PR #146, opened 2014-02-05
Update Coop.ch.xml

J0WI:
rebased

### DIFF
--- a/src/chrome/content/rules/Coop.ch.xml
+++ b/src/chrome/content/rules/Coop.ch.xml
@@ -1,15 +1,67 @@
-<ruleset name="Coop.ch" platform="mixedcontent">
-        <target host="coop.ch" />
-        <target host="www.coop.ch" />
-        <target host="bankcoop.ch" />
-        <target host="www.bankcoop.ch" />
-        <target host="onlinebanking.bankcoop.ch" />
-        <target host="supercard.ch" />
-        <target host="www.supercard.ch" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Coop.ch.xml
```diff
@@ -1,15 +1,67 @@
-<ruleset name="Coop.ch" platform="mixedcontent">
-        <target host="coop.ch" />
-        <target host="www.coop.ch" />
-        <target host="bankcoop.ch" />
-        <target host="www.bankcoop.ch" />
-        <target host="onlinebanking.bankcoop.ch" />
-        <target host="supercard.ch" />
-        <target host="www.supercard.ch" />
+<ruleset name="Coop">
```
nelegalno: "**HTTPS Everywhere WARN: Error: found duplicate rule name Coop**" error is displayed in the Error Console.

One of this two file names must be changed to something unique (currently they use ruleset name="Coop"):
[src/chrome/content/rules/Coop.ch.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Coop.ch.xml)
[src/chrome/content/rules/Coop.ch-falsemixed.xml ](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Coop.ch-falsemixed.xml)

J0WI: Thanks, filled  #4693

Issue #49, opened 2014-01-02
Causes issue with Access-Control-Allow-Origin:*


danielchatfield:
Browser: chrome
Example (~~may be fixed soon though~~ Fixed by using https directly): http://yeoman.io/community-generators.html

Page tries to load http://yeoman-generator-list.herokuapp.com/ which results in: `XMLHttpRequest cannot load http://yeoman-generator-list.herokuapp.com/. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://yeoman.io' is therefore not allowed access.`.

If the page requests the https version directly then it works.

diracdeltas:
anne van kesteren says this Firefox bug might be relevant: https://bugzilla.mozilla.org/show_bug.cgi?id=881830

diracdeltas:
Bug in chrome: https://code.google.com/p/chromium/issues/detail?id=387198

diracdeltas:
https://bugzilla.mozilla.org/show_bug.cgi?id=881830#c12 seems like the right explanation here. When a subresource is redirected to HTTPS by HTTPS-E or HSTS, the browser says, "I have no CORS information for this new URL, so just block it."

Did a quick test and this seems to be fixed in Chrome but not in Firefox. To repro, make sure you have the Heroku rule enabled in the dev version of HTTPS Everywhere Firefox, then do an XHR to `http://yeoman-generator-list.herokuapp.com/` in the context of an HTTP page at a different origin. In Firefox 36, the XHR silently fails. OTOH, `https://yeoman-generator-list.herokuapp.com/` loads fine.

SimonSapin:
That Firefox bug is [marked as fixed in Firefox 38](https://bugzilla.mozilla.org/show_bug.cgi?id=881830#c40), but I can reproduce #1365 (which I think is a case of this bug) on Firefox Nightly 40.0a1 (2015-04-04)

AnAkIn1:
I still have the same issue with Firefox 39.

AnAkIn1:
Is there any progress on this? I can't enable https everywhere because I have this issue on one of the websites I often visit.

jsha:
No progress so far, I'm afraid. In the meantime, if you would like to use HTTPS Everywhere, you can disable the rules for the specific site that has problems using the icon menu.

TheNavigat:
https://github.com/EFForg/https-everywhere/issues/2637

TheNavigat:
[...]

PR #158, opened 2014-02-13
Fix OpenStreetMap

yardenac:
This fixes an old ruleset that's been disabled since last summer, and re-enables it.

Not sure if that's appropriate during a ruleset freeze. It definitely needs testing, as OSM is embedded in many many sites, but they've made progress on deploying SSL and it seems to work just fine now.

Your call!

### DIFF
--- a/src/chrome/content/rules/OpenStreetMap.xml
+++ b/src/chrome/content/rules/OpenStreetMap.xml
@@ -9,7 +9,7 @@
 	<rule from="^http://tile\.openstreetmap\.org/"
 		to="https://a.tile.openstreetmap.org/"/>
 
-	<rule from="^http://(blog|help|lists|piwik|taginfo|[abc]\.tile|trac|wiki)\.openstreetmap\.org/"
+	<rule from="^http://(blog|help|lists|nominatim|piwik|taginfo|[abc]\.tile|trac|wiki)\.openstreetmap\.org/"
 		to="https://$1.openstreetmap.org/"/>
 
 </ruleset>


drzraf:
did you tried recently openlayers/openstreetmap with httpse 5-dev3 ?
I get pink tiles (disable httpse and the map tiles appear again).
Sample: http://openlayers.org/two/

jsha:
Thanks for spotting! I'm disabling the rule: https://github.com/EFForg/https-everywhere/pull/1221

Firefishy:
OpenStreetMap.org sysadmin here. Please do not conflate the OpenStreetMap.org website with OpenLayers.org map library.

OpenStreetMap.org works well using HTTPS and all [abc].tile.openstreetmap.org caches are SSL enabled.

jsha:
Hi @Firefishy! Thanks for the great work on OSM, and for making it work well with HTTPS.

Would you mind taking a look at the ruleset at #1221 and the website @drzraf linked to help me figure out what might be triggering the pink tiles problem? I've verified that disabling the ruleset resolves the problem, but obviously I'd like to fix it instead.

jmorahan:
I think this is related to #49. OpenLayers requests the images as eg. ``. Edit out the crossorigin attribute in developer tools, and the tiles appear.

KommX:
So, what's the status here?
I would like to see the rule enabled again.

koops76:
@Firefishy Can you fix the mismatched certificate at https://shop.openstreetmap.org? I am currently fixing the OpenStreetMap ruleset. I'll exclude shop.openstreetmap.org for now since it just redirects to http://wiki.openstreetmap.org/wiki/Merchandise.

koops76:
@Firefishy Also please fix the same issue with https://maps.openstreetmap.org/.

PR #273, opened 2014-05-07
[Github] separate rule for github pages

yardenac:
Some github pages are not designed for https due to mixed content, and aren't motivated to fix because their primarily purpose is to CNAME a vanity domain (which can't support SSL). This makes it easier to work with those pages - especially for dev & testing - without disabling all of Github

### DIFF
--- a/src/chrome/content/rules/Github-Pages.xml
+++ b/src/chrome/content/rules/Github-Pages.xml
@@ -1,0 +1,11 @@
+<!--
+	For other GitHub coverage, see Github.xml.
+-->
+<ruleset name="GitHub Pages">
+
+	<target host="*.github.io" />
+
+	<rule from="^http://([^/@:\.]+)\.github\.io/"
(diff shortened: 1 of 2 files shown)


cben:
Turns out Github Pages https is not end-to-end secure.  Got this from Github's support:

> While HTTPS requests may appear to work, our CDN provider is adding and removing the encryption at their end, and then the request is transmitted over the open internet from our CDN provider to our GitHub Pages infrastructure, creating the appearance of trustability.
> 
> This is why we do not yet officially support HTTPS for GitHub Pages.

[more discussion at https://konklone.com/post/github-pages-now-sorta-supports-https-so-use-it]

This raises question if benefit of https at least part if the way outweighs harm from illusion of security?
I suppose yes for HTTPS Everywhere users, that are generally looking to get as much security as possible from sites that don't necessary try to be secure.

PR #515, opened 2014-08-29
Speed up builds.

jsha:
Only build / validate rulesets when they have changed.

Also break out validation steps into their own script, shared by
makecrx/makexpi.


jsha:
Updated so this now merges cleanly and passes tests.

pde:
This seems to no longer cleanly apply because of changes to <tt>--fast</tt> that were a (probably unintended) component of f415c0ee38ae4d3959dfe0cd06387a24036504f9.  Please figure out whether you want to kill <tt>--fast</tt> altogether now.  If so, you should also remove all the places that test for it; there are a few stragglers!

jsha:
Good point! Fixed.

jsha:
Re-review @pde?

diracdeltas:
Works fine for me on OSX, but @pde should sign off on this since he knows the build scripts much better.

jsha:
I'm going to go ahead and merge this on the strength of @pde's earlier review, since I think he's too busy to do a second one.

PR #603, opened 2014-09-22
Add HTTP Nowhere mode to Chromium

diracdeltas:
Adds mode for blocking HTTP requests entirely after HTTPS Everywhere
rewrites. Switches UI button from PageAction to BrowserAction.


semenko:
Ok -- taking a look at this now. @jsha should we be merging this into 4.0 vs master?

First notes: HTTPS Nowhere mode doesn't seem to be working 100% for me. Some pages (Chrome prerender replaced?) seem to bypass this mode. Will update / patch.

jsha:
Let's merge to master, since I'll be promoting master to stable sometime this month.

chromium/manifest.json
```diff
@@ -35,6 +37,7 @@
         "webRequestBlocking", 
         "tabs", 
         "cookies", 
+        "storage",
```
semenko: Just being explicit, since this is a dangerous file to change: 'storage' does *not* cause additional permissions warnings, per:
https://developer.chrome.com/extensions/permission_warnings

jsha:
How's this going?

semenko:
Fell to the back burner for a while -- will take a look soon

diracdeltas:
Let me know if you need help on getting this merged!

Also, note that this introduces use of `chrome.storage.sync` for extension prefs. I think that's a good move, since it can be used for #1022.

semenko:
Sorry, still on the back burner until a few more things are hammered out for a manuscript.

Last I checked, this wasn't working as expected -- I *think* prerendered tabs were ignored, or something similar.

(At the least, I'll break out the UX change into a separate PR)

semenko:
Ok -- tracked down the bug in this. Will rebase and probably open a new PR (since i'm hesitant to ever use git --force ...)

semenko:
rebased & opened #1295

PR #750, opened 2014-11-19
added rules for voiceofsandiego.org

jeffhammett:
Fairly new to this sort of thing, copied syntax from another rule, but it seems to work in my browser.


julianladisch:
I propose to delete this pull request as both www.voiceofsandiego.org and voiceofsandiego.org redirect vom https to http.

PR #779, opened 2014-12-02
use Firefox latest stable(v35.0, CI build passed)

PeterDaveHello:


jsha:
There's a bug in Firefox 34 that causes a spurious test failure when using the addon-sdk. It's fixed in 35.

PeterDaveHello:
Will test with v35 betas, thanks @jsha.

jsha:
Actually I may have spoken too soon. It is definitely fixed in 37. I just tested 35 and got a failure. Haven't tried 36.

PeterDaveHello:
But there is no 36, right?

jsha:
Aha, I did not know that!

PeterDaveHello:
Passed in v35.0!

PR #847, opened 2014-12-29
Update Yemeksepeti.xml

yymmk:
Added exclusions


reedy:
Are the Sanalmarket commits supposed to be part of this?

jsha:

Thanks,
Jacob

PR #863, opened 2015-01-03
add governikus.com and governikus.de

hauke:


reedy:
It looks like from a bit of testing that http://www.governikus.de and http://www.bos-bremen.de always gets redirected to https://www.governikus.com (at least in my browser). With a /de suffix, but that's the same for either site, so I'm not sure if it's worth adding that too.

Either way, I think we should just move any of those straight over to the HTTP version...

jsha:
Closing infavor of #2132.

PR #865, opened 2015-01-04
Don't exclude bugs.mysql.com

dveeden:
The MySQL Bugs website now also has SSL:
https://bugs.mysql.com/bug.php?id=73457

But there is no http->https redirect and the http url is still used in
email notifications:
https://bugs.mysql.com/bug.php?id=75367


reedy:
WFM

PR #868, opened 2015-01-05
Rule for strasbourg.eu

Rudloff:


reedy:
I see a load of mixed content warnings for this site...

All from http://media.strasbourg.eu/ but HTTPS doesn't seem to work on that subdomain

Rudloff:
Should I add ```platform="mixedcontent"``` to the ruleset?
(The warning seems to only be triggered by images, so the website still works correctly.)

reedy:
Yeah, I think so. Plus a comment stating why it's there like other rule sets like https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/10_Minute_Mail.com.xml

Rudloff:
I added a comment. Is it OK like this?
I also added a rule for portail.strasbourg.eu.

src/chrome/content/rules/Strasbourg.eu.xml
```diff
@@ -0,0 +1,15 @@
...
+
+    * Unsecurable
+-->
+<ruleset name="Strasbourg.eu" platform="mixedcontent">
+  <target host="strasbourg.eu" />
+  <target host="www.strasbourg.eu" />
+  <target host="portail.strasbourg.eu" />
+
+  <rule from="^http://(www\.)?strasbourg\.eu/" to="https://www.strasbourg.eu/" />
+  <rule from="^http://portail\.strasbourg\.eu/" to="https://portail.strasbourg.eu/" />
```
reedy: The first rule is fine, but this second one can just become the new style simpler http -> https rewrite

Rudloff:
I have updated the rule.

PR #869, opened 2015-01-05
Add OEIS.org rule

PiRSquared17:
Add basic rule for OEIS.org.


reedy:
Looks like the SSL cert is bad for https://www.oeis.org/

So yay for this :)

PR #871, opened 2015-01-05
[Glerups] Added ruleset.

jpds:


reedy:
Seems https://glerups.dk/ has a redirect loop. So I guess this prevents it :)

PR #872, opened 2015-01-05
toggleHttpNowhere function appeared twice in code

emanueLczirai:
Unless I'm missing something, this function shouldn't appear twice. Checked both occurrences to be exactly the same(with diff). This removes one of them.


reedy:
Ping @diracdeltas

https://github.com/EFForg/https-everywhere/commit/599f270ad36dff4b0857e7858a68be97e618f659

https://github.com/EFForg/https-everywhere/commit/07cdc4675ceddc7e4d3e0655b15ff7d92f615e60


Looks like it's somewhat of a merge conflict

PR #879, opened 2015-01-07
Create Gamesrocket.de, Kinguin.net, Whosgamingnow.net, Wdc.com, Meo.pt, Yoyogames.com

john-kk:


src/chrome/content/rules/Wdc.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Wdc.com">
+  <target host="wdc.com" />
+  <target host="www.wdc.com" />
+
+  <rule from="^http://(www\.)?wdc\.com/" to="https://www.wdc.com/" />
```
reedy: This server could not prove that it is www.wdc.com; its security certificate is from a248.e.akamai.net. This may be caused by a misconfiguration or an attacker intercepting your connection.

semenko:
Looks like this PR has some outstanding edits -- and it hasn't been updated in a while. I'm going to close this for now, but if you'd like it merged, feel free to re-open. Thanks! :)

PR #883, opened 2015-01-08
adding deutsche-rentenversicherung.de

julianladisch:
Germany's social security insurance authority


src/chrome/content/rules/Deutsche-Rentenversicherung.de.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Deutsche-Rentenversicherung.de">
+	<target host="www.deutsche-rentenversicherung.de" />
+	<target host=    "deutsche-rentenversicherung.de" />
+	<target host="www.deutsche-rentenversicherung-bund.de" />
+	<target host=    "deutsche-rentenversicherung-bund.de" />
+	<target host="www.drv-bund.de" />
+	<target host=    "drv-bund.de" />
+
+	<rule from="^https?://(?:www\.)?(?:deutsche-rentenversicherung\.de|deutsche-rentenversicherung-bund\.de|drv-bund\.de)/"
```
reedy: The \.de is common to all 3 capture groups... You might aswell just put it after

"^https?://(?:www\.)?(?:deutsche-rentenversicherung|deutsche-rentenversicherung-bund|drv-bund)\.de/"

But you can simplify it further..

"^https?://(?:www\.)?(?:deutsche-rentenversicherung(-bund)?|drv-bund)\.de/"

julianladisch: JavaScript's RegExp compiles the `from` pattern.  Therefore no optimization is needed beforehand.  Instead rules files can be written in the most comprehensible way for humans.

reedy: Maybe so, in which case (-bund)? is probably premature optimisation. But removing the triple repeated \.de seems sensible at least. It's still readable.

Computer optimisation requires CPU time too

julianladisch: Building the deterministic finite automaton (DFA) automatically does the optimization, I cannot measure a time difference between the three pattern compilations.

jsha:
Please simplify these rules as:

```
<rule from="^http:"
           to="https:" />
```

donced:
@julianladisch
Enhancement of this PR #2132.

PR #884, opened 2015-01-08
Elster.de: Germany's tax and revenue authorities

julianladisch:


jsha:
Please simplify these rules as:

```
<rule from="^http:"
           to="https:" />
```

src/chrome/content/rules/Elster.de.xml
```diff
@@ -0,0 +1,16 @@
...
+	<target host="www.elsteronline.de" />
+	<target host=    "elsteronline.de" />
+
+	<test url="https://elster.de/" />
+	<test url="https://elsteronline.de/" />
+
+	<rule from="^https?://elster\.de/"
+		to="https://www.elster.de/" />
+	<rule from="^https?://elsteronline\.de/"
+		to="https://www.elsteronline.de/" />
```
reedy: These 2 rules above should be unneeded with the rule addition below

julianladisch: Note the addition of www.

reedy: Mmm. The generic new style rewrite doesn't do anything for these rules

julianladisch: http://elster.de/ needs rule in line 10 to add www.
https://elster.de/ needs rule in line 10 to add www.
http://www.elster.de/ needs rule in line 14.
http://elsteronline.de/ needs rule in line 12 to add www.
https://elsteronline.de/ needs rule in line 12 to add www.
http://www.elsteronline.de/ needs rule in line 14.

Please explain which rule you think is not needed and why.

donced: I guess lines 10-13 (and line 7-8 as well) are not required because https://elster.de/ and https://elsteronline.de/ display just fine. Some website ([example](https://pbebank.com/)) requires url rewrite (to add www) because without it, the link is either unreachable or SSL error.

julianladisch:
All comments have been resolved, the ruleset may be merged.

PR #885, opened 2015-01-09
Unbreak SSL Observatory.

jsha:
md5Fingerprint is no longer available for certs in Firefox, so switch to SHA1 for private cert lookup. Note: we originally used md5+sha1 on the theory that if an adversary is capable of a second preimage attack on one hash they may not be able to do both. Sha256 is newly available so if we still find it worthwhile we can do the same trick with sha256+sha1, and regenerate the contents of Root-CAs.js. But for now we should just fix the code because on current Firefox all certs are being considered 'private', which blocks submission to the Observatory.

cc @cooperq @pde


cooperq:
Seems good, I will merge it now in the interest of unbreaking things but I do think that we should switch to SHA2 asap. @schoen has some thoughts about whether it's worthwhile to combine SHA2 and SHA1 or not I think.

pde:
We chose to use the concatenation of md5 and sha1 because it was (1) decently secure and (2) more readily computable in client code.  Though we could probably have grabbed a sha256 library.

Unfortunately the serverside code may check that the submitted certs have the same submitted fp as the server recomputes for them.  So we'll need to set CertificateParser.skipfpcheck to True on the server if we want this to work.

pde:
Alternatively to the serverside kludge, we could [use an md5 implementation](http://www.myersdaily.org/joseph/javascript/md5.js).

pde:
https://bugzilla.mozilla.org/show_bug.cgi?id=622332 
https://hg.mozilla.org/mozilla-central/rev/f2a5edce0a57

This broke when FF32 released in September, so presumably everyone not running TBB / ESR has been not submitting to the Observatory since then :(

pde:
I don't think it's wise to ship a client that just uses sha1.  It would create two populations of incompatible client submissions, meaning:

* we'd have to either turn skipfpcheck off forever, or have two different URL parameters for the fingerprint submissions (fp=MD5+SHA1&fp2=SHA1)
* features like [this one](https://github.com/EFForg/https-everywhere/pull/276) would have to exist in two versions for the two client populations

So I've been trying to get md5 fingerprints going in pure JS.  They don't match, yet...

jsha:
@pde: Aha, I missed the fact that this is used as part of the certificate submission as well as for checking whether root CAs are public. What's the reason behind doing fp check?

Assuming we keep the fp check, I think the right thing to do is to add a new URL parameter.

It's okay if the whitelist updating breaks in older clients, since it's just a load reduction measure and the population of those clients is small.

PR #889, opened 2015-01-10
Added digikey.de to DigiKey ruleset

ulikoehler:
This pull request modifies the Digikey ruleset so that not only `digikey.com` but also `digikey.de` is included in it.

### DIFF
--- a/src/chrome/content/rules/Digi-Key.xml
+++ b/src/chrome/content/rules/Digi-Key.xml
@@ -28,6 +28,9 @@ Fetch error: http://digikey.com/ => https://www.digikey.com/: Redirect for 'http
 
 	<target host="digikey.com" />
 	<target host="*.digikey.com" />
+
+	<target host="digikey.de" />
+	<target host="*.digikey.de" />
 		<!--
 			These paths 404:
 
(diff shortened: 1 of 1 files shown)


jsha:

Thanks,
Jacob

ulikoehler:
@jsha I hope I did it right this time ;-)

The ruleset actually contained some old exclusion rules which are no longer required (particularly the DigiKey catalog) and needed to be replaced with a new (simpler) exclusion rule. I will write to DigiKey regarding the remaining 404s.

jsha:
You're close! A couple more things:

The previous version of the rule didn't rewrite all digikey subdomains, only the ones matched by:

> ^http://(digikeytest|dkc\d|form|media|ordering(?:test)?|punchout(?:test)?|www)\.digikey\.com/

Did you intend to expand the ruleset to cover all subdomains? If not, you should replace the "*.digikey.com" in target hosts with a list of the subdomains you want (probably the ones from the regex above). If you do want to cover all subdomains, update the comment.

Also, in the time since you submitted this pull request, I ran the https-everywhere-checker and auto-disabled the rule. You can see in the ruleset tag it has `default_off='failed ruleset test'`. In the comment above that, it says:

```
Disabled by https-everywhere-checker because:
Fetch error: http://digikey.com/ => https://www.digikey.com/: Redirect for 'http://digikey.com/' missing Location
```

I looked into that error, and it appears to be spurious (I filed #1079) to fix it. You should delete the error and the default_off attribute to re-enable this rule.

Thanks,
Jacob

semenko:
Closed per #2242

PR #893, opened 2015-01-13
StackExchange: duplicated meta beta rule removed, added support to the new localized SO sites and to their metas

HorvathAkosPeter:
* remove duplicated meta beta rules
* added rules to the new localized stackoverflow sites, and to their metas


HorvathAkosPeter:
Could I ask, is there any problem with the patch, or it is only matter of time to be merged?

jsha:
Hi,

Sorry for the delay - it can take a while for us to review and merge a
patch sometimes, since we receive so many. Thanks for your contribution!
We'll review it soon.

HorvathAkosPeter:
Ok, thank you!

PR #894, opened 2015-01-13
Rule for contribuez.cnnumerique.fr

Rudloff:
Hello,

Here is a rule for contribuez.cnnumerique.fr.

Regards,


semenko:
LGTM -- thanks!

PR #895, opened 2015-01-14
Forocoches ruleset

i-rme:
Forocoches.com is the top 1 Spanish Forum with more than 2 million unique visitors per day.

Is the top 3601 most popular website worldwide (http://www.alexa.com/siteinfo/forocoches.com) and the top 60 in Spain.


i-rme:
They are 4 commits because the first 3 failed the Travis build test because I was using the wrong file extension (yml vs xml).

That was my fault, the final commit is the right one.

semenko:
Awesome -- LGTM

PR #897, opened 2015-01-15
Garfield.com now redirects back to HTTP

q--:
Licensee.Garfield.com is still working, though.


semenko:
NAK. The bare name seems to work for me: https://garfield.com/

q--:
Weird, now it's working for me as well. Maybe they were testing different configurations; it was clearly redirecting me back when I tested it.

PR #898, opened 2015-01-15
Some Latvian government websites

kristapsk:


src/chrome/content/rules/Mod.gov.lv.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="Mod.gov.lv">
+  <target host="www.mod.gov.lv" />
+
+  <rule from="^http://www\.mod\.gov\.lv/" to="https://www.mod.gov.lv/" />
```
reedy: https://www.mod.gov.lv/ doesn't seem to work for me.

Also, you should use the new ruleset format too, especially for a simple rewrite like this

kristapsk:
Just checked, yes it does not work now. Did work back then. I think you can simply reject this pull request, I will look at this at some moment and will resubmit a new one.

PR #901, opened 2015-01-16
Reinstate md5

jsha:
This reverts #885 and, instead, does the MD5 calculation using nsICryptoHash. I found that rewiring the SSL Observatory submission and whitelisting code to use pure SHA256 was too big a task to get done in the appropriate timeframe for fixing this bug. So now we're back to MD5+SHA1.

cc @pde @cooperq for review. I'd like to try and put this in a release tomorrow or Monday.


cooperq:
Looks fine to me.

PR #906, opened 2015-01-17
Improvements in response to feedback from AMO-editors.

jsha:
- Use XPCOMUtils.generateQI instead of our own hand-rolled one.
- Stringify rule.id.

CC @cooperq for code review. @pde I plan on doing a release next Tuesday, but if you decide to do one before then, include this branch.

### DIFF
--- a/src/chrome/content/code/ApplicableList.js
+++ b/src/chrome/content/code/ApplicableList.js
@@ -212,7 +212,7 @@ ApplicableList.prototype = {
 
   add_command: function(rule) {
       var command = this.document.createElement("command");
-      command.setAttribute('id', rule.id+'-command');
+      command.setAttribute('id', JSON.stringify(rule.id)+'-command');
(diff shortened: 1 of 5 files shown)


src/chrome/content/code/ApplicableList.js
```diff
@@ -212,13 +212,9 @@ ApplicableList.prototype = {
 
   add_command: function(rule) {
       var command = this.document.createElement("command");
-      if (typeof rule.id !== 'number') {
-        this.log(WARN, "Rule has invalid id " + rule.id);
-        return;
-      }
       command.setAttribute('id', rule.id+'-command');
```
cooperq: I suspect that they would ding you for this too. You may want to json stringify this and the next line as well.

cooperq:
Looks good other than my one note. Merge whenever you feel good about it.

PR #911, opened 2015-01-18
Added Mentalhealth.asn.au.xml

StevenRoddis:
MCB blocks google web fonts, no big deal.

### DIFF
--- a/src/chrome/content/rules/Mentalhealth.asn.au.xml
+++ b/src/chrome/content/rules/Mentalhealth.asn.au.xml
@@ -7,11 +7,15 @@
 			- $
 
 		Doesn't Work:
+			nil
+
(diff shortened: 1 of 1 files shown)


reedy:
https://www.mentalhealth.asn.au/templates/t3_blank/js/jquery.sticky.js is also showing in the console as 404, but doesn't work under HTTP either

https://www.mentalhealth.asn.au/ works fine

https://mentalhealth.asn.au/ redirects to http://www.mentalhealth.asn.au/

I think we should force the www in the rewrite?

StevenRoddis:
>I think we should force the www in the rewrite?
Agreed, will fix and squash.

https://www.mentalhealth.asn.au/templates/t3_blank/js/jquery.sticky.js 404s for me http or https too. So no problem there.

PR #912, opened 2015-01-18
Bing: rule for ts?.mm.bing.net

julianladisch:
Needed for eg. https://ixquick.com/eng/pics.html


jsha:

Thanks,
Jacob

PR #916, opened 2015-01-18
Added a number of .dk domains

fuglede:
Out of the Watch Medier ones, only a few work completely well (as in, everything but a few trackers work), but all are included to allow for the interconnection between the various sites.

Since the first commit, also added a few assorted .dk sites.


fuglede:
Split into seperate requests.

PR #917, opened 2015-01-18
Turn on Disqus rule by default and remove unreliable flag.

### LINKED ISSUE #900: Make Disqus not experimental
I'm not sure how we can do this, but I'd say that the Disqus support is solid and I've been helping maintain the list personally to make sure it works well. There's no need to have it still flagged as experimental.

jsha:
cc @mattrobenolt for review.

Fixes #900.


mattrobenolt:
:+1: Thanks. :) Will be happy to see this.

PR #925, opened 2015-01-20
Update Disqus.com.xml

mattrobenolt:
We now have https on {blog,community}.disqus.com

This should now cover every *.disqus.com and disqus.com property. :sparkling_heart:


jsha:
:sparkling_heart: :lock:

PR #930, opened 2015-01-21
Revert "Update Disqus.com.xml"

mattrobenolt:
As it turns out, Tumblr shut is down for doing this. I had to revert what we did to get https in front of our blogs.

I just routed traffic through our load balancers, which was terminating SSL, then passing the traffic back along to them and it worked fine. They apparently didn't like that and our blogs started 404ing today. Their service requires validating that the DNS record is pointed to `domains.tumblr.com`, and if it's not, they apparently just turn it off.

Any help from Tumblr here would be cool if possible. Fortunately, we're moving this away from Tumblr probably this quarter.

:(

### DIFF
--- a/src/chrome/content/rules/Disqus.com.xml
+++ b/src/chrome/content/rules/Disqus.com.xml
@@ -4,13 +4,34 @@
 
 	Nonfunctional domains:
 
-		- none
+		disqus.com subdomains:
+
+			- blog		(tumblr)
+			- community     (tumblr)
+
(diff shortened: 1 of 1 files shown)


jsha:
That's a bummer. I had similar problems when I was working on HTTPS deployment at Twitter. The last stubborn domain that doesn't support HTTPS was status.twitter.com, a Tumblr blog. I tried for years to get it moved. :-(

PR #934, opened 2015-01-22
Remove tls.version.min setting now that it is incorporated in Firefox

jsha:
cc @cooperq for review.


jsha:
Self-merging in order to get this into today's release. @cooperq, please review post-hoc.

cooperq:
This would still leave vulnerable someone who was installing httpse for the first time in an old version of firefox. But thats probably not a very large number of people.

PR #171, opened 2014-02-22
Hospital and clinic rulesets

adamjones443:
Rulesets for some hospital and clinic websites.

PR #937, opened 2015-01-22
[ClevelandClinic] adding several rules

julianladisch:
Fixes "Cleveland Clinic Ruleset" from "Hospital and clinic rulesets" #171.


jsha:

Thanks,
Jacob

PR #947, opened 2015-01-24
Fixed Reddit-subdomains with two-character country-codes (e.g. en.reddit.com)

mrshell:
Somehow, domains like "i.reddit.com" where working, but "en.reddit.com" not. I don't understand, why [a-z]{1,2} didn't work, because I've tested the rule with en.reddit.com back then when I've modified the rule to cover more subdomains. Changed to [a-z][a-z]? now and it works locally on my machine.

### DIFF
--- a/src/chrome/content/rules/Reddit.xml
+++ b/src/chrome/content/rules/Reddit.xml
@@ -102,7 +102,7 @@
 	<securecookie host="^\.redditmedia\.com$" name="^__cfduid$" />
 
 
-	<rule from="^http://((?:[a-z][a-z]?|[a-z]{2}-[a-z]{2}|blog|pay|redditama|ssl|static|www)\.)?reddit\.com/"
+	<rule from="^http://((?:[a-z]{1,2}|[a-z]{2}-[a-z]{2}|blog|pay|redditama|ssl|static|www)\.)?reddit\.com/"
 		to="https://$1reddit.com/" />
 
 	<rule from="^http://(?:www\.)?np\.reddit\.com/"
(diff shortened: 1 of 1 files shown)


jsha:
Can you double-check that en.reddit.com doesn't get rewritten for you? I just tested it with latest master and it works for me. What revision / browser are you on?

mrshell:
I've just disabled all other add-on's except HTTPS-Everywhere. I'm on version 4.0.3 of HTTPS-Everywhere and version 35.0 of Firefox. All on a Linux machine. With the rule included in the addon, en.reddit.com doesn't get rewritten to use https. But for example i.reddit.com or en-us.reddit.com works just fine.

jsha:
Aha - the ruleset is different between 4.0.3 and master. Try the latest 5.0development.2, it should have the latest version of the ruleset, which should rewrite en.reddit.com.

mrshell:
Oh! I think, I've just noticed my mistake. The "View XML source" button in the enable/disable rules GUI of the installed addon actually doesn't display the rule for the current version, but of the HEAD within the source tree. So I've viewed the HEAD version of the Reddit rule instead of the 4.0.3 one. I did just download the 4.0.3 tarball and there is the old rule still in it.

I guess then, my original Reddit regexp (the one currently in HEAD) actually works. That explained why I was sure to have been checking the rule before commiting it and it still didn't seem to work now.

mrshell:
I guess this pull request is not necessary...

jsha:
Thanks for looking into it! Yes, the 'view XML source' tool can be a little buggy. In particular if it can't find the exact revision, it will quietly fall back to HEAD.

For future testing of latest rules, I recommend `./test.sh --justrun`, which will start Firefox with a temporary profile containing only the latest version of HTTPS Everywhere from your repo, with all the latest rules.

PR #948, opened 2015-01-24
Added Jabber.dk.xml

fuglede:
Simple site, simple rule.


fuglede:

PR #949, opened 2015-01-24
Added Alternativet.dk.xml

fuglede:
Marked partial as a few ressources are fetched over HTTP. Still, the site itself works fine.


reedy:
Test failure is unrelated, it's in WeeklyPlus.xml which is now fixed

reedy:
But the site loads a lot of mixed content when visited in HTTPS

fuglede:
A few stylesheets and a Google API font, right? I just noticed now that this actually causes the four gray links at the top to become hidden. I notified the website devs about this, so hopefully they will fix the hardcoded references.

default_off="mixedcontent" until then?

fuglede:

PR #951, opened 2015-01-24
Added Altinget.dk.xml

fuglede:


fuglede:

reedy:
Re-running travis before merge

PR #952, opened 2015-01-24
Added Mathjobs.org.xml

fuglede:


jsha:
From the tests:
> Invalid attribute deafault_off for element ruleset

I'd recommend running `./install-dev-dependencies.sh`. In addition to installing all the prerequisites to build and test HTTPS Everywhere, it will set up a pre-push hook that runs the tests, so you can catch these sorts of errors early on.

jsha:
Whoops, my mistake. I missed that this was the failure in the WeeklyPlus rule that caused so many builds to fail.

PR #954, opened 2015-01-24
Added Rejsekort.dk.xml

fuglede:
Included with default_off as the entire site works, except for its contact forms. However, if you ever visit the site, it's probably to complain about something, so the contact forms are essential.

### DIFF
--- a/src/chrome/content/rules/Rejsekort.dk.xml
+++ b/src/chrome/content/rules/Rejsekort.dk.xml
@@ -1,0 +1,13 @@
+<!--
+	Not covered here:
+		doubleclick.net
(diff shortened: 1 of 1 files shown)


jsha:
I'd prefer not to merge new rulesets that are default_off from the beginning. I think that is more useful for rules that worked once, but stopped working, in order to preserve some history and make it easy to fix them.

fuglede:
@jsha: Fair enough; I figured this one might be useful to users who actively activate rules as most of the site works fine, and that moreover it might save some time if someone else decides to write a rule for it later on (twice I've found myself beginning to write a rule, only to realize that it already exists).

PR #956, opened 2015-01-24
Added Evm.dk.xml

fuglede:


jsha:
You don't have a target host that covers the non-www version.

fuglede:
Fixed.

PR #957, opened 2015-01-24
Added Jobzonen.xml and ClickDimensions.xml

fuglede:


fuglede:
Adapted to the new ruleset style

PR #958, opened 2015-01-24
Added rules relating to Watch Medier

fuglede:
Out of the Watch Medier ones, only a few work completely well on the mixedcontent browsers (as in, everything but a few trackers work), but all are included to allow for the interconnection between the various sites, of which there is plenty.


fuglede:
Adapted to new rule style

PR #962, opened 2015-01-24
Changed from kickass.to to kickass.so

fuglede:


jsha:
@fuglede Thanks for the contributions! Each of your recent pull requests fails the Travis build. I think this may be because we recently had a test breakage committed on master. If you pull the latest master into each of your branches and push them again, the Travis build should pass.

fuglede:
Heh, alright, will have a look at it in a bit; I checked the build with the modified WeeklyPlus before submitting the pull requests, so hopefully they work.

jsha:
Actually, it looks like if I tell Travis to re-do all those builds, it will Just Work without you having to re-push. I'll do that now.

What's WeeklyPlus?

fuglede:
Great.

Sorry, WeeklyPlus was the ruleset that caused the build to fail.

src/chrome/content/rules/KickassTorrents.xml
```diff
@@ -2,25 +2,29 @@
...
+  <target host="kickass.so" />
+  <target host="www.kickass.so" />
   <target host="kickass.to" />
   <target host="www.kickass.to" />
   <target host="kickasstorrents.com" />
   <target host="www.kickasstorrents.com" />
 
+  <securecookie host="^kickass\.so$" name=".+" />
   <securecookie host="^kickass\.to$" name=".+" />
+  <securecookie host="^\.kat\.ph$" name=".*"/>
```
guiweber: It looks like the cookies are never set under the kat.ph or kickass.to domains anymore, even if you access the site from those addresses. I think these two lines can be removed since they are now always set under kickass.so.

fuglede: Fixed

guiweber:
It looks like kickass.so was taken down and that they are going back to kickass.to for now.

fuglede:
Alright, I'll close this PR for now.

PR #964, opened 2015-01-24
Updated translations from Transifex

fuglede:
Most translations were already pretty good but some rather typical mistakes were made in recent additions


fuglede:
Contained in #966.

Issue #961, opened 2015-01-24
Localisation: Hardcoded string in toolbar_button.xul


fuglede:
The string "Block all HTTP requests" is hardcoded in toolbar_button.xul:46.

StevenRoddis:
Thanks! I also noticed line 35 has "HTTPS" hardcoded.

fuglede:
Sure, but it sounds to me like it would be a lot faster for a Transifex maintainer, if I understand [jsha's comment here](https://github.com/EFForg/https-everywhere/pull/966#discussion_r23559337); the other ones I added to the individual .dtd's manuall.

By the way, is that HTTPS label actually shown anywhere?

jsha:
Yes, it's probably easier for me to just copy over all the latest strings before each release.

The HTTPS label is displayed if you customize your toolbar and drag the HTTPS Everywhere icon off of it and into the area of available icons. I think the proper fix is to change the string to be a reference to the "HTTPS Everywhere" string.

fuglede:
Ah, that makes sense. I changed it to the extension name string in the pull request.

Issue #960, opened 2015-01-24
Localisation: Hardcoded string in about.xul


fuglede:
The string "or" in about.xul:63 is hardcoded, thus not a priori translatable.

PR #966, opened 2015-01-25
Patch hardcoded strings and updated Danish locale

fuglede:
Added a few new strings to https-everywhere.dtd, cf. issue #960 and #961. I feel like there was probably an easier way to do this by using Transifex?

Moreover, updated the Danish localisation. Most of the strings were already fine but a few recent ones contained some common misspellings.

### DIFF
--- a/src/chrome/locale/ar/https-everywhere.dtd
+++ b/src/chrome/locale/ar/https-everywhere.dtd
@@ -8,12 +8,14 @@
 <!ENTITY https-everywhere.about.contribute  "إن أعجبتك إضافة HTTPS Everywhere، فكر بـ">
 <!ENTITY https-everywhere.about.donate_tor "التبرع لتور">
 <!ENTITY https-everywhere.about.tor_lang_code "en">
+<!ENTITY https-everywhere.about.or "or">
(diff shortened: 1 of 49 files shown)


src/chrome/locale/ar/https-everywhere.dtd
```diff
@@ -8,12 +8,14 @@
 <!ENTITY https-everywhere.about.contribute  "إن أعجبتك إضافة HTTPS Everywhere، فكر بـ">
 <!ENTITY https-everywhere.about.donate_tor "التبرع لتور">
 <!ENTITY https-everywhere.about.tor_lang_code "en">
+<!ENTITY https-everywhere.about.or "or">
```
jsha: I believe you can leave these entries absent until they are translated in Transifex. HTTPS Everywhere will default to the English version if there is no available translation.

fuglede: Yeah, I actually tried that since that's what I gathered from the Transifex manual, but makexpi gave a warning without them.

jsha:
Thanks!

PR #980, opened 2015-01-26
Added Ems-ph.org.xml

fuglede:


jsha:
Can you clarify the comment? I don't understand why loading an HTTPS resource would fail as mixed content in any browser.

fuglede:
Right, that didn't make much sense; fixed.

PR #989, opened 2015-01-27
[Milepoint] Secure site is no longer on a separate host.

andrewachen:


jsha:
It looks like this ruleset doesn't cover www.milepoint.com, so you can simplify the rule to just:

```
<rule from="^http:"
        to="https:" />
```

julianladisch:
Simplify the rule to this:

    <ruleset name="Milepoint.com">
        <target host="milepoint.com"/>
        <target host="www.milepoint.com"/>
        <rule from="^http:" to="https:"/>
    </ruleset>

PR #995, opened 2015-01-29
Add Forocoches ruleset into 4.0

i-rme:
Forocoches.com is the top 1 Spanish Forum with more than 2 million unique visitors per day.

Is the top 3601 most popular website worldwide (http://www.alexa.com/siteinfo/forocoches.com) and the top 60 in Spain.


jsha:
We're in the middle of preparing to upgrade master to stable, so we're not merging new rulesets into stable unless they are critical. Fixes to existing rulesets in stable can be merged though.

PR #999, opened 2015-01-30
Added Altmetric.com.xml

fuglede:
Had to be a bit creative to avoid having /blog/ served on https; the attached .xml works, but I imagine that line 16 of this one is /not/ the correct way of doing it, but I couldn't find a way around it -- in particular, this fails the included test.


fuglede:
As I read the [ruleset manual](https://www.eff.org/https-everywhere/rulesets), downgrade="1" seems to be what I want but the test still fails.

jsha:
The relevant line from the failing test:

> failure: Altmetric.com failed test: Rule redirects to something other than https.

You need to add the ruleset name to utils/downgrade-whitelist.txt.

I believe this ruleset will not do the correct thing because rules are evaluated in order, and the first result is used. So the blog URL will hit the first rewrite and not make it to the following ones. You can reorder the rules to fix this - make sure to manully test, though!

Also, per the new [ruleset testing](https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md) requirements and [ruleset style](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) guide, can you change this rule to list specify hosts in the target hosts, rather than wildcards?

Thanks,
Jacob

fuglede:
Fixed

jsha:
Failing test.

fuglede:
What would be the correct test domains in this case? Any combination I come up with seems to fail. The test seems to be complaining about the exclusion pattern in particular. On the other hand, the new [ruleset style](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) mentions that "important" subdomains need tests specified, but it's not clear what would be an important subdomain in this case.

jsha:
You need to have test URLs that match each case of your rules and exclusions. So in this case:

http://www.altmetric.com/blog
https://www.altmetric.com/blog
http://altmetric.com/blog
https://altmetric.com/blog

PR #1001, opened 2015-01-31
Disable ReadWrite by default for MCB & cert mismatch

### LINKED ISSUE #973: Mixed content completely breaks ReadWrite
> Zoinks! There's a problem
>
> It looks like you can't access some of the data needed to view this site. Often this is because the institution where this site is being accessed is blocking unrecognized domain names. If reloading the page does not fix this error, please notify your IT department and contact bugs@readwrite.com if you have any further questions.

This is because of mixed content.

StevenRoddis:
Fixes #973

### DIFF
--- a/src/chrome/content/rules/ReadWrite.xml
+++ b/src/chrome/content/rules/ReadWrite.xml
@@ -1,6 +1,7 @@
-<!--
-	Note: server is configured for rc4 only.
-
+<!--	
+	Problematic subdomains:
+		- api ** (site breaks without it)
+		- www (no secure protocols supported)
 
 	Mixed content:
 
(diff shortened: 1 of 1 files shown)


jsha:
Thanks! Typo in the comment: *Problematic.

Also, please put issue numbers in the body of the pull request, e.g. "Fixes #973." This makes them linkable. And if you specifically prefix with "Fixes ", then GitHub auto-closes the issue on merge.

Also, the subject of the pull request should be more descriptive, e.g. Disable ReadWrite by default for mixed content.

StevenRoddis:
@jsha thanks for your help on writing better PRs.
Typo fixed.

nyuszika7h:
For future reference, the commit message could be better too, something like this:

    Disable ReadWrite by default because of mixed content

    Fixes #973.

The fact that your commit fixes an issue doesn't say anything if someone is looking at the commit log without remembering which issue that is.

jsha:
I should have caught this in the first round of review: For sites that break because of mixed content, instead of setting default_off, we set platform="mixedcontent". That way, if either Chrome or Firefox fix the bugs that prevent us from rewriting subresources before they get blocked by mixed content, we can easily re-enable the appropriate rules.

And it's fine to leave the name of the rule as-is, without adding (Broken).

StevenRoddis:
@jsha `api.readwrite.com` presents a cert mismatch, it's so even if MCB was off it still wouldn't work.

https://www.ssllabs.com/ssltest/analyze.html?d=api.readwrite.com

-----
@jsha I'll fix up the rule name.
@nyuszika7h I'll try fix up the commit message.

jsha:
Merged manually.

PR #1011, opened 2015-02-03
Added ingdirect.com.au

StevenRoddis:


StevenRoddis:
Already covered in /src/chrome/content/rules/ING-Direct.xml

PR #1013, opened 2015-02-03
[RawGit] rawgit.com

StevenRoddis:


src/chrome/content/rules/RawGit.xml
```diff
@@ -0,0 +1,18 @@
...
+			- cdn
+
+		Doesn't Work:
+			- www (mismatch)
+-->
+<ruleset name="RawGit">
+  <target host="rawgit.com" />
+  <target host="*.rawgit.com" />
+
+  <rule from="^http://(?:(cdn\.)|www\.)?rawgit\.com/"
```
reedy: This syntax took a little while to grok :)

Could we just use not match such as (?!www\.) ?

reedy: Though, now I notice we use this code elsewhere

StevenRoddis: Interesting, I wonder if it would be better to have two rules?
```xml
<rule from="^http://cdn\.rawgit\.com/"
          to="https://cdn.rawgit.com/" />

<rule from="^http://(?:www\.)?rawgit\.com/"
          to="https://$1rawgit.com/" />
```

StevenRoddis:
Updated ruleset to reflect new style guide

Issue #1010, opened 2015-02-02
Can't add loans to basket on kiva.org with HTTPS-E enabled (5.0-development-2)


austin987:
Originally filed at https://trac.torproject.org/projects/tor/ticket/11645.

You'll need a kiva account for this. Once logged in, using FF 28.0 on Fedora 20, trying to add a loan to my account has no effect. Disabling HTTPS-E for kiva.org alone works around it.

Also present in Iceweasel 31.3.0, 5.0-development-2, debian jessie.

StevenRoddis:
Hi @austin987 are you able to post the console output from Firefox when you try to add a loan?

austin987:
Hi @StevenRoddis, there is no output at that time (and interestingly, iceweasel gives no outout at all, aside from the G-lib error.

There are some earlier messages, which I've attached. Do I need to enable logging somewhere first?

austin987:
(process:5427): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed
GLib-GIO-Message: Using the 'memory' GSettings backend.  Your settings will not be saved or shared with other applications.
info: Error while loading localization: There is no localization file for your language.
error: An exception occurred.
Traceback (most recent call last):
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/l10n/loader.js:18", line 5, in readURI/request.onload
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:138", line 32, in resolve
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:32", line 43, in then
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:112", line 34, in resolved
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:138", line 32, in resolve
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:32", line 43, in then
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:112", line 34, in resolved
  File "resource://7ca9cf31-1c73-46cd-8377-85ab71ea771f/api-utils/lib/promise.js:138", line 32, in resolve
[...]

PR #1016, opened 2015-02-05
[Kiva] Remove /ajax/ from being excluded from the exclusion.

StevenRoddis:
(ie. Exclude /ajax/)
See GitHub Issue #1010 & https://trac.torproject.org/projects/tor/ticket/11645
Was preventing users from adding a loan to a basket.


StevenRoddis:
Added test coverage

PR #1021, opened 2015-02-05
qtio: broaden rule for downloads/bugreporting links

DomT4:
Just broadens the Qt.io rule a bit to include the download link, which doesn't autoredirect to SSL/TLS, and the bug reporting link, which does.

### DIFF
--- a/src/chrome/content/rules/Qt.io.xml
+++ b/src/chrome/content/rules/Qt.io.xml
@@ -5,10 +5,11 @@
 <ruleset name="Qt.io">
 
 	<target host="qt.io" />
+	<target host="download.qt.io" />
+	<target host="bugreports.qt.io" />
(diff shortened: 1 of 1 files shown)


jsha:
Thanks! We recently added a style guide suggesting that rules like this should enumerate hosts with target tags rather than in the regex: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

Would you mind changing this to have a target host for each hostname, and then a simple rule `<rule from='^http:' to='^https:' />` ?

DomT4:
Darn, I'd only just gotten used to the old style :wink:. I'll read through the new guide and fix this up ASAP :+1:

DomT4:
Done, I think.

jsha:
Excellent, thanks for bearing with us through the style changes!

PR #1025, opened 2015-02-06
Tv2.dk rules (testing needed)

fuglede:
Tv2.dk is one of the most visited news sites in Denmark and so having any kind of rules here would be rather neat; devs have said that they have been working on it, but as always with this kind of site, they appear to be too dependent on trackers.

As is clear from the rule comments, current support is partial at best. The domains that do work do work reasonably well though, outside of the ordinary mixedcontent problems. Given the scope of the site, I wouldn't mind some extra pairs of eyes to ensure that I didn't miss anything.


jsha:

Thanks,
Jacob

PR #1030, opened 2015-02-07
Added Censurfridns.dk.xml

fuglede:


jsha:

Thanks,
Jacob

reedy:
@jsha I guess we should bulk comment that on (recently?) open pull requests?

jsha:
@reedy: Yep. I'll go through and do that.

reedy:
No point me bulk leaving these comments on all of your updated commits...

Shouldn't these commits have <test url="http://www.example.com/" /> too now?

jsha:
@fuglede: I'm not sure I understand your question.

This ruleset doesn't trigger the coverage tests because it is default_off.

Additionally, we don't yet test securecookie rules in the ruleset test coverage checker.

fuglede:
But if the rule is active, all cookies sent should have the secure tag, right?

jsha:
@fuglede: What's the host on the cookie you get? If it's ".blog.censurfridns.com" it might not get the tag correctly.

Regarding default_offs: I am willing to be convinced in some cases. You made a good point about having the file in place so that others don't have to replicate your work. But I am wary of bloating the extension with rulesets that never get used. In this case it looks like the site is not close to working under HTTPS because it has a self-signed cert, and the rule does not have a lot of exclusions or notes that would be useful in the future when turning it on, so I'm inclined to say no to this ruleset.

Regarding "important" subdomains: The ruleset coverage rules enforce that you have a minimum number of test URLs for each rule, based on the complexity of the rule. Usually meeting that requirement is sufficient. However, a wildcard target host like "*.example.com" can cover infinitely many domains, so we can't necessarily enumerate them. When those target hosts are present, you should make sure to include test URLs from at least three subdomains.

fuglede:
@jsha: Yeah, I tried changing the host attribute to "^\.blog\.censurfridns\.dk$" but that doesn't seem to help.

Also, agreed, that point on default_offs certainly does not apply in this case. As far as I understood, at some point in the future, Chrome (and friends?) will deal with plain HTTP even more harshly than it will with self-signed/bad certificates, so perhaps at some point in the future rules like these would make sense? I have no idea when that's going to happen though.

About test subdomains: I'm still not sure I follow. Shouldn't those test domains then also be included in all the other rules I submitted a while ago? (And as far as I understand, as long as one does not use wildcards, wouldn't it be possible to automate the inclusion of subdomains for testing?)

jsha:
I think we're on the same page regarding test subdomains. If you list all subdomains explicitly, they will be automatically included in testing. It's only when you use a wildcard that you need to add test URLs for some of the subdomains.

fuglede:
Ah, I see; the one in #999 confused me greatly as it had no wildcards and already had the relevant domains covered as targets.

Anyway, thanks. I'll close this one.

PR #1031, opened 2015-02-08
Delete Privacybox.de.xml

J0WI:
The projects has been discontinued. The site will no longer support https.


jsha:
Merged manually, thanks!

PR #1032, opened 2015-02-09
Added trakt.xml

qwerty12:
I didn't run into any mixed content issues (for instance, trakt.us is automatically accessed over HTTPS by trakt.tv if browsing it with HTTPS) whilst browsing trakt.tv with this rule on in Firefox. I tried to adhere to the common patterns I saw in other rules, but forgive me if there's any oddities.

### DIFF
--- a/src/chrome/content/rules/trakt.xml
+++ b/src/chrome/content/rules/trakt.xml
@@ -1,0 +1,20 @@
+<ruleset name="trakt">
+
+	<target host="trakt.tv" />
+	<target host="*.trakt.tv" />
+
+	<!-- For securecookie rule below (used for some images) -->
+	<target host="*.trakt.us" />
+
+	<!--	Not secured by server:
+					-->
+	<!--securecookie host="^\.trakt\.(tv|us)$" name="^__cfduid$" /-->
+	<!--securecookie host="^trakt\.tv$" name="^(_traktsession|AWSELB|trakt_username|remember_user_token)$" /-->
+
+	<securecookie host="^(www\.|\.)?trakt\.tv$" name=".+" />
+	<securecookie host="^\.trakt\.us$" name=".+" />
+
+	<rule from="^http://(www\.)?trakt\.tv/"
+		to="https://$1trakt.tv/" />
+
+</ruleset>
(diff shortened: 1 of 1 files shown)


jsha:
The tests are failing because you need to add some test URLs to your ruleset using <test url="..." />. See https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md for more details.

qwerty12:
Thanks. Of all the subdomains I found, only support.trakt.tv seems to be in use. `dev.trakt.tv` seems to host the old, abandoned trakt.tv interface and `go.trakt.tv` is a redirector. 

My apologies, but I've had to mark this as mixedcontent. As I use uBlock on my normal Firefox profile with it set to block Gravatar requests, I did not notice trakt.tv downloading the placeholder image if a Gravatar is not set over HTTP.

src/chrome/content/rules/trakt.xml
```diff
@@ -0,0 +1,33 @@
...
+-->
+<ruleset name="trakt" platform="mixedcontent">
+
+	<target host="trakt.tv" />
+	<target host="*.trakt.tv" />
+
+	<!-- For securecookie rule below (host used for movie etc. images) -->
+	<target host="*.trakt.us" />
+
+	<test url="http://www.trakt.tv/" />
```
jsha: I'm recommending that for rules like this, you replace

`<target host="*.trakt.tv">`

```
<target host="www.trakt.tv" />
<target host="support.trakt.tv" />
```

That way you can leave out the test URLs, since each non-wildcard target tag creates an implicit test URL.

Also, that way you can greatly simplify the rewrite rule, so it becomes:

```
<rule from="^http:"
  to="https:" />
```

Thanks,
Jacob

StevenRoddis:
@qwerty12 
>My apologies, but I've had to mark this as mixedcontent

Wouldn't that still be ok since it's just passive mixed content (images)?

qwerty12:
I'm sorry for the delay.

@jsha
Thanks for the improvements.

@StevenRoddis 
Yes, it is just an image, according to the console. If that's the case, then I'll remove the attribute (I put it in originally as it was causing Firefox to replace the padlock with a triangle warning sign and I, unaware of HTTPS Everywhere policies, would rather err on the safe side).

Otherwise, yeah, it's only Gravatar's showing of trakt's default placeholder image from a HTTP source that causes Firefox to throw up the warning sign. If I'm looking at a userpage of someone who has a Gravatar image set or if I run uBlock with a rule blocking Gravatar, then the padlock remains in place... Thanks.

PR #1035, opened 2015-02-10
Idea/suggestion: Include total count when some rules are inactive

fuglede:
One thing that has bugged me a few times is that there's no way to tell if one is browsing a site that supports a ruleset which is disabled per default. These commits fix this by adding the total count of rules, in the case where some are inactive. [See this screenshot](https://hallon.fuglede.dk/dump/https-e-counter.png).

Pro:
- Makes it easier to identify the possibility of activating rules that are otherwise hidden in the toolbar button menu.

Some cons:
- Only really useful to people who don't mind experimenting with broken rulesets
- Clutters the interface ever so slightly. This is mitigated by only changing the label when the total count differs from the active count.

(Also, a question: does the ApplicableList have a counter itself?)

### DIFF
--- a/src/chrome/content/toolbar_button.js
+++ b/src/chrome/content/toolbar_button.js
@@ -181,21 +181,29 @@ httpsEverywhere.toolbarButton = {
     }
     // Make sure the list is up to date
     alist.populate_list();
-
-    var counter = 0;
+    
+    var totalCounter = 0;
+    for (var x in alist.all) {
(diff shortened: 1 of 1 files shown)


jsha:
I think for the average user, this is too much clutter. If you want to add this as a preference, I think that would be fine. Also, if you want to port the counter code into ApplicableList that would be great.

PR #1037, opened 2015-02-11
Use system python-lxml in Travis.

jsha:
This avoids having to 'pip install' lxml, which takes a long time because it
compiles the package each time. Instead we install the precompiled version from
apt-get and tell virtualenv it's okay to use it. Note that this works for Python
2.7 only.

This brings test times from ~3 minutes to ~1 minute.


cooperq:
lgtm

PR #1040, opened 2015-02-12
[jsFiddle] Added jsfiddle.net

StevenRoddis:


src/chrome/content/rules/jsFiddle.xml
```diff
@@ -0,0 +1,17 @@
...
+			- $
+
+		Problematic:
+			- www (cert mismatch)
+-->
+<ruleset name="jsFiddle">
+  <target host="jsfiddle.net" />
+  <target host="*.jsfiddle.net" />
+
+  <rule from="^http://(?:www\.)?jsfiddle\.net/"
```
jsha: It looks like this rule only covers `jsfiddle.net` and `www.jsfiddle.net`. If so, please change the `*.jsfiddle.net` above to `www.jsfiddle.net`.

That will have the added benefit that the ruleset coverage checker can create an implicit test URL of `http://www.jsfiddle.net/`, and you will have sufficient ruleset test coverage: https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md

StevenRoddis: Fixed.

PR #1041, opened 2015-02-12
Copying my changes that exist in 4.0 to master

StevenRoddis:
Added some test urls too.


jsha:
I'm planning to do another merge of 4.0 into master soon. I assume these rules will be automatically brought in with that merge?

Also you've got two test failures:

```
failure: Lanik.us failed test: The 'from' rule contains unescaped period in regular expression. Try escaping it with a backslash.
failure: Hostname *.blender.org shows up in 2 different rulesets.
```

StevenRoddis:
@jsha but I've made a few changes, added test urls and fixed some bugs.

src/chrome/content/rules/iTerm2.xml
```diff
@@ -0,0 +1,24 @@
...
+		doing rewrite here for performance.
+-->
+<ruleset name="iTerm2">
+  <target host="iterm2.com" />
+  <target host="*.iterm2.com" />
+
+  <rule from="^http://(?:www\.)?iterm2\.com/"
+          to="https://iterm2.com/" />
+
+  <test url="http://iterm2.com/" />
```
jsha: This test URL is unnecessary, since the ruleset checker creates an implicit URL for each target host.

src/chrome/content/rules/iTerm2.xml
```diff
@@ -0,0 +1,24 @@
...
+  <test url="http://www.iterm2.com/" />
```
jsha: What I've been doing with rulesets like these, where there's a wildcard but the actual rules only rewrite the base name + www, is to replace the wildcard with "www." That takes advantage of the implicit test URL to save some typing, and make it a little clearer to immediately see which domains are rewritten.

In some cases, like this one, narrowing down the targets like that lets you dramatically simplify the rule to this:

```
<rule from="^http:"
  to="https:"
```

You lose the redirect-skipping performance gain you mentioned in the comment, but since HTTPS Everywhere is not intended to try and improve end-user performance, I think the gain in ruleset simplicity is worth it.

src/chrome/content/rules/Tampermonkey.xml
```diff
@@ -0,0 +1,30 @@
...
+		nil
+-->
+<ruleset name="Tampermonkey.net">
+  <target host="tampermonkey.net" />
+  <target host="*.tampermonkey.net" />
+
+  <rule from="^http://tampermonkey\.net/"
+          to="https://tampermonkey.net/" />
+
+  <rule from="^http://(www|forum|safari)\.tampermonkey\.net/"
```
jsha: Similar to my other comment, I would suggest breaking out each of these domains into its own target host and then using a simple http: -> https: redirect. I'm working on a style guide that I'll post soon.

So it would be like:
```
<target host="tampermonkey.net" />
<target host="www.tampermonkey.net" />
<target host="forum.tampermonkey.net" />
<target host="safari.tampermonkey.net" />
```

And again, this will allow you to drop the redundant test URL tags.

jsha:
Looking at the test errors, I think they are because the Relax-NG grammar specification I added was too strict and does not allow for tests in arbitrary location. I will fix that.

src/chrome/content/rules/Tampermonkey.xml
StevenRoddis: Sounds good I'll wait for the style guide.

jsha: Style guide is live: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md. Sorry for all the churn in how we write rules. I think in the end it will make for a much more maintainable ruleset library!

jsha:
I checked into the test errors, and they are not due to the Relax-NG grammar being wrong. You've got an escaped dot in the 'to' section of your Lanik rule.

StevenRoddis:
Good catch! Fixed.

StevenRoddis:
I'll rewrite these with the new style guide.

StevenRoddis:
Should be good now.

PR #1045, opened 2015-02-12
Remove MokaFive from ruleset

geofft:
We don't currently support HTTPS properly on our website (this makes me very sad, but we tried and failed, and we aren't linking to HTTPS / guaranteeing that the cert will still be valid). I'll try to send a rule update if this changes, or preferably we'll just do HSTS.


jsha:
@geofft: Very sorry to hear that! I'd be interested to hear more about what problems you ran into, since a big part of my job is helping more web sites get HTTPS deployed. So I'm always looking for tough spots that could be made easier with better infrastructure or tools.

PR #1046, opened 2015-02-12
Ruleset style guide

jsha:
Here's a proposed style guide for writing rulesets. It represents a significant departure from typical ruleset writing practice today, but makes it much easier to comply with the new test coverage rules, which in turn should make managing our ruleset library much easier.

cc @cooperq @pde @2d1

### DIFF
--- a/ruleset-style.md
+++ b/ruleset-style.md
@@ -1,26 +1,26 @@
 # Ruleset style guide
 
 Goal: Rules should be written in a way that is consistent, easy for humans to
-read and debug, and makes errors less likely.
+read and debug, and reduces the chance of errors, and makes testing easy.
 
 To that end, here are some style guidelines for writing or modifying rulesets.
 They are intended to help and simplify in places where choices are ambiguous,
 but like all guidelines they can be broken if the circumstances require it.
 
-Avoid using the left-wildcard ("target host='*.example.com'") for unless you
+Avoid using the left-wildcard ("&lt;target host='*.example.com'&gt;") unless you
 really mean it. Many rules today specify a left-wildcard target, but the
 rewrite rules only rewrite an explicit list of hostnames.
 
-Prefer listing explicit target hosts and a single rewrite from "^http:" to
+Instead, prefer listing explicit target hosts and a single rewrite from "^http:" to
 "^https:". This saves you time as a ruleset author because each explicit target
-host automatically creates a test URL, reducing the need to add your own test
-URLs. These also make it easier for someone reading the ruleset to figure out
+host automatically creates a an implicit test URL, reducing the need to add your
+own test URLs. These also make it easier for someone reading the ruleset to figure out
 which subdomains are covered.
 
 If you know all subdomains of a given domain support HTTPS, go ahead and use a
 left-wildcard, along with a plain rewrite from "^http:" to "^https:". Make sure
 to add a bunch of test URLs for the more important subdomains. If you're not
-sure what subdomains might exist, check the 'subdomain tab on Wolfram Alpha:
+sure what subdomains might exist, check the 'subdomain' tab on Wolfram Alpha:
 http://www.wolframalpha.com/input/?i=_YOUR_DOMAIN_GOES_HERE_.
 
 If there are a handful of tricky subdomains, but most subdomains can handle the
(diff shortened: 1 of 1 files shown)


ruleset-style.md
```diff
@@ -0,0 +1,90 @@
+# Ruleset style guide
+
+Goal: Rules should be written in a way that is consistent, easy for humans to
```
cooperq: Nit: No capital on rules.

ruleset-style.md
```diff
@@ -0,0 +1,90 @@
+# Ruleset style guide
```
cooperq: nit: title caps: Ruleset Style Guide

ruleset-style.md
```diff
@@ -0,0 +1,90 @@
...
+
+Goal: Rules should be written in a way that is consistent, easy for humans to
+read and debug, and reduces the chance of errors, and makes testing easy.
```
cooperq: nit: Remove the first 'and' here, before 'reduces'

ruleset-style.md
```diff
@@ -0,0 +1,90 @@
...
+left-wildcard, along with a plain rewrite from "^http:" to "^https:". Make sure
+to add a bunch of test URLs for the more important subdomains. If you're not
+sure what subdomains might exist, check the 'subdomain' tab on Wolfram Alpha:
+http://www.wolframalpha.com/input/?i=_YOUR_DOMAIN_GOES_HERE_.
+
+If there are a handful of tricky subdomains, but most subdomains can handle the
+plain rewrite from "^http:" to "^https:", specify the rules for the tricky
+subdomains first, and then then plain rule last. Earlier rules will take
+precedence, and processing stops at the first matching rule.
+
```
cooperq: If this is the case, wouldn't you want the most common subdomains at the top for the most efficient rule processing?

ruleset-style.md
```diff
@@ -0,0 +1,90 @@
...
+(i.e. www.beta). Avoid more visually complicated options like `([^/:@\.]+\.)?`.
+
+For `securecookie` tags, it's common to match any cookie name. For these, prefer
+`.+` over `.*`. They are functionally equivalent, but it's nice to be
+consistent.
+
+Avoid the negative lookahead operator `?!`. This is almost always better
+expressed using positive rule tags and negative exclusion tags. Some rulesets
+have exclusion tags that contain negative lookahead operators, which is very
+confusing.
```
cooperq: :+1:  The lookahead operator is the bane of all readable regexes

cooperq:
Have a couple of grammar nits and one question but other than those this looks good to me.

ruleset-style.md
jsha: Yes, but I think the hit is very small and the readability is worth it.

jsha:
Ready for re-review @cooperq.

ruleset-style.md
```diff
@@ -0,0 +1,92 @@
...
+
+Prefer capturing groups `(www\.)?` over non-capturing `(?:www\.)?`. The
+non-capturing form adds extra line noise that makes rules harder to read.
```
q--: I disagree with this as having capturing groups as opposed to non-capturing groups have a negative impact on performance: http://jsperf.com/regex-capture-vs-non-capture.

jsha: Thanks for the reference! That's a much bigger relative difference in speed that I would have expected. I have a couple of follow-up questions:

- Does the performance impact scale with the number of capturing groups? Many of our rules that use non-capturing groups use them alongside capturing groups.
- How much absolute overhead does this add to an example page load, in milliseconds? On my fast laptop, the test you linked indicates 5m ops/sec for capturing groups vs 12m ops/sec for non-capturing groups. That's the difference between 200 nanoseconds vs 83 nanoseconds, or a delta of 117ns. If an average page loads 100 URLs, and each URL evaluates 3 regexes, that still only adds up to 35 microseconds of difference per page load. Adding in a factor of 20 slowdown between my laptop and a slower machine, we're still only looking at 700 microseconds.

q--: As for your first question, I put together another JSPerf test: http://jsperf.com/regex-capture-vs-non-capture/8. So yes, the number of capturing groups matters.

As for your second question, I haven't tested kthis but some searching suggests [even sub-second delays matter](https://blog.kissmetrics.com/wp-content/uploads/2011/04/loading-time.pdf).

PR #1053, opened 2015-02-13
[Pomf.se] Added a.pomf.se

StevenRoddis:


src/chrome/content/rules/Pomf.se.xml
```diff
@@ -15,18 +15,28 @@
...
 
 	<securecookie host="^\.pomf\.se$" name="^__cfduid$" />
 
 
-	<rule from="^http://(?:www\.)?pomf\.se/"
-		to="https://pomf.se/" />
+	<rule from="^http://(www\.)?pomf\.se/"
+			to="https://pomf.se/" />
 
+	<test url="http://pomf.se/" />
```
jsha: These are redundant since they are automatically created by the target tag, so you can leave them out.

StevenRoddis:
Done

src/chrome/content/rules/Pomf.se.xml
jsha: You missed these two. :-)

jsha:
FYI (since the comment structure doesn't make it clear): I'm waiting for this branch on the deletion of the two remaining redundant test URLs.

StevenRoddis:
Cheers, should be good to go.

PR #1064, opened 2015-02-15
[Flybuys] Added flybuys.com.au

StevenRoddis:


src/chrome/content/rules/Flybuys.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Flybuys">
+	<target host="flybuys.com.au" />
+	<target host="www.flybuys.com.au" />
+
+	<rule from="^http://(www\.?)flybuys\.com\.au/"
```
jsha: Question mark is in the wrong place.

This should have been caught by the coverage test, since one of the hosts is not covered by any rule. I'll look into why it was missed.

StevenRoddis:
Good catch, fixed.

PR #1070, opened 2015-02-15
Bump up earliest date for mandatory coverage rulesets.

jsha:
Turns out --date=local means something different on VMs running on UTC.

The ideal thing would be to specify an explicit timezone, but this is easier and makes little difference in the long run.

cc @cooperq submitting as TBR.


cooperq:
lgtm

PR #1078, opened 2015-02-15
Improve validation

jsha:
Remove old cases for different paths to trivial-validate.py.
Pass --quiet option to trivial-validate so it doesn't spam error output with warnings about duplicate hostnames.
Remove two rulests from downgrade whitelist that no longer need to be there.

cc @cooperq @reedy for code review.


cooperq:
LGTM

PR #1084, opened 2015-02-16
[Tmblr.co] Redirect to the right domain

mnordhoff:
It was redirecting to www.tumblEr.com instead of www.tumblr.com.

This wasn't too much of a privacy issue, since https://www.tumbler.com/ has a
mismatching cert, so a user would have to click through the warning to load it.
Besides, it's only Tumblr's URL shortener.

### DIFF
--- a/src/chrome/content/rules/Tmblr.co.xml
+++ b/src/chrome/content/rules/Tmblr.co.xml
@@ -14,7 +14,7 @@
 
 	<!--	Redirect keeps path and args:
 						-->
-	<rule from="^http://tmblr\.co/+"
+	<rule from="^http://tmblr\.co/"
 		to="https://www.tumblr.com/" />
 
 </ruleset>
--- a/src/chrome/content/rules/Tmblr.co.xml
+++ b/src/chrome/content/rules/Tmblr.co.xml
@@ -1,14 +1,12 @@
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://tmblr.co/ => https://www.tumbler.com/: (51, "SSL: certificate subject name 'www.tervis.com' does not match target host name 'www.tumbler.com'")
-	For other Tumblr coverage, see Tumbler.xml.
+	For other Tumblr coverage, see Tumblr.xml.
 
 
 	^: refused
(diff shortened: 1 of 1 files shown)


reedy:
This needs rebasing/remaking ontop of master

mnordhoff:
Travis failed because... I haven't added tests in my two-byte typo fix?

reedy:
Heh, yeah. It's a because you've touched a file that doesn't now comply with the updated testing standards

mnordhoff:
Now what? I'm new here and don't know anything. Should I figure out how to add tests and do so? Will someone else? Can this fix be merged anyway without new tests?

src/chrome/content/rules/Tmblr.co.xml
```diff
@@ -1,21 +1,20 @@
...
-<ruleset name="Tmblr.co" default_off='failed ruleset test'>
+<ruleset name="Tmblr.co">
 
 	<target host="tmblr.co" />
 
+	<test url="http://tmblr.co/ZE5Fby1dP0xGe" />
 
 	<!--	Redirect keeps path and args:
 						-->
 	<rule from="^http://tmblr\.co/+"
```
jsha: The + is somewhat misleading here, since it indicates any number of slashes can be matched. It winds up working by coincidence, but I'd prefer if you removed it.

jsha:
@mnordhoff: Thanks for helping out with testing, and sorry we're being so strict about adding tests if you touch a file. We've got an enormous number of rules, and have been having a hard time testing them all thoroughly, so we've started requiring that any modified files be brought up to date on testing standards. It looks like you found the relevant files, but just in case:

I'll add these to the test error message so they are more obvious. Thanks for your contribution!

src/chrome/content/rules/Tmblr.co.xml
mnordhoff: I agree the `+` is weird. It's harmless, so I decided not to make extraneous changes and complicate the diff. I can fix it later, but i don't have time right now.

jsha:
Thank again @mnordhoff!

PR #1086, opened 2015-02-17
Fix ruleset National-Security-Agency.xml by removing non-resolving host

mnordhoff:
Yes, "nsa.gov" really doesn't resolve. "www.nsa.gov" works.

By the way, I suspect the website no longer has the mixed-content issues it used to, but I can't say for certain and don't know how to test.


jsha:
Thanks!

PR #1087, opened 2015-02-17
install-dev-dependencies.sh fails after cloning

NovemberZulu:
Fix: install-dev-dependencies.sh fails if run just after cloning due to missing https-everywhere-checker/requirements.txt. This file is created when the corresponding submodule is updated.


jsha:
Thanks!

NovemberZulu:
You are welcome!

PR #1088, opened 2015-02-17
Improve error message for ruleset coverage checker.

jsha:
Indicates which docs to read for improvements.

cc @cooperq for code review.


StevenRoddis:
If possible could the line number where the problem is also be displayed?

jsha:
That may be a little hard. It depends on whether libxml gives you the line number for tags it parsed.

Filed an issue: https://github.com/EFForg/https-everywhere/issues/1097

cooperq:
LGTM

PR #1089, opened 2015-02-17
Escape HTML tags in ruleset-testing.md

mnordhoff:
HTML in ruleset-style.md is escaped, but it's not in ruleset-testing.md, so it doesn't get displayed properly on Github:

See:

https://github.com/mnordhoff/https-everywhere/blob/ruleset-testing-html/ruleset-testing.md


jsha:
Thanks!

PR #1098, opened 2015-02-18
Fix MySQL rule.

jsha:
foo.mysql.com and similar test URLs failed the fetch test. Also there were more
exclusions than inclusions, so turned the rule into an explicit list of good
subdomains.

cc @cooperq for review


julianladisch:
add <target host="bugs.mysql.com"/>

cooperq:
looks good.

PR #1101, opened 2015-02-20
[UPS.com] Reenabled Exceptions

### LINKED ISSUE #1039: Track a package using UPS - requires login
##### What happened:

Went to UPS website with clean cache/cookies/etc. Selected a country, clicked on "Tracking" tab above, entered tracking code into "Track by Number" box. Clicked "Track" button. Was redirected to a login page.

##### What I expected to happen:

View track package page without requiring a login. Called UPS who confirmed this was the expected behavior. After some testing, identified that the expected behavior works with HTTPS Everywhere disabled for UPS.com.

StevenRoddis:
Fixes #1039

### DIFF
--- a/src/chrome/content/rules/UPS.com.xml
+++ b/src/chrome/content/rules/UPS.com.xml
@@ -40,16 +40,11 @@
 		<exclusion pattern="^http://www\.ups\.com/WebTracking/track($|\?)" />
 			<test url="http://www.ups.com/WebTracking/track" />
 			<test url="http://www.ups.com/WebTracking/track?" />
-		<!--
-			Exceptions:
-					-->
-		<exclusion pattern="^http://www\.ups\.com/+(?!assets/|favicon\.ico|img/|styles/|stylesheets/)" />
-
+			<test url="http://www.ups.com/WebTracking/track?foo" />
 
 	<securecookie host="^www\.campusship\.ups\.com$" name=".+" />
 
-
-	<rule from="^http://(?:www\.)?ups\.com/"
+	<rule from="^http://(?:www\.)?ups\.com/(assets/|favicon\.ico|img/|styles/|stylesheets/)"
 		to="https://www.ups.com/" />
 
 	<rule from="^http://www(\.)?apps\.ups\.com/"
--- a/src/chrome/content/rules/UPS.com.xml
+++ b/src/chrome/content/rules/UPS.com.xml
@@ -46,8 +46,8 @@
 
 	<securecookie host="^www\.campusship\.ups\.com$" name=".+" />
 
-	<rule from="^http://(?:www\.)?ups\.com/(assets/|favicon\.ico|img/|styles/|stylesheets/)"
-		to="https://www.ups.com/" />
+	<rule from="^http://(www\.)?ups\.com/(assets/|favicon\.ico|img/|styles/|stylesheets/)"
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/UPS.com.xml
```diff
@@ -22,28 +22,42 @@
...
+			<test url="http://www.ups.com/WebTracking/track" />
+			<test url="http://www.ups.com/WebTracking/track?" />
+			<test url="http://www.ups.com/WebTracking/track?foo" />
 
 	<securecookie host="^www\.campusship\.ups\.com$" name=".+" />
 
-
-	<rule from="^http://(?:www\.)?ups\.com/"
+	<rule from="^http://(?:www\.)?ups\.com/(assets/|favicon\.ico|img/|styles/|stylesheets/)"
 		to="https://www.ups.com/" />
```
jsha: I think you need to add a $2 to copy over the path, right? Have you manually tested this?

StevenRoddis: I tested each of those paths in my browser, I forgot to put the (?=asserts/....) lookahead by the looks of it. Will fix.

src/chrome/content/rules/UPS.com.xml
```diff
@@ -22,28 +22,42 @@
...
 		<!--
-			Exceptions:
-					-->
-		<exclusion pattern="^http://www\.ups\.com/+(?!assets/|favicon\.ico|img/|styles/|stylesheets/)" />
-
+			Now redirects to login:
+						-->
+		<exclusion pattern="^http://www\.ups\.com/WebTracking/track($|\?)" />
+			<test url="http://www.ups.com/WebTracking/track" />
+			<test url="http://www.ups.com/WebTracking/track?" />
```
jsha: Ideally for these types of test URLs we should have query parameters that are meaningful. The goal is not just to satisfy the letter of the coverage rule, but to let the coverage rule be a reminder that the more complex the rewrite rule is, the more likely it is that one of the types of pages it covers will break, so we want to test all the different types of pages.

It may not be possible to get a useful tracking URL that does not expose private data, though. In this case I'd be willing to merge non-valid test URLs for that reason.

StevenRoddis: The exclusions were already there just commented out, unfortunately I don't have UPS tracking number handy.

I did try `123456`

Linkname: UPS: Tracking Information
    URL: http://wwwapps.ups.com/WebTracking/track
Post Data:
loc=en_US&HTMLVersion=5.0&USER_HISTORY_LIST=&trackNums=123456&track.x=Track
Post Content Type: application/x-www-form-urlencoded

StevenRoddis: https://wwwapps.ups.com/WebTracking/track
HTTP/1.1 302 Found
Date: Wed, 25 Feb 2015 07:46:51 GMT
Server: Apache
X-Frame-Options: SAMEORIGIN
Location: https://www.ups.com/one-to-one/login?returnto=https%3a//wwwapps.ups.com/WebTracking/track&reasonCode=-1&appid=ETRACK

jsha:
Looks like tests are failing because you are missing ".ups.com" after www(\.)?apps.

StevenRoddis:
Fixed, I probably should stop coding while tired.

PR #1113, opened 2015-02-23
[0xbadc0de.be] *.0xbadc0de.be → hostlist, add tests, use ^http: → https:

julianladisch:


reedy:
At a quick glance, the test failure(s) are unrelated. None of the errors are in the 0xbadc0de file, which is what changed

src/chrome/content/rules/0xbadc0de.be.xml
```diff
@@ -29,20 +29,19 @@
...
-		<test url="http://stats.0xbadc0de.be/" />
-		<test url="http://www.0xbadc0de.be/" />
+	<test url=    "http://0xbadc0de.be/?p=67" />
+	<test url="http://www.0xbadc0de.be/?p=67" />
 
-
-	<!--	Redirect drops path but not args:
-							-->
-	<rule from="^http://(?:www\.)?0xbadc0de\.be/[^?]*"
+	<!--	Redirect drops path but not args:         -->
```
jsha: You should also have a test URL that includes a path as well as args, ideally one that is linked to from somewhere on the site.

PR #1114, opened 2015-02-24
[dhl.de] Multiple subdomains + test coverage

johnp:
Already supports HSTS, but currently doesn't default to https.
The official account-page on paket.de for example links from https to http which prevents HSTS from working.

### DIFF
--- a/src/chrome/content/rules/DHL.de.xml
+++ b/src/chrome/content/rules/DHL.de.xml
@@ -10,11 +10,6 @@
 
   <securecookie host="^(?:.*\.)?dhl\.de$" name=".*" />
 
-  <rule from="^http://(?:www\.)?dhl\.de/" to="https://www.dhl.de/"/>
-  <rule from="^http://mobile\.dhl\.de/" to="https://mobile.dhl.de/"/>
-  <rule from="^http://nolp\.dhl\.de/" to="https://nolp.dhl.de/"/>
-  <rule from="^http://nolb\.dhl\.de/" to="https://nolb.dhl.de/"/>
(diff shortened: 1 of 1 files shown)


johnp:
The travis build shows that the file now has complete test-coverage (read: test does not fail anymore).

src/chrome/content/rules/DHL.de.xml
```diff
@@ -1,8 +1,20 @@
...
+  <target host="nolp.dhl.de"/>
+  <target host="nolb.dhl.de"/>
+  <target host="partner.dhl.de"/>
+  <target host="entwickler.dhl.de"/>
+  <target host="fcm.dhl.de"/>
 
   <securecookie host="^(?:.*\.)?dhl\.de$" name=".*" />
 
   <rule from="^http://(?:www\.)?dhl\.de/" to="https://www.dhl.de/"/>
+  <rule from="^http://mobile\.dhl\.de/" to="https://mobile.dhl.de/"/>
```
jsha: These can be simplified as

```
<rule from="^http:"
  to="https:" />
```

johnp: This would then still /only/ apply to the subdomains as given in the ``target`` rules? Because there are subdomains that are http-only.

jsha: That's correct. Because you no longer include the "*.dhl.de" domain, this would only apply to subdomains listed explicitly in the `target` rules.

johnp:
Updated accordingly.

PR #1117, opened 2015-02-25
Fix tests: disable failing rulesets.

jsha:
These rulesets were updated after the test coverage cutoff date,
but don't have sufficient test coverage. Disabling them for now.

Also, add test-ruleset-coverage.sh to test.sh and remove from .travis.yml.

Also, increase clone depth in .travis.yml as a short-term fix.

cc @cooperq @2d1 for code review.

PR #1124, opened 2015-02-26
[Github] replace *.github.com and *.githubusercontent.com, use ^http: → https:

julianladisch:
Fix #1117 by adding test coverage and reenabling rule

### DIFF
--- a/src/chrome/content/rules/Github.xml
+++ b/src/chrome/content/rules/Github.xml
@@ -25,8 +25,6 @@
 			- assets-cdn
 			- bounty
 			- cloud
-			- f.cloud
-			- codeload
 			- developer
 			- eclipse
 			- edu
@@ -40,7 +38,6 @@
 			- jobs
 			- mac
 			- mobile
-			- nodeload
 			- octodex
 			- pages
 			- raw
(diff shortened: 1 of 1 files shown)


jsha:
This looks good. Have you run the ruleset checker in manual test mode? That will check each domain to make sure it still exists and answers on 443.

```
python2.7 https-everywhere-checker/src/https_everywhere_checker/check_rules.py https-everywhere-checker/manual.checker.config src/chrome/content/rules/GitHub.xml
```

2d1:
Note: This has made github tracking cookies insecure.

If @jsha doesn't care about that, be warned I won't be using/testing the same github ruleset as is in master (because I care). I guess the use of the ruleset is widespread enough for that not to matter.

julianladisch:
I was not my intention to make secure cookies insecure. Can you give an example which cookie may be affected?

jsha:
@2d1: I do care about fixing the securecookie issue you flagged on IRC, thanks for the reminder. I've been punting it down the road in favor of improving the ruleset library, but I'll consider it a blocker for the next dev release: https://github.com/EFForg/https-everywhere/issues/1145.

PR #1127, opened 2015-02-26
Improve detection of updated rulesets.

jsha:
Previous, we tried to detect when rulesets changed by using git log --since.
However, this doesn't work very well for uncommited changes, and it works poorly
on Travis, which does a shallow clone and therefore is not able to get a good
history of which files have changed since when.

Instead, we create a list of SHA256 hashes of files as they stood on 2014-02-12
(plus some extras that slipped through because the changes were made before then
but merged later). When running coverage tests, and files that match those
hashes are ignored.

Also add a test-generator script to aid in generating test URLs for rulesets.

See also https://github.com/hiviah/https-everywhere-checker/compare/master...jsha:disable-rules?expand=1 for the support on the https-everywhere-checker side.

cc @cooperq for code review.


jsha:
Merging TBR since I think this unbreaks tests on some pending rulesets. @cooperq

cooperq:
lgtm

PR #1135, opened 2015-02-27
[Cryptonit.net] replace *.cryptonit.net by www.cryptonit.net, use ^http: → https:

julianladisch:
remove Cryptonit.net's default_off originating from #1117 (jsha/disable-uncovered-rulesets)


jsha:
This looks like only a rename, no changes?

julianladisch:
Forgot to `git add` the new filename. Fixed.

PR #1141, opened 2015-03-01
Add rule for hendrik.luup.info

hluup:


kampsun:
I viewed the build log and it seemed to me that my rule wasn't the one failing. I can add some tests if needed. I understood that they are only required if the rule has capture groups.

src/chrome/content/rules/Hendrik.luup.info.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="Hendrik Luup">
+  <target host="hendrik.luup.info" />
+
+  <rule from="^http://hendrik\.luup\.info/" to="https://hendrik.luup.info/" />
```
reedy: You should update this rule to use the new rewrite format

rule from="^http:" to="https:"/>

PR #1142, opened 2015-03-01
[IEEE] replace *.ieee.org, use ^http: → https:, add tests

julianladisch:


jsha:
Excellent tests, thanks!

PR #1146, opened 2015-03-02
Disable rulesets that break coverage test.

jsha:


cooperq:
Looks fine, but there is a lot of stuff in here that should be a high priority to fix, especially the linux distros. (Though most of them seem to do https by default now so thats good)

PR #1147, opened 2015-03-03
Fix cookie securing under new style guide.

jsha:
Cookies are only eligible to be secured if a ruleset matches them using
potentiallyApplicableRulesets. Previously, cookie wildcard domains
(".example.com") happened to match target host wildcards ("<target
host=*.example.com />") through a quirk in the implementation of
potentiallyApplicableRulesets.

However, the new ruleset style guide recommends against including
target host wildcards unless all subdomains are actually rewritten
to HTTPS. Under the old quirky behavior, the cookie domain attribute
".example.com" does not match the target host "example.com".

This change clarifies the function of potentiallyApplicableRulesets and modifies
the cookie-related functions to strip a leading dot before calling
potentiallyApplicableRulesets. This restores cookie securing functionality for
cookies with wildcard domain attributes.

### DIFF
--- a/src/chrome/content/code/HTTPSRules.js
+++ b/src/chrome/content/code/HTTPSRules.js
@@ -640,8 +640,15 @@ const HTTPSRules = {
     return output;
   },
 
+  /**
+   * Return a list of rulesets that declare targets matching a given hostname.
+   * The returned rulesets include those that are disabled for various reasons.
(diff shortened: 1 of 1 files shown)


jsha:
Fixes #1145

cooperq:
LGTM, go ahead and merge if you are ready.

PR #1150, opened 2015-03-04
Fix user rules in Chromium.

### LINKED ISSUE #741: Chrome: User-added rules don't persist enabled/disabled state across restart
When there's a site with either defective https cert, or one that I manually added (add rule set), there's no way to permanently remove that defective rule set for that site! I have to manually uncheck the site's name from the extension for it to work, every time I visit it!

Can you please provide a way to permanently remove a site?

jsha:
An undefined reference was preventing them from saving.

Also, addUserRule was not correctly checking active status on reload, meaning
that rules would not stay disabled across restarts.

Fixes #741

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -116,9 +116,12 @@ RuleSets.prototype = {
     if (!(params.host in this.targets)) {
       this.targets[params.host] = [];
     }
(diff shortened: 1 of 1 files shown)


jsha:
cc @cooperq for code review.

PR #1159, opened 2015-03-05
new rule for geocaching.com

PetziAt:


reedy:
So this is actually reverted now. There's already a (new) ruleset for geocaching.com at https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Geocaching.com.xml

PR #1164, opened 2015-03-07
Added trivial rule for nxtgn.org

fuglede:


semenko:
Awesome -- thanks! :)

Issue #983, opened 2015-01-26
Install dev dependencies error on OSX


jish:
I am having issues running `bash install-dev-dependencies.sh`

```
$ bash install-dev-dependencies.sh
+ type apt-get
install-dev-dependencies.sh: line 5: type: apt-get: not found
+ type brew
+ brew install python libxml2 gnu-sed
==> Installing python dependency: sqlite
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/sqlite-3.8.7.4.yosemite.bottle.
######################################################################## 100.0%
==> Pouring sqlite-3.8.7.4.yosemite.bottle.tar.gz
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local.

Mac OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

OS X provides an older sqlite3.

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/sqlite/lib
    CPPFLAGS: -I/usr/local/opt/sqlite/include

==> Summary
🍺  /usr/local/Cellar/sqlite/3.8.7.4: 9 files, 2.1M
==> Installing python
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/python-2.7.9.yosemite.bottle.7.
######################################################################## 100.0%
==> Pouring python-2.7.9.yosemite.bottle.7.tar.gz
==> Caveats
Setuptools and Pip have been installed. To update them
  pip install --upgrade setuptools
  pip install --upgrade pip

You can install Python packages with
  pip install <package>

They will install into the site-package directory
  /usr/local/lib/python2.7/site-packages

See: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Homebrew-and-Python.md

.app bundles were installed.
Run `brew linkapps python` to symlink these to /Applications.
==> /usr/local/Cellar/python/2.7.9/bin/python -s setup.py --no-user-cfg install --force --verbose --
==> /usr/local/Cellar/python/2.7.9/bin/python -s setup.py --no-user-cfg install --force --verbose --
==> Summary
🍺  /usr/local/Cellar/python/2.7.9: 4846 files, 79M
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/libxml2-2.9.2.yosemite.bottle.t
######################################################################## 100.0%
[...]

PR #1166, opened 2015-03-08
Remove the `--user` flag when installing dev dependencies.

jish:
Suggested fix from https://github.com/EFForg/https-everywhere/issues/983

### DIFF
--- a/install-dev-dependencies.sh
+++ b/install-dev-dependencies.sh
@@ -17,9 +17,9 @@ git submodule init
 git submodule update
 
 # Install Python packages
-pip install --user --no-allow-insecure --no-allow-external -r requirements.txt
+pip install --no-allow-insecure --no-allow-external -r requirements.txt
 cd https-everywhere-checker
-pip install --user -r requirements.txt
+pip install -r requirements.txt
 cd -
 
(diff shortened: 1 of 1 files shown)


jsha:
The `--user` flag is correct for most environments: we don't want to install globally. There was something about your environment that caused an issue, though:

```
    error: can't combine user with prefix, exec_prefix/home, or install_(plat)base
```

Can you figure out if you have default flags set for pip somewhere? E.g. environment variables or an rc file?

jish:
I use a pretty stock MacBook Air, and I do not do any Python development. Whatever is set, came with Mac OS.

I tried this a few months ago when I first filed the issue #983 and it was broken, and I actually just wiped and re-installed my MacBook Air a few days ago and I'm getting the same issue. So it seems the install script with the `--user` flag is just not really compatible with Mac OS / MacBook Airs. =/

jsha:
Good to know, thanks. I will try and get my hands on a Mac to do some testing. Do you have Xcode installed?

jish:
Yes, I have XCode installed now, and I did when I originally opened the issue.

I'm not sure if it was installed yet when I opened this PR.

Josh

PR #1170, opened 2015-03-08
Exclusion pattern for video streaming for politiken.dk

### LINKED ISSUE #1026: Politiken.dk rule breaks video streaming
While they work perfectly well for viewing articles, the rules in Politiken.dk.xml break video streaming; [here's an example](https://politiken.dk/tv/ECE2536531/direkte-med-lidegaard-islamisk-stat-vinder-frem-i-syrien-men-vi-har-bremset-dem-i-irak/).

Since the rules is on by default, this may or may not be a problem; I imagine that the most common use case for the webpage is reading articles, but still.

fuglede:
To fix #1026. I suppose it would make sense to include the pattern again in a platform="mixedcontent" ruleset, but maybe that's a little excessive.


fuglede:
This one fails the test as it lacks some of the tests required for the February 2015 Travis test. In fact I'm not entirely sure what some of the rules *do*; I'll drop the original author a mail about it.

fuglede:
In adding the various test criteria, I realized that I have no idea what the web.politiken.dk rule does. I asked the original author but did not hear back and so opted to simply remove that rule.

fuglede:
Closing; largely superceded by #1230 (cf. the discussion there).

PR #1171, opened 2015-03-08
Rule for moncompteformation.gouv.fr

Rudloff:
Hello,

Here is a rule for moncompteformation.gouv.fr.

Regards,

### DIFF
--- a/src/chrome/content/rules/moncompteformation.gouv.fr.xml
+++ b/src/chrome/content/rules/moncompteformation.gouv.fr.xml
@@ -1,0 +1,6 @@
+<ruleset name="moncompteformation.gouv.fr">
+    <target host="espaceprive.moncompteformation.gouv.fr" />
+    <target host="gestespaceprive.moncompteformation.gouv.fr" />
+
+    <rule from="^http:" to="https:" />
+</ruleset>


src/chrome/content/rules/moncompteformation.gouv.fr.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="moncompteformation.gouv.fr">
+    <target host="espaceprive.moncompteformation.gouv.fr" />
+    <target host="gestespaceprive.moncompteformation.gouv.fr" />
+
+    <rule from="^http://espaceprive\.moncompteformation\.gouv\.fr/" to="https://espaceprive.moncompteformation.gouv.fr/" />
+    <rule from="^http://gestespaceprive\.moncompteformation\.gouv\.fr/" to="https://gestespaceprive.moncompteformation.gouv.fr/" />
```
reedy: This ruleset can be simplified by just using the new rewrite format instead

```
<rule from="^http:"
to="https:" />
```

Rudloff:
Oh, I didn't know you could do it like this. I have updated the rule.

PR #1174, opened 2015-03-09
Fix tumblr image CDN and a ruleset validation bug on stable branch

### LINKED ISSUE #1168: Tumblr redirects incorrectly redirect some images to Edgecast CDN
When viewing Tumblr, approximately half of the blog-posted images on my dash do not show up. If I right-click and view the image then I get an error from Edgecast CDN [1]. The beginning parts of the URLs for working images are identical.

If I disable the "Tumblr (partial)" ruleset then everything works fine (but I'm obviously not protected). If I clone the Tumblr ruleset and comment out the line:

    <rule from="^http://data\.tumblr\.com/"
                to="https://gs1.wac.edgecastcdn.net/8019B6/data.tumblr.com/" />

(which gives me all of the protection of the original rules, except on data.tumblr.com) then everything works fine. I don't know why this works, since the images are on ##.media.tumlr.com and so shouldn't be matching the data.tumblr.com rule, but it seems to work.

Browser: Firefox 35 and Chromium 39 on Linux
Extension versions: 4.0.3 and 2015.1.22
No other custom rules used

[1] http://ibboard.tumblr.com/post/112897748412/thank-goodness-for-content-delivery-networks

diracdeltas:
Fix #1168

### DIFF
--- a/src/chrome/content/rules/Lanik.us.xml
+++ b/src/chrome/content/rules/Lanik.us.xml
@@ -17,8 +17,8 @@
   <rule from="^http://lanik\.us/"
           to="https://lanik.us/" />
 
-  <rule from="^http://(www|forums)\.lanik.us/"
-          to="https://$1\.lanik.us/" />
+  <rule from="^http://(www|forums)\.lanik\.us/"
+          to="https://$1.lanik.us/" />
 
   <securecookie host="^forums\.lanik\.us$" name=".+" />
 </ruleset>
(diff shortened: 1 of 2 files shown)


jsha:
Can you confirm this will merge cleanly into master? I'm going to merge 4.0 into master soon and discard all conflicts in preference for the master copy.

diracdeltas:
It's already fixed in master

Issue #58, opened 2014-01-07
StackExchange OpenID Login fails


ulikoehler:
I'm using Chrome Stable on Ubuntu Linux 13.10, x86_64. When trying to enable the _Stack overflow (Mixed Content)_ ruleset, I can't login via StackExchange OpenID.
Steps to reproduce that work for me:
- Ensure you're not logged in in StackOverflow and that the ruleset is enabled
- Click the Login Button, click on StackExchange button (I was able to reproduce this with Google login, too)
- Message is shown: "Third Party Cookies Appear To Be Disabled", click on "continue and login manually
- Enter StackExchange credentials and click login button
- You are redirected (via OpenID) to the StackOverflow site. The following error message is shown in bold red :

```
Unable to log in with your OpenID provider:

The openid.return_to parameter (http://stackoverflow.com/users/authenticate/?s=[.......] does not match the actual URL (https://stackoverflow.com/users/authenticate/?s=[.......]) the request was made with.
```

(Note: Any query parts were removed to protect me from potential session hijacking, but I'm happy to send you the full error message by E-Mail if neccessary, please use the email listed on Github and my PGP key 1BAA8BBC.)

This issue reproducibly doesn't occur when the StackExchange rule in HTTPS Everywhere is disabled.

The only difference I can spot in the two URLs is that one is HTTP and the other is HTTPS. I think its therefore safe to assume that HTTPS Everywhere kills the OpenID login process by changing the URL scheme

Possible solution: Add _http://stackoverflow.com/users/authenticate_ to blacklist, so HTTPS everywhere leaves this request unencrypted. Maybe it's even possible to encrypt the originally requested URL (I don't know too much about the OpenID internals)?

Thank in advance for reviewing this & for this awesome tool!

PR #1182, opened 2015-03-11
Updated Stack-Exchange.xml

guiweber:
Removed meta.*.stackexchange.com as the certificate is not valid for this pattern

### DIFF
--- a/src/chrome/content/rules/Stack-Exchange.xml
+++ b/src/chrome/content/rules/Stack-Exchange.xml
@@ -151,7 +151,7 @@ Fetch error: http://app.stacktack.com/ => https://s3.amazonaws.com/stacktackapp/
 	<rule from="^http://(?:www\.)?stackapps\.com/"
 		to="https://stackapps.com/" />
 
-	<rule from="^http://((?:\w+|meta\.\w+|discuss\.area51)\.)?stackexchange\.com/"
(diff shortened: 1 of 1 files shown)


jsha:
Looks good. Keep in mind this ruleset is currently disabled, and will need updates to the ruleset test in order to re-enable it.

guiweber:
Right, I had not seen that. It is still enabled in 5.0 development 2. I'll work on it when I have time.

jsha:
Great, thanks. BTW, please try installing the latest 5.0development.3:
https://www.eff.org/https-everywhere#install-development. Your dev2
install won't auto-update due to a packaging mistake, and we could use
all the testing help we can get.

guiweber:
I have a new rule almost ready to go, but I was thinking of merging the following files into this one. Is this ok with you?
- Ask_Ubuntu.com.xml
- MathOverflow.net.xml
- Server_Fault.com.xml

Also, do you know if issue #58 got resolved? I removed the exclusion and I can login just fine on stackoverflow through openID using my google credentials on firefox.

jsha:
Yes, it's fine to merge those files into it.

Regarding #58: It's entirely possible and likely that something changed on their site so the exclusion is no longer necessary. If you've tested thoroughly and it works, go ahead and remove the exclusion.

PR #1186, opened 2015-03-11
Update UBS.xml

J0WI:
removed obsolete domain


jsha:
Looks good. Keep in mind this rule is still disabled until you remove the default_off.

PR #1194, opened 2015-03-12
Fix some Google-related rulesets

jsha:
Re-enable Youtube ruleset.
Remove broken rule within GoogleVideos ruleset, and disable the ruleset because it needs ruleset tests.
Remove /url from list of paths that fail on cctld and need to be redirected to .com.

cc @2d1 @reedy @StevenRoddis could one of you code review this? Cooper and Peter are unavailable for reviews currently. Thanks!


StevenRoddis:
LGTM

Issue #271, opened 2014-05-04
Is there any reason for separating Ask Ubuntu, MathOverflow and Server Fault?


nyuszika7h:
It doesn't make sense to me, as they belong to the same Stack Exchange network.

semenko:
Closing, as this issue is quite old and my no longer be valid. Feel free to reopen if you think it's still a bug.

PR #1195, opened 2015-03-13
Merged and updated all StackExchange rulesets

guiweber:
Also addresses issue #271


src/chrome/content/rules/Stack-Exchange.xml
```diff
@@ -1,177 +1,131 @@
...
 
-	<rule from="^http://discuss\.area51\.stackexchange\.com/"
-		to="https://discuss.area51.stackexchange.com/" />
+<!-- Rules -->
+	
+	<!-- https links from other pages to these two sections only will cause MCB for important elements, hence the downgrade -->
+	<rule from="^https://careers\.stackoverflow\.com/(cities|employer)" 
+		to="http://careers.stackoverflow.com/$1" downgrade="1" />
+	
+	<rule from="^http://([\w-]+)\.stackexchange\.com/$"
```
jsha: I think this is covered by the http: -> https: rule below.

guiweber:
You are right. Fixed.

src/chrome/content/rules/Stack-Exchange.xml
```diff
@@ -1,177 +1,128 @@
...
+	<securecookie host="^meta\.serverfault\.com$" name=".+" />
+	<securecookie host="^serverfault\.com$" name=".+" />
+	
+	<!-- Stackauth -->
+	<securecookie host="^stackauth\.com$" name=".+" />
+	<securecookie host="^dev\.stackauth\.com$" name=".+" />
+	
+	<!-- Stackexchange -->
+	<securecookie host="^stackexchange\.com$" name=".+" />
+	<securecookie host="^([\w-]+)\.stackexchange\.com$" name=".+" />
```
jsha: It looks like you changed this from `host="\.stackexchange.com$"`. But there is a subtle semantic difference. A site can actually send `Set-Cookie: foo=bar; domain=.stackexchange.com`, which means "Send this cookie on requests to stackexchange.com and all subdomains."

guiweber: I actually forgot to include an important exclusion in the ruleset and I'm going to push another commit for that. With this in mind, maybe the cookies should stay like they are since some subdomains will be excluded:
```<exclusion pattern="^http://.*\.([\w-]+)\.stackexchange.com"/>```

But I'm not really familiar with how cookies work behind the scene, so please let me know if I should I should put ```domain=.stackexchange.com``` anyway.

Should I also add a ```domain=.***.com``` cookie tag for the other domains (stackauth, stackoverflow, serverfault, etc...) in the ruleset?

And one more question to make sure it's fine. I removed the following rule from the ruleset:
```<rule from="^http://(?:www\.)?blogoverflow\.com/$" to="https://stackexchange.com/blogs" />``` since blogoverflow.com redirects to stackexchange.com/blogs by itself. Is this ok? We already catch stackexchange.com/blogs with the other rules.

**Comment by guiweber** [comment]
I pushed my latest commit with the new exclusion without changing the cookies, pending your reply.

However it looks like Travis doesn't like something about the new rule:
*failure: Stack-Exchange.xml failed test: The 'from' rule contains an invalid extended regular expression.*
would you know what doesn't work with this?
```<rule from="^https://([\w-.]+)\.([\w-]+)\.stackexchange\.com" to="http://$1.$2.stackexchange.com" downgrade="1" />```
I tested the regexp in and it works well, as does the updated rule in Firefox.

**Comment by jsha** [comment]
You can test in Python:

```
import re
re.compile("...")
```

It looks like the issue is that `[\w-.]` looks like a character range, and it's invalid for that. Try `\w.-` instead.

File: src/chrome/content/rules/Stack-Exchange.xml (hunk 2 above)
Comment (jsha) [inline src/chrome/content/rules/Stack-Exchange.xml]: Okay, it looks like your fix was correct: Since `meta.*.stackexchange.com` does have a login capability, it probably relies on cookies. So the only cookies we can reliably secure are the ones set for a specific hostname, as you've got it set up. Have you checked the actual cookies set by the server to see if there are any of these, and if they are important?

File: src/chrome/content/rules/Stack-Exchange.xml [hunk 3]
```diff
@@ -1,177 +1,138 @@
...
-						-->
-		<exclusion pattern="^http://(?:\w+\.)?stack(?:exchange|overflow)\.com/users/authenticate/" />
+	<!-- This wildcard covers 125 subdomains -->
+		<test url="http://academia.stackexchange.com/" />
+		<test url="http://crypto.stackexchange.com/" />
+		<test url="http://reverseengineering.stackexchange.com/" />
+		<test url="http://unix.stackexchange.com/" />
+		<test url="http://writers.stackexchange.com/" />
+
+		<!-- deeper domains cause certificate missmatch errors - there is also a downgrade rule below -->
```
jsha: *mismatch

guiweber:
Thanks. I actually copied/pasted this directly from ruleset-style.md when I checked for the preferred way to match DNS labels (I know, I'm lazy ;- ). So I updated the document as well in this new pull request : https://github.com/EFForg/https-everywhere/pull/1206

src/chrome/content/rules/Stack-Exchange.xml
guiweber: Yes, there are indeed cookies set as ```domain=.stackexchange.com```. There are several analytics cookies, but some others do seem important like "usr" and "acct", so it's probably better to leave it like this.

jsha: Right, so the cookies with domain=.stackexchange.com won't get secured under these rules, which is correct. Are there any cookies that *will* get secured under these rules?

jsha:
Thanks for the fix to the matching regex. Please do make sure to test this modified ruleset very thoroughly by hand as well.

Thanks,
Jacob

src/chrome/content/rules/Stack-Exchange.xml
guiweber: Yes, there are cookies that will get secured. I didn't check all subdomains, but I guess it wouldn't be bad to have them all there even if there were no cookies for some of them.

Again, I don't know a lot about cookies but I noticed something that seems odd to me. In many cases, both secure and unsecure cookies are set at the same time for pages secured by HTTPS-E, is this normal? If I disable HTTPS-E and load the same pages, then only unsecure cookies are created. Some cookies also appear never to be secured, even though they should be according to the ruleset.

Is there any documentation explaining how HTTPS-E interacts with cookies, and what the best practices are for creating and testing the securecookie instructions?

PR #1205, opened 2015-03-18
Adding dailyinfo.co.uk

mapkyca:


reedy:
I see this site is loading content via HTTP from 3rd party sites (fonts from Google in this case)

mapkyca:
Gave them a nudge, should be fixed soon :)

Issue #1216, opened 2015-03-21
Google URL redirector no longer rewritten


jsha:
I think this is due to a mistake in #1194 where I removed the /url exception for cctlds, but didn't add an explicit rewrite for it elsewhere.

Issue #1215, opened 2015-03-21
Google ruleset no longer applies to google.co.uk and google.com.au


jsha:
I think this is due to the changes in #1063 and some confusion over how deep google.* matches (https://github.com/hiviah/https-everywhere-checker/issues/13).

StevenRoddis:
hmmm

Would google.co.* and google.com.* work?

jsha:
@StevenRoddis: They would. If you have time, could you revert a378c6bb3afc5bdd1311691a4638e980c33a0e2c to roll-forward your earlier Google changes? Then, before submitting fix this bug by adding google.co.* and google.com.* and any other wildcard targets that were removed: best to do a thorough review of all targets that changed. Also before we submit, we'll need to fix #1216, which I think is my bug, but if you have time to fix it I would be very grateful!

PR #1217, opened 2015-03-21
Revert all Google rules to pre-checker version.

jsha:
There were two major bugs in Google rules related to recent changes:

  - https://github.com/EFForg/https-everywhere/issues/1216
  - https://github.com/EFForg/https-everywhere/issues/1215

The simplest fix is to simply revert these rules to the way they were before
I disabled them and @StevenRoddis rewrote them to comply with new testing and
style guides.

We should roll this change forward again after the 5.0 release and fix the
specific bugs, but this is a quick fix so we can proceed with the release.

cc @StevenRoddis @cooperq for code review.


jsha:
Merging TBR (To Be Reviewed) cc @cooperq @StevenRoddis

cooperq:
LGTM assuming we think these are fully tested

Issue #443, opened 2014-08-18
palemoon compatibility


a923315:
HTTPS Everywhere for Android is not currently compatible with Palemoon Android.

Palemoon is an increasingly popular firefox alternative based on an older firefox version without all the recent "masses against the powerusers" decisions.

A thread on the palemoon forums has identified the incompatibility problem as a setting in the install.rdf file, down to the minVersion and MaxVersion settings.

Quoting JSB2000:
"changed the minVersion to 21.0, the MaxVersion to 23.* and reinserted it into the XPI."
and we have a working version.

Link to palemoon thread:
https://forum.palemoon.org/viewtopic.php?f=39&t=5382&sid=9442e6f5b8e9aa05b17c08e0e58865d4

Hopefully we can bump the maxVersion and solve the problem. Otherwise, please advise any compatibility issues so that the community may solve them.

StevenRoddis:
The current xpi (4.0.3) has
`<em:minVersion>21.0</em:minVersion>`

Looks like it's all fixed.

semenko:
Closing, as this issue is quite old and my no longer be valid. Feel free to reopen if you think it's still a bug.

PR #1220, opened 2015-03-21
Don't block HSTS requests in HTTP Nowhere mode

### LINKED ISSUE #1219: "Block all HTTP requests" disregards HSTS
Steps to reproduce:

1. Visit twitter.com.
2. In HTTPS Everywhere menu, disable Twitter rule (if it's not already disabled).
3. In HTTPS Everywhere menu, enable "Block all HTTPS requests."
4. Open a new tab with twitter.com.

Expected result:

Twitter's built-in HSTS preload sends the site to its HTTPS version and it load successfully.

Actual result:

The site is blocked by "Block all HTTP requests."

diracdeltas:
Adds a simple detector for whether the channel URI has STS state. This will not work in Firefox <26 because of https://bugzilla.mozilla.org/show_bug.cgi?id=887052.

Fix #1219

### DIFF
--- a/src/chrome/content/code/HTTPS.js
+++ b/src/chrome/content/code/HTTPS.js
@@ -1,5 +1,8 @@
 INCLUDE('Cookie');
 
+var securityService = CC['@mozilla.org/ssservice;1']
+    .getService(CI.nsISiteSecurityService);
+
 // Hack. We only need the part of the policystate that tracks content
 // policy loading.
 const PolicyState = {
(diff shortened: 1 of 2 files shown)


diracdeltas:
I noticed in `install.rdf` that we're supporting Firefox versions down to FF 20. Is that still necessary? ESR is now on the FF 31 series.

jsha:
I don't think that's necessary anymore. Want to include a bump up in here?

I asked on #extdev recently about the versions in install.rdf and they said they aren't really checked by anything in recent versions. So I haven't been bothering to update the top end.

diracdeltas:
Bumped to 26!

diracdeltas:
Hmm, wouldn't the version bump break https://github.com/EFForg/https-everywhere/issues/443? Or does that not matter anymore?

jsha:
Pale Moon changed the way it deals with versions in install.rdf so it doesn't matter anymore. Additionally I've decided not to support Pale Moon.

diracdeltas:
I applaud that decision

SJ-Diamond:
Applauding? Only because the Pale Moon guys do not jump anymore on the Mozilla codebase train it does not mean the browser is bad or outdated.

jsha:
@SJ-Diamond: It's more a matter of resources and priorities. We just don't have enough staff to support every possible browser.

nelegalno:
**@jsha** How about you get you're facts straight and implement other peoples patches to do it: http://forum.palemoon.org/viewtopic.php?f=16&t=7287&start=20 ?

Ref #1315

**P.S.** I  wouldn't discard the PaleMoon users so easily if I ware you ...

PR #1222, opened 2015-03-22
Create Steamrep.com.xml

john-kk:


src/chrome/content/rules/Steamrep.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Steamrep.com">
+  <target host="steamrep.com" />
+  <target host="www.steamrep.com" />
+
+  <rule from="^http://(www\.)?steamrep\.com/" to="https://www.steamrep.com/" />
```
semenko: Looks like this actually doesn't work with the full www.steamrep.com -- maybe remove the www prefix?

donced: Just simplify to
```
<rule from="^http:"
    to="https:" />
```

jsha: I think @semenko's suggestion is better: Since only steamrep.com works over HTTPS, both steamrep.com and www.steamrep.com should redirect to the HTTPS version of steamrep.com.

jsha: I think @semenko's suggestion is better: Since only steamrep.com works over HTTPS, both steamrep.com and www.steamrep.com should redirect to the HTTPS version of steamrep.com.

donced: https://www.steamrep.com works fine (i.e. no SSL error). [2d1](https://github.com/EFForg/https-everywhere/pull/2083#discussion_r34428201) prefer not to preempting server's redirect.

jsha: Ah, I see what you mean now. https://www.steamrep.com redirects to
http://steamrep.com, but then that would get rewritten by HTTPS
Everywhere also and you'd wind up on https://steamrep.com. Okay, I'm
convinced your way is better. :thumbsup:

ekaris:
Hi, I'm the author of this pull, I created a new account and I don't use the old one anymore, do I need to create new pull requests for this and the others pulls? Or if not how can I edit this pull with my new account?

donced:
@ekaris if you still can access your old account, I suggest closing this pull and create a new pull under your current account. or else, just create a new pull and reference this pull.

semenko:
Closed in favor of #3211

PR #1223, opened 2015-03-22
Create Dispenser.tf.xml

john-kk:


src/chrome/content/rules/Dispenser.tf.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Dispenser.tf">
+  <target host="dispenser.tf" />
+  <target host="www.dispenser.tf" />
+
+  <rule from="^http://(www\.)?dispenser\.tf/" to="https://www.dispenser.tf/" />
```
reedy: You can (should?) use the new ruleset format here for the rewrite

<rule from="^" to="https:" />

Issue #1224, opened 2015-03-22
meta.statuspage.io gives untrusted certificate


strugee:
StatusPage.io's status page is on Heroku, not their infrastructure. So connecting via TLS gives a certificate name mismatch.

This looks like just an overzealous wildcard. Easy to fix.

strugee:
@jsha does HTTPS Everywhere not have support for excluding certain subdomains? I'm not familiar enough with the ruleset grammar to know... :/

jsha:
@strugee: Yes, you can use the exclusion tag for that, and I'd encourage you to do that to fix the rule. I tend to disable broken rules rather than trying to fix them. It can take up huge amounts of my time trying to fix rules, and I'm the only official maintainer on HTTPS Everywhere. Our awesome community members sometimes fix up and re-enable the rules.

strugee:
@jsha you might consider having a GitHub label for that. Then interested community members could more easily find problems like this :)

PR #1226, opened 2015-03-23
Disable StatusPage rule.

jsha:
Breaks per https://github.com/EFForg/https-everywhere/issues/1224

cc @cooperq @diracdeltas for code review.


jsha:
Merging TBR (To Be Reviewed).

cooperq:
I'm aware that we probably wanted to move fast for the 5.0 release but shouldn't we just add some exclusion rules instead of turning this off?

PR #1230, opened 2015-03-23
Disable Politiken.dk.

jsha:
Per report in IRC, rule breaks streaming video.

Possible fix at https://github.com/EFForg/https-everywhere/pull/1170,
but not tested.

cc @cooperq @diracdeltas for code review.

### DIFF
--- a/src/chrome/content/rules/Politiken.dk.xml
+++ b/src/chrome/content/rules/Politiken.dk.xml
@@ -40,7 +40,7 @@
 	* Secured by us
 
 -->
-<ruleset name="Politiken.dk (partial)">
+<ruleset name="Politiken.dk (partial)" default_off="Breaks streaming video">
 
 	<target host="multimedia.pol.dk" />
 	<target host="politiken.dk" />


jsha:
Merging as TBR (To Be Reviewed).

cooperq:
LGTM

fuglede:
I revamped the rule in #1170, including only the parts where it's clear what it does. Tried to get in touch with the original author without success, unfortunately.

Anyway, my guess is that the site is probably either the most popular or second most popular Danish news site among https-e users so having any kind of covering would be very beneficial.

jsha:
Thanks! Sorry I missed that pull request for so long. Can you merge latest master and confirm that it still fixes streaming video?

fuglede:
No problem; I could have certainly timed that better.

But good thing you asked: they've changed their design recently and it seems like the URL pattern for pages with video streams have changed so that the old fix no longer applies (or I might have simply missed those before). Roughly: before, video pages were all called politiken.dk/tv/..., but now some individual articles also include videos. On the bottom line, the exclusion pattern from my branch fixes the issue on many, but not quite all pages. Seems impossible to fix on mixed content-skeptical platforms though. I'll try to get in touch with a dev and get back.

PR #1239, opened 2015-03-23
Improve bugzilla rule

J0WI:


jsha:
Thanks for this patch! It will need some additional test URLs to pass the tests. See the linked Travis build: https://travis-ci.org/EFForg/https-everywhere/builds/55557038
```
ERROR src/chrome/content/rules/Bugzilla.xml: Not enough tests (0 vs 3) for <Rule from '^http://(www|landfill|lists)\.bugzilla\.org/' to 'https://\g<1>.bugzilla.org/'>
```

PR #1240, opened 2015-03-23
Update kantonalbanken.xml

J0WI:
false positive, works in all major browsers (with extra download).


jsha:
You're correct about the false positive, thanks for catching and fixing that.

This will fail the ruleset coverage test, can you add some test URLs and/or update to match the latest style guide (which will usually add sufficient coverage)?

You can use utils/test-generator.py to assist in writing test URLs.

J0WI:
@jsha
test-generator.py returns ImportError: No module named exrex

jsha:
You need to pip install exrex.

J0WI:
Sadly, the script is nor very helpful in this case. I need much more test urls -.-

J0WI:
fixed important ruleset, please merge soon

PR #1241, opened 2015-03-23
Update swissgov.xml

J0WI:
commented out safe redirect to www. (Browser will correct the url itself).
removed false positive (extra download of issuer cert)


src/chrome/content/rules/swissgov.xml
```diff
@@ -76,12 +70,12 @@ Fetch error: http://gr.ch/ => https://www.gr.ch/: (6, 'Could not resolve host: g
 		to="https://www.jura.ch/"/>
 	<rule from="^http://([^/:@]*)\.jura\.ch/"
 		to="https://$1.jura.ch/"/>
-	<rule from="^http://(?:www\.)?zh\.ch/"
-		to="https://www.zh.ch/"/>
+	<!--<rule from="^http://(?:www\.)?zh\.ch/"
```
jsha: Assuming these really don't exist, please delete rather than commenting out.

J0WI: So we prefer an unsafe redirect?

PR #1242, opened 2015-03-23
Update Raiffeisen.xml

J0WI:
false positive? WFM

### DIFF
--- a/src/chrome/content/rules/Raiffeisen.xml
+++ b/src/chrome/content/rules/Raiffeisen.xml
@@ -1,8 +1,4 @@
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://tb.raiffeisendirect.ch/ => https://tb.raiffeisendirect.ch/: (28, 'Connection timed out after 10000 milliseconds')
--->
-<ruleset name="Raiffeisen.ch" default_off='failed ruleset test'>
+<ruleset name="Raiffeisen.ch">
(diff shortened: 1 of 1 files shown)


jsha:
The HTTP version of tb.raiffeisendirect.ch times out for me. But we've decided for now that it's okay to have a simple rewrite rule for non-existent HTTP sites, so I'll accept this change.

Travis hasn't completed yet, but I think it will fail: There don't appear to be enough tests for the second rule. You can probably change it to a simple http->https rule, or you can add raiffeisendirect.ch as a target host, if it exists.

src/chrome/content/rules/Raiffeisen.xml
```diff
@@ -10,6 +6,7 @@ Fetch error: http://tb.raiffeisendirect.ch/ => https://tb.raiffeisendirect.ch/:
         <securecookie host="^(?:.*\.)?raiffeisen\.ch$" name=".*" />
         <securecookie host="^(?:.*\.)?raiffeisendirect\.ch$" name=".*" />
 
-        <rule from="^http://(?:www\.)?raiffeisen\.ch/" to="https://www.raiffeisen.ch/"/>
```
jsha: The previous rule would send `raiffeisen.ch` to `www.raiffeisen.ch`, but the new one won't. Is that intentional?

J0WI: Yes, it's the same host/content.

PR #1254, opened 2015-03-24
[OpenStreetMap.org] reenable most subdomains, www.openstreetmap.org MCB

julianladisch:


jsha:
Have you tested against the site listed in the linked bug?

julianladisch:
I've tested these sites and they work with this pull request's rulesets:
http://openlayers.org/two/
http://www.openstreetmap.org/ (mixed content, but ruleset has platform="mixedcontent")
https://www.openstreetmap.org/edit?editor=id

PR #1275, opened 2015-03-25
Delete Peertech.org.xml

J0WI:
host does not exist


jsha:
Host works for me. Try again?

Gedankenleser:
I cant rech the Website as well.

J0WI:
port 80 is refused, but 443 works now

PR #1283, opened 2015-03-25
Updates

2d1:
Just enlisting travis, nothing to see here.


jsha:
@2d1: Can you elaborate on what you mean by effects of nonlinear time? If I accidentally re-disabled some rules that you had previously enabled, I'd like to know about it so I can fix.

A spot check shows a few of these are still broken. For instance https://foundry.myfonts.net/ still doesn't resolve and https://yotta.mbed.org/ does time out.

Issue #1297, opened 2015-03-27
Add a block all mixed content mode to Chrome


semenko:
Chrome allows limited mixed content (IIRC only images?).

We could add a mode that blocks *all* mixed content.

domainedma:
Firefox also allows mixed content

domainedma:
Firefox also is allowing limited content https://support.mozilla.org/en-US/kb/how-does-content-isnt-secure-affect-my-safety#w_the-icon-is-a-gray-triangle

"Only the potentially harmful part of HTTP content is blocked so some websites may still have some HTTP content (such as images)"

semenko:
Mechanistically in chrome, we can probably do this with the `frameId` and `parentFrameId` fields in the addListener callback (and can add a custom listener only for images).

mikewest:
You could also opt-into Chrome's built-in (as of Chrome 43) support for https://w3c.github.io/webappsec/specs/upgrade/, which would have more or less the effect you're looking for.

semenko:
Whoa -- cool!

We'd need to inject a CSP header into all HTTPS requests (if I'm reading that spec correctly)? That could be hard, as we might want to eventually drop the https://*/* permissions (and only request http://*/* to reduce our attack surface).

Just noticed the bugs for --enable-strict-mixed-content-checking / https://code.google.com/p/chromium/issues/detail?id=450458 -- which we could suggest to users.

Issue #760, opened 2014-11-26
Add UI to temporarily disable HTTPS Everywhere for Chrome


Superluigi6:
I'm using HTTPS Everywhere for Chrome, and it's great. My only suggestion, is to add a feature that temporarily disables HTTPS Everywhere and turns back on when you tell it too.

diracdeltas:
You can do this by going to `chrome://extensions` and unchecking the "enabled" box for now.

semenko:
Love this idea.

I might implement this as a temporary disable button (e.g. "press to disable HTTPSe for 1 hour") or a toggle that auto-reenables itself whenever chrome is restarted -- so users don't forget and disable the extension for an extended period of time.

jsha:
I like the ideas for making it harder to forget and leave HTTPS
Everywhere off. I'm happy with whichever of these you choose. Let's get
some feedback on the Chrome version and update the Firefox version not
too long after so the behavior and messaging matches.

semenko:
Also worth considering what UI changes we want -- perhaps before pushing a version with HTTP Nowhere mode. 

HTTP Nowhere mode makes the icon to RED.

One option:
Have HTTP Nowhere mode to turn the icon GREEN and have the disabled extension RED

Another:
Have HTTP Nowhere mode stay RED, and have the disabled extension GREY.

jsha:
> Have HTTP Nowhere mode stay RED, and have the disabled extension GREY.

I like this because it matches the current Firefox behavior and I don't
want to change that for fear of confusing people.

Red also makes sense for HTTP Nowhere because it breaks a lot of sites,
and at least in Firefox it's not clear why the site is failing to load.
The red is a nice reminder that HTTPS Everywhere might be causing it.

cooperq:
I think that red is a good color for http-nowhere mode since it 'breaks' non-encrypted sites. It is a good reminder to the user that they have turned on a feature that will break things. I think that the icon should be more distinct though if https everywhere is disabled, since this is a potentially dangerous state. Maybe something like a warning sign on top of it? Or a crossed out sign? Something to strongly indicate to the user that security is turned off. 

Also :+1: for both of these features in https-everywhere for chrome.

Superluigi6:
Wait what is HTTP Nowhere?

reedy:
"Block all HTTP requests"

PR #1304, opened 2015-03-28
[Chrome] Add mode to block all mixed content

semenko:
Add a mode for Chrome to block all mixed content (mostly images & media content). (For issue #1297)

Unfortunately, the webRequest API doesn't provide any parent information (only information about `<iframe>` parents, it seems), so we have to do this manually. We record whether tabs are on HTTPS sites (the main_frame request is https), then block non-HTTPS requests within the same tab.

Give it a test at: https://www.bennish.net/mixed-content.html

One question: What should we do about UX here?

Right now, there's no UX change (except the hidden checkbox inside the popup extension window). It'sp probably worth changing the icon color -- since this will definitely break some sites.

There was a bit of a discussion in #760 -- of note, I'm not 100% sure RED for HTTP Nowhere mode is necessary -- Chrome is /very vocal/ when extensions break entire sites, showing this page:

screen shot 2015-03-28 at 1 38 56 pm

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -162,9 +162,6 @@ function onBeforeRequest(details) {
   // todo: check that this is enough
   var uri = new URI(details.url);
 
-  // Should the request be canceled?
-  var shouldCancel = (httpNowhereOn && uri.protocol() === 'http');
-
   // Normalise hosts such as "www.example.com."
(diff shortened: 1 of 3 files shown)


semenko:
Closing until we figure out issue #1327 (plus Chrome may be adding a flag for this)

PR #1308, opened 2015-03-30
Bug 12218: toolbar_button.js should do more null checks.

MarkCSmith:
Add null checks to avoid various errors that occur when the
HTTPS Everywhere button is not on the toolbar. This also fixes:
  Bug 13550: 'counterItem is null' if no HTTPS Everywhere button.


diracdeltas:
Migrated from https://trac.torproject.org/projects/tor/ticket/12218

(turns out my old trac.torproject.org account still owns a lot of tickets!)

jsha:
Thanks!

PR #1310, opened 2015-03-30
Remove securecookie rule for OkCupid.com

mmaxim:
The securecookie rule for OkCupid.com breaks login for those users that don't have access to https on the site (puts Chrome into an infinite redirect loop). I am the CTO of OkCupid, so feel free to ask any questions, or offer suggestions for how this could be better handled, but I think just removing the rule is a good way to go here.

### DIFF
--- a/src/chrome/content/rules/OkCupid.xml
+++ b/src/chrome/content/rules/OkCupid.xml
@@ -64,7 +64,7 @@
 					-->
 	<!--securecookie host="^\.okcupid\.com$" name="^(__cfuid|cf_clearance|guest)$" /-->
 
-	<securecookie host="^\.okcupid\.com$" name=".+" />
+    <!--<securecookie host="^\.okcupid\.com$" name=".+" /> -->
 
 
 	<rule from="^http:"


jsha:
Hi Mike! Pleased to have you contributing to HTTPS Everywhere. I met my partner on your site!

Since HTTPS doesn't work for some users of the site, I think the better approach would be to disable the entire ruleset, or remove the domains that don't work- i.e., remove `okcupid.com` and `www.okcupid.com`, leaving `blog.okcupid.com`. It's surprising that Chrome gets into an infinite redirect loop. We are supposed to have detection for that situation and automatically disable the rule in the user's browser.

Of course, the *best* way to solve this would be to allow HTTPS for all users on OkCupid. I'm guessing you've considerd this. What are the current obstacles that stand in the way? The larger context of our project is to try and remove those obstacles. Feel free to followup to jsha@eff.org, I'd love to hear from you.

Thanks,
Jacob

mmaxim:
Glad to hear it!

Unfortunately, OkCupid cannot enable https for accounts that see ads. Many of our ad partners do not have secure ad tags at this moment, so it is not practical for us to be turn https on for everyone.

How exactly does that securecookie directive work in Chrome? I think that is the root cause of the whole problem (and honestly I just wanted to make a minimal change here). It seems like it might set the cookie as secure at the browser level (even though the server did not)? If that is the case, then I can see how the redirect detection in the extension wouldn't be able to kick in. I do notice that at some point I stop seeing 307 redirects, and it just starts becoming straight 302s, entirely because Chrome is not sending the session cookie on the non-secure requests.

Issue #1309, opened 2015-03-30
Broken Bitly Branded Short Domain Links


jehiah:
Unfortunately 1c7baccb61d39fcca25d6018d4642c97d8d44565 breaks functionality for some links on Bitly Branded Short Domains (like ift.tt, nyti.ms, etc) because those domains are not completely interchangeable with bit.ly. The only bitly domains that can reliably be redirected from http->https are bit.ly, j.mp and bitly.com

kevinebaugh:
@2d1 Can we get your input on this issue? This is a major bug as all ift.tt links will redirect to their bit.ly counterparts.

For example, our IFTTT newsletter this morning linked to ift.tt/newsletter28. For people running HTTPS Everywhere, they are redirected to bit.ly/newsletter28 instead.

jsha:
Fixed by @kevinebaugh.

PR #1311, opened 2015-03-30
Remove ift.tt from Bit.ly vanity domain ruleset

kevinebaugh:
Ahead of a more permanent fix for #1309, this should fix the issues affecting ift.tt URLs.

### DIFF
--- a/src/chrome/content/rules/Bit.ly_vanity_domains.xml
+++ b/src/chrome/content/rules/Bit.ly_vanity_domains.xml
@@ -49,8 +49,6 @@
 	<target host="www.hub.am" />
 	<target host="huff.to" />
 	<target host="www.huff.to" />
-	<target host="ift.tt" />
-	<target host="www.ift.tt" />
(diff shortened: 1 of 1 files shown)


jsha:
Hi Kevin,

Thanks for spotting and fixing this bug. Could you instead submit a patch that disables the whole ruleset until we get a chance to add enough test URLs to bring this ruleset up to our current testing standard?

Thanks,
Jacob

kevinebaugh:
@jsha I'm in a bit over my head, but I'd be happy to help. Any chance you could point me in the direction of another pull request that disables a ruleset?

Cc'ing @jehiah to see if someone on the bit.ly team could take a look.

jsha:
Hi @kevinebaugh,

Sorry, I should have added more context! Here's an example you can follow: https://github.com/EFForg/https-everywhere/pull/1230/files

kevinebaugh:
Thanks!

PR #1319, opened 2015-03-31
add rules for lebsanft.org

piratenpanda:


src/chrome/content/rules/lebsanft.org.xml
```diff
@@ -0,0 +1,13 @@
...
+  <target host="*.lebsanft.org" />
+  <securecookie host="^www\.lebsanft\.org$" name=".*" />
+  <exclusion pattern="^http://physik\.lebsanft\.org/"/>
+
+  <test url="http://lebsanft.org/" />
+  <test url="http://physik.lebsanft.org/" />
+  <test url="http://www.lebsanft.org/" />
+
+  <rule from="^http://lebsanft\.org/" to="https://lebsanft.org/"/>
+  <rule from="^http://www\.lebsanft\.org/" to="https://www.lebsanft.org/"/>
```
reedy: You should just use the updated/simplified rule format here

```
```

You could then just explicitly only target www and no subdomain, and remove the exclusion. Simplies the ruleset somewhat

piratenpanda:
Hopefully this is how it should be

reedy:
I wouldn't have deleted the test urls, nor the secure cookie probably (unless it's unused?).

This ruleset now doesn't secure http://lebsanft.org/ either - is this intended?

piratenpanda:
Oops let me fix that

piratenpanda:
Do I need a rule when the folder without www is just a symlink to www. Will this work for https-everywhere just fine? If yes, then I can just omit it.

reedy:
The server adds the www with a 301, so that's fine

```
chronos@localhost / $ curl -I http://lebsanft.org
Date: Fri, 17 Apr 2015 13:33:32 GMT
Server: Apache/2.2.15 (CentOS)
X-Powered-By: PHP/5.4.4
X-Pingback: http://www.lebsanft.org/xmlrpc.php
Location: http://www.lebsanft.org/
Connection: close
Content-Type: text/html; charset=UTF-8

chronos@localhost / $
```

piratenpanda:
then I guess this can be merged

PR #1333, opened 2015-04-02
New and updated rules

KommX:


install-dev-dependencies.sh
```diff
@@ -4,7 +4,7 @@
 set -o errexit -o xtrace
 if type apt-get >/dev/null ; then
   sudo apt-get install libxml2-dev libxml2-utils libxslt1-dev python-dev \
-    firefox chromium-browser zip sqlite3 python-pip libcurl4-openssl-dev
+    firefox chromium-browser zip sqlite3 python-pip libcurl4-openssl-dev libssl-dev
```
reedy: Out of interest, why do you need to add libssl-dev in?

@jsha I'm guessing you're ok with this change?

KommX: On a Ubuntu system (that I was running in a VM) I had the problem that running install-dev-dependencies.sh ended in a compiler error not finding "crypto.h" (or something similar). Installing libssl-dev solved this problem for me.

But on another machine running Ubuntu I can't reproduce this right now?!

KommX:
I will close this and open up separate pull request.

install-dev-dependencies.sh
jsha: The machine on which you can't reproduce - does that already have libssl-dev?

I'd prefer to not add the dependency unless we're sure it's needed.

KommX: Now I can't reproduce anymore.
We should just let this one go.

PR #1336, opened 2015-04-02
Add additional rules for Open Computing Facility

chriskuehl:
This adds a few additional rules for the Open Computing Facility at UC Berkeley.


chriskuehl:
I've started a substantial rewrite of the UC Berkeley rules into what my understanding of the preferred format is (more specific targets, less complicated regular expressions). I'm working out a few kinks and a couple of outdated rules, and will follow up with another PR shortly with the changes.

I'm closing this one for now since it can't be merged until these are fixed.

jsha:
Great, thanks!

PR #1338, opened 2015-04-02
Add andrewmunsell.com Rules

andrewmunsell:
Adds `www.andrewmunsell.com` and `andrewmunsell.com` to HTTPS Everywhere. HSTS is also enabled for these two domains.

### DIFF
--- a/src/chrome/content/rules/Andrew-Munsell.xml
+++ b/src/chrome/content/rules/Andrew-Munsell.xml
@@ -1,0 +1,9 @@
+<ruleset name="Andrew Munsell">
+
+	<target host="andrewmunsell.com" />
+	<target host="www.andrewmunsell.com" />
+
+	<rule from="^http://(?:www\.)?andrewmunsell\.com/"
+		to="https://www.andrewmunsell.com/" />
+
+</ruleset>
--- a/src/chrome/content/rules/Andrew-Munsell.xml
+++ b/src/chrome/content/rules/Andrew-Munsell.xml
@@ -3,7 +3,7 @@
 	<target host="andrewmunsell.com" />
 	<target host="www.andrewmunsell.com" />
 
-	<rule from="^http://(?:www\.)?andrewmunsell\.com/"
+	<rule from="^http://(www\.)?andrewmunsell\.com/"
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Andrew-Munsell.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Andrew Munsell">
+
+	<target host="andrewmunsell.com" />
+	<target host="www.andrewmunsell.com" />
+
+	<rule from="^http://(?:www\.)?andrewmunsell\.com/"
+		to="https://www.andrewmunsell.com/" />
```
reedy: You should probably use the new ruleset format for these

<rule from="^http:"
to="https:" />

jsha: Note that this one is specifically rewriting the non-www form to the www form. For those I've generally been fine with this style.

I will note that current style is to do regular capturing groups for better readability: `^http://(www\.)?andrewmunsell\.com/`

reedy: https://www.andrewmunsell.com and https://andrewmunsell.com both work fine in testing. That, and the server 301s the non www version to www too

andrewmunsell:
I updated the capture group style to reflect your earlier comments. Would it be preferential to change the rules to be shorter so that they look like the following?:

```
<ruleset name="Andrew Munsell">

	<target host="andrewmunsell.com" />
	<target host="www.andrewmunsell.com" />

	<rule from="^http://(www\.)?"
		to="https://www." />

</ruleset>
```

It would be slightly shorter, but it would also sacrifice some readability.

reedy:
I've wondered why we've not got other rules doing that in these simple cases that want to force the www. Presumably it's possible, but I'm not sure what the reason for not doing it - presumably readability (disk space saved here doesn't seem advantageous ;)). But a question for the higher ups :)

jsha:
I agree that the alternate form is somewhat less reasonable, although if it became idiomatic we would probably start finding it easier to read. But let's leave it as-is for now.

Issue #1335, opened 2015-04-02
Update URI.js


semenko:
We haven't updated this in a while -- and a bunch of performance bugs are fixed.

@jsha could you update this? (As it's a big, not-so-easily diffable blob.)

Via: https://github.com/medialize/URI.js

We want: https://github.com/medialize/URI.js/blob/gh-pages/src/URI.min.js

semenko:
Hm, alternatively we could drop URI.js and switch to IDL attributes:
https://jsperf.com/idl-attributes-vs-uri-js

Let me ponder.

jsha:
IDL attributes sound good too. I will wait for your conclusions.

If we stick with URI.js I'd like to commit the unminified version so that future diffs are meaningfully auditable. We could either minify it ourselves during the release process or decide that the infrequent downloading of the extension means we can simply not minify.

semenko:
I'm thinking IDL, since that's standard blazing-fast code we can forget about updating / auditing, etc. PR inbound in a bit here.

PR #1340, opened 2015-04-03
[Chrome] drop URI.js

semenko:
Remove URI.js (which we keep forgetting to update) and default to the better tested & higher performance IDL attributes (see also issue #1335).


jsha:
Thanks!

PR #1345, opened 2015-04-03
oops.kernel.org does not support HTTPS

mineo:


jsha:
Hi @mineo! Thanks for the fix. We're asking for some additional help with each changed ruleset, in order to bring our ruleset library up to the latest testing standards. In particular we're asking that all modified rulesets be brought up to date with regards to https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md.

If you have time to make these changes, awesome! If not, please let us know and one of us can incorporate this fix into an overhaul of this ruleset.

mineo:
I've added some test URLs and travis is now happy :-)

One thing I've noticed is that the instructions in the file you linked to don't work,

>  python2.7 https-everywhere-checker/src/https_everywhere_checker/check_rules.py https-everywhere-checker/checker.config.sample rules/Example.xml

fails for me with

    DEBUG Loaded certificate platforms: 
      File "https-everywhere-checker/src/https_everywhere_checker/check_rules.py", line 404, in <module>
        sys.exit(cli())
      File "https-everywhere-checker/src/https_everywhere_checker/check_rules.py", line 288, in cli
        raise RuntimeError("Platform 'default' is missing from certificate directories")
    RuntimeError: Platform 'default' is missing from certificate directories

Using `manual.checker.config` worked, but I'm not sure what the difference between them is. The `requirements.txt` file of the checker also seems to be missing the dependency on lxml (`tox.ini` on the other hand lists it explicitly, as does `.travis.yml`), so I had to install that manually as well. Not sure if you want separate tickets for those or if they're just quick fixes for you.

jsha:
Thanks for the test URLs, and for the bug reports. You're right, they're quick fixes-- I'll put a branch up shortly. You can now run tests more easily with:

```
./fetch-test.sh rules/Example.xml
```

PR #1350, opened 2015-04-03
Update bitly vanity domain comment.

jsha:
Remove sched.co and explain recurring problem with domains that don't point at
bit.ly anymore.


cooperq:
lgtm

Issue #1207, opened 2015-03-19
Use Firefox's translation strings in Chrome version


jsha:
Chrome mostly reuses strings from Firefox. We should remove the explicit messages.json and automatically generate it from the Firefox strings at build time, so we benefit from the Firefox localization.

reedy:
:+1:

reedy:
This falls somewhat under the purview of #1681

jsha:
I've got a branch that fixes this (linked), but need to finish testing it.

J0WI:
@jsha can we close this since #1351 is merged?

jsha:
Yep, thanks!

PR #1351, opened 2015-04-03
Share translations between Chrome & FF

jsha:
Most of the messages for i18n are the same between Chrome and Firefox. This automates turning the Firefox local files into Chrome ones at build time, so Chrome will always have the same translations FF does, and the benefit of the larger translation community.

This removes one message that was unique to Chrome ("What is this?"), replacing it with "About HTTPS Everywhere." It also copies four other unique-to-Chrome messages into the FF locale files, preserving the already-translated copies of those.

cc @semenko @cooperq @diracdeltas


jsha:
Also, fair notice: I haven't actually tested these translations because I couldn't figure out how to set the interface language in Chromium. Help appreciated!

semenko:
Cool -- will test more in a bit. IIRC you can change the language at chrome://settings/languages

jsha:
Unfortunately chrome://settings/languages only changes my Accept-Language header, it doesn't seem to change the interface language. This is on Linux with Chromium and Chrome, both. Does it work differently for you?

semenko:
Hm, good call. I got this to work exporting a LANG= variable on Linux. I also set a temporary user profile directory:

`LANGUAGE=es ./google-chrome-unstable --user-data-dir=/tmp/profile-test-ES`

The `--lang` flag suggested at https://developer.chrome.com/extensions/i18n didn't seem to work for me either.

jsha:
Awesome, this worked, thanks! Here's a neat trick to make it easier:

`LANGUAGE=es ./run-chromium.sh`

Builds the extension and runs it in a temporary profile.

It seems to be missing the "Stable rules" / "Experimental rules" translations for ES at least. That sort of makes sense - they haven't been translated - but ideally it should have the English strings there rather than blank space. I'll take a look next week.

jsha:
Fixes https://github.com/EFForg/https-everywhere/issues/1207

cooperq:
Can you resolve the conflicts in this PR?

jsha:
Okay, brought this branch up-to-date with the latest code.

Note: The newly added strings (e.g. "Stable rules") will show up as blank in non-English languages at the moment. Once this change hits master, Transifex will pick up the new strings and add them to all locales. So, once we update the `translations` submodule, we'll at least have English strings for those. And, as volunteers translate them and we continue to update the `translations` submodule, they will become available in the relevant languages.

semenko:
I'm seeing blanks for "Stable Rules" in Chrome. Looks like the HTML is prefixed with `chrome_` e.g. `chrome_stable_rules` but messages.json provides `stable_rules`.

screen shot 2016-02-01 at 10 33 20 am

jsha:
Fixed, thanks.

semenko:
FYI, I think this broke the "Add new rule" button in chrome (because it wasn't prefixed by chrome_). I might fix in a bit.

hilalgruopcompanies:
HILAL GROUP COMPANIES
Live animals meat and Skin, Intestines food and petroleum products trade Inc.
Viber – WhatsApp +905326512082 
Email : hilalcasing@mail.ru hilalcasing@gmail.com  hilalmeat.kz@yandex.com 
Web: http://hilalmeatkz.wix.com/hilalgruopcompanies 
 Skype: hilalcasing1968

PR #1370, opened 2015-04-06
Add strugee.net

strugee:


strugee:
I should note that I maintain strugee.net, so I can guarantee this ruleset is correct.

PR #1371, opened 2015-04-06
Removing broken How-To Geek rule

howtogeek:
How-To Geek does not support HTTPS at this time, and this rule breaks images on the site.


jsha:
Since it looks like there are some URLs that work, I'd prefer adding a default_off="Breaks images" attribute to the ruleset tag.

jsha:
Since it looks like there are some URLs that work, I'd prefer adding a default_off="Breaks images" attribute to the ruleset tag.

howtogeek:
There are no URLs that work correctly. This is a completely broken rule.

Every URL that does respond to HTTPS just redirects you to HTTP. 

I am working toward eventually possibly enabling HTTPS but this rule is causing a huge problem for my eventual implementation. Just because we bought a certificate doesn't mean the site actually works.

If you continue this, I will be forced to disable HTTPS everywhere since it's quite unnecessary for a static content website.

jsha:
Apologies, I was reviewing pull requests in a hurry and didn't see that you are in fact the site owner.

Part of the reason we like to default_off rulesets is that if someone else comes along and tries to write a rule, they can see what the history is. For instance, in this case you might want to turn off the ruleset and add a comment saying "Per site owner, this site doesn't yet support HTTPS. Please don't re-enable until requested." That way if another ruleset author makes a similar mistake and thinks your site supports HTTPS, there will be a clear indication that it doesn't.

Thanks,
Jacob

howtogeek:
That seems reasonable then. As long as the rule gets turned off, we're getting a ton of complaints from readers.

PR #1379, opened 2015-04-07
Enhance danbooru.donmai.us.xml

DeathlyX:


jsha:
Current style recommends, instead of the three `<rule>` tags:

```
```

DeathlyX:
All right, I'll keep that in mind. It still needs one additional rule though to redirect.

PR #1381, opened 2015-04-07
Disable rules and rename "HTTP Nowhere"

jsha:
Rules are disabled per user reports.

cc @cooperq for code review.


cooperq:
lgtm

PR #1382, opened 2015-04-07
Modify makexpi to build EFF and AMO versions.

jsha:
AMO version doesn't include updateKey and updateURL.

cc @cooperq for review.


cooperq:
looks good

PR #1390, opened 2015-04-08
Update LastPass.xml

DeathlyX:


src/chrome/content/rules/LastPass.xml
```diff
@@ -1,11 +1,8 @@
...
   <target host="lastpass.com" />
   <target host="*.lastpass.com" />
 
   <securecookie host="^(?:.+\.)?lastpass\.com$" name=".*"/>
 
-  <rule from="^http://(?:www\.)?lastpass\.com/" to="https://lastpass.com/"/>
-
-	<rule from="^http://forums\.lastpass\.com/"
-		to="https://forums.lastpass.com/" />
+  <rule from="^http:" to="https:" />
```
jsha: The old version of this rule rewrote only forums and www, but the new version will rewrite all subdomains. I suggest you should probably keep the old semantics unless you have a very strong reason not to.

That means changing the target host = "*.lastpass.com" to target host = "forums.lastpass.com", target host = "www.lastpass.com".

DeathlyX: I've done that deliberately. All the other sub-domains worked too apparently. The certificate name of forum.lastpass.com for instance says it applies to *.lastpass.com.

jsha: Which other subdomains have you tested? You should test at least three, preferably more (you can often find subdomains using Wolfram Alpha). You should also add those subdomains as `<test url=... />`.

DeathlyX: I've actually used Wolfram Alpha to find the sub-domains. I've tested all the sub-domains which were listed and they all worked. One sub-domain, download-do.lastpass.com, gave a 404, but it has nothing to do with http or https.

DeathlyX: I've added some <test url=... /> lines.

DeathlyX:
I can remove the wildcard if really necessary, but I believe it's justified. Should I change the ruleset?

Issue #1393, opened 2015-04-09
Breaks HBO GO


cweagans:
Videos won't play on HBO GO when HTTPS Everywhere is enabled in Chrome 42.0.2311.68 beta (64-bit) on Mac OS Yosemite 10.10.3.

reedy:
What version of HTTPS Everywhere?

cweagans:
https-everywhere-chrome-2015.4.7.crx (== 5.0.2): Broken
https-everywhere-chrome-2015.3.31.crx (== 5.0.1): Broken
https-everywhere-chrome-2015.3.23.crx (== 5.0): Broken
https-everywhere-chrome-2015.1.22.crx (== 4.0.3): Works

reedy:
Thanks for the info/testing.

Looks like the HBO ruleset was new in 5.0. Do you see any relevant errors in the browser console? And/or Chrome blocking mixed content?

Ping @2d1

jagtalon:
I think I have a related fix here: https://github.com/EFForg/https-everywhere/pull/1440. Looks like @jsha has fixed it as well but maybe `platform="mixedcontent"` is more apt. :) Tell me what you think!

Schmactor:
I don't know if this helps, but it seems the issue is with HTTPS Everywhere encrypting Conviva.  Unchecking Conviva under the HTTPS Everywhere "force encrypted" settings while on HBOGO seems to fix the problem.

Issue #1387, opened 2015-04-08
Huffington Post - Top Headline Image Not Displayed


garyploski:
Using the latest version of Chrome the top headline image does not load.

ENABLED
![screen shot 2015-04-07 at 10 14 29 pm](https://cloud.githubusercontent.com/assets/4456404/7039238/857cce52-dd73-11e4-849d-97801ccbb6bd.png)

DISABLED
![screen shot 2015-04-07 at 10 15 33 pm](https://cloud.githubusercontent.com/assets/4456404/7039248/a92f4974-dd73-11e4-8bb4-e1ea38723f7c.png)

semenko:
Exists:
http://i.huffpost.com/gen/2811624/images/r-WALTER-SCOTT-huge.jpg

Doesn't (403)
https://s-i.huffpost.com/gen/2811624/images/r-WALTER-SCOTT-huge.jpg
[neither does https://i.huffpost.com/gen/2811624/images/r-WALTER-SCOTT-huge.jpg ]

PR #1396, opened 2015-04-09
Misc fixes.

jsha:
Formatting in rulesets.html (anchors go inside headings).
Makexpi package name corrections.
Disable HBO and HuffPo.

Fixes https://github.com/EFForg/https-everywhere/issues/1393
Fixes https://github.com/EFForg/https-everywhere/issues/1387

@cooperq for review


reedy:
#1440 marks HBO as mixedcontent instead of disabling it completely

jsha:
Merged master so now HBO is just mixedcontent rather than disabled. @pde can you review? or @reedy?

makexpi.sh
```diff
@@ -44,14 +44,9 @@ if [ -n "$1" ] && [ "$2" != "--no-recurse" ] && [ "$1" != "--fast" ] ; then
 
   # Now escape from the horrible mess we've made
   cd ..
-	XPI_NAME="$APP_NAME-$1"
-  # In this mad recursive situation, sometimes old buggy build scripts make
-  # the xpi as ./pkg :(
-  if ! cp $SUBDIR/pkg/$XPI_NAME pkg/ ; then
```
pde: Are you eliminating this case on the theory that it couldn't possibly have been working and doing anything beforehand?

makexpi.sh
```diff
@@ -44,14 +44,9 @@ if [ -n "$1" ] && [ "$2" != "--no-recurse" ] && [ "$1" != "--fast" ] ; then
...
-    echo Recovering from hair-raising recursion:
-    echo cp $SUBDIR/pkg pkg/$XPI_NAME
-    cp $SUBDIR/pkg pkg/$XPI_NAME
-  fi
+  XPI_NAME="$APP_NAME-$1"
+  cp $SUBDIR/pkg/$XPI_NAME.xpi pkg/
```
pde: You've added ".xpi" to the copied filename here.  Was that buggy beforehand?

pde: Oh I see, it's a fix for a bug that was introduced in commits 6a7b8761bf1167a78049073f6452deec7bb20a6c and 55a7ab05f334ed50dfdfdd15ebc2cf8940882e04...

makexpi.sh
pde: This may break some historical builds from circa 2012; not sure if that matters...

pde:
I think there's another bug in makexpi.sh, the third last line <tt>cp $SUBDIR/$XPI_NAME pkg</tt> is probably missing a ".xpi".  Other than that, I think this is good to merge.

makexpi.sh
jsha: I'm eliminating this case because if I updated it I would need to test it, and the case is too obscure and weird to bother testing. So instead I'm deciding that if we want to make builds from 2012 we will have to check out the correct tag by hand and build using that version of makexpi.sh.

jsha:
Updated makexpi.sh to include that and a handful of other fixes. Merging TBR (To Be Reviewed). Would appreciate a final run-through.

PR #1403, opened 2015-04-10
New rule for cfos.de

KommX:


KommX:
I just noticed that it breaks some sites. So, this one shouldn't be merged.

PR #1404, opened 2015-04-10
additional subdomains for freecycle.org

KommX:


src/chrome/content/rules/Freecycle.xml
```diff
@@ -4,10 +4,13 @@
...
+	<target host="my.freecycle.org"/>
+	<target host="www.freecycle.org"/>
+	<target host="freecycle.org"/>
+	<target host="groups.freecycle.org"/>
 
 	<securecookie host="^(?:web)?mail\.freecycle\.org$" name=".*"/>
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http:"	to="https:" />
```
reedy: Any reason to change this to a single line? Generally we don't "care" too much for stylistic changes like this

KommX: No, no particular reason. I just saw that 'make-trivial-rule' does it like that. The main part are the additional subdomains here.

src/chrome/content/rules/Freecycle.xml
```diff
@@ -4,10 +4,13 @@
 	<target host="butler.freecycle.org"/>
 	<target host="mail.freecycle.org"/>
 	<target host="webmail.freecycle.org"/>
+	<target host="my.freecycle.org"/>
+	<target host="www.freecycle.org"/>
+	<target host="freecycle.org"/>
```
reedy: https://freecycle.org/ gives a 403

KommX:
Yes, you're right. I now removed the target host 'freecycle.org'.

PR #1406, opened 2015-04-10
New rule for forecast.io

KommX:


reedy:
Thanks for doing all these as individual pull requests. Makes reviewing and merging individual ones much easier! :+1:

Issue #1412, opened 2015-04-10
FF 37 doesn't connect to https://www.deutschepost.de/


KommX:
It works in Chrome, but I think we should disable this rule (temporarily).

julianladisch:
I don't have any problems with https://www.deutschepost.de/ using my Firefox 37.0.1 for Windows. The activated deutschpost rule works just fine for me. Please explain in more detail how to reproduce the problem.

KommX:
That's interesting.

I am also on FF 37.0.1 and I have tested with Windows and Linux. I just visit https://www.deutschepost.de/ and get the following error message:
![deutschepost de](https://cloud.githubusercontent.com/assets/7859245/7091695/16c07d98-dfac-11e4-99ec-6b4ffce46c7f.png)

jsha:
I get the same exact error. It could be that we are hitting different frontends or load balancers, and only one is broken?

julianladisch:
Yes.

It is a certificate problem that will be fixed soon:

https://groups.google.com/forum/#!msg/mozilla.dev.security.policy/nEM8cfLf324/Lgs6rjXJEOIJ

PR #1415, opened 2015-04-10
Disabled rule for deutschepost.de, conflicts with FF37

KommX:
This addresses #1412.


src/chrome/content/rules/deutschepost.xml
```diff
@@ -1,5 +1,5 @@
-<ruleset name="deutschepost">
+<ruleset name="deutschepost" default_off="FF37+ doesn't connect">
```
jsha: Please add a link to the issue in a comment above the ruleset, so future editors can find out more details.

KommX: Done.

reedy:
Already disabled

PR #1420, opened 2015-04-10
New rulesets for Belgium banks

DeathlyX:
These new rulesets will mostly secure the sites against SSL stripping. Occasionally it turns an unencrypted page, like a blog subdomain, to its encrypted counterpart. I didn't secure the cookies though, since I'm not 100% familiar with that.


DeathlyX:
Maybe I pushed a bit to early on second consideration. I'll refine the rulesets and push the result later.

PR #1424, opened 2015-04-11
New rule for detektor.fm

KommX:


reedy:
https://www.detektor.fm/ gives an SSL cert error as the cert is issued to webserver.ispgateway.de

On either site, there's a load of mixed content but the site looks to work fine otherwise

KommX:
Ok, I removed www.detektor.fm from the ruleset. Should be fine now?!

PR #1425, opened 2015-04-11
Reenabled bahn.de

KommX:


reedy:
They WFM in a browser.. Did you test them via https-everywhere-checker? Just out of interest...

KommX:
What exactly do you mean?
I ran install-dev-dependencies.sh, so that test.sh gets executed on every push.

Manually running fetch-test.sh yields the same error again:
Fetch error: http://fahrkarten.bahn.de/ => https://fahrkarten.bahn.de/: (7, 'Failed to connect to fahrkarten.bahn.de port 80: Connection refused')
but according to this discussion on the mailing list, they can be ignored:
https://lists.eff.org/pipermail/https-everywhere/2015-April/002486.html

reedy:
Right, same as #1435

PR #1426, opened 2015-04-11
arbitrary.ch: initial addition

pgerber:
add domain arbitrary.ch


reedy:
Does this site do anything?

```
Intentionally left blank!
```

pgerber:
There are a few subpages. But they are not open to the wide-public, at least for now. Still I'd like to have my domain covered by HTTPS Everywhere rules before actually any public content goes online. After all, foresight is always better than hindsight.

Issue #1249, opened 2015-03-24
Question about rulesets


cypherpunk:
If a site is in https by default, should it still be added as a ruleset?

jsha:
If the site has HSTS and the cookies are all marked secure: no.

Otherwise: yes, but lower priority to implement than sites that don't use HTTPS by default.

cypherpunk:
Should I report sites that have https, but don't load cookies until the user have logged in and the cookies are therefore unchecked?

PR #1430, opened 2015-04-11
Various rulesets for Belgium banks

DeathlyX:
This pull request contains some various rulesets for some mayor Belgium bank. I didn't include those who have HSTS enabled. This is mainly as a protection against SSL stripping.

I've also deleted rabobank.nl (Netherlands) since it includes HSTS.


anand-bhat:
The first request to a website with HSTS enabled (unless it is included in the browser's preload list) may still be done over HTTP, giving the potential for an attacker to perform an SSL strip attack. I believe HTTPS Everywhere needs to include domains that have HSTS to rewrite the first HTTP request into HTTPS even before it gets sent out.

The only case where a domain may not need to be added to HTTPS Everywhere is if it supports HSTS *and* is included in the browser's HSTS preload list.

DeathlyX:
You've got a point, but I thought @jsha mentioned in #1249 not to add those if the cookies are already secured. Shouldn't it be better to submit them to the HSTS preload list?

anand-bhat:
Ah, I missed seeing that. Considering that I'm not affiliated to EFF or to this extension besides being a user/ contributor, I'll defer this to @jsha and/ or the other project members..

PR #1432, opened 2015-04-11
Update University-of-Exeter.xml

Deiz:


src/chrome/content/rules/University-of-Exeter.xml
```diff
@@ -26,6 +30,10 @@
 	<securecookie host="^(?:.*\.)?exeter\.ac\.uk$" name=".+" />
 
 
+	<!-- Cert only matches empslocal.ex.ac.uk -->
```
reedy: I presume the comment is meaning it doesn't match empslocal.exeter.ac.uk (from testing), but I'm not sure what the reason for adding this is for.. The site forces HTTPS anyway it seems. Did you mean to rewrite from exeter to ex?

src/chrome/content/rules/University-of-Exeter.xml
```diff
@@ -26,6 +30,10 @@
...
+	<rule from="^http://empslocal\.ex\.ac\.uk/"
+		to="https://empslocal.ex.ac.uk/" />
+
 	<rule from="^http://newton\.ex\.ac\.uk/"
 		to="https://newton.ex.ac.uk/" />
```
reedy: Looks like this one should rewrite to newton.exeter.ac.ukas the cert is only for the full version...

```
This server could not prove that it is emps.ex.ac.uk; its security certificate is from *.exeter.ac.uk. This may be caused by a misconfiguration or an attacker intercepting your connection.
```

src/chrome/content/rules/University-of-Exeter.xml
Deiz: No, I'm just preempting the later rewrite so that http://empslocal.ex.ac.uk doesn't turn into https://empslocal.exeter.ac.uk/. Is there a preferred way to do that?

src/chrome/content/rules/University-of-Exeter.xml
Deiz: Ah, so presumably newton.exeter.ac.uk should be handled by the subsequent rule, since if it should always be rewritten to exeter, it's not a special case.

src/chrome/content/rules/University-of-Exeter.xml
reedy: Ah. Makes more sense. Maybe expand the comment slightly?

With regards to my comment about the newton.ex.ac.uk below... I wonder if that can be removed, and noted in the observed subdomain

src/chrome/content/rules/University-of-Exeter.xml
reedy: Right, AFAIS :+1

Deiz:
Alright, nixed the newton-specific rule, expanded the comment and made empslocal rule match `ex(?:eter)?` so that it also catches links to http://empslocal.exeter.ac.uk/ without causing certificate errors.

PR #1434, opened 2015-04-11
Make trivial rule for resin.io.

jish:


reedy:
Looks like they "force" HTTPS anyway...

PR #1439, opened 2015-04-12
Reactivated TED.com ruleset

KommX:


jsha:
Have you confirmed that videos now work with this rule? Also on third-party sites?

KommX:
The videos on ted.com work. I haven't tested on third-party sites because I don't know where to test. The only location known to me was linked in an old mail to the mailing list:
https://blog.mozilla.org/blog/2012/05/03/ted-u-talk-gary-kovacs-tracking-the-trackers/

But there the video doesn't work with Firefox or Chrome, independently from the ruleset.

jsha:
Please do find some working examples of third-party TED embeds before re-enabling this ruleset. You could try by tweeting at the TED organizers, or perhaps look through the BoingBoing archives. I think they sometimes embed TED.

KommX:
I found a test case:
http://blog.ted.com/teds_embeddable/

The video works, but with the ruleset of this pull request I get an error message regarding the subtitles.

numismatika:
I have been working on that file too in #1512
have a  look at https://github.com/EFForg/https-everywhere/blob/346aa62f0a4f3d369ac3ca9661f26ebd86868061/src/chrome/content/rules/TED.xml 
for some additional lines

PR #1452, opened 2015-04-14
New rule for sky.de

KommX:


ridwande:
Please add subdomain https://www.skygo.sky.de/ in new rule

PR #1464, opened 2015-04-15
Delete 2020mobile.es.xml

J0WI:
Miss configured Server


jsha:
I think it's sufficient to leave this host in default_off mode. If they later fix their TLS config, someone can come along and re-enable it rather than starting from scratch.

PR #1468, opened 2015-04-15
Delete 99.se.xml

J0WI:
host does not support https anymore :'(


reedy:
Per other commits, should this be changed to just disabling it, rather than deleting it?

jsha:
My general rule of thumb has been that configuration errors (self-signed cert, expired, etc) generally warrant disablement. For sites that stop answering on 443, I don't have a firm rule: If I get a pull request deleting the rule or disabling it, I'll generally accept either. Definitely open to feedback on the right thing to do. We probably don't want dead rules to bloat the extension forever.

PR #1469, opened 2015-04-15
Delete 95Bet.com.xml

J0WI:
host does no longer exist


reedy:
If the site was still up, but no HTTPS, disabling would seem a sensible option. But if it's gone completely... Deleting seems sane?

PR #1476, opened 2015-04-15
Delete Symbian.xml

J0WI:
host does no longer support https :'(


reedy:
Per other commits, should this be changed to just disabling it, rather than deleting it?

jsha:
FYI I'm being fairly loosey-goosey on disable vs delete on rulesets that don't work anymore, and generally leaving it up to the submitter unless there is a lot of embedded knowledge in the ruleset.

PR #1482, opened 2015-04-15
Update myGov.xml

J0WI:
fixed


StevenRoddis:
@jsha has the checker been fixed so it won't barf on [Ll]ocation headers?

jsha:
The Location bug is in theory fixed, although I've found a few weird cases where the checker still can't find the location header, even though it is present. If you'd like to help debug, you can run:

```
python2.7 https-everywhere-checker/src/https_everywhere_checker/check_rules.py https-everywhere-checker/disable-broken-rulesets.checker.config
```

And look for complaints about missing Location in the diffs.

PR #1489, opened 2015-04-15
Update Anonbox.net.xml

J0WI:
cacert

### DIFF
--- a/src/chrome/content/rules/Anonbox.net.xml
+++ b/src/chrome/content/rules/Anonbox.net.xml
@@ -1,17 +1,6 @@
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://anonbox.net/ => https://anonbox.net/: (60, 'SSL certificate problem: self signed certificate in certificate chain')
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Anonbox.net.xml
```diff
@@ -1,17 +1,6 @@
...
-Fetch error: http://anonbox.net/ => https://anonbox.net/: (60, 'SSL certificate problem: self signed certificate in certificate chain')
-Fetch error: http://www.anonbox.net/ => https://anonbox.net/: (60, 'SSL certificate problem: self signed certificate in certificate chain')
-	Cert only matches ^anonbox.net.
-
--->
-<ruleset name="anonbox.net" platform="cacert" default_off='failed ruleset test'>
-
+<ruleset name="anonbox.net" platform="cacert">
 	<target host="anonbox.net" />
-	<target host="www.anonbox.net" />
```
reedy: Why is the www host being removed?

J0WI: The certificate is only valid for anonbox.net

PR #1496, opened 2015-04-16
New rule for phoronix.com

KommX:


reedy:
There's some mixed content stuff from offsite, but the site seems to work fine anyway

gitarra:
This breaks the forum styling due to mixed content.

See here:

https://www.phoronix.com/forums/showthread.php?117228-QtWebEngine-Poses-Problems-For-Debian-Distribution-Vendors&p=485227#post485227

PR #1506, opened 2015-04-17
Disable Razer due to broken redirect

chriskuehl:
razerzone.com redirects https -> http, creating a redirect loop with the existing rule.


DamonGant:
I have contacted Razer and they escalated this issue internally promising to take a look at fixing it.

PR #1513, opened 2015-04-19
Update for disqus.com ruleset

DeathlyX:
I've changed the unreliable ruleset with the wildcard into a partial ruleset. It seems to work for me at the moment, but since I don't have an account there I can't test things like the login process.

### DIFF
--- a/src/chrome/content/rules/Disqus.com.xml
+++ b/src/chrome/content/rules/Disqus.com.xml
@@ -4,6 +4,9 @@
 	Nonfunctional domains:
 		- blog.disqus.com
 		- community.disqus.com
+		
+	Subdomains like SITENAME.disqus.com aren't covered in this ruleset.
+	Probably some other subdomains aren't covered either.
 -->
 <ruleset name="Disqus (partial)">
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Disqus.com.xml
```diff
@@ -1,52 +1,32 @@
...
-
-		<exclusion pattern="^http://disqus\.com/next/lounge/client\.html" />
-		<exclusion pattern="^http://\w+\.disqus\.com/\d+/build/system/embed\.js"/>
-
-			See comments at top.
-							-->
-		<exclusion pattern="^http://(?:blog|community)\.disqus\.com/" />
-
-
-	<rule from="^http://(?:www\.)?disqus\.com/"
```
jsha: This rule used to strip the 'www' from www.disqus.com. Have you confirmed that's not necessary anymore?

DeathlyX: Apparently http(s)://(www.)disqus.com redirects to https://disqus.com automatically now.

PR #1548, opened 2015-04-21
Reactivated fraunhofer.de

KommX:


src/chrome/content/rules/Fraunhofer.xml
```diff
@@ -3,21 +3,19 @@
...
 	<target host="www.scai.fraunhofer.de"/>
-	<!--	mismatch	-->
-	<target host="izb.fraunhofer.de"/>
-	<target host="www.izb.fraunhofer.de"/>
+	<target host="fraunhofer.de"/>
+	<target host="www.fraunhofer.de"/>
 
 	<securecookie host="^(?:.*\.)?(?:izb|scai)\.fraunhofer\.de$" name=".*"/>
 
-	<rule from="^http://(?:www\.)?(izb|scai)\.fraunhofer\.de/"
```
jsha: This rule used to rewrite the non-www version of these names to the www version. Have you confirmed both work?

KommX: scai.fraunhofer.de ends up at https://www.scai.fraunhofer.de/

The https-version of izb.fraunhofer.de breaks the design, so it isn't part of this ruleset anymore.

PR #1554, opened 2015-04-22
New rule for ishares.com

KommX:


reedy:
www.ishares.co doesn't seem to even resolve for me

KommX:
Did you misspell the URL?
This rule is about www.ishares.com.

reedy:
Sweet, no idea wtf I did there :)

PR #1558, opened 2015-04-22
Enable Dreamwidth

afuna:
https now works on Dreamwidth for both www and journal pages

Work on Dreamwidth here. We've done a bunch of work on https support recently, and https now works with journal subdomains and for logged out users.


jsha:
Excellent, thanks for writing this, and thanks for improving HTTPS support. Could you add at least one more subdomain as a test URL? The more the better - if you have a list of known subdomains that are expected to exist long term, that is ideal, so we can check them in our automated tests.

afuna:
Added a couple more subdomains, and a comment explaining that the "system.dreamwidth.org" test is for testing https on journals.

jsha:
Great, thank you!

afuna:
And thank you as well!

PR #1561, opened 2015-04-23
Create Dogecoin-P2Pool.xml

CohibAA:


reedy:
Attackers might be trying to steal your information from dogecoin-p2pool.com (for example, passwords, messages or credit cards). NET::ERR_CERT_AUTHORITY_INVALID

This server could not prove that it is dogecoin-p2pool.com; its security certificate is not trusted by your computer's operating system. This may be caused by a misconfiguration or an attacker intercepting your connection.

```
This CA Root certificate is not trusted because it is not in the Trusted Root Certification Authorities store.
```

CohibAA:
The certificate is not trusted because it is [self-signed](https://en.wikipedia.org/wiki/Self-signed_certificate).  I wasn't able to find specifications for submitting a self-signed SSL xml ruleset if they are different than CA signed rulesets.  Are self-signed certs not accepted to the master httpseverywhere rules?  If they are permitted for inclusion but I missed something, can you please point me in the right direction if I need to modify something in the ruleset?

jsha:
Generally no, we don't accept rulesets for sites with self-signed certs.

CohibAA:
Understood, thank you for the quick reply.

reedy:
Closing as self signed cert

PR #1566, opened 2015-04-23
www.moverall.com

ramonsmits:
This adds support for moverall.com which is a removal comparison website. It supports SSL but has no autoredirect. It also loads al its resources from SSL when accessed via SSL.

### DIFF
--- a/src/chrome/content/rules/Moverall.com.xml
+++ b/src/chrome/content/rules/Moverall.com.xml
@@ -2,6 +2,7 @@
 
 	<target host="moverall.com" />
 	<target host="*.moverall.com" />
+	<target host="az698912.vo.msecnd.net" />
 
 	<test url="http://moverall.com/" />
 	<test url="http://www.moverall.com/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Moverall.com.xml
```diff
@@ -0,0 +1,15 @@
+<ruleset name="Moverall">
+
+	<target host="moverall.com" />
+	<target host="*.moverall.com" />
+
+	<test url="http://moverall.com/" />
+	<test url="http://www.moverall.com/" />
+	<test url="http://cdn.moverall.com/" />
+	<test url="http://az698912.vo.msecnd.net/" />
```
jsha: You need to include a target host for msecnd.net and/or some of its subdomains.

ramonsmits: @jsha I thought that this would only apply this rule when the current url is matching the target hosts? As in, when a page patching the target url's then apply the following rules?

I'll add the host to the target's.

jsha: No, it's done on a url-by-url basis. For each URL, first we do a fast match against target hosts, and then for all rulesets that match that way we check rule regexps.

PR #1573, opened 2015-04-23
[Nerdsniped] use significant parallelism for warmup IO

pde:
- This will let the OS disk scheduler seek efficiently

### DIFF
--- a/makexpi.sh
+++ b/makexpi.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -o errexit
 APP_NAME=https-everywhere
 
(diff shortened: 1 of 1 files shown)


pde:
@2d1 this reverts a change you made to have the build script invoked under <tt>sh</tt> rather than <tt>bash</tt>; does that matter?

jsha:
So as long as you were getting nerdsniped, you tested the current time difference on your slow machine, right? :-)

pde:
Of course not :)

pde:
Build times from scratch on a laptop with slowish disk:
* no cat: 59 seconds
* one cat: 50 seconds
* many cats: 45 seconds

pde:
But it took me 20+ minutes to determine that :\

jsha:
According to XKCD, shaving five seconds off a daily task is worth up to two hours of work (provided a five year amortization period): https://xkcd.com/1205/. But of course HTTPS Everywhere will no longer be necessary in five years because everyone will be on HSTS.

PR #1576, opened 2015-04-24
[uppsalahem.se] Added ruleset

fuglede:


reedy:
It's nearly a year since I was in Uppsala!

Minor issue for future, mind keeping the whitespace consistent in your rulesets? :) I don't care if you use tabs or spaces, just consistency! :+1:

fuglede:
Sure; I must have copied the template without noticing the difference -- that also means that I might have overlooked that in other rules as well. Do you want a PR fixing it?

reedy:
Can if you want, it's not a major issue

PR #1577, opened 2015-04-24
Fix Flickr login.

jsha:
cc @pde @cooperq for review

### DIFF
--- a/src/chrome/content/rules/Flickr.xml
+++ b/src/chrome/content/rules/Flickr.xml
@@ -46,7 +46,8 @@
 	<!--securecookie host="^\.flickr\.com$" name="^(BX|admsort|cookie_session|current_identity|density|ffs|fldetectedlang|flrb|flrbcr|flrbgdrp|flrbgmrp|flrbgrp|flrbp|flrbs|fltoto|just_registered|localization|memsort|modsort|photoPHS|photoPWS|vp|xb)$" /-->
 	<!--securecookie host="^www\.flickr\.com$" name="^(liqph|liqpw|ywandp)" /-->
 
-	<securecookie host=".*\.flickr\.com$" name="^(?!current_identity$).+" />
+	<!-- If we secure this cookie, login fails. -->
+	<!-- <securecookie host=".*\.flickr\.com$" name="^(?!current_identity$).+" /> -->
 
 
 	<rule from="^http:"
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Flickr.xml
```diff
@@ -46,7 +46,8 @@
 	<!--securecookie host="^\.flickr\.com$" name="^(BX|admsort|cookie_session|current_identity|density|ffs|fldetectedlang|flrb|flrbcr|flrbgdrp|flrbgmrp|flrbgrp|flrbp|flrbs|fltoto|just_registered|localization|memsort|modsort|photoPHS|photoPWS|vp|xb)$" /-->
 	<!--securecookie host="^www\.flickr\.com$" name="^(liqph|liqpw|ywandp)" /-->
 
-	<securecookie host=".*\.flickr\.com$" name="^(?!current_identity$).+" />
+	<!-- If we secure this cookie, login fails. -->
+	<!-- <securecookie host=".*\.flickr\.com$" name="^(?!current_identity$).+" /> -->
```
cooperq: Should we delete this and the few lines above it instead of commenting it out?

jsha: This one in particular I chose to comment rather than deleting, because it's been fixed in the past and has reappeared. So it's useful to have a comment saying "don't do this."

cooperq: Fair enough

2d1:
@jsha, I understand that your time is far more valuable than this lowly peon's.

But could you please take just 5 minutes or so reviewing perl-style regular expressions and git usage.  This clearly specifies that any cookie whose name begins with the string ((beginning)current_identity(end)) is *not* secured.

If you had even more time, you could have used git log, git show, to determine that your statement "... it's been fixed in the past and has reappeared" is false. What you commented out *is* the fix.

Ta. (and sorry for misattributing to @pde)

jsha:
@2d1: Apologies, I was stressed and in a rush and messed this up. I tested Flickr on Friday in response to https://twitter.com/AlanDejecacion/status/591631531392663553, and found that login did in fact break with the same cookie error you fixed in November. I also found that removing the securecookie rule fixed it.

However, as of this morning I can no longer reproduce, so this may be intermittent.

Also, apologies for maligning your code: I actually did remember that you had fixed this in November by removing a cookie, and had assumed that because the site was breaking in the exact same way, someone *else* had undone your fix. This seemed like such a straightforward diff that I didn't think to check the git log.

PR #1579, opened 2015-04-24
Add new locales.

jsha:
And sort all. Note: Some locale directories from Transifex have underscores, but
the locale passed to the 'locale' directive in chrome.manifest must have a dash.

cc @cooperq @pde for review.


src/chrome.manifest
```diff
@@ -1,39 +1,53 @@
...
-locale https-everywhere ja chrome/locale/ja/
-locale https-everywhere si chrome/locale/si/
+locale https-everywhere sl-SI chrome/locale/sl_SI/
+locale https-everywhere sq chrome/locale/sq/
+locale https-everywhere sr chrome/locale/sr/
+locale https-everywhere sv chrome/locale/sv/
+locale https-everywhere th chrome/locale/th/
+locale https-everywhere tr chrome/locale/tr/
+locale https-everywhere uk chrome/locale/uk/
+locale https-everywhere zh-CN chrome/locale/zh-CN/
```
cooperq: I assume the path is meant to have an underscore here

cooperq:
LGTM modulo my one comment.

src/chrome.manifest
jsha: Actually that's the real path on disk. I'm not sure why some of the directories from Transifex have dashes vs spaces, but *shrug*.

PR #1585, opened 2015-04-25
Energy&isp

numismatika:
This PR has new and updated rules for small german ISPs and related energy companies


J0WI:
You may simplify the rules a bit using
```
<rule from="^http:"
          to="https:" />
```

Issue #1584, opened 2015-04-25
cookie changed to "secure" even if connection was NOT https


dan42:
According to https://www.eff.org/https-everywhere/rulesets#secure-cookies the &lt;securecookie> tag should **only** set the secure flag on a cookie **if** it was sent over HTTPS. However if you go to http://fast.animenewsnetwork.com/ the session cookie's secure flag is set. This is in direct contradiction of the docs.

DeathlyX:
My bad. Apparently the subdomain sets a session cookie originating from ".animenewsnetwork.com", which is secured in the ruleset I've pushed.

Nevertheless, the subdomain you mentioned and all others which I can find seem securable too. It should be for the better to redirect those to HTTPS instead of commenting out the cookie ruleset, presuming no functionality is lost of course.

dan42:
Thanks for improving the ruleset... but it's not what my bug report was about. Honestly I think the original ruleset didn't really have a problem; the problem is with the fact that https-everywhere changes the cookie the "secure" *even if it was not sent over a secure connection*. Not only is that in contradiction of the docs, it also makes no sense; a cookie set by a http://whatever.com will never be sent back on future visits to  http://whatever.com if it's marked secure; it's completely useless.

This is a bug with https-everywhere, not with this specific rule.

More specifically, a few of our subdomains do not support HTTPS. For example http://www.d.animenewsnetwork.com/ is meant to bypass Cloudflare (because sometimes it's wonky) and therefore doesn't have HTTPS. But the session cookie is still set with domain=.animenewsnetwork.com and therefore is converted to "secure". Which makes it impossible to be logged in at that url.

I very much appreciate the fact that Anime News Network has been added to https-everywhere. An easy fix would be to remove the &lt;securecookie> tag in this ruleset, but the **real** fix is to set the secure flag **only** for cookies sent over a HTTPS connection.

DeathlyX:
Ah, I've misunderstood you a bit first because of your previous title. Well, some additional coverage isn't bad neither, right? :)

You've got a point though, I can confirm this strange behaviour with FireFox. Clearly the fast subdomain shouldn't be triggered with the current ruleset. You can definitely notice the ruleset isn't listed on the menu when visiting the subdomain, as it should be. But appearently it still triggers the securecookie rule for animenewsnetwork.com, overriding the listed target hosts rules. The ruleset shouldn't have been loaded at all.

Hopefully someone of the development team can give some more clarification...

[...]

PR #1588, opened 2015-04-26
Update for Anime News Network

DeathlyX:
Added some more HTTPS coverage as I mentioned in #1584.


DeathlyX:
The problem appears to be more complex, as you can read on the issue. I'm not certain if it's better to keep this pull request as it is or to remove the ".animenewsnetwork.com" securecookie rule...

DeathlyX:
I've disabled the securecookie ruleset for (.)animenewsnetwork.com now.

PR #1591, opened 2015-04-27
wiki.eclipse.org, marketplace.eclipse.org support HTTPS

MilanKral:


src/chrome/content/rules/Eclipse.xml
```diff
@@ -1,22 +1,21 @@
...
+	<target host="wiki.eclipse.org" />
+	<target host="marketplace.eclipse.org" />
 
 	<securecookie host="^bugs\.eclipse\.org$" name=".*" />
 
-
-	<rule from="^http://((?:bugs|dev|git|www)\.)?eclipse\.org/"
-		to="https://$1eclipse.org/" />
-
-	<rule from="^http://marketplace(1)?\.eclipse\.org/(eclipse\.org-common|misc|modules|sites)/"
```
jsha: It looks like this rule used to only do HTTPS on certain paths on marketplace. Have you checked some other paths on the site and made sure they support HTTPS properly?

MilanKral: I have checked different paths on marketplace.eclipse.org and the site works.

PR #1594, opened 2015-04-27
[GoogleServices] add coverage for Google movies

pde:


jsha:
Tests are failing because of the need to add test url coverage, which would be a ton of work.

Luckily, @fuzzyroddis did a bunch of work to modernize the Google rulesets. Unluckily, I had to last-minute revert those changes to do the 5.0.0 release, because they missed coverage for a bunch of country TLDs. We need to roll it forward again to pick up the good stuff and fix the bugs. Here's the issue: https://github.com/EFForg/https-everywhere/issues/1595

semenko:
I think this is superseded by #1595 - closing

PR #1614, opened 2015-05-01
Print out all case insensitive filename conflicts

### LINKED ISSUE #1401: List all conflicting filenames due to case-insensitivity issues
Sometimes when I'm being "lazy", I work directly from github.com. And as it doesn't like all the files in the rules folder, you can't view them easily.

Merging #1159 broke the build, so I deleted said file when I noticed. However, it doesn't tell you what the names of the conflicting files actually are

```
Generating sqlite DB
src/chrome/content/rules/geocaching.com.xml failed case-insensitivity testing.
```

Can we have it spit out all the conflicts in https://github.com/EFForg/https-everywhere/blob/master/utils/make-sqlite.py#L54 ?

reedy:
Fixes #1401

### DIFF
--- a/utils/make-sqlite.py
+++ b/utils/make-sqlite.py
@@ -5,10 +5,10 @@
 import glob
 import locale
 import os
+import re
 import sqlite3
 import subprocess
 import sys
-import re
 
 from collections import Counter
 from lxml import etree
--- a/utils/make-sqlite.py
+++ b/utils/make-sqlite.py
@@ -8,6 +8,7 @@ import os
 import sqlite3
 import subprocess
 import sys
+import re
 
 from collections import Counter
 from lxml import etree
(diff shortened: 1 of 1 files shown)


reedy:
```
reedy@ko-kra:~$ ./test.py                                                                       
['test/FOO.xml', 'test/Foo.xml', 'test/foo.xml'] failed case-insensitivity testing.
reedy@ko-kra:~$
```

My python isn't great!

Not the nicest output, but I wasn't sure if it was worth making it prettier for the intended audience of this script

utils/make-sqlite.py
```diff
@@ -8,6 +8,7 @@
 import sqlite3
 import subprocess
 import sys
+import re
```
reedy: I've just noticed the list above is alphasorted

jsha:
Looks great, thanks!

PR #1615, opened 2015-05-01
Update YourBittorrent.xml: strip "www."

PotcFdk:
The "www" subdomain doesn't seem to work on this domain. (See [this](https://www.ssllabs.com/ssltest/analyze.html?d=yourbittorrent.com))


jsha:
Thanks for the fix! To get the tests to past, could you change the "*.yourbittorent.com" target host to "i.yourbittorrent.com"?

PotcFdk:
There, I hope this is better.

PR #1636, opened 2015-05-06
www.oeffentliche-it.de

MilanKral:


reedy:
Visiting https://oeffentliche-it.de/ gives

This server could not prove that it is oeffentliche-it.de; its security certificate is from www.oeffentliche-it.de. This may be caused by a misconfiguration or an attacker intercepting your connection.


Might be worth forcing the www in the ruleset?

MilanKral:
I have fixed the redirection from oeffentliche-it.de to www.oeffentliche-it.de.

PR #1658, opened 2015-05-09
[Peytz.dk] New ruleset

fuglede:


reedy:
I get a 404 for this...

fuglede:
@reedy: Yeah, the root gives a 404 on port 80 as well. It's an analytics company and some medias refer to scripts on this domain -- one concrete that I came across is https://opinion.peytz.dk/decide2.js.

reedy:
Aha, maybe add that as a test url? :)

fuglede:
Can do. (Although I'm not completely sure I understand which purpose adding that one in particular could serve?)

fuglede:
Guess that makes sense; I could also add a few words on the intent in a comment if that's better.

PR #1662, opened 2015-05-10
WDR rule extensions

numismatika:
added a few more domains to the ruleset and test-urls as well


reedy:
Looks like this ruleset could be massively simplified by adding the hosts explicitly, and using the catch all http -> https redirect

numismatika:
@reedy yea no complicated rewrites are necessary  here

numismatika:
@reedy
is it appropriate now?

PR #1663, opened 2015-05-10
Improved Thirdparty content

numismatika:
this handles a few more subdomains for ad-networks and images used on newspages


reedy:
```
ERROR src/chrome/content/rules/Nugg.ad.xml: Not enough tests (2 vs 4) for <Rule from '^http://([\w-]+-s|gwp|mtm)\.nuggad\.net/' to 'https://\g<1>.nuggad.net/'>
ERROR src/chrome/content/rules/Outbrain.xml: Not enough tests (0 vs 1) for <Rule from '^http://(?:wp\.|www\.)?outbrain\.com/' to 'https://www.outbrain.com/'>
ERROR src/chrome/content/rules/Outbrain.xml: Not enough tests (0 vs 10) for <Rule from '^http://(editorial|u|widgets|images|log|my|paid|traffic|vrp|vrt)\.outbrain\.com/' to 'https://\g<1>.outbrain.com/'>
ERROR src/chrome/content/rules/Outbrain.xml: Not enough tests (1 vs 5) for <Exclusion pattern '^http://(?:www\.)?outbrain\.com/+(?!wp-content/|wp-includes/)'>
```

numismatika:
i am not entirely sure on 
>^http://([\w-]+-s|gwp|mtm)\.nuggad\.net/'

what kind of string does [\w-]+-s match?

PotcFdk:
`[\w-]+` matches multiple word characters (i.e. alphanumeric and underscores)
So `^http://([\w-]+-s).nuggad.net/` could match something like `http://test-s.nuggad.net/`.

jsha:
Could you update these rulesets to follow the current ruleset style guide? In particular, we'd like to avoid using wildcard targets (`<target host="*.foo.com" />`) when the hostnames are explicitly enumerated in rules. Instead, it's better to enumerate those hostnames in target tags. Wildcard targets are still fine when there is a corresponding wildcard rule.

PR #1671, opened 2015-05-10
Add rule for pao-pao.net

PotcFdk:
Note, that even though this page has mixed content, it 301-s to https,
so this rule shouldn't do any harm.

```
$ curl -I http://pao-pao.net/
[...]
Location: https://pao-pao.net/
```


reedy:
Looks like it's trying to load some fonts over http

You could mark the ruleset platform="mixedcontent"

PotcFdk:
If that's the recommended way, even for websites that 301 to https themselves, fine.
Added a new commit.

PR #1685, opened 2015-05-12
Added rule for pcgarage.ro

mnemonicflow:
The rule has been tested to work properly in a standalone Firefox profile

### DIFF
--- a/src/chrome/content/rules/pcgarage.ro.xml
+++ b/src/chrome/content/rules/pcgarage.ro.xml
@@ -1,0 +1,6 @@
+<ruleset name="pcgarage.ro">
+  <target host="pcgarage.ro" />
+  <target host="www.pcgarage.ro" />
+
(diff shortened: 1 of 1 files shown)


reedy:
https://pcgarage.ro/ redirects to http://www.pcgarage.ro/, but then HTTPSE would catch this and put it to https://www.pcgarage.ro/


Why curl gives a 403 for that site though... :/

mnemonicflow:
@reedy Seems like the HTTP server on the other end doesn't like the User-Agent request header curl sends and it refuses to serve the response if it sees the UA something like `User-Agent: curl/7.38.0`

This works:
` curl -v -s -H 'User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:37.0) Gecko/20100101 Firefox/37.0' -L -i -o /dev/null https://pcgarage.ro`

PR #1690, opened 2015-05-13
[Givskudzoo.dk] New ruleset

fuglede:


reedy:
```
Mixed Content: The page at 'https://www.givskudzoo.dk/' was loaded over HTTPS, but requested an insecure script 'http://www.jscache.com/wejs?wtype=cdswritereviewlg&uniq=185&locationId=1876176&lang=da&border=false'. This request has been blocked; the content must be served over HTTPS.
```

Do we care? :) Does it break anything? Looks trivial

fuglede:
As far as I can tell, it does break the TripAdvisor box towards the bottom; I'll let you decide how much breaking is too much.

Unfortunately jscache.com only provides 301s on https requests. However, the scripts themselves only serve to include other scripts which *can* be secured. Let me just ask the givskudzoo.dk devs if they want to change chat (there was some other mixed content as well, but they were really quick to fix that.)

PR #1692, opened 2015-05-13
Add new rulesets for pollin.de and dict.cc

KurtKrampmeier:


reedy:
```
failure: non-whitelisted downgrade rule in dict.cc redirects to http.
failure: dict.cc.xml failed test: Rule redirects to something other than https.
```

KurtKrampmeier:
So what would be the correct solution? Their frameset breaks https://www.dict.cc/dict/options.php, even without HTTPS Everywhere. I just followed the documentation on 
https://www.eff.org/https-everywhere/rulesets#downgrades, which did not provide any information about whitelisting being necessary. What would be the policy for whitelisting?

Should I remove the downgrade rule and add a `platform="mixedcontent"` to the ruleset?

reedy:
That's curious, you already have the downgrade="1" in place...

reedy:
Oh, unless it's because of the exclusion AND the downgrade

reedy:
Ah, you need to add it to https://github.com/EFForg/https-everywhere/blob/master/utils/downgrade-whitelist.txt

Also, use branches so you don't add other rulesets to existing PR

J0WI:
@KurtKrampmeier there will be a new release soon. It would be great to have the dict.cc rule in.

semenko:
Looks like our tests are failing on this PR -- and it hasn't been updated in a while. I'm going to close this for now, but if you'd like it merged, feel free to re-open. Thanks! :)

PR #1707, opened 2015-05-16
Add ruleset for Phillips Academy Andover

TonyZTan:
Add ruleset for Phillips Academy Andover websites.


reedy:
There's some mixed content on the home page, but it doesn't seem to obviously break anything

PR #1708, opened 2015-05-16
Added ruleset for PACTF.com

TonyZTan:


reedy:
Also should be secured by the cloudflare ruleset

TonyZTan:
@reedy It is indeed running on CloudFlare; however, I don't think sites running on CloudFlare are automatically covered. I cannot find any rules for that, and there are many websites that run on CloudFlare that do not get redirected to HTTPS. (such as http://www.tonytan98.com) Thanks!

reedy:
HTTPSE shows that the CloudFlare rule is active for this site... For your personal site I don't get redirected to HTTPS, but I do for this site, and some of your other CloudFlare hosted PRs. I wonder if there's some server side setting for forcing HTTPS?

I'm not adversed to merging these PRs, just a little confused/intrigued by it all :)

TonyZTan:
Hmm.. You are right, the extension does show CloudFlare.com. But it shows for my personal site as well, without switching to HTTPS. I control both sites, and I know that PACTF.com redirects users to HTTPS automatically server-side, while my personal site does not, even though both could support it.

I think the extension is currently not doing anything to rewrite the requests to CloudFlare; it is just the server settings. I don't think it should, either, because many sites on CloudFlare only support HTTPS partially (i.e. mixed content). The rules should be on a per-site basis, and I think they are right now? Looking at the https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/CloudFlare.xml, I don't see how it rewrites requests to other websites.

reedy:
Isn't it done by CloudFlare essentially being one big CDN?

https://www.eff.org/deeplinks/2014/10/how-cloudflare-moved-web-toward-ubiquitous-https


There is obviously the Server side 301 too

```
chronos@localhost / $ curl -I -L www.pactf.com
Date: Sat, 16 May 2015 23:35:40 GMT
Connection: keep-alive
Set-Cookie: __cfduid=de9f32cf1a710400a478bee07d205ffe31431819340; expires=Sun, 15-May-16 23:35:40 GMT; path=/; domain=.pactf.com; HttpOnly
Location: https://www.pactf.com/
X-Content-Type-Options: nosniff
Server: cloudflare-nginx
CF-RAY: 1e7ae67b7a820328-MIA

HTTP/1.1 200 OK
Server: cloudflare-nginx
Date: Sat, 16 May 2015 23:35:40 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Set-Cookie: __cfduid=df6c96b0ae912829b69c0c53c44f1b4d91431819340; expires=Sun, 15-May-16 23:35:40 GMT; path=/; domain=.pactf.com; HttpOnly
Last-Modified: Sun, 03 May 2015 22:43:09 GMT
Access-Control-Allow-Origin: *
Expires: Sat, 16 May 2015 23:45:40 GMT
Cache-Control: max-age=600
Via: 1.1 varnish
Age: 0
X-Served-By: cache-mia1326-MIA
X-Cache: MISS
X-Cache-Hits: 0
X-Timer: S1431819340.456140,VS0,VE28
Vary: Accept-Encoding
Strict-Transport-Security: max-age=15552000; includeSubDomains; preload
X-Content-Type-Options: nosniff
CF-RAY: 1e7ae67dab6c07cd-MIA

chronos@localhost / $
```

jsha:
FYI the Cloudflare rule is active because of an embedded file: https://ajax.cloudflare.com/cdn-cgi/nexp/dok3v=7e13c32551/cloudflare.min.js. The Cloudflare rule doesn't rewrite all the top-level Cloudflare-hosted domains. It would be nice if it did, but there are two obstacles:

1) We don't know about them
2) Many of those domains are likely to have mixed content problems if we rewrite them.

We just had some of the Cloudflare folks over for a visit at EFF to discuss this very issue. It seems like (1) is easy to fix, but (2) is hard. The solution will probably wind up requiring something like https://w3c.github.io/webappsec/specs/upgrade/.

TonyZTan:
@jsha CloudFlare support told me a few days ago that they would consider supporting the Upgrade Insecure Requests directive, though they could not provide a time frame. Do you think it would be difficult for CloudFlare to add a rule to their systems rewriting all HTTP references on HTTPS pages to HTTPS? (perhaps with an option for webmasters to disable this?) Since they power >2 million websites, getting this fixed could help the effort to encrypt the web by a lot.

reedy:
Having good access to the software and the content (in this case, Wikipedia/Wikimedia and MediaWiki) does make this sort of thing easy. The amount of hardcoded HTTP we found in various site js and such... Years after we put proper HTTPS support in too. Scary :)


As for CloudFlare... @FiloSottile is a big HTTPS supporter and also on the CloudFlare Security team... Maybe he can chip in :)

jsha:
@TonyZTan: The inline-rewrite thing was also something we discussed. It's not a silver bullet. The most troublesome mixed content comes from external resources, and some of those cannot be converted to HTTPS. Even with a full crawl of a site, it's hard to know with certainty whether that site will generate mixed content errors when moved to HTTPS, since resources are often loaded by JS.

It would be a potentially powerful 90% solution, though. But there's also a matter of resources and time to implement it and make it fast.

TonyZTan:
I see. Thank you.

PR #1710, opened 2015-05-16
Add ruleset for CerberusApp.com

TonyZTan:


reedy:
Site seems to force HTTPS. @jsha Is there any use of having rules for sites that already force to HTTPS?

TonyZTan:
@reedy Yes, including sites that already force HTTPS can prevent SSL striping attacks, in which the MITM attacker prevents the 301/302 redirect to HTTPS from reaching the user. In this case, the protection offered by HTTPS-Everywhere is similar to HSTS, except for the fact that many site owners (including this one) fail to send HSTS headers for both the domain with "www" and without. Please see: https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?

TonyZTan:
@reedy (Off-topic) Are you also https://en.wikipedia.org/wiki/User:Reedy? That's cool. I am https://en.wikipedia.org/wiki/User:Tony_Tan. I thought your name was familiar but did not realize where I saw it until now. ;)

reedy:
Yup, that'd be me ;)

jsha:
@reedy: Yep, Tony's got it right. Eventually I'd like to be able to delete rules that are truly 100% redundant, i.e. those that overlap with preloaded HSTS rules. But the sslstrip attacks are real, so in the meantime we accept rulesets even for sites that do a 301 redirect to HTTPS.

PR #1714, opened 2015-05-16
Fix shell scripts for BSD-ish systems

### LINKED ISSUE #1504: test.sh doesn't work in BSD-ish (OSX) systems
In order to facilitate running as a git hook, [Line 6 of test.sh](https://github.com/EFForg/https-everywhere/blob/master/test.sh#L6) says:

```Bash
cd $(dirname $(readlink -f $0))
```

BSD `readlink`'s `-f` is "format", not "canonicalize" like in GNU `readlink`. Tiresome. I tried patching it to this:

```Bash
ABSPATH=$(readlink -f $0) || ABSPATH=$(readlink $0)
cd $(dirname "$ABSPATH")
```

But then invoking test.sh directly doesn't work, because BSD `readlink` doesn't print anything at all if its argument isn't a symlink. At that point I gave up.

BSD `mktemp` also diverges from GNU's version. This is a problem with [the invocation on line 12](https://github.com/EFForg/https-everywhere/blob/master/test.sh#L12). I haven't dug into it in any detail.

ErinCall:
fixes #1504 

Basically I added the `coreutils` Homebrew package to `install-dev-dependencies.sh` and added a script that sets variables pointing to the location of a GNU version of the utils in question.


ErinCall:
Oh, meant to mention--I've tested this approach on OSX, for launching `test.sh` as a git hook, either from the project root or a subfolder, and for running `./test.sh` directly. It doesn't work right when running `test.sh` from a subfolder (e.g. `../test.sh`). It gets bollixed up trying to find `locate-gnu-utils.sh` in order to successfully do a `chdir`. I'm not sure if that's a dealbreaker.

jsha:
This is terrific! Thank you for working on it. The failure to run test.sh from a subdirectory is a dealbreaker, I'm afraid. Because install-dev-dependencies adds the symlink to git hooks, a lot of people have it.

I think you can probably avoid the GNU vs BSD distinction for readlink if we drop the `-f` flag. By calling readlink directly you can avoid calling out to `locate-gnu-utils.sh` until after the cd. And I think you can then remove the readlink detection from `locate-gnu-utils.sh`.

Also, I'd prefer for locate-gnu-utils.sh to live under `utils/`.

What's the behavior difference between BSD sha256sum vs GNU sha256sum?

ErinCall:
> The failure to run test.sh from a subdirectory is a dealbreaker, I'm afraid. Because install-dev-dependencies adds the symlink to git hooks, a lot of people have it.

Just to clarify: running it as a git hook works from anywhere. It only fails if you manually invoke test.sh from a subdirectory. I can probably get it working, though.

> What's the behavior difference between BSD sha256sum vs GNU sha256sum?

There doesn't seem to be a BSD sha256sum at all. :) There is a `shasum` with a `--algorithm sha256` option on both OSX and Ubuntu, which might be a better solution. I just went with the approach I'd been using.

jsha:
I tried pushing this branch to my own repo, triggering the git hook, and the script failed in the same way as manually invoking test.sh from a subdirectory.

>  There is a shasum with a --algorithm sha256 option on both OSX and Ubuntu, which might be a better solution.

On my system that appears to belong to the `perl` package. Not sure whether it's more common or less. There's also `openssl sha256`, which is probably on ~all systems that we care about. Let's use that one instead?

The `mktemp -d` problem can be worked around inline: http://unix.stackexchange.com/a/84980

With those three changes we could avoid the need for locate-gnu-utils.sh entirely, and I think make the scripts work on *BSD without having to install GNU coreutils.

ErinCall:
I'm back! The new version should work from any directory, and doesn't need `coreutils`. :)

jsha:
Looks great, thanks for working on this!

PR #1720, opened 2015-05-18
[pr0gramm.com] Add ruleset

cloehle:


reedy:
Does this site have any subdomains other than www? www 301s back to no www on HTTP, but then HTTPS everywhere would secure it again

cloehle:
Yes, the static links to images like this:
https://img.pr0gramm.com/2015/04/24/44e0aa4b89360b29.jpg

reedy:
Do you know all the subdomains this site actually uses? Or does it use a lot?

Generally, best practice now is to explicitly list all subdomains in the target hosts rather than using the * wildcard

PR #1725, opened 2015-05-19
add ÖBB ruleset

J0WI:


J0WI:
This is an important site in Austria

reedy:
```
This server could not prove that it is oebb.at; its security certificate is from www.oebb.at. This may be caused by a misconfiguration or an attacker intercepting your connection.
```

J0WI:
fixed

PR #1736, opened 2015-05-21
[Billund.dk] New ruleset

fuglede:


reedy:
Some mixed content from google fonts, no big deal

PR #1745, opened 2015-05-22
add Jadi.net ruleset

akfoss:


reedy:
There seems to be a lot of mixed content on this site when viewed over https

akfoss:
Yeah. The Iranian government has blocked access to this website for http access. But the https version has not been blocked (or could not be blocked).

The author regularly posts photos from other websites, and this habit may have caused the problem of having mixed contents.

I added this ruleset to circumvent censorship.

PR #1748, opened 2015-05-23
Esc

numismatika:
This adds rules for the Eurovision  Song Contest streaming site and the thirdparty content by Akamai that is used there


pde:
Pity we didn't catch this & ship the patch *before* the contest :/

This isn't the first time we've had trouble with a giant event using custom streaming code.

numismatika:
The amount of issues and pullrequests here is growing.
The patch , for which i wasn't even sure if it solved the problem completely, was made just a couple of hours before the finals and as far as i know all people with write access to the repo are us-based.
It wasn't possible to get it done in time. Not to mention that a new release wouldn't have made it in time as well.

PR #1780, opened 2015-05-29
Add rule for cemetech.net

tari:


reedy:
```
chronos@localhost / $ curl -I -L https://cemetech.net
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
chronos@localhost / $ curl -I -L -k https://cemetech.net
Date: Sat, 30 May 2015 00:25:35 GMT
Server: Apache/2.4.10 (Debian)
Location: http://www.cemetech.net/
Content-Type: text/html; charset=iso-8859-1

HTTP/1.1 200 OK
Date: Sat, 30 May 2015 00:25:35 GMT
Server: Apache/2.4.10 (Debian)
Set-Cookie: phpbb_data=a%3A2%3A%7Bs%3A11%3A%22autologinid%22%3Bs%3A0%3A%22%22%3Bs%3A6%3A%22userid%22%3Bi%3A-1%3B%7D; expires=Sun, 29-May-2016 00:25:35 GMT; Max-Age=31536000; path=/; domain=.cemetech.net
Set-Cookie: phpbb_sid=603c507456c87109db23a14612321f52; path=/; domain=.cemetech.net
Cache-Control: no-cache, pre-check=0, post-check=0
Expires: 0
Pragma: no-cache
Content-Type: text/html; charset=iso-8859-1

chronos@localhost / $
```

Non www HTTPS redirects to non HTTP with www. But HTTPSE would catch this and redirect anyway

PR #1782, opened 2015-05-30
Create telebasel.ch.xml

J0WI:


reedy:
I see some mixed content, are these ok?

J0WI:
player is affected by mixed content.

jsha:
Marking as needswork based on player being affected by mixed content. Let me know if that's incorrect.

J0WI:
There is nothing more we can do here. The player requires resources that are only available over http. So I marked it with ```platform="mixedcontent"```

PR #1798, opened 2015-05-31
Create hieber.de.xml

J0WI:


reedy:
Mixed content but it's only for fonts

PR #1813, opened 2015-05-31
Update Spreadshirt.xml

J0WI:


reedy:
ERROR src/chrome/content/rules/Spreadshirt.xml: Not enough tests (17 vs 18) for <Rule from '^http://(\w+)\.spreadshirt\.(at|be|co\.uk|com|com\.au|de|dk|es|fi|fr|ie|it|net|nl|no|pl|se)/' to 'https://\g<1>.spreadshirt.\g<2>/'>

J0WI:
I didnt' find the missing regex to test. Can you give me a hint?

jsha:
The heuristic used to pick the required number of tests is: add one for each `+*?|` character (with a few small caveats). There are 17 `|` characters, but there is also the `\w+`. So you should add at least one test URL that matches the rule and starts with something other than wikileaks (the more the better).

PR #1815, opened 2015-06-01
[Cyon.ch] Resolved insecure cookie and include all subdomains

dol:
The reported issue with secure cookie are resolved on the server side. The securecookie tag is obsolete.
All subdomain content is served over https. HSTS header is constructed to includeSubDomains.

### DIFF
--- a/src/chrome/content/rules/Cyon.ch.xml
+++ b/src/chrome/content/rules/Cyon.ch.xml
@@ -6,31 +6,11 @@
 		- order
 		- webmail
 
-
-	Insecure cookies are set for these domains and hosts:
-
-		- .cyon.ch
-		- my.cyon.ch
-
 -->
 <ruleset name="cyon.ch">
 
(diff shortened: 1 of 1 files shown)


reedy:
Generally, we are moving away from using *.foo.bar, but that is mostly for sites where they then have loads of different rewrites for various subdomains.

I *think* that if it's just being done for ALL subdomains, this should be ok

dol:
@reedy Thank you for your advice. Is there somewhere a information site about best practices that I can refer to?

reedy:
See https://www.eff.org/https-everywhere/rulesets for more info

jsha:
This looks great. When all subdomains really do support HTTPS, it's fine to have a wildcard target and a corresponding blanket rule.

dol:
@jsha Thx for merging. We make sure that every subdomain is secured.

PR #1818, opened 2015-06-01
Create olx.pl.xml

matrix0123456789:


src/chrome/content/rules/olx.pl.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="RabbitMQ">
```
reedy: The name doesn't look right either for this site

reedy:
```
This server could not prove that it is olx.pl; its security certificate is from ssl.olx.pl. This may be caused by a misconfiguration or an attacker intercepting your connection.
```

semenko:
Looks like this PR has some outstanding edits -- and it hasn't been updated in a while. I'm going to close this for now, but if you'd like it merged, feel free to re-open. Thanks! :)

PR #1846, opened 2015-06-03
Update AllTube Download rule

Rudloff:
Hello,

The app is now hosted on a Heroku free plan so HTTPS is not available on alltubedownload.net anymore, so I updated the rule.

Regards,


reedy:
Might be temporary, but https://alltube.herokuapp.com/ is down for me atm

Rudloff:
Strange, it works for me.

reedy:
Does for me now too, must've been transient

PR #1849, opened 2015-06-04
[Federal Soup] Added ruleset.

jpds:


reedy:
https://federalsoup.com redirects back to http with Home.aspx at the end. Not a big deal

But there's a lot of mixed content, including loading jquery itself

semenko:
Blast from the past! Still mixed content, but the site seems to work fine. Merging!

PR #1856, opened 2015-06-06
Update TagesWoche.xml

J0WI:


reedy:
Few bits of mixed content, but nothing major

PR #1858, opened 2015-06-06
Update LibreOffice.xml

J0WI:


J0WI:
Important ruleset, would be great to have this changes in the next release

Issue #1803, opened 2015-05-31
Add ecosia.org


blobgo:
Please add [Ecosia.org] (https://www.ecosia.org). It's an eco-friendly search engine used by many users so it should be in HTTPS Everywhere.
But exclude https://blog.ecosia.org/ and https://blog.ecosia.org as they are not accessible with HTTPS.

PR #1864, opened 2015-06-07
Add Ecosia.org rule

donced:
#1803


blobgo:
This rule doesn't look as if it would exclude https://blog.ecosia.org/ and https://blog.ecosia.org...

donced:
@blobgo exclusion is not necessary in this case because the rule "ecosia.org" "www.ecosia.org" don't apply to blog.ecosia.org. Only "*.ecosia.org" or (obviously) "blog.ecosia.org" would apply.

blobgo:
Ah okay I overlooked this. So now all should be all right.

blobgo:
Anything which speaks against merging this?

PR #1875, opened 2015-06-08
New rule for arlt.com

KommX:


reedy:
```
This server could not prove that it is arlt.com; its security certificate is from *.arlt.com. This may be caused by a misconfiguration or an attacker intercepting your connection.
```

Looks like this needs a !www -> www rewrite

KommX:
I simply removed the target 'arlt.com'.

PR #1892, opened 2015-06-09
Create SemperVideo.xml

J0WI:


reedy:
Do we care about the mixed content on this site?

J0WI:
The site works for the most part and is secured by us. I'll contact the owner anyway.

PR #1893, opened 2015-06-09
Create kinoX.to.xml

J0WI:


J0WI:
very popular domain, would be great to have a review soon.

PR #1894, opened 2015-06-10
Add zdf.de rule

m0jo:
Added www.ZDF.de rule. SSL Only on www

### DIFF
--- a/src/chrome/content/rules/hackaday.com.xml
+++ b/src/chrome/content/rules/hackaday.com.xml
@@ -1,0 +1,7 @@
+<ruleset name="hackaday.com">
+	<target host="www.hackaday.com" />
+	<target host="hackaday.com" />
+
+	<rule from="^http:"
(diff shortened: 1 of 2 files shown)


reedy:
```
Mixed Content: The page at 'https://www.zdf.de/' was loaded over a secure connection, but contains a form which targets an insecure endpoint 'http://www.zdf.de/form/globale-suche-10.html'. This endpoint should be made available over a secure connection.
(index):210 Mixed Content: The page at 'https://www.zdf.de/' was loaded over HTTPS, but requested an insecure image 'http://wwwdyn.zdf.de/images/weather/icons/weather_12.png'. This content should also be served over HTTPS.
```

Minor issue

m0jo:
Updated.
Can this be merged?

J0WI:
Please open a separate pull request for Hackaday.com and fix the issues mentioned in the test result.

PR #1901, opened 2015-06-10
[StartCom] Attempt to fix problem

TonyZTan:
Attempt to fix problem that caused the ruleset to be disabled.

### DIFF
--- a/src/chrome/content/rules/StartCom.xml
+++ b/src/chrome/content/rules/StartCom.xml
@@ -39,6 +39,17 @@
 
   <!-- should mitigate against exploitation of the above exclusions -->
   <securecookie host=".*" name=".*" />
+  
+  <test url="http://startcom.org/" />
+  <test url="http://linux.startcom.org/" />
+  <test url="http://forum.startcom.org/" />
+  
+  <test url="http://www.startssl.com/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/StartCom.xml
```diff
@@ -43,7 +39,19 @@ Fetch error: http://startcom.org/ => https://startcom.org/: (28, 'Connection tim
...
+  <test url="http://forum.startcom.org/" />
+  
+  <test url="http://www.startssl.com/" />
+  <test url="http://auth.startssl.com/" />
+  <test url="http://www.startssl.net/" />
+  <test url="http://www.startssl.org/" />
+  <test url="http://www.startssl.eu/" />
+  <test url="http://www.startssl.us/" />
 
+  <rule from="^http://startcom\.org/" to="https://www.startcom.org/" />
```
reedy: Don't the 2 rules below take care of this? Obviously with a more complex rule...

TonyZTan: @reedy It seems like https://startcom.org/ does not respond (times out), and that is why the rule was disabled by the checker. My rule redirects http://startcom.org/ to https://www.startcom.org/ instead. Thanks!

PR #1903, opened 2015-06-11
[ssl-tools.net] New ruleset

TonyZTan:


reedy:
Is it just me, or would it be hilarious if a site called ssl-tools.net didn't have HTTPS support? :D

TonyZTan:
Haha I didn't even think about that :smile:

PR #1916, opened 2015-06-13
Create sogou.com.xml

J0WI:


J0WI:
popular domain

reedy:
Visiting http://www.sogou.com/ keeps redirecting me back to HTTP serve side

J0WI:
WFM using curl or browser, even if I enable JS.

PR #1919, opened 2015-06-13
Update AliCDN.com.xml

J0WI:


J0WI:
The refactor of Taobao.xml, which is a very important ruleset, depends on this.

PR #1924, opened 2015-06-14
[YouNeedABudget] youneedabudget.com

fuzzyroddis:


src/chrome/content/rules/YouNeedABudget.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="You Need A Budget">
+	<target host="youneedabudget.com" />
+	<target host="www.youneedabudget.com" />
+	<target host="purchase.youneedabudget.com/" />
```
reedy: Looks like the trailing / here in the host is breaking the tests

fuzzyroddis: Fixed thanks.

PR #1928, opened 2015-06-14
add ruleset for Wien-IT

numismatika:
adds a rule to cover most of the domains administrated by the it department of the austrian capital city


semenko:
Hm, looks like there are some invalid cert names in here, e.g. mobile.buergersolarkraftwerk.at

numismatika:
The data was ripped from the alternative names on the main certificate, which seems to have changed a little meanwhile.

numismatika:
@semenko 
I removed the mobile page now and added a few other domains.
Anything else that needs attention before merging?

semenko:
Awesome -- thanks! :+1:

PR #1945, opened 2015-06-17
Add appsflyer ruleset

graingert:


graingert:
oops didn't spot it already exisited

PR #1948, opened 2015-06-18
New rule for thetvdb.com

KommX:


reedy:
I see some mixed content, but nothing important, I think

semenko:
Looks fine to me too

Issue #1911, opened 2015-06-11
Certain features on Steam Community sites break when using https


NJC2:
There are some problems when using the Steam Community (steamcommunity.com) site with https.  One that I run into most often is screenshots not loading.  When viewing forums, the screenshot thumbnails load, but when you click one to see the larger size and details, it never loads.
![steamssnoload](https://cloud.githubusercontent.com/assets/6441752/8117725/2a4e2d4a-1050-11e5-86f2-1ff9f9adb0cf.png)

It's not entirely an https-everywhere problem because disabling https-everywhere and manually going to https://steamcommunity.com I still have this problem, but the only way around the problem is to completely disable https-everywhere because it forces steamcommunity.com to use https.

I wish I could disable https-everywhere for specific sites instead of globally disabling rules.  I usually end up disabling the addon completely and then forgetting to re-enable it.

reedy:
See also #1453 and #1702

NJC2:
I have already seen those.  Those are problems with the Steam store (store.steampowered.com), this is a dealing with the Steam Community site (steamcommunity.com).  It also can't be solved by disabling rules as certain parts of steamcommunity.com just don't work at all using https.  Currently the only way to use the Steam Community is to completely disable the addon because it always forces you to https.

reedy:
I know, and I didn't say they were duplicates.

Though, I imagine most of the issues come from the cdn/akamai rewrites

```
Mixed Content: The page at 'https://steamcommunity.com/#scrollTop=1500' was loaded over HTTPS, but requested an insecure image 'http://images.akamai.steamusercontent.com/ugc/711914012598753408/F70D59485BBE5DCDBCF197AAABA07BFD973B6535/'. This content should also be served over HTTPS.
```

```
	<rule from="^http://cdn\.akamai\.steamstatic\.com/"
		to="https://steamcdn-a.akamaihd.net/" />
```

Tailszefox:
The Monster Summer game, hosted on steamcommunity.com, is also completely broken when using HTTPS: it loads, but just sits there failing to get any assets.

The only way to play is to disable both the Steamcommunity rule and the Akamai rule, or disable HTTPS Everywhere entirely. Only disabling one or the other doesn't work.

From the console:
```
[...]

PR #1949, opened 2015-06-18
Disable all steam rulesets

pde:
Unil a fix for https://github.com/EFForg/https-everywhere/issues/1911 is
arranged.


jsha:
This looks good, though I think it will not fix all of the steam problems. I think some of those are caused by an interaction of the Akamai rule and CORS.

PR #1956, opened 2015-06-19
Fix bashisms in mktemp.sh and test-ruleset-coverage.sh

mnordhoff:


mnordhoff:
Ah... You just fixed this by changing `#!/bin/sh` to `#!/bin/bash`. Never mind?

jsha:
Yep, sorry for the trouble, and thanks for the fix!

PR #1957, opened 2015-06-19
[Employeeexpress.gov][WTF is my IP.com] Reenable IPv6-only rulesets

mnordhoff:
The checker only failed because the subdomain is IPv6-only. It passes when run from a dual-stack system.


jsha:
Thanks!

Issue #1595, opened 2015-04-27
Roll forward Google rulesets refactors


jsha:
@fuzzyroddis did a bunch of work to refactor the Google rulesets for the new coverage rules, but I had to revert them at the last minute for the 5.0 release because the refactor broke coverage for deep country TLDs like .co.uk. Here's the revert: https://github.com/EFForg/https-everywhere/pull/1217

We should roll forward that change, fixing #1216 and #1215 at the same time.

fuzzyroddis:
I wish I had time to do this but my days are spent looking after my health. Anyone feel free to do this :+1:

jsha:
Sorry to hear you are having health problems, @fuzzyroddis! No worries, we will take care of the roll forward.

fuzzyroddis:
Thank you so much for doing this, I really wish I had the time to do it.

PR #1964, opened 2015-06-21
[Google Services] Fix, reenable. Still lacks a billion tests.

mnordhoff:
Last time around, GoogleServices.xml was grandfathered through... This pull request fixes the simple immediate issue that got it disabled this time, but doesn't make the other changes that probably ought to be done.

Shortage of test coverage aside, it does pass the current tests.

I'm not sure if this is useful to merge or not.

(Also, should I have edited `ruleset-coverage-whitelist.txt`? If so, by hand?)

### DIFF
--- a/src/chrome/content/rules/GoogleServices.xml
+++ b/src/chrome/content/rules/GoogleServices.xml
@@ -1,8 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


jsha:
Thanks! I'm planning to fix this with https://github.com/EFForg/https-everywhere/issues/1595, which has been on my plate for way too long. There's a branch, forward-google, on the EFF repo that has a start on this. I'll try to wrap that up today since it unblocks a lot of fixes on the Google ruleset.

jsha:
Fixed by #1595.

Issue #1960, opened 2015-06-20
Active HTTPS-Everywhere rule "Spiegel" prevents loading of SPIEGEL Online (www.spiegel.de) videos


Wikinaut:
Loading of Flash videos from SPIEGEL Online http://www.spiegel.de is **broken** when the HTTPS Everywhere (HE) rule "Spiegel" is active.

### System information

I noticed the problem under these circumstances:
* Page/s where the problem is present: http://www.spiegel.de/video/
* System: OpenSuse Tumbleweed
* Browser and AddOns: Firefox 38.0.6 with HTTPS Everywhere 5.0.5 and with Shockwave 11.2.202.457

### Observed problem when having the HE rule "Spiegel" active and when you try to start a video

A loading spinner is shown for some seconds when ads are trying to be loaded at the begin, then a message is shown that the video cannot be played.

### Quick fix
Deactivation of the rule for "Spiegel" helps and the videos can then be loaded and shown as expected.

### Request
Asking the HE developers to fix the rule so that playing of videos becomes possible even when the "Spiegel" rule is active.

I am not experienced to analyze what the exact reason is, but as SPIEGEL Online and their videos are a major, worldwide player, I suggest that someone of your teams looks in the reported problem to help any HE users who want to access SPIEGEL Online and their videos.

finaljason:
Thanks for your report. I experienced the same issue and wrote a fix for it (see #1972)
The proposed change excludes the flash videos from HTTPS and does some other fixes.

Wikinaut:
@finaljason great. Confirmed, fix is working.

I copied Spiegel.xml into the User rule folder as explained on the HE web site, and this solved the problem. After restarting Firefox, and having HE enabled, I can now see the videos playing.

Thanks for the swift reply and fix.

jsha:
Fixes by #1972.

J0WI:
#closeme

PR #1972, opened 2015-06-21
[Spiegel.de] fixing ruleset

finaljason:
Based on issue #1960 I fixed the broken flash videos per exclusion, added test urls and did some general fixes

img/ gets rewritten to pics/ by the site itself, so i corrected the rule
shop.spiegel.de and wissen.spiegel.de are not longer functional due to bad ssl domain


Wikinaut:
@finaljason great. Confirmed, fix is working.

I copied Spiegel.xml into the User rule folder as explained on the HE web site, and this solved the problem. After restarting Firefox, and having HE enabled, I can now see the videos playing.

Thanks for the swift reply and fix.

jsha:
Excellent, thanks!

PR #1981, opened 2015-06-23
Remove exclusions for careers.stackoverflow.com

strout:
I fixed the mixed content on the site (I'm a dev on the Careers.SO team). The downgrade rule breaks parts of the /employer section of the site, such as [employer login](https://careers.stackoverflow.com/employer/login), by causing redirect loops.

### DIFF
--- a/src/chrome/content/rules/Stack-Exchange.xml
+++ b/src/chrome/content/rules/Stack-Exchange.xml
@@ -135,9 +135,6 @@
 <!-- Rules -->
 	
 	<!-- https links from other pages to these will cause MCB for important elements, hence the downgrades -->
-	<rule from="^https://careers\.stackoverflow\.com/(cities|employer)" 
-		to="http://careers.stackoverflow.com/$1" downgrade="1" />
-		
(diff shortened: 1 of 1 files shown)


strout:
Whoops, didn't get rid of the download rule; I'll fix that.

jsha:
Excellent, thanks for fixing that! Any chance you can also fix the mixed content on *.*.stackexchange.com (the downgrade rule below this one)?

strout:
@jsha I'll let the Q&A devs know about that so they can fix it.

jsha:
Great, thank you!

aalear:
Q&A dev @ SE here. Unfortunately we don't have certs for child meta sites since you can't buy a cert with a wildcard in the middle (*.*.stackexchange.com and meta.*.stackexchange.com  are both out), and maintaining 100+ individual certs and getting a new one for every new site launch isn't awesome. Moving meta sites to a different route is somewhere on the roadmap, but we're a ways off from that as far as I know. For the time being, the downgrade has to stay in place.

jsha:
Thanks for the insight, that makes sense. I hope when Let's Encrypt launches its automation can simplify some of these use cases.

PR #1982, opened 2015-06-23
php.net updates

LawnGnome:
This fixes two problems:

* www.php.net cannot be used with HTTPS any more, as the domain name will redirect to a mirror that (almost certainly) doesn't have a valid HTTPS configuration.
* bugs.php.net now sends a secure PHPSESSID cookie, so that rule can be enabled.

### DIFF
--- a/src/chrome/content/rules/PHP.xml
+++ b/src/chrome/content/rules/PHP.xml
@@ -9,10 +9,12 @@
 		- people ²
 		- talks ³
 		- windows ¹
+    - www ⁴
 
 	¹ Refused
 	² Dropped
 	³ Shows viewvc interface, valid cert
+  ⁴ Depends on the mirror, but the "best case" is an invalid certificate
 
 
 	Problematic subdomains:
@@ -24,7 +26,7 @@
 
 	Fully covered subdomains:
 
-		- (www.)?
+		- (null)
 		- bugs
 		- downloads
 		- edot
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/PHP.xml
```diff
@@ -65,14 +67,26 @@
 	<target host="*.php.net" />
 
 
-	<!--	Not secured by server:
-					-->
-	<!--securecookie host="^bugs\.php\.net$" name="^PHPSESSID$" /-->
-
+	<securecookie host="^bugs\.php\.net$" name="^PHPSESSID$" />
```
jsha: I think this is redundant with the securecookie rule below.

src/chrome/content/rules/PHP.xml
```diff
@@ -65,14 +67,26 @@
...
 	<securecookie host="^(?:\w*\.)?php\.net$" name=".+" />
 
 
-	<rule from="^http://((?:bugs|downloads|edit|git|master|museum|pear|qa|shared|snaps|static|svn|wiki|www)\.)?php\.net/"
+	<rule from="^http://((?:bugs|downloads|edit|git|master|museum|pear|qa|shared|snaps|static|svn|wiki)\.)?php\.net/"
 		to="https://$1php.net/" />
 
+  <test url="http://php.net/" />
+  <test url="http://bugs.php.net/" />
```
jsha: You can simplify this ruleset by moving these to `<target>`s, e.g.

```
<target host="bugs.php.net" />
<target host="downloads.php.net" />
```

Then you can replace the rule with:

```
<rule from="^http:"
        to="https:" />
```

And you won't need the test URLs because they are autogenerated by the target tags.

This is optional: If you'd like to submit as-is, that's fine. The changes I suggest will just make maintenance slightly simpler in future.

LawnGnome:
@jsha Thanks for the feedback. I've made those changes — I was obviously trying to make the most minimal changes possible, but let's get it right in one hit for maintainability.

I've also re-added pecl.php.net and added the new secure.php.net, since those are also fine with HTTPS (and, indeed, should prefer that where possible).

src/chrome/content/rules/PHP.xml
```diff
@@ -9,22 +9,17 @@
 		- people ²
 		- talks ³
 		- windows ¹
+		- www ⁴
 
 	¹ Refused
 	² Dropped
```
bjori: people.php.net has valid certificate

LawnGnome: @bjori Thanks; updated.

Issue #1980, opened 2015-06-23
Amazon Web Services breaking PremiumBeat.com samples


MDawg957:
http://www.premiumbeat.com/sfx/view/sweepers

When attempting to play a sample before purchase, nothing plays, but the moment Amazon Web Services rule is disabled the audio plays without any problems.

finaljason:
Thanks for your report. The fix above should solve your problem. Can you also test by adding the AmazonAWS.xml from the referenced commit into your UserRules folder (as described on https://www.eff.org/de/node/58830#testing)?

MDawg957:
I tested out the replacement ruleset and it still doesn't want to work.  Amazon Web Services doesn't show as something to disable when you first visit the page and play something, but if you go to another tab in Firefox and come back, it shows up.  Disabling it at that point allows the sample to play.

finaljason:
Hmm, this is strange. The fix i did was meant to exclude the sound files for the player from rewrite. The flash application which plays them doesn't seem to support HTTPS.
Before the fix I could confirm your issue, but now, with the rule, it works perfectly fine for me.

That the AWS rule didn't show up should mean it wasn't applicated. Have you tried using a clean Firefox profile? Alternatively, you can try to provide the URL where the content is loaded from for you (could be region specific). Just open the Firefox Network Analyzer (under Development Tools) and try to play a one of the previews with the AWS rule disabled. When you click the Play button, a connection to amazonaws.com will be shown in the analyzer window. Right-click onto it and provide me the link please.

MDawg957:
http://s3.amazonaws.com/pb_sfx_previews/0333_production_element_hit_sweeper/0333_production_element_hit_sweeper

http://s3.amazonaws.com/pb_sfx_previews/ seems to be the common address for any previews across the site from a few spot checks.

finaljason:
Exactly. Those addresses are excluded by the fix. The new rule:
``<exclusion pattern="^http://s3\.amazonaws\.com/pb_(sfx_)?previews/" />``
catches both http://s3.amazonaws.com/pb_sfx_previews/ and http://s3.amazonaws.com/pb_previews/ (which is for previews from the music section). I have no idea why this doesn't work out for you. Isn't it working with a completely clean firefox profile and just https-everywhere 5.0.5 + fix in the User rule directory?

MDawg957:
[...]

Issue #1927, opened 2015-06-14
forums.na.leagueoflegends.com doesn't support HTTPS


methanols:
https://forums.na.leagueoflegends.com/ produces Error 500.

An exclusion in [League_of_Legends.com.xml](https://github.com/EFForg/https-everywhere/blob/stable/src/chrome/content/rules/League_of_Legends.com.xml) probably fixes this.

finaljason:
Thanks for your report. The commit referenced above should solve your issue. If you want you can do a little more extensive testing with the rewritten ruleset, but there shouldn't be any greater problems.

PR #1983, opened 2015-06-23
Fix issues #1980/#1927

finaljason:
This pull request contains fixes for issue #1980 and #1927
Additionally I rewrote League_of_Legends.com.xml according to the style guidelines and tested it a bit for compatibility.


jsha:
Thanks!

PR #1993, opened 2015-06-25
remove nginx.org

nanaya1:
The only site in the rule, `trac`, doesn't actually work.


reedy:
Looks like an old SSL config or something based on Chromes response: ERR_SSL_VERSION_OR_CIPHER_MISMATCH

nanaya1:
it's like the ssl/sni is active but nothing is configured.

```
openssl s_client -connect trac.nginx.org:443 -servername trac.nginx.org
CONNECTED(00000003)
140010415498912:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:s23_clnt.c:770:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 318 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
---
```

reedy:
Do you know if it ever/recently worked?

I do agree with just deleting the whole file if it hasn't worked in a long time...

nanaya1:
I don't remember it ever worked. I'll try contacting relevant people to see how it should be.

nanaya1:
For reference, https is indeed not supported for that (sub)domain:

https://marc.info/?l=nginx&m=143598891015623&w=2

Issue #1897, opened 2015-06-10
"WP Engine.com" rule broken on x.y.wpengine.com subdomains


ivoanjo:
wpengine.com's SSL cert is only valid for  `*.wpengine.com, wpengine.com`, and not `*.*.wpengine.com`.

This breaks any site in the x.y.wpengine.com format, such as the following link:
<http://talkdesk.staging.wpengine.com/wp-content/uploads/2014/12/desk-widget-2.png>

Thanks for looking into this!

jeremyn:
It looks like pull request #2000 fixed this issue. Closing.

PR #2000, opened 2015-06-26
Update WP-Engine.xml

donced:
Exclude `*.*.wpengine.com` #1897

### DIFF
--- a/src/chrome/content/rules/WP-Engine.xml
+++ b/src/chrome/content/rules/WP-Engine.xml
@@ -76,10 +76,7 @@
 		<test url="http://talkdesk.staging.wpengine.com/"/>
 		<test url="http://kelvin.staging.wpengine.com/"/>
 		<test url="http://cfps.staging.wpengine.com/"/>
-		<test url="https://talkdesk.staging.wpengine.com/"/>
-		<test url="https://kelvin.staging.wpengine.com/"/>
-		<test url="https://cfps.staging.wpengine.com/"/>
-		
+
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/WP-Engine.xml
```diff
@@ -69,7 +70,19 @@
...
+	<exclusion pattern="^http://([\w.-]+)\.([\w-]+)\.wpengine\.com"/>
+
+		<test url="http://talkdesk.staging.wpengine.com/"/>
+		<test url="http://kelvin.staging.wpengine.com/"/>
+		<test url="http://cfps.staging.wpengine.com/"/>
+		<test url="https://talkdesk.staging.wpengine.com/"/>
+		<test url="https://kelvin.staging.wpengine.com/"/>
+		<test url="https://cfps.staging.wpengine.com/"/>
+		
+	<!-- https links from other pages to these will cause MCB for important elements, hence the downgrades -->
```
jsha: Can you confirm there are other pages that link to the HTTPS version of these domains?

I ask because we'd really like to decrease the number of downgrade rules, since they make certain optimizations harder (i.e. listening only on HTTP URLs).

donced: To be honest, I'm not familiar with wpengine.com. Now that I think of it, judging by the "staging" keyword, I don't think other pages would link to `*.*.wpengine.com`. I removed the downgrade rule.

PR #2004, opened 2015-06-26
[Pol.dk] Disabled ruleset: connections are refused

fuglede:
As of today or yesterday, the existing rule would cause all images on https://politiken.dk not to be loaded. Devs are contacted, and hopefully it will be fixed soon enough.

As an aside: For PRs like these, fixing broken websites, do you want an companion issue submitted as well?


jsha:
Submitting ruleset-fixing PRs without a companion issue is fine. Thanks!

PR #2008, opened 2015-06-27
Added ethicaldata.co.uk

jacobgelling:


src/chrome/content/rules/Ethical_Data.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="Ethical Data">
+
+	<target host="ethicaldata.co.uk" />
+	<target host="*.ethicaldata.co.uk" />
```
reedy: Is there any benefit from this being a wildcard? Are they using numerous and random domains?

jacobgelling: @reedy Yes, each client of them gets one.

reedy: Aha, that makes sense then. Might be worth adding a comment to this end, but it's not a big deal

Issue #1778, opened 2015-05-28
Latest translations from Transifex missing "or"


jsha:
Steps to reproduce:

```
git clone https://git.torproject.org/translation.git
cd translation
git checkout https_everywhere_completed
cd ~/https-everywhere/src/chrome/locale/
for LOCALE in * ; do rsync -av ~/translation/$LOCALE/ $LOCALE/ ; done
cd ../../../
./test.sh
```

Expected result: Test passes.
Actual result: 
```
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
\_ missing for locale et
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
\_ missing for locale fo
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
ssl-observatory.prefs.show_cert_warning
ssl-observatory.prefs.show_cert_warning_tooltip
\_ missing for locale he
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
\_ missing for locale km
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
\_ missing for locale ko
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
ssl-observatory.prefs.show_cert_warning
ssl-observatory.prefs.show_cert_warning_tooltip
\_ missing for locale lt
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
ssl-observatory.prefs.show_cert_warning
ssl-observatory.prefs.show_cert_warning_tooltip
\_ missing for locale ms_MY
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
https-everywhere.menu.showCounter
https-everywhere.popup.keep
https-everywhere.popup.paragraph1
https-everywhere.popup.paragraph2
https-everywhere.popup.paragraph3
https-everywhere.popup.revert
https-everywhere.popup.title
https-everywhere.ruleset-tests.status_cancel_button
https-everywhere.ruleset-tests.status_start_button
https-everywhere.ruleset-tests.status_title
ssl-observatory.prefs.show_cert_warning
ssl-observatory.prefs.show_cert_warning_tooltip
\_ missing for locale si_LK
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
\_ missing for locale sk_SK
https-everywhere.about.or
https-everywhere.menu.blockHttpRequests
ssl-observatory.prefs.show_cert_warning
[...]

PR #2009, opened 2015-06-27
Various fixes and changes to Firefox toolbar button

fuglede:
As an extension of #1035, this adds the option to display the total number of applicable rulesets on a given site, rather than just the active ones. For users who don't mind breakage, this has the advantage of easily identifying the opportunity to use less-than-optimal rulesets.

I didn't play around with this code before, so apologies in advance if things are messier than they should be. On the other hand, I did clean up a few very minor things myself.

~~Note: There is one hardcoded string in this PR that should get fixed. After seeing #1778, though, it is not entirely clear to me how to match up the locales with Transifex optimally; should the string be added (in English) to all .dtd's manually -- or what can be automated in this?~~ Localisation added in third commit.


src/chrome/content/toolbar_button.js
```diff
@@ -163,14 +148,75 @@ httpsEverywhere.toolbarButton = {
...
+    
+    if (counterItem) { // If we can find the counterItem like this, we can find everything in the UI.
+      uiFound = true;
+      var counterTotalItem = document.getElementById('https-everywhere-counter-total-item');
+      var httpNowhereItem = document.getElementById('http-nowhere-item');
+      var toolbarbutton = document.getElementById('https-everywhere-button');
+    } else {
+      // If the toolbar button is hidden away in the Firefox hamburger menu,
+      // we won't find it by the above, and need to look somewhere else
+      let palette = document.getElementById("navigator-toolbox").palette;
```
jsha: I suspect this can return undefined, or may in the future if Firefox changes its XUL. We should check the result of getElementById before dereferencing .palette.

src/chrome/content/toolbar_button.js
```diff
@@ -163,14 +148,75 @@ httpsEverywhere.toolbarButton = {
...
   /**
    * Update the rulesets applied counter for the current tab.
    */
   updateRulesetsApplied: function() {
     var toolbarbutton = document.getElementById('https-everywhere-button');
     if (!toolbarbutton) {
-      return;
+      // This becomes relevant for instance when the toolbar button is
+      // hidden away in the Firefox hamburger menu.
+      let palette = document.getElementById("navigator-toolbox").palette;
```
jsha: Same here- check for undefined.

jsha:
Other than those two comments, looks good. @diracdeltas was going to double-check the appearance on OS X, since we've had some OS X display issues with the counter in the past.

src/chrome/content/toolbar_button.js
fuglede: Done. I guess one could also check if palette itself is undefined afterwards.

fuglede:
Actually, it had three (rather than two) appearances. I suppose that's an argument for abstraction.

fuglede:
One more note: When the label is shown while the button is in the hamburger menu, the button (+ label) becomes too big to fit in the default grid, which can make it look a bit wonky. One obvious fix to this would be to simply not show the label when the button is hidden away.

semenko:
@fuglede Sorry this didn't end up getting merged before a conflict -- if you'd like to update it, i'd be happy to merge it.

fuglede:
I'll see if I can get around to that at some point. Great job on the review spree though!

semenko:
Thanks! :+1:

fuglede:
Fingers crossed.

fuglede:
Okay, the merge was trivial enough, but something else has changed, as now the button is gone altogether. I don't quite have an overview of why that could be, but perhaps #3727 is related?

cowlicks:
Closing since we've moved to webextensions. @fuglede let me know if you'd like to re-open this.

PR #2011, opened 2015-06-27
Create new buzer.de.xml rule

zerozan:


src/chrome/content/rules/buzer.de.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="buzerde">
+    <target host="buzer.de" />
+    <target host="www.buzer.de" />
+
+    <rule from="^http://(www\.)*buzer\.de/" to="https://www.buzer.de/" />
```
reedy: The use of * doesn't make sense here

This could be http://buzer.de. http://www.buzer.de or even http://www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.www.buzer.de


Change it for a ? then this should be good to merge

semenko:
Closing since this has outstanding work, and hasn't been touched in a while. Feel free to reopen if you'd like to have this merged.

PR #2022, opened 2015-06-30
[StartCom] Should not be disabled

TonyZTan:
This rule should not be disabled. The checker encounters an error at https://auth.startssl.com/ because it requires a client certificate for the handshake to complete, but this does not mean that https-everywhere should not include this rule. Is there a way to override this error?

Thanks!

### DIFF
--- a/src/chrome/content/rules/StartCom.xml
+++ b/src/chrome/content/rules/StartCom.xml
@@ -1,10 +1,4 @@
-
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://auth.startssl.com/ => https://auth.startssl.com/: (35, 'error:14094410:SSL routines:SSL3_READ_BYTES:sslv3 alert handshake failure')
-
--->
-<ruleset name="StartCom" default_off='failed ruleset test'>
+<ruleset name="StartCom">
   <target host="startssl.com" />
   <target host="*.startssl.com" />
   <target host="startssl.net" />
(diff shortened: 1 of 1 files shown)


rugk:
:+1: :smiley:

src/chrome/content/rules/StartCom.xml
```diff
@@ -62,6 +56,6 @@ Fetch error: http://auth.startssl.com/ => https://auth.startssl.com/: (35, 'erro
 
   <!-- host startcom.org responds neither on 80 nor on 443,
     but should be protected from MitM and redirected to https://www.startcom.org -->
-  <rule from="^http://startcom\.org/" to="https://www.startcom.org/" />
```
jsha: Why add a `*` here?

TonyZTan: I was worried that "^http://startcom.org/" wouldn't match "http://startcom.org" (without the final /), so I used the * to indicate there could be a slash or not. Does it actually affect matching?

TonyZTan: Do you know if there is a way to prevent the checker from failing at https://auth.startssl.com, which requires a client certificate?

2d1: Yes, instead of excluding anything in auth\\\.startssl\\\.com/.*\\\.cr[lt], exclude \\\.cr\[lt\](?:$|\?) for everything, i.e.
```xml
<exclusion pattern="^http://.*\.cr[lt](?:$|\?)" />
```
and don't include auth.startssl.com among the tests required by the test coverage checker heuristics for that exclusion.

TonyZTan:
This rule should be re-enabled.

PR #2023, opened 2015-06-30
[Andover.edu] Should not be disabled

TonyZTan:
This rule should not be disabled. The checker encounters an error at three hosts because they either require special ports to access or are on an internal network, but this does not mean that https-everywhere should not include this rule set. Is there a way to override this error?

Thanks!

### DIFF
--- a/src/chrome/content/rules/Phillips_Academy_Andover.xml
+++ b/src/chrome/content/rules/Phillips_Academy_Andover.xml
@@ -1,6 +1,5 @@
 <ruleset name="Phillips Academy">
 
-	<target host="andover.edu" />
 	<target host="www.andover.edu" />
 	<target host="panet.andover.edu" />
 	<target host="canvas.andover.edu" />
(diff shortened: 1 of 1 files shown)


reedy:
I'm not sure exactly, but this is good examples of stuff that @jsha was looking for ways to improve the checker. I believe it's being worked on so that the ruleset checker will disable only the broken hosts, not the whole file

One way of doing this would be to have a default off separate ruleset, and people can enable it who need it. That, or we have some way of annotating the hosts (extra attribute) that they're not globally accessible or similar

TonyZTan:
@reedy Is there a way to keep these hosts there without them being disabled? Somehow marking them as "do not check"? Because #2022 is an example where a working host is being disabled by the checker. Moreover, there is no harm in keeping domains that are internal or require special ports to access in the rulesets :smile:

jsha:
@TonyZTan: Thanks for the patch! It looks like @2d1 has beat you to the fix, by disabling the relevant hostnames.

I do think we should not include internal-only hostnames, since they are impossible to verify and they are not useful for 99% of HTTPS Everywhere users.

Externally visible hostnames that require an alternate port are an interesting case. I think it makes sense to allow those, with a test URL that uses that alternate port. We would also have to suppress the implicit test URL for the hostname on 443. We could do that with an exclusion (not optimal) or with a new `no-implicit-test` tag on the target host.

TonyZTan:
@jsha I did a check, and it seems like mail.andover.edu is actually accessible on port 443, but will return a 401 unauthorized page. For users that use the site, the correct link is https://mail.andover.edu/owa, which requires login credentials. Perhaps, we could create a rule that only covers the "owa" link. but would the checker let it through, given that the checker does not have the credentials? Maybe this becomes a similar issue to #2022, which requires a client authentication certificate.

TonyZTan:
Moving to pull request #2042.

PR #2024, opened 2015-06-30
Rules for orange.fr

LoHub:


src/chrome/content/rules/orange.fr.xml
```diff
@@ -0,0 +1,23 @@
+<ruleset name="Orange France (Partial)" >
+  <target host="boutique.orange.fr" />
+  <target host="id.orange.fr" />
+  <target host="iz.orange.fr" />
+  <target host="web.cloud.orange.fr" />
+  <target host="assistance.orange.fr" />
+  <target host="jeu.orange.fr" />
+  <target host="*.espaceclientv*.orange.fr" />
```
2d1: Given there are multiple targets with misplaced asterisks, I assume that you mean to target $(subdomain).*\.orange\.fr. Targets accept only wildcards or only literal text for any given domain level.

jsha:
Test output:

```
['src/chrome/content/rules/Orange.xml', 'src/chrome/content/rules/orange.xml'] failed case-insensitivity testing.

```

Can you incorporate your change into the existing Orange.xml?

PR #2026, opened 2015-06-30
Create netvibes.com.xml

MatthieuBizien:
Inspired from 131002.net.xml


fuglede:
This one fails, because Netvibes.com.xml (with a capital N) already exists.

2d1:
Per @fuglede's comment, please make additions and changes for netvibes.com to Netvibes.com.xml, and more generally grep all rulesets for existing domains before making a new one.

PR #2027, opened 2015-06-30
Create slate.fr.xml

MatthieuBizien:


fuglede:
You should probably capitalize the "s" of slate; that's what all other ruleset files do.

MatthieuBizien:
It's done!

Issue #1228, opened 2015-03-23
Fonts broken on slate.com


abbeyj:
I'm using Firefox 36.0.4 on Windows and HTTPS-Everywhere 5.0development.4.

The rule "CDN Slate.com (partial)" is breaking custom fonts on http://slate.com/ .  Go to that URL and note that the headlines are all in a serif font (Times New Roman?).  If you disable the rule, the headlines appear in the custom font "sl-Apres" as intended.

This seems to be a problem with MIME types.  For instance, http://static.cdnslate.com/etc/designs/slate/fonts/ApresRegular/c60bb3c2-10e5-4eee-a0d4-e0e221eca5c0-3.woff has "Content-Type: application/font-woff".  With the rule enabled, this URL is rewritten to https://a248.e.akamai.net/f/248/1/1/www.slate.com/etc/designs/slate/fonts/ApresRegular/c60bb3c2-10e5-4eee-a0d4-e0e221eca5c0-3.woff which has "Content-Type: text/html; charset=UTF-8", disallowing use as a font.

jsha:
Thanks for reporting! I'll disable the rule.

PR #2028, opened 2015-06-30
Create slate.com.xml

MatthieuBizien:


fuglede:
Note that rules/Slate.com.xml already exists.

jsha:
Indeed, and it was caught by the test. @MatthieuBizien, it looks like the existing rule is disabled per https://github.com/EFForg/https-everywhere/issues/1228. Closing this ticket. If you can confirm custom fonts are now working on the HTTPS version, we'll accept a pull request re-enabling the existing ruleset.

PR #2032, opened 2015-06-30
New rule computerguard.de

rugk:
As a security forum we recently enabled HTTPS on [our website] (https://www.computerguard.de/).
Now we think it would be a good idea to include us in HTTPS-Everywhere.

Just a small note: As we allow including images from outer sources some users choose http:// image host and so mixed content errors can occur. However this is quite rarely and we always encourage our users to host the images on https sites or on our own forum to prevent this.
I've also described this in the comments of this rule.
But this is not really a problem as most of today's browsers does not block mixed content images anyway.

### DIFF
--- a/src/chrome/content/rules/Threema.xml
+++ b/src/chrome/content/rules/Threema.xml
@@ -1,5 +1,18 @@
 <ruleset name="Threema">
-<target host="*.threema.ch" />
-<target host="threema.ch" />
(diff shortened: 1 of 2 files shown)


KommX:
If you are the admin you could also think about submitting your site in the HSTS preload list:
https://hstspreload.appspot.com/

rugk:
Yes I am the admin. And we have already thought about this, but as it could be difficult to remove it from the list again until now we didn't added our website to the list.
In HTTP Everywhere it's much simpler to delete or modify a rule.

PR #2033, opened 2015-06-30
Add rules for franckgirard.net

LoHub:


src/chrome/content/rules/franckgirard.net.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="franckgirard.net">
+  <target host="franckgirard.net" />
+  <target host="www.franckgirard.net" />
```
2d1: www.franckgirard.net redirects to www.www.franckgirard.net, which doesn't exist. So www.franckgirard.net must be rewritten to ^franckgirard.net.

J0WI:
LGTM. Sorry for the delay!

PR #2035, opened 2015-07-01
Create schema.org.xml

J0WI:


Jeni4:
Wouldn't this be a little better way to write the ruleset?
```xml
<ruleset name="schema.org">
	<target host="schema.org"/>
	<target host="www.schema.org"/>
	<target host="bib.schema.org"/>

	<rule from="^http:\/\/(www\.|bib\.)?schema\.org\/" to="https://$1schema.org/" />
</ruleset>
```

J0WI:
Nope, RegEx would be an additional overhead here.
Please have a look at our Ruleset-Styleguide: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

Jeni4:
Can see what you mean.
My experience with creating contributing rulesets predates the Ruleset-Styleguide. Hence my thought.

jsha:
Note: At present, will will still use a regex for the simple http -> https rewrite, so it's probably not a big overhead savings. It's more about an improvement in clarity. And, eventually, we should change the code to skip the regex in such simple cases.

PR #2042, opened 2015-07-02
[Andover.edu] Fix mail.andover.edu rule

TonyZTan:
Continuation of pull request #2023. Attempts to fix the mail.andover.edu rule and adds a new rule for the root domain.

Please note that https://mail.andover.edu/owa requires login credentials, and thus may cause the rule to fail checks by the automatic checker. This is similar to #2022, where the website requires authentication with a client certificate.

### DIFF
--- a/src/chrome/content/rules/Phillips_Academy_Andover.xml
+++ b/src/chrome/content/rules/Phillips_Academy_Andover.xml
@@ -18,6 +18,11 @@
 	<!-- mail.andover.edu returns an error, but should be included to protect mail.andover.edu/owa -->
 	<target host="mail.andover.edu" />
 	
+	<test url="http://andover.edu" />
+	<test url="http://andover.edu/" />
+	<test url="http://mail.andover.edu" />
+	<test url="http://mail.andover.edu/" />
+	 
 	<securecookie host=".+" name=".+" />
 	
 	<!-- redirects andover.edu to https://www.andover.edu -->
--- a/src/chrome/content/rules/Phillips_Academy_Andover.xml
+++ b/src/chrome/content/rules/Phillips_Academy_Andover.xml
@@ -20,7 +20,6 @@
 	<target host="mail.andover.edu" />
 		-->
 	
-	<test url="http://andover.edu" />
 	<test url="http://andover.edu/" />
 	 
 	<securecookie host=".+" name=".+" />
--- a/src/chrome/content/rules/Phillips_Academy_Andover.xml
+++ b/src/chrome/content/rules/Phillips_Academy_Andover.xml
@@ -26,12 +26,15 @@
 	<securecookie host=".+" name=".+" />
 	
 	<!-- redirects andover.edu to https://www.andover.edu -->
-	<rule from="^http://andover\.edu/*"
+	<rule from="^http://andover\.edu/"
 		to="https://www.andover.edu/" />
 	
+	
 	<!-- redirects http://mail.andover.edu/ (with or without "/") to https://mail.andover.edu/owa -->
+	<!-- commented out until 'no-test' attribute is available
 	<rule from="^http://mail\.andover\.edu/*$"
 		to="https://mail.andover.edu/owa" />
+		-->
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Phillips_Academy_Andover.xml
```diff
@@ -17,13 +6,32 @@
...
+	
+	<test url="http://andover.edu" />
+	<test url="http://andover.edu/" />
+	<test url="http://mail.andover.edu" />
+	<test url="http://mail.andover.edu/" />
+	 
 	<securecookie host=".+" name=".+" />
+	
+	<!-- redirects andover.edu to https://www.andover.edu -->
+	<rule from="^http://andover\.edu/*"
```
jsha: Assuming you intend the `*` to indicate repeats of the `/`, it should be `+` instead, to require at least one slash.

src/chrome/content/rules/Phillips_Academy_Andover.xml
```diff
@@ -17,13 +6,32 @@
...
 	<target host="*.patechmasters.com" />
 	
+	
+	<!-- andover.edu does not respond, but should be redirected to https://www.andover.edu to prevent MitM. -->
+	<target host="andover.edu" />
+	
+	<!-- mail.andover.edu returns an error, but should be included to protect mail.andover.edu/owa -->
+	<target host="mail.andover.edu" />
+	
+	<test url="http://andover.edu" />
```
jsha: Both of the test urls here are duplicated with each other. Also, those test urls will be implicitly created by the corresponding `<target>` tag. Extra test URLs are only needed then the implicit test URL is not sufficient, e.g. when a domain has some problematic parts or needs a more complex URL.

src/chrome/content/rules/Phillips_Academy_Andover.xml
TonyZTan: @jsha I was worried that "^http://andover\.edu/" wouldn't match "http://andover.edu" (without the final /), so I used the <code>*</code> to indicate there could be a slash or not. Does it actually affect matching? Is the final "/" something that is automatically added in requests?

src/chrome/content/rules/Phillips_Academy_Andover.xml
TonyZTan: @jsha I don't understand, but the [Travis-CI build](https://travis-ci.org/EFForg/https-everywhere/builds/69226922) asked for test URLs:

    ERROR src/chrome/content/rules/Phillips_Academy_Andover.xml: Not enough tests (1 vs 2) for <Rule from '^http://andover\.edu/*' to 'https://www.andover.edu/'>
    ERROR src/chrome/content/rules/Phillips_Academy_Andover.xml: Not enough tests (1 vs 2) for <Rule from '^http://mail\.andover\.edu/*$' to 'https://mail.andover.edu/owa'>

TonyZTan: Do you know if the ruleset checker is going to fail on https://mail.andover.edu/owa?

src/chrome/content/rules/Phillips_Academy_Andover.xml
jsha: I'm 95% positive the final "/" is automatically added to all requests before matching.

src/chrome/content/rules/Phillips_Academy_Andover.xml
jsha: Aha, now I see the difference between the test cases. After changing the * to a + below, you can have test URLs like:

```
http://andover.edu/
http://andover.edu////
```

Yes, the ruleset checker will fail on mail.andover.edu currently. I've filed a ticket to add a `no-test` attribute on target tags, to fix this problem.

TonyZTan:
Modified to remove the mail.andover.edu rule which requires the `no-test` attribute. Please merge the remaining change to cover the root domain. I will open a new pull request when the attribute becomes usable. Thanks!

PR #2043, opened 2015-07-02
[Washington Post] Now supports HTTPS

TonyZTan:
Initial test in Firefox shows no major problems with this rule.

https://www.washingtonpost.com/blogs/the-switch/wp/2015/06/30/washington-post-starts-to-automatically-encrypt-part-of-web-site-for-visitors/


jsha:
Great news about Washington Post! It looks like this change is probably too aggressive. I tried a couple of the subdomains from the existing comment (findnsave, knowmore), and they still exhibit the problems listed. Could you scale it back so it doesn't doesn't try to rewrite those domains? The conservative thing would probably be to add rewrites for www.washingtonpost.com and washingtonpost.com.

jsha:
@TonyZTan: Thanks for kicking off this pull request. It looks like @2d1 has incorporate these fixes and more into the laster master, so I'm going to close this out as a duplicate.

PR #2045, opened 2015-07-02
Adding rule for oxfam.org.uk

mapkyca:


src/chrome/content/rules/Oxfam.org.uk.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Oxfam.org.uk">
+  <target host="oxfam.org.uk" />
+  <target host="www.oxfam.org.uk" />
```
2d1: This ruleset doesn't cover donate, which seems to work when rewritten directly. What's here looks fine, though.

PR #2048, opened 2015-07-02
Better translation

LoHub:


jsha:
Thanks for the pull request! We actually maintain our translations in Transifex, and then copy them into this repository. Can you update the translation there? https://www.transifex.com/signup/?join_project=torproject

PR #2050, opened 2015-07-03
dd-wrt.com

GdrjirfRtrUhYUdrjUMm:
Cleanup to indicate partial rule and addition of self-signed rule.


donced:
Why not just make a note in `DD-WRT.xml` indicates that `svn` is a non-functional subdomain?

PR #2054, opened 2015-07-03
Add rule for konachan.com

PotcFdk:


semenko:
Hm, looks like the WWW might have a redirect loop: https://www.konachan.com

PotcFdk:
Are you sure? It looks fine to me.
With the rule enabled, your link opens the home page just fine.

CLI test:
```
$ curl -LI https://www.konachan.com/
Server: nginx
Date: Sun, 06 Sep 2015 22:06:43 GMT
Content-Type: text/html
Content-Length: 178
Connection: keep-alive
Location: http://konachan.com/

HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Content-Length: 5062
Connection: keep-alive
Vary: Accept-Encoding
Status: 200 OK
[...]
```

semenko:
Ah good call -- yep! Thanks!

PR #2055, opened 2015-07-05
Improve Unhosted.xml

donced:
The site's SSL cert is only valid for unhosted.org and apps.unhosted.org, so *.unhosted.org is dropped. There is also redirect rule for www.unhosted.org because it presents self-signed cert.


jsha:
Fixed by #2055.

PR #2057, opened 2015-07-05
Delete Agar.io

### LINKED ISSUE #2056: Remove Agar.io
Hello.

I'm the developer of Agar.io. Could you remove it from the list of sites that get redirected to the https version? Right now, the https version has to use wss, which means that we have to pass through a reverse proxy to kill it and turn into unsecure websockets before handing it to the game servers, which results in an unnecessary increase in latency. We currently only support https because Facebook demands it for their Canvas app stuff, but we prefer to send users to the http version.

Thank you.

(See here for domain ownership proof: http://agar.io/2056.txt)

Matheus28:
Fixes #2056


jsha:
Instead of deleting, please edit the file to say:

```
<ruleset name="Agar.io" default_off="webmaster request">
```

That way, if somebody else comes along and wants to add the rule, they will see there is an existing one and why it is turned off.

Please also add a comment at the top linking to the GitHub issue (https://github.com/EFForg/https-everywhere/issues/2056).

Thanks for reaching out to us, and I look forward to re-enabling HTTPS once you've finished the necessary server-side work!

Matheus28:
Ok I'll create another pull request as I can't find a way to do it through GitHub without cloning the repo locally.

Thank you.

PR #2063, opened 2015-07-06
[CIMB Indonesia] New ruleset 

donced:


src/chrome/content/rules/CIMB-Indonesia.xml
```diff
@@ -0,0 +1,48 @@
...
+	<target host="www.cimbclicks.co.id" />
+	<target host=    "itradecimb.co.id" />
+	<target host="www.itradecimb.co.id" />
+
+	<securecookie host="^www\.cimbclicks\.co\.id$" name=".*" />
+	<securecookie host="^www\.itradecimb\.co\.id$" name=".*" />
+	<securecookie host="^www\.cimbniaga\.com$" name=".*" />
+
+	<rule from="^http://cimbniaga\.com/"
+		to="https://www.cimbniaga.com/" />
```
2d1: It would be nice if you added comments stating why you must rewrite from ^domain to www.domain in each of these rules, for the sake of webmasters reading the Atlas.

src/chrome/content/rules/CIMB-Indonesia.xml
```diff
@@ -0,0 +1,48 @@
...
+
+		<test url="http://www.cimbniaga.com/" />
```
2d1: All three explicit tests are redundant, as literal targets generate implicit tests.

2d1:

PR #2064, opened 2015-07-06
Update BroadbandReports.xml

### LINKED ISSUE #2058: 5.0.5 Breaks redirect on DSLReports.ca
http://www.dslreports.ca is redirected to http://www.dslreports.com and this has worked fine with HTTPS Everywhere until 5.0.5, re-writing it properly as https://www.dslreports.com.  However, with this latest update it's now re-writing it as https://secure.dslreports.com which is not a valid subdomain for their certificate.  There are no rules that I could find for that site so I'm not sure what's causing it.

donced:
Fix #2058 and rename ruleset


src/chrome/content/rules/BroadbandReports.xml
```diff
@@ -1,45 +0,0 @@
...
-
-
-	<securecookie host="^\.dslr\.net$" name=".+" />
-
-
-	<!--	Redirect loop:
-
-		https://mail1.eff.org/pipermail/https-everywhere-rules/2012-October/001352.html
-												-->
-		<exclusion pattern="^http://(?:www\.)?(?:broadband|dsl)reports\.com/forum/[a-qs-z][_a-z]+(?:,[a-z]+|,?\d+)?(?:~[=a-z]+)?$" />
```
jsha: Looks like you got rid of the exclusion patterns. Did you confirm that these are no longer necessary?

donced: the forum link changed to `/forums` and HTTPS works fine over there.

PR #2065, opened 2015-07-06
Added policies.tilt.com to exclusion pattern

godsflaw:
I work at tilt.com, and we just added this new domain (hosted on a third party, so we cannot provide SSL).


amackera:
+1 (I also work at Tilt)

jsha:
Great, thanks very much for the pull request!

PR #2073, opened 2015-07-07
Update US-government.xml (whitehouse.gov)

youdly:
I'm thinking of "decentralizing" the US-government.xml.
Federal websites should be in better shape (as in HTTPS support) by now judging by the recent [memorandum](https://www.whitehouse.gov/sites/default/files/omb/memoranda/2015/m-15-13.pdf).


jsha:
Thanks! Yes, refactoring US-government.xml into various files would be a nice maintainability improvement.

PR #2082, opened 2015-07-08
Remove podbean mixcontent & subdomains

andychase:
The site isn't displayed correctly due to content being blocked on browsers that block mixed-content content and on many other podcast sites (such as pagebreakpodcast.com) the podcast doesn't load with no indication that the cause is https everywhere.

screen shot 2015-07-07 at 8 10 21 pm
screen shot 2015-07-07 at 7 21 39 pm

### DIFF
--- a/src/chrome/content/rules/Podbean.com.xml
+++ b/src/chrome/content/rules/Podbean.com.xml
@@ -1,17 +1,7 @@
-<ruleset name="Podbean.com (partial)">
+<ruleset name="Podbean.com (partial)" platform="mixedcontent">
 
 	<target host="podbean.com" />
-	<target host="*.podbean.com" />
-		<!--
-			At least media57 504s
-
-			Only hosts images & audio files
-							-->
-		<exclusion pattern="^http://media\d+\." />
-
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Podbean.com.xml
```diff
@@ -1,19 +0,0 @@
-<ruleset name="Podbean.com (partial)">
```
donced: Suggest to use `<ruleset name="Podbean.com (partial)" platform="mixedcontent"> ` as per [guideline](https://www.eff.org/https-everywhere/rulesets). It will disable the ruleset for browsers that block mixed content.

jsha:
I agree with @donced. It's nice to support this ruleset in browsers that don't block mixed content (i.e., the Tor Browser).

Also, in general if a site still exists but is broken, we like to keep the rule around rather than delete, as a hint to future maintainers that might want to re-add it.

andychase:
@jsha 

I made @donced's change. I had to remove subdomains as well because on the pagebreak example I mentioned the url is `http://pagebreakpodcast.podbean.com/[..]`

PR #2083, opened 2015-07-08
[JustPaste.it] New ruleset

donced:


src/chrome/content/rules/JustPaste.it.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="JustPaste.it">
+	<target host=    "justpaste.it" />
+	<target host="www.justpaste.it" />
+
+  	<securecookie host="^justpaste\.it$" name=".*" />
+
+	<rule from="^http://www\.justpaste\.it/"
+		to="https://justpaste.it/" />
```
2d1: Why redirect to ^justpaste.it? www.justpaste.it seems to work fine. If there's no problem, please don't preempt the server's redirect. Doing so only gives an adversary on the wire greater confidence that the user is using https-everywhere.

src/chrome/content/rules/JustPaste.it.xml
```diff
@@ -0,0 +1,14 @@
...
+
+		<test url="http://justpaste.it/" />
```
2d1: This test is redundant, as literal targets generate implicit tests.

2d1:

src/chrome/content/rules/JustPaste.it.xml
donced: I didn't consider such risk. I add it to save server's redirection.

PR #2084, opened 2015-07-08
[word.camera] New ruleset

donced:


src/chrome/content/rules/word.camera.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="word.camera">
+	<target host=    "word.camera" />
+	<target host="www.word.camera" />
+
+	<rule from="^http://www\.word\.camera/"
+		to="https://word.camera/" />
+
```
2d1: It would be nice if you added a comment stating that the cert doesn't match www, for the sake of webmasters reading the Atlas.

src/chrome/content/rules/word.camera.xml
```diff
@@ -0,0 +1,12 @@
...
+		<test url="http://word.camera/" />
```
2d1: This test is redundant, as literal targets generate implicit tests.

2d1:

PR #2092, opened 2015-07-08
exclusion for wacken.com anfahrt-widget

phibo23:
the travel planning widget for the Wacken festival does not work with https, I added an exclusion. The festival will take place from July 30 to August 1, 2015, so please merge this soon.

### DIFF
--- a/src/chrome/content/rules/Wacken.com.xml
+++ b/src/chrome/content/rules/Wacken.com.xml
@@ -12,4 +12,6 @@
 
 	<rule from="^http:"
 		to="https:" />
+	<!-- https breaks the widget -->
+	<exclusion pattern="^http:\/\/www\.wacken\.com\/anfahrt-widget\/index-(de|en)\.php" />
 </ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Wacken.com.xml
```diff
@@ -12,4 +12,8 @@
 
 	<rule from="^http:"
 		to="https:" />
+	<!-- https breaks the widget -->
+	<exclusion pattern="^http:\/\/www\.wacken\.com\/anfahrt-widget\/index-(de|en)\.php" />
```
donced: you don't have to escape forward slash.

phibo23: ah, ok, I did that habitually since / is a pattern delimiter in many languages.

2d1:
It would be nice if you explained why it doesn't work, and since it's due to mixed "active content", ideally the excluded coverage into a platform=mixedcontent ruleset. It works fine for torbrower.

phibo23:
Not sure what happens there. There widget itself is an iframe to a site that definitely does not work with https. However, something on these pages causes css and js resources that are available via https to be requested via http, causing the mixed content warnings and blocks in FF and Chrome. 
I didn't spend too much time looking into the issue because these pages are only relevant for a few more weeks and I just wanted to provide a hot fix so people can use the planning tool without running into this issue.

PR #2093, opened 2015-07-09
[CloudMagic] New ruleset

donced:


src/chrome/content/rules/CloudMagic.xml
```diff
@@ -0,0 +1,25 @@
...
+
+<ruleset name="CloudMagic">
+	<target host=       "cloudmagic.com" />
+	<target host=   "www.cloudmagic.com" />
+	<target host="static.cloudmagic.com" />
+	<target host=   "api.cloudmagic.com" />
+
+  	<securecookie host="^cloudmagic\.com$" name=".*" />
+
+	<rule from="^http://www\.cloudmagic\.com/"
```
2d1: Why redirect to ^cloudmagic.com? www.cloudmagic.com seems to work fine. If there's no problem, please don't preempt the server's redirect. Doing so only gives an adversary on the wire greater confidence that the user is using https-everywhere.

src/chrome/content/rules/CloudMagic.xml
```diff
@@ -0,0 +1,25 @@
...
+		to="https://cloudmagic.com/" />
+
+		<test url="http://cloudmagic.com/" />
```
2d1: This test is redundant, as literal targets generate implicit tests.

2d1:

PR #2094, opened 2015-07-09
[iplocation.net] New ruleset

donced:


src/chrome/content/rules/iplocation.net.xml
```diff
@@ -0,0 +1,23 @@
...
+
+<ruleset name="iplocation.net">
+
+	<target host=    "iplocation.net" />
+	<target host="www.iplocation.net" />
+
+  	<securecookie host="^www\.iplocation\.net$" name=".*" />
+
+	<rule from="^http://iplocation\.net/"
+		to="https://www.iplocation.net/" />
```
2d1: Why redirect to www? ^iplocation.net seems to work fine. If there's no problem, please don't preempt the server's redirect. Doing so only gives an adversary on the wire greater confidence that the user is using https-everywhere.

src/chrome/content/rules/iplocation.net.xml
```diff
@@ -0,0 +1,23 @@
...
+
+		<test url="http://www.iplocation.net/" />
```
2d1: This test is redundant, as literal targets generate implicit tests.

2d1:

Issue #2091, opened 2015-07-08
Add www.bplaced.net


blobgo:
Please add a ruleset for https://www.bplaced.net. It's a hosting provider.

Please note that (at least) https://forum.bplaced.net/ must be excluded as this displays a completely different site in HTTPS.

PR #2095, opened 2015-07-09
[bplaced.net] New ruleset

donced:
Fix https://github.com/EFForg/https-everywhere/issues/2091

### DIFF
--- a/src/chrome/content/rules/bplaced.xml
+++ b/src/chrome/content/rules/bplaced.xml
@@ -1,0 +1,23 @@
+<!--
+
+	Non-functional subdomain:
+		- forum
+
+-->
+
+<ruleset name="bplaced">
+
+	<target host=    "bplaced.net" />
+	<target host="www.bplaced.net" />
+
+  	<securecookie host="^www\.bplaced\.net$" name=".*" />
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/bplaced.xml
```diff
@@ -0,0 +1,23 @@
...
+
+-->
+
+<ruleset name="bplaced">
+
+	<target host=    "bplaced.net" />
+	<target host="www.bplaced.net" />
+
+  	<securecookie host="^www\.bplaced\.net$" name=".*" />
+
```
2d1: Why redirect to www? ^bplaced.net seems to work fine. If there's no problem, please don't preempt the server's redirect. Doing so only gives an adversary on the wire greater confidence that the user is using https-everywhere.

src/chrome/content/rules/bplaced.xml
```diff
@@ -0,0 +1,23 @@
...
+	<rule from="^http://bplaced\.net/"
+		to="https://www.bplaced.net/" />
+
+		<test url="http://www.bplaced.net/" />
```
2d1: This test is redundant, as literal targets generate implicit tests.

2d1:

Issue #1327, opened 2015-04-01
Proposal: Consider removing https -> https/http rewrite rules?


semenko:
We have a small subset of rules that rewrite HTTPS sites to other HTTPS/HTTP sites. A rough grepping shows ~63 rules like this.

```
$ grep -F "from=\"^https" default.rulesets  | wc -l
63
```

I propose removing these rules, and the code needed to handle them.

Benefits:
- HTTPSe still uses nonzero resources & can slow down browsing on underpowered systems (e.g. Chromebooks). We can remove blocking all HTTPS requests in Chrome, which would be a great performance benefit.
- Reduce ruleset fragility -- rewriting broken pages to fix incorrect cert names seems really fragile
- Reduce code complexity -- We can remove some of the downgrade="1" rules and eliminate the downgrade whitelist.
- Reduce rules that probably decrease security -- I imagine some of these downgrades are ancient & aren't needed anymore (e.g. the GoogleMaps whitelist that downgrades https://maps.googleapis.com/map ?) -- users also probably don't expect us to send https -> http

Downsides:
- We may increase breakage on sites that aren't very HTTPS friendly, but many are broken for other reasons (mixedcontent). e.g. https://fellowsblog.kiva.org

Thoughts?

semenko:
FWIW, here's a rough list of files with downgrade/https matching rules.

Aeriagames.xml
Apple.xml
Austin_Fit_Magazine.com.xml
CERN.xml
CafePress.xml
Caltech.xml
Cashback.co.uk.xml
Cheezburger.xml
Comic-Con-Intl.xml
Contextly.xml
Cru.org.xml
Dafont.com.xml
DailyDot.xml
EFF.xml
Elder_Scrolls_Online.com.xml
Epic-Systems.xml
EvoTronix.xml
Fasthosts.xml
Fourmilab.xml
GO.com.xml
Ghostery_metrics_optout.xml
GoogleMaps.xml
ImageShack.us.xml
InkFrog.com.xml
Intel.xml
Intelliworks_Chat.com.xml
Jabber.ru.xml
Kat_de_Paris.xml
Kiva.org.xml
Lenovo.xml
Levo_League.com-problematic.xml
Link-Plus-Catalog.xml
Linuxaria.com.xml
Marketwatch.com.xml
Minijob-zentrale.de.xml
MobyGames.xml
MoveOn.xml
MovieTickets.com.xml
Mutterschaftsgeld.de.xml
NYTimes.xml
[...]

PR #2104, opened 2015-07-10
Cleanups & prep for reducing HTTPS listeners

semenko:
For issue #1327 

Start to split out HTTPS listening from HTTP in Chrome, and some other minor tweaks.


jsha:
Thanks!

PR #2105, opened 2015-07-10
Switch to new Travis infra?

semenko:
Travis is pushing a Docker-based infra:
http://docs.travis-ci.com/user/migrating-from-legacy/

This PR will probably break ... let's see!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
   - DISPLAY=':99.0'
 script:
   - ./test.sh
+sudo: false
 notifications:
   email:
     recipients:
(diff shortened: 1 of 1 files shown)


semenko:
So the build passed, but threw one error:
```
Disallowing packages: libxslt-dev
If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process
```
Do we need libxslt-dev?

And TODO: add caching flags, per http://docs.travis-ci.com/user/caching/

jsha:
This is terrific! I was just looking at the new infra stuff this afternoon and thinking of porting HTTPS Everywhere's build. Thank you for doing it.

It appears we don't need libxslt-dev, since the tests pass. Merging.

PR #2113, opened 2015-07-10
Create rules for  leboncoin.fr

LoHub:


src/chrome/content/rules/leboncoin.fr.xml
```diff
@@ -0,0 +1,21 @@
+<ruleset name="leboncoin.fr (Partial)">
```
youdly: why it's partial? any non-functional subdomain? Please list them as ruleset comment.

semenko:
Awesome -- thanks!

Issue #1771, opened 2015-05-25
Settings window freezes Firefox for a long time / sometimes crashes it


alex42ka:
As described above. I'm using https-everywhere 5.0.4 within Firefox 38 on Ubuntu 14.04.

reedy:
Can you try with a new ff profile to see if that helps?

reedy:
Also, where did you build of FF 38 for 14.04 come from?

http://packages.ubuntu.com/trusty/firefox and http://packages.ubuntu.com/trusty-updates/firefox only offer 37.0.2

alex42ka:
The package came via update:

https://launchpad.net/ubuntu/+source/firefox/38.0+build3-0ubuntu0.14.04.1

I will try to add a new profile and to test if the bug appears there as well.

alex42ka:
Tried it with a new clean profile. Firefox didn't crash but produced high CPU-load and a lot of memory was in use. Then I tried my old profile again and realized that Firefox there as well wasn't crashing but in a freeze for more then a minute. The CPU load lasted even longer on 100% and up to 900MiB memory were in use. After more then a minute the settings window appeared but in the list there were no visible entries. I clicked "OK" and reopened the settings windows. This time the window appeared directly and the list were visible.
After closing and reopening the Firefox window again the first attempt to open the settings window of https-everywhere produced high load and made Firefox freeze and on the second attempt the window appeared directly.

reedy:
See also #1775

pde:
I can reproduce this in FF 39 / HTTPSE 5.0.5, but may not have time to work on it till late in the week.  May be able to provide some data to others, though...

jsha:
This happens because the settings window includes a list of all rulesets packages with HTTPS Everywhere. To show that list, HTTPS Everywhere loads all rulesets into memory at once (normally it loads rulesets on an as-needed basis). That's a lot of memory, and it's also a lot of CPU time to parse all that XML.

My preferred approach is to remove the list of all rulesets, since it is rarely used. It's not that useful in a world of 15k rulesets, either. However, it would also be possible to show an empty list, wait for the user to type a query, and only load the rulesets matching that query.

Issue #1683, opened 2015-05-12
Decide on future of HTTPS Everywhere Preferences window


reedy:
The "Enable / Disable Rules" aka the HTTPS Everywhere Preferences window should be added to Chrome, along with the dependant view xml source windows.

Subtask of #1681

jsha:
BTW, I would actually be excited about *removing* this window from Firefox to reach parity. I think it is used very seldom, and the dropdown checkboxes serve 95% of the use cases it serves. Interested to hear your thoughts on the matter though.

jsha:
The other things we have in that window are `Reset to defaults` (could be moved to dropdown) and `Disable all` (superseded by `Disable HTTPS Everywhere` in the menu, should probably be removed.

reedy:
I'd be up for that, yeah. It has some use for a layman to see "what's supported", I think, Some way of getting from from the ruleset (a hyperlink) to viewing the ruleset in a web browser tab (#1436 or similar) would be useful.

PR #2121, opened 2015-07-10
Remove Enable / Disable rules window.

jsha:
Loading all the rulesets at once for the Enable / Disable rules window
causes browser lockup and sometimes crashes. All of the functionality
in the window is now available in other places: Rules can be enabled
and disabled directly from the drop down menu. "Disable all" is
available as the menu item "Disable HTTPS Everywhere." And viewing
the source of rulesets is available through the HTTPS Everywhere Atlas.

The one missing piece of functionality was "Reset rulesets to defaults," which I
moved to a menu item.

Now, the preferences link in Firefox's Add-On pane will take you to the SSL
Observatory preferences.

Fixes https://github.com/EFForg/https-everywhere/issues/1771
Fixes https://github.com/EFForg/https-everywhere/issues/1683


src/chrome/content/toolbar_button.xul
```diff
@@ -43,6 +43,8 @@
       <menupopup id="https-everywhere-context" onpopupshowing="show_applicable_list(this)">
         <!-- entries will be written here by ApplicableList.populate_menu() -->
         <menuseparator class="hide-on-disable"/>
+        <menuitem label="&https-everywhere.prefs.reset_defaults;"
```
pde: Logically, it seems this option should be in the same menu subsection as the ruleset togglers...

pde:
Overall this looks okay, though if there was an easy way to visually indicate that "Enable / Disable rules" is now a heading, and not an option of its own, that would be desirable.

src/chrome/content/toolbar_button.xul
jsha: You spotted my corner cutting. The problem is, the string for "Reset to defaults" is in the .dtd locale file (accessible only from XUL), but the menu subsection with the ruleset togglers is built dynamically in Javascript. I considered porting over the "Reset to defaults" string into the .properties locale file so it can be accessed from Javascript, but it didn't seem worth the time.

jsha:
I'm going to go ahead and merge this for the upcoming release, and possible nice-ify the layout for a subsequent release.

alice-margatroid:
Nothing but inane changes because old hardware cannot handle it.

Why not implement an "Expert mode" for people capable of making such decisions to load a "slow" but meaningful preferences window?

berrythesoftwarecodeprogrammar:
i agree. seems like clicking the entries in the dropdown menu doesnt even work most of the time anymore

margentieri:
This is insanely stupid. Completely removing this functionality has damaged the usability of this addon, and has FORCED me to downgrade to 3.4.5 to retain the preferences window.  You are ignoring the fact that not all sites that are being converted to an HTTPS connection are actually that of the website being visited, and therefore do not show up as options to check or uncheck on the site.  A prime example is Netflix, which uses Silverlight to play their videos and provide DRM.  Silverlight needs to communicate with Microsoft, but none of Microsoft's websites are listed on the dropdown when you are on Netflix's site.  Without disabling HTTPS Everywhere for Microsoft, Silverlight fails to properly communicate with Microsoft, and the site will not play ANY videos.  In the past, I simply had to go into the addon's settings, search for Microsoft, and uncheck it.  Now I can't do that, which completely breaks Netflix.  I should note that I am using Firefox on Ubuntu Linux, not Windows or Mac.  Functionality in an addon like this should never be totally removed for the sake of convenience or the laymen.  If you want to remove it from sight for the laymen, fine, but like alice-margatroid suggested, at least provide a means to re-enable it though an expert mode check box or an about:config setting that can be changed.

Issue #2006, opened 2015-06-26
canon.ca


joshuaauger:
canon.ca support and drivers page uses an iframe with the custhelp.com domain. When using HTTPS Everywhere, this page will not load and will not give an error.

http://www.canon.ca/inetCA/en/inFrame?m=showInFrame&url=http://canoncanada.custhelp.com/ the important part being http://canoncanada.custhelp.com/

semenko:
Only half closed

jeremyn:
@joshuaauger If I go to http://www.canon.ca/inetCA/en/inFrame?m=showInFrame&url=http://canoncanada.custhelp.com/ and click the "Drivers, Software & Manuals" image, I'm brought to a page where I can enter a model name. This looks working to me. Is this still an active problem for you?

joshuaauger:
Seems fine now.

jeremyn:
@joshuaauger Thanks for the quick reply. Closing issue.

PR #2124, opened 2015-07-11
[Canon] Fixed problem with www.usa.canon.com

LoHub:


donced:
Interested in fixing #2006?

donced:
Improved by https://github.com/EFForg/https-everywhere/pull/2452.

jsha:
The relevant issue is fixed in #2452, but this change adds some more domains. Can you pull in the latest master and change the rules to be simple rules (i.e. `<rule from='^http:' to='https:' />`)?

J0WI:
Fixed in #2452 and #2725

PR #2126, opened 2015-07-11
Improvements to Deviantart.xml

BoffinbraiN:
2nd rule would previously fail for deviantart.net because it changed the TLD to into deviantart.com. Added 2nd capture group like the first rule.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,20 @@
 # Config copied from
 #  http://www.theautomatedtester.co.uk/blog/2012/using-travis-ci-for-building-and-testing-firefox-addons.html
-# Use node_js because there are maybe more workers?
 language: python
 python:
   - "2.7"
+cache: pip
 addons:
+  apt:
+    packages:
+      - libxml2-dev
+      - python-dev
+      - libcurl4-openssl-dev
+      - python-lxml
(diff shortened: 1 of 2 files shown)


2d1:
The comment above the rule you changed is still true for at least (www.)?deviantart.net.

Please either create a third rule for names for which the comment is still true, or add the names that you've observed as being broken to the first rule.

BoffinbraiN:
Good spot. deviantart.net now has its own separate certificate, so I'm going to condense both rules into a single rule and test for a few days before proposing it.

semenko:
Looks like this PR has conflicts -- and it hasn't been updated in a while. I'm going to close this for now, but if you'd like it merged, feel free to re-open. Thanks! :)

BoffinbraiN:
I'm not exactly sure what's going on with this thread any more. My original changes still work fine on my local modified version of the extension. If I were to make another PR, where should I fork from?

semenko:
Cool -- I think another user edited Deviantart in the meantime, which broke your PR:
https://github.com/EFForg/https-everywhere/commit/d915ffc02ffa0f66e4bd4379b3bd4c23073c0d40

So you'd need to rebase/pull from master, to make sure you have those changes.

PR #2128, opened 2015-07-11
Create administrator.de.xml

fuu0:
ruleset for http://www.administrator.de with test URLs added

### DIFF
--- a/src/chrome/content/rules/administrator.de.xml
+++ b/src/chrome/content/rules/administrator.de.xml
@@ -1,8 +1,6 @@
 <ruleset name="administrator.de">
 <target host="*.administrator.de" />
-<rule from="^http:" to="https:" />
+<rule from="^http://administrator\.de/"
+    to="https://www.administrator.de/" />
 <test url="http://www.administrator.de/" />
-<test url="http://www.administrator.de/fragen/" />
-<test url="http://www.administrator.de/wissen/kabel-deutschland-bridge-mode-echten-router-verwenden-darf-incl-ipv6-201095.html" />
-<test url="https://www.administrator.de/members/login.php" />
 </ruleset>
--- a/src/chrome/content/rules/administrator.de.xml
+++ b/src/chrome/content/rules/administrator.de.xml
@@ -1,0 +1,8 @@
+<ruleset name="administrator.de">
+<target host="*.administrator.de" />
+<rule from="^http:" to="https:" />
+<test url="http://www.administrator.de/" />
+<test url="http://www.administrator.de/fragen/" />
+<test url="http://www.administrator.de/wissen/kabel-deutschland-bridge-mode-echten-router-verwenden-darf-incl-ipv6-201095.html" />
+<test url="https://www.administrator.de/members/login.php" />
+</ruleset>


src/chrome/content/rules/administrator.de.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="administrator.de">
+<target host="*.administrator.de" />
```
donced: can you confirm there is any subdomain (aside from www.)? I don't see any in my testing.

src/chrome/content/rules/administrator.de.xml
```diff
@@ -0,0 +1,8 @@
...
+<rule from="^http:" to="https:" />
+<test url="http://www.administrator.de/" />
+<test url="http://www.administrator.de/fragen/" />
+<test url="http://www.administrator.de/wissen/kabel-deutschland-bridge-mode-echten-router-verwenden-darf-incl-ipv6-201095.html" />
+<test url="https://www.administrator.de/members/login.php" />
```
donced: Line 5-7 is not required and caused build fail.

src/chrome/content/rules/administrator.de.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="administrator.de">
+<target host="*.administrator.de" />
+<rule from="^http:" to="https:" />
+<test url="http://www.administrator.de/" />
```
donced: Suggest to add:
```
<rule from="^http://administrator\.de/"
	to="https://www.administrator.de/" />
```
To illustrate, without above lines, the redirection will be:
`http://administrator.de -> http://www.administrator.de -> https://www.administrator.de`
with above rules:
`http://administrator.de -> https://www.administrator.de`
In the end, you still get https://www.administrator.de though, so it's not strictly essential.
If you decide to take my suggestion, keep line 4, as the above lines require it.

src/chrome/content/rules/administrator.de.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="administrator.de">
+<target host="*.administrator.de" />
+<rule from="^http:" to="https:" />
```
donced: usually we put this at last line (before `</ruleset>`).

donced:
Anyway, please read the [ruleset style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

src/chrome/content/rules/administrator.de.xml
fuu0: No, that was only for future, if there come new URLs.

src/chrome/content/rules/administrator.de.xml
fuu0: okay, i thougt that i need more URLs for testing

src/chrome/content/rules/administrator.de.xml
fuu0: But if someone goes to http://www.administrator.de he isnt redirectet, or i am wrong?

donced: My apologies, I have mislead you. Just simplify to:
```
<ruleset name="administrator.de">
<target host="*.administrator.de" />
</ruleset>
```
But I personally still prefer to list all subdomains instead of `*.administrator.de`. I understand your argument that * can cover future subdomain. But in my experience, it often causes mixed content blocking (MCB) issue as some subdomains don't serve HTTPS. We never know whether new subdomain can support HTTPS. By listing all subdomains, we know that those subdomains *work*.

src/chrome/content/rules/administrator.de.xml
donced: If website doesn't use any subdomain (except www), most probably they won't in future.

jsha: I'm fine with using a rule that rewrites all subdomains when there is only www. It has the big advantage that it makes it impossible for a MITM to make you load nonexistent.example.com over HTTP and potentially leak cookies if they are scoped to *.example.com.

semenko:
Looks like our tests are failing on this PR -- and it hasn't been updated in a while. I'm going to close this for now, but if you'd like it merged, feel free to re-open. Thanks! :)

PR #2132, opened 2015-07-12
[deutsche-rentenversicherung.de & governikus.com] New ruleset

donced:
Improves previous PR https://github.com/EFForg/https-everywhere/pull/883 https://github.com/EFForg/https-everywhere/pull/863

### DIFF
--- a/src/chrome/content/rules/Deutsche-Rentenversicherung.de.xml
+++ b/src/chrome/content/rules/Deutsche-Rentenversicherung.de.xml
@@ -1,0 +1,24 @@
+<ruleset name="Deutsche-Rentenversicherung.de">
+
+	<target host=       "deutsche-rentenversicherung.de" />
+	<target host=   "www.deutsche-rentenversicherung.de" />
+	<target host="static.deutsche-rentenversicherung-bund.de" />
+	<target host=   "www.deutsche-rentenversicherung-bund.de" />
+	<target host=       "deutsche-rentenversicherung-bund.de" />
+	<target host="www.drv-bund.de" />
+	<target host=    "drv-bund.de" />
+
+  	<securecookie host="^www\.deutsche-rentenversicherung\.de$" name=".+" />
+
+	<rule from="^http://(www\.)?deutsche-rentenversicherung(-bund)?\.de/"
+		to="https://www.deutsche-rentenversicherung.de/" />
+
+		<test url="http://www.deutsche-rentenversicherung.de/" />
+
+	<rule from="^http://(www\.)?drv-bund\.de/"
+		to="https://www.deutsche-rentenversicherung.de/" />
+
+	<rule from="^http:"
+		to="https:" />
+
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Governikus.com.xml
```diff
@@ -0,0 +1,25 @@
...
+	<target host=    "bos-bremen.de" />
+
+	<securecookie host="^www\.governikus\.com$" name=".+" />
+
+	<!-- cert only matches www -->
+	<rule from="^http://governikus\.com/"
+		to="https://www.governikus.com/" />
+	
+	<!-- cert only matches www -->	
+	<rule from="^http://?bos-bremen\.de/"
```
jsha: I think you have a stray `?` here.

src/chrome/content/rules/Deutsche-Rentenversicherung.de.xml
```diff
@@ -0,0 +1,26 @@
...
+	<target host="static.deutsche-rentenversicherung-bund.de" />
+	<target host=   "www.deutsche-rentenversicherung-bund.de" />
+	<target host=       "deutsche-rentenversicherung-bund.de" />
+	<target host="www.drv-bund.de" />
+	<target host=    "drv-bund.de" />
+
+  	<securecookie host="^www\.deutsche-rentenversicherung\.de$" name=".+" />
+
+	<!-- could not resolve host -->
+	<rule from="^http://deutsche-rentenversicherung(-bund)?\.de/"
```
jsha: When rewriting to add the www, it's a little nicer to keep the rest of the URL the same when possible, since it's easier to trivially validate that the redirect is correct.

jsha:
Copying @reedy's comment from #863:

>  It looks like from a bit of testing that http://www.governikus.de and http://www.bos-bremen.de always gets redirected to https://www.governikus.com (at least in my browser). With a /de suffix, but that's the same for either site, so I'm not sure if it's worth adding that too.
> Either way, I think we should just move any of those straight over to the HTTP version...

@reedy: I'm not sure I follow what you're saying. Is it that was should downgrade certain URLs to HTTP? That seems incorrect.

src/chrome/content/rules/Deutsche-Rentenversicherung.de.xml
donced: http://www.deutsche-rentenversicherung-bund.de http://deutsche-rentenversicherung-bund.de https://www.deutsche-rentenversicherung-bund.de are redirected (by server) to http://www.deutsche-rentenversicherung.de just fine, except for https://deutsche-rentenversicherung.de/ & https://deutsche-rentenversicherung-bund.de/ so line 14 is necessary.

semenko:
Looks like this has some outstanding conflicts, and hasn't been touched in a while. Feel free to reopen if you'd like this PR merged. Thanks!

PR #2136, opened 2015-07-12
Update ALDI.xml

J0WI:
- fixed hofer.si


2d1:
It would nice if you added a comment stating that the cert doesn't match ^hofer.si, for the sake of webmasters reading the Atlas.

PR #2137, opened 2015-07-12
Update Coop.ch.xml

J0WI:
false positive

### DIFF
--- a/src/chrome/content/rules/Coop.ch.xml
+++ b/src/chrome/content/rules/Coop.ch.xml
@@ -1,12 +1,4 @@
-
-<!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


2d1:
This seems to have hit a case that isn't accounted for in the fetch tester. Namely 200 followed by redirect to 404 which occurs over both http and https. If that is the case, this ruleset will have to be reenabled after each fetch-test triggered deluge.

jsha:
Are you sure this is a false positive? If I load the HTTP version of http://www.cooperation-online.ch/, I get redirected to http://www.cooperation.ch/10288, which works. If I load the HTTPS version of https://www.cooperation-online.ch/, I get redirected to https://www.cooperation.ch/startseite, a 404 page. I think the checker actually did the right thing here, and the cooperation-online domains should be removed or rewritten to a different domain.

PR #2138, opened 2015-07-12
Update migros.xml

J0WI:
- fixed errors (Migros implements http redirects)
- cleanup redirect rules


2d1:
It would be nice if you'd add comments about m-budget and twistymania for the sake of webmasters reading the atlas.

PR #2142, opened 2015-07-12
Update swissgov.xml

J0WI:
fixes for regressions in gov infrastructure


2d1:
It would be nice if you added comments about the problematic names your commit works around, for the sake of webmasters reading the Atlas.

PR #2145, opened 2015-07-12
Delete Lolicore.ch.xml

J0WI:
remove ruleset because
1) it's outdated
2) it's disabled
3) domain supports preloaded HSTS https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsSTSPreloadList.inc#1037


J0WI:
I'll release a patch for all the HSTS dupes soon.

PR #2146, opened 2015-07-12
Update upc-cablecom.xml

J0WI:
fixed some server issues


2d1:
It would be nice if you added comments stating that ^upc-cablecom.(biz|ch) aren't listening for tls connections, for the sake of webmasters reading the Atlas.

PR #2148, opened 2015-07-13
[Verisign] fixed various errors and reenable ruleset

J0WI:


J0WI:
I think the strange error on mygatekeeper.verisign.com.au is a chain issue, but I'm not sure. I disabled it anyway to pass the tests.

Issue #2134, opened 2015-07-12
Streamfarm.net rule breaks jumpradio.de


finaljason:
The Streamfarm.net rule seems to break live radio streaming on jumpradio.de
The underlying streamfarm.net server (http://c22033-ls.i.core.cdn.streamfarm.net) times out on HTTPS requests and should be excluded from rewrite.

jsha:
Fixed by #2150.

PR #2149, opened 2015-07-13
[YTS.to] Revert "add new domain" (fixing the build!)

mnordhoff:
The domain, `s.ynet.io`, is a duplicate: someone else concurrently added it in `Ynet.io.xml` in fafaa2838c361a2e7a698ead200a14e3398286f6.

This reverts commit f0722e55a4f27f07fcb4585d6cdee09bf2c0970d.

PR #2150, opened 2015-07-13
Fix jumpradio.de livestream

finaljason:
See issue #2134


finaljason:
Note that the Travis CI build fails because of commit f0722e5. The test coverage is fine and when #2149 gets merged this should be not a problem anymore.

Issue #2151, opened 2015-07-13
boingboing.net no longer using cachefly


orenwolf:
boingboing.net is in the process of implementing HTTPS. As part of this change, we've changed CDNs on our end.

It looks to me that at some point in the past, someone decided they could jury-rig boingboing to be active by piggybacking on our old CDNs existing HTTPS certificate. they did this by rewriting requests to our CDN to instead use their domain.

Imagine their (and our!) surprise when suddenly after changing providers, users of HTTPS everywhere suddenly had a broken BB experience! I'm not sure where that ruleset came from, or the idea for this hack, but unfortunately everyone using the current ruleset is now broken.

For now, boingboing (mostly) works via HTTPS. Over the next week or so we'll correct the rest of the mixed-content errors (most of which are the result of third parties), however the media.boingboing.net domain *is* ready to go with HTTPS.

*so*, my recommendation is to re-enable the ruleset, but without the domain remapping for media. It should simply rewrite any http: media requests to https (which is fine, and should work without issue if they visit the http version of boingboing.net).

In the next few weeks, I'll open a new issue to enable the site fully, once the rest of the site has cleaned out mixed content errors and is ready to go. :)

Thanks!

-- 
Ken Snider
Sysadmin, Boing Boing

jsha:
Hi Ken! Thanks for the ping. We've got the ruleset disabled in the current master branch, which I'm going to release today. For an upcoming release, I'll re-enable the rule with a rewrite from http to https on the media domain. I'm looking forward to full HTTPS support on Boing Boing!

jsha:
Fixed in latest release.

orenwolf:
Thanks, Jacob!

PR #2160, opened 2015-07-13
#2151

numismatika:
For #2151
- remove Cachefly from Boing-Boing


jsha:
Note: This is still default_off, but that's fine. It sounds like Boing Boing is in the middle of an HTTPS transition, so I'm happier leaving this disabled until things stabilize a little bit.

PR #2168, opened 2015-07-13
readd www.sbb.ch

J0WI:


J0WI:
@jsha would be great to have this rule back in next release

Issue #2098, opened 2015-07-09
http://www.antenna.gr/webtv/live broken


fateswarm:
http://www.antenna.gr/webtv/live does not work with it on. (It normally plays a TV channel live).

(it's a legal site, the official site of the channel)

stullig:
Bit hard to reproduce cause region locked to greece

fateswarm:
Well, ask a few others from that region to confirm

reedy:
There's at minimum, loads of mixed content...

J0WI:
Can you check if #2175 solves the problem?
You may report the problem on https://github.com/kargig/https-everywhere-greek-rules as offered in the code comment.

semenko:
Closing, per rule update above.

PR #2175, opened 2015-07-14
Update Greek-media.xml

J0WI:


jsha:
It looks like this rule formerly rewrote all first-level subdomains of antenna.gr, but now only rewrites antenna.gr and www.antenna.gr. Can you add detail about why? I'm also not sure why this would fix #2098.

J0WI:
I removed the wild card to avoid writing test urls. I didn't found any subdomains exept www. I also added ```static.ant1.gr``` which fixes mixed content issues and may be #2098.

jsha:
Shouldn't this be `platform="mixedcontent"` if it requires the `static.ant1.gr` rewrite to work properly?

semenko:
Want to simply remove antenna.gr from this? They have some cert name issues.

J0WI:
Oh, thats new.
I removed the host.

PR #2178, opened 2015-07-14
[Calameo] Fix #1880

### LINKED ISSUE #1880: calameo.com (partial) rule breaks fr.calameo.com
Here's a file that cannot be opened when the calameo.com (partial) rule is on (default) : http://fr.calameo.com/read/004260307690d779c9a4e
it opens all right when rule is off.

J0WI:


jsha:
Looks good. For future, please put "Fixes #NNN" in the pull request body, so GitHub makes it linkable.

PR #2194, opened 2015-07-14
Update NYDailyNews.com-falsemixed.xml

J0WI:
false positive


jsha:
Thanks. Updated the checker to include the needed intermediate here.

PR #2196, opened 2015-07-14
Update Host2.bg.xml

J0WI:
false positive


jsha:
Thanks. This intermediate was already added recently.

PR #2197, opened 2015-07-14
Update Olvi.fi.xml

J0WI:
false positive and domains added


jsha:
Thanks! Updated the checker with the missing intermediate.

PR #2199, opened 2015-07-14
Update WWE.com.xml

J0WI:
false positive


jsha:
Added the intermediate to the checker.

PR #2200, opened 2015-07-14
Update Upplandsvasby.se.xml

J0WI:
false positive


jsha:
Thanks! Added the intermediate to the checker.

PR #2203, opened 2015-07-14
Update SLF.se.xml

J0WI:
false positive, removed platform, cleanup


src/chrome/content/rules/SLF.se.xml
```diff
@@ -1,11 +1,9 @@
...
 	<target host="slf.se" />
 	<target host="www.slf.se" />
-	<rule from="^http://slf\.se/" to="https://www.slf.se/"/>
-	<rule from="^http://www\.slf\.se/" to="https://www.slf.se/"/>
-</ruleset>
 
+	<rule from="^http://slf\.se/"
+		to="https://www.slf.se/"/>
+	<rule from="^http:"
+		to="https://www.slf.se/"/>
```
jsha: This line is incorrect; should be just `https`.

J0WI:
(done)

Issue #2220, opened 2015-07-14
Chrome build likely broken


semenko:
FYI, @jsha -- it looks like https://github.com/EFForg/https-everywhere/commit/08a1f4acae648be8b76bdb74a8e720b193e20797 broke the Chrome build because of the unterminated:
```<securecookie host="^\.paste2\.org$" name="^(?:__cfduid|__qca$" />```

This throws:
```Uncaught SyntaxError: Invalid regular expression: /^(?:__cfduid|__qca$/: Unterminated group```

I'm guessing the version submitted to the Chrome store (and stuck) is broken (!)

* We should fail more gracefully in Chrome if a rule is malformed.
* We should expand tests to catch that.

jsha:
Good catch, thanks! Fixed and test added in #2222.

PR #2222, opened 2015-07-14
[Paste2.org] Fix bad regexp.

jsha:
Also, expand trivial-validate to catch this case.
Fixes https://github.com/EFForg/https-everywhere/issues/2220


semenko:
Cool! LGTM.

Issue #2223, opened 2015-07-14
Chrome: Recover from failures when reading rulesets


jsha:
In #2220, a bad regexp made it into the released package. This would have caused rewrites to fail for all users, but fortunately the release got hung up in the Chrome Web Store, probably by automated validation.

We should put try/catches around the relevant blocks, so a problem loading a single ruleset does not break all rulesets.

nallar:
> but fortunately the release got hung up in the Chrome Web Store

Actually, it did get released :(

Getting `Uncaught SyntaxError: Invalid regular expression: /^(?:__cfduid|__qca$/: Unterminated group` on current version from the Chrome Web Store.

PR #2229, opened 2015-07-15
Add a try/catch around ruleset processing.

jsha:
This will help mitigate future bugs like
https://github.com/EFForg/https-everywhere/issues/2220, so an error processing a
single ruleset doesn't break the whole extension. cc @semenko for review.


semenko:
LGTM, closes #2223 -- manually tested that it catches bug #2220

Issue #2227, opened 2015-07-15
Chrome: Cannot create / modify rules


smarek:
In latest version of extension, I'm unable to enable/disable applied rules or create custom rules, the dialog contains only "block all http request" item, see screenshot. Is this expected behavior?

Using Chrome Canary 46.0.2455.0

![screen shot 2015-07-15 at 16 51 41](https://cloud.githubusercontent.com/assets/384824/8701406/d99dd22e-2b11-11e5-897f-21adc99c8582.png)

nallar:
Caused by #2220.

It's not just not displaying them, no rules are being applied at all. The broken build was released for some reason.

smarek:
Well, can I request comment from devs please? @semenko @jsha

jsha:
@smarek: Apologies for the broken build. The fixed version (2015.7.15) should be available now. Has your browser auto-updated?

smarek:
@jsha well, thank you wery much, i didn't notice that. now rules appear to be workin, "block all http requests" too, however not "add rule for this site", that one will show the buttons, but all three actions do nothing ("Show advanced", "Add new rule", "Cancel"), collected error is on screenshot, only error shown for this version:
![screen shot 2015-07-16 at 01 02 33](https://cloud.githubusercontent.com/assets/384824/8711898/732d8c1a-2b56-11e5-994a-f351b1f4041c.png)

semenko:
Hm, yeah, I see it on some sites but not others. Missing on nytimes.com, but present on github.com.

jsha:
Thanks for point this out, @smarek! I fixed it in #2234, which will be included in the next release.

PR #2234, opened 2015-07-15
Fix user rules in Chromium.

jsha:
An unclosed <p> tag meant that i18n message replacement was eliminating some UI
elements that were expected to be there.

Fixes https://github.com/EFForg/https-everywhere/issues/2227

cc @semenko @smarek for review.

### DIFF
--- a/chromium/popup.html
+++ b/chromium/popup.html
@@ -22,7 +22,7 @@
   <a href="#" id="add-rule-link" i18n="menu_add_rule"></a>
   <div id="add-new-rule-div" style="display:none">
     <h3 i18n="about_add_new_rule"> </h3>
-    <p i18n="menu_always_https_for_host">
+    <p i18n="menu_always_https_for_host"></p>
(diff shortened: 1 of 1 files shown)


semenko:
LGTM.

I also didn't realize this was gated on the site actually being HTTPS in the open tab. Makes sense, though it's a confusing UX to have elements disappear.

````
  // Only show the "Add a rule" link if we're on an HTTPS page
  if (/^https:/.test(tab.url)) {
    show(e("add-rule-link"));
  }
````

jsha:
Yeah, it was a stopgap measure, since we were getting lots of reports that people had added rules for sites with no HTTPS version and effectively broken the site for themselves.

Probably a better UX treatment would be to show the elements unconditionally, but trigger an HTTPS fetch either when opening the "add rule" section or when clicking the button to create the rule, and refusing to create a rule unless there was a successful HTTPS response.

PR #2235, opened 2015-07-16
Update https-everywhere.dtd

Karlingue:
update fr


jsha:
Please join Transifex and use their interface to submit translation fixes: https://www.transifex.com/signup/?join_project=torproject. Thanks!

PR #2245, opened 2015-07-16
New rules (imirhil.fr, chiffrofete.fr, café-vie-privée.fr)

aeris:
Hello,

Here are 3 new rules.

For one (café-vie-privée.fr), the domain use Unicode symbols.
First, I test with punny-code inside rule file, but the rule doesn’t seems match (too simple char to char matching ?)…
Secondly, I test with Unicode directly on the rule file, but the relaxng schema is too restrictive for host check.
I just remove the pattern restriction from host, not sure if it can really be filter with regex since IDN exist (any Unicode symbol is allowed).

Perhaps IDN can/must be manage with the corresponding punny-code ?
Will avoid charset encoding problem on rule files and direct punny-code input by the user…

I will open a standalone ticket for this.

Regards,

### DIFF
--- a/utils/relaxng.xml
+++ b/utils/relaxng.xml
@@ -32,9 +32,7 @@
     <oneOrMore>
       <element name="target">
         <attribute name="host">
-          <data type="string">
-            <param name="pattern">(([A-Za-z0-9äö-]+|\*)\.)*([A-Za-z0-9äö-]+|\*)</param>
-          </data>
+          <data type="string" />
         </attribute>
       </element>
     </oneOrMore>
--- a/src/chrome/content/rules/Cafe-vie-privee.fr.xml
+++ b/src/chrome/content/rules/Cafe-vie-privee.fr.xml
@@ -1,0 +1,8 @@
+<ruleset name="café-vie-privée.fr">
+	<target host="café-vie-privée.fr" />
+	<target host="*.café-vie-privée.fr" />
+	
+	<test url="http://mail.café-vie-privée.fr/" />
+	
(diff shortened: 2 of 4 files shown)


TheNavigat:
@semenko, @aeris mentioned this PR on the IRC. It has been here for too long and no one has checked it yet. Can you please take a look?

TheNavigat:
Turns out the ```utils/relaxng.xml``` edits are made to fix that issue, https://github.com/EFForg/https-everywhere/issues/2246. Sorry for not noticing.

LGTM as well, but again, I just had a quick look, nothing more.

utils/relaxng.xml
```diff
@@ -32,9 +32,7 @@
     <oneOrMore>
       <element name="target">
         <attribute name="host">
-          <data type="string">
-            <param name="pattern">(([A-Za-z0-9äö-]+|\*)\.)*([A-Za-z0-9äö-]+|\*)</param>
-          </data>
+          <data type="string" />
```
jsha: I haven't had a chance to look into HTTPS Everywhere's overall handling of IDNs, but removing the regex check here is definitely too much. If the goal is to include all alphanumeric characters in Unicode, I think you want to use one of the Posix character classes, problably `:alnum:`: http://www.regular-expressions.info/posixbrackets.html. Don't forget to also include `-` which is a valid character in DNS names.

semenko:
Closed since this has some outstanding work and hasn't been touched in a while. Feel free to re-open if you'd like this merged.

PR #2260, opened 2015-07-18
[TWRP.me] New ruleset

youdly:


semenko:
Hm, I can't seem to connect to www.teamw.in -- and are you sure these domains are related?

Not clear that teamw.in -> twrp.me

youdly:
Yes, they are related. twrp.me is the new domain replacing teamw.in (previously these two sites exist side-by-side). This [post](https://plus.google.com/+DeesTroy/posts/5uV24qFJauS) could be a proof.

semenko:
Cool -- thanks! This is a bit nicer than forcing domain redirects (That may disappear in the future, etc.)

PR #2272, opened 2015-07-20
book stores added

J0WI:


semenko:
Hm, there's an annoying HTTP redirect at the root of www.staempfliverlag.com, but I think these work.

PR #2279, opened 2015-07-20
[LibreOffice] Update ruleset

### LINKED ISSUE #2271: ask.libreoffice.com needs exception
Given that LibreOffice has a third-party platform for managing Q&A, it is important to exempt links to ask.libreoffice.org and templates.libreoffice.org, which will be completely unavailable when https:// is prepended to them. For example:

http://ask.libreoffice.org/en/question/52740/i-would-like-to-make-a-letter-template-or-use-one/ 
http://templates.libreoffice.org/

versus:

https://ask.libreoffice.org/en/question/52740/i-would-like-to-make-a-letter-template-or-use-one/
https://templates.libreoffice.org/
(Both of which appear as broken links in Chromium and Firefox on Ubuntu 14+)

I want to encourage novice users to use HTTPS Everywhere as a default option, and I appreciate the fact that rarely does such an inconvenience like this happens. Thanks for the effort to raise the standard of privacy users can expect online.

youdly:
Explicit subdomains which would exclude invalid subdomains.
Fixes #2271

### DIFF
--- a/src/chrome/content/rules/LibreOffice.xml
+++ b/src/chrome/content/rules/LibreOffice.xml
@@ -1,12 +1,13 @@
 <!--
 	Nonfunctional subdomains:
 
-		- ask *
-		- api		(refused)
-		- dev-builds	(refused)
-		- tinderbox *
+		- ask 			¹ 
+		- api			¹ 
+		- dev-builds	¹
+		- templates		(mixed content)
+		- tinderbox		¹
 
-	* Times out
+	¹ Connection refused
 
 
 	Mixed content:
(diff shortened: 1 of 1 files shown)


jsha:
Thanks! For future reference, for a domain like this with many good subdomains and a handful of bad ones, I'd also be fine with a wildcard target host, plus exclusions for the bad subdomains. I don't think this needs to be rewritten though.

yaaqov:
Thanks guys. What's the easiest way to update local plugins to use this commit?

youdly:
"Copy the ruleset source into a custom ruleset in your user dir, make the necessary changes, and disable the old ruleset so your new custom ruleset takes control." https://github.com/EFForg/https-everywhere/issues/2283#issuecomment-122975763
However, my suggestion is to disable the ruleset for now and wait until the next version is rleased. Since libreoffice.org offers HTTPS by default (no HSTS though), it's not really an issue.

PotcFdk:
Why is `extensions.libreoffice.org` included when it fails to load CSS files due to active mixed content getting blocked?

youdly:
https://github.com/EFForg/https-everywhere/pull/2279#issuecomment-124851598
I didn't know? Anyway, thanks for reporting.

PR #2295, opened 2015-07-21
[Hackaday.com] New ruleset

m0jo:


m0jo:
This already exists in wordpress-blogs (not enabled by default). sorry.

PR #2309, opened 2015-07-23
Repair disabled ruleset iapc.utwente.nl

jornane:
The DNS policy at Twente University only allows us to use subdomains under iapc.nl and iapc.utwente.nl. The ruleset has been updated to reflect this policy and is enabled again.

A comment is added that shows the illegal rulesets and an explanation on why they are disabled, so that someone may enable them again when policy changes.


jsha:
Thanks!

PR #2312, opened 2015-07-23
makexpi.sh: Don't copy and delete rules a second time

mnordhoff:
This reduces `make` runtime ~40% on one of my (slow) systems.

(The compare page is showing four other commits from master as part of this branch, though it's not showing them in the diff. I... don't know why.)


jsha:
Nice, thanks! I merged manually and confirmed there were no spurious diffs.

BTW you may also be interested in https://github.com/EFForg/https-everywhere/pull/515. It's mostly useful if you're iterating on code rather than rulesets, but still nice.

PR #2328, opened 2015-07-26
Update rule for Scriptfodder.com

PotcFdk:
Re-enable the ruleset and handle `cdn.scriptfodder.com/` with special care, to prevent the 403 error, that caused https-everywhere-checker to disable the ruleset.
Provided an alternative test url, too.


PotcFdk:
Rebased onto master to fix the CI check.

PR #2347, opened 2015-07-29
[algonquincollege.com] New ruleset

Manouchehri:


semenko:
Some mixed content on the landing, but seems to work. +1

Manouchehri:
Yeah, it's not perfect. Everything important seems to work though.

Issue #2316, opened 2015-07-24
HTTPS Everywhere logo blurry on retina Macbook Pro


adikamdar:
<img width="101" alt="screenshot 2015-07-24 00 49 06" src="https://cloud.githubusercontent.com/assets/1112044/8869870/ebfea5d2-319d-11e5-8467-49d5bc1e5192.png">

reedy:
I agree. It looks like we need a high definition logo, and it's the same in FF :)

reedy:
https://github.com/EFForg/https-everywhere/blob/master/chromium/manifest.json#L25-L29

Those would look like they're supposed to dish out better icons, but I've not really much idea

S-ed:
https://github.com/EFForg/https-everywhere/pull/2351

S-ed:
Heh, found this icon later:
https://github.com/EFForg/https-everywhere/blob/master/src/chrome/skin/https-everywhere-128.png

jsha:
Yes, I think the issue is not lack of high-res logos, but failure to choose the proper high-res logo. Or, perhaps, the browser is choosing the right logo but then needs to resize it for some reason, leading to blurriness.

jsha:
I think I found the cause! On Chromium we set the icon manually based on whether `Block all HTTP requests` is enabled: https://github.com/EFForg/https-everywhere/blob/master/chromium/background.js#L86, and the code doesn't take into account resolution.

Not sure the best way to fix, but thought I'd post what I found so far.

jsha:
@reedy: Can you provide a screenshot in Firefox on a retina Mac Pro? Preferably including neighboring icons for reference?

reedy:
<img width="273" alt="screen shot 2015-08-13 at 09 46 58" src="https://cloud.githubusercontent.com/assets/67615/9245980/4e52f9c4-41a0-11e5-966d-fd4221e073c9.png">

jsha:
@reedy: Looks like you've got 16px icons, same as I do on a regular-resolution PC running Linux. I get the same blurriness on the 16px icon, which I think is just because the icon doesn't shrink well. I've talked a little to our in-house graphic designer about a completely new icon for HTTPS Everywhere that would be more legible / less blurry.

In the meantime, we should at least be able to do better for Retina displays.

gloomy-ghost:
close as fixed

PR #2351, opened 2015-07-31
Vector logo for HTTPS Everywhere

S-ed:
Made in Adobe Illustrator CC using 'Glow' so can't be exported to SVG, sorry.
Also PNG w/o background in 1844x1627 px (should be enough for icon)


S-ed:
Made solely for icons requested in
https://github.com/EFForg/https-everywhere/issues/2316

jsha:
Thanks for working on this! I'm going to close out this branch since I think the solution will be a matter of choosing the correct one of the icons we have, rather than getting a higher-res icon.

PR #2364, opened 2015-08-03
Create sberbank.xml

J0WI:


J0WI:
URL with port seems not no be allowed in tests

J0WI:
@jsha is it possible to make rules for a domains running on a different port?

jsha:
I believe it is possible. The target host matching happens purely on hostname though, so you'd want to leave off the port there, and specify what you want in the `<rules>` tags.

J0WI:
updated

J0WI:
Please merge before this conflicts again.

TheNavigat:
Looks good to me. @semenko Ping! :D

PR #2158, opened 2015-07-13
Externals 13.07.2015

numismatika:
This adds a couple of Subdomains to ad- and tracking networks as well as adds rules for new ones.

PR #2388, opened 2015-08-03
Updates to Politiken.dk and its ad service nugg.ad

fuglede:
The nugg.ad ruleset test currently fails due to a lack of tests. I would add those but I am not sure what the intent of the original ruleset is (i.e. what <code>[\w-]+-s\.nuggad\.net/</code> is supposed to match).

### DIFF
--- a/src/chrome/content/rules/Nugg.ad.xml
+++ b/src/chrome/content/rules/Nugg.ad.xml
@@ -1,26 +1,32 @@
 <!--
-	Nonfunctional subdomains:
+	Problematic subdomains:
 
-		- (www.) *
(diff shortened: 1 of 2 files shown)


jsha:
This PR is failing tests. Do you want to update it or subsume it into https://github.com/EFForg/https-everywhere/pull/2158/files?

fuglede:
Updated with tests from #2158. Travis failure appears to be unrelated (tests succeed locally).

jsha:
Looks good. I re-triggered the test and it passed. Looks like we may have some test flakiness on our hands. :-(

PR #2432, opened 2015-08-05
Added rule for dellin.ru

KOLANICH:


src/chrome/content/rules/dellin.ru.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Delovie Linii (transport company)">
+	<target host="dellin.ru" />
+	<target host="*.dellin.ru" />
+	<securecookie host="(.+\.)?dellin\.ru$" name=".+" />
+	<rule from="^http://(.+\.)?dellin\.ru/" to="https://www.dellin.ru/" />
```
reedy: failure: dellin.ru.xml failed test: The 'from' rule contains unescaped period in regular expression.  Try escaping it with a backslash.



I guess it's supposed to be there, but I guess it doesn't like the .+ syntax

KOLANICH:
> I guess it's supposed to be there, but I guess it doesn't like the .+ syntax

Did you mean that ```.dellin.ru``` is also valid domain or you just don't like wildcards?

jsha:
The preferred format for rules like these is `[\w-]+` or `[\w.-]+` (see the ruleset style guide: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md). The 'unescaped period in from' check is mostly to avoid writing 'www.example.com' which unexpectedly matches 'wwwXexample.com'.

KOLANICH:
Dislike this style guide:
 [RFC 3696](https://tools.ietf.org/html/rfc3696) says:
>**Any characters, or combination of bits (as octets), are permitted in DNS names.** However, there is a **preferred form** that is required by most applications. This preferred form has been the only one permitted in the names of **top-level** domains, or TLDs. In general, it is also the only form permitted in **most second-level** names registered in TLDs, although some names that are normally not seen by users obey other rules.
As you see, third level domain names can have any character.

semenko:
Looks like this has outstanding work and hasn't been touched in a while. Feel free to reopen if you'd like this PR merged -- thanks!

PR #2456, opened 2015-08-05
Delete Toolserver.xml

J0WI:
content moved to Wikimedia


reedy:
Most of these hosts should go, but I wonder if we should put the basic redirect into the Wikimedia ruleset

reedy:
Perfect, thanks!

PR #2484, opened 2015-08-08
[City of Portland] Improved cosmetics, added a test URL, secured cook…

galenide:
…ies, and added the (www.)?portlandoregon.gov domain

### DIFF
--- a/src/chrome/content/rules/CityofPortland.xml
+++ b/src/chrome/content/rules/CityofPortland.xml
@@ -1,6 +1,18 @@
 <ruleset name="City of Portland, OR">
-  <target host="www.portlandonline.com"/>
-  <target host="portlandonline.com"/>
+	<target host="www.portlandonline.com"/>
+	<target host="portlandonline.com"/>
+	<target host="www.portlandoregon.gov"/>
+	<target host="portlandoregon.gov"/>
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/CityofPortland.xml
```diff
@@ -1,6 +1,18 @@
...
+	<target host="portlandoregon.gov"/>
 
-  <rule from="^http://(?:www\.)?portlandonline\.com/" to="https://www.portlandonline.com/"/>
-</ruleset>
+	<securecookie host="^www\.portlandonline\.com$"
+			name=".+" />
+	<securecookie host="^www\.portlandoregon\.gov$"
+			name=".+" />
+
+	<test url="https://portlandonline.com/" />
```
jsha: This test URL is unnecessary.

src/chrome/content/rules/CityofPortland.xml
```diff
@@ -1,6 +1,18 @@
...
+
+	<rule from="^(http://(?:www\.)?|https://)portlandonline\.com/"
```
jsha: The preferred form here is:

```
<rule from="^http://portlandonline\.com/"
  to="https://www.portlandonline.com/" />
<rule from="^http:"
  to="https:" />
```

jsha:
BTW, since I haven't said it yet: Thanks for these ruleset submissions!

PR #2497, opened 2015-08-08
[Quantix] Added a new partial ruleset

galenide:


semenko:
Seems to work. FWIW actual urls look like https://www.quantixtickets1.com/php-bin/quantix.php?fn=135

PR #2511, opened 2015-08-10
add mobile.de.xml

J0WI:


semenko:
Huh, ebay has some interesting brands.

PR #2514, opened 2015-08-11
Use a data-id element for rule toggling.

jsha:
Per amo-validator advice, we should avoid setting oncommand attributes using
interpolated JS. Instead this puts one constant oncommand attribute in the XUL,
which pull the necessary rule id from a data attribute.

cc @cooperq for review.


cooperq:
lgtm

PR #2516, opened 2015-08-11
Remove in-browser ruleset tests.

jsha:
These were implemented over two years ago, when Firefox first turned on mixed
content blocking. They're no longer practical to run because they are so slow
and there are too many rulesets. We will probably want to consult this code in
the future to create a similar test suite that can run headless on multiple
Firefoxes in response to requests from the https-everywhere-checker.
cc @cooperq for review


cooperq:
lgtm

PR #2530, opened 2015-08-12
[OptOutPrescreen.com] Added test URLs and note about geolocation rest…

galenide:
…rictions

### DIFF
--- a/src/chrome/content/rules/OptOutPrescreen.com.xml
+++ b/src/chrome/content/rules/OptOutPrescreen.com.xml
@@ -9,10 +9,8 @@
 	<securecookie host="^www\.optoutprescreen\.com$"
 			name=".+" />
 
-	<test url="http://optoutprescreen.com/" />
-	<test url="http://www.optoutprescreen.com/" />
-	<test url="https://optoutprescreen.com/" />
-
-	<rule from="^(?:http://(?:www\.)?|https://)optoutprescreen\.com/"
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/OptOutPrescreen.com.xml
```diff
@@ -1,15 +1,18 @@
...
 -->
-<ruleset name="OptOutPrescreen.com" default_off='failed ruleset test'>
+<ruleset name="OptOutPrescreen.com">
 	<target host="optoutprescreen.com" />
 	<target host="www.optoutprescreen.com" />
 
 	<securecookie host="^www\.optoutprescreen\.com$"
 			name=".+" />
 
+	<test url="http://optoutprescreen.com/" />
```
jsha: These test URLs are unnecessary, because they will be implicitly generated by the checker.

src/chrome/content/rules/OptOutPrescreen.com.xml
```diff
@@ -1,15 +1,18 @@
...
+	<test url="http://www.optoutprescreen.com/" />
+	<test url="https://optoutprescreen.com/" />
+
 	<rule from="^(?:http://(?:www\.)?|https://)optoutprescreen\.com/"
```
jsha: While you're in this file, could you update this rule to match the style I described in https://github.com/EFForg/https-everywhere/pull/2538/files?

PR #2536, opened 2015-08-12
German Unix User Group

MilanKral:


reedy:
Mixed Content: The page at 'https://www.guug.de/' was loaded over HTTPS, but requested an insecure script 'http://piwik.guug.de/piwik.js'. This request has been blocked; the content must be served over HTTPS.

Do we care about this?

jsha:
It looks like piwik is an analytics script, so is unlikely to break site functionality. So I'm fine with it. Though it raises a question of how a future mixed-content detector would deal with such subtle cases.

PR #2532, opened 2015-08-12
[Wells Fargo] Improved cosmetics, secured cookies, and added test URLs.

galenide:

PR #2538, opened 2015-08-12
Added a new ruleset for the ACLU of Idaho

galenide:


src/chrome/content/rules/The-ACLU-of-Idaho.xml
```diff
@@ -0,0 +1,14 @@
...
+	<target host="www.acluidaho.org" />
+
+	<securecookie host="^acluidaho\.org$"
+			name=".+" />
+
+	<test url="http://acluidaho.org/" />
+	<test url="http://www.acluidaho.org/" />
+	<test url="https://www.acluidaho.org/" />
+
+	<rule from="^(http://(www\.)?|https://www\.)acluidaho\.org/"
```
jsha: The preferred way to write this type of rule is like this:

```
<rule from="^http://www\.acluidaho\.org/"
  to="https://acluidaho.org />
<rule from="^http:"
  to="https:" />
```

Note that this version doesn't rewrite links from https://www.acluidaho.org/ to the no-www version. Since https://www.acluidaho.org/ doesn't work, it's very unlikely that anyone would link to it, so there's no need to fix it up. And we try to minimize rewrites of URLs that are already HTTPS wherever possible, to enable future performance optimizations.

src/chrome/content/rules/The-ACLU-of-Idaho.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="The ACLU of Idaho">
+	<target host="acluidaho.org" />
+	<target host="www.acluidaho.org" />
+
+	<securecookie host="^acluidaho\.org$"
+			name=".+" />
+
+	<test url="http://acluidaho.org/" />
```
jsha: These test URLs are actually unnecessary: They are generated implicitly by the checker.

galenide:
@jsha

The code for the ACLU of Idaho ruleset has been changed in line with your feedback. Pull requests #2532, #2530, and #2484 have also had their code updated.

When a pull request has been labeled as "needs-revision," is there a way for the user who opened the pull request to remove the label or otherwise indicate that the issue has been dealt with after they have revised the affected code in the pull request?

jsha:
Unfortunately GitHub doesn't allow people without push access to change labels on issues. Makes it a little hard to work with labels.

The best way to update me and the other reviewers that a pull request has been updated is to comment on the pull request, which will trigger and email, and we'll come take a look. Thanks

Issue #2313, opened 2015-07-23
Fix Transifex integration


jsha:
We have tests that fail if a message is present in the english locale
file but absent from some of the other locales, which makes it hard to
add a new message. Our integration with Transifex, our translation platform,
means that we should add new messages to the english locale, commit, then
wait a short while and pull the latest copy from Transifex, which will have an
instance of that message for every locale, defaulted to English (and updated
over time). Currently this would make our tests fail.

We should keep only the English locale in our source repo, with a submodule
that tracks the upstream Transifex repo. We should remove the test for "all
locales present" and move it to a separate set of pre-release tests.

reedy:
In WMF projects, we have translations done to translatewiki.net 99.9% of the time, but some authors do them manually to source files. Those get pulled in, and periodically we rebuild all language files, to remove any erroneous translations etc.

PR #2543, opened 2015-08-13
Use a submodule to pull locales from Transifex

jsha:
Previously we would manually copy over locale files from a separately synced
Transifex repo. This formalizes that arrangement by making that Transifex repo a
submodule and pulling in its translations at build time.

This change also makes compare-locales.sh operate on an XPI rather than the repo
directory, and check a file after building, when we expect all the locales to be
incorporated.

Builds should still be reproducible because the submodule sha1 specifies a specific commit.

Fixes https://github.com/EFForg/https-everywhere/issues/2313.

cc @cooperq for review.


cooperq:
your branch has conflicts, but I guess it looks fine other than that.

PR #2574, opened 2015-08-16
[Redd.it] Updated and activated

fuglede:
Redd.it now replies to HTTP requests with a 302 redirect to their HTTPS versions so that we can translate the URL directly.

Note that the ruleset test fails due to an unrelated error in <code>Alphabet.xml</code>/<code>Abc.xyz.xml</code> (see #2575).

Edit: With the above fixed, the test now fails due to another unrelated error.


PotcFdk:
LGTM

PR #2575, opened 2015-08-16
[Alphabet] Remove duplicate rule

fuglede:
There are duplicate rules for abc.xyz and www.abc.xyz in the current master, causing ruleset validation to fail. This removes one of those rulesets containing these.

### DIFF
--- a/src/chrome/content/rules/Alphabet.xml
+++ b/src/chrome/content/rules/Alphabet.xml
@@ -1,9 +1,0 @@
-<ruleset name="Alphabet">
-
-	<target host="abc.xyz" />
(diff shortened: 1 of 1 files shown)


TonyZTan:
@fuglede I don't really care but I just want to let you know that my rule (the one you deleted) was merged  on Aug 12 while the one kept was authored on the 15th. :wink:

fuglede:
@TonyZTan: Ah, thanks for the heads-up. I took the other one partly because it had a bit more info, but primarily because I think it's a lot easier to find rulesets if their file names and target hosts correlate. (Yours might prove more foresightful when Alphabet starts taking over the internet though.)

PR #2584, opened 2015-08-18
Removed 'spamcop.net' from target hosts since this domain doesn't sen…

### LINKED ISSUE #2583: spamcop.net doesn't offer valid certificate
spamcop.net (without subdomain www) doesn't offer a valid certificate (anymore?).

KommX:
…d a valid vert

This should fix #2583.

### DIFF
--- a/src/chrome/content/rules/spamcop.xml
+++ b/src/chrome/content/rules/spamcop.xml
@@ -1,6 +1,9 @@
 <ruleset name="SpamCop">
     <target host="spamcop.net" />
     <target host="www.spamcop.net" />
+    
+    <test url="http://www.spamcop.net" />
+    <test url="http://spamcop.net" />
 
     <rule from="^http://(www\.)?spamcom\.net/" to="https://www.spamcop.net/" />
 </ruleset>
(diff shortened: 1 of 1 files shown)


fuglede:
Hi @KommX. This certainly works: another solution that you'll find is quite normal is to leave in the target host but to change the rule so that it says

    <rule from="http://(www\.)?spamcom\.net/" to="https://www.spamcop.net/" />

instead. This way, the user is taken directly to the https://www version in both cases. While at the end of the day this will lead to the same thing, this does save a redirect and more importantly, all parts of the connection become secure.

KommX:
Ok, I understand. I now did this but the Travis CI build failed and I don't understand the error message.

fuglede:
Ah, sorry, my bad, the 'to' attribute should have been <code>^http://(www\\.)?spamcom\\.net/</code> rather than <code>http://(www\\.)?spamcom\\.net/</code> (with the extra <code>^</code>).

KommX:
You mean the 'from' attribute, don't you?

fuglede:
I did ...

fuglede:
Also I wrote spamcom instead of spamcop ... :)

semenko:
Cool -- thanks!

PR #2585, opened 2015-08-18
[Drugcom.de] New ruleset

cypherpunk:


fuglede:
Hi @cypherpunk. You actually want '.' instead of '\.' in your 'to'-attribute. The test fails because you should have trailing slashes in both 'to' and 'from' (so that 'drugcom.de' becomes 'drugcom.de/').

cypherpunk:
Thanks. Done.

TheNavigat:
LGTM. @fuglede Ping!

fuglede:
Thanks for the fix. This ruleset is actually not entirely good: the site does not serve a complete certificate chain which means that the user will be presented with a warning unless she has obtained the intermediate certificate via some other website (in which case the intermediate certificate might be stored). The [SSL Labs test](https://www.ssllabs.com/ssltest/analyze.html?d=drugcom.de&latest) shows this error by indicating that the intermediate certificate is an "Extra download".

semenko:
wfm now, merging

fuglede:
@semenko: It's still broken:

    $ curl -I https://www.drugcom.de

PR #2592, opened 2015-08-18
[Freifunk-Goettingen.de] new ruleset

julianladisch:


semenko:
goettingen.freifunk.net 301s to freifunk-goettingen.de, so redirect looks good. Thanks!

PR #2594, opened 2015-08-19
makexpi.sh: Fix cp (I think)

mnordhoff:
I think there's a mistake in a `cp` in the recent changes to `makexpi.sh`, but I don't use that part of it, so I'm not certain.

`cp $SUBDIR/$XPI_NAME.xpi pkg` was changed to `cp $SUBDIR/${XPI_NAME}-eff.xpi $SUBDIR/${XPI_NAME}-amo pkg` -- shouldn't that be `${XPI_NAME}-amo.xpi`?


jsha:
Thanks, good catch!

PR #2598, opened 2015-08-19
Delete Exactly-About.info.xml & Create GoNewSoft.com.xml

MojtabaFaal:
Delete Exactly-About.info.xml
Old Website

&

Create GoNewSoft.com.xml
New Website

### DIFF
--- a/src/chrome/content/rules/GoNewSoft.com.xml
+++ b/src/chrome/content/rules/GoNewSoft.com.xml
@@ -1,6 +1,7 @@
 <ruleset name="GoNewSoft.com">
         <target host="gonewsoft.com" />
         <target host="www.gonewsoft.com" />
-		
-		<rule from="^http://(?:www\.)?gonewsoft\.com/" to="https://gonewsoft.com/"/>
+	
+	<rule from="^http:"
+	to="https:" />
 </ruleset>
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/GoNewSoft.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="GoNewSoft.com">
+        <target host="gonewsoft.com" />
+        <target host="www.gonewsoft.com" />
+		
+		<rule from="^http://(?:www\.)?gonewsoft\.com/" to="https://gonewsoft.com/"/>
```
jsha: This rule removes the initial www but it is not necessary. You can instead use:

```
<rule from="^http:"
  to="https:" />
```

MojtabaFaal: I would like to be removed www

for example :
http://GoNewSoft.com
http://www.GoNewSoft.com
https://www.GoNewSoft.com

Redirect to :
https://GoNewSoft.com

jsha: Yes. But with the simple rule I suggested, visitors will go first to https://www.gonewsoft.com/, and then be redirected to https://gonewsoft.com/ by the server. We tend to prefer that style, unless the www version doesn't have a certificate.

MojtabaFaal:
Better?

jsha:
Great, thanks!

PR #2599, opened 2015-08-19
Add www.hs-karlsruhe.de and qis.hs-karlsruhe.de to ruleset

jannispinter:
This commit adds www.hs-karlsruhe.de and qis.hs-karlsruhe.de to the ruleset


semenko:
Thanks!

PR #2606, opened 2015-08-19
Fix justrun CLI params.

jsha:
Prior to this, Firefox would get the `--justrun` flag, which would prevent it from processing the next arg passed.

cc @cooperq for review


cooperq:
:+1:

PR #2615, opened 2015-08-19
Use LibSSH.org as test URL.

jsha:
Reddit now redirect to HTTPS on its own, so is not a good test.

cc @cooperq for review.


cooperq:
lgtm

PR #2623, opened 2015-08-21
Added Chromium testing

TheNavigat:
Also created test/ directory and renamed run-chromium.sh to test-chromium.sh with the ability to only run Chromium with the extension using --justrun


test-chromium.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+# Build the extension and run a chromium extension with it installed.
+#
+set -o errexit -o xtrace
+
+source makecrx.sh
+
+# If you just want to run Chromium with the latest code:
+if [ "$1" == "--justrun" ]; then
```
jsha: Move this above makecrx and add a `shift` here so the `--justrun` doesn't get passed to makecrx.

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
+# Run Chromium tests for HTTPS Everywhere
+#
+# This script may be executed as `python script.py [directory of CRX]`
+#
+# The script is compatible with Python 2. Python 3 is not tested.
```
jsha: This should include `#!/usr/bin/env python2.7` at the top.

test-chromium.sh
```diff
@@ -0,0 +1,22 @@
...
+	source utils/mktemp.sh
+
+	PROFILE_DIRECTORY="$(mktemp -d)"
+	trap 'rm -r "$PROFILE_DIRECTORY"' EXIT
+	chromium-browser \
+		--user-data-dir="$PROFILE_DIRECTORY" \
+		--load-extension=pkg/crx/
+else
+	echo "running tests"
+	python test/chrome/script.py pkg/
```
jsha: Run with `python2.7`

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
...
+	WARNING = '\033[93m'
+	FAIL = '\033[91m'
+	ENDC = '\033[0m'
+	BOLD = '\033[1m'
+	UNDERLINE = '\033[4m'
+
+extension = None
+
+for file in os.listdir(sys.argv[1]):
+	if file.endswith(".crx"):
```
jsha: This finds any `.crx`, not necessarily the latest one. Instead this should take a filename as an argument and the caller should pass the most recently built `.crx`. This can be done with `ls -tr | tail -1`, see `test.sh` for an example.

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
...
+		extension = file
+
+if(extension == None):
+	try:
+		raise Exception()
+	except:
+		traceback.print_exc(file=sys.stdout)
+		exit(3)
+
+print '' #New line
```
jsha: Why is this here?

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
+# Run Chromium tests for HTTPS Everywhere
+#
+# This script may be executed as `python script.py [directory of CRX]`
+#
+# The script is compatible with Python 2. Python 3 is not tested.
+# Selenium, WebDriver and Google Chrome (or Chromium) must be installed
```
jsha: Can you include a requirements.txt that install Selenium? Is WebDriver a Python package or ... ? If the former, include it in the requirements.txt.

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
...
+	except:
+		traceback.print_exc(file=sys.stdout)
+		exit(3)
+
+print '' #New line
+
+chromeOps = webdriver.ChromeOptions()
+chromeOps.add_extension(sys.argv[1] + '/' + extension)
+
+driver = webdriver.Chrome('/usr/bin/chromedriver', chrome_options = chromeOps)  # Optional argument, if not specified will search path.
```
jsha: What package does `chromedriver` come from on Ubuntu? Can you add it to `install-dev-dependencies.sh`? Also, the comment makes the line too long, please put it on the previous line and specify which argument is optional.

test/chrome/script.py
TheNavigat: I'm not sure what should be done when something fails. My first thought was "just raise an exception".

test/chrome/script.py
TheNavigat: Unfortunately ChromeDriver has to be downloaded, extracted, and moved to /usr/bin. It's not available as an Ubuntu package. Otherwise, I've added Selenium to the requirements.txt file and updated install-dev-dependencies.sh accordingly.

makecrx.sh
```diff
@@ -131,4 +131,4 @@ if [ -n "$BRANCH" ]; then
   cd ..
   cp $SUBDIR/$crx pkg
   rm -rf $SUBDIR
-fi
+fi
```
jsha: It looks like your text editor removes newlines at the end of files, which generates spurious diffs. Could you either configure it not to do so, or manually add an extra newline so you don't generate these diffs?

jsha:
@semenko: I'm going on vacation next week and probably won't have time to set up Chrome Driver and test this before I go. Can you take a look at this PR? Also, @TheNavigat is looking forward to adding additional tests once this one lands, and it would be great to have a mentor to help shepherd those through.

semenko:
@jsha Will do!

TheNavigat:
@semenko: Anything yet?

semenko:
Hey -- sorry for the delay. I'll try to get to this tonight. (~6-7 hours from now)

semenko:
Looks great so far -- I'm running into a chromium crashing but, so it'll be a day or two before I merge.

Thoughts on using the apt package `chromium-chromedriver` & pointing to `/usr/lib/chromium-browser/chromedriver`?

It'll probably be easier to integrate into Travis that way.

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
...
+	BOLD = '\033[1m'
+	UNDERLINE = '\033[4m'
+
+chromeOps = webdriver.ChromeOptions()
+chromeOps.add_extension(sys.argv[1])
+
+# First argument is optional, if not specified will search path.
+
+try:
+	driver = webdriver.Chrome('/usr/bin/chromedriver', chrome_options = chromeOps)
```
semenko: Please switch to /usr/lib/chromium-browser/chromedriver and add chromium-chromedriver to install-dev-dependencies.sh

test/chrome/script.py
```diff
@@ -0,0 +1,58 @@
...
+		print "ChromeDriver isn't installed. Check test/chrome/README.md for instructions on how to install ChromeDriver"
+		sys.exit(0)
+	else:
+		raise e
+
+driver.get('http://libssh.org/robots.txt')
+
+#Page Loaded
+
+if driver.current_url.startswith('https'):
```
semenko: Can you make these return exit codes too, with sys.exit()?  Super easy to add to Travis-CI.

semenko:
Hey -- awesome work! This looks great.

I'm happy to merge now if you like -- or you can add the apt chromium-chromedriver + exit codes, and we can then integrate this with Travis-CI.

TheNavigat:
Wow, I never knew that chromium-chromedriver existed. Turns out that the package exists, and even homebrew supports it.

I was going to add support for Mac OS X as well after finding out that Homebrew supports installing ChromeDriver as well. I checked this page, https://github.com/Homebrew/homebrew/blob/master/Library/Formula/chromedriver.rb , but I wasn't really sure where does {opt_bin} point to.

semenko:
Awesome -- nice work! I'll go ahead and merge this -- and feel free to add Mac support in a later PR.

Thanks again! I think this'll really help with Chrome testing.

semenko:
Awesome -- thanks again!

If you want to keep playing with Chrome testing, it'd be awesome to integrate this with Travis and the pre-push hook at:
https://github.com/EFForg/https-everywhere/blob/master/test.sh

You'll probably need to update the Travis apt packages to include chromium & chromium-webdriver in: https://github.com/EFForg/https-everywhere/blob/master/.travis.yml

TheNavigat:
You don't need to thank me, it's an honour for me to contribute in something like this.

I think I might need to make sure things work on Mac OS X before integrating stuff with Travis. Check #2721 . I'm running on Mac OS X right now, trying to figure out how stuff works there.

PR #2643, opened 2015-08-22
reenable LongTail.xml

J0WI:


J0WI:
This needs some additional tests for all the exclusions.

J0WI:
This rule needs a lot of work and the JW player is quite popular. Any help is welcome :)

TheNavigat:
I might try helping with these. Can you give me a brief on the problem you're facing with JW player?

J0WI:
fixed in #3531 :)

PR #2646, opened 2015-08-22
[Habrahabr.ru] Fix icons

### LINKED ISSUE #2620: Habrahabr icons are broken by rule
Rule name: Habrahabr.ru (partial)

fuglede:
The test wants a few more tests for the exclusion pattern. Not quite sure what they're supposed to be. Otherwise fixes #2620.

### DIFF
--- a/src/chrome/content/rules/Habrahabr.ru.xml
+++ b/src/chrome/content/rules/Habrahabr.ru.xml
@@ -14,7 +14,6 @@
 	Fully covered subdomains:
 
 		- auth
-		- special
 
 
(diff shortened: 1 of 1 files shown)


jsha:
The `auth(?:$|[?/]..` part of the exclusion pattern generates:

```
auth
auth?
auth/
```

So the test URLs should contain a URL that matches each of those (assuming they can be fetched).

fuglede:
Looks like that part of the rule (and some other parts) simply referred to 404 pages, so I opted to simply remove them.

PR #2651, opened 2015-08-23
[Akamai] Exclude imagesmtv-a.akamaihd.net

mnordhoff:
http://imagesmtv-a.akamaihd.net/ needs to be excluded to avoid breaking all the images on <http://www.mtv.com/news/>.

Edit: Several paragraphs about testing removed. I was able to mint enough mediocre tests for the other rules in the ruleset to satisfy the scripts.

### DIFF
--- a/src/chrome/content/rules/Akamai.xml
+++ b/src/chrome/content/rules/Akamai.xml
@@ -24,8 +24,9 @@
 -->
 <ruleset name="Akamai">
 
-	<target host="*.abmr.net" />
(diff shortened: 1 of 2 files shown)


jsha:
Test failure: `failure: Hostname a248.e.akamai.net shows up in 2 different rulesets.`

Need to add to duplicate whitelist.

mnordhoff:
That's what I get for making a trivial clean-up without testing it. ;_;

jsha:
Thanks for the fix!

PR #2655, opened 2015-08-24
Add ruleset for docs.saltstack.com.

aneeshusa:
The main site (saltstack.com) currently has mixed content warnings over HTTPS. I'm not sure if it's still ok to add to the ruleset or not (I don't think any functionality breaks, it looks like it's just some favicons?), so I would appreciate some guidance on that.


jsha:
Looking at the set of warnings on the main page, it appears the only active mixed content getting blocked on saltstack.com is two tracking scripts. Since the rest of the page works, I'd be fine with including a rewrite for the main page.

aneeshusa:
Great, went ahead and added the main site as well.

PR #2668, opened 2015-08-25
reenable Beck_PC.ch.xml

J0WI:


J0WI:
@jsha is 'missing certificate chain' obsolete since the ruleset tests has been fixed?

jsha:
Missing certificate chain is a potential problem, not just for the ruleset tests, but also for browsers. Browser behavior when confronted with a missing intermediate is unpredictable: Effectively, if the browser has a cached copy of that intermediate, it will work; otherwise not.

You can see this in action by using a fresh Firefox profile to fetch the page:

`firefox -no-remote -profile `mktemp -d` https://beckpc.ch`

jsha:
Also, just to be clear, this differs from "can't find issuer cert," which means that the chain was present, but the checker wasn't aware of the intermediate and so wasn't able to validate it (since the checker is bad at path building).

J0WI:
okay, closing for now.

PR #2681, opened 2015-08-26
[US Government] Simplified format, improved cosmetics, and added and …

galenide:
…removed domains.

### DIFF
--- a/src/chrome/content/rules/US-government.xml
+++ b/src/chrome/content/rules/US-government.xml
@@ -124,6 +124,8 @@
 		- www.cdc.gov				(Akamai; "temporarily unavailable")
 		- factfinder2.census.gov
 		- (www.)civicfed.gov			(cert: secure115.inmotionhosting.com; shows that domain's data)
+		- downloads.cms.gov			(unable to connect)
+		- innovation.cms.gov			(times out)
 		- (www.)compliance.gov			(cert: Plesk; shows default Plesk page)
 		- (www.)copyright.gov			(cert: wwws.loc.gov; shows that domain's data)
 		- (www.)cybercrime.gov			(ssl_error_rx_record_too_long)
@@ -138,6 +140,8 @@
 		- www.gpo.gov				(times out)
 		- archive.hhs.gov			(timeout)
 		- aspe.hhs.gov				(refused)
+		- cob.cms.hhs.gov			(cert mismatch)
+		- data.cms.hhs.gov			(cert mismatch; cert is for data.cms.gov)
 		- www.hhs.gov				(Akamai, shows 404 page)
 		- appropriations.house.gov		(400)
 		- clerk.house.gov			(timeout)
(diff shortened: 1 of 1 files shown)


youdly:
Good work! Would you like to refactor the domains into separate ruleset? I'm [thinking](https://github.com/EFForg/https-everywhere/pull/2073) of doing it, but there are so many domains involved.

galenide:
@youdly 

Moving domains from this ruleset into separate rulesets could be useful (an example would be moving all of the domains for Medicare to a Medicare-specific ruleset), particularly if an issue arises with some (but not all) of the domains that are currently in this ruleset. At the same time, the current instance of the US government ruleset should be an improvement over the currently disabled version. One possibility might be to enable the current instance of this ruleset and to subsequently disable it in the future if the domains that it contains are moved into separate rulesets.

youdly:
Indeed, re-enabling this ruleset should be prioritised over refactoring (which would involve lots of work). My idea is moving domains of specific branch/agency/program into separate ruleset, as you mentioned. Then at the same time, the domains would be removed from this ruleset (just like https://github.com/EFForg/https-everywhere/pull/2073/files). In the end, this ruleset would be left with www.usa.gov domain only.

TheNavigat:
@galenide Not sure if I'm missing something, but if you're waiting for someone to review this then I think you should follow @youdly 's advice.

semenko:
@galenide Oof, sorry this didn't get merged. Would you mind taking a look again -- and fixing the conflict?

We could also consider getting info from https://pulse.cio.gov/

konklone:
> We could also consider getting info from https://pulse.cio.gov/

That could be a helpful way to stay in sync, with the significant limitation that pulse.cio.gov is currently only covering second-level domains, not subdomains. (It's also only covering the federal government's 3 branches, not state/local/tribal domains.)

There's bulk data, updated whenever Pulse's data is update, that you could use as a source:
https://pulse.cio.gov/static/data/tables/https/https-domains.csv

konklone:
One source of subdomains is [the list of participating websites](https://analytics.usa.gov/data/sites.csv) in [analytics.usa.gov](https://analytics.usa.gov). It might be interesting to grab them from Certificate Transparency logs or censys.io data, too.

src/chrome/content/rules/US-government.xml
```diff
@@ -167,79 +185,87 @@ Fetch error: http://wtr.nhtsa.gov/ => https://wtr.nhtsa.gov/: (7, 'Failed to con
...
+	<target host="mymedicare.gov" />
+	<target host="www.mymedicare.gov" />
+	<target host="usajobs.gov" />
+	<target host="help.usajobs.gov" />
+	<target host="login.usajobs.gov" />
+	<target host="my.usajobs.gov" />
+	<target host="www.usajobs.gov" />
+	<target host="postalinspectors.uspis.gov" />
+	<target host="uspsoig.gov" />
+	<target host="www.uspsoig.gov" />
```
konklone: [uspsoig.gov is HSTS preloaded](@USGPO ) with an includeSubDomains policy, so you can use `*.uspsoig.gov` here instead of just `www`.

konklone: If you want HTTPS Everywhere to still use HSTS preloaded domains, you might go check out the [HSTS preload list](https://chromium.googlesource.com/chromium/src/+/master/net/http/transport_security_state_static.json) and take any `.gov` domains that have an `includeSubDomains` policy and add `*.` records for them.

Though, since HTTPS Everywhere is only available on Chrome and Firefox, and they both enforce HSTS preloading, you can **remove** rules from HTTPS Everywhere already covered by the preload list?

jsha: Yep, the current approach is to not include HSTS preloaded domains unless there is some additional reason, like securing cookies or including subdomains where the preloaded stuff doesn't have `includeSubdomains`.

konklone: In that case, you can remove:

* `uspsoig.gov` from this file
* `ftccomplaintassistant.gov` from [the FTC's file](https://github.com/galenide/https-everywhere/blob/improved-us-government-ruleset/src/chrome/content/rules/Federal_Trade_Commission.xml)
* `18f.gsa.gov` from the [GSA file](https://github.com/galenide/https-everywhere/blob/improved-us-government-ruleset/src/chrome/content/rules/GSA.gov.xml)
* the entirety of [the `notalone.gov` file](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Not_Alone.gov.xml)
* the entirety of [the `whitehouse.gov` file](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/WhiteHouse.gov.xml)

(I may have missed some.)

J0WI:
@galenide are you still interested to continue here?

PR #2694, opened 2015-08-26
Ruleset for uibk.ac.at

dschallert:
University of Innsbruck, Austria

LFU, OLAT, VPN should all be forced to HTTPS


semenko:
Thanks!

PR #2696, opened 2015-08-26
Fix Fastmail sending the wrong certificate

carbin:
The server sends a different, non-EV, certificate if you go to https://fastmail.com rather than https://www.fastmail.com, even though you get redirected to www the non-EV certificate persists.

I let Fastmail know about this a few months ago but they haven't fixed it.

### DIFF
--- a/src/chrome/content/rules/Fastmail.xml
+++ b/src/chrome/content/rules/Fastmail.xml
@@ -33,7 +33,17 @@
 	<rule from="^http://(www\.)?fastmail\.(com|fm)/"
 		to="https://www.fastmail.com/" />
 
+	<test url="http://fastmail.com/" />
+	<test url="http://www.fastmail.com/" />
+	<test url="http://fastmail.fm/" />
+	<test url="http://www.fastmail.fm/" />
+
 	<rule from="^http://(beta|qa)\.fastmail\.(com|fm)/"
 		to="https://$1.fastmail.com/" />
+
+	<test url="http://beta.fastmail.com/" />
+	<test url="http://qa.fastmail.com/" />
+	<test url="http://beta.fastmail.fm/" />
+	<test url="http://qa.fastmail.fm/" />
 
 </ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Fastmail.xml
```diff
@@ -27,8 +27,23 @@
 
 	<securecookie host="^\.www\.fastmail\.(?:com|fm)$" name=".+" />
 
-
-	<rule from="^http://((?:beta|qa|www)\.)?fastmail\.(com|fm)/"
-		to="https://$1fastmail.com/" />
+	<!--
+		The server sends an incorrect (non-EV) certificate when www. isn't used
```
jsha: I'd prefer not to redirect between domains when the only difference is an EV cert vs a non-EV cert. So long as fastmail.com (sans www) answers HTTPS correctly, I think serving the EV cert on that domain is up to the site operator.

I do appreciate the test url additions, though! If you'd like to submit this ruleset as a straight rewrite without changing domain names.

Also, looking at the ruleset more closely it looks like the existing rule rewrites fastmail.fm to fastmail.com unnecessarily: fastmail.fm answers HTTPS correctly.

carbin:
> I'd prefer not to redirect between domains when the only difference is an EV cert vs a non-EV cert. So long as fastmail.com (sans www) answers HTTPS correctly, I think serving the EV cert on that domain is up to the site operator.

It's not really a conscious decision. 

http://fastmail.com does a 301 redirect to https://www.fastmail.com

https://fastmail.com also does a 301 redirect to https://www.fastmail.com

The problem is when you hit https://fastmail.com first, it negotiates a TLS connection and gets a cert for *.fastmail.com instead of the EV-cert for www.fastmail.com and then because the two hosts share the same IP the browser doesn't renegotiate the TLS connection when it redirects and keeps the non-EV wildcard cert it already has. Essentially it still says "www.fastmail.com" in the address bar but because you came in through https://fastmail.com you have a different cert that if you had gone directly to https://www...

The thing is most people will never encounter this (without HTTPS Everywhere) because no one types "https://fastmail.com" in the address bar. They type "fastmail.com", which hits the http:// site first and get immediately redirected to https://www and therefore get the EV cert. The problem is HTTPS Everywhere makes typing "fastmail.com" equivalent to typing "https://fastmail.com"

jsha:
You make a good point that this is something we trigger rather than the provider triggering. Given that, and the fact that this is a no-www -> www rewrite, I'm okay with this.

PR #2707, opened 2015-08-28
[Wikimedia.xml] Remove some redundant rules

nextPrime:
https://cs.wikimedia.org/, https://cz.wikimedia.org/, https://download.wikimedia.org/, https://download.wikipedia.org/, https://sitemap.wikimedia.org/ no longer use invalid certificates, so these can be covered by the general http to https rewrite.


fuglede:
Looks good to me; thanks for the clean up!

PR #2710, opened 2015-08-28
Add rules for echo24.cz and echomedia.cz domains

djasa:
The rules were hand tested and the site work 99 % (only some
trackers/ads aren't loaded because of mixed content restrictions)


semenko:
Agreed -- looks liked mixed content isn't needed for the site to work.

PR #2715, opened 2015-08-29
Add rule for www.sainsburys.co.uk

alexmuller:
This rule only supports `www` because the bare domain doesn't have a valid SSL certificate.

Some things (like login) are already using HTTPS, but this rule ensures that all pages use HTTPS and we don't end up with silly things like login forms on an HTTP page with an HTTPS action.


semenko:
Thanks! LGTM

PR #2726, opened 2015-08-30
[THC.org] Disable ruleset due to expired cert

youdly:
Cert expired a month ago (01/08/15 15:29).


fuglede:
It looks like the HTTP version of the site itself serves a 301 Moved Permanently to its HTTPS counterpart.

Also worth noting, in the current ruleset, the subdomains ircs-web and wiki are never caught.

PR #2728, opened 2015-08-30
Update Freifunk.net

J0WI:


numismatika:
```
<securecookie host="forum\.freifunk\.net$" name=".+" />
```
for the forum might be a good idea as well, as it is not secured by default

J0WI:
*bump*

J0WI:
cc @loveithateit

PR #2731, opened 2015-08-30
Getting Travis CI to work with Chromium testing

TheNavigat:
This PR is not complete. test.sh does not start Chromium testing because of a bug I'm still trying to figure out how to fix. However, it's fine for me if someone would like to merge it and let me create another PR to fix this issue. Otherwise, I'm still figuring out how to make Chromium testing run properly through test.sh.


TheNavigat:
Okay, the PR's ready @semenko

TheNavigat:
On a side note, I'm delighted as fuck that we finally got this working.

semenko:
Whoa -- awesome!

Looks like Travis isn't quite as happy as it claims. Click the little green arrow by the commit for the full log:
https://travis-ci.org/EFForg/https-everywhere/builds/78120167

```
+./test/chromium.sh
++dirname ./test/chromium.sh
+++readlink ./test/chromium.sh
++dirname
dirname: missing operand
Try `dirname --help' for more information.
+cd ./test/../
+'[' '' == --justrun ']'
+./makecrx.sh
```

and early on

```
Disallowing packages: chromium-chromedriver
```


1. Looks like chromedriver for Travis might be tricky. There's probably info in this issue (I haven't read it yet): https://github.com/travis-ci/travis-ci/issues/272
2. We'll want to make sure that the ./test.sh script hard fails on these errors (so Travis can't erroneously report success). Not sure what's up here -- maybe we haven't set pipefail somewhere (e.g. `set -o pipefail`)

TheNavigat:
I mistakenly returned a 0 exit code when ChromeDriver's missing.

Regarding Travis, whoa... "Disallowing"?!

I checked the ticket again, read it line by line, and now I get why it isn't (and will never) work. It's because of Xvfb, Travis doesn't support it. I used a trick to run Chrome as a headless instance, by using pyvirtualdisplay, but as I checked right now, that relies on Xvfb as well, which is (again) not supported by Travis.

So yea... What to do now?

semenko:
Perhaps something like: https://github.com/travis-ci/travis-ci/issues/938#issuecomment-90791432 ? Not 100% clear that's successful for the commenter.

You can either remove Chrome from Travis here (and I'll merge this after a more careful review) and try a separate PR for Chrome + Travis (or Sauce Labs, I guess...), or try the above here.

TheNavigat:
This looks good, https://github.com/andy11/ScriptObservatory/blob/4587f5400dd95edc4fbac7a7a0a59c81532e3bf9/.travis.yml . I'll try it out in an hour or two. If it failed I'll remove comment Chrome's testing in test.sh and then we can merge this PR and start another one to slowly get the whole thing fixed step by step.

TheNavigat:
Wait a second... We do have Xvfb working...

I'll try out my trick, hopefully it'll work. Should me much more stable. However, apparently we need to download Chromedriver directly, since the package is basically "disallowed".

TheNavigat:
How can I test Travis locally?

TheNavigat:
I think I finally got it working...

semenko:
- Cool -- looks like there are a bunch of dirname errors still. There's one in head (https://travis-ci.org/EFForg/https-everywhere/builds/78068691) though this adds ~4 more. 

I haven't had a chance to see if they're sane -- but we should try to have as little spew there as possible.
```
++dirname
dirname: missing operand
Try `dirname --help' for more information.
```

- Does this actually run Chromium? I don't see that in the logs.

.travis.yml
```diff
@@ -16,10 +16,22 @@ virtualenv:
...
+  - pip install -r test/chromium/requirements.txt
 env:
   - DISPLAY=':99.0'
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+  - wget https://chromedriver.storage.googleapis.com/2.18/chromedriver_linux64.zip
+  - unzip chromedriver_linux64.zip
+  - sudo mv chromedriver /usr/bin/chromedriver
+  - sudo apt-get install -y python-software-properties
```
semenko: You should put this in .travis.yml's `apt:` section instead

.travis.yml
```diff
@@ -16,10 +16,22 @@ virtualenv:
...
+  - sudo apt-add-repository "deb http://dl.google.com/linux/chrome/deb/ stable main"
+  - sudo sed -i s/deb-src.*google.*//g /etc/apt/sources.list
+  - wget -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+  - sudo apt-get update -qq
+  - sudo apt-get install -y google-chrome-stable
+  - sudo chmod 1777 /dev/shm
```
semenko: Hm, does this really require sudo? Have you tried adding google-chrome-stable to apt directly (maybe it's whitelisted by Travis) and just looking in the working path for chromedriver? Could we just download the Chrome .deb, `ar xv` & run it?

My concern with `sudo: required` (below) is that it forces us off of Travis's newer / preferred container infra: http://docs.travis-ci.com/user/workers/container-based-infrastructure/

TheNavigat:
https://travis-ci.org/EFForg/https-everywhere/builds/78260977

Check the last few lines.

```
Created pkg/https-everywhere-2015.8.25~pre.crx
+echo 'running tests'
running tests
++tail -1
++ls -tr pkg/https-everywhere-2015.8.25~pre.crx
+CRX_NAME=pkg/https-everywhere-2015.8.25~pre.crx
+python2.7 test/chromium/script.py pkg/https-everywhere-2015.8.25~pre.crx
HTTP to HTTPS redirection successful
```

It does run Chromium.

And yea, I noticed those errors, but I'm not sure how to correctly fix them.

.travis.yml
TheNavigat: Not really, no. Chromedriver and every single Google Chrome package are disallowed, so you can't use apt for that.

semenko: Just `python-software-properties`

semenko: And fwiw, `chromium-browser` from the distro is allowed -- maybe that'd work? https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L296

.travis.yml
TheNavigat: I've tried, yes. It's disallowed.

There are two concerns regarding that, someone on Travis's issue mentioned that it'll never work without ```sudo```.

Second, the directory where ```chromedriver``` is extracted isn't a PATH directory, and I'm not sure how to move it correctly to the directory where the Selenium Python script exists.

For the first concern, here. https://github.com/travis-ci/travis-ci/issues/3354#issuecomment-90791268. Would you like me to try that?

For the second concern, I might need a little bit of help.

.travis.yml
TheNavigat: Ah! I see. Alright, I'll move python-software-properties to apt.

And I'll try chromium-browser. Hopefully it'll work.

semenko: Cool -- yeah, may or may not work. Only downside is we'll be using an older version of Chrome (M37), but probably OK http://packages.ubuntu.com/precise/chromium-browser

TheNavigat: So yea...

https://travis-ci.org/EFForg/https-everywhere/builds/78606378, "Chrome version must be >= 43.0.2357.0"

semenko: Try: https://chromedriver.storage.googleapis.com/2.12/chromedriver_linux64.zip

FWIW, I'm just trying to keep this as minimalist as possible -- since in *any* future version of Travis (when they upgrade beyond 12.04 LTS), we should be able to just apt-get install chromium + chromedriver

TheNavigat: Looks like it worked.

semenko:
Looking good! Some last thoughts:
- We should clean up the new dirname errors. I bet there's a quick fix, or some if[] test, basename, etc.
- Can you test real quick if it's necessary to chmod /dev/shm? (I'm trying hard to keep sudo out of this, again so we can stay on Travis's new infra: http://docs.travis-ci.com/user/workers/container-based-infrastructure/ )

TheNavigat:
Sorry for being late.

Remove the /dev/shm line works, but removing the mv chromedriver line makes the whole thing fail.

I'm only an amateur when it comes to bash, so I have no idea what to do in order to fix the dirname errors, and I don't know how can we move the chromedriver binary correctly to the script's location. I'll try looking up the latter though.

semenko:
Awesome -- looks good! I can take care of the dirname errors later on.

I'll do some manual testing (been mostly looking at Chrome/Travis stuff here), but then hopefully it'll be ready to merge!

TheNavigat:
```selenium.common.exceptions.WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally```

:(

semenko:
Heh, this has gone pretty deep down the rabbit hole. I'm guessing this breaks because a deprecated setuid/setgid requirement for Chromium (that we can ultimately fix with --disable-setuid-sandbox).

Switch sudo: false to sudo: true, and I'll do some offline tests then hopefully merge :)

TheNavigat:
I tried using --disable-setuid-sandbox, https://travis-ci.org/EFForg/https-everywhere/builds/78634510. Didn't work.

I'll switch it to required (I think we need required, not true) and push.

TheNavigat:
There's something weird happening. The path part is what makes Chrome fail.

See here, https://travis-ci.org/EFForg/https-everywhere/builds/78636591
And here, https://travis-ci.org/EFForg/https-everywhere/builds/78637370

TheNavigat:
So, the PATH command won't work, and moving ```chromedriver``` to ```test/chromium/chromedriver``` won't work either. https://travis-ci.org/EFForg/https-everywhere/builds/78647768

Selenium expects ChromeDriver to be in a PATH location, and does not try to locate chromedriver in its own directory. I can explicitly force it to do, though. I'll try that right now.

TheNavigat:
The non-sudo script is ready. This deserves a party.

semenko:
Awesome! Thanks so much for your hard work!

I'll do some offline testing, and hopefully merge this tonight! :+1:

semenko:
Cool -- a couple nits that I'll fix post merge. Thanks again!

PR #2739, opened 2015-09-01
update telnic rulesets

rdark:
updated telnic rulesets

### DIFF
--- a/src/chrome/content/rules/Telnic.xml
+++ b/src/chrome/content/rules/Telnic.xml
@@ -1,26 +1,31 @@
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://nic.tel/ => https://www.telnic.org/: (6, 'Could not resolve host: nic.tel')
-	Nonfunctional domains:
-
-		- dev.telnic.org	(ssl_error_rx_record_too_long)
-
--->
-<ruleset name="Telnic" default_off='failed ruleset test'>
-
-	<target host="nic.tel" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Telnic.xml
```diff
@@ -1,26 +1,31 @@
...
+                to="https://d0.webproxy.nic.tel/" />
+        <rule from="^http://d1\.webproxy\.nic\.tel/"
+                to="https://d1.webproxy.nic.tel/" />
+        <rule from="^http://s0\.webproxy\.nic\.tel/"
+                to="https://s0.webproxy.nic.tel/" />
+        <rule from="^http://s1\.webproxy\.nic\.tel/"
+                to="https://s1.webproxy.nic.tel/" />
+        <!-- registry/corporate site
+             cert valid for telnic.org and www.telnic.org -->
+        <rule from="^http://(?:www\.)?(?:nic\.tel|telnic\.org|telnic\.com|telnic\.net)/" to="https://www.telnic.org/"/>
```
TheNavigat: I believe all these ```<rule from="..." to "..." />``` tags should be replaced with ```<rule from="^http:" to="https:" />```

semenko:
Closing since this has outstanding work, and hasn't been touched in a while. Feel free to reopen if you'd like to have this merged.

[Also not sure we should be redirecting telnic's domains for them, in case they use them later. I prefer to keep http->https redirects within the same domain.]

PR #2743, opened 2015-09-02
We've changed e-commerce platforms for store.pfsense.org, and this ca…

cbuechler:
…uses a redirect loop on the new one. It's being fixed, but in the mean time, don't list it here.


semenko:
Thanks Chris!

PR #2749, opened 2015-09-04
[Hertz] New rulesets

youdly:
All the websites of Hertz and its subsidiaries that I can find.


semenko:
Wow, nice work -- especially with the weird affiliates like http://cinelease.com/.

By-region seems reasonable here, though that's a mega comment at the beginning.

Issue #2733, opened 2015-08-31
Common Application supports TLS


strugee:
[The Common Application][1] supports TLS.

 [1]: https://commonapp.org/

youdly:
There is existing ruleset https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Common_App.xml

strugee:
Hmm, weird. I definitely checked for existing rulesets... not sure what happened here.

PR #2752, opened 2015-09-04
[Commonapp.org] New ruleset

youdly:
Fix https://github.com/EFForg/https-everywhere/issues/2733


youdly:
There is existing ruleset https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Common_App.xml

PR #2769, opened 2015-09-05
ikvc.slovakrail.sk doesn't support HTTP, only HTTPS

MilanKral:


J0WI:
Removing the rewrite is a bad practice, sice you can now MITM http://ikvc.slovakrail.sk.
I think just reenable is fine.

PR #2784, opened 2015-09-06
reenable Cloudwear.com.xml

J0WI:


semenko:
Oh the irony... "World-class security for the connected world."

PR #2797, opened 2015-09-06
[growth.supply] New ruleset

youdly:
New Medium-related domain.

### DIFF
--- a/src/chrome/content/rules/Medium.xml
+++ b/src/chrome/content/rules/Medium.xml
@@ -1,16 +1,29 @@
 <!--
 	Other Medium rulesets:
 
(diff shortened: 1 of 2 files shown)


semenko:
Looks like there's a conflict -- can you take a look?

youdly:
Modification I made on Medium.xml is based on [master](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Medium.xml), but somehow this PR diff on [stable](https://github.com/EFForg/https-everywhere/blob/stable/src/chrome/content/rules/Medium.xml) branch instead.

PR #2800, opened 2015-09-06
rules: add Mikaela.info.xml

Mikaela:
I hope everything is OK with this, the test script didn't understand Arch (pacman) and my regexp isn't the best possible and I hope the \* matches everything like HSTS IncludeSubDomains.


semenko:
Looks good to me -- and our automated tests pass, so should be good. Thanks!

PR #2806, opened 2015-09-06
Cache Travis's python packages

semenko:
Just caching pip only saves the downloads, but the most time is spent compiling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 cache:
   directories:
     - $HOME/.cache/pip
+    - $HOME/virtualenv/python2.7_with_system_site_packages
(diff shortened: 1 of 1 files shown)


semenko:
Safe Travis-only code, I'm going to go ahead and merge this. (fyi @TheNavigat )

TheNavigat:
Good one. Are you sure the directory is correct, though?

semenko:
Yep! see: https://travis-ci.org/EFForg/https-everywhere/builds/79017800#L189

(Either way, worst case it just caches an empty dir.)

TheNavigat:
Oh, perfect, then :D

PR #2808, opened 2015-09-06
[Agar.io] Reenable ruleset

### LINKED ISSUE #2799: Re-add Agar.io
Hey,

Could you revert #2056? We changed the way that TLS is handled on the game servers, there shouldn't be any downsides to using it now.

Thank you.

fuglede:
Fixes #2799.

The ruleset was previously disabled upon webmaster request, but the webmaster has now requested that it be reenabled (GH #2799).


semenko:
Not sure why Travis didn't trigger, but merging and will watch master.

PR #2810, opened 2015-09-06
[Discourse.org] Fix subdomain cdn and simplify ruleset

### LINKED ISSUE #2775: incorrect redirect on cdn.discourse.org
https://www.eff.org/https-everywhere/atlas/domains/discourse.org.html

	<rule from="^http://cdn\.discourse\.org/" to="https://try-discourse.global.ssl.fastly.net/"/>

cdn.discourse.org is now HTTPS without a redirect, and try-discourse.global.ssl.fastly.net is now a dead domain. The rule should be changed to a plain upgrade.

fuglede:
Fixes #2775

### DIFF
--- a/src/chrome/content/rules/Discourse.org.xml
+++ b/src/chrome/content/rules/Discourse.org.xml
@@ -1,7 +1,6 @@
 <!--
 	CDN buckets:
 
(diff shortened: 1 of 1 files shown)


semenko:
LGTM

PR #2817, opened 2015-09-07
Adding required package to install-dev-dependencies.sh

TheNavigat:


semenko:
Seems fine, though not sure it solves the FF build issue.

TheNavigat:
It doesn't, but it's a required package.

PR #2821, opened 2015-09-08
Disable PrintFriendly Rules

j-manu:
The rules are invalid currently and HTTPS for our main domain will not
work 100% since we embed third party content which may not be HTTPS

### DIFF
--- a/src/chrome/content/rules/PrintFriendly.xml
+++ b/src/chrome/content/rules/PrintFriendly.xml
@@ -6,7 +6,7 @@
 	* Desk.com
 
 -->
-<ruleset name="PrintFriendly.com (partial)">
+<ruleset name="PrintFriendly.com (partial)" default_off="webmaster request">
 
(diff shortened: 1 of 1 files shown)


semenko:
Cool -- thanks!

semenko:
(Manu's listed on the about page, so seems legit.)

fuglede:
Certainly appears to behave a bit suboptimally when one uses the service on an insecure webpage. I had a brief look and if you want, you could secure parts of your site yourself:

* On the front page, the Wikipedia example could be changed from http to https.
* On the about page, the embedded video http://www.youtube.com/embed/SOYKL7KYeSQ could be changed from http to https.
* Likewise, https://www.printfriendly.com/browser_tool has a few securable YouTube videos.
* On the page https://www.printfriendly.com/button you could secure the blogspot images and those in your AWS bucket.

PR #2827, opened 2015-09-08
[Kabel Deutschland] Added kabeldeutschland.com & cleanup

rugk:


rugk:
Is this okay or should something be changed?

rugk:
@SoGood13 
Could you please write English?
With Google Translator I only get this (which does not make sense at all):
`` 
Why we do this to you ~
It amuses you ??
```

And I have no idea why this change got into a commit in https://github.com/EFForg/https-everywhere/commit/90a4768642747e36e2fb5004ddeecda23ca80edf

rugk:
@jsha and all other collaborators of this repo
How could I add [this commit] (https://github.com/EFForg/https-everywhere/commit/90a4768642747e36e2fb5004ddeecda23ca80edf)?

As you can see nobody merged my pull request and only this strange commit made me aware of it as it noticed me.

---
Okay, as the [real file](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/KabelDeutschland.xml) is the old one nothing seems to be changed, but this view is confusing...

rugk:
What about this PR?

rugk:
I don't want to jostle, but what about this? Any recommendation, questions or things to change?

rugk:
No one wants to review this PR?

rugk:
Still no review? :cry:

rugk:
Any progress here?

rugk:
This must be a difficult ruleset... :stuck_out_tongue_winking_eye: 

But seriously KabelDeutschland is a big German cable network operator and they seem to support HTTPS. So let's start using HTTPS there (on more sites than the old ruleset)! :smiley:

PR #2829, opened 2015-09-08
[1Cart] fix and reenable rules

julianladisch:


semenko:
Cool. Redirect from www....co.nz -> .com seems fine.

PR #2841, opened 2015-09-09
Set the size of favicons to 16x16 on Chrome extension

ngyikp:
On high DPI/Retina displays, Chrome uses a higher resolution image for the generic favicon. This can cause some odd layout placement on the popup page.

screen shot 2015-09-09 at 5 42 26 pm

The fix is to just simply set the width and height of the `````` tag to 16.

screen shot 2015-09-09 at 5 42 57 pm

### DIFF
--- a/chromium/popup.css
+++ b/chromium/popup.css
@@ -25,6 +25,8 @@ section.rules {
 
 /* Favicons */
 .rule img {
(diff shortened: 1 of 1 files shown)


semenko:
Awesome -- thanks for the super clean fix!

PR #2843, opened 2015-09-09
[SelfHTML.org] New ruleset

rugk:


fuglede:
More or less looks good to me; I noticed that some people on the site forum mention problems with mixed content, but perhaps that is irrelevant for HTTPS-E? Also, it looks like src.selfhtml.org could probably be added to the targets as well, cf. e.g. [this image](https://src.selfhtml.org/selfhtml_wiki.png).

rugk:
Done.
See also the discussion at https://forum.selfhtml.org/meta/2015/sep/9/https-bei-selfhtml/1649624 so possibly more SelfHTML.org sites will support HTTPS soon.

fuglede:
Cool! And thanks for the update to the ruleset.

PR #2835, opened 2015-09-09
[QuiteRSS.org] Update to new standard

ldgbc:

PR #2844, opened 2015-09-09
Revert "[QuiteRSS.org] Update to new standard"

semenko:
Reverts EFForg/https-everywhere#2835

(meant for master)


semenko:
Sorry for the noise -- didn't realize that PR was against stable, instead of master, so reverting.

PR #2845, opened 2015-09-09
[Intevation] Correcting rules & cleanup

rugk:
* switched default-on/off (problematic) files as they are now different
* cleaned them up
* activated wald.intevation.org (formerly intevation-problematic.xml)


rugk:
Fixes https://github.com/EFForg/https-everywhere/issues/2846

src/chrome/content/rules/Intevation-GmbH.xml
```diff
@@ -1,3 +1,10 @@
+<!--
+	Other Intevation rulesets:
+
+		- Greenbone.xml
+		- Intevation-problematic.xml
+		- Wald.Intevation.xml
```
semenko: please fix up this comment w/ the new filenames

semenko:
Awesome -- thanks! Also gets rid of a right-sided TLD wildcard -- which is great! :)

rugk:
> Also gets rid of a right-sided TLD wildcard -- which is great! :)

Yes, this had to be changed. :smiley: 
So any more ideas for https://github.com/EFForg/https-everywhere/issues/2846?

PR #2849, opened 2015-09-09
Exclude irc.t411.io from rewrite

Cilyan:
Special server irc.t411.io (also accessible through irc.t411.me) doesn't
provide any secured channel. Trying to access the server with https will
fail. An exclusion rule has been added thus.

Also added a coverage test for the server api.t411.io (and its .me
counterpart).

Note: yes, the server includes a specific web server, in addition to the
IRC service. The IRC service can be accessed with SSL on a specific
port, the web service cannot. The web server at irc.t411.io is a
different one than www.t411.io.

Signed-off-by: Cilyan Olowen <gaknar@gmail.com>


semenko:
LGTM, I see ERR_CONNECTION_REFUSED on both

PR #2856, opened 2015-09-10
Moving https-everywhere-tests to test/firefox

TheNavigat:


semenko:
Whoa -- so clean! :+1: 

Tests seem to execute and pass, so let's do it!

PR #2865, opened 2015-09-11
Add salesMANAGO.pl.

tomwys:


fuglede:
Hi tomwys, and thanks for the ruleset addition. The test fails, because part of your rules are already covered in the ruleset in <code>rules/SALESmango.pl.xml</code>. Ruleset filenames usually begin with an uppercase letter, but it looks like the existing one has a typo which should be fixed. It looks like you found a few more subdomains as well, so perhaps you want to include those in the existing ruleset?

While we're at it: you may have already seen it, but the (semi-new) [ruleset style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) suggests adding explicitly all functional subdomains rather than excluding non-functional. This has the benefit that new subdomains added to the site are not included in HTTPS-E automatically when they possibly shouldn't be.

tomwys:
I have fixed issues you have pointed. I'm not sure if "host" field in "securecookie" is correct by your standards.

fuglede:
Thanks for the fixes! A few more things:

* You can rename the file to remove the typo from the filename.
* It looks like www.salesmanago.pl (with the www) works as well (right?), so that you could add that as a target host.
* I *think* from a quick look at the source code that your securecookie host trick works. Normally, though, the hosts are specified explicitly; I'm not the right person to say why that is, but if nothing else, it has the advantage that it's more stable with respect to changes in the securecookie decision logic. That is, in this case, you could do something like <code>host="(\\.|app2|app3|www)?\\.?salesmanago\\.pl"</code>. (And yeah, this is not super consistent what's otherwise in the ruleset guide).

tomwys:
I have added "www" and fixed filename. I would like to leave securecookie as it is if no one argues. Do you have rights to merge or should we wait or ask someone else to do it?

fuglede:
Great. Before merging, I've opened #2907 regarding the securecookies.

tomwys:
I see that there is some development in #2907 topic. Do you think that this pull request can be merged now?

I would really love to use this rule inside my company and I don't want to patch every computer in office with custom rule.

Issue #2859, opened 2015-09-10
[broken ruleset] embedly.com


terrorist96:
http://genius.com/Calvin-harris-how-deep-is-your-love-lyrics

embedly.com *(NOT embed.ly (partial))* breaks video at top.

TheNavigat:
https://github.com/EFForg/https-everywhere/pull/2867

semenko:
Fixed in #2867

PR #2867, opened 2015-09-11
Fixing #2859

TheNavigat:


semenko:
(embed.com vs embed.ly threw me for a sec). LGTM.

PR #2869, opened 2015-09-11
Fixes #2572

### LINKED ISSUE #2572: [broken ruleset] gannett-cdn.com (partial)
http://thefreethoughtproject.com/cop-arrested-body-cam-refuted-report-shows-pummeling-teen/

gannett-cdn.com (partial) breaks the video in the article.

terrorist96:


semenko:
Yeah, definitely that rule. Seems to result in an error 400 w/ a248.e.akamai.net/crossdomain.xml

terrorist96:
How can it be fixed?

semenko:
Don't think it can -- it's ok. This rule is pretty specific / partial, and Gannett's TLS/Akamai setup is just broken. :/

PR #2879, opened 2015-09-12
add rule for DN.no

### LINKED ISSUE #2860: HTTPS Everywhere breaks videos on DN.no
HTTPS Everywhere breaks videos on dn.no. Disabling the "Brightcove (partial)" ruleset fixes this.
Sample: http://www.dn.no/magasinet/2015/09/07/1606/Popcorn-Time/inside-popcorn-time--the-worlds-fastest-growing-piracy-site

numismatika:
fixes #2860

### DIFF
--- a/src/chrome/content/rules/DN.no.xml
+++ b/src/chrome/content/rules/DN.no.xml
@@ -1,0 +1,8 @@
+<ruleset name="DN.no">
+
+	<target host="www.dn.no" />
+
+	<rule from="^http:"
+		to="https:" />
+
+</ruleset>


src/chrome/content/rules/DN.no.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="DN.no">
+
+	<target host="www.dn.no" />
```
semenko: Is there a reason you aren't including the bare name too? (host="dn.no"?)

numismatika:
The page redirects calls to the bare name to www anyway, doesn't seem to be used.

loveithateit:
@numismatika If it's okay for Eve to control the response when I enter 'dn.no' and press return, then the network is trusted. If the network is trusted, there's no point in encryption.

Issue #2880, opened 2015-09-12
HTTPS Everywhere breaks Symantec Connect Blogs


anonsubmitter:
HTTPS Everywhere breaks the layout of Symantec Connect Blogs to the degree that it is impossible to read any blog post or navigate the blog's frontpage without disabling the "Symantec (partial)" ruleset.

Samples:
http://www.symantec.com/connect/blogs
http://www.symantec.com/connect/blogs/firefox-vulnerability-could-allow-attackers-steal-documents

dholbert:
I hit this as well, with this blog post which is making the tech press rounds over the past few days:
http://www.symantec.com/connect/blogs/tough-day-leaders

@TheNavigat, looks like you've got a patch for this -- do you know if it's already being considered for inclusion? It'd be great to see this fixed.

TheNavigat:
@dholbert That's the PR link, https://github.com/EFForg/https-everywhere/pull/2881. It's not merged yet :/

semenko:
merged per above

PR #2881, opened 2015-09-12
Fixing #2880

TheNavigat:
I had to push using ```no-verify``` to get this through. It's a bit too much to refactor the whole thing while the fix is that simple.

### DIFF
--- a/src/chrome/content/rules/Symantec.xml
+++ b/src/chrome/content/rules/Symantec.xml
@@ -35,6 +35,8 @@
 		- service1		(works, expired, mismatched, CN: secure1.symantec.com)
 		- sfdoccentral		(works, expired 2011-12-14)
 		- www4			(breaks blog)
+		- corp.sarcscan (mistmatched, CN: submit.symantec.com)
+		- retail.sarcscan (mistmatched, CN: submit.symantec.com)
 
 	* Works, akamai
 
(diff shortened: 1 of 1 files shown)


TheNavigat:
Ugh I hate this...

TheNavigat:
I tried to fix this, but I ended up with a failure: ```Symantec.xml failed test: Rule redirects to something other than https.``` error. I commented stuff, trying to find the error, and then I ended up with this.

http://pastebin.com/BTty3dGE

That still returned the same error.

Any help, @fuglede ?

fuglede:
If I run <code>test/rules.sh</code> against your commit, I get the same result as Travis which is just that a bunch of tests are missing from the ruleset; I'm not actually seeing the error you're mentioning.

In your pastebin link though, <code>http</code> should be <code>http:</code> and <code>https</code> should be <code>https:</code>.

TheNavigat:
Yea, the errors are different indeed. I figured that I need to restructure the code in order to get this passing, and the error in the comment above was what the code in Pastebin produced.

Thanks for the ```http:``` and ```https:``` note, that fixed it.

TheNavigat:
@fuglede Can you check this PR again, please?

semenko:
Seems good + fixes a user-reported bug. Merging.

Issue #2253, opened 2015-07-17
CSS scaling issue for Firefox toolbarbutton in hamburger menu


fuglede:
This is probably of very low priority, but here goes: When the counter label is shown while the button is in the hamburger/Australis menu, the button (+ label) can become too big to fit in the default grid, which makes it act a bit wonkily, as it forces other buttons to move around.

To reproduce, place the button in the horizontal center of the menu grid, surrounded by other buttons on both sides, and run

```
document.getElementById("rsapplied").value = "10"
```

in the browser console.

One obvious fix to this would be to simply not show the label when the button is hidden away; a better one might be to move label a bit (for instance, Ghostery has its label fixed in the lower right corner of its toolbarbutton, which I personally like). It looks like <code>/src/chrome/content/toolbar_button_binding.xml</code> would be the place to change the style.

jeremyn:
@fuglede I can't tell if this is still a problem. Is it?

fuglede:
It is. Note that one now has to set `devtools.chrome.enabled` to `true` to run commands in the browser console interpreter.

Here's what it looks like on Firefox 48.0.2 on Windows; it is impossible to have 3 buttons in the row containing the HTTPS Everywhere button:

![untitled](https://cloud.githubusercontent.com/assets/6169306/18230025/f993bb9a-728f-11e6-8316-4687003e00ec.png)

jeremyn:
Okay, I see it now. The picture and the `devtools.chrome.enabled` reference helped.

Is this really a problem though? You're saying that if you increase the size of the add-on, it pushes other buttons around. What else would happen?

fuglede:
No other side effects, I would imagine. As mentioned in the original post, I do consider this low priority. Nonetheless, it did [show up in another issue thread as well](https://github.com/EFForg/https-everywhere/pull/2882#issuecomment-141923976).

jeremyn:
What is the side effect happening here?

fuglede:
One unintended effect is that if you have three buttons in the relevant row, then the third button will be forced down when the label grows in size, which in turn can mess up the ordering in the rest of the menu.

jeremyn:
If one button becomes too big, what else could Firefox do besides push the other buttons down?

Or is the issue that you can affect the size of the counter using CSS at all?

fuglede:
Well, I simply don't think that we should give rise to the misalignment in the first place. If, for instance, the label extended in the opposite direction, this would not be an issue.

[...]

PR #2882, opened 2015-09-12
New toolbar icons for Firefox and Chrome/ium

snwh:
I use https-everywhere on Firefox and Chrome, however the toolbar icons aren't the greatest or don't fit in, so I'd thought I'd design and propose some.

Proposed example:
https-everywhere-toolbars


semenko:
Whoa, I like these a lot! :100:

snwh:
@semenko thanks. :smile:

semenko:
FYI, the EFF folks are pretty busy with the https://letsencrypt.org/ launch.

(This might not get much discussion for a month or so.)

snwh:
@semenko no worries. I'm patient.

J0WI:
Would it be possible/better to use a single SVG instead of multiple PNGs?

snwh:
@J0WI it is possible, I wouldn't say better. These are rendered from an SVG, but if I recall correctly don't think extensions support SVG files for icons.

jsha:
@snwh: Thanks for designing these! As it happens, I was thinking the same thing, and had asked EFF's in-house graphic designer, Hugh D'Andrade, for a new logo for HTTPS Everywhere. The design he came up with wound up being very similar to yours: a larger circle, with clearer arrows pointing out towards the four corners. The main difference: I asked Hugh for an alternative to the lock icon. Since the lock in the URL bar is intended as an indicator of the security state, I felt like it would be confusing to have an always-on lock in the HTTPS Everywhere logo.

Here's Hugh's version: https-active-24

My current plan is to roll this out in the next release of HTTPS Everywhere, along with new banners for the homepage and Twitter account.

PotcFdk:
I agree with @jsha, having multiple locks near the URL bar might cause confusion with less experienced users who might not even know that they have HTTPSE installed, seeing as, for example, it might have been installed by a computer-savvy friend to improve the user's security experience.

@snwh while your version seems to *work well* with the other icons in your screenshot, have you checked how the icons look with a Firefox theme that has a dark background?

semenko:
Hm, agreed regarding padlock confusion, although the "S" strikes me as a bit random.

I'd love to see some mocks with other icon options (a shield? a key?)

snwh:
@semenko A lock is already the common metaphor for https in most (if not all browsers).

@jsha hmm, just an "S" doesn't really get across that a user is more secure when using https-everywhere.

@PotcFdk trying to account for all the various themes is a losing battle. It's the best case to design with the default theme colors in mind –which are all light-background.

snwh:
@jsha also, I still think making the icon fit the style of extensions' toolbar icons is a better route, which that one (no offence to Hugh) doesn't do.

semenko:
@snwh I actually like the lock a *ton* -- and definitely prefer it over the "S" and the current icon.

However, I'm also concerned about users mistaking the HTTPSe lock for "everything being secure" or other dangerous ideas. (I'm guessing that UX folks like @adrifelt would shame us for using a big padlock for HTTPSe.)

I'm reminded of this video of a non-technical user trying to interpret an SSL error: https://www.youtube.com/watch?v=ngCIU7b5j-M

(The top comments are now from confused users who've figured out that typing "danger" will silently bypass the TLS warning :/ )

snwh:
@semenko for what it's worth, I am in UX myself. ;)

semenko:
Oh of course! (Sorry, that that came out wrong. I meant more security interaction folks.)

I just have a sneaking suspicion that if we show a clearer large padlock to the right of every site, some subset of users will think "Hey, everything is hidden from my ISP/employer now!" or "This free coffeeshop wifi is now secure!"

That said, maybe they think that now, just by the extension's name?

There's no great way for us to A/B test, and the data are pretty weak (if anything, they show users don't care about padlocks at all...). 

From http://www.usablesecurity.org/emperor/emperor.pdf [PDF]:

"""
Prior studies have reported that few users notice the presence of HTTPS indicators such as the browser lock icon [6, 20]. Our results corroborate these findings and extend them by showing that even participants whose passwords are at risk fail to react as recommended when HTTPS indicators are absent.
The failure of all of our participants to respond to the removal of HTTPS indicators cannot be entirely attributed to ignorance. In the post-task questionnaire, three participants mentioned HTTPS indicators, though only in the context of explaining another attack clue.
"""

There's an excerpt in https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/41323.pdf [PDF] that's relevant:

"""
Friedman et al. asked participants whether screenshots of websites depicted secure connections; many participants could not reliably determine whether a connection was secure [15]. Whalen and Inkpen used eye-tracking software to determine that none of their 16 participants looked at the lock or key icon in the URL bar, HTTP(S) status in the URL bar, or the SSL certificate when asked to browse websites “normally” [33]. Some browsers modify the lock icon or color of the URL bar to tell the user when a website has an Extended Validation (EV) SSL certificate. Jackson et al. asked 27 study subjects to classify 12 websites as either phishing or legitimate sites, but the EV certificates did not help subjects identify the phishing sites [19].
"""

snwh:
@semenko I don't think a GitHub merge proposal is the place for a discussion on the bigger implications of the lock icon. ;) 

This is simply me proposing an update to the existing icons with using a symbolic version of https-everywhere's logo which is poorly rendered as these small toolbar icons.

ldgbc:
Will you be adding some color to this? It look plain like Firefox's hamburger icon. The Australis icon are all plain and boring. It cheaper to not add color I suppose.

snwh:
@ldgbc the whole idea is to match the menu icon and other toolbar icons.

adrifelt:
fwiw, I think it's OK to have the lock in the icon as long as it's clear that it's an action-icon and not a status-icon. meaning that you have to click on it for something to happen, rather than telling you whether the page you're currently looking at is secure. i think the arrows do that.

semenko:
Well, in that case, I'm 100% in favor of @snwh's icons. Thanks @adrifelt!

thecosmicfrog:
Other than @snwh's wonderful talents at work again, there is now a major reason to use this icon as the HTTPS Everywhere icon in Firefox is now affected by this bug which causes it to completely mess up the order of the Hamburger Menu:

https://bugzilla.mozilla.org/show_bug.cgi?id=1148826

fuglede:
@thecosmicfrog: Hm, how would the new icons change this? (For HTTPS-E, this bug is #2253 btw.)

snwh:
@fuglede it doesn't. The badge is independent of the icon.

J0WI:
This is our only ope PR with "High Priority". Have the EFF anything on the roadmap with the HTTPS Everywhere brand?

jsha:
We also have https://github.com/EFForg/https-everywhere/pull/4100/files open, but after a round of internal feedback, we've asked our graphic designers for another revision, so currently blocked on that.

sampablokuper:
Please can this PR no longer be tagged "High Priority"?

Security and basic functionality are high priority. A new logo, nice though it is (although I like the existing logo, too), is not. It is almost certainly lower priority than many of the other 194 pull requests currently in the queue, including some that are not labelled "High Priority".

Thanks :)

J0WI:
We have a complete new logo now, so I think I can close this one.

snwh:
A logo doesn't equal an icon, and as an icon designer, I'm kind of disappointed. Too many projects miss that concept and overuse their logo. But I won't argue the point further.

PR #2890, opened 2015-09-14
Rewrite subdomains for strugee.net

strugee:
As a side note, it would be nice if the test framework let me disable tests if I know the website maintainer (me) will check this.

### DIFF
--- a/src/chrome/content/rules/Strugee.net.xml
+++ b/src/chrome/content/rules/Strugee.net.xml
@@ -9,8 +9,8 @@
 	<!--	Direct rewrites:
 				-->
 	<target host="strugee.net" />
-	<target host="piwik.strugee.net" />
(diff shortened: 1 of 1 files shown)


strugee:
Also, the Travis build hasn't completed yet, but it's going to fail for unrelated reasons.

fuglede:
There's a typo in the comment and some inconsistent usage of <code>'</code> vs. <code>"</code>. Otherwise looks good to me; merging.

strugee:
@fuglede I'll prepare another PR; nice catch!

Looks like `"` is preferred, is that correct?

fuglede:
Yep, <code>"</code> is more popular.

strugee:
Great. I'll send a patch to the style guide as well.

PR #2891, opened 2015-09-14
Make ./test.sh --justrun work on OS X

strugee:
I've tested the OS X side of this fairly thoroughly. In theory it should also be tested on GNU/Linux, but the change is so simple, who cares. I'm sure it works.


semenko:
Whoa, cool! Thanks a bunch! (The signed extensions breakage by Mozilla is pretty sad .. )

PR #2896, opened 2015-09-14
Removed bad dashboard redirect.

epelletier:


semenko:
We generally prefer to default_off these rules, but OK removing, since @epelletier is from Power Reviews.

PR #2899, opened 2015-09-15
[UToronto.ca] (www.)library.utm.utoronto.ca is no good anymore.

mnordhoff:
Reported on IRC.

Update `ruleset-coverage-whitelist.txt` hash because omg it needs so many tests.


semenko:
Thanks!

PR #2900, opened 2015-09-15
Don't redirect gettingstarted.exposure.co to HTTPS

Tenzer:
It runs on Tumblr and only supports HTTP.


fuglede:
Looks good to me; thanks!

PR #2901, opened 2015-09-15
Provide a fallback chromedriver for other Linuxes as well

Tenzer:
When running the test scripts (and thus pre-commit hook) on a Linux distribution which is neither Ubuntu or Debian, the `chromedriver_path` variable isn't set at all, and the script will then always fail with a error such as this:

```plain
  File "test/chromium/script.py", line 51, in <module>
    driver = webdriver.Chrome(chromedriver_path, chrome_options=chromeOps)
NameError: name 'chromedriver_path' is not defined
```

This fixes that by providing the fallback value of just "chromedriver" to other Linux distributions, which at least makes it work on my machine (running Arch Linux).

### DIFF
--- a/test/chromium/script.py
+++ b/test/chromium/script.py
@@ -41,6 +41,9 @@ elif sys.platform.startswith("linux"):
         #Debian is lowercase when platform.linux_distribution() is used.
         #This is not a mistake.
         chromedriver_path = "/usr/lib/chromium/chromedriver"
+    else:
+        # Let's hope it's in the user's path.
+        chromedriver_path = "chromedriver"
 else:
     # Let's hope it's in the user's path.
     chromedriver_path = "chromedriver"


semenko:
Ooof, yeah, good call. That logic got pretty silly over a couple of merges (that's a new script since ~last week). 

Thanks a bunch!

Tenzer:
You're welcome. It could probably be cleaner by putting `chromedriver_path = "chromedriver"` above the entire if statement, and then scrapping the last two branches, but I wanted to make the change as small as possible.

semenko:
Yeah -- definitely. Testing got significantly restructured last week, so there are a bunch of hacks hanging around to make things work.

@TheNavigat want to take a look now that Travis/Debian, etc. have been fixed, and see if anything can be cleaned up?

TheNavigat:
The ```chromedriver_path = "chromedriver"``` above the entire if statement is a good idea. However, not doing that has a reason. I simply wanted to make sure that every listed distro is fully supported instead of adding support for distros that might not work. Quite close to the reason why rulesets guides were changed.

Since this PR basically ruins my whole "fully supported distros only" point, you can go ahead and add the whole path setting thing above the if statement.

@semenko Sure, I'll check that in a few!

Tenzer:
Oh, sorry, I wasn't aware of that. What does it take to be a fully supported distro then?
One problem with that - at least when using Python 2, is that Arch Linux apparently doesn't fill in a distribution name while it does however work on Python 3:

```plain
[jeppe@jeppe ~]$ python2 -c 'import platform; print(platform.linux_distribution())'
('', '', '')
[jeppe@jeppe ~]$ python3 -c 'import platform; print(platform.linux_distribution())'
('arch', '', '')
```

semenko:
I disagree with @TheNavigat  -- I think it's totally fine to provide a best effort "hey this might work" on other distros.

Clearly people are developing on non-Debian flavors -- and if people want to submit patches to make those distros work better, that's great!

Let's just set chromedriver to look in the path by default -- and if Arch, BSD, etc. run into fun problems, hopefully we'll get equally fun fixes.

TheNavigat:
@Tenzer, there are no specific rules, but basically I thought that we need to make sure everything works on the distro before adding it to the script.

And that's weird... However, if it works then we're good.

@semenko Alright then, let's do that instead.

Tenzer:
I have submitted #2903 which makes the change suggested.

PR #2908, opened 2015-09-15
Fix kinopoisk.ru charts

galeksandrp:
Kinopoisk.ru use amcharts to draw charts (for example, [here](http://www.kinopoisk.ru/votes/)). Some time ago charts were broken by extension.


semenko:
Awesome -- thanks!

PR #2910, opened 2015-09-16
[Surfeasy.com] Update ruleset

youdly:
New subdomains.


fuglede:
The securecookie host change disables cookie securing for cookies whose host is just surfeasy.com, so you want to make sure that that's still covered.

fuglede:
Looks fine; both new subdomains already redirect to HTTPS (with no HSTS).

Issue #2837, opened 2015-09-09
HTTPS Everywhere Atlas ignore case-senstive


ldgbc:
Issue: Search result search use case sensitive (Uppercase)
Solution: Resolve all uppercase as lowercase or ignore uppercase.

Reproduce: Go to https://www.eff.org/https-everywhere/atlas/index.html
Search keyword: Eff.org / https://www.eff.org/https-everywhere/atlas/index.html?Eff.org
Result is 0.
Now search keyword: eff.org / https://www.eff.org/https-everywhere/atlas/index.html?eff.org
Find 1.

TheNavigat:
I think that's easy to fix but I'm not sure if the public has access to the atlas's code.

jsha:
That's correct, we just have a copy on our website, though we should bring it into the HTTPS Everywhere repo.

I've fixed the case sensitivity, thanks!

PR #2914, opened 2015-09-16
Create archive.moe.xml

ldgbc:


fuglede:
Thanks, but the ruleset already exists with the same rules (and a few more) than this one contains. Generally, all ruleset XML files begin with an uppercase letter, which may be why you missed it.

ldgbc:
https://github.com/EFForg/https-everywhere/issues/2837

I only knew that when Travis failed.

PR #2920, opened 2015-09-17
[Threema.xml] New subdomain

rugk:
Added subdomain avatar

### DIFF
--- a/src/chrome/content/rules/Threema.xml
+++ b/src/chrome/content/rules/Threema.xml
@@ -11,6 +11,7 @@
     <test url="http://gateway.threema.ch/en/developer/api" />
     <test url="http://shop.threema.ch/download" />
     <test url="http://myid.threema.ch/revoke" />
+    <test url="http://avatar.threema.ch/*THREEMA" />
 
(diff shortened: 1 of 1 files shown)


fuglede:
Do you have a URL to exemplify what this subdomain serves? Currently, the HTTP/HTTPS versions of the web root serve different things (a 301 and a 404 respectively).

rugk:
It serves avatar images (used on gateway.threema.ch), e.g. https://avatar.threema.ch/*THREEMA
You're right. When accessing the domain via http it redirects to threema.ch (even when accessing an using an existing ID BTW: http://avatar.threema.ch/*THREEMA), but that's not the purpose of thid domain AFAIK. The purpose is getting user avatars. And if you do not pass a valid Threema ID there it just returns an 404 error as the ID is not found.

Maybe @threema-ch can say more about this, so I'm pinging them here.

threema-ch:
@rugk this is correct: the threema app try to fetch the avatar of the gateway ids with this url (e.g. https://avatar.threema.ch/*THREEMA). the service return a 404 if no avatar was found/set.

Issue #2907, opened 2015-09-15
Include securecookie host style in ruleset-style.md


fuglede:
Often we want to secure all cookies on all target hosts in a ruleset. As far as I can tell from the code ([relevant Firefox bit](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/code/HTTPSRules.js#L690) and [relevant Chromium bit](https://github.com/EFForg/https-everywhere/blob/master/chromium/rules.js#L308)), one easy way to obtain this is to let <code>host=".+" name=".+"</code>. Currently, most rulesets use long strings of subdomains (a la [this monstrous one](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/KK.dk.xml#L66)) which the new ruleset style explicitly wants to avoid for the actual rules.

Thus, for consistency, and if it is indeed possible, the ruleset style should probably suggest doing it the shorter way for securecookies as well.

jsha:
I agree! Thanks for checking in the code.

fuglede:
I guess this would also follow from #77 if one had some kind of 'complete coverage' attribute for rulesets, specifying that <code>\<securecookie host="." name="." /\></code> and <code>\<rule from="^http:" to="https" \></code> be automatic, so that one only needs to specify target hosts.

jeremyn:
@fuglede In pull request #5639 , @J0WI pointed out to me that a `.+/.+` `securecookie` can cause problems if one of the domains is only partially included. Maybe this is what you meant by requiring a "complete coverage" attribute. I agree that it would be nice if, in the simple cases, a `securecookie` could be automatically applied.

Anyway this is an old issue, can we wrap it up somehow?

fuglede:
Another thing one might need to be aware of is that subdomain-including `securecookie`s will be covered by `.+`, which can become an issue if not all subdomains are securable. I have yet to see an instance of this in the wild though.

In terms of wrapping it up, it would just be a matter of stating something to the extent of "If you are positive that all cookies can be safely secured, use `.+`/`.+`" in the style documentation.

jeremyn:
What's the difference, if any, between my comment [here](https://github.com/EFForg/https-everywhere/pull/6731#issuecomment-244774660) (seen in the wild) and the below?

>Another thing one might need to be aware of is that subdomain-including securecookies will be covered by .+, which can become an issue if not all subdomains are securable.

fuglede:
[...]

PR #2921, opened 2015-09-17
[Let's Encrypt.org] New ruleset

rugk:


reedy:
Any reason we couldn't rename it (to save some history) and then just update the file?

src/chrome/content/rules/LetsEncrypt.org.xml
```diff
@@ -0,0 +1,22 @@
...
+        and the cross-signature is not in place too. However this will be changed later.
+-->
+<ruleset name="Let's Encrypt">
+    <target host="letsencrypt.org" />
+    <target host="www.letsencrypt.org" />
+    <target host="helloworld.letsencrypt.org" />
+    <target host="community.letsencrypt.org" />
+
+    <!--
+        testing URl which does not redirect to HTTPS by default, see
```
reedy: mixed case on URL?

rugk:
Okay, renamed.

src/chrome/content/rules/LetsEncrypt.org.xml
rugk: done

fuglede:
Seems to me that adding helloworld.letsencrypt.org partially defeats the purpose of that page and might lead to possible confusion. I'd suggest leaving that out until the CA is live.

fuglede:
Regarding the securecookie, I wasn't actually able to find one that was not already secured. Can you confirm that those exist? Otherwise there's not much point to that one. (Not that it hurts -- #2907 is relevant here.) As a minor note, <code>www.</code> was probably intended to be <code>www\\.</code> (and similarly for <code>community.</code>).

Finally, the included test URL does in fact redirect to HTTPS (now?);

    $ curl -is http://letsencrypt.org/repository | grep "Location"
    Location: https://letsencrypt.org/repository

rugk:
> Can you confirm that those exist?

No, just added it like it is. As you said it does not hurt.

> As a minor note, www. was probably intended to be www\. (and similarly for community.).

Indeed. I'll change that.

> Finally, the included test URL does in fact redirect to HTTPS (now?);

Yes it seems like it is.

rugk:
So everything was fixed? Does someone wants to merge it?

fuglede:
I would hold that this still breaks the intended behaviour of http://helloworld.letsencrypt.org.

rugk:
Okay, okay. I commented `helloworld.letsencrypt.org` out.

fuglede:
Also, the comment on /repository is still inaccurate.

rugk:
`helloworld.letsencrypt.org` is [now trusted](https://letsencrypt.org/2015/10/19/lets-encrypt-is-trusted.html) by all major browsers, so should I possibly add it back to this ruleset?

rugk:
So what about this PR?

fuglede:
👍

PR #2928, opened 2015-09-18
Adding YoungArts, fixes #2923

### LINKED ISSUE #2923: youngarts.org supports TLS
At least, the homepage does. Triggers MCB, though I'm not sure if it really matters.

TheNavigat:


fuglede:
Hm, it would appear to me that the CA signing their cert may not necessarily be in browser store.

Firefox (Debian's shipped version with the default store):

    youngarts.org uses an invalid security certificate.

    The certificate is not trusted because the issuer certificate is unknown.

Details:

	$ openssl s_client -connect youngarts.org:443 2>&1 1>/dev/null
	depth=0 serialNumber = 1ts7pADuf88hxLGRA21QM4xqLtaV-knn, OU = GT54699839, OU = See www.geotrust.com/resources/cps (c)14, OU = Domain Control Validated - QuickSSL(R), CN = www.youngarts.org
	verify error:num=20:unable to get local issuer certificate
	verify return:1
	depth=0 serialNumber = 1ts7pADuf88hxLGRA21QM4xqLtaV-knn, OU = GT54699839, OU = See www.geotrust.com/resources/cps (c)14, OU = Domain Control Validated - QuickSSL(R), CN = www.youngarts.org
	verify error:num=27:certificate not trusted
	verify return:1
	depth=0 serialNumber = 1ts7pADuf88hxLGRA21QM4xqLtaV-knn, OU = GT54699839, OU = See www.geotrust.com/resources/cps (c)14, OU = Domain Control Validated - QuickSSL(R), CN = www.youngarts.org
	verify error:num=21:unable to verify the first certificate
	verify return:1

So it looks like the site can be supplied with a valid chain but doesn't supply it itself. To the extent that browsers fetch the remaining certs themselves this is not an issue, but I hesitate to merge. The [SSL Labs test](https://www.ssllabs.com/ssltest/analyze.html?d=youngarts.org&latest) marks this by noting that the intermediate cert is an "extra download".

TheNavigat:
Oh, I see. Chrome didn't report this problem on Ubuntu 14.04 LTS (not sure which Chrome version, but it's the latest stable one AFAIK - can't access my laptop right now :/ ) so that's the reason I didn't know that happens.

I'm not sure either whether the PR should be merged or not, knowing that the supplied certificate is incomplete. I'm biased towards not merging, though. Better safe than sorry.

PR #2929, opened 2015-09-18
Fixes #2917

### LINKED ISSUE #2917: SNstatic.fi ruleset is broken
The SNstatic.fi ruleset breaks http://www.iltasanomat.fi/

You cannot view some of the static content like images.

Example here: http://is12.snstatic.fi/img/626/1289033434966.jpg

TheNavigat:


fuglede:
Looks like is11 and is13 are broken for the same reason is12 is. On the other hand, parts of the main page (such as lehti.iltasanomat.fi) appear to be securable.

TheNavigat:
https://lehti.iltasanomat.fi/ returns a mismatching certificate, and I think it redirects to a login page that redirects page to http://lehti.iltasanomat.fi/ - I don't understand the language.

So yea, I'm not sure if we should add that to the ruleset.

fuglede:
Right, but at least some of the content (looks like the web root is excluded) on that domain could be redirected to e-pages.dk, in the same way that the hs1[0-3] domains are redirected to what appears to be a third party. In any case, that's for a different ruleset.

PR #2931, opened 2015-09-19
Fixes #2887

### LINKED ISSUE #2887: HTTPS Everywhere breaks Torcache
HTTPS Everywhere breaks torcache.net so that torrent files won't load and pop up in a "save window". Disabling the "Torcache" ruleset fixes this.
Sample: https://torcache.net/torrent/72F242DB89E763B6CE390F25D576195C2169B149.torrent

TheNavigat:


fuglede:
It [seems to me](https://github.com/EFForg/https-everywhere/issues/2887#issuecomment-142396027) that this exclusion is unnecessary.

TheNavigat:
The file never actually downloads. The URL loads correctly but you can't download the file.

fuglede:
That's peculiar. It works in both Firefox and Chrome here.

TheNavigat:
Weird... I'll double check in a couple hours, maybe they fixed it on their own side.

TheNavigat:
Yep, peculiar. Works here as well. I guess I'll just close the issue.

fuglede:
Great; thanks for checking!

PR #2933, opened 2015-09-19
Create noblogs.org.xml

J0WI:


TheNavigat:
For some reason people included noblogs here. https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Autistici-Inventati.xml

J0WI:
*ping* @fuglede

PR #2934, opened 2015-09-19
Update QuiteRSS.org.xml

ldgbc:


TheNavigat:
You have a couple of syntax errors there.

```<ruleset name="QuiteRSS.org"``` should be followed by a ```>``` at the end.

And there's an extra ```=``` here, ```to=="https:" />```. It should be ```to="https:" />``` instead.

**Comment by ldgbc** [comment]
Thanks @TheNavigat 
Appreciate it, still trying to figure this whole thing out.

File: src/chrome/content/rules/QuiteRSS.org.xml [hunk 1]
```diff
@@ -1,14 +1,15 @@
 <!--
-Disabled by https-everywhere-checker because:
+Pre-2015-09-18 Disabled by https-everywhere-checker because:
 Fetch error: http://www.quiterss.org/ => https://www.quiterss.org/: (51, "SSL: no alternative certificate subject name matches target host name 'www.quiterss.org'")
+2015-09-18 Remove default_off since website has been update to fix the error and update to new standard.
 -->
-<ruleset name="QuiteRSS.org" default_off='failed ruleset test'>
+<ruleset name="QuiteRSS.org"></ruleset>
```
TheNavigat: ```</ruleset>``` should be removed. Think of it as an array, ```<ruleset name="QuiteRSS.org"></ruleset>``` is simply declaring an empty ruleset. Instead, what closes the opened tag is the ```</ruleset>``` at the bottom.

src/chrome/content/rules/Answers.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="Answers">
+        <target host="answers.com/" />
+        <target host="www.answers.com/" />
+        <target host="file*.answcdn.com/" />
```
TheNavigat: Hosts in ```target``` tags shouldn't be followed by /. Think of those as domains/hostnames, those are not URLs

TheNavigat: You can't really have ```file*.answcdn.com``` as a host. Either use ```*.answcdn.com```, or explicitly define the target for each ```file*.answcdn.com``` subdomain.

TheNavigat:
Just a tiny bit of advice, you should run ```install-dev-dependencies.sh``` to install development dependencies, and then you'll be able to run ```test.sh``` to test your updates before you commit/push them.

TheNavigat:
You should also have each new rule in its own independent PR and its own branch. Makes it easier for people who merge stuff :)

semenko:

PR #2938, opened 2015-09-19
[Spiegel.de] fix Fußball live

hauke:
This fixes this page for example:
http://www.spiegel.de/sport/fussball/bundesliga-live-ticker-tabelle-ergebnisse-spielplan-statistik-a-842988.html#contest=bl1

This page uses javascript and it fails to load the additional parts without this change.


fuglede:
Looks good to me; thanks for the fix!

PR #2939, opened 2015-09-19
[spkrb.de] activate again

hauke:
http://banking.spkrb.de/ works for me, this looks like a temporary error to me.


fuglede:
It times out here. Is it just super unstable?

hauke:
There is only a https version of this website, how should this be handled?

fuglede:
Ah, good catch, what you did makes sense. Let me open an issue about it so it doesn't just get automatically disabled again in the future.

PR #2940, opened 2015-09-19
Fix tldrlegal.com

galeksandrp:
Website cert valid for `tldrlegal.com`, `community.tldrlegal.com`, `dev.tldrlegal.com`, but not for `www.tldrlegal.com`.


fuglede:
Looks good to me; thanks!

PR #2941, opened 2015-09-20
[ForestSangha.org] New Ruleset

lucaspetter:


fuglede:
Site serves a 301 Moved Permanently to its https version. No HSTS though.

PR #2944, opened 2015-09-20
Create Nyaa.eu.xml

Chris344:
Force HTTPS for the nyaa.eu domain


TheNavigat:
Would be nice if you could fix the indentations. Otherwise, LGTM.

youdly:
I don't think `test url` is necessary.

Chris344:
Removed the test url

PR #2946, opened 2015-09-20
Create derstandard.at.xml

J0WI:


src/chrome/content/rules/derstandard.at.xml
```diff
@@ -0,0 +1,42 @@
...
+	<test url="http://mobil.derstandard.at/img/ui/searchLupe.gif" />
+	<test url="http://www.derstandard.at/img/ui/searchLupe.gif" />
+	<test url="http://derstandard.at/js/fixedfixed.js" />
+	<test url="http://mobil.derstandard.at/js/fixedfixed.js" />
+	<test url="http://www.derstandard.at/js/fixedfixed.js" />
+	<test url="http://derstandard.at/wetter/Content/img/search-icon2.png" />
+	<test url="http://mobil.derstandard.at/wetter/Content/img/search-icon2.png" />
+	<test url="http://www.derstandard.at/wetter/Content/img/search-icon2.png" />
+
+	<rule from="^http://(www\.)?derstandard\.at/(ajax|css|img|js|wetter/Content/img)/"
```
TheNavigat: That line causes many 404s. The page loads without any CSS.

src/chrome/content/rules/derstandard.at.xml
```diff
@@ -0,0 +1,42 @@
...
+		to="https://derstandard.at/$1/"/>
+	<rule from="^http://mobil\.derstandard\.at/(ajax|css|img|js)/"
+		to="https://mobil.derstandard.at/$1/"/>
```
TheNavigat: The dot here is not escaped correctly.

src/chrome/content/rules/derstandard.at.xml
```diff
@@ -0,0 +1,42 @@
...
+	<rule from="^http://epaper.derstandard.at/"
```
TheNavigat: The dot here is not escaped correctly. Same applies to the following rule.

J0WI: Thanks for the feedback. The test server was down today, so I wasn't able to continue.

J0WI:
Now it seems to work and passed all tests.

src/chrome/content/rules/derstandard.at.xml
```diff
@@ -0,0 +1,49 @@
...
+<ruleset name="derStandard.at (partial)">
+	<target host="derstandard.at"/>
+	<target host="www.derstandard.at"/>
+	<target host="epaper.derstandard.at"/>
+	<target host="images.derstandard.at"/>
+	<target host="mobil.derstandard.at"/>
+
+	<exclusion pattern="^http://derstandard\.at/$" />
+	<exclusion pattern="^http://www\.derstandard\.at/$" />
+	<exclusion pattern="^http://mobil\.derstandard\.at/$" />
```
TheNavigat: I think those are useless since the following rules don't include those links anyway.

J0WI: These are necessary to pass the tests.

TheNavigat: Oh, I see... That doesn't look like a good way to solve that problem, but I'm not sure if there's a better way to do it without restructuring the whole rewrite system.

So yea, LGTM.

J0WI:
*ping* @fuglede

numismatika:
I would add 
```
	<rule from="^http://livestat\.derstandard\.at/"
		to="https://livestat.derstandard.at/" />	
```
as well.

J0WI:
Thanks!

This site is on Alexa rank 12 in Austria, hopefully someone has time to review this :)
@fuglede @reedy

semenko:
LGTM, :fireworks: Thanks! :fireworks:

PR #2951, opened 2015-09-21
support unprefixed box-sizing

J0WI:


J0WI:
Should we still support Firefox <29 (removing the prefix fallback)?

J0WI:
*ping* @fuglede

semenko:
I don't quite understand this -- does `box-sizing: border-box` fix a visual issue for you?

J0WI:
I have no idea why we use box-sizing, but the moz prefix is deprecated.

semenko:
Oh, sure, reasonable. :+1:

PR #2955, opened 2015-09-22
Add Simplednscrypt.org

bitbeans:


TheNavigat:
Technically Simple DNSCrypt automatically redirects users on HTTP to the HTTPS site so HTTPS Everywhere doesn't actually do anything here, so I'm not sure if the rule needs to exist in the ruleset or not.

However, it doesn't break anything, so when it comes to that the PR LGTM.

fuglede:
It does do something: when the user first connects to the website, and before they receive the 301 redirect to HTTPS they are still open to MITM attacks.

In this particular case though, it's true that the ruleset achieves very little: the site in question uses HSTS,

    $ curl -vs https://simplednscrypt.org 2>&1 | grep "Strict"
    < Strict-Transport-Security: max-age=31536000; includeSubdomains; preload

which means that if the user is using a HSTS compatible browser, this ruleset does nothing for the following year. And it's worse (well, better) than that: since 'preload' is specified, the site is set for inclusion in the browsers' HSTS preload lists, and when it's included (which it is not at the moment but it won't be long), this ruleset truly does nothing.

fuglede:
And [now it's included](https://code.google.com/p/chromium/codesearch#chromium/src/net/http/transport_security_state_static.json&l=3731). I guess we could still use this ruleset for legacy browsers or in case the browsers drop the preload lists again in the future. (But by that argument, we could include the entire preload list which is probably not the best idea when there's already performance issues.)

PR #2956, opened 2015-09-22
New rule for motorola.de

KommX:
One could merge this rule into the existing "Motorola.xml" but I haven't done it because there are missing tests.

### DIFF
--- a/src/chrome/content/rules/Motorola.xml
+++ b/src/chrome/content/rules/Motorola.xml
@@ -63,20 +63,39 @@
 -->
 <ruleset name="Motorola (partial)">
 
-	<target host="*.email-motorola.com" />
 	<target host="*.motorola.com" />
 		<exclusion pattern="^http://(?:www\.)?motorola\.com/(?:Business|staticfiles)/" />
 	<target host="cdn.motorola-2.com" />
 	<target host="motorolasolutions.com" />
 	<target host="*.motorolasolutions.com" />
+	<target host="www.motorola.de" />
 
 
 	<securecookie host="^(?:globalrepairtools|mediacenter|membership)\.motorola\.com$" name=".+" />
 	<securecookie host="^mediacenter\.motorolasolutions\.com$" name=".+" />
+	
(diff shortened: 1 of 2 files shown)


TheNavigat:
I believe that it's cleaner to add the tests to Motorola.xml and then add the .de domain to the hosts.

KommX:
Ok, I think I managed to do it.

TheNavigat:
LGTM

TheNavigat:
LGTM.

src/chrome/content/rules/Motorola.xml
```diff
@@ -63,23 +63,39 @@
...
+		to="https://www.motorolasolutions.com/" />
+	
 	<rule from="^http://motorola\.com/"
 		to="https://www.motorola.com/" />
+		
+	<rule from="^http://www\.motorola\.de/"
+		to="https://www.motorola.de/" />
 
 	<rule from="^http://(globalrepairtools|mediacenter|membership|sdc|www)\.motorola\.com/"
 		to="https://$1.motorola.com/" />
```
youdly: For cleaner rule, you can simply line 97-101 to
`<rule from="^http:"
    to="https:" />`

KommX: This doesn't seem to be possible.
After your recommended change 'test.sh' complains:

ERROR src/chrome/content/rules/Motorola.xml: Not enough tests (0 vs 1) for <Rule from '^http://cdn\.motorola-2\.com/' to 'https://cdn.motorola-2.com/'>
ERROR src/chrome/content/rules/Motorola.xml: Not enough tests (0 vs 1) for <Rule from '^http://mediacenter\.motorolasolutions\.com/' to 'https://mediacenter.motorolasolutions.com/'>

youdly: Remove line 103-104 & 109-110.

KommX: Thank you for your suggestions. I now committed them.

PR #2960, opened 2015-09-22
Fix access to default_off attribute in RuleSets lib.

### LINKED ISSUE #2959: rules.js ignores default_off attributes
RuleSets library used internally in `rewriter` (and probably in other places) ignores `default_off` attribute on rulesets due to incorrect access.

RReverser:
`.attributes.default_off` returns `undefined` since `attributes` returned from `DOMParser` is array-like `NodeList` and not key-value object. Attribute access is correctly implemented via `getAttribute` in all the other places of the code, so this could be just an oversight.

Fixes #2959.


jsha:
Thanks!

PR #2966, opened 2015-09-22
removed mixedcontent flag from weltbild.ch

J0WI:


fuglede:
Looks good; site serves a 301 Moved Permanently to https.

PR #2971, opened 2015-09-22
Update ruleset for eigenLab.

VittGam:
- Added etherPad -> pad rule.
- Added eXploitPisa.org rules.
- ownCloud and GitLab will be restored soon. Keep the rules for them in the meantime.

Cheers,
Vittorio

### DIFF
--- a/src/chrome/content/rules/eigenLab.xml
+++ b/src/chrome/content/rules/eigenLab.xml
@@ -1,19 +1,20 @@
-<!--
-	Expired:
-		- owncloud
-	
-	503: 
-		- git
--->
 <ruleset name="eigenLab">
 	<target host="eigenlab.org"/>
 	<target host="www.eigenlab.org"/>
 	<target host="pad.eigenlab.org"/>
+	<target host="etherpad.eigenlab.org"/>
+	<target host="git.eigenlab.org"/>
+	<target host="owncloud.eigenlab.org"/>
 	<target host="wiki.eigenlab.org"/>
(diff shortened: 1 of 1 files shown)


fuglede:
Thanks for the ruleset update. It's a bit hard to review at the moment, but here are some comments;

- The etherpad rule looks good to me.
- The git subdomain (HTTPS and HTTP) and owncloud (HTTP) subdomains serve 503s. The HTTPS version of owncloud serves a self-signed certificate. Is this intentional?
- The exploitpisa pages time out at the moment.

VittGam:
Thank you for your reply!

The git and owncloud servers are currently down, but we're working on it. They're going to use HTTPS and HSTS for sure when they'll be restored.

Also, the eXploitPisa.org domain is down since two days because there are problems at our noc. This is going to be solved shortly.

Anyway I'm going to move the rules for eXploitPisa.org to a new "eXploit!" ruleset, since it makes more sense to us.

So please don't merge for now; I'm going to add commits here.

Best,
Vittorio

VittGam:
(Actually the HTTPS owncloud certificate that's being served now is not self-signed but expired; unfortunately, recent "modern" versions of browsers have got the warning pages too simplified so it's quite difficult to understand what the actual technical problem is... But `curl` can always help. ;) )

semenko:
Closed since this has some outstanding work and hasn't been touched in a while. Feel free to re-open if you'd like this merged. :+1:

PR #2977, opened 2015-09-22
Update CPanel.guru.xml

J0WI:


fuglede:
Hm, is it possible to add some context for this one? Currently, http://cpanel.guru and https://cpanel.guru serve somewhat different content. Perhaps the former is never used?

J0WI:
I have no idea, just tried to fix the rule.

semenko:
Yeah, gonna NAK this since the domains are serving different content.

PR #2978, opened 2015-09-22
Delete Echo.xml

J0WI:


tomwys:
It seams that website was deleted.

fuglede:
Regarding the aboutecho.com one, is it clear that this is not simply a temporary issue? http://blog.aboutecho.com/ is still live and has fairly recent updates.

The JS-Kit one looks deletable though.

J0WI:
I just checked on archive.org: The site seems to be down since June.

PR #2980, opened 2015-09-23
Fix static.websimages.com rewrite

RReverser:
Current rewrite was either typo or something changed, anyway test URL below failed.

### DIFF
--- a/src/chrome/content/rules/Webs.xml
+++ b/src/chrome/content/rules/Webs.xml
@@ -71,8 +71,10 @@
 	<target host="static.websimages.com" />
 
 
-	<rule from="^http://images\.(?:free)?webs\.com/"
-		to="https://a248.e.akamai.net/f/1129/8333/3m/$1s.com/" />
+	<rule from="^http://(images\.(?:free)?webs\.com)/"
+		to="https://a248.e.akamai.net/f/1129/8333/3m/$1/" />
+
+		<test url="http://images.webs.com/static/global/js/webs/usersites/ie6subnav.js" />
 
 	<!--	Protocol-relative links on members.webs.com:
 								-->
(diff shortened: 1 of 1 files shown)


fuglede:
Good catch. Seems like the site is down right now so it's hard to verify but from the looks of it, the rule above the one you fixed is also broken.

RReverser:
Maybe. I was just running automatic comparison of results of own rewriter vs rules.js and found this one.

fuglede:
I wasn't actually able to find any links that get rewritten by it but my guess would be that the issue is the same: that there's a capture group missing.

fuglede:
Right, found a useful test URL:
http://images.webs.com/static/global/js/webs/usersites/ie6subnav.js can be rewritten to https://a248.e.akamai.net/f/1129/8333/3m/images.webs.com/static/global/js/webs/usersites/ie6subnav.js

Could you include this one in your PR as well, simply adding the relevant capture rule where it's currently missing?

RReverser:
@fuglede Sure. Can you please also show some example for the "free" part?

fuglede:
http://images.freewebs.com/Members/Generationz/Standard/PrintCSS/print.css

Looks like both 
  https://a248.e.akamai.net/f/1129/8333/3m/images.freewebs.com/Members/Generationz/Standard/PrintCSS/print.css
and
  https://a248.e.akamai.net/f/1129/8333/3m/images.webs.com/Members/Generationz/Standard/PrintCSS/print.css
are working so I'm not sure why the 'free' part has been uncaptured.

RReverser:
Probably just to avoid capture group, dunno.

One more question - are you sure that rewrite was indented to target `https` (I mean in `^https?://(static\.websimages\.com)/`)?

fuglede:
Yes, it looks like that becomes necessary on https://members.webs.com where there's a reference to "//static.websimages.com/vfefa01f/active-static/lib/modernizr.js".

RReverser:
Ah, certificate mismatch. I see.

RReverser:
Done!

fuglede:
Looks good; thanks!

PR #2988, opened 2015-09-24
T411 got a new .in domain

Cilyan:
The website T411 got an additional domain name.
It already responded to t411.me and t411.io, now it also responds on
t411.in.
The change adds support for this new domain.

Signed-off-by: Cilyan Olowen <gaknar@gmail.com>

### DIFF
--- a/src/chrome/content/rules/T411.xml
+++ b/src/chrome/content/rules/T411.xml
@@ -3,6 +3,8 @@
   <target host="*.t411.io" />
   <target host="t411.me" />
   <target host="*.t411.me" />
+  <target host="t411.in" />
+  <target host="*.t411.in" />
   
   <test url="http://wiki.t411.io/" />
   <test url="http://www.t411.io/" />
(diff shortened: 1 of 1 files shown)


fuglede:
I'm having some issues with the wiki subdomain. For instance, if one clicks "Linux" on https://wiki.t411.in/doku.php/start?idx=console, nothing happens as http://wiki.t411.in/lib/exe/ajax.php is (falsely) blocked as mixed content.

The rest of the ruleset seems good.

Cilyan:
I've already reported that they had a problem in their redirection configuration, but until now I hadn't seen any case where this could block the access to pages. Thanks for the report. Let's suspend shortly this PR, time to report the problem again a see if something is changed. If not, I'll add `platform="mixedcontent"` in the rule.

fuglede:
Thanks for checking with them. If it ends up not being fixed, you could include the problematic subdomain in a separate ruleset with <code>platform="mixedcontent"</code> and keep the working ones in the current one; like the separation in <code>Webs.xml</code> and <code>Webs-MixedContent.xml</code>.

Micdu70:
@Cilyan Hello, nothing changed yet... so please add the ruleset to this subdomain. Forcing https on www.t411.in is expected since 2 months now. :(

semenko:
Lot of mixed content here still.

PR #2991, opened 2015-09-25
[21.co] New ruleset

youdly:


fuglede:
Looks good; site itself serves a redirect to its HTTPS counterpart. No HSTS though.

PR #2993, opened 2015-09-25
[ESET] Add subdomains

rugk:
Added subdomain *help.eset.com* and *ela.eset.com* and more...

### DIFF
--- a/src/chrome/content/rules/Eset.xml
+++ b/src/chrome/content/rules/Eset.xml
@@ -70,7 +70,6 @@
 	<target host="go.eset.com" />
 	<target host="socialmediascanner.eset.com" />
 	<target host="anti-theft.eset.com" />
-	<target host="anti-theft.eset.com" />
 	<target host="parentalcontrol.eset.com" />
 	<!--target host="kb.eset.com" /-->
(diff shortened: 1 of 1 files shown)


youdly:
Seems like Eset has many subdomains, why not simplify them into *.eset.com and then exclude non-functional subdomains.

rugk:
Because you should avoid using wildcards. It's better to include every domain.
See https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

youdly:
Maintainer seems fine with wildcard though (https://github.com/EFForg/https-everywhere/pull/2279#issuecomment-123026249). Perhaps this ruleset doesn't have too many subdomains.

rugk:
I also think in this case there are more domains which do not support HTTPS than those who do.

rugk:
So is this mergeable now?

rugk:
Anything which speaks against merging?

fuglede:
Looks good to me. As a somewhat minor point: the targets of the original rulesets were ordered alphabetically, which made it a bit easier to navigate.

rugk:
Ah okay. Nice time I'll pay attention to this.

PR #2994, opened 2015-09-26
[Mega] Update ruleset

youdly:
New domain.


J0WI:
Could you also remove the obsolete RC4 comment?
It would be great to have some test urls for this ruleset.

youdly:
Added Add test url for subdomains.

semenko:
:fireworks: thanks! :fireworks:

PR #2997, opened 2015-09-26
Update Transport for London

alexmuller:
TfL has got a lot better since this rule was written in September 2012. Almost everything supports HTTPS and there are just a few subdomains which don't. The certificate is a wildcard cert now.

### DIFF
--- a/src/chrome/content/rules/Transport_for_London.xml
+++ b/src/chrome/content/rules/Transport_for_London.xml
@@ -7,14 +7,20 @@
 	<target host="tfl.gov.uk" />
 	<target host="*.tfl.gov.uk" />
 
+	<exclusion pattern="^http://art\.tfl\.gov\.uk/" />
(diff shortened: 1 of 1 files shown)


alexmuller:
I had a bit of trouble running the tests locally - I think because the location of https-everywhere-checker was changed in 39c3297f9d25b405455f7c6c17da2aba5e6da74a but the certificate directory in the `manual.checker.config` of https-everywhere-checker points to the wrong place. But I'm not sure about that.

semenko:
Hey -- looks great -- I'd love to merge this, but it has some conflicts. Could you please resolve them? (Or even submit the tiny testing change as a different pr?)

alexmuller:
Thanks for the prompt, I've pulled out the other commit because ae60b4179cbf19db14627f9dc185ea7a520da6f7 is on master and solves the same problem.

semenko:
Awesome -- thanks!

PR #2999, opened 2015-09-26
Add Esolangs.org

q--:
Notes:

* The cookie name regex can just be `.`; it just has to match, no need for it to capture the entire string. (In fact, you can do `new RegExp("")` with an empty string, but I don't know how this works under the hood so I used a dot just to be safe.)
* Non-capturing groups perform significantly better than capturing groups.


fuglede:
Looks good to me; thanks! Regarding the first comment, #2907 is also relevant.

PR #3006, opened 2015-09-27
Add www.lloydsbank.com

alexmuller:
The certificate doesn't support the bare domain.


anand-bhat:
You can force the non-www domain to www with something like this:

	<target host="lloydsbank.com" />
	<target host="www.lloydsbank.com" />

	<rule from="^http://(?:www\.)?lloydsbank\.com/"
		to="https://www.lloydsbank.com/" />

semenko:
Or simply exclude the bare domain, since we don't want to prevent the site owner from using barenames in the future.

alexmuller:
I think this'll work as is, only the `www` subdomain is included as a target.

semenko:
Fine by me. Thanks!

PR #3012, opened 2015-09-28
[Ruhr-Uni-Bochum.de] Add some exceptions

fuglede:
This fixes https://github.com/EFForg/https-everywhere/issues/3009


J0WI:
Here it would make sense to split the rule in a falsemixed ruleset.

PR #3013, opened 2015-09-29
Nativo additions and cleanup

robinkunde:
Nativo has a new version of their JS library that's served through a new domain. While adding a rule file for that, I noticed that the other rule files covering them needed to be brought up to the new standards, so I did that as well.

### DIFF
--- a/src/chrome/content/rules/Ntv.io.xml
+++ b/src/chrome/content/rules/Ntv.io.xml
@@ -1,0 +1,19 @@
+<!--
+	Ads.
+
+	For other Nativo coverage, see:
+		Nativo.net.xml
+		Postrelease.com.xml
+
+	Fully covered subdomains:
+
+		- s.ntv.io
+		- a.ntv.io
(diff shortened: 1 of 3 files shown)


src/chrome/content/rules/Ntv.io.xml
```diff
@@ -0,0 +1,19 @@
...
+
+	Fully covered subdomains:
+
+		- s.ntv.io
+		- a.ntv.io
+-->
+
+<ruleset name="ntv.io (partial)">
+	<target host="a.ntv.com" />
+	<target host="s.ntv.com" />
```
loveithateit: These don't wfm.

robinkunde: wfm?

robinkunde:
Ok, I fixed the typo for ntv.io

semenko:
Aweome -- thanks! :fireworks:

PR #3015, opened 2015-09-29
[SwitchVPN.net] New ruleset

youdly:


fuglede:
On https://www.secure.switchkonnect.com/cart.php?a=view&language=english, one encounters blocked mixed content from http://secure.switchkonnect.com/index.php?m=PGPC&modaction=getConfiguration. Is this something we care about?

You could make a rule from http://www.switchvpn.net to https://switchvpn.net to have that covered as well. Also, do you want to squash the four commits into a single one?

PR #3017, opened 2015-09-29
Remove redundant regex parts.

### LINKED ISSUE #2842: Some regexps use `?` quantifiers for look-aheads
Probably purely a code style issue, but some regular expressions use constructs like `(?=...)?` which is valid in JavaScript but don't make sense and only complicate expressions. Here's a list of all the found ones:

```javascript
{ input: '^http://www\\.heteml\\.jp/[^?]*(?=\\?)?',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 36,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
{ input: '^http://store\\.mandriva\\.com/[^?]*(?=\\?.*)?',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 43,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
{ input: '^http://soe\\.stanford\\.edu/+(?=\\?.*)?$',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 37,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
{ input: '^http://(?:www\\.)?tescofinance\\.com/+(?=\\?.*)?$',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 46,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
{ input: '^http://hilfe\\.udmedia\\.de/[^?]*(?=\\?.*)?',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 41,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
{ input: '^http://livechat\\.udmedia\\.de/[^?]*(?=\\?.*)?',
  error: 
   { name: 'RegexSyntaxError',
     type: 'NothingRepeat',
     lastIndex: 44,
     message: 'Nothing to repeat!Repeat after assertion doesn\'t make sense!' } }
```

RReverser:
Closes #2842.


fuglede:
> Not enough tests (0 vs 185)

Ouch!

As you may already be aware, some old rulesets are currently whitelisted from the test coverage test as they were made before the implementation of that. Optimally those rulesets are updated as time goes on, but in this particular case, it might be more reasonable to simply update the whitelist.

In practice, the whitelist is implemented by having [utils/ruleset-coverage-whitelist.txt](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist.txt) contain sha256sums of the whitelisted filenames, so that for example the Stanford one could be updated to

<code>82403e71eec1d2e3304cf33e63a77a3aee0821586edacf0091fa98478ac5542e  Stanford-University.xml</code>

RReverser:
Uh oh. Thanks for noting.

fuglede:
Here are the others; if you could add them to utils/ruleset-coverage-whitelist.txt (removing the existing entries), we should be good.

    cd57e308cea1dcbb49b9eaca13934dd2b1d6b8fa1b6b441367814d996b4389c3  Stanford-University.xml
    95365d0076da114c98b41049f775b4e36a52bf8297842da9d533802ef3317761  Heteml.jp.xml
    ea8f9bf69a5d5029ab1c217760e6e27c3b5a4e9672ace704a2f9f7d8332bb497  UD_Media.de.xml
    510cc20ce2abc399d5826465e6f232cf1d9b205f1750c9906c177cbcecf7324f  Mandriva.com.xml
    80a2b0d54ed1eabf4ff9c4d5b2d7955a37a11eb7ea64736afefe5df15052f142  Tesco_Bank.com.xml

RReverser:
Sure. Got busy yesterday.

RReverser:
Let's see if it worked.

fuglede:
👍 Thanks!

fuglede:
Hm, I should have checked this before, but I'm a bit confused now, as it looks like some of the rewrites would give inequivalent results; for instance, <code>http://soe\.stanford\.edu/?whatever</code> I would have thought would be caught by <code>^http://soe\\.stanford\\.edu/+(?=\\?.*)?$</code> but not by <code>^http://soe\\.stanford\\.edu/+$</code> (because of the <code>$</code>). [Debuggex](https://www.debuggex.com/) seems to disagree though -- am I missing something or is this a JavaScript peculiarity?

Coincidentally, soe.stanford.edu seems to not exist. If this were an issue, it would also be so on tescofinance.com, but I'm not actually seeing any problems there.

RReverser:
> `http://soe.stanford.edu/?whatever` I would have thought would be caught by `^http://soe\.stanford\.edu/+(?=\?.*)?$`

Hm, why do you think it would be?

fuglede:
Ah, sorry, I was being confused. Makes me wonder why there's a <code>$</code> at all though, since the rewrites would appear to work find without it.

RReverser:
> Ah, sorry, I was being confused. Makes me wonder why there's a $ at all though, since the rewrites would appear to work fine without it.

Yeah, looks like for Stanford it shouldn't be there at all.

PR #3020, opened 2015-09-29
Remove adainitiative.org: no longer supports HTTPS

puzzlement:
Unfortunately Ada Initiative moved hosting to WordPress.com, which does not support HTTPS on mapped domains. Therefore, I request that our entry in HTTPS Everywhere be removed per this pull request.


semenko:
LGTM

PR #3026, opened 2015-10-01
[Coralproject.net] New ruleset

rugk:


fuglede:
Looks good to me.

Note: Site serves the HSTS header with 'preload' and 'includeSubDomains' specified, meaning that if the hostname gets included in the browser preload lists, this ruleset will not do very much for a user who uses a browser with the preload list active.

PR #3027, opened 2015-10-02
fix makexpi.sh when no files in ./translations/

ghost:
makexpi.sh nolonger fails when there are no files in translations/ directory
#2556


semenko:
Breaks on OS X :( -- also I don't entirely understand when there wouldn't be translations available, though I generally build only for Chrome

semenko:
```
cp -a --no-target-directory -- test1/ test2/
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file ... target_directory
```

semenko:

PR #3031, opened 2015-10-03
[MP3Ninja] New ruleset

fuglede:
Some of these already redirect to HTTPS, but http://mp3ninja.skull doesn't.


fuglede:
Hm, messed up the name; closing and reopening.

PR #3032, opened 2015-10-03
[MP3Skull] New ruleset

fuglede:
Some of these already redirect to HTTPS but http://mp3skull.ninja does not.

### DIFF
--- a/src/chrome/content/rules/MP3Skull.xml
+++ b/src/chrome/content/rules/MP3Skull.xml
@@ -1,0 +1,17 @@
+<ruleset name="MP3Ninja">
+
+    <target host="mp3skull.is" />
+    <target host="www.mp3skull.is" />
+
+    <target host="mp3skull.ninja" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/MP3Skull.xml
```diff
@@ -0,0 +1,17 @@
...
+    <target host="mp3skull.ninja" />
+    <target host="www.mp3skull.ninja" />
+
+    <securecookie host=".+" name=".+" />
+
+    <rule from="^http://(?:www\.)?mp3skull\.is/"
+            to="https://mp3skull.ninja/" />
+
+    <rule from="^http://(?:www\.)?mp3skull\.ninja/"
+            to="https://mp3skull.ninja/" />
```
loveithateit: The site appears to have moved to .wtf

fuglede:
So it would seem; thanks for checking. I updated the ruleset to reflect this change.

PR #3033, opened 2015-10-03
Added Tor2web.xml

virgil:
These are the modifying rules for ensuring a Tor Browser Bundle user doesn't accidentally use a Tor2web node.

* I do not know to ensure this only occurs for Tor Browser Bundle users and not all HTTPS-everywhere users.  I presume the admins of this repository do.

* I will be doing an additional modification for adding Tor2web to the downgrade whitelist

### DIFF
--- a/src/chrome/content/rules/Tor2web.xml
+++ b/src/chrome/content/rules/Tor2web.xml
@@ -1,0 +1,25 @@
+<!--
+This ruleset ensures that Tor Browser Bundle users don't accidentally use a Tor2web node. 
+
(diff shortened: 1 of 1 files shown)


virgil:
Perhaps it's just my cluelessness, but it's unclear to make the patch:

https://github.com/EFForg/https-everywhere/pull/3034

to also be included in the travis-ci build for  8a50410

FWIW it works on my machine?

virgil:
Replaced by: https://github.com/EFForg/https-everywhere/pull/3035

PR #3034, opened 2015-10-03
Added Tor2web rules to downgrade-whitelist

virgil:
The needed whitelist change for: https://github.com/EFForg/https-everywhere/pull/3033

PR #3035, opened 2015-10-03
adding Tor2web.xml

virgil:
This ruleset ensures that Tor Browser Bundle users don't accidentally use a Tor2web node.       

* Wanted to include <target host="*.onion.to" /> but it wouldn't work due to a bug associated with the keyword "to".

* commit https://github.com/virgil/https-everywhere/commit/2d1104be3cae589a88b27f7ef58bd2a62d946c2f adds Tor2web to the downgrade whitelist

Perhaps it's just my cluelessness, but it's unclear to make patch https://github.com/EFForg/https-everywhere/pull/3034 also be included in the travis-ci build.  But when you combine them together it works.

### DIFF
--- a/src/chrome/content/rules/Tor2web.xml
+++ b/src/chrome/content/rules/Tor2web.xml
@@ -1,0 +1,21 @@
+<!-- This ruleset ensures that Tor Browser Bundle users don't accidentally use a Tor2web node. -->
+
+<ruleset name="Tor2web">
+  <target host="*.tor2web.com" />
+  <target host="*.tor2web.org" />
+
+  <target host="*.onion.cab" />
+  <target host="*.onion.city" />  
+  <target host="*.onion.direct" />  
+  <target host="*.onion.link" /> 
+  <target host="*.onion.nu" />
+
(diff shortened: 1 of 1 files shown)


youdly:
from the build log, it seems the ruleset still need two more test url similar to line 13-17, in addition to 4 test url for `*.tor2web.com`.

cypherpunk:
I think no Tor rules should be included in https-everywhere. Https-everywhere also used by users without Tor.
A better place to contribute this would be https://github.com/chris-barry/darkweb-everywhere.

chris-barry:
@cypherpunk darkweb-everywhere already has this. Most known tor2web instances will be re-written to go directly to the onion.

virgil:
Revised to: https://github.com/chris-barry/darkweb-everywhere/pull/43/files

How does one get Darkweb-everywhere into TBB by default?

semenko:
Closing, since looks like this is in darkweb-everywhere -- happy to reopen if we decide we should add Tor rules into HTTPSe.

PR #3037, opened 2015-10-04
[Syncplicity] New ruleset

youdly:


fuglede:
Looks good to me; except from the dev and stating subdomains, all included targets already redirect to HTTPS, and some of them specify HSTS. You could also add syncplicity.com as a target and just redirect it to www.syncplicity.com directly.

Do you want to [rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) this one to the latest commit? That will make it a bit easier for users who need to read the log on <code>SwitchVPN.xml</code> in the future.

youdly:
Even though some subdomains already specify HSTS, I think ruleset for those subdomains is still necessary.
`https://syncplicity.com` works now. Squashed the commits.

fuglede:
Thanks for the update. The squashing seems to have not been done entirely. The file <code>SwitchVPN.xml</code> shows up in one commit, then disappears again in your latest one, which is kind of unfortunate. If you <code>pick</code> the latest commit and <code>squash</code> the others, this should not happen.

youdly:
Squashed the `SwitchVPN.xml` commit.

fuglede:
👍 Thanks.

PR #3042, opened 2015-10-05
Add Buycott.com

chris-barry:


fuglede:
On campaign pages such as https://buycott.com/campaign/1250/boycott-dupont one encounters false mixed content blocking where Twitter's bootstrap JS and Google's recaptcha is blocked. Is this something we care about? As far as I can tell, it doesn't really make any difference.

chris-barry:
The bootstrap js thing seems to be a bug. I've sent them an email and will report back. It does not seem to be causing any issues on the page though.

PR #3043, opened 2015-10-05
Update SANS.org.xml

anand-bhat:


fuglede:
Unfortunately it's hard to test pre-software-security.sans.org without being authorized but otherwise it seems good; thanks for the update!

PR #3047, opened 2015-10-05
[TorrentShack] New ruleset

lucaslampe:
Adding a ruleset for the TorrentShack private tracker.


fuglede:
Thanks for the ruleset! It's a bit hard to verify coverage without a user but it seems good from the looks of it.

PR #3049, opened 2015-10-07
Make trivial rule for redwoodsinyosemite.com

jish:


fuglede:
At least one font is blocked by false mixed content blocking; this is not really a big problem.

On other pages, such as https://www.redwoodsinyosemite.com/property-info-2/17817.html (click "calendar"), mixed content blocking causes [some scripts](http://www.resortpro.net/pmt_common/js/calendar/calendar.js) to not load. In this particular case, the relevant scripts can be retrieved over secure connections as well but would be blocked in Firefox and Chrome even then. It seems like this would break relevant page content, so the ruleset should probably at least have the <code>platform="mixedcontent"</code> attribute specified.

jish:
Oh, neat. I didn't know about that attribute. I've updated the pull request with the `platform` attribute set to `"mixedcontent"`.

fuglede:
Thanks for the update.

> I didn't know about that attribute.

In that case, I should elaborate a bit on why it makes sense here. So, http://www.resortpro.net/pmt_common/js/calendar/calendar.js is currently blocked, but one finds that the resortpro.net links [could actually be secured as well](https://github.com/EFForg/https-everywhere/pull/3100), so it would be natural to assume that if we just add a rule for resortpro.net, everything would work well for your ruleset as well. However, due to an [extremely unfortunate bug](https://bugzilla.mozilla.org/show_bug.cgi?id=878890) in both Firefox and Chromium, this is not generally the case. So, the idea with <code>platform="mixedcontent"</code> is that if either browser gets fixed, all the rulesets specifying that (and currently there are 2010 such rulesets) could be put into effect.

jish:
Cool, thanks for the info :)

PR #3051, opened 2015-10-07
Support for engineering.riotgames.com.

KeenRivals:
Created a ruleset for engineering.riotgames.com. Other riotgames domains don't seem to support https.


fuglede:
Thanks for the ruleset; looks good to me!

PR #3055, opened 2015-10-08
Updated HTTPS rule for HollywoodReporter.com

jazwiecki:
We moved The Hollywood Reporter from Akamai to Edgecast a year ago and it has fully SSL support.

### DIFF
--- a/src/chrome/content/rules/Hollywood-Reporter.xml
+++ b/src/chrome/content/rules/Hollywood-Reporter.xml
@@ -33,15 +33,6 @@
 
 	<target host="hollywoodreporter.com" />
 	<target host="*.hollywoodreporter.com" />
-		<!--
-			Avoid user-visible paths:
-							-->
-		<exclusion pattern="^http://(?:www\.)?hollywoodreporter\.com/(?!favicon\.ico|sites/)" />
-		<!--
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Hollywood-Reporter.xml
```diff
@@ -37,26 +33,17 @@
...
+    <test url="http://www.hollywoodreporter.com/" />
+    <test url="http://subscribe.hollywoodreporter.com/" />
 
-	<rule from="^http://(?:www\.)?hollywoodreporter\.com/"
-		to="https://a248.e.akamai.net/f/1682/7219/8m/www.hollywoodreporter.com/" />
 
-	<rule from="^http://m\.hollywoodreporter\.com/mobify/"
-		to="https://hollywoodreporter.mobify.com/mobify/" />
+	<rule from="^http://(?:www\.)?hollywoodreporter\.com/"
+		to="https://www.hollywoodreporter.com/" />
```
loveithateit: ^hollywoodreporter.com seems to work. We prefer not to 'snap' redirects if possible, since doing so is both more complex and more distinguishing.

loveithateit:
Thanks for this. Site devs contributing rulesets is always nice to see.

PR #3060, opened 2015-10-11
Update ruleset for T411

laomaiweng:
T411 has migrated to t411.in.
t411.io now redirects there.


fuglede:
Thanks for the update to the ruleset! It appears to be identical to that of #2988 though, so the update should be taken from that instead.

laomaiweng:
Whoops, hadn't seen that PR, sorry for the noise!

PR #3065, opened 2015-10-13
Update Blendle ruleset

koenrh:
This removes the exclusion for the UserVoice sub domain, because it's no longer available. It also extends the `securecookie` rule to include all sub domains.

### DIFF
--- a/src/chrome/content/rules/Blendle.xml
+++ b/src/chrome/content/rules/Blendle.xml
@@ -19,7 +19,7 @@
 
   <test url="http://blog.blendle.nl/" />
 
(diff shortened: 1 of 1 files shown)


fuglede:
Thanks for the update. The securecookie change is actually not quite optimal, since it means that those cookies whose host is set to "blendle.com" exactly will not be matched. You can cover both cases by letting the host be, for instance, <code>^\\.?blendle\\.com</code>.

koenrh:
You're right, it should match the bare domain as well. Fixed it!

fuglede:
👍

PR #3070, opened 2015-10-14
Changes to ruleset validation output

fuglede:
This achieves a couple of different things:
- Previously, when some rulesets contained errors, xmllint would be run twice when running it only once was sufficient. This thus saves a fair amount of time.
- Previously, the end of the script would never actually be executed since xmllint would break it. In particular, the obtained exit status might not have been the expected one.
- Echo errors to stderr instead of stdout.
- Most importantly: Give some additional output when validation failure is due to one of the two most common ruleset errors, where the existing error output was not particularly helpful.

### DIFF
--- a/utils/validate.sh
+++ b/utils/validate.sh
@@ -33,19 +33,26 @@ fi
 type xmllint >/dev/null || die "xmllint not available"
 
 GRAMMAR="relaxng.xml"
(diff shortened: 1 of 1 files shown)


semenko:
Nice work -- thanks!

PR #1946, opened 2015-06-17
New rule for dict.cc

KommX:

PR #3071, opened 2015-10-14
added dict.cc

J0WI:
rebased #1692 / #1946 
Couldn't reproduce the problems on options.php.


J0WI:
Very popular domain in Germany (cc @fuglede)

tomwys:
It seams that they have wildcard domain configuration because random domains such as https://asdfasdfsad.dict.cc/ work. So maybe it is the case we should use *.dict.cc rather than enumerating domains? @fuglede any opinion on that topic?

fuglede:
I don't necessarily disagree entirely. However, there are already quite a few subdomains which would have to be excluded, so it boils down to the following: Do we expect more subdomains to arise which would then be automatically broken by not being excluded? And sort-of-conversely: do we expect more language-style subdomains to arise which would then have to be maintained individually? Both are pains, and we want the least of them.

J0WI:
It's harder to maintain wildcard rules with wildcard exclusions, so I decided to support just the common domains.

fuglede:
Some issues:
- The ruleset redirects http://dict.cc to https://dict.cc which serves a bad cert.
- https://www9.dict.cc [serves an incomplete chain](https://www.ssllabs.com/ssltest/analyze.html?d=www9.dict.cc&latest) and should be removed from the ruleset.
- pocket.dict.cc might be securable.
- you could add contribute.dict.cc to the list of bad subdomains.
- if one goes to a non-secured page, say http://contribute.dict.cc, and presses "Optionen", then [things break](https://www.dict.cc/dict/options.php?ref=http%3A%2F%2Fcontribute.dict.cc%2F%3Faction%3D%26id%3D%26p%3D%26goback%3D%26f%3D%26o%3D).

J0WI:
Thanks for the feedback. All points should be fixed now.

fuglede:
Thanks for the fixes. The first four points are fixed, but the last one persists: [this page](https://www.dict.cc/dict/options.php?ref=http%3A%2F%2Fcontribute.dict.cc%2F%3Faction%3D%26id%3D%26p%3D%26goback%3D%26f%3D%26o%3D) is not downgraded, for instance.

J0WI:
(updated)

PR #3072, opened 2015-10-15
[IRS.gov] New Ruleset

lucaspetter:


fuglede:
This does not appear to be working as intended: https://irs.gov time out; instead, http://irs.gov should redirect to https://www.irs.gov.

It is also not clear to me, if the rule for sa.www4.irs.gov is working; certainly http://sa.www4.irs.gov/ and https://sa.www4.irs.gov/ give different responses.

lucaspetter:
@fuglede Thanks for catching the https://irs.gov redirect error. I've added a redirect from http://irs.gov to https://www.irs.gov to fix it.

Regarding sa.www4.irs.gov having different responses over HTTP and HTTPS:
- Any HTTP address on sa.www4.irs.gov gives a 403 access denied error. It seems they are blocking all HTTP access instead of redirecting it to HTTPS. The server also doesn't use HSTS.
- I don't think IRS intends for anyone to access the domain's root https://sa.www4.irs.gov because no content is actually hosted there. Rather, there are links on their site that point to specific locations on that domain for different programs, eg. https://sa.www4.irs.gov/irfof-efp/start.do or https://sa.www4.irs.gov/irfof/lang/en/irfofgetstatus.jsp. I've added both these links as test URLs and can confirm the ruleset redirects them to HTTPS properly.

fuglede:
Great, thanks. In the same spirit as with www4, it could be helpful to have URLs to describe the purpose of apps.irs.gov. Not a big deal, though; in particular not since its HTTP version currently serves a 301 redirect to HTTPS.

PR #3074, opened 2015-10-15
Add justseed.it

gricey432:


fuglede:
Looks good to me; thanks for the ruleset!

PR #3080, opened 2015-10-16
[Carezone.cz] New ruleset

Roman3349:
Added ruleset for webhosting Carezone.cz.


fuglede:
Thanks for the ruleset! Looks good to me.

Note that you don't actually need to add test hosts in a simple rule like this; the target hosts will be automatically included. Also, it looks like you could also add roundcube.carezone.cz and squirrel.carezone.cz to the list.

Issue #2263, opened 2015-07-18
Add GoExplore.net


rugk:
Please add [GoExplore.net] (https://www.goexplore.net), a kind of news site from [ESET] (https://www.eset.com) (which already has a [ruleset] (https://www.eff.org/https-everywhere/atlas/domains/eset.com.html)).

Interestingly this only causes mixed content warnings because even if it's accessed via https it still loads some assets via http.

PR #3083, opened 2015-10-17
[GoExplore.net] Disable ruleset

rugk:
HTTPS connections are now redirected to HTTP... Unfortunately.

### DIFF
--- a/src/chrome/content/rules/GoExplore.net.xml
+++ b/src/chrome/content/rules/GoExplore.net.xml
@@ -8,7 +8,7 @@
 		- ESET_static.com.xml
 		- Sicontact.at.xml
 -->
-<ruleset name="GoExplore.net">
+<ruleset name="GoExplore.net" default_off="http redirect">
 	<target host=       "goexplore.net" />
 	<target host=   "www.goexplore.net" />
 	<target host="static.goexplore.net" />
--- a/src/chrome/content/rules/GoExplore.net.xml
+++ b/src/chrome/content/rules/GoExplore.net.xml
@@ -8,10 +8,16 @@
 		- ESET_static.com.xml
 		- Sicontact.at.xml
 -->
-<ruleset name="GoExplore.net" default_off="http redirect">
+<ruleset name="GoExplore.net">
 	<target host=       "goexplore.net" />
 	<target host=   "www.goexplore.net" />
 	<target host="static.goexplore.net" />
+	
+	<!--
+	Exclusion for buggy URL (which sometimes redirects to HTTP), more information: https://github.com/EFForg/https-everywhere/pull/3083
+	https://regex101.com/r/vC6jX8/1
+	-->
+	<exclusion pattern="^http:\/\/www\.goexplore\.net\/they-did-what\/payments-evolution-sheep-to-smartphones" />
 
   	<securecookie host="^www\.goexplore\.net$" name=".+" />
 
--- a/src/chrome/content/rules/GoExplore.net.xml
+++ b/src/chrome/content/rules/GoExplore.net.xml
@@ -18,6 +18,7 @@
 	https://regex101.com/r/vC6jX8/1
 	-->
 	<exclusion pattern="^http:\/\/www\.goexplore\.net\/they-did-what\/payments-evolution-sheep-to-smartphones" />
+	<test url="http://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones" />
 
   	<securecookie host="^www\.goexplore\.net$" name=".+" />
 


fuglede:
I can't reproduce; seems to work fine.

rugk:
I've just tested it with my browser and indeed now it seems to work on the main pages and most other... Strange...

But let's try this page: http://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones ([HTTPS link](https://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones))

fuglede:
Thanks for checking again. Your last link works here as well though;

    $ curl -I https://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones 
    HTTP/1.1 200 OK

Could you double check? So their site seems pretty indecisive. Perhaps we're hitting different nodes on a CDN, or something like that?

    $ dig +short www.goexplore.net
    91.228.166.80

rugk:
At first: BTW `https://static.goexplore.net` redirects to the homepage, so that's okay. E.g. https://static.goexplore.net/wp-content/uploads/2015/09/black-mirror-episodes-come-true-2-150x150.png does not redirect.

>  Are we hitting the same server?

I hope so... :fearful: 
(nslookup returns `91.228.166.80`)

## So let's test...
OS: Windows 7

### Firefox 41.0.2
redirection
--> **301 redirect**

### Chromium 48.0.2538.0 (64-bit)
chromium
--> **200, uses HTTPS**

### IE 11

ie
--> **301 redirect**

### curl 7.45.0
```
>curl -I https://www.goexplore.net/they-did-what/payments-evolution-sheep-to-sma
rtphones
HTTP/1.1 200 OK
Server: Apache
Vary: Accept-Encoding,User-Agent
Last-Modified: Thu, 01 Jan 1970 00:00:00 GMT
Expires: Sun, 18 Oct 2015 14:53:38 GMT
Pragma: public
Cache-Control: max-age=86400, public, must-revalidate, proxy-revalidate
Etag: 8831fff31af3007ea05cace4a785419a
X-Powered-By: W3 Total Cache/0.9.4
X-Frame-Options: sameorigin
Content-Type: text/html
X-Cache-TTL: 86400.000
Date: Sat, 17 Oct 2015 14:53:50 GMT
X-Varnish: 1107732346 1107732345
Age: 11
Via: 1.1 varnish
X-VCache: HIT
X-VCache-Hits: 1
X-VarnishServer: lb-h1
```
--> **200, uses HTTPS**

## Summary
Software | redirects to HTTP
-------- | ----
Firefox | **yes**
IE | **yes**
Chrome | no
curl | no

fuglede:
Huh, if I go to the HTTP version of that page in Firefox 41, HTTPS-E triggers a redirect loop. If I go directly to the HTTPS version though, there's no problem.

rugk:
> Huh, if I go to the HTTP version of that page in Firefox 41, HTTPS-E triggers a redirect loop.

That's why I created this PR.

> If I go directly to the HTTPS version though, there's no problem.

In my case not. It always redirects to HTTP - even with HTTPS-E completely disabled.
And as it also happens in IE - which of course has no HTTPS-E extension (installed), so it does not seem to be caused by HTTPS-E.

fuglede:
Very peculiar. It does redirect in Chrome here as well though (Chrome/45.0.2454.85 Safari/537.36). Supplying a custom UA header with curl gives the same behaviour:

    $ URL=https://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones
    $ curl -I $URL
    HTTP/1.1 200 OK
    $ curl -I -A "Mozilla/5.0 (X11; Linux x86_64; rv:41.0) Gecko/20100101 Firefox/41.0" $URL
    $ curl -I -A "What on Earth is going on" $URL

rugk:
Okay I've made GoExplore and @eset [on Twitter](https://twitter.com/rugkme/status/655408391259598848) aware of this issue here.
As they know their server config they maybe know more about this.

rugk:
> Supplying a custom UA header with curl gives the same behaviour

Yes you're right. I can confirm this.

Also when I submitted the UA of my Chromium browser it does indeed return 200:
```
# curl -I -A "Mozilla/5.0 (Windows NT 6.1; rv:31.0) Gecko/20100101 Firefox/31.0"
 %URL%
HTTP/1.1 200 OK
```
And now I also see why it does so: This UA is of course spoofed and does not belong to Chromium.
Using my real UA it also returns 301:
```
# curl -I -A "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML
, like Gecko) Chrome/48.0.2538.0 Safari/537.36" %URL%
```

So the issue seems to be discovered: GoExplore redirects to HTTP (at least on this site) based on the User Agent.
I also tested some other GoExplore sites, but I could find none which also has such a strange behaviour. Possibly they were concerned that their 'interactive infographic'.
But I don't see any reason why this should only break with HTTPS in some (or the majority of) browsers. Especially as I see no special resources loaded from third-party sites.

---

So what are we doing now? Certainly HTTPS-E does not support exclusions based on the user agent of the browser.
So the only thing I can think of would be to exclude `https://www.goexplore.net/they-did-what/payments-evolution-sheep-to-smartphones` completely. We just don't know whether there are more of this 'special redirecting sites'.
But maybe GoExplore replies to my tweet or here at GitHub. That would of course the best way.

fuglede:
Yep, let's put this on standby until further notice, or until the next release. If they don't fix it, I agree that over-exclusion is the way to go here.

rugk:
I have received a confirmation that the ESET staff is noticed.

Just a note about the mixed content errors (completely BTW here):
As I wrote when requesting this ruleset in https://github.com/EFForg/https-everywhere/issues/2263 all assets, which caused mixed content errors and I have noticed, are only loaded from goexplore.net, so the only thing which has to be done is make the link protocol-independent.

rugk:
New status: "developer informed", they are "working on it"

fuglede:
Thanks for the update; was that the only page that exhibited the strange behaviour? Seems a bit random.

rugk:
>was that the only page that exhibited the strange behaviour?

It was the only one I found. If you find more let me know. Indeed it seems a bit random.
The only special thing about the page was the fixed bar at the left, whcih is not there in other articles.

semenko:
This is a pretty random exclusion, but LGTM

rugk:
Okay, ESET has said in the new year they'll probably fix this, so we should keep an eye an it. (I'll do it)

PR #3086, opened 2015-10-17
[Getbootstrap.com] New ruleset

rugk:
CC: @twbs

### DIFF
--- a/src/chrome/content/rules/Getbootstrap.com.xml
+++ b/src/chrome/content/rules/Getbootstrap.com.xml
@@ -6,8 +6,9 @@
     <target host="www.getbootstrap.com" />
     <target host="expo.getbootstrap.com" />
     <target host="blog.getbootstrap.com" />
+    <target host="themes.getbootstrap.com" />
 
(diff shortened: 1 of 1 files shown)


fuglede:
Looks good to me. Did you try adding http://themes.getbootstrap.com? Seems to be possible; a single (securable) Google font is blocked, but that's probably not a big problem.

And some minor things: Some (securable) ads are blocked throughout. Not a huge deal. On https://blog.getbootstrap.com/ the (securable) GitHub buttons are blocked. Also not a huge deal.

rugk:
> Looks good to me. Did you try adding http://themes.getbootstrap.com? Seems to be possible; a single (securable) Google font is blocked, but that's probably not a big problem.

Oh, forgot this subdomain. And again this font does not seem to be important - it was also blocked by Privacy Badger. :smile: 

And as the other blocked things are securable at least in Tor Browser they should be redirected. But as you noticed they are also not important. As I pinged the bootstrap team in this PR, maybe they can even just change the URLs so this third-party content is secured.

PR #3087, opened 2015-10-17
[wisc.edu] Add multiple subdomains

austinhartzheim:
Add support for subdomains:
* cs.wisc.edu
* discovery.wisc.edu
* doit.wisc.edu
* go.wisc.edu
* housing.wisc.edu
* kb.wisc.edu
* library.wisc.edu
* math.wisc.edu

Updated rule documentation (old comments not valid for all domains). Further updated this documentation with new domains to explain why they are not yet supported (for example, mixed content warnings caused by Google site search).


semenko:
:fireworks: thanks! :fireworks:

PR #3089, opened 2015-10-17
[tetadrogerie.cz] Added ruleset

Roman3349:
Added ruleset for drugstore teta (tetadrogerie.cz).


fuglede:
Note that as in #3080, the test URLs aren't actually necessary here.

Roman3349:
Now it should be okay.

fuglede:
👍

PR #3097, opened 2015-10-18
[Telecolumbus] New ruleset

rugk:


src/chrome/content/rules/Telecolumbus.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+
+-->
+<ruleset name="Telecolumbus.com">
+    <target host="telecolumbus.com" />
+    <target host="www.telecolumbus.com" />
+    <target host="wohnungsunternehmen.telecolumbus.com" />
```
loveithateit: wohnungsunternehmen doesn't appear to reply from :443.

PR #3098, opened 2015-10-18
[Kabelkiosk.de] Rewritten, activated

rugk:


fuglede:
The ruleset validation test fails as some trailing slashes are missing. In any case, the original problem -- i.e. that the site delivers an incomplete certificate chain -- still persists, and the ruleset should not be enabled.

rugk:
It works in Firefox and Chrome, but indeed it seems a intermediate cert is missing... But is this really so bad to disable it completely?
root

I've reached out to the kabelkiosk.de and gave them a few tips what they should do, so maybe they'll fix this.

fuglede:
The browsers do not download the certificate automatically, so if it works for you, it is likely because you received it via some other site. In the case of Firefox, this stores the certificate as what the browser calls a "Software Security Device" in its certificate store.

If you run Firefox on a clean profile (which in my opinion makes writing rules a lot easier), you will get a warning page:


firefox-warning

Since this error is a little subtle, it might actually be a good idea to have the integration test check for it through https-everywhere-checker.

fuglede:
So let's close this for now. If they fix the chain (or if browsers change the way they handle certificate downloads), we can have a look at it again.

PR #3099, opened 2015-10-18
[An.de] New ruleset

rugk:
Note that it only covers pk.an.de. Should I possibly rename it to
`pk.an.de` because of this?

### DIFF
--- a/src/chrome/content/rules/An.de.xml
+++ b/src/chrome/content/rules/An.de.xml
@@ -1,0 +1,13 @@
+<!--
+    Non-functional subdomains:
+        - www
+		- forum
+-->
+<ruleset name="An.de">
+    <target host="pk.an.de" />
+
+    <securecookie host="^pk\.an\.de$" name=".+" />
+
+    <rule from="^http:"
+            to="https:" />
+</ruleset>


fuglede:
The filename is fine. And so's the ruleset; the HTTP version serves a 301 redirect to HTTPS. Merging.

rugk:
:+1: 

> And so's the ruleset; the HTTP version serves a 301 redirect to HTTPS.

How often do you want to say this? Read this in several PMs now. Do you already have a template/shortcut key for this? :wink:
Okay seriously: Should I already imply this in the PR description, so you don't have to add this any time? I mean is this a kind of convention to make say if it 301 redirects?

fuglede:
Nah, think of it more as review notes (which may or may not be helpful to people other than myself): if the site redirects to HTTPS itself, there's usually not very much to check. However, if any parts of the site do not themselves redirect to HTTPS, then I missed it in my review.

On a related note, there has been some talk about picking out rules like these which are 'easily HSTS-able' and to have HTTPS Everywhere treat those differently (which in particular could help remedying the memory consumption issue).

rugk:
Thanks for the information.

PR #3106, opened 2015-10-19
yplanapp supports wildcard TLS now

graingert:


src/chrome/content/rules/yplanapp.com.xml
```diff
@@ -1,11 +1,12 @@
 <ruleset name="YPlan">
 
-	<target host="yplanapp.com" />
-	<target host="www.yplanapp.com" />
+	<target host="*.yplanapp.com" />
 
-  <securecookie host="^(www\.)?yplanapp.com$" name=".+" />
+	<securecookie host="^.*\.yplanapp\.com$" name=".+" />
```
graingert: does this work with `https://yplanapp.com` as well?

rugk: No. ([just test it by yourself](https://regex101.com/r/jG4bG2/2))
The issue is just that dot `\.` is always required.
But you can e.g. [fix it like this](https://regex101.com/r/jG4bG2/3).

graingert: I got the example from` 	<securecookie host="^.*\.123-reg\.co\.uk$" name=".*" />` is that wrong too?

rugk:
The test currently fails, because you have to add test URLs for wildcard rulesets.
More information: https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md

src/chrome/content/rules/yplanapp.com.xml
rugk: It [does not](https://regex101.com/r/zG4kM9/1) match the domain without any subdomain (here: `https://123-reg.co.uk`). So if this is the aim of the ruleset it's indeed wrong.

graingert:
@rugk the tests pass now

rugk:
Nice, so you only have to wait until this is merged.

PR #3108, opened 2015-10-19
Update LegitScript rule

ryansouza:
We no longer use a separate hostname for ssl connections


fuglede:
Thanks for the update. It looks like the blog could be covered as well?

PR #3111, opened 2015-10-20
[Badssl.com] Rewrite ruleset

rugk:
Rewritten to:
* add test URLs
* exclude URLs which (potentially) manipulate tests
* convert to spaces

### DIFF
--- a/src/chrome/content/rules/Badssl.com.xml
+++ b/src/chrome/content/rules/Badssl.com.xml
@@ -1,6 +1,7 @@
 <!--
     Not secured subdomains:
-        - All subdomains are not secured, because this can influence tests and cause unexpected results.
+        - All subdomains except of the HTTPS Everywhere test are not secured, because this can
+          influence tests and cause unexpected results.
 -->
 <ruleset name="Badssl.com">
     <target host="badssl.com" />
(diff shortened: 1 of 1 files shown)


lgarron:
`badssl.com` maintainer here. If we're excluding any of `[*.]badssl.com`, I'd prefer to exclude all of it.
We're frequently adding and changing tests, and there is no benefit for selectively upgrading subdomains – only potential confusion for developers.

rugk:
So what do you suggest? To disable every subdomain and only upgrade `http://badssl.com`?

fuglede:
Note that the missing tests that Travis is complaining about should match the exclusion pattern; i.e. be mixed.badssl.com, hsts.badssl.com, etc.

rugk:
> Note that the missing tests that Travis is complaining about should match the exclusion pattern; i.e. be mixed.badssl.com, hsts.badssl.com, etc.

Yes, but know I've already used another way. :smiley:

rugk:
It would be nice if it could be merged now, because https://github.com/lgarron/badssl.com/issues/84 depends on this.

rugk:
Does someone want to merge this?

rugk:
Thanks. ☺️

PR #3118, opened 2015-10-21
[Pastemarkdown.com.xml] New ruleset

rugk:
Cloudflare

### DIFF
--- a/src/chrome/content/rules/Pastemarkdown.com.xml
+++ b/src/chrome/content/rules/Pastemarkdown.com.xml
@@ -1,7 +1,7 @@
 <!--
     Cloudflare SSL
 -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Pastemarkdown.com.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+    Cloudflare SSL
+-->
+<ruleset name="Pastemarkdown.com">
+    <target host="pastemarkdown.com" />
+    <target host="www.pastemarkdown.com" />
```
loveithateit: Connections are rejected for https://www

rugk: You're right. I removed it...

Issue #2875, opened 2015-09-12
Denic.de


rugk:
On Tuesday, 15 September, Denic will change it's denic.de domain to HTTPS. Currently you only can use *https://www.secure.denic.de*.
We should have a rule in place, which can be published at this time. 

https://www.secure.denic.de/en/denic-in-dialogue/news/4079.html

rugk:
BTW the change has now happened. Although they [could improve] their configuration at least they support it by default now.
But a HTTPS everywhere rule is still missing.

BTW *secure.denic* still works.

PR #3128, opened 2015-10-23
[Denic.de] New ruleset; [WebGo24.de] Rewrite

rugk:
Fixes https://github.com/EFForg/https-everywhere/issues/2875, 301
redirects to HTTPS

As Denic.de was mentioned in WebGo24.de I wanted to change this. However
WebGo24.de was a very bad wildcard ruleset without test URLs and without
any use of the wildcard. So I had to rewrite this too.

### DIFF
--- a/src/chrome/content/rules/WebGo24.de.xml
+++ b/src/chrome/content/rules/WebGo24.de.xml
@@ -33,9 +33,6 @@
 					-->
 	<!--securecookie host="^login\.webgo24\.de$" name=".+" /-->
 	
-	<test url="https://webgo24.de/" />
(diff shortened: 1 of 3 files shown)


src/chrome/content/rules/WebGo24.de.xml
```diff
@@ -25,20 +21,18 @@
 <ruleset name="WebGo24.de">
 
 	<target host="webgo24.de" />
-	<target host="*.webgo24.de" />
-
+	<target host="www.webgo24.de" />
+	<target host="support.webgo24.de" />
```
loveithateit: Cert doesn't match support.

PR #3137, opened 2015-10-26
Create solarizde.xml

solariz:


src/chrome/content/rules/solarizde.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="solariz.de">
+	<target host="solariz.de" />
+	<test url="http://solariz.de/" />
```
loveithateit: Missing www target, redundant test.

fuglede:
Could you add a <code>\<target host="www.solariz.de" /\></code> to this one? Otherwise that part of the rule does nothing (and it would also save your users a redirect as currently http://www.solariz.de redirects to http://solariz.de instead of https://solariz.de).

PR #3149, opened 2015-10-31
Add two sites

hoschiCZ:
Namely [stremio.io](https://stremio.io) and [www.gml.cz](https://www.gml.cz) .
Stremio is a young app for streaming video content from the internet & p2p networks.
GML is our school website.

Both have HTTPS, but the don't have it enabled by default.

### DIFF
--- a/src/chrome/content/rules/Gml.cz.xml
+++ b/src/chrome/content/rules/Gml.cz.xml
@@ -1,0 +1,7 @@
+<ruleset name="Gml.cz">
+        <target host="www.gml.cz" />
+
+        <rule from="^http:"
+                to="https:" />
+</ruleset>
+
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Gml.cz.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Gml.cz">
+        <target host="www.gml.cz" />
```
loveithateit: This is fine

src/chrome/content/rules/Stremio.io.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Stremio.io">
+        <target host="stremio.cz" />
```
loveithateit: stremio.cz doesn't exist, stremio.io is parked. Did you mean to target strem.io?

PR #3156, opened 2015-11-02
[bitmask.net] cover all but deb in *.bitmask.net

varac:
https://deb.bitmask.net uses a self-signed cert which CN is not matching
the subdomain.

### DIFF
--- a/src/chrome/content/rules/Bitmask.net.xml
+++ b/src/chrome/content/rules/Bitmask.net.xml
@@ -18,9 +18,15 @@
 	<!--	Direct rewrites:
 				-->
 	<target host="bitmask.net" />
-	<target host="demo.bitmask.net" />
-	<target host="dl.bitmask.net" />
+	<target host="*.bitmask.net" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Bitmask.net.xml
```diff
@@ -18,9 +18,15 @@
 	<!--	Direct rewrites:
 				-->
 	<target host="bitmask.net" />
-	<target host="demo.bitmask.net" />
-	<target host="dl.bitmask.net" />
+	<target host="*.bitmask.net" />
```
loveithateit: deb wasn't targeted before this change, so you're introducing a problem and fixing it.

Would a better description for your intention be "cover all but deb in *.bitmask.net"? If so, are you reasonably certain there aren't any other problematic hosts?

varac:
@loveithateit i changed the desciption as you proposed, and yes, i'm certain that there are no other problematic hosts.

PR #3174, opened 2015-11-08
[DataSavesLives] New ruleset

fuglede:
Some mixed content here and there; a Google font is blocked but the site works fine.


J0WI:
lgtm

Issue #2430, opened 2015-08-05
BROKEN RULESET (Scene7)


terrorist96:
https://www.verizonwireless.com/wcms/consumer/products/verizon-messages.html

Scene7 breaks video at bottom.

terrorist96:
You have to be logged in to see this, but I assure you, Scene7 breaks the video.

terrorist96:
Can this be merged?
@fuglede

terrorist96:
Verizon Wireless along with Scene7 are now breaking the video here:
https://www.verizonwireless.com/wcms/myverizon/smart-rewards.html

No login required to reproduce the issue. And I doubt a default=off rule for verizonwireless.com is going to be merged. Other than that, I don't know how to fix this...

fuglede:
Thanks for the fix. You mean [this video](https://www.verizonwireless.com/wcms/overlays/tips-video.html), right? If so, then you're right that there's no reason to turn off the entire ruleset. In my testing, adding a

    <exclusion pattern="http://s7mbrstream\.scene7\.com/" />

to <code>Scene7.xml</code> does the job.

terrorist96:
Thanks, I made a PR with that code:
https://github.com/EFForg/https-everywhere/pull/3177

fuglede:
Thanks, I'll comment there. Maybe I should add for the record that I didn't pull that exclusion pattern out of the air: if you open the browser console (F12 in Firefox) with the broken ruleset active, then you'll see a bunch of errors relating to s7mbrstream.scene7.com when loading the video.

terrorist96:
@fuglede With the new version out, the videos on these pages still don't load. Verizon Wireless and Scene7 have to be disabled, and only then will they load.
https://www.verizonwireless.com/wcms/consumer/products/verizon-messages.html
https://www.verizonwireless.com/wcms/myverizon/smart-rewards.html

fuglede:
Thanks for the update; is there any pattern of problematic pages or is it just those two?

terrorist96:
Not sure. Those are the two that I know of. I haven't found anything else that is broken, but that doesn't mean they don't exist. :)

semenko:
Fixed in #3630

PR #3181, opened 2015-11-10
[Scene7] Update ruleset; disable wildcard rule

fuglede:
This fixes at least two different issues. One is GH issue #2430, in which the ruleset would cause embedded videos at
  https://www.verizonwireless.com/wcms/overlays/tips-video.html
to not load.

Another is a flawed redirect for http://s7sps3.scene7.com/.

Note that at the moment, the www1 subdomain appears to serve nothing.

### DIFF
--- a/src/chrome/content/rules/Scene7.xml
+++ b/src/chrome/content/rules/Scene7.xml
@@ -9,9 +9,10 @@
 
 	Problematic subdomains:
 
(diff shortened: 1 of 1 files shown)


J0WI:
You may also update/remove the ```Fully covered subdomains``` comment

semenko:
lgtm as is, thanks!

PR #3184, opened 2015-11-10
[New ruleset] PHPDoc

rugk:
disabled because expired cert; [I've asked](https://twitter.com/rugkme/status/664158508808581121) what PhpDoc wants to do about this.

### DIFF
--- a/src/chrome/content/rules/Phpdoc.org.xml
+++ b/src/chrome/content/rules/Phpdoc.org.xml
@@ -1,0 +1,12 @@
+<!--
+
+-->
+<ruleset name="PhpDocumentor" default_off="expired cert">
+    <target host="phpdoc.org" />
+    <target host="www.phpdoc.org" />
+
+    <!-- <securecookie host="^(www\.)?phpdoc\.org/$" name=".+" /> -->
+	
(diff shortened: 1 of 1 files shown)


fuglede:
@jsha is wary of adding default_off rules that will only cause bloat, but, given [their answer on Twitter](https://twitter.com/phpDocumentor/status/664327932056240128), let's give this one a bit of time.

rugk:
IMO this does not cause bloat because some users (like I would) enable the rule and just add an exception to the rule.

But let's not discuss this now, because you're right we can wat now.

fuglede:
Hm, okay, looks like they replaced their self-signed cert with another self-signed one, so that didn't help all that much from the point of view of https-e.

rugk:
Okay... https://twitter.com/rugkme/status/671745557371731969

semenko:
Bleh, I hate adding default_off rules, though you claim you'll use it ...

rugk:
Well... I still hope they will switch to a CA soon.

rugk:
Finally they announced they'll do. :smiley: 
https://twitter.com/phpDocumentor/status/677591475354476544

PR #3187, opened 2015-11-11
[Codeclimate] New ruleset

rugk:


fuglede:
Looks like <code>.org</code> should have been <code>.com</code> in two places.

rugk:
Yes, in fact I forgot this.

PR #3183, opened 2015-11-10
[Radware.com] New ruleset

rugk:

PR #3189, opened 2015-11-11
[Radware.com] New ruleset

rugk:
Followup of https://github.com/EFForg/https-everywhere/pull/3183

### DIFF
--- a/src/chrome/content/rules/Radware.com.xml
+++ b/src/chrome/content/rules/Radware.com.xml
@@ -1,5 +1,6 @@
 <!--
-
+    Nonfunctional subdomains:
+        - $      -> times out
 -->
 <ruleset name="Radware.com">
 	<target host="radware.com" />
@@ -9,6 +10,9 @@
 	<target host="portals.radware.com" />
 
 	<securecookie host="^(www\.|blog\.|kb\.|portals\.|)?radware\.com$" name=".+" />
+
+	<rule from="^http://radware.com"
+		to="https://www.radware.com" />
 
 	<rule from="^http:"
 		to="https:" />
--- a/src/chrome/content/rules/Radware.com.xml
+++ b/src/chrome/content/rules/Radware.com.xml
@@ -1,0 +1,15 @@
+<!--
+
+-->
+<ruleset name="Radware.com">
+	<target host="radware.com" />
+	<target host="www.radware.com" />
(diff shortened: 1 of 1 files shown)


fuglede:
More or less looks good to me. Only problem would appear to be that https://radware.com times out (so it should probably just be redirected to www.radware.com instead).

rugk:
> Only problem would appear to be that https://radware.com times out (so it should probably just be redirected to www.radware.com instead) 

Would be possible, for sure.

However to fix that
> http://portals.radware.com/Not-Logged-In/Login/ triggers a redirect loop

I think it would be much more better if this is fixed by Radware itself. Especially as this is a login page and they are a company in the security area...
[I've notified them](https://twitter.com/rugkme/status/668881978306011137).

What's worse is that some of their servers used do not support HTTPS at all [according to SSLLabs](https://www.ssllabs.com/ssltest/analyze.html?d=radware.com).

rugk:
They seem to have fixed http://portals.radware.com/Not-Logged-In/Login/ and I now also added the redirection to www.

fuglede:
Thanks for the update; you need to add trailing slashes to the <code>to</code> and <code>from</code> attributes in the rule (in order to not have the rule trigger on a different hypothetical host that happens to also being with <code>radware.com</code>).

semenko:
@rugk I'd love to merge this -- would you please escape those special characters to fix the build error? Thanks! :+1:

rugk:
> Missing caret (^) in 'from' attribute: it should be "^http:" and not "http:".

That's not true. There is a caret in both.

> Missing trailing slashes in 'from' or 'to' when specifying full hostnames

Also not true.

There is a validation error, but I still do not know why it fails here.

fuglede:
It fails because of the backslashes in the <code>to</code>-attribute.

rugk:
This syntax is quite confusing. You have to use RegExp (escaping) in the `from` part but *must not* use it in the `to`part...

fuglede:
I don't disagree; it might be worth adding as one of the 'common reasons' for failure.

semenko:
Yeah, it's pretty silly. We should probably rethink rulesets as we move forward.

I'd also like to make the "name" attribute of cookie rules optional (most are .* or .+)

rugk:
Okay, so back to this ruleset: The tests seem to succeed.

PR #3194, opened 2015-11-12
Removed target host www.rzuser.uni-heidelberg.de because it breaks so…

KommX:
…me sites, e.g. http://www.rzuser.uni-heidelberg.de/~hu020/EWTS/index.html

### DIFF
--- a/src/chrome/content/rules/Uni-heidelberg.de.xml
+++ b/src/chrome/content/rules/Uni-heidelberg.de.xml
@@ -6,7 +6,6 @@
   <target host="katalog.ub.uni-heidelberg.de" />
   <target host="www.mathinf.uni-heidelberg.de" />
   <target host="www.math.uni-heidelberg.de" />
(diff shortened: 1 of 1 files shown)


fuglede:
Thanks for the report. Which other sites are broken? (Is there a pattern to it that could be excluded?)

KommX:
I don't know if other sites are affected but I guess so. This server hosts many different sites from university people.
Normally, you should get a list of sites at http://www.rzuser.uni-heidelberg.de/list/rzuser.html but that doesn't work obviously.

fuglede:
Yeah, looks like a problem with http://whitepages.urz.uni-heidelberg.de -- unfortunate. I also agree that excluding the domain is probably the safest bet. In any case, do you want to add a comment on the removal in the ruleset file? That makes it a bit easier to navigate for future ruleset editors.

KommX:
Done.

fuglede:
Thanks.

PR #3196, opened 2015-11-12
[QRpedia.org] New ruleset

rugk:
Causes mixed content errors; disabled by default because of missing chain.

I've contacted the site authors - maybe they'll fix it: https://twitter.com/rugkme/status/664840438503899136


semenko:
LGTM, broken on https

PR #3202, opened 2015-11-14
Updated Unity.xml

mercurytw:
Extended regex to match "docs.unity3d.com"


J0WI:
You should add some test urls or remove regex and wildcards.

semenko:

PR #3204, opened 2015-11-15
[Symbolab.com] New ruleset

rugk:
301 redirects to HTTPS


fuglede:
Looks like .org should have been .com in this one?

rugk:
No,
* https://www.symbolab.com/ works
* https://symbolab.com works

rugk:
This way it is right. :smile:

PR #3210, opened 2015-11-15
[MyDealz.de] New ruleset

rugk:
Cloudflare SSL

Issues are [reported](http://www.mydealz.de/bugreports/https-fehler-633128)

### DIFF
--- a/src/chrome/content/rules/Mydealz.de.xml
+++ b/src/chrome/content/rules/Mydealz.de.xml
@@ -1,0 +1,18 @@
+<!--
+    Cloudflare SSL
+
+    - Mixed content:
+	    - mydealz.de
+		    - images *
+	
+	* Secured by us
+-->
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks for reporting the errors!

WFM:
https://hukd.mydealz.de/

Mismatch:
https://visit.hukd.mydealz.de/
https://gutscheine.mydealz.de/
https://links.mydealz.de/
https://live.mydealz.de/

Most of them are to track clicks, but having them secure would be great anyway.

semenko:
I don't really want to add default_off domains -- they clutter up things and add some technical debt.

Closing for now -- feel free to re-open if they fix their TLS.

rugk:
Okay, if I should not notice if they do so pinging me (or replying to this PR) should be enough to make me aware of it.

PR #3212, opened 2015-11-16
Trivial rule for careerfoundry.com

ErinCall:


ErinCall:
#3214 should resolve the CI failure.

ErinCall:
CI passes now; this is ready to go :)

PR #3216, opened 2015-11-17
[MyMoustache.net] New ruleset

youdly:
New ruleset and update comment in existing rulesets.


src/chrome/content/rules/MyMoustache.net.xml
```diff
@@ -0,0 +1,17 @@
...
+
+<ruleset name="MyMoustache.net">
+	<target host=    "mymoustache.net" />
+	<target host="www.mymoustache.net" />
+
+  	<securecookie host="^.*\.mymoustache\.net$" name=".+" />
+
+	<!-- connection timeout -->
+	<rule from="^http://mymoustache\.net/"
+		to="https://www\.mymoustache.net/" />
```
fuglede: \\. should be . here.

PR #3219, opened 2015-11-17
Create Img-9gag-fun.9cache.com.xml

intchloe:
The 9gag image CDN supports HTTPS wihtout problem.


fuglede:
Thanks for the update. Could you add this to <code>9gag.xml</code> instead? That ruleset already has some rules for 9cache.com.

semenko:
closing in favor of #3342

PR #3239, opened 2015-11-20
add slimerjs wildcard TLS

graingert:


fuglede:
More or less looks good to me. You will probably want to add a <code>\<target host="slimerjs.org" /\></code> though, as that is not covered by the wildcard target.

graingert:
@fuglede are you sure? Doesn't this test https://github.com/EFForg/https-everywhere/pull/3239/files#diff-7aa40122d814f704ff980cf613e0fcb7R15 prove that it's covered?

fuglede:
Yeah, the test url itself does not ensure coverage; it needs to be caught by one of the targets.

PR #3246, opened 2015-11-21
[Oculus.com] Update ruleset

youdly:
Rename file and update domains.


loveithateit:
@youdly How do you view history prior to renames? I've tried $SHA1^ and $SHA1@{n}, but neither seem to work. I guess you could take the first revision, find it in full history, from that take the one prior and do log $(SHA1-1) -- $previous_filename, but that is a bit awkward, and takes a while if it's not already cached.

If there isn't an easy way to do so, please don't rename files unless it's absolutely necessary.

Issue #3234, opened 2015-11-19
https://microsoft.com/ uses a wildcard subdomain cert


nealey:
If you go to http://microsoft.com/, something other than their web server (I am guessing HTTPS everywhere, I can't reproduce it, unfortunately, even in incognito mode) redirects to https://microsoft.com/, which uses a cert issued for *.microsoft.com: this triggers a security message in Chrome and FF.

fuglede:
Yep, that's a problem with HTTPS Everywhere. You can reproduce it by disabling the ruleset in the dropdown menu and going to http://microsoft.com. It looks like this could be fixed by redirecting instead http://microsoft.com to https://www.microsoft.com.

PR #3252, opened 2015-11-21
[Microsoft] Redirect microsoft.com to www.microsoft.com

fuglede:
https://microsoft.com currently serves a bad certificate, its CN only covering *.microsoft.com. This fixes https://github.com/EFForg/https-everywhere/issues/3234.


J0WI:
lgtm

PR #3259, opened 2015-11-22
Create adiirc.com.xml

itsdf:


J0WI:
You should also add https://dev.adiirc.com/ and a comment, that the cert doesn't match https://forum.adiirc.com/


Same for the other host:

Mixed content on:
https://blogg.sydsvenskan.se/
https://apps.sydsvenskan.se/

Refused:
https://sydsvenskan.se/ <= You can fix that with a redirect from ^ to www
https://stjarnklubb.sydsvenskan.se/

semenko:
Closing in favor of #3262

PR #3263, opened 2015-11-23
Add lib.rus.ec

galeksandrp:


fuglede:
The host <code>www.lib.rus.ec</code> does not currently serve anything;

	$ curl -I https://www.lib.rus.ec
	curl: (6) Could not resolve host: www.lib.rus.ec
	$ curl -I http://www.lib.rus.ec
	curl: (6) Could not resolve host: www.lib.rus.ec

PR #3269, opened 2015-11-23
Create hd.se.xml

itsdf:


herrniemand:
@itsdf 
['src/chrome/content/rules/HD.se.xml', 'src/chrome/content/rules/hd.se.xml'] failed case-insensitivity testing.

herrniemand:
@itsdf Why dont you rename it to `HelsingborgsDagblad.se.xml`?

herrniemand:
@itsdf actually, there already Helsingborgs Dagblad
https://github.com/itsdf/https-everywhere/blob/719ca209be62bc53ef25d78cbd37bff406ae96c8/src/chrome/content/rules/HD.se.xml

itsdf:
mb didn't know that hd.se already existed!

PR #3270, opened 2015-11-23
Fix amara login

galeksandrp:
Login only saved for non-www domain, therefore it is a pain to manually
add www to access logged version of pages.


semenko:
Seems ok -- thanks!

PR #3273, opened 2015-11-24
[Globalclimatemarch.org] New ruleset

rugk:


semenko:
```

../src/chrome/content/rules/Globalclimatemarch.org.xml:15: element rule: Relax-NG validity error : Element rule failed to validate attributes
../src/chrome/content/rules/Globalclimatemarch.org.xml fails to validate
```

rugk:
What? In line 15 there is just the redirect rule.

rugk:
Does anybody know why the test fails here?

semenko:
You need to escape the from rule. At the bottom of travis:

    failure: Globalclimatemarch.org.xml failed test: The 'from' rule contains unescaped period in regular expression.  Try escaping it with a backslash.

rugk:
Ah thanks. Found it.

fuglede:
You missed the one after www.

rugk:
Thanks. Fixed. And as you somewhere else said I have to add `/` at the end I also did it here.

PR #3279, opened 2015-11-25
Add rkn.gov.ru

galeksandrp:


loveithateit:
Server sends incomplete or no certificate chain, please add default_off="missing certificate chain".

PR #3282, opened 2015-11-25
[Idenyt.dk] New ruleset

fuglede:
Some Emediate scripts (secured by us) are blocked as mixed content in Firefox/Chrome but this has no effect on site function.


J0WI:
timeout on:
- foto
- jul2k8
- (www.)tv

PR #3325, opened 2015-11-26
Readd Nginx.org

J0WI:


fuglede:
There's a typo in the ruleset comment, and as in the old ruleset, https://wiki.nginx.org is a thing but one that doesn't quite work. Otherwise it looks good to me; merging.

PR #3326, opened 2015-11-26
Delete ServerBeach.com.xml

J0WI:
subdomain dropped, host still does not support HTTPS


fuglede:
Even though the ruleset itself is certainly useless, I could see the comments existing providing some value so I hesitate to support deleting the entire thing. That being said, having a bunch of worthless rulesets around is of course not very good for performance.

PR #3342, opened 2015-11-30
Update 9gag.xml

intchloe:
Add Img-9gag-fun.9cache.com from #3219

### DIFF
--- a/src/chrome/content/rules/9gag.xml
+++ b/src/chrome/content/rules/9gag.xml
@@ -53,4 +53,7 @@
 	<rule from="^http://assets-9gag-lol\.9cache\.com/"
 		to="https://assets-9gag-lol.9cache.com/" />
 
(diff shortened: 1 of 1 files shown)


youdly:
failed ruleset due to lack of test url `<test url="http://example.com" />`. Instead of adding test URLs, I suggest cleaning up this ruleset by specifying each supported subdomain to `<target host>` (assuming not all subdomain has HTTPS), and then  a single rewrite "^http:" to "^https:" to conform to the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md). [example](https://github.com/youdly/https-everywhere/blob/comparitech/src/chrome/content/rules/Comparitech.com.xml) ruleset (notice line 3-4 & 9-10).

semenko:

PR #3349, opened 2015-12-02
Added subdomains to qt.io

quazgar:
Tested & works for me.


fuglede:
Looks good to me; thanks.

PR #3350, opened 2015-12-02
Fixed Colorado_Secretary_of_State.xml regexp

Vanderdecken:
Using a non-selecting group for the www. broke the result and users were sent to https://$1sos.state.co.us/pubs/info_center/holidayCalendar.html - as there was no $1 selected to be re-inserted. I barely know regexp but brief testing with https://regex101.com/ suggests my change works - but it needs some oversight.

### DIFF
--- a/src/chrome/content/rules/Colorado_Secretary_of_State.xml
+++ b/src/chrome/content/rules/Colorado_Secretary_of_State.xml
@@ -9,7 +9,7 @@
 	<target host="www.sos.state.co.us" />
 
 
-	<rule from="^http://(?:www\.)sos\.state\.co\.us/pubs/"
+	<rule from="^http://(www\.)sos\.state\.co\.us/pubs/"
 		to="https://$1sos.state.co.us/pubs/" />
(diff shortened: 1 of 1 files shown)


semenko:
```
+python2.7 test/rules/src/https_everywhere_checker/check_rules.py test/rules/coverage.checker.config
ERROR src/chrome/content/rules/Colorado_Secretary_of_State.xml: No rule or exclusion applies to test URL http://sos.state.co.us/
ERROR src/chrome/content/rules/Colorado_Secretary_of_State.xml: No rule or exclusion applies to test URL http://www.sos.state.co.us/
ERROR src/chrome/content/rules/Colorado_Secretary_of_State.xml: Not enough tests (0 vs 1) for <Rule from '^http://(www\.)sos\.state\.co\.us/pubs/' to 'https://\g<1>sos.state.co.us/pubs/'>
+echo 'Ruleset test coverage was insufficient.'
+echo ''
```

semenko:

PR #3362, opened 2015-12-04
Create Meetic.pt.xml

ekaris:


J0WI:
You should also add https://touch.meetic.pt/

ekaris:
It is good with my last patch?

J0WI:
lgtm
cc @loveithateit

PR #3372, opened 2015-12-05
[Sigdet.nu] [Dein.dk] New rulesets

fuglede:
This adds two related hosts whose certificates are signed by Let's Encrypt.


J0WI:
lgtm

PR #3374, opened 2015-12-06
add test urls for TED and re-enable the ruleset

numismatika:


semenko:
Awesome! Ted.com looks fixed -- want to remove the (Buggy) label?

semenko:
Rather than make more work, let's just assume it could be buggy, and roll with this ruleset.

PR #3379, opened 2015-12-07
Add French I18N (Chromium)

Micdu70:


semenko:
Awesome! Thanks so much! :100:

PR #3390, opened 2015-12-10
Fix typos

jwilk:
This PR was brought to you by [mwic](http://jwilk.net/software/mwic).

### DIFF
--- a/chromium/lru.js
+++ b/chromium/lru.js
@@ -31,7 +31,7 @@ function LRUCache (limit) {
  */
 LRUCache.prototype.put = function(key, value) {
   var entry = {key:key, value:value};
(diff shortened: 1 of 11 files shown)


reedy:
Unfortunately, as you modified XML rulesets, travis doesn't like it.

We could merge this anyway, and ignore travis, but that's not brilliant

If we could split the commits, so that they're not all in one PR, that might be easier

jwilk:
Would it be sufficient to move the commits that touch rules (5564e1b14534 and 9dcc7c92526d) to a separate PR?

reedy:
Yup, exactly, that'd do the job

jwilk:
Moved the two commits to #3449.

Issue #77, opened 2014-01-07
Project: Find all the HTTPS Everywhere rules that can be HSTS rules


diracdeltas:
Someone should take this project; it sounds fun!

The idea is to write a program that parses HTTPS Everywhere rules and detects which ones can be equivalent to HSTS preloads. Seth took a stab at it in utils/simple.py, but it needs some work.

Once we have a list of rules from HTTPS Everywhere that can be used to expand the built-in browser HSTS preload list, there's a couple of awesome things we can try:
* Offer this list as a download to developers writing software/apps that would benefit from STS.
* Since it's impossible to make HTTPS Everywhere for Safari, the next-best thing seems to be adding this list to ~/Library/Cookies/HSTS.plist in Mavericks. This file remembers STS for Safari and possibly other apps as well.
* Perhaps we can completely replace these rules in FF and Chrome by STS preloads so that we use the browsers' built-in STS mechanism for switching domains to HTTPS rather than HTTPS Everywhere's. This might help with performance issues.

jsha:
Closing. I don't think it's safe to inject HSTS headers for sites, since that means any certificate error turns into a hard fail (can't click through it). I think it really needs to be at the site owner's discretion to turn on HSTS.

PR #3454, opened 2015-12-15
[Wikimedia.xml] Remove already preloaded domains

nextPrime:
The removed domains have been included in Chrome's HSTS-preloaded list for quite a long time. So no need to include them in HTTPS Everywhere. Wikimedia.org is not currently HSTS-enabled, so it remains here.

### DIFF
--- a/src/chrome/content/rules/Wikimedia.xml
+++ b/src/chrome/content/rules/Wikimedia.xml
@@ -30,8 +30,6 @@
 	<target host="enwp.org" />
 	<target host="frwp.org" />
 
-	<target host="mediawiki.org" />
-	<target host="*.mediawiki.org" />
 	<target host="wikimedia.org" />
 	<target host="*.wikimedia.org" />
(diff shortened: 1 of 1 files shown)


reedy:
Feels like a stupid question, but, what about FF too?

nextPrime:
@reedy IE, Edge, Firefox, and Chrome all use the Chrome's HSTS preload list.

semenko:
Hm, this is probably reasonable, though if we were to go this route, let's:

- Make sure there's no use case for HTTPSe without a preload list
- Do this on a broader scale, removing all the pre-loaded domains

IIRC, a while back, we added all the Chrome preload sites to HTTPSe, perhaps when there wasn't broader pre-load support.

Sorta related #77

jsha:
Yep, when we added the preload sites into HTTPSe, there wasn't support for preload in FF at the time. Now there is. One of our community members did a purge of some of those auto-created rulesets a while back, but didn't touch rulesets that weren't auto-created.

semenko:
In that case, ltgm. Confirmed these are in https://code.google.com/p/chromium/codesearch#chromium/src/net/http/transport_security_state_static.json

PR #3464, opened 2015-12-16
Update Favstar.fm rules

mathiasbynens:
Favstar has finally removed their HTTPS to HTTP redirects. See https://twitter.com/TimHaines/status/676986243822014464.


semenko:
Thanks!

PR #3478, opened 2015-12-16
Add an on/off switch for Chrome

### LINKED ISSUE #760: Add UI to temporarily disable HTTPS Everywhere for Chrome

semenko:
Allow users to enable/disable HTTPSe from the extension popup itself -- otherwise it's hard to temporarily suspend the extension.

Two notes:
- On Chrome restart, the extension is always enabled
- I tried allowing users to disable HTTPSe per-tab [see commits], but that's a mess — pre-rendered tabs randomly replace current tabs, cookie rules don't understand the concept of a tab, etc.

This looks like:
screen shot 2015-12-16 at 5 24 12 pm
screen shot 2015-12-16 at 5 24 23 pm

Closes #760

Signed-off-by: Nick Semenkovich \<semenko@alum.mit.edu\>

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -183,7 +183,9 @@ var redirectCounter = {};
  * */
 function onBeforeRequest(details) {
   // If HTTPSe has been disabled by the user, return immediately.
-  if (!isExtensionEnabled) { return; }
+  if (!isExtensionEnabled) {
+    return;
+  }
 
   var uri = document.createElement('a');
   uri.href = details.url;
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -37,6 +37,11 @@ var switchPlannerEnabledFor = {};
 // rw / nrw stand for "rewritten" versus "not rewritten"
 var switchPlannerInfo = {};
 
+// Is HTTPSe enabled, or has it been manually disabled by the user?
+var isExtensionEnabled = true;
+// The setBadgeText API has an abandoned bug: https://crbug.com/170413
+chrome.browserAction.setBadgeText({ text: "" });
+
 // Load prefs about whether http nowhere is on. Structure is:
 //  { httpNowhere: true/false }
 var httpNowhereOn = false;
@@ -177,6 +182,9 @@ var redirectCounter = {};
  * @param details of the handler, see Chrome doc
  * */
 function onBeforeRequest(details) {
+  // If HTTPSe has been disabled by the user, return immediately.
+  if (!isExtensionEnabled) { return; }
+
   var uri = document.createElement('a');
   uri.href = details.url;
 
@@ -459,7 +467,7 @@ function switchPlannerDetailsHtmlSection(tab_id, rewritten) {
  * @param changeInfo Cookie changed info, see Chrome doc
  * */
 function onCookieChanged(changeInfo) {
-  if (!changeInfo.removed && !changeInfo.cookie.secure) {
+  if (!changeInfo.removed && !changeInfo.cookie.secure && isExtensionEnabled) {
     if (all_rules.shouldSecureCookie(changeInfo.cookie, false)) {
       var cookie = {name:changeInfo.cookie.name,
                     value:changeInfo.cookie.value,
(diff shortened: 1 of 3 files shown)


chromium/popup.js
```diff
@@ -77,12 +77,35 @@ function createRuleLine(ruleset) {
...
+  document.getElementById('onoffswitch').checked = backgroundPage.isExtensionEnabled;
+}
+
+// Toggle extension enabled/disabled status
+function toggleEnabledDisabled() {
+  if (backgroundPage.isExtensionEnabled) {
+    // User wants to disable us
+    backgroundPage.isExtensionEnabled = false;
+    chrome.browserAction.setBadgeText({ text: "OFF" });
+    updateEnabledDisabledButtonState();
```
jsha: Move this call outside the if/else since it's the same in both clauses.

jsha:
Generally looks good, and I applaud the execution and the idea of resetting start on startup.

I also like the cool on/off slider, but I'd rather not introduce a new UI element piecemeal. Could you pull that out of this PR, and possibly reintroduce it as an overhaul of our checkboxes generally?

Also, please squash the commits for a cleaner history.

semenko:
@jsha updated & squashed, PTAL.

Looks like this now, using an existing translation :tada: 

screen shot 2016-01-19 at 8 12 56 pm

screen shot 2016-01-19 at 8 13 03 pm

corysimmons:
yay :100:

semenko:
@jsha Friendly ping here

chromium/background.js
```diff
@@ -177,6 +182,9 @@ var redirectCounter = {};
  * @param details of the handler, see Chrome doc
  * */
 function onBeforeRequest(details) {
+  // If HTTPSe has been disabled by the user, return immediately.
+  if (!isExtensionEnabled) { return; }
```
jsha: Newline after brace.

chromium/popup.js
```diff
@@ -77,12 +77,34 @@ function createRuleLine(ruleset) {
...
+
 /**
  * Create the list of rules for a specific tab
- * @param tab
+ * @param tabArray
  */
-function gotTab(tab) {
-  var rulesets = backgroundPage.activeRulesets.getRulesets(tab.id);
+function gotTab(tabArray) {
+  var activeTab = tabArray[0];
```
jsha: Are we guaranteed to have at least one result? If not, we should return earlier if tabArray.length === 0.

chromium/popup.js
```diff
@@ -77,12 +77,34 @@ function createRuleLine(ruleset) {
...
+    // User wants to disable us
+    backgroundPage.isExtensionEnabled = false;
+    chrome.browserAction.setBadgeText({ text: "OFF" });
+  } else {
+    // User wants to enable us
+    backgroundPage.isExtensionEnabled = true;
+    chrome.browserAction.setBadgeText({ text: "" });
+  }
+   updateEnabledDisabledButtonState();
+  // The extension state changed, to reload this tab.
```
jsha: I think this sentence got mangled.

chromium/popup.js
semenko: We are -- since this is called from the popup extension window.

If Chrome can't find the active rules popup window, something is catastrophically wrong -- and I don't think we can/should work around it.

chromium/popup.html
```diff
@@ -13,6 +13,12 @@
 <header>
   <h1 i18n="about_ext_name"></h1>
 </header>
+<section id="disableButton" class="options">
+  <div class="onoffswitch">
+    <input id="onoffswitch" type="checkbox" checked>
+    <label i18n="menu_globalEnable" for="onoffswitch"></label>
+</div>
```
jsha: Identation wrong on this div.

jsha:
LGTM other than that one last nit. Fix and merge.

semenko:
Fixed. Merging per above LGTM.

corysimmons:
Any ETA on when this will hit the Chrome Extension marketplace?

jsha:
We're going to try and do a release this week.

corysimmons:
Thanks :+1:

PR #3482, opened 2015-12-17
Create Twitch.tv.xml

ekaris:


ekaris:
I cant find any file named twitch.tv so where is the duplicate?

Roman3349:
Rule for twitch.tv is in ruleset Justin.tv.xml (https://github.com/EFForg/https-everywhere/blob/stable/src/chrome/content/rules/Justin.tv.xml).

ekaris:
Going to revert this, twitch.tv was working all in HTTPS this morning but not anymore :/

PR #3491, opened 2015-12-17
[Mmister.com] Remove ruleset (domains using nontrust/expired shared w…

Roman3349:
[Mmister.com] Remove ruleset (domains using nontrust/expired shared webhosting certificates)


semenko:
It's already default_off, but looks like half the rule is broken/changed, so lgtm.

PR #3502, opened 2015-12-18
Add rule for podnapisi.net

Liudvikam:
podnapisi.net is a subtitle website that supports HTTPS, but sadly does not redirect to it, making you connect via unencrypted HTTP whenever you forget to include the 'S' in HTTPS. An HTTPS Everywhere rule would partially remedy this.


Liudvikam:
Thanks!

Issue #3510, opened 2015-12-18
add ruleset for blog.fefe.de


sjjh:
Following rule should add support for https://blog.fefe.de

    <ruleset name="Fefes Blog">
            <target host="blog.fefe.de" />

            <rule from="^http:"
                    to="https:" />
    </ruleset>

semenko:
Thanks! Do you think you could add create a [pull request](https://help.github.com/articles/using-pull-requests/) with that rule?

sjjh:
Done:  #3515
I hope I did everything correct. The rule is untested (but looks trivial to me).

sjjh:
The change seems non-trivial to me any more, I don't have the necessary knowledge of the xml file structure. I hope you can get the idea of what I wanted to archive in my second pull request.

numismatika:
@semenko did you check the certs before merging?
https://www.ssllabs.com/ssltest/analyze.html?d=blog.fefe.de&latest and related seem to be back at the CACert stuff with:
Strict Transport Security (HSTS) 	Yes   TOO SHORT (less than 180 days)
Issuer 	CA Cert Signing Authority

seems like either trolling or a rather short test of LE

sjjh:
I can confirm that the site is from Let's encrypt back to CAcert. Maybe rollback and wait until the sites certificates get more stable.

semenko:
Sorta confused -- the entry at https://www.ssllabs.com/ssltest/analyze.html?d=blog.fefe.de shows LE certs. Their blog post seems to say they fixed the issue, and chrome was showing stale data.

PR #3515, opened 2015-12-18
added rule for blog.fefe.de

sjjh:
the new rule is untested (but looks trivial to me).


sjjh:
fixes #3510

sjjh:
seems that someone was faster than me?

numismatika:
Fefe.xml exists for some time now, but it was off by default for using Cacert.
You should check its domains and base your PR on that file.

sjjh:
I'm just talking about blog.fefe.de, not about any possible other subdomains. blog.fefe.de uses a cert from Let's Enrcrypt and thus could be enabled by default.

youdly:
Since `blog.fefe.de` is added in https://github.com/EFForg/https-everywhere/pull/3521, this PR can be closed.

PR #3516, opened 2015-12-18
make-trivial-rule: Use tabs in output

jwilk:


semenko:
LGTM -- aligns w/ style guide: `Use tabs and double quotes`

PR #3517, opened 2015-12-18
Create Amd.com.xml

ekaris:


ekaris:
I know this rule is incomplete, can anyone tell me which tools are used to automate testing of all URLs in the domain @J0WI

J0WI:
You should have a look at https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Advanced-Micro-Devices.xml
It would be great if you could refactor that instead of doing the work twice.

You can search for DNS names in the certificate, on Google using the ```site:```parameters (AFAIK only Google supports wildcards) and maybe there are loading some external resources on the page itself.

PR #3531, opened 2015-12-19
LongTail:

numismatika:
- remove longtail.tv domain, as it no longer exists
- remove comment from the checker script
- removed unnecessary host
- added test-urls
- enable jwplayer.com
- move from wildcards to explicit target hosts
- add account subdomain


semenko:
Tentative LGTM, but I'm going to test this a little more, since videos are notoriously fragile.

semenko:
Seems to be working. Awesome work! :100:

PR #3532, opened 2015-12-19
[Apkmirror.com.xml] New ruleset

rugk:
https://twitter.com/rugkme/status/678190574474797058

### DIFF
--- a/src/chrome/content/rules/Apkmirror.com.xml
+++ b/src/chrome/content/rules/Apkmirror.com.xml
@@ -1,7 +1,7 @@
 <!--
 
 -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Apkmirror.com.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+
+-->
+<ruleset name="Let's Encrypt">
```
semenko: This should probably be APK Mirror or something else.

semenko:
:100:

Issue #3447, opened 2015-12-14
Web.de Rule


IDKwhattoputhere:
"breaks videos"
Not true anymore since they seem to have changed the site.
Example: http://web.de/magazine/unterhaltung/lifestyle/katzenvideos/katze-hypnotisiert-fenster-30576910
Works fine with rule enabled since they reject SSL for all articles and sites with videos.
Please set to default on. Please also add:
bap.navigator.web.de
navigator.web.de
which host the webmail interface,
v.web.de
which hosts the video files,
i1.web.de
i2.web.de
also host images like i0.web.de which is already in the rule.

J0WI:
a http redirect would still break the rule

semenko:
Thanks for the report!

If you have the time, it'd be great if you could submit a small pull request enabling those domains.

numismatika:
i have a first attempt here #3534 
It may need more work on the exclusion patterns and is still largely untested.

IDKwhattoputhere:
Thank you. This can be closed then?

PR #3534, opened 2015-12-19
Web.de: convert to new style & add more subdomains

numismatika:
Attempt at fixing #3447


semenko:
Awesome -- thanks! Seems to work, and that akamai endpoint resolves.

PR #3547, opened 2015-12-19
[Debian] Add appstream.debian.org, jenkins.debian.org

jwilk:
Forwarded from @pabs3:
https://lists.eff.org/pipermail/https-everywhere-rules/2015-November/002270.html


semenko:
:100:  thanks!

PR #3550, opened 2015-12-19
Add rule for teo.lt

Liudvikam:
teo.lt is the website of a Lithuanian ISP that has the same problem as the one described in #3502.


semenko:
awesome - thanks! :)

Liudvikam:
You're welcome.

PR #3558, opened 2015-12-20
Improve the rule for thecthulhu.com

Liudvikam:
The rule for thecthulhu.com is missing a few subdomains, some of which do not redirect to HTTPS and therefore are in need of an HTTPS Everywhere rule.

This would also rename the rule to "Thecthulhu.com", since the space that is there is unnecessary.


Liudvikam:
Thanks!

PR #3562, opened 2015-12-20
N1- add some new rules

numismatika:
adds a couple of new rules


J0WI:
Could you split up this in multiple pull requests?
It's much easier to review when there is only one new ruleset per pull request.

PR #3566, opened 2015-12-20
12 2015 3dparty

numismatika:
todays batch of updates for external content that is embedded on websites.


semenko:
Thanks! LGTM, including the wildcard removals (reuters is now at Akamai and broken, etc.)

PR #3572, opened 2015-12-20
Dreamhost subdomain transferapproval.dreamhost.com does not work over https

richard4339:
I modified the exclusion line with the two existing inclusions to include transferapproval.dreamhost.com

Modified the spacing on the header since this new domain is longer and threw off the tabbing


semenko:
LGTM -- thanks!

PR #3579, opened 2015-12-21
[Private-Layer.com] New ruleset

youdly:
disabled for now due to cert issue.


src/chrome/content/rules/Private-Layer.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="Private Layer" platform="cacert" default_off="missing certificate chain">
```
semenko: is this still cacert? [not a huge fan of adding default_off rulesets in general]

youdly: my mistake, it's not cacert. the website comes with intermediate cert "GeoTrust DV SSL CA - G4" which causes the cert issue. anyway, I'll re-open this PR when cert issue has been resolved.

PR #3521, opened 2015-12-18
updated Fefe.xml to reflect new certificates from Let's Encrypt

sjjh:
to fix #3510 (instead of #3515)

PR #3581, opened 2015-12-21
reflect certificate changes on fefe.de

Yannik:
full chain is sent, lets encrypt instead of cacert


semenko:
Whoops, just noticed this was on stable. Could you please make this against master instead?

semenko:
Handled in #3521 now.

PR #3591, opened 2015-12-22
Add HTTPS Everywhere rule for several sites

lqf96:


semenko:
Awesome -- thanks! :)

PR #3592, opened 2015-12-22
[Open_Source.org] Add opensource.org

jwilk:


src/chrome/content/rules/Open_Source.org.xml
```diff
@@ -1,20 +1,11 @@
...
 -->
-<ruleset name="Open Source.org (partial)" default_off="missing certificate chain">
 
-	<!--	Direct rewrites:
-				-->
-	<target host="lists.opensource.org" />
+<ruleset name="Open Source.org (partial)">
 
+	<target host="opensource.org" />
+	<target host="www.opensource.org" />
```
semenko: www doesn't work for me.
````
~ $ curl -Iv https://www.opensource.org/
*   Trying 217.70.184.38...

^C
````

jwilk:
My bad. I removed the www.opensource.org host.

PR #3594, opened 2015-12-22
[Docracy] New ruleset

rugk:


ukaccount123:
Presumably the opening quote block should be "<!--" and not "!--" (missing less-than symbol, as picked up by travis)

fuglede:
https://docracy.com is problematic:

    $ curl -I https://docracy.com
    curl: (7) Failed to connect to docracy.com port 443: Connection refused

semenko:
@rugk I'd love to merge this -- would you please escape those special characters here too, to fix the build error? Thanks! :+1:

rugk:
> Missing caret (^) in 'from' attribute: it should be "^http:" and not "http:".

That's not true. There is a caret in both.

> Missing trailing slashes in 'from' or 'to' when specifying full hostnames

Also not true.

There is a validation error, but I still do not know why it fails here.

semenko:
I do think the second half is true:
```
- Missing trailing slashes in 'from' or 'to' when specifying full hostnames: it should be "https://eff.org/" and not "https://eff.org".
```

e.g.
```
+	<rule from="^http://docracy\.com"
 +		to="https://www\.docracy\.com" />
```
vs
```
+	<rule from="^http://docracy\.com/"
 +		to="https://www\.docracy\.com/" />
```

fuglede:
Also, the backslashes in the <code>to</code>-attribute.

PR #3598, opened 2015-12-23
New ruleset for douban.com

lqf96:


semenko:
Some mixed content but seems functional

PR #3603, opened 2015-12-23
Fix non-standard wildcard in Live.xml

### LINKED ISSUE #2981: Non-standard wildcard in Live.xml
Live.xml is the only ruleset that contains wildcard as following:
```xml
<target host="secure.*.live.com" />
```

From my understanding, such wildcard can be considered incorrect given that documentation explicitly states:
> To cover all of a domain's subdomains, you may want to specify a wildcard target like `*.twitter.com`. Specifying this type of left-side wildcard matches any host name with .twitter.com as a suffix, e.g. `www.twitter.com` or `urls.api.twitter.com`. You can also specify a right-side wildcard like `www.google.*`. Right-side wildcards, unlike left-side wildcards, apply only one level deep. So if you want to cover all countries you'll generally need to specify `www.google.*`, `www.google.co.*`, and `www.google.com.*` to cover domains like `www.google.co.uk` or `www.google.com.au`. You should use wildcard targets only when you have rules that apply to the entire wildcard space. If your rules only apply to specific hosts, you should list each host as a separate target.

Such wildcard is not neither left-side nor right-side, and it's unclear how many domain levels it might contain when placed in between. Due to such ambiguity, I believe we should either document the behavior explicitly or just replace this single rule with broader left-side one.

RReverser:
Fixes #2981. /cc @fuglede

### DIFF
--- a/src/chrome/content/rules/Live.xml
+++ b/src/chrome/content/rules/Live.xml
@@ -103,9 +103,7 @@
 		- profiles
 		- rad
 
-		- secure.*
-
-			- secure.shared
+		- secure.shared
 
 		- directory.services
 		- livecontacts.services
@@ -153,7 +151,7 @@
 
 	<!--	Direct rewrites:
 				-->
-	<target host="secure.*.live.com" />
+	<target host="secure.shared.live.com" />
 	<target host="account.live.com" />
 	<target host="calendar.live.com" />
 	<target host="c.live.com" />


RReverser:
Yay, thanks :)

RReverser:
```
node_modules/.bin/babel-node  test.js
8866/8866 passed (errors: 0, critical: 0).
```

Seems to work :)

fuglede:
🎉

PR #3605, opened 2015-12-23
[Debian.net] Remove live.debian.net

jwilk:


semenko:
LGTM (I get a connect refused.)

PR #3606, opened 2015-12-24
Add rule for tickets.gwr.com

ukaccount123:
tickets.gwr.com uses http (despite seemingly supporting https with no issues) whilst all other subdomains of gwr.com use https.


semenko:
:100: Thanks!

PR #3619, opened 2015-12-25
Virgin media update

ukaccount123:
I've updated the rules for Virgin Media, but haven't added any test urls.

Presumably the "correct" course of action is to change the (a|b|c).example.com rule to separate rules per subdomain to reduce the number of tests required as per https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md#ruleset-coverage-requirements ?

Also I haven't touched the tracking cookie rules so somebody should check this is still correct.

### DIFF
--- a/src/chrome/content/rules/Virgin_Media.com.xml
+++ b/src/chrome/content/rules/Virgin_Media.com.xml
@@ -30,7 +30,7 @@
 		- (www.)
 		- assets 
 		- da
-		- help		(requires assets assets rule to prevent mixed-content)
+		- help		(requires assets rule to prevent mixed-content)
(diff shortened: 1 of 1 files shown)


ukaccount123:
...or should I have a <target> block per subdomain and a single http to https <rule> e.g. like

https://github.com/renyouguo/https-everywhere/blob/patch-3/src/chrome/content/rules/JD.com.xml ?

J0WI:
yes, that's easier and more fail-safe

ukaccount123:
I'll write a neater pull request then submit separately :)

ukaccount123:
New pull request at https://github.com/EFForg/https-everywhere/pull/3626

PR #3620, opened 2015-12-25
Create Nuuvem.com.xml

ekaris:


J0WI:
You may also add ```secure.nuuvem.com``` and comments, that the following subdomains are broken:
- alendadoheroi (timeout)
- forum (refused)
- support (mismatch)

PR #3626, opened 2015-12-25
Updated Virgin Media rules

ukaccount123:
Tidied up and fixed rules for virginmedia.com

### DIFF
--- a/src/chrome/content/rules/Virgin_Media.com.xml
+++ b/src/chrome/content/rules/Virgin_Media.com.xml
@@ -7,9 +7,43 @@
 
 	Nonfunctional subdomains:
 
+		- about *
+		- admin *
+		- adserver *
 		- anywhere *
+		- careers *
+		- cms *
+		- contact *
+		- dev *
+		- downloads *
+		- email *
+		- energy *
+		- events *
+		- fedora *
+		- finance *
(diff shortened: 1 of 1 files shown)


J0WI:
I didn't check all the rules yet, but the following hosts are known to be RC4 only (#3627):
identity.virginmedia.com
national.virginmedia.com

ukaccount123:
Yep I've marked them as such in the rules - should I disable those two for now?

identity.virginmedia.com is used by all customers to sign in to their accounts (via a redirect from the "My Virgin Media" Sign In link in the top right at http://www.virginmedia.com) so I presume this will be very quickly fixed once nobody can login.

national.virginmedia.com I'm not so hopeful about.

J0WI:
I would disable them for now, because this result in a broken site and users have to use http anyway.
So it's better to have a working ruleset than disabling the whole ruleset.

An alternative would be to have a separate ruleset for problematic domains, that is disabled by default. But most users won't notice that.


Other meanings? (see also #3627)
cc @semenko @fuglede @jsha

jsha:
Yep, I agree on disabling the specific hosts that have problems. Thanks!

ukaccount123:
I've commented out the affected subdomains

PR #3628, opened 2015-12-26
Add and update several rules

pgerber:


src/chrome/content/rules/Wiss.ch.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="WISS">
+	<target host="lms.wiss.ch" />
+	<target host="lehre.wiss.ch" />
+
+	<rule from="^http://lehre\.wiss\.ch/"
+		to="https://ecolm.com/" />
```
semenko: Hm, I'm not a huge fan of off-domain redirects. Could you exclude this instead?

semenko:
Awesome -- thanks!

Issue #3523, opened 2015-12-18
calendars.com is completely broken


ANogin:
Right now it looks like HTTPS EveryWhere breaks CSS and images on calendars.com, making site unusable. Disabling calendars.com rule makes it usable again.

fuglede:
Fixed by #4662.

PR #3633, opened 2015-12-27
[Calendars] Fix typo

fuglede:
This fixes a typo in a rule which broke most site resources. In particular, this fixes https://github.com/EFForg/https-everywhere/issues/3523.

### DIFF
--- a/src/chrome/content/rules/Calendars.com.xml
+++ b/src/chrome/content/rules/Calendars.com.xml
@@ -14,6 +14,6 @@
 
 
 	<rule from="^http://(www\.)?calendars\.com/(css/|img/|user)"
-		to="https://$1calendars.com/$1" />
+		to="https://$1calendars.com/$2" />
 
 </ruleset>
--- a/src/chrome/content/rules/Calendars.com.xml
+++ b/src/chrome/content/rules/Calendars.com.xml
@@ -16,4 +16,9 @@
 	<rule from="^http://(www\.)?calendars\.com/(css/|img/|user)"
 		to="https://$1calendars.com/$2" />
 
+
+	<test url="http://calendars.com/css/common.css" />
+	<test url="http://www.calendars.com/img/common/calendarsLogo.gif" />
+	<test url="http://www.calendars.com/user/login.jsp" />
+
 </ruleset>


fuglede:
The ruleset test coverage test fails with the message

    ERROR src/chrome/content/rules/Calendars.com.xml: No rule or exclusion applies to test URL http://calendars.com/

    ERROR src/chrome/content/rules/Calendars.com.xml: No rule or exclusion applies to test URL http://www.calendars.com/

yet none of the listed URLs are actually test URLs in the ruleset. What gives?

donced:
The test coverage reminds me of [Mobile01.com.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Mobile01.com.xml) whereby only a single page supports HTTPS. I used reverse exclusion (line 9). This is what I come up with:
```
<!--
	Nonfunctional subdomains:

		- app		(404, mismatched, CN: *.e-dialog.com)
		- reviews	(mismatched, CN: *.ugc.bazaarvoice.com)


	Some (most?) pages redirect to http

-->
<ruleset name="Calendars.com (partial)">

	<target host="calendars.com" />
	<target host="www.calendars.com" />

  	<exclusion pattern="^http://(www\.)?calendars\.com/(?!(css/|img/|js/|user/))" />

		<test url="http://calendars.com/" />
		<test url="http://www.calendars.com/" />
		<test url="http://www.calendars.com/Baby-Animals/cat00172/" />
		<test url="http://www.calendars.com/Bears/cat00173/" />
		<test url="http://www.calendars.com/Big-Cats/cat00174/" />
		<test url="http://www.calendars.com/Jigsaw-Puzzles/" />

	<rule from="^http:"
		to="https:" />

		<test url="http://www.calendars.com/css/common.css" />
		<test url="http://www.calendars.com/img/common/calendarsLogo.gif" />
		<test url="http://www.calendars.com/js/jquery.timer.js" />
		<test url="http://www.calendars.com/user/login.jsp" />
</ruleset>
```

J0WI:
@fuglede still interested in fixing this? Page looks weird atm.

fuglede:
@J0WI: Hm, yeah, my plan was to trace down whatever part of the test coverage checker that was causing this particular behaviour, rather than pushing a workaround. Since I'm a bit low on time these days, having a workaround would provide more value than nothing though.

J0WI:
As far as I can see you just need to add 3 more tests. css, img and user each with and without www. And you should do the same with /js.
You need to exclude both hosts on /$.

There are two more mismatches on link and topcat.
There is some mixed content on blog.

fuglede:
Yeah, the test checker complains because of the lack of tests for <code>/$</code>, but it makes no sense that we require those as the URLs appear in no rules.

suryapatt:
Hi,

I am leading the development team that works for Calendars.com site. Seems the issue is not yet fixed. Calendars.com  still breaks.

fuglede:
Closing as the fix itself was made redundant by #4662 (but it would still be nice to have the test checker fixed).

PR #3637, opened 2015-12-27
remove duplicate files for HEG and Picload

numismatika:
somewhere in the last merged pull requests some conflicts were introduced.
HEG.com and Picload.org both are spread over two files each now which causes the check script on travis to fail.
I merged the 4 files into 2 to avoid these errors.


semenko:
Thanks!

Issue #1997, opened 2015-06-25
Fix rules that use a248.e.akamaihd.net


jsha:
Per https://blogs.akamai.com/2014/10/poodle-faq-what-akamai-customers-need-to-know.html and outreach from Brian Sniffen, the a248.e.akamaihd.net hostname is going away soon. There are 325 rulesets in HTTPS Everywhere that use that name.

Brian Sniffen recommended, for hostnames that are known to be Akamai-fronted but don't have a valid cert for their own name, we could go ahead and contact the hostname on port 443, and look for a cert containing the wildcard name `*.akamaihd.net`. We can probably do this on Firefox but not on Chrome, so we'd additionally want to tag the relevant rules with platform=firefox, so they don't apply on Chrome.

semenko:
Oof, this name disappeared. We should take a look at this soon.

jsha:
Yep. I think just delete or default_off the rules that rely on rewriting to this hostname. Thanks for noticing this!

PR #3649, opened 2015-12-28
Remove some a248.e.akamaihd.net redirects.

jsha:
Previously, it was possible to redirect to this domain to get working HTTPS, but
it no longer exists. Remove the rewrites from some rulesets, and delete some
rulesets that consisted only of Akamai rewrites.

Part of https://github.com/EFForg/https-everywhere/issues/1997

cc @semenko @fuglede @loveithateit for review.


semenko:
LGTM (especially given the active breakage that redirect causes)

PR #3657, opened 2015-12-28
Framasoft

AtomiKe-FR:


fuglede:
The build fails since framasphere.org is already covered in <code>Framasphere.org.xml</code>.

AtomiKe-FR:
Oups. This pull request should be good. https://github.com/EFForg/https-everywhere/pull/3734

AtomiKe-FR:
Oups. This pull request should be good. https://github.com/EFForg/https-everywhere/pull/3734

PR #3659, opened 2015-12-29
Update nzz.xml

J0WI:
add CDN stuff


semenko:
LGTM, including resrc CDN

PR #3680, opened 2015-12-29
[RC4] Update Uni-Kl.de.xml

J0WI:
update comments


semenko:
Closing since this is comments-only & throws rule issues. You can exclude this in `utils/ruleset-coverage-whitelist.txt` if you like

PR #3691, opened 2015-12-30
[elektro-paloucek.cz] Disable ruleset (Heartbleed, CVE-2014-0224)

Roman3349:


semenko:
Mixed feelings about this. Not sure the added risk of heartbleed exposing user data is greater than the risk of cleartext over the wire.

PR #3698, opened 2015-12-31
Create Smzdm.com.xml and Update Blogspot.com_blogs.xml

renyouguo:


src/chrome/content/rules/Smzdm.com.xml
```diff
@@ -0,0 +1,20 @@
+<ruleset name="smzdm.com">
```
semenko: Tons of mixed content including jquery -- does this break functionality?

renyouguo: @semenko No problem.Ihave test it.

PR #3702, opened 2015-12-31
[grandmargherita.com] New ruleset

youdly:
Also include new ruleset for riversidemajestic.com


fuglede:
For grandmargherita.com:

* On https://grandmargherita.com/meeting.php, if one clicks an image, nothing happens (expected behaviour is that an image shows up).
* Same issue on https://grandmargherita.com/room.php and https://grandmargherita.com/gallery.php (and possibly other pages).

Curiously, the same issue does not exist on riversidemajestic.com; that ruleset looks good to me.

youdly:
The issue also affects HTTP. It seems those page are not using js function properly.

fuglede:
Hm, yeah, it works on both HTTP and HTTPS here now. Caching issues? I'm not going to look into it further.

PR #3715, opened 2015-12-31
Made doi.org more general

KOLANICH:


J0WI:
You should also add https://www.doi.org/ as a test domain.

lgtm then

PR #3717, opened 2015-12-31
Added rule for dellin.ru

KOLANICH:


fuglede:
The build fails because no test URLs apply to the exclusion pattern (while two such tests should be given);

    ERROR src/chrome/content/rules/dellin.ru.xml: Not enough tests (0 vs 2) for <Exclusion pattern '^http://m.dellin\.ru/.+'>

PR #3731, opened 2016-01-01
[Senate of the Philippines] Enable the ruleset by default

jwilk:


J0WI:
You may also add https://janus.senate.gov.ph/

everything else lgtm

jwilk:
I've simplified the rule and added `janus.senate.gov.ph`.

J0WI:
great, lgtm

PR #3726, opened 2016-01-01
added Bernina.xml

J0WI:

PR #3732, opened 2016-01-01
[Rasende.dk] New ruleset

fuglede:


J0WI:
Doesn't this need more test urls since it contains a ```?``` expression?

rule lgtm

fuglede:
Good question; I never looked into the details of the test URL calculation myself; it [looks like](https://github.com/jsha/https-everywhere-checker/blob/updates/src/https_everywhere_checker/rules.py#L184) two tests would be needed, so maybe we're just happy because those two tests come from the target hosts. It's worth looking into, though, as it seems (from #3633 and #3726) that something is not quite right.

semenko:
Gonna merge this, but we definitely need to look into ruleset testing a big more in those other issues.

PR #3737, opened 2016-01-02
[Irssi] New ruleset

jwilk:


fuglede:
Looks good to me; site already redirects to HTTPS on its own.

PR #3739, opened 2016-01-02
Update Blendle ruleset

koenrh:
Both blendle.nl and blendle.com are now HTTPS-only. The exclusion for `blog.blendle.nl` is no longer necessary.


fuglede:
https://blog.blendle.nl/ times out here.

koenrh:
I didn't check it properly. My bad :-(

PR #3742, opened 2016-01-02
[Randi Schumann] New rulesets

fuglede:


J0WI:
lgtm

J0WI:
you may consider merging this rules into one file, because it's the same host and content.

PR #3743, opened 2016-01-02
[Reyn.ir] New ruleset

fuglede:


J0WI:
Mismatch on ```paste.```

I think it would make sense to merge them into one file since they have the same content, host and cert.

fuglede:
I agree completely.

PR #3744, opened 2016-01-02
[Agol.dk] New ruleset

fuglede:


J0WI:
serves preloaded HSTS header including subdomains, but isn't valid for www.
This is really bad.

The owner should fix this issue and then the domain will be accepted in the preload list. I'm not sure if we should merge this rule as a temp. workaround.

J0WI:
I think in this case a documented https redirect would make sense.

fuglede:
Yeah, I informed the owner back when I submitted this one; looks like nothing much has happened since then. And yeah, normally I would hesitate to redirect from https because of the maintenance overhead, but since we might be the ones to trigger the faultily set up HSTS, it's probably well warranted here.

PR #3745, opened 2016-01-02
[Kommaer.dk] New ruleset

fuglede:


J0WI:
lgtm if this doesn't requires additional test urls.

PR #3750, opened 2016-01-02
[Hvidovrekajakklub.dk] New ruleset

fuglede:
Some mixed content here and there, but nothing that appears to break site function.


J0WI:
The site contains mixed content.
rule lgtm

PR #3757, opened 2016-01-02
Fix standalone chromium profile runner in Debian

oldhill:
Fixes `bash test/chromium.sh --justrun` on Debian systems, where the chromium package is named differently than on Ubuntu.

The package name logic is copied from the existing `install-dev-dependencies.sh` script: https://github.com/EFForg/https-everywhere/blob/5effa1541a5a8dd9299a9a7820cec2a6a8052655/install-dev-dependencies.sh#L8


semenko:
Awesome -- thanks!

PR #3777, opened 2016-01-04
Create netease.com.xml

renyouguo:
Test:
https://nos.netease.com/edu-image/001DF8829C2FD685596FF3AFF484BEE2.jpg


fuglede:
Thanks for the PR. The build fails since there is already a ruleset whose filename is the same up to case. In this case, it also contains the same target host, so I am closing this PR as redundant.

PR #3778, opened 2016-01-04
Add a few rules

Liudvikam:
zimperium.com is a website that supports but fails to redirect to HTTPS and therefore is in need of an HTTPS Everywhere rule.


Liudvikam:
And I'll add rules for two other websites, lanayachocolate.com and torrentbutler.eu, both of which have the same problem.

semenko:
Great, thanks!

Liudvikam:
You're welcome.

PR #3779, opened 2016-01-04
[PaulDreik.se] New ruleset

jwilk:


fuglede:
Looks good to me; both target hosts already redirect to HTTPS on their own.

PR #3791, opened 2016-01-06
[procensus.com] New ruleset

graingert:


src/chrome/content/rules/procensus.com.xml
```diff
@@ -0,0 +1,17 @@
+<ruleset name="Procensus">
+
+	<target host="*.procensus.com" />
+	<target host="procensus.com" />
+
+	<securecookie host="^.*\.?procensus\.com$" name=".+" />
+
+	<rule from="^http://(www\.)?procensus\.com/"
```
semenko: Awesome -- thanks! Since https://www.procensus.com/ seems to work fine, please remove this line.

(Perhaps they'd like to use the www. domain in the future, etc.)

semenko:
Awesome -- thanks! :)

PR #3796, opened 2016-01-06
Create Movie2k.tv.xml

J0WI:


src/chrome/content/rules/Movie2k.tv.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Movie2k.tv">
+	<target host="movie2k.cm"/>
+	<target host="www.movie2k.cm"/>
+	<target host="movie2k.nu"/>
+	<target host="www.movie2k.nu"/>
+	<target host="movie2k.pe"/>
+	<target host="www.movie2k.pe"/>
+	<target host="movie2k.tv"/>
```
semenko: does this domain exists? seems to be a squatter

J0WI:
The name of the site doesn't match their domain. Okay...

PR #3805, opened 2016-01-08
Create new rules, fix two issues and tweak Serveriai.lt.xml

Liudvikam:
Here are 5 more websites that fail to redirect to HTTPS, despite supporting it. This also tweaks [Serveriai.lt.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Serveriai.lt.xml) to remove the unnecessary wildcard and disables [Jodel-app.com.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Jodel-app.com.xml) because the website's certificate has expired.

### DIFF
--- a/src/chrome/content/rules/Dedikuoti.lt.xml
+++ b/src/chrome/content/rules/Dedikuoti.lt.xml
@@ -1,0 +1,7 @@
+<ruleset name="Dedikuoti.lt">
+	<target host="dedikuoti.lt" />
+	<target host="www.dedikuoti.lt" />
+	
+	<rule from="^http:"
+			to="https:" />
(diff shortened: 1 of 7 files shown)


J0WI:
Thanks for your contribution.
It would me much easier for us to review, if you split all the changes into multiple PRs.

Liudvikam:
@J0WI You mean make a separate PR for each ruleset?

J0WI:
exactly :)

Liudvikam:
Okay, sure. The PRs are #3826, #3827,  #3828, #3829,  #3830, 3831, and #3832.

PR #3807, opened 2016-01-08
[LWN.net] Simplify the rule; secure the cookies

jwilk:


J0WI:
You should also add:
old.lwn.net
static.lwn.net

vena.lwn.net (mismatch, self-signed, expired)

PR #3809, opened 2016-01-09
Itx

numismatika:
rule for Computerwoche.de
rework for Golem.de, Computerbase.de,PCwelt.de and a minor edit for CCC

### DIFF
--- a/src/chrome/content/rules/CCC.xml
+++ b/src/chrome/content/rules/CCC.xml
@@ -11,12 +11,9 @@
 
 		- dasalte ¹
 		- halfnarp.events ²
-		- streaming.media ²
 		- wahlcomputer ²
 
 	¹ Shows wuppa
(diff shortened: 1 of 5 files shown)


J0WI:
There are multiple unrelated changes in this PR.
It's much easier to review if each rule set change has it's own PR.

E.g. while the change on CCC would be fine, the computerbase rule still needs some work.

numismatika:
They are all about it-related news pages, so they share a common theme.
What is the issue with computerbase, have they changed stuff around again?

J0WI:
Sorry, I meant "ComputerWoche".

Unrelated means, that these are all completely different hosts and each ruleset needs individual changes and review.
And then we have the problem, that one rule change is fine (like CCC) and others still need some work (like ComputerWoche). But we can't merge them separately.

PR #3810, opened 2016-01-09
add rule for Pfennigparade

numismatika:


J0WI:
Could you please also add a secure redirect for ^pfennigparade.de?

PR #3811, opened 2016-01-09
add rule for Hehner Reus Systems GmbH

numismatika:


J0WI:
There is a bunch of subdomains that are refused (it seems that there exist a subdomain for each project listed on "Referenzen"). Same if you try to connect without the www prefix.

I propose to expose in the comments that www is the only working domain here and to add e secure redirect for the domain without prefix.

numismatika:
References are links to pages for customers that they designed the websites for, they are not owned by the company. So no need to mention that.

J0WI:
e.g. https://rottenburg.gipsprojekt.de/

PR #3812, opened 2016-01-09
add rule for Lotto.de

numismatika:


J0WI:
Please also add
lotto.de
m.lotto.de

PR #3813, opened 2016-01-09
add rule for IT:Agenten

numismatika:


J0WI:
You should also add the following domains:
it-agenten.com
crm.it-agenten.com

And also comments for non working domains:
fw.it-agenten.com (refused)
stage.it-agenten.com (mismatch)
css.stage.it-agenten.com (mismatch)
tcw.staging.it-agenten.com (mismatch)
blog.it-agenten.com (mixed content)

numismatika:
just forgot to push that change.

J0WI:
Thanks! We use to put comments above the ruleset, but I think this is trivial.

PR #3815, opened 2016-01-09
add rule for Repugraph.com

numismatika:


J0WI:
The cert expired a few days ago. And the host seems to be down.

If they renewed it, could you also add comments for (www.)repugraph.com?

numismatika:
Obviously the website went out of service.

PR #3816, opened 2016-01-09
Switchads.com:

numismatika:
- fix style


J0WI:
Could you add comments that all other domains are refused?
switchads.com
www.switchads.com
dev.switchads.com
w.switchads.com
switchconcepts.com
www.switchconcepts.com

You may also add a redirect for support.switchconcepts.com

numismatika:
We should probably merge Switchads.com.xml and Switch_Adserver.xml

J0WI:
At least it should be mentioned that they are related.

numismatika:
I merged them , added the broken domains and renamed the file.
I guess there are more domains (a-z probably), but i haven't checked their existence.
There is a catch-all dns entry for stuff like testgarbage.switchadhub.com as well (which doesn't serve a valid cert)

J0WI:
I got `proxy` up to `proxy6` and the letters up to `h`.
Please put all the comments on the top of the file.

J0WI:
There was something wrong with the test server. Let me reopen.

PR #3817, opened 2016-01-09
KDE:

numismatika:
- remove rewrite of developer.kde.org


J0WI:
Thanks for fixing this.
Could you add some comments about the issue, that nobody reads the rule in the future?

You may also remove files.kde.org as explained in the mail.

Issue #3721, opened 2015-12-31
RT.com now uses HTTPS by default for all pages


ekaris:
Should we delete the ruleset?

ekaris:
What is done in these cases @J0WI

J0WI:
Nothing, because RT.com still lacks of preloaded HSTS including all subdomains.

fuglede:
And even actually HSTS-preloaded sites [still appear](https://lists.eff.org/pipermail/https-everywhere/2015-October/002639.html) in the ruleset database.

loveithateit:
@fuglede I'm sorry if this comment is offensively patronizing, but in case it isn't (as this issue existing would seem to imply...), remember that torbrowser's HSTS preload list is quite outdated compared to Chromium.

loveithateit:
@fuglede I'm sorry if this comment is offensively patronizing, but in case it isn't (as this issue existing would seem to imply...), remember that torbrowser's HSTS preload list is quite outdated compared to Chromium's.

fuglede:
@loveithateit No offense taken; my comment was only to follow along the same line of J0WI's and say that even if the site *were* HSTS preloadable and included in the HTTPS-E database, that does not mean that it should necessarily be deleted from the database.

PR #2163, opened 2015-07-13
Cleanup HSTS rules

J0WI:
- removed broken rules completely for nonexistent hosts
- fixed/rewrote broken rules
- removed rules that duplicates the HSTS preload list
- removed platform="firefox" and HSTS comment, since Firefox supports preloaded HSTS
- merged auto generated rules with existing rules
- set rules to default off if they need a complete rebasing

PR #3820, opened 2016-01-09
[keycdn.com] New ruleset

PeterDaveHello:


fuglede:
Hey @PeterDaveHello. You do not actually need to add those test URLs; the test engine will figure those out automatically from the target hosts. Other than that, LGTM: the domain is already HSTS preloaded.

PeterDaveHello:
So should I remove them? Thanks.

PeterDaveHello:
Hello ~~~

PeterDaveHello:
I removed it.

J0WI:
This domain is already in the HSTS preload list of browsers: https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsSTSPreloadList.inc#2135
This means the browser is forcing https and refuses any connection over http.

So there is no need for a ruleset.
Thanks anyway!

fuglede:
@J0WI: While certainly less useful than many other rulesets at the moment, they are not completely without value. https://github.com/EFForg/https-everywhere/issues/3721#issuecomment-168814096 and https://github.com/EFForg/https-everywhere/pull/3454 are relevant here.

J0WI:
I already tried to clean them up in https://github.com/EFForg/https-everywhere/pull/2163 refering to the discussion in https://lists.eff.org/pipermail/https-everywhere/2015-May/002525.html

Would be great to have a list of the affected ruleset somewhere instead of just stats to finish this task.

fuglede:
Thanks for the reference; I [ran some stats](https://lists.eff.org/pipermail/https-everywhere/2015-October/002639.html) back in October and noticed an overlap with the preload list as well (but took the opposite angle of trying to find new preloadables among the target hosts). It would be nice with an overview of exactly how slow various browsers are at pulling in HSTS preload lists (and I suppose any decision on what browser versions to support (\*cough\*https://github.com/EFForg/https-everywhere/pull/3876#issuecomment-172965331) would be relevant here as well).

J0WI:
Referring to the [commit log] (https://hg.mozilla.org/mozilla-central/filelog/aa90f482e16db77cdb7dea84564ea1cbd8f7f6b3/security/manager/ssl/nsSTSPreloadList.inc), Mozilla updates it's list each week (7 days).
The SSL Labs API does also [provide] (https://github.com/ssllabs/ssllabs-scan/blob/stable/ssllabs-api-docs.md#endpointdetails) information about the preloaded state of a host in each browser, even for Tor.

fuglede:
Hm, just checked with my own site, which I submitted to the preload list ages ago; according to SSL Labs, that one is not in Tor's list. It's a bit unfortunately that the SSL Labs list does not appear to cover older browser versions in general.

J0WI:
Tor uses Firefox ESR, that uses a different release cycle.
Firefox ESR Release Overview

fuglede:
Ah, good to know. Any other notable release cycles that one should be aware of/take into account for compatibility considerations?

PR #3821, opened 2016-01-09
[Decadent.org.uk] New ruleset

jwilk:


fuglede:
Looks good to me; site already serves redirects to HTTPS with HSTS headers on its own.

PR #3822, opened 2016-01-09
[FileFormat.info] New ruleset

jwilk:


fuglede:
Thanks for the PR. I see what you were trying to do with the rule, but unfortunately it won't work quite like that. Instead, <code>\<rule from="^http://(www\\.)\?fileformat\\.info/" to="https://www.fileformat.info/" /\></code> should do it.

jwilk:
Bummer. Fixed.

PR #3824, opened 2016-01-09
[jwilk.net] New ruleset

jwilk:


J0WI:
I think it doesn't make any sense to add a ```default_off="webmaster request"``` rule. What's the reason to have no HSTS?
You can use the ```HTTPSEverywhereUserRules``` directory for custom rules.

How could I confirm that you are the webmaster of jwilk.net btw? ;)

semenko:
Adding default_off rules clutter up stuff. Please default on if TLS works :)

PR #3825, opened 2016-01-09
add rule for PopCash.net

numismatika:
PopUnder Advertiser , nevertheless it should be secured.

### DIFF
--- a/src/chrome/content/rules/PopCash.net.xml
+++ b/src/chrome/content/rules/PopCash.net.xml
@@ -1,0 +1,10 @@
+<ruleset name="PopCash.net">
+
+	<target host="popcash.net"/>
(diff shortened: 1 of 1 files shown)


J0WI:
I got several mixed content warnings.
And you should add ```tb.popcash.net``` too.

numismatika:
Yea they are embedding googleapis on their main page (active mixed content) and pictures (passive mixed content), it is a browser problem with the detection of mixed content.

J0WI:
Sadly most users won't notice that and interpret the rule/Add-On as broken.
We have ```platform="mixedcontent"``` for such cases.

J0WI:
mixed content problems seems to be fixed.

PR #3831, opened 2016-01-09
Add rule for irc.lc

Liudvikam:
Per the request to split #3805 into separate PRs for each of the rulesets, here's the PR for adding a rule for irc.lc.


Liudvikam:
Thanks!

PR #3835, opened 2016-01-11
Create cathaypacific.com.xml

Riotism:
https://cathaypacific.com could not be connected to, and must be redirected to https://www.cathaypacific.com


J0WI:
Are cathaypacific.com and dragonair.com related?

youdly:
even though Dragon Air is subsidiary of Cathay Pacific, but it's a separate brand. To avoid confusion, I think it's better to separate them.
`book` and `lifewelltravelled` of cathaypacific.com supports HTTPS. asiamiles.com as well.

src/chrome/content/rules/Cathay_Pacific.xml
```diff
@@ -0,0 +1,26 @@
...
+	<target host="cathaypacificcargo.com" />
+	<target host="www.cathaypacificcargo.com" />
+	
+	<target host="cxholidays.com" />
+	<target host="www.cxholidays.com" />
+	
+	<target host="dragonair.com" />
+	<target host="www.dragonair.com" />
+	
+	<rule from="^http://(?:www\.)?cathaypacific\.com/" to="https://www.cathaypacific.com/" />
```
youdly: suggest to put a comment why lines 16 - 22 are necessary (timeout) for clarification. `(?:www\.)?` is not necessary since www.cathaypacific.com is covered by line 24.

PR #3837, opened 2016-01-11
Two ruleset additions

Riotism:


J0WI:
Thanks for your contribution.
It's much easier for us to review if we have one PR for each ruleset. Could you please split them up?

PR #3838, opened 2016-01-11
New ruleset for LCSD

Riotism:


fuglede:
More or less looks good to me, but some pages cause redirect loop with this ruleset; those include, but may not be limited to,

- http://www.lcsd.gov.hk/en/news.php
- http://www.lcsd.gov.hk/en/contactus/access.php
- http://www.lcsd.gov.hk/en/contactus/email.php

Other subdomains that are securable to some extent:

- eaps.lcsd.gov.hk
- www2.lcsd.gov.hk
- sc.lcsd.gov.hk

J0WI:
let me retest this one

PR #3842, opened 2016-01-11
[gwhois.org] New ruleset

youdly:


fuglede:
Looks good to me; site redirects to HTTPS on its own and serves a HSTS header.

PR #3852, opened 2016-01-12
Implement async reading of rules from SQLite.

jsha:
This borrows a hack from AdBlock Plus: since the replaceChannel APIs expect a
synchronous response, any request for which we don't yet know whether we want to
rewrite it gets redirected to its own URL, and channel gets suspended. Once we've
loaded the necessary state from SQLite and can decide on a rewrite based on
in-memory information only, we unsuspend the channel. The internal redirect
fires and we get another opportunity to rewrite that request.

Note: Mozilla is currently blocking new versions from being published on AMO until
we have asynchronous SQLite loading.

cc @pde @diracdeltas @cooperq for code review.

### DIFF
--- a/src/chrome/content/code/HTTPS.js
+++ b/src/chrome/content/code/HTTPS.js
@@ -39,7 +39,53 @@ const HTTPS = {
    *   untouched or aborted.
    */
   replaceChannel: function(applicable_list, channel, httpNowhereEnabled) {
-    var blob = HTTPSRules.rewrittenURI(applicable_list, channel.URI.clone());
+    var that = this;
+    // If the callback gets called immediately (not async), it will be called
+    // before the return from the function sets this variable, so we default it
+    // to true.
+    var callbackedImmediate = true;
+    callbackedImmediate = HTTPSRules.rewrittenURI(applicable_list, channel.URI.clone(), function(blob) {
+      if (callbackedImmediate) {
+        that.replaceChannelCallback(applicable_list, channel, httpNowhereEnabled, blob);
+      } else {
+        // If we can't redirect right away because we're waiting on some disk
+        // I/O to read the rules, we will have told the channel to redirect to
+        // itself and suspend, to insure it doesn't actually open while it's
+        // waiting for us to decide what to do. See
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1170197.
+        // Now that we're in the callback, we know the rules are loaded. So we
+        // tell the channel to go ahead and continue with the redirect-to-self.
+        // That will trigger on-modify-request again, and we'll wind up here
+        // again, except with the rules already loaded. We don't try a
+        // redirectTo here because it would fail with NS_ERROR_IN_PROGRESS.
+        try {
+          channel.resume();
+        } catch (e) {
+          that.log(WARN, 'Failed to resume ' + channel.URI.spec + ': ' + e);
+          return;
+        }
+        that.log(DBUG, 'Succeeded resuming ' + channel.URI.spec + ' ');
+      }
+    });
+    if (!callbackedImmediate) {
+      try {
+        channel.redirectTo(channel.URI);
+      } catch (e) {
+        this.log(WARN, 'Failed to redirect to self ' + channel.URI.spec + ': ' + e);
+        return;
+      }
+      try {
+        channel.suspend();
+      } catch (e) {
+        this.log(WARN, 'Failed to suspend ' + channel.URI.spec + ': ' + e);
+        return;
+      }
+      this.log(DBUG, 'Succeeded suspending ' + channel.URI.spec);
+    }
+  },
+
+  replaceChannelCallback: function(applicable_list, channel, httpNowhereEnabled, blob, callbackedImmediate) {
+    var that = this;
     var isSTS = securityService.isSecureURI(
         CI.nsISiteSecurityService.HEADER_HSTS, channel.URI, 0);
     if (blob === null) {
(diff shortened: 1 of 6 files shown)


src/chrome/content/code/HTTPSRules.js
```diff
@@ -472,8 +519,9 @@ const HTTPSRules = {
     }
   },
 
-
-  rewrittenURI: function(alist, input_uri) {
+  // return true iff callback has been called already, false if callback will be
+  // called asynchronously
```
pde: There is a comment a few lines below that documents a different, now-obsolete return value.

pde:
You should probably update the [documentation of the main rewrite call graph](https://github.com/EFForg/https-everywhere/blob/async2/src/components/https-everywhere.js#L176) to reflect this set of changes.

src/chrome/content/code/HTTPS.js
```diff
@@ -39,7 +39,53 @@ const HTTPS = {
    *   untouched or aborted.
    */
   replaceChannel: function(applicable_list, channel, httpNowhereEnabled) {
-    var blob = HTTPSRules.rewrittenURI(applicable_list, channel.URI.clone());
+    var that = this;
+    // If the callback gets called immediately (not async), it will be called
+    // before the return from the function sets this variable, so we default it
+    // to true.
+    var callbackedImmediate = true;
+    callbackedImmediate = HTTPSRules.rewrittenURI(applicable_list, channel.URI.clone(), function(blob) {
```
pde: nit: with complex async logic like this, I'd find it more legible if the callback was assigned to a variable with a good name rather than being defined in the middle of the `rewrittenURI` call.

src/chrome/content/code/HTTPSRules.js
```diff
@@ -487,44 +535,55 @@ const HTTPSRules = {
...
-        if (rs[i].wouldMatch(uri, alist)) alist.moot_rule(rs[i]);
-        continue;
-      } 
-    }
-    return null;
+      callback(null);
+      return true;
+    }
+    var that = this;
+    return this.potentiallyApplicableRulesets(host, function(rs) {
```
pde: nit: again, give this callback a helpful name rather than defining it inside a call?

src/chrome/content/code/HTTPSRules.js
```diff
@@ -576,44 +635,82 @@ const HTTPSRules = {
...
+        }
+      },
+      handleError: function(aError) {
+        that.log(WARN, "SQLite error: " + aError.message);
+        callback();
+      },
+
+      handleCompletion: function(aReason) {
+        if (aReason != Components.interfaces.mozIStorageStatementCallback.REASON_FINISHED) {
+          that.log(WARN, "SQLite query canceled or aborted!");
```
pde: log `aReason` here

src/chrome/content/code/HTTPSRules.js
```diff
@@ -576,44 +635,82 @@ const HTTPSRules = {
...
-  // so we only hit the DB when we know there is something to be had.
-  loadRulesetById: function(ruleset_id) {
-    this.queryForRuleset.params.id = ruleset_id;
-
-    try {
-      if (this.queryForRuleset.executeStep()) {
-        RuleWriter.readFromString(this.queryForRuleset.row.contents, this, ruleset_id);
-      } else {
-        this.log(WARN,"Couldn't find ruleset for id " + ruleset_id);
+  loadRulesetById: function(ruleset_id, callback) {
```
pde: Perhaps, rather than just `callback`, give the callback a name that allows the reader to follow where execution is going next, in this case `callback_loadOne`?

src/chrome/content/code/HTTPSRules.js
```diff
@@ -622,17 +719,15 @@ const HTTPSRules = {
    * This function is only defined for fully-qualified hostnames. Wildcards and
    * cookie-style domain attributes with a leading dot are not permitted.
    * @param host {string}
-   * @return {Array.<RuleSet>}
+   * @return true iff we didn't have to go async to load rules
    */
-  potentiallyApplicableRulesets: function(host) {
+  potentiallyApplicableRulesets: function(host, callback) {
+    if (!callback) {
```
pde: It looks like there aren't any call sites that could cause this branch to execute, but OK...

src/chrome/content/code/HTTPSRules.js
```diff
@@ -387,43 +391,86 @@ const HTTPSRules = {
...
+    if (this.targetsLoaded) {
+      callback();
+      return;
+    }
+    // loadTargets can be called multiple times before it resolves. We store a
+    // list of callbacks to call when done, and make sure we only actually do
+    // the query once.
+    this.targetsLoadingCallbacks.push(callback);
+    if (this.targetsLoadingCallbacks.length > 1) {
+      this.log(DBUG, "Skipping loadTargets, a query is already in progress.");
```
pde: What implements the skip here?

src/chrome/content/code/HTTPSRules.js
```diff
@@ -692,13 +790,33 @@ const HTTPSRules = {
...
+    // When checking for potentially applicable rulesets, we have to wait for a
+    // callback, because we may need to load the rulesets from disk. However, in
+    // practice this callback will always be run immediately, because the
+    // ruleset for the necessary host will have been loaded already for the HTTP
+    // request.
+    var result;
+    var callbackedImmediate = this.potentiallyApplicableRulesets(host, function(rs) {
+      result = this.shouldSecureCookieWithRulesets(applicable_list, c, known_https, rs);
+    }.bind(this));
+    if (callbackedImmediate) {
```
pde: Where is this value of `callbackedImmediate` set?

pde:
Well, I'm convinced that async code is a nuisance and makes things less readable :)

I think this looks OK, but let's talk through it in person?

jsha:
Fixing in #4110 instead.

PR #3853, opened 2016-01-12
Add Endchan Rules

gitgud-software:
This commit adds HTTPS Everywhere rules for the [Endchan](https://infinow.net/) imageboard.

### DIFF
--- a/src/chrome/content/rules/Endchan.xyz.xml
+++ b/src/chrome/content/rules/Endchan.xyz.xml
@@ -1,0 +1,6 @@
+<ruleset name="Endchan.xyz">
+	<target host="endchan.xyz" />
+	<target host="www.endchan.xyz" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Endchan.xyz.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Endchan.xyz">
+	<target host="endchan.xyz" />
+	<target host="www.endchan.xyz" />
```
semenko: Looks like this doesn't work (cert name doesn't include www)

gitgud-software:
This adjustment should resolve the matter for now, I'll add the "www" resolution back in when the SSL certificate properly covers the subdomain.

semenko:
Looks like you removed the wrong www

gitgud-software:
Alright, now we're good.

semenko:
Awesome -- thanks! :100:

PR #3854, opened 2016-01-12
Make Firefox run on the latest version in Travis.

jsha:
This also adds an option to override the path to Firefox when running tests (e.g. to run a developer version), and updates the extensions.json to the latest schema, without which the test won't run.

cc @cooperq @pde for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,20 +17,22 @@ addons:
       - python-lxml
       - python-software-properties
       - chromium-browser
-  firefox: "40.0"
 virtualenv:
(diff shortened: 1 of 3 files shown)


jsha:
This change was failing for a while while I wrestled with Travis, but now it's good. It runs both firefox-latest and firefox-esr-latest.

jsha:
I'm going to go ahead and self-merge this since it's test code only, and it's blocking running tests for other branches locally.

cooperq:
lgtm

PR #3859, opened 2016-01-12
[Diffoscope] New ruleset

jwilk:


fuglede:
Even though the bare domain serves a HSTS header, the www subdomain is problematic:

    $ curl https://www.diffoscope.org
    curl: (51) SSL: no alternative certificate subject name matches target host name 'www.diffoscope.org'

jwilk:
Ooops. I poked the webmaster and it's fixed now.

fuglede:
Cool, looks good to me. Perhaps if you poke even harder it becomes preloaded? ;)

PR #3861, opened 2016-01-12
[W3C] Re-enable www.w3.org

jwilk:


semenko:
Cool, LGTM
ref: https://www.w3.org/blog/2016/01/w3-org-supporting-https-and-hsts/

PR #3863, opened 2016-01-12
new ruleset

shibumi:
hello,
I've added my website.
I hope the ruleset is okay like this was struggling first.. but should be ok like this.


fuglede:
Looks good to me; thanks for the PR!

PR #3867, opened 2016-01-13
Remove a248 from some more rulesets.

jsha:
cc @semenko for review.

Part of #1997 

Approximately two more chunks like this one still coming.


src/chrome/content/rules/Caller.com.xml
```diff
@@ -53,24 +53,18 @@ Fetch error: http://caller.com/ => https://www.caller.com/: (51, "SSL: no altern
...
+	<target host="www.caller.com" />
+	<target host="login.caller.com" />
 	<target host="*.caller.com" />
 
 
 	<securecookie host="^(?:login)?\.caller\.com$" name=".+" />
 
 
-	<rule from="^http://(?:www\.)?caller\.com/"
+	<rule from="^http://caller\.com/"
```
semenko: I get an invalid name here (and connection refused @ https://caller.com).

Perhaps limit this ruleset to login.caller.com only?

semenko:
Despite my note, I'm going to go ahead and merge.

We can worry about other ruleset fixes here later on -- this is a huge improvement over the status quo.

PR #3868, opened 2016-01-13
[Filepicker.io] Rename ruleset, add new domains.

juwancase:
Filepicker has been re-branded to Filestack. Rename ruleset but not filename (to retain history).


semenko:
Awesome -- thanks!

PR #3870, opened 2016-01-13
Remove some more a248 rules.

jsha:
Part of #1997.

cc @semenko for review.

This is actually the last of the batch AFAICT! Searched for "to=.*a248"


semenko:
Awesome :fireworks:

PR #3872, opened 2016-01-13
Remove directory names from ruleset-coverage-whitelist entries

jwilk:


semenko:
Fixing in separate non-conflicting pr

PR #3875, opened 2016-01-13
Netdna (CDN) ruleset rewritten

rugk:
Netdna now seems to belong to MaxCDN. I think it is appropiate to create
a ruleset for this domain as it 1) has ssl in it's name so web devs are
most likely aware that it can be served via HTTPS and b) it has a valid
wildcard cert for all subdomains, so there should be no connection
errors.
Additionally this ruleset may prevent mixed content errors on sites
which use this CDN.


fuglede:
From the Travis log:

    failure: Hostname *.netdna-ssl.com shows up in 2 different rulesets.

The existing ruleset is <code>NetDNA.xml</code>.

semenko:
@rugk I'd love to merge this -- would you please consider merging this with the other NetDNA.xml ruleset (or closing this PR?) Thanks! :fireworks:

rugk:
Well... everything except this one URL in the `NetDNA.xml` ruleset is broken. The only other working this is one MaxCDN secure cookie, but this does not belong in this ruleset and as far as I see it is also covered by the MAxCDN-ruleset.
So you can consider this a merge.

semenko:
Awesome :tada:

PR #3877, opened 2016-01-13
update chromedriver to 2.20 on travis

nikolas:


semenko:
Cool, thanks! Will merge when tests pass (Travis is on fire now: https://www.traviscistatus.com/ )

Any thoughts on moving to an apt-based install, now that we're on sudo + trusty? I think we can sudo apt-get install chromium-chromedriver

nikolas:
I tried using the `chromium-chromedriver` apt package, which we should be able to use. At the moment, it's not whitelisted on travis, so it didn't work. See: https://github.com/travis-ci/apt-package-whitelist/issues/574

semenko:
Yeah, though since we're on their new containerized sudo infra, I *think* we can sudo apt-get install it (that is, not in the `apt` section, but in the `before_script` section).

(see https://github.com/travis-ci/apt-package-whitelist/issues/574#issuecomment-167933084 )

PR #3881, opened 2016-01-13
[HTWK Leipzig] New rulesets

rugk:


J0WI:
You should have a look at https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/HTWK-Leipzig.xml and merge you changes there.

PR #3887, opened 2016-01-14
Switch to apt chromedriver with latest Travis

semenko:
Now that Travis containers on GCE support sudo, let's try installing this package from apt.


semenko:
@jsha ptal

semenko:
Self merge to improve tests, + improve travis :fire:

PR #3888, opened 2016-01-15
[ProtonMail.ch & ProtonMail.com] Merge rulesets, fix securecookie

lucaspetter:
- Merge rulesets from ProtonMail.ch.xml and ProtonMail.com.xml.
- Fix broken regex in previous .ch ruleset's securecookie tag.


loveithateit:
@lucaspetter Firstly what is broken about an unanchored wildcard?

Secondly, those of us less skilled than you with git find reading history more difficult when filenames are changed, so please don't do that.

lucaspetter:
> Fix broken regex in previous .ch ruleset's securecookie tag.

@loveithateit Sorry for the confusion, I don't know what I was thinking. The regex was clearly fine the way it was originally. And I'll try to avoid changing filenames in the future to keep the history readable.

I'll submit a better pull request soon, as there are still changes to make here:
- The securecookie tag is missing altogether in ProtonMail.com.xml.
- The 2 rulesets should be combined, since protonmail.ch redirects to protonmail.com. They are the same site.

PR #3893, opened 2016-01-15
[tsbohemia.cz] Added securecookie

Roman3349:


fuglede:
Looks good to me, given that the site serves a HSTS header with includeSubDomains specified. On the same note, link.tsbohemia.cz, interlink-static.tsbohemia.cz, and mailing.tsbohemia.cz could also be added as target hosts (and a wildcard target would be reasonable as well).

PR #3894, opened 2016-01-15
[PIPNI.cz] Added ruleset

Roman3349:


src/chrome/content/rules/PIPNI.cz.xml
```diff
@@ -0,0 +1,29 @@
...
+    <target host="manual.pipni.cz" />
+    <target host="web.pipni.cz" />
+    <target host="support.pipni.cz" />
+    <!--
+    Active Mixed-content:
+    <target host="stats.pipni.cz" />
+    -->
+
+    <securecookie host="^pipni\.cz$" name=".+" />
+    <securecookie host="^\.pipni\.cz$" name=".+" />
```
fuglede: No cookies of this form are used on stats.pipni.cz? Otherwise this would make them unavailable. It's hard to verify from here as the site requires authentication.

Roman3349:
@fuglede on stats.pipni.cz is running AWStats. AWStats does not use cookies.

src/chrome/content/rules/PIPNI.cz.xml
Roman3349: @fuglede on stats.pipni.cz is running AWStats. AWStats does not use cookies.

semenko:
LGTM then :tada:

PR #3895, opened 2016-01-15
Add more subdomains for serveriai.lt

Liudvikam:


J0WI:
Thanks for figuring them out!
You may also add https://roundcube.serveriai.lt/

Sadly it seems that every customer/server gets a subdomain too, but without a valid cert. Could you document this in a comment for future contributors?

Liudvikam:
> You may also add https://roundcube.serveriai.lt/

The server seems to redirect to HTTPS on that subdomain all by itself.

> Sadly it seems that every customer/server gets a subdomain too, but without a valid cert. Could you document this in a comment for future contributors?

Every customer that decides not to buy a domain gets a subdomain. It also seems that they get a subdomain on puslapiai.lt, not serveriai.lt.

J0WI:
>>You may also add https://roundcube.serveriai.lt/

>The server seems to redirect to HTTPS on that subdomain all by itself.

As long as they don't provide preloaded HSTS we should include this domain, because the first request is over insecure http and cloud be compromised.

>>Sadly it seems that every customer/server gets a subdomain too, but without a valid cert. Could you document this in a comment for future contributors?

>Every customer that decides not to buy a domain gets a subdomain. It also seems that they get a subdomain on puslapiai.lt, not serveriai.lt.


e.g. http://laidara.lt.povas.serveriai.lt/

Liudvikam:
>>> You may also add https://roundcube.serveriai.lt/

>> The server seems to redirect to HTTPS on that subdomain all by itself.

> As long as they don't provide preloaded HSTS we should include this domain, because the first request is over insecure http and cloud be compromised.

Ah, that's right. I'll add that subdomain, then.



>>> Sadly it seems that every customer/server gets a subdomain too, but without a valid cert. Could you document this in a comment for future contributors?

>> Every customer that decides not to buy a domain gets a subdomain. It also seems that they get a subdomain on puslapiai.lt, not serveriai.lt.

> e.g. http://laidara.lt.povas.serveriai.lt/

I see. I'll go ahead and add that comment.

J0WI:
Great, thank you!

PR #3897, opened 2016-01-15
Create Baidupcs.com.xml

renyouguo:


fuglede:
Thanks for the PR. The build fails do to the slash in the target host;

      failure: nbct01.baidupcs.com/ failed: The target host must be a hostname, not URL, and must use at most one wildcard.

PR #3899, opened 2016-01-15
[DN.dk] New ruleset

fuglede:
Most references seem to be securable, but all content pages redirect to HTTP.

### DIFF
--- a/src/chrome/content/rules/DN.dk.xml
+++ b/src/chrome/content/rules/DN.dk.xml
@@ -1,18 +1,21 @@
 <!--
 	Non-functional domains:
 
+		- fordel ¹
+		- historie ¹
 		- tur ¹
 
 	¹: Refused
 -->
-<ruleset name="DN.dk">
+<ruleset name="DN.dk (partial)">
 	<target host="dn.dk" />
 	<target host="www.dn.dk" />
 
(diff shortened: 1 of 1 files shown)


J0WI:
fordel and historie are also refused.
The [Dd]efault.aspx is optional (but https://www.dn.dk/?ID=7244 also redirects to http)

I would mark the rule as partial since it only loads resources over https.

fuglede:
Agreed on all accounts.

src/chrome/content/rules/DN.dk.xml
```diff
@@ -0,0 +1,24 @@
...
+		- historie ¹
+		- tur ¹
+
+	¹: Refused
+-->
+<ruleset name="DN.dk (partial)">
+	<target host="dn.dk" />
+	<target host="www.dn.dk" />
+
+	<exclusion pattern="^http://www\.dn\.dk/([dD]efault\.aspx|\?).*" />
```
J0WI: This also matches ```http://www.dn.dk/default.aspxfoo```. Do we care about it?

fuglede: I didn't come across any parts of the site where that would be a problem, at least.

PR #3901, opened 2016-01-16
Create Baidupcs.com.xml

renyouguo:


semenko:
Looks like you copied some plus signs into this file. Would you please update without them? Thanks! :)

Issue #3864, opened 2016-01-12
Bad redirect for http://telepathy.freedesktop.org/doc/book/


benwaffle:
http://telepathy.freedesktop.org/doc/book/

PR #3907, opened 2016-01-17
[FreeDesktop.org] Remove redundant rules, add exceptions.

youdly:
Fixed https://github.com/EFForg/https-everywhere/issues/3864


semenko:
Awesome fixes -- thanks! :100:

PR #3911, opened 2016-01-17
[flokinet.is] New ruleset

youdly:


fuglede:
This causes a redirect loop on http://www.billing.flokinet.com/index.php and related pages.

youdly:
Added exception for affected pages. (Reminds me of https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/SwitchVPN.xml , same software perhaps)

semenko:
I think travis is confused. Merging anyway & watching on master.

PR #3916, opened 2016-01-17
Add a few strings for localisation in the 'about' window

fuglede:
This changes a few hardcoded strings so as to allow for localisation. ~~Obviously the way this is done is not ideal, though, as translations are handled through a torproject.org based git submodule. The corresponding changes to this, which in turn is based on Transifex in a way that I do not remember, would be of the same form as those in https://github.com/fuglede/torproject_translation/commit/67875f8bf227798d8b46b77a4f6557e70cd365c3 which contains the changes to the English and Danish language strings. Cc @jsha for review/lesson on what I should have actually done ...~~

Scratch that; I remembered the logic.

### DIFF
--- a/src/chrome/content/about.xul
+++ b/src/chrome/content/about.xul
@@ -43,17 +43,17 @@
     
     <groupbox>
       <caption label="&https-everywhere.about.created_by;" />
-      <label>Mike Perry, Peter Eckersley, and Yan Zhu</label>
+      <label>Mike Perry, Peter Eckersley&https-everywhere.about.and; Yan Zhu</label>
     </groupbox>
     
     <groupbox>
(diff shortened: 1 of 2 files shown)


fuglede:
@semenko: Do you have access to the Transifex setup? This one depended on having something like https://github.com/fuglede/torproject_translation/commit/67875f8bf227798d8b46b77a4f6557e70cd365c3 in the submodule; without it, this breaks the about window:


master-broken

semenko:
Whoops! I don't have access to the Transifex setup (chrome handles missing translations more gracefully). Thoughts on what to do here?

fuglede:
@semenko: Maybe just revert and reopen until @jsha gets the opportunity to have a look; I believe he has got it automated.

Issue #1775, opened 2015-05-27
HTTPS Everywhere uses too much memory


bcroq:
*Member note: One straightforward fix for this is [described here (issue comment)](https://github.com/EFForg/https-everywhere/issues/1775#issuecomment-256973768), along with an example PR in the Firefox codebase*

I use HTTPS Everywhere Chrome 43.0.2357.65 (64-bit Linux) and the extension eats a little more than 100MB of memory. Isn't it too much for this kind of extension?

reedy:
What do you mean by "this kind of extension"?

Noting the raw XML rulesets totalling 62M

bcroq:
When I wrote "this kind of extension" I had uBlock in mind. This extension has 53869 filters on my computer and uses 68MB of memory.

jsha:
@bcrow: Yes, my gut feeling is that HTTPS Everywhere can and should be optimized for better memory and CPU use. We haven't had the time to do it yet, but would very much welcome help from someone who is experienced at profiling extensions.

The uBlock number you quote is actually encouraging! HTTPS Everywhere has 14.5k 'rulesets', but each ruleset contains many targets and rules, which are presumably more akin to uBlock's filters. There are 39361 rules and 41287 targets, for a total of ~80k entities. The memory usage for those entities sounds proportional to uBlock, which is a surprise to me. Still, we can probably do a better job of keeping inactive rulesets out of memory.

foresto:
For the record, this is not a Chrome-specific issue. In my Firefox profile, HTTPS Everywhere 5.0.5 is using almost seven times as much memory as uBlock Origin 0.9.8.2.

jsantell:
Scoping this out for Firefox, will report back what I find

imdjh:
Taking 100+ MB usage for my 44.0.2403.89 chromium on Debian 8.1, installed via chrome webstore.

widefox:
Agree uBlock has shown the way forward to reduce memory usage. Can we get this prioritised?

stokito:
Maybe it is possible to make online service API and make pre-request to fetch a rule?
Also it may be a small cache for last used domains.

txtsd:
It would really help if this issue were prioritized. HTTPS Everywhere uses more RAM than any of my tabs or extensions on Chrome.

![chrome task manager](http://i.imgur.com/6Ol1HmW.png)

jaens:
200M memory usage, this is getting a bit excessive.

[...]

PR #3918, opened 2016-01-17
Memory usage improvements

semenko:
(For #1775 & inspired by @blockoperation's awesome work in https://github.com/blockoperation/https-everywhere/tree/chromium-memory-usage-v2 )

Don't create rule arrays when unneeded.


semenko:
@jsha ptal

PR #3921, opened 2016-01-18
[ProtonMail] Combine two rulesets

lucaspetter:
- Combines the ProtonMail.ch.xml and ProtonMail.com.xml rulesets into one. Since protonmail.ch redirects to protonmail.com, they are the same website, so we don't need two separate rulesets. A single ruleset will also make future maintenance easier.
- Adds securecookie coverage for all ProtonMail domains. Previously, there was no securecookie tag for protonmail.com, only for protonmail.ch.
- Adds a test URL for app.protonmail.ch. Some users [choose](https://protonmail.com/support/knowledge-base/what-is-the-difference-between-protonmail-com-and-protonmail-ch/) to log in there instead of at protonmail.com.
- Changes the securecookie regex from `.` to `.+` for consistency, as per the [Ruleset Style Guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

(This is an updated attempt at #3888 (closed), with clearer git history, better explanations, and a simpler regex.)


semenko:
Awesome -- nice work! :fireworks:

lucaspetter:
Thank you!

semenko:
Thanks for the rules! :100: :100: :tada:

PR #3929, opened 2016-01-18
Update Booking.com, add Villas.com (same company)

TvdW:
Hi, Booking.com dev here. Nowadays everything should take HTTPS properly, let's update the rules to match that!


semenko:
Awesome -- thanks!

PR #3937, opened 2016-01-19
[Libreboot.org] New ruleset

jetomit:


semenko:
Awesome -- thanks! :100:

PR #3943, opened 2016-01-20
Add baomoi.com

chatch:


semenko:
Thanks! I'll merge this, but in general, we don't like adding "default_off" rules -- they generally just add clutter & technical debt.

PR #3946, opened 2016-01-20
Create Dragonair.xml

Riotism:
Split from #3835.

### DIFF
--- a/src/chrome/content/rules/Dragonair.xml
+++ b/src/chrome/content/rules/Dragonair.xml
@@ -3,6 +3,6 @@
 	<target host="dragonair.com" />
 	<target host="www.dragonair.com" />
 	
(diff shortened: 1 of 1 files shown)


semenko:
Tests fail:

```
../src/chrome/content/rules/Dragonair.xml:6: element rule: Relax-NG validity error : Element rule failed to validate attributes
../src/chrome/content/rules/Dragonair.xml fails to validate
Two very common reasons for this are the following:
- Missing caret (^) in 'from' attribute: it should be "^http:" and not "http:".
- Missing trailing slashes in 'from' or 'to' when specifying full hostnames: it should be "https://eff.org/" and not "https://eff.org".
ERROR: Validation of rulesets against relaxng.xml failed.
+rm -r /tmp/tmp.IyQl6WkKI5
``

semenko:
Awesome -- thanks! :)

PR #3883, opened 2016-01-14
Remove match_rule and f attributes.

jsha:
They're no longer used.

cc @semenko for review.

PR #3948, opened 2016-01-20
Revert #3883.

jsha:
This caused a regression: Rules that are active but not taking effect (formerly
moot rules) are simply not listed, when my intention was to list them (i.e.,
match the Chrome behavior).

Will roll forward again when I have a fix, but didn't want to leave things in a
broken state until then.

cc @semenko for review.


semenko:
Hm, this is FF only behavior, yes? I so, could we do a partial revert? That chrome code in `chromium/rules.js` seems dead.

jsha:
Updated to leave the Chromium code deleted.

semenko:
Cool! Thanks -- will re-merge when tests pass

Issue #2926, opened 2015-09-18
Publish https-everywhere-rewriter


RReverser:
Would you be open to publishing [rewriter](https://github.com/EFForg/https-everywhere/tree/master/rewriter) to npm? We would like to use it for custom tool testing needs, and would be very convenient not to include it from git submodule just for that.

RReverser:
Ideally would be nice to have `ruleSets` exposed out-of-module as well.

jsha:
Yes, I'd definitely be open to that. Would you be interested in being the maintainer for the npm module?

RReverser:
Sure, why not.

RReverser:
@jsha It looks like `rewriter` got broken due to restructuring of `chromium` page and changes to `RuleSets` constructor, so our code that used it, broke :( (one more reason to keep them as separate packages)

RReverser:
@jsha @fuglede On a related note, would you be open to modularizing https-everywhere a bit more? In particular, splitting rules, browser extension and generic code into own submodules so that one could safely update XML data without code breakage and so?

fuglede:
That idea actually aired in https://github.com/EFForg/https-everywhere/issues/3999#issuecomment-175782777.

Personally, I have no strong opinions on the matter; my first thought was that it might make doing projects like https://github.com/chris-barry/darkweb-everywhere a bit more straightforward to maintain, but then again, that's not really a very strong argument for anything.

Issue #2640, opened 2015-08-22
Try implementing HTTPS Everywhere using WebExtensions


jsha:
According to this Mozilla blog post, Firefox is moving to a new WebExtensions model, which is supposed to be compatible with Chrome extensions: https://blog.mozilla.org/addons/2015/08/21/the-future-of-developing-firefox-add-ons/.

We have a long lead time before XPCOM is deprecated, and there are some big issues to solve, like how to make SSL Observatory work in such a world (probably we need Mozilla to implement some extra APIs).

We should try implementing HTTPS Everywhere using the new API, to see if it is possible. Probably starting from the Chromium code rather than the existing Firefox code makes the most sense.

Mikaela:
I was wondering whether SSL Observatory is coming to Chrome or disappearing from Firefox too at https://github.com/EFForg/https-everywhere/issues/1681#issuecomment-137374049 and I was linked here after asking at OFTC and this issue answers my question.

> We have a long lead time before XPCOM is deprecated, and there are some big issues to solve, like how to make SSL Observatory work in such a world (probably we need Mozilla to implement some extra APIs).

Would *[Proposal: native.js to “embrace & extend” the WebExtensions API](https://discourse.mozilla-community.org/t/proposal-native-js-to-embrace-extend-the-webextensions-api/3457)* help if Mozilla implements it?

semenko:
So I took a cursory look here, and it'll be surprisingly easy to adapt the Chrome code to support the new WebExtensions model, and move to one unified extension codebase.

There's great documentation here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions

And some issues to look at here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Chrome_incompatibilities

As of now, it looks like:
- SSL Observatory support may die: It's never going to work in Chrome, and it may stop working in Firefox without a new Mozilla API (see above comment) 
- We'll need to change some handlers to be platform specific and listen in onBeforeSendHeaders in Firefox -- Chrome lets us rewrite requests earlier (see https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest)
- It's not clear we can catch redirect loops in Firefox

vgturtle127:
@semenko Can you explain why in detail that we won't be able to have the SSL Observatory? I saw and read the links, but I don't understand what you're referring to.

semenko:
[...]

PR #3952, opened 2016-01-21
Simplify caching using ES6 Map()

semenko:
With EMCA2015, we can use a simple Map() instead of the more complex LRU engine.

This is no longer an LRU (now a FIFO), but seems to be more performant -- and it's certainly cleaner.

(+ minor cleanups.)


semenko:

jsha:
I definitely like the simplification. It's interesting that `Map.prototype.keys()` returns keys in insertion order rather than random or alphabetically. I didn't know that! It's probably worth commenting about it on the line where you remove.

You said it seems to be more performant. Is that based on performance testing, or general feel? If the former, can you share your methodology?

chromium/rules.js
```diff
@@ -287,6 +282,10 @@ RuleSets.prototype = {
 
     // Insert results into the ruleset cache
     this.ruleCache.set(host, results);
+
+    // Cap the size of the cache. (Limit chosen somewhat arbitrarily)
+    if (this.ruleCache.size > 1000) { this.ruleCache.delete(this.ruleCache.keys().next().value) }
```
jsha: Newline after the brace, please.

chromium/rules.js
```diff
@@ -357,6 +356,9 @@ RuleSets.prototype = {
     var nonce_path = "/" + Math.random().toString();
     var test_uri = "http://" + domain + nonce_path + nonce_path;
 
+    // Cap the size of the cookie cache (limit chosen somewhat arbitrarily)
+    if (this.cookieHostCache.size > 250) { this.cookieHostCache.delete(this.cookieHostCache.keys().next().value) }
```
jsha: Same here.

semenko:
Thanks for the review! Making those changes and will rebase after you merge #3961 since they'll conflict on a line or two.

Alas, performant is based on a squishy "pseudo-testing" with Chrome's Javascript CPU profiler while opening sets of tabs (in a fresh profile) and observing the % cpu & timing info.

I'd love a better way to do this programmatically, or in Travis -- any thoughts?

jsha:
Another question: I think this is the first hard dependency on ECMA2015 in the Chrome code, right?

I would like, at some point not too far off, to move the rules code into a shared directory, and migrate the Firefox extension to it, so we can get the benefit of these recent performance improvements (and generally cleaner code). We need to support Firefox ESR for the purposes of the Tor browser, and well as Firefox latest. If you could investigate whether `Map()` and similar ECMA2015 features are available in ESR, I'd greatly appreciate it. If they're not, I'd prefer to use other functionality where possible.

semenko:
Yep -- this is the first hard ES6 dependency, and it's compatible with FF38 ESR, per: https://kangax.github.io/compat-table/es6/

I commented a bit in #2640 too -- I'm trying to clean the Chrome implementation and stick to APIs (unrelated to this PR) that are supported by FF's new WebExtension model.

jsha:
LGTM. Needs merge.

semenko:
Merging per above LGTM.

RReverser:
This broke [rewriter](https://github.com/EFForg/https-everywhere/issues/2926#issuecomment-181845601) that is used for various tests.

semenko:
Cool -- any thoughts on fixing it? (I'm not that familiar with how it works -- though it looks like you could just rip out the lrucache reference.)

RReverser:
@semenko Looks like pretty much that's it (plus fix RuleSets constructor - with new approach it should receive smth like `Object.create(null)` by default?). This fix should be easy, but more generally, as mentioned in that comment, would be nice to have separate parts of HTTPS Everywhere decoupled so that they could be used and updated independently without future code breakages. Do you think that would be possible?

semenko:
Yeah, it's definitely something good to do, though we'll need think about the best way to do it while preserving git history, etc.

RReverser:
@semenko Git history can be rewritten AFAIK - e.g. you can see entire UNIX history converted to Git few days ago https://github.com/dspinellis/unix-history-repo/commits/master?page=3951

PR #3958, opened 2016-01-21
Regexp reduction for trivial rules + cookie names

semenko:
Performance & memory usage improvements (for #1775), inspired in part by @blockoperation's suggestions.

- For trivial rules (`^http: -> https:`), set the Rules values equal to null (and handle null as a trivial rule), instead of creating a new, silly RegExp for every single one
- For trivial cookie names (~50% of cookie names are `.*` or `.+`), do the same

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -469,7 +469,7 @@ function switchPlannerDetailsHtmlSection(tab_id, rewritten) {
  * */
 function onCookieChanged(changeInfo) {
   if (!changeInfo.removed && !changeInfo.cookie.secure && isExtensionEnabled) {
-    if (all_rules.shouldSecureCookie(changeInfo.cookie, false)) {
+    if (all_rules.shouldSecureCookie(changeInfo.cookie)) {
       var cookie = {name:changeInfo.cookie.name,
                     value:changeInfo.cookie.value,
                     path:changeInfo.cookie.path,
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -284,18 +284,17 @@ RuleSets.prototype = {
 
   /**
    * Check to see if the Cookie object c meets any of our cookierule criteria for being marked as secure.
-   * knownHttps is true if the context for this cookie being set is known to be https.
    * @param cookie The cookie to test
-   * @param knownHttps Is the context for setting this cookie is https ?
    * @returns {*} ruleset or null
    */
-  shouldSecureCookie: function(cookie, knownHttps) {
+  shouldSecureCookie: function(cookie) {
     var hostname = cookie.domain;
     // cookie domain scopes can start with .
-    while (hostname.charAt(0) == ".")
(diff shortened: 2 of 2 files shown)


semenko:

jsha:
I'm uncomfortable with using null as a sentinel value with special meaning. If we instantiate a single RegExp value and create references to it, that should achieve the memory benefit we want. That leaves the question of whether it's faster to do the manual replacement than it would be to do the RegExp replacement. If they're about the same speed I'd prefer to keep it as a RegExp just to keep the code complexity down. Could you benchmark the difference?

Assuming the manual replacement is faster than RegExp, I'd like to a define a special sentinel value, e.g.

```
var simpleRule = {};
...
if (rule === simpleRule) {
```

semenko:
Good idea -- I'll give that a try. I'd bet the speed difference is ~negligible.

semenko:
@jsha PTAL. Removed conflicts, and switched to shared `const` regexp for code clarity.

The ES6 `const` feature is supported in FF 38 ESR and higher, per https://kangax.github.io/compat-table/es6/

semenko:
@jsha Friendly ping here, too :+1:

funkydude:
What about rules with `name="."` should they not also be considered?

semenko:
@funkydude Hm, you may have found a fun error!

A quick grep shows we have ~3,700 of those rules.

I don't think they're behaving as people expected -- a `name="."` rule will only match cookies with a single character name.

This is definitely something we should fix --  though in the rulesets themselves + ruleset testing, instead of mopping up everything in the extension code.

funkydude:
@semenko If I understand the code correctly your assumption that it won't work would only be true if it were `name="^.$"` but as it stands using `"."` in this case will have the exact same effect as `".*"`

What I'm trying to say is the rules actually DO work because they are simply matching 1 character of anything, but I think? you are assuming that it will only match cookies with 1 character.

semenko:
Yep, good call. Thoughts on fixing the heterogeneity? Require cookierules have no name attribute if they should matchable names? 

(There's also a fun edge case here, since cookies can have *no* name, too.)

funkydude:
It would be great if we could avoid having to set a name in the rules, which would just allow any named or unnamed cookie and skip a regex check altogether. But for now, my suggestion is to add `"."` to the if statement [here](https://github.com/semenko/https-everywhere/blob/c654124eac405408602842fbb0a7ca86a26cf987/chromium/rules.js#L47).

semenko:
Mind sending a tiny pull request? :)

funkydude:
@semenko https://github.com/EFForg/https-everywhere/pull/4058

funkydude:
@semenko https://github.com/EFForg/https-everywhere/pull/4064
Some rules match any host also, maybe there should be some way to simply flag "allow all".

PR #3959, opened 2016-01-21
ubi.com updates

galeksandrp:
- [x] Add legal.
- [x] Remove ghost-recon., tomclancy-thedivision. due to mixedcontent
- [x] Thanks `@akamai` and `@aws` that they don't rely on his clients
and use additional https urls.


semenko:
Awesome! Thanks :tada:

PR #3961, opened 2016-01-22
Expand "use strict" use & drop (broken) old Opera mixedcontent support

semenko:
I found this fun bug while expanding "use strict" for some ES6 features:

Because of the IIFE (immediately-invoked function expression) surrounding `localPlatformRegexp`, it's never worked. `this.userAgent` is undefined during execution, and the else is always taken, resulting in `chromium` being the platform -- always.

Let's just rip this out, since it hasn't worked and `mixedcontent` support has been dead for a while in Chrome (and FF 23+, and Opera 23+).

In the future, we should consider allowing Chrome users to enable `mixedcontent` rules, similar to the FF behavior.

Signed-off-by: Nick Semenkovich \<semenko@alum.mit.edu\>

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -1,3 +1,4 @@
+"use strict";
 // Stubs so this runs under nodejs. They get overwritten later by util.js
 var DBUG = 1;
 function log(){};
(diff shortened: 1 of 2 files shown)


semenko:

chromium/rules.js
```diff
@@ -274,6 +261,7 @@ RuleSets.prototype = {
     }
     // now eat away from the left, with *, so that for x.y.z.google.com we
     // check *.z.google.com and *.google.com (we did *.y.z.google.com above)
+    var t;
```
jsha: `t` should be scoped inside the for loop, right?

jsha:
LGTM except for that one nit.

semenko:
Fixed -- thanks!

semenko:
Merging per LGTM above.

PR #3962, opened 2016-01-22
[filecrypt.cc] New ruleset

Balgden:


semenko:
:fireworks: thanks!

rolak4gh:
This ruleset prevents successfull Click'n'Load when cooperating with JDownloader - [example](https://filecrypt.cc/Container/14e1fu7vew0a0jy), password is "serienjunkies.org" (without the "" as usually). Via the incorporated [script](https://filecrypt.cc/js/container.js?9823982) the intermediate page "http://filecrypt.cc/helper.html?..." executes a "_DOMAIN.replace('https:', 'http:')", obviously with reasons.

PR #3971, opened 2016-01-23
[Mumble.info] New ruleset

rugk:


J0WI:
You missed forums.mumble.info and nonfunctional stats.mumble.info

rugk:
Thanks. Did not know these.

PR #3973, opened 2016-01-23
add rule for Voobly

numismatika:


J0WI:
I get a an error from CloundFlare when I try to access the site over https.
There is also a bunch of subdomains that don't support https.

semenko:
Closing given error above.

PR #3976, opened 2016-01-24
Fixed BECU.org ruleset.

breadtk:
BECU.org appears to have enabled HTTPS throughout its entire website. This pull request re-enables the BECU rule with new tests and using the domains that they still use.


J0WI:
Thanks for updating the rule! You don't need test urls as long as they are already covered by the target hosts.

I found a few more hosts. It would be great if you could add them too and add comments for the broken ones.

Working:
onlineappointments.becu.org
accessassistant.becu.org

Mismatch:
locatorsearch.becu.org

Timeout:
news.becu.org

breadtk:
Closing this request for now.

PR #3979, opened 2016-01-24
[butterflynetinc.com] New ruleset

youdly:


J0WI:
The cert doesn't match www.butterflynetinc.com

youdly:
There is a new cert that begins on 10 March 2016, should be working by now.

PR #3983, opened 2016-01-24
[automattic.com] New ruleset

youdly:


J0WI:
This domain was added in 73b763ffb70fe06e5ab4514c6eb3badc53a0aada and removed in 143128b3f9b647c3a07aa3063239e42f6a3e055f

How do we handle this?
@fuglede @semenko @jsha

semenko:
We don't have a great way of handling this :(

We might want to rethink our mechanism for storing rules, especially as TLS adoption expands. Our current "throw a bunch of .xml files in a folder" strategy could definitely be improved.

fuglede:
To make sure I understand: what is the issue exactly? That one is not automatically exposed to the information about the ruleset having existed previously, or something more concrete than that?

semenko:
I think that's correct. Pondering some other rule storage options that might address this.

fuglede:
I don't necessarily think that that's a big issue, but one simple way out of it -- if we were better at keeping disabled rulesets out of memory at least -- would be to only ever default_off ruleset files instead of deleting them.

semenko:
Agreed -- though I do think we should ponder reorganizing our ruleset chaos. I'll make an issue at some point...

J0WI:
My question was about how we handle such scenarios:
Reverting the remove or just merging the new one?

PR #3984, opened 2016-01-24
[theaustralianalps.wordpress.com] Remove redundant rule

youdly:
Covered by https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/WordPress.xml


J0WI:
Thanks!

PR #3986, opened 2016-01-24
[tenderapp.com] Apply ruleset to all subdomains, test URLs.

youdly:


J0WI:
Could you also add the www test domain? LGTM then.

youdly:
Added www test domain.

PR #3996, opened 2016-01-25
Add benchmarkjs.com

mathiasbynens:


fuglede:
Thanks for the PR! I can see what you were trying to do with the rule, but for it to work with the current syntax, it would have to be something like

    <rule from="^http://(?:www\.)?benchmarkjs\.com/" to="https://benchmarkjs.com/" />

mathiasbynens:
@fuglede Thanks! Patch updated.

PR #3998, opened 2016-01-25
Uncomment formerly rc4-only viginmedia.com subdomains

Liudvikam:
national.virginmedia.com and identity.virginmedia.com now support proper cipher suites and no longer support RC4 (see links below), so they should be uncommented in Virgin Media's ruleset.

https://www.ssllabs.com/ssltest/analyze.html?d=identity.virginmedia.com

https://www.ssllabs.com/ssltest/analyze.html?d=national.virginmedia.com


Liudvikam:
Thanks!

PR #4016, opened 2016-01-28
Hide rules if extension is disabled [Chrome]

### LINKED ISSUE #4014: Disabling HTTPS Everywhere should hide menu elements
When HTTPS Everywhere is disabled, menu elements in the popup are still visible. They should be hidden instead.

Probably we should leave "What's this" and the version string visible, though.

cc @semenko

semenko:
PTAL @jsha 

Closes #4014

### DIFF
--- a/chromium/popup.html
+++ b/chromium/popup.html
@@ -1,62 +1,69 @@
+<html>
 <head>
-<title i18n="about_ext_name"></title>
+    <title i18n="about_ext_name"></title>
 
-<!-- Local copy of chrome://resources/css/chrome_shared.css -->
-<link href="chrome-resources/css/chrome_shared.css" rel="stylesheet" type="text/css"/>
+    <!-- Local copy of chrome://resources/css/chrome_shared.css -->
+    <link href="chrome-resources/css/chrome_shared.css" rel="stylesheet" type="text/css"/>
 
-<link href="popup.css" rel="stylesheet" type="text/css"/>
-<script src="popup.js"></script>
+    <link href="popup.css" rel="stylesheet" type="text/css"/>
+    <script src="popup.js"></script>
 
 </head>
 
 <body>
 <header>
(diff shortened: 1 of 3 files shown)


jsha:
This leaves the "Block all HTTP requests" UI visible, and allows you to check and uncheck it (but it doesn't take effect when HTTPS Everywhere is disabled).

chromium/popup.js
```diff
@@ -77,9 +77,14 @@ function createRuleLine(ruleset) {
   return line;
 }
 
-// Change the UX to reflect extension enabled/disabled
-function updateEnabledDisabledButtonState() {
+// Change the UI to reflect extension enabled/disabled
+function updateEnabledDisabledUI() {
   document.getElementById('onoffswitch').checked = backgroundPage.isExtensionEnabled;
+  // Hide or show the rules sections
```
jsha: Probably makes more sense to set a class on some top-level element within the popup (html, or body?), and have CSS to hide elements when that class is present.

semenko: Updated!

chromium/popup.html
```diff
@@ -1,62 +1,69 @@
...
-</section>
-<section id="StableRules" class="rules">
-  <h3 i18n="chrome_stable_rules"></h3>
-  <p class="description" i18n="chrome_stable_rules_description"></p>
-</section>
-<section id="UnstableRules" class="rules">
-  <h3 i18n="chrome_experimental_rules"></h3>
-  <p class="description" i18n="chrome_experimental_rules_description"></p>
-</section>
+<div id="RuleManagement">
```
semenko: The diff here isn't pretty. It's just indented correctly, plus this div.

jsha: You can add ?w=1 to a diff to ignore whitespace, which is very helpful!

semenko: Holy crap -- thanks!

PR #4018, opened 2016-01-28
[nugg.ad] Fix subdomain handling for nuggad.net

snadorp:
We are serving for http and https versions for all our tracking subdomains. I've also added our main company site.


semenko:
Awesome -- thanks!

Issue #3963, opened 2016-01-22
Fix World Community Grid


funkydude:
In regards to this line: https://github.com/EFForg/https-everywhere/blob/bcb101723edb31a933b1f4e26cce17c2364da31d/src/chrome/content/rules/WorldCommunityGrid.xml#L15

It redirects to the `secure` subdomain, this is not needed. The main domain now fully supports HTTPS (a recent change by the website owners)

PR #4019, opened 2016-01-28
[WorldCommunityGrid] Don't redirect to "secure" subdomain

funkydude:
Stop redirecting worldcommunitygrid.org requests to
secure.worldcommunitygrid.org. The former now supports HTTPS.

Closes https://github.com/EFForg/https-everywhere/issues/3963


J0WI:
The redirect for ^worldcommunitygrid.org does not work for deep links. You should redirect it to www or secure then.

funkydude:
@J0WI Good catch, I saw that worldcommunitygrid.org redirected to www.worldcommunitygrid.org and figured it would be the same for deep links also. I'll adjust it to www.

PR #4029, opened 2016-01-29
Expand use of Set()

semenko:
Replace blacklists and potentiallyApplicableRulesets with actual Set()s, trying to clean up the Chrome code.

Compatible with FF 38 ESR and above per https://kangax.github.io/compat-table/es6/

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -229,8 +229,8 @@ RuleSets.prototype = {
     var tmp;
     var results = [];
     if (this.targets[host]) {
-      // Copy the host targets so we don't modify them. // TODO: Check if this comment makes sense.
-      results = results.concat(this.targets[host].slice());
+      // Copy the host targets so we don't modify them.
+      results = results.concat(this.targets[host]);
     }
 
     // Replace each portion of the domain with a * in turn
@@ -348,7 +348,9 @@ RuleSets.prototype = {
     log(INFO, "Testing securecookie applicability with " + test_uri);
     var potentiallyApplicable = this.potentiallyApplicableRulesets(domain);
     for (let ruleset of potentiallyApplicable) {
-      if (!ruleset.active) continue;
+      if (!ruleset.active) {
+        continue;
+      }
       if (ruleset.apply(test_uri)) {
         log(INFO, "Cookie domain could be secured.");
         this.cookieHostCache.set(domain, true);
@@ -370,8 +372,9 @@ RuleSets.prototype = {
     var newuri = null;
     var potentiallyApplicable = this.potentiallyApplicableRulesets(host);
     for (let ruleset of potentiallyApplicable) {
-      if (ruleset.active && (newuri = ruleset.apply(urispec)))
+      if (ruleset.active && (newuri = ruleset.apply(urispec))) {
         return newuri;
+      }
     }
     return null;
   }
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -213,19 +213,6 @@ RuleSets.prototype = {
   },
 
   /**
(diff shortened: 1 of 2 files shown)


chromium/rules.js
```diff
@@ -242,8 +229,8 @@ RuleSets.prototype = {
     var tmp;
     var results = [];
     if (this.targets[host]) {
-      // Copy the host targets so we don't modify them.
-      results = this.targets[host].slice();
+      // Copy the host targets so we don't modify them. // TODO: Check if this comment makes sense.
```
jsha: I'm pretty sure you no longer need the `.slice()` here:


> [Strings and numbers (not String and Number objects): concat copies the values of strings and numbers into the new array.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat)

chromium/rules.js
```diff
@@ -251,24 +238,31 @@ RuleSets.prototype = {
...
     // now eat away from the left, with *, so that for x.y.z.google.com we
     // check *.z.google.com and *.google.com (we did *.y.z.google.com above)
     for (var i = 2; i <= segmented.length - 2; ++i) {
       var t = "*." + segmented.slice(i,segmented.length).join(".");
-      this.setInsert(results, this.targets[t]);
+      results = results.concat(this.targets[t]);
     }
+
+    // Clean the results list, which may contain duplicates or undefined entries
+    var resultSet = new Set(results);
```
jsha: Why not just create the Set as you go?

chromium/rules.js
```diff
@@ -353,10 +346,10 @@ RuleSets.prototype = {
     }
 
     log(INFO, "Testing securecookie applicability with " + test_uri);
-    var rs = this.potentiallyApplicableRulesets(domain);
-    for (var i = 0; i < rs.length; ++i) {
-      if (!rs[i].active) continue;
-      if (rs[i].apply(test_uri)) {
+    var potentiallyApplicable = this.potentiallyApplicableRulesets(domain);
+    for (let ruleset of potentiallyApplicable) {
+      if (!ruleset.active) continue;
```
jsha: As long as you're touching this code, would you put brackets around this `if`?

chromium/rules.js
```diff
@@ -375,9 +368,9 @@ RuleSets.prototype = {
    */
   rewriteURI: function(urispec, host) {
     var newuri = null;
-    var rs = this.potentiallyApplicableRulesets(host);
-    for(var i = 0; i < rs.length; ++i) {
-      if (rs[i].active && (newuri = rs[i].apply(urispec)))
+    var potentiallyApplicable = this.potentiallyApplicableRulesets(host);
+    for (let ruleset of potentiallyApplicable) {
+      if (ruleset.active && (newuri = ruleset.apply(urispec)))
```
jsha: Here too?

jsha:
Generally LGTM modulo the comments above.

chromium/rules.js
semenko: I had this in a previous version, but I don't know of a clean way to add a list to a set (and we have to test that we actually have a list to iterate over).

e.g.
```
result = result.concat(this.targets[segmented.join(".")])
```
becomes
```
if (this.targets[segmented.join(".")]) {
   this.targets[segmented.join(".")].forEach(function(s) {resultSet.add(s)});
}
```

Maybe I'm missing some obvious approach here?

jsha: Hm, you're right. I though maybe we could workaround by storing the targets as sets too, but it looks like there is no actual support in ES6 for Set unions, and you have to convert to arrays and back again(??). So, this seems fine.

PR #4036, opened 2016-01-30
[Wj32.org] New ruleset

rugk:
Some notes for the web admin @wj32:
* Pretty bad SSL config (already submitted you some more information)
* On https://wj32.org/wp/ there is some mixed content. You can easily fix it as it as it is just on your own site.

### DIFF
--- a/src/chrome/content/rules/Wj32.org.xml
+++ b/src/chrome/content/rules/Wj32.org.xml
@@ -1,0 +1,14 @@
+<!--
+
+-->
+<ruleset name="Let's Encrypt">
+    <target host="wj32.org" />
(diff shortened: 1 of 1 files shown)


J0WI:
You should add the mixed content platform attribute.

rugk:
The mixed content does not break the site. On https://wj32.org/wp/ it is just the ico file.

J0WI:
At least [here](https://wj32.org/wp/about/) we have mixed display content in articles.
But I can live with that since the site sends a HSTS header. It's a bit weird but it works.

rugk:
Yeah, possibly @wj32 notices this issue and can fix it.

PR #4044, opened 2016-01-31
Computerbase.de:

numismatika:
- add test-urls and more files
- remove stylesheets path, not used anymore
- remove wildcard
- fix forum editor

### DIFF
--- a/src/chrome/content/rules/Computerbase.de.xml
+++ b/src/chrome/content/rules/Computerbase.de.xml
@@ -1,5 +1,22 @@
 <ruleset name="Computerbase.de (partial)">
-<target host="*.computerbase.de"/>
 
-<rule from="^http://pics\.computerbase\.de/" to="https://pics.computerbase.de/"/>
+	<target host="pics.computerbase.de"/>
+	<target host="www.computerbase.de"/>
+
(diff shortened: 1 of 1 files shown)


J0WI:
I found some more:

/api/stats
/css/
/login/
/abo/
/push/

They are also working for ^computerbase.de

numismatika:
have you checked that those don't break the page and especially the forum?
Forum was a bit evil on my first attempts.

J0WI:
Resource rewriting should be quite safe. Do you have an account to check the behaviour for logged in visitors?

The following pattern is working for me:
```
<rule from="^http://www\.computerbase\.de/(api/|css/|login/|abo/|push/|clientscripts/|js/|design/|img/|forum/images/|forum/clientscript/)" to="https://www.computerbase.de/$1"/>
```
They are just some broken 404, but this is trivial I think.

You can also safely redirect videos.computerbase.de.

numismatika:
Yes i have, that is how i uncovered the forum issue.
I'll check a few more days before i consider it working.
Shouldn't be a problem though, they offer TLS for their paid users.

numismatika:
so far it looks fine @J0WI

J0WI:
Thanks!

PR #4047, opened 2016-01-31
[hdvest.com] New ruleset

youdly:


src/chrome/content/rules/hdvest.com.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Problematic domains:
+		- (my|www).myhdvest.com		(incomplete cert chain)
```
J0WI: same for my.hdvest.com

J0WI:
Please also add bcp.hdvlink.com

youdly:
Added new subdomain.

PR #4048, opened 2016-01-31
[fastmail.com] Add blog subdomain, remove wildcard

youdly:


J0WI:
I also found:
fbl.fastmail.com (mismatch)
classic.fastmail.com (works)

There are multiple non working customer subdomains on fastmail.fm and user.fm. You should leave some comments about that.

youdly:
> There are multiple non working customer subdomains on fastmail.fm and user.fm. You should leave some comments about that.
Example?

youdly:
Added classic subdomain, awaiting clarification on ruleset comment.

J0WI:
>> There are multiple non working customer subdomains on fastmail.fm and user.fm. You should leave some comments about that.

> Example?

http://joewalton.fastmail.fm/
http://copyrightx.org.user.fm/

youdly:
Updated comment.

PR #4053, opened 2016-01-31
[rezdy.com] New ruleset

youdly:


J0WI:
It seems that every event/location got a own subdomain, so I agree with the wildcard.
Sadly the host does not provide a HSTS header.

PR #4056, opened 2016-01-31
[WineHQ] Update rule

funkydude:
Update WineHQ rule to add "dl" and (now working) "wiki" subdomains.

### DIFF
--- a/src/chrome/content/rules/WineHQ.xml
+++ b/src/chrome/content/rules/WineHQ.xml
@@ -5,6 +5,7 @@
 	<target host="appdb.winehq.org" />
 	<target host="bugs.winehq.org" />
 	<target host="forum.winehq.org" />
+	<target host="forums.winehq.org" />
 	<target host="source.winehq.org" />
 	<target host="test.winehq.org" />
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks for the additions. Could you also add forums.winehq.org?

funkydude:
@J0WI that entry already exists (forum not forums)

J0WI:
Yes, but the subdomain forums does also exist.

funkydude:
@J0WI As far as I can see any attempt to access `forums` is redirected to `forum`. What is the benefit of adding `forums`?

J0WI:
Every unsecured access to forums.winehq.org could be MITM attacked. That's the same reason why we add forum.winehq.org and all other domains, that doesn't support preloaded HSTS.

funkydude:
@J0WI I see, added. How did you come across the forums subdomain anyway?

J0WI:
Thanks!

> How did you come across the forums subdomain anyway?

Just found it on Google, see https://github.com/EFForg/https-everywhere/pull/3517#issuecomment-165904237

PR #4059, opened 2016-01-31
Switch to spanning incognito

### LINKED ISSUE #4034: Settings not synced with/across Chrome incognito sessions
Since the move to using split mode[1] for incognito windows/sessions the settings are not saved for use across different incognito sessions.  They're also not synced properly with the non-incognito extension process as you would expect it to.  This behavior isn't what you really expect from an extension even if it's using split mode.  The settings should still be synced between them and across sessions.  uBlock Origin[2] behaves this way and keeps settings synced between them.

To test what I'm describing:

1. Make sure all chrome processes are closed before starting.
2. Start chrome with the incognito command (i.e. "chrome.exe --incognito").
3. Browse to any site which has a rule and change this rule
4. Close chrome and ensure all processes end.
5. Start Chrome again in incognito mode as in step 2.
6. Browse to the same site as in step 3 and check the rules.  It will revert to the defaults.

You can also change various rules in normal (non-incognito) mode then browse to those same sites in incognito mode and see the settings for the rules are not the same.  The incognito window will be using whatever the default rules are.

There are two possible ways to fix this.  Obviously you could remove the split mode (is it even necessary anymore?) so that it only runs one extension process.  This fixes all problems described above and is the easiest fix provided it would not have security implications.  However if you do want to keep using split mode you'll have to sync the settings between them somehow (e.g. how uBlock Origin does it).

[1] https://developer.chrome.com/extensions/manifest/incognito
[2] https://github.com/gorhill/uBlock

semenko:
Fixes:
- Stops us from ~doubling memory usage on incognito (by instantiating again)
- Rule preferences & custom rules are shared in incognito

This is a little verbose, since:
- There's no easy way to see when incognito sessions are destroyed, and we
don't want to maintain additional state about window IDs.
- We need to clear caches when incognito sessions are destroyed, like Chrome
does for about:net-internals/#dns, for example.

Closes #4034

### DIFF
--- a/chromium/incognito-cache-clearing.js
+++ b/chromium/incognito-cache-clearing.js
@@ -1,0 +1,61 @@
+"use strict";
+// This file keeps track of incognito sessions, and clears any caches after
+// an entire incognito session is closed (i.e. all incognito windows are closed).
+
+let incognito_session_exists = false;
+
+/**
+ * Detect if an incognito session is created, so we can clear caches when it's destroyed.
+ *
+ * @param window: A standard Window object.
+ */
+function detect_incognito_creation(window) {
+    if (window.incognito === true) {
+        incognito_session_exists = true;
+    }
+}
+
+/**
+ * Clear any caches we have.
+ * Called if an incognito session is destroyed.
+ */
+function destroy_caches() {
+    log(DBUG, "Destroying caches.");
+    all_rules.cookieHostCache.clear();
+    all_rules.ruleCache.clear();
+}
+
+/**
+ * Check if any incognito window still exists. If not, destroy caches.
+ * @param arrayOfWindows: A array of all open Window objects.
+ */
+function check_for_incognito_session(arrayOfWindows) {
+    for (let window of arrayOfWindows) {
+        if (window.incognito === true) {
+            // An incognito window still exists, so don't destroy caches yet.
+            return;
+        }
+    }
+    // All incognito windows have been closed.
+    incognito_session_exists = false;
+    destroy_caches();
+}
+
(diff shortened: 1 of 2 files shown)


semenko:
Ptal @jsha (and @TheRyuu).

These APIs are supported by WebExtensions. See:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/Windows/onCreated
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/Windows/onRemoved
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/Windows/getAll

jsha:
This looks nice, but can you double check what the intended promises of incognito mode are?

If the goal is "state is gone when you leave incognito," this accomplishes it.

If the goal is "state does not leak from cognito to incognito," this does not accomplish it: incognito mode windows can theoretically whether you've visited a given site from a cognito window, by the timing information available in the cache.

semenko:
Chrome's consumer description: 
> "Incognito mode only prevents Chrome from saving your site visit activity." https://support.google.com/chrome/answer/95464?hl=en

The developer summary:
> "When in Incognito mode, Chrome on desktop platforms (Windows, Linux, and Mac OS X), Chrome on Android, and Chrome OS do not store any new history, cookies, or other state in non-volatile storage. However, Incognito windows will be able to access some previously-stored state, such as browsing history."
https://www.chromium.org/Home/chromium-security/security-faq#TOC-What-are-the-security-and-privacy-guarantees-of-Incognito-mode-

So this should be good to merge.

PR #4064, opened 2016-02-01
Some rules also match any host, use trivial_cookie_name_c for those also

funkydude:


chromium/rules.js
```diff
@@ -42,7 +42,11 @@ function Exclusion(pattern) {
  * @constructor
  */
 function CookieRule(host, cookiename) {
-  this.host_c = new RegExp(host);
+  if (host === ".*" || host === ".+" || host === ".") {
+    this.host_c = trivial_cookie_name_c;
```
semenko: Could you please make a new `const` above for host rules, e.g. trivial_cookie_host?

semenko:
Awesome -- thanks! LGTM other than the one comment above.

funkydude:
If it's the name that bothers you, then why not rename it to e.g. `trivial_cookie_regex` rather than create another const of the same regex?

semenko:
A single variable is cheap, so please make a new one, since it helps clarity & fits well with the parallelism for rules.

(Long term, it'll also make testing easier, since we can more cleanly test object equivalence.)

funkydude:
Personally I think it's a waste and I'd go for renaming the var but oh well, changed.

semenko:
Awesome - thanks :)

jsha:
LGTM, thanks!

PR #4065, opened 2016-02-01
Fix comment in Virgin_media.com.xml

### LINKED ISSUE #4024: Virgin Media domains no longer require RC4
identity.virginmedia.com and national.virginmedia.com subdomains no longer need to be commented out in Virgin_Media.com.xml as they now support modern cipher-suites and not just RC4 (also in the comments they can be moved to the "Fully covered subdomains" list).

https://www.ssllabs.com/ssltest/analyze.html?d=national.virginmedia.com
https://www.ssllabs.com/ssltest/analyze.html?d=identity.virginmedia.com

Liudvikam:
The target tags for identity.virginmedia.com and national.virginmedia.com were uncommented in   #3998, but the comment in the ruleset wasn't changed to reflect that.

This fixes #4024.

### DIFF
--- a/src/chrome/content/rules/Virgin_Media.com.xml
+++ b/src/chrome/content/rules/Virgin_Media.com.xml
@@ -77,13 +77,6 @@
 	** certificate name mismatch
 	*** certificate expired
 
-	Future nonfunctional subdomains:
-
-		- identity ****
-		- national ****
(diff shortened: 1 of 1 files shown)


Liudvikam:
Thanks!

PR #4067, opened 2016-02-01
Add rule for tele2.lt

Liudvikam:


funkydude:
Don't name your rules generically. Name the XML file tele2.lt.xml and the rule tele2.lt

Liudvikam:
Okay, sure. Done.

J0WI:
There is a lot of mixed content, so you should add `platform="mixedcontent"`
Please remove the www => ^ redirect, because not all deep links get redirected to ^. We only use such redirect if it would run in a error (e.g. cert is not valid for www).

Please also add the following sites:
mano.tele2.lt
loyaloop.tele2.lt

And comments for broken ones:
pagalba.tele2.lt (wrong content)
narsyk.tele2.lt (wrong content)
pyptonas.tele2.lt (refused)
webmail.tele2.lt (refused)
wap.tele2.lt (refused)
mail.tele2.lt (refused)
blog.tele2.lt (self-signed)
web.tele2.lt (self-signed)
parkuok.tele2.lt (self-signed)
internetas.tele2.lt (mismatch)
editorial.tele2.lt (mismatch, expired)

This makes it easier to maintain the ruleset.

Liudvikam:
> There is a lot of mixed content, so you should add `platform="mixedcontent"`

I don't see why, because all of the mixed content is passive, and therefore isn't blocked by browsers and doesn't break the site when loaded over HTTPS. Is there some other reason I haven't thought of?

> Please remove the www => ^ redirect, because not all deep links get redirected to ^. We only use such redirect if it would run in a error (e.g. cert is not valid for www).

Done.

> Please also add the following sites:
> mano.tele2.lt
> loyaloop.tele2.lt
> 
> And comments for broken ones:
> pagalba.tele2.lt (wrong content)
> narsyk.tele2.lt (wrong content)
> pyptonas.tele2.lt (refused)
> webmail.tele2.lt (refused)
> wap.tele2.lt (refused)
> mail.tele2.lt (refused)
> blog.tele2.lt (self-signed)
> web.tele2.lt (self-signed)
> parkuok.tele2.lt (self-signed)
> internetas.tele2.lt (mismatch)
> (www.)editorial.tele2.lt (mismatch, expired)

Done.

J0WI:
@semenko @fuglede how do we handle mixed passive content?

fuglede:
@J0WI: Indeed passive mixed content only triggers the browsers' "This page serves mixed content warning"; when the actual request is performed, it is upgraded just fine. If I recall correctly, it does make a difference when (and thus only when) the user themselves have enabled passive mixed content blocking (e.g. with <code>security.mixed_content.block_display_content</code> in Firefox), in which case it will be blocked falsely along with the active content.

J0WI:
@fuglede so do we require the mixedcontent directive here or not?

fuglede:
@J0WI: If all mixed content is passive, then no. (And on the same note, in the active case, we do so only if it actually breaks the site.)

J0WI:
Okay, thanks for clarify.

Issue #4028, opened 2016-01-29
Https Everywhere breaks Nvidia driver downloads


random6969:
example: http://www.nvidia.com/download/driverResults.aspx/97767/en-us

Clicking "download" will lead to a page with broken formatting. 
Disabling "nVidia.com (partial)" fixes it. 

- chrome (v48) 64-bit, Windows 10 64-bit

funkydude:
This seems like an HTTPSE issue to me. The CSS files Chrome is blocking from being downloaded works fine when accessed over HTTPS (the rule kicks in correctly), however HTTPSE isn't changing the HTTP request to HTTPS.

The workaround would be setting the following exclusion which I have not tested:
`<exclusion pattern="^http://www\.nvidia\.com/(download|content)/" />`

funkydude:
cc @semenko
Am I misunderstanding the functionality or should the requests being made (and blocked by Chrome) not be re-written to HTTPS by the extension?

semenko:
Outstanding chrome bug #793 https://code.google.com/p/chromium/issues/detail?id=122548

funkydude:
@semenko 2012 seriously? Ugh... Should we be working around these Chrome specific issues with exclusions in the rules themselves?

semenko:
Generally yes :(

PR #4068, opened 2016-02-01
[NVidia] Workaround Chrome bug

funkydude:
Untested change that should resolve https://github.com/EFForg/https-everywhere/issues/4028


J0WI:
Note that this is not only chrome specific.

funkydude:
You mean this change will also apply to Firefox? Yes I know, I'm not aware of any way to make an exclude rule truly exclusive to Chrome.

Or do you mean the problem this is trying to workaround also affects other browsers? I assumed it would just be Chromium browsers.

J0WI:
The false positive mixed content triggering affects all browsers AFAIK.

PR #4071, opened 2016-02-02
Make tests headless with the help of xvfb

Hainish:
I'm not sure if this is categorically useful.  There may be instances where you actually want to have the ability to interact with the webdriver browser.  We could also add this as an option, like `--headless` if desired.

### DIFF
--- a/test/chromium.sh
+++ b/test/chromium.sh
@@ -34,5 +34,5 @@ else
 	./makecrx.sh
 	echo "running tests"
 	CRX_NAME="`ls -tr pkg/*.crx | tail -1`"
-	python2.7 test/chromium/script.py $CRX_NAME
+	xvfb-run python2.7 test/chromium/script.py $CRX_NAME
 fi
--- a/test/chromium.sh
+++ b/test/chromium.sh
@@ -11,6 +11,10 @@ else
     git rev-parse && cd "$(git rev-parse --show-toplevel)"
 fi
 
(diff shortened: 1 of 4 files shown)


semenko:
Hey cool, thanks! This definitely could be useful, especially as we're trying to expand testing.

I'll need to think about this a bit more (and not sure what's up with Travis failing)

Hainish:
Since the xvfb daemon was run as a travis `before_install` step, the addition of xvfb-run was causing the tests to fail.  Looks like it's working now..

semenko:
Cool! I'm thinking adding xvfb-run into the actual test scripts is going to break testing on Mac -- thoughts?

Hainish:
This latest commit should fix that.  It won't be headless for OSX users, though.

test/chromium.sh
```diff
@@ -11,6 +11,10 @@ else
     git rev-parse && cd "$(git rev-parse --show-toplevel)"
 fi
 
+# Make sure we have xvfb-run and it's not already set.
+if [ ! -v XVFB_RUN -a -n "$(which xvfb-run)" ]; then
```
semenko: Errors for me:
```
++ which xvfb-run
+ '[' '!' -v XVFB_RUN -a -n '' ']'
./chromium.sh: line 15: [: -v: unary operator expected
```

Hainish:
@semenko try now?

semenko:
Awesome! :fireworks:

PR #4085, opened 2016-02-04
Delete HasGeek.xml

karthikb351:
jobs.hasgeek.com rule no longer needed.


J0WI:
The host had been moved to hasjob.co.
But there are still a lot of hosts on hasgeek.com.
So an updated version of the ruleset would be much more helpful than just removing.

jeremyn:
@J0WI This is over six months old. Is it all right if we merge this in, and I make a new issue saying that hasgeek.com needs a new ruleset?

J0WI:
I would rather disable and rebase it.

jeremyn:
@karthikb351 Are you interested in further updating this ruleset?

karthikb351:
We've enforced HTTPS on all our domains, so you can go ahead and merge this. There is no need for a ruleset for hasgeek.com domains.

jeremyn:
HTTPS Everywhere can cover sites that redirect all HTTP to HTTPS. See [here](https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?).

J0WI:
@karthikb351 or even better: You should consider preloaded HSTS: https://hstspreload.appspot.com/

Hainish:
Since `jobs.hasgeek.com` is not preloaded, this site is still vulnerable on first load to the [SSLStrip attack](https://moxie.org/software/sslstrip/).  In order to qualify for removal, this site must be included in the HSTS preload list for all the browsers that we support.  See https://github.com/EFForg/https-everywhere/issues/7126 for additional context

PR #4091, opened 2016-02-04
Excluded http://elections.stackexchange.com

stevenatwork:
As it does not have https


stevenatwork:
Which is better: `<exclusion pattern="^http://elections\.stackexchange\.com" />` or `<exclusion pattern="^http://elections\.stackexchange\.com/" />`

fuglede:
Thanks for the PR. Regarding your question: The first one would match <code>http://elections.stackexchange.com.otherfunkystuff.stackexchange.com</code> whereas the second would not, so the second one is more to the point. Kind of a far-fetched example though.

PR #4092, opened 2016-02-04
cs-ellpic.yandex.net supports https

stevenatwork:
https://github.com/EFForg/https-everywhere/issues/1327


J0WI:
You should also remove the rule from the downgrade whitelist: https://github.com/EFForg/https-everywhere/blob/master/utils/downgrade-whitelist.txt

stevenatwork:
fixed

J0WI:
Thanks!

PR #4095, opened 2016-02-05
Fix: custom rules should be disableable

Hainish:
JSON.stringify was transforming custom rule ids:

From: custom_0
To: "custom_0"

(quotes included in the string)

This was breaking the ability to disable custom rules.  This fixes it by making sure rules fit a certain regex for validation, but removing the JSON.stringify.

### DIFF
--- a/src/chrome/content/code/ApplicableList.js
+++ b/src/chrome/content/code/ApplicableList.js
@@ -213,9 +213,13 @@ ApplicableList.prototype = {
   },
 
   add_command: function(rule) {
+      // basic validation for data to be added to xul
+      if(!String(rule.id).match(/^[a-zA-Z_0-9]+$/))
+        return;
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/code/ApplicableList.js
```diff
@@ -213,9 +213,13 @@ ApplicableList.prototype = {
   },
 
   add_command: function(rule) {
+      // basic validation for data to be added to xul
+      if(!String(rule.id).match(/^[a-zA-Z_0-9]+$/))
```
jsha: Local style is a space between `if` and `(`, and braces around all blocks.

Also, this would probably be better as:

```
var ruleId = String(rule.id);
if (!ruleId.match(/.../)) {
  return;
}
setAttribute(... ruleId + 'blah')
```

That is, reuse the same variable so it's clearer that it's the one you checked.

Hainish:
Ready for review @jsha

PR #4099, opened 2016-02-06
Bitly vanity domains: direct http to https redirect (resolves #4097)

### LINKED ISSUE #4097: Re-enable Bit.ly vanity domains
src/chrome/content/rules/Bit.ly_vanity_domains.xml has a lot of rules for Bit.ly vanity domains that we had to disable. They redirected vanity domains to bitly.com, which would fail horribly when the vanity domain moved off of bitly.

Now, it looks like many bitly domains have valid HTTPS certificates, so in theory this whole ruleset can be turned into a simple http->https rule.

To do this, (1) comment out all existing rules, and add an http->https rewrite. (2) run .`/fetch-test.sh src/chrome/content/rules/Bit.ly_vanity_domains.xml` and check which domains fail. Remove those domains from the target list.

We can probably also remove most of the special rewrites. Most of these are there because the root URL of a bitly vanity domain would general redirect to the owner, e.g. nyt.ms -> nytimes.com. With the simple http->https this is no longer necessary. Still, we should read each particular special rewrite and understand its purpose before removing it.

Hainish:
I went through the instructions in #4097 so the target domains all have valid https endpoints. It looks like all the rules are snapping redirects, other than the few very particular mzl.la path redirects. Some of those were simply mimicking server redirects that no longer exist, others were very specific paths that are still redirecting. Either way, https://mzl.la and all paths behind it redirect to https endpoints it looks like, so this should no longer be needed.

I also removed the exclusion patterns, as they seem brittle to me. Why should we only redirect bit.ly patterns that we've observed as valid instead of them all? What if they change this pattern? Better to just redirect them all.

I'm not completely sure if I've done everything right here and would appreciate any feedback you have.

### DIFF
--- a/src/chrome/content/rules/Bit.ly_vanity_domains.xml
+++ b/src/chrome/content/rules/Bit.ly_vanity_domains.xml
@@ -40,279 +40,60 @@
   -->
 <ruleset name="bit.ly vanity domains" default_off="Breaks many redirectors">
 
-	<target host="1tw.org" />
-	<target host="www.1tw.org" />
-	<target host="4mn.ca" />
 	<target host="7ny.tv" />
-	<target host="www.7ny.tv" />
-	<target host="a21.tv" />
 	<target host="abcn.ws" />
-	<target host="www.abcn.ws" />
-	<target host="alj.am" />
-	<target host="www.alj.am" />
 	<target host="apne.ws" />
 	<target host="atfp.co" />
-	<target host="www.atfp.co" />
 	<target host="bbc.in" />
 	<target host="bitly.is" />
 	<target host="bloom.bg" />
 	<target host="brook.gs" />
-	<target host="www.brook.gs" />
 	<target host="buff.ly" />
-	<target host="www.buff.ly" />
-	<target host="bzfd.it" />
-	<target host="www.bzfd.it" />
 	<target host="cbsn.ws" />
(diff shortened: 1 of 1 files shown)


jsha:
Agreed that now, with a simple rewrite, we no longer need the "Avoid matches that are legibly cased" exception.

I was surprised by how many of these domains had to be removed. I spot-checked a few, and found that they provide a bit.ly cert. My current guess is that perhaps bit.ly hasn't yet HTTPS-enabled *all* their vanity domains. Could you do one more pass to find the domains that don't have a valid cert, but the invalid one they present is for bit.ly? That list should go in a comment, listing "possible future hostnames for this ruleset," along with an explanation.

jsha:
This might also be interesting / useful: http://dev.bitly.com/domains.html.

Hainish:
The domains with TLS errors I stuck in a file named `errors` and ran this over it to filter by those which have a cert valid for `bit.ly`:

```
for x in `cat errors`; do
  if [ `echo 'x' | openssl s_client -connect $x:443 | openssl x509 -noout -text | grep -e "^        Subject:.* CN=bit.ly$" | wc -l` -ne 0 ]; then
    echo $x >> errors_bitly
  fi
done
```

Hainish:
Okay, I added that note. Ready for review @jsha

PR #4100, opened 2016-02-06
Add new icons and banner.

jsha:
The old icons were hard to make out at their typical size. The new icons are
clearer, and avoid the lock iconography in order to avoid creating confusion
with the lock icon in the URL bar.

Adds new icon type:
  "inactive" indicates that no rules applied on a given site.
  "disabled" indicates HTTPS Everywhere is disabled.

Disables the counter by default on Firefox (you can still turn
it on manually), and adds a new icon state "inactive."

cc @semenko, @cooperq, and @Hainish for review and test.


Hainish:
This looks great! I can do a more comprehensive code review on Monday.

chromium/manifest.json
```diff
@@ -9,35 +9,35 @@
...
-        "128": "icon128.png", 
-        "16": "icon16.png", 
-        "48": "icon48.png"
-    }, 
-    "incognito": "spanning",
-    "manifest_version": 2, 
+        "48": "icons/icon-active-48.png",
+        "128": "icons/icon-active-128.png"
+    },
+    "incognito": "split",
```
semenko: Any reason you went back to split incognito? (I'm guessing this was a merge issue)

jsha: Yep, merge issue, thanks for catching!

chromium/background.js
```diff
@@ -88,26 +92,35 @@ loadStoredUserRules();
...
+ * blocking: extension is in "block all HTTP requests" mode.
+ * active: extension is enabled and rewrote URLs on this page.
+ * disabled: extension is disabled from the popup menu.
  */
-var setIconColor = function() {
-  var newIconPath = httpNowhereOn ? './icon38-red.png' : './icon38.png';
-  chrome.browserAction.setIcon({
-    path: newIconPath
+var updateState = function() {
+  chrome.tabs.query({active: true}, function(tabs) {
```
semenko: You'll want to pass currentWindow:true here, too.

Otherwise, creation order determines the `tabs` array, and with multiple windows, multiple tabs can be active.

Example:
- Open a tab to a weird page (e.g. chrome://extensions)
- The extension icon will show as inactive
- Open a new window, and browse -- the icon state won't change

jsha: Fixed, thanks.

This has the unfortunate side effect that if you have two windows visible at once, the icon in both windows will reflect the state of whichever window is currently focused, and switching windows will change icon state (if it's different). We could probably fix this by switching back from browserAction to pageAction, but I'm not very excited about that. Probably the current behavior is fine.

semenko:
Awesome -- Chrome bits LGTM (& tested on Stable) and they're supported in WebExtensions for an eventual FF code merger (e.g. https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/Tabs/query ).

I agree the current behavior is fine -- going back to pageAction would be pretty annoying.

Hainish:
When you move HTTPS Everywhere to behind the hamburger menu, the icon becomes much blurrier:

screenshot - 02162016 - 03 49 30 pm

Can we make it clearer, like in the `about:addons` page?

Hainish:
Within the `SSL Observatory` preferences, the older banner image has been replaced with a new HTTPSE image.  I discussed this with @jsha, who lmk that the old HTTPSE iconography is embedded into the observatory image, looks like on the right side, in red.  The original image can be found here, without this iconography: https://www.eff.org/files/sslobservatory.jpg

IMHO it seems more desirable to have a banner that's more specific to the observatory, just so people have a clearer indication of what they're looking at.

It also looks like the image width/height ratio is lessened from the original to give it a squished look, probably worth it to fix that.

Hainish:
everything else lgtm

PR #4104, opened 2016-02-06
Update I3wm.org.xml

J0WI:
Accidentally pushed to master in c522683f384e03822148d41790a8808ec16319f5 and reverted in a0421d77159b6f66f8ea533091f5fe0cc6d6a677

Cherry-picked to request regular review here.

### DIFF
--- a/src/chrome/content/rules/I3wm.org.xml
+++ b/src/chrome/content/rules/I3wm.org.xml
@@ -1,21 +1,18 @@
 <!--
 	Nonfunctional subdomains:
 
-		- (www.) *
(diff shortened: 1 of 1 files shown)


funkydude:
When I try to access https://www.i3wm.org (without a rule) the site redirects to http://i3wm.org. You may want to redirect www in your rule.

Other than that, LGTM.

J0WI:
...and ^i3wm.org will be redirected to www.
We should not force redirects on our site if it doesn't break anything.

funkydude:
Fair enough.

PR #4109, opened 2016-02-07
Check scheme before attempting to rewrite.

jsha:
This avoids "uri.host is explosive" errors in the logs on URIs like
about:startpage. This check already happens in https-everywhere.js before
calling replaceChannel, but this other path into rewrittenURI wasn't checked
previously.

### DIFF
--- a/src/chrome/content/code/ApplicableList.js
+++ b/src/chrome/content/code/ApplicableList.js
@@ -73,7 +73,11 @@ ApplicableList.prototype = {
   populate_list: function() {
     // The base URI of the dom tends to be loaded from some /other/
     // ApplicableList, so pretend we're loading it from here.
-    HTTPSEverywhere.instance.https_rules.rewrittenURI(this, this.uri);
+    var uri = this.uri;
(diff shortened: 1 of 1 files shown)


src/chrome/content/code/ApplicableList.js
```diff
@@ -73,7 +73,11 @@ ApplicableList.prototype = {
   populate_list: function() {
     // The base URI of the dom tends to be loaded from some /other/
     // ApplicableList, so pretend we're loading it from here.
-    HTTPSEverywhere.instance.https_rules.rewrittenURI(this, this.uri);
+    var uri = this.uri;
+    if (!(uri.schemeIs("http") || uri.schemeIs("https"))) {
+      return true;
```
cooperq: Is `true` what we want to be returning here?

cooperq:
lgtm other than one question above.

src/chrome/content/code/ApplicableList.js
jsha: Good point. This method normally doesn't return anything, so I removed the return value for clarity.

PR #4110, opened 2016-02-08
Switch to using JSON to store rulesets.

jsha:
This loads much faster, and avoids reads from disk after startup.

Previously, loading all targets from SQLite took about 268ms. Now it takes about
60ms, including reading the file from disk and parsing JSON.

This replaces #3852 by removing SQLite support altogether.

### DIFF
--- a/src/chrome/content/code/HTTPSRules.js
+++ b/src/chrome/content/code/HTTPSRules.js
@@ -402,11 +402,16 @@ const HTTPSRules = {
     return;
   },
 
+  /**
+   * Read and parse the ruleset JSON.
+   * Note: This only parses the outer JSON wrapper. Each ruleset is itself an
+   * XML string, which will be parsed on an as-needed basis.
+   */
   loadTargets: function() {
     var file = new FileUtils.File(RuleWriter.chromeToPath("chrome://https-everywhere/content/rulesets.json"));
     var rules = JSON.parse(RuleWriter.read(file));
     this.targets = rules.targets;
-    this.rules_list = rules.rules_list;
+    this.rulesetStrings = rules.rulesetStrings;
   },
 
   checkMixedContentHandling: function() {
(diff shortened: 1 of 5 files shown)


src/chrome/content/code/HTTPSRules.js
```diff
@@ -424,6 +402,13 @@ const HTTPSRules = {
     return;
   },
 
+  loadTargets: function() {
```
cooperq: Should this function have documentation?

utils/make-sqlite.py
```diff
@@ -89,7 +97,14 @@
     ruleset_id = c.lastrowid
     for target in targets:
         c.execute('''INSERT INTO targets (host, ruleset_id) VALUES(?, ?)''', (target, ruleset_id))
```
cooperq: This line should probably be removed

utils/make-sqlite.py
```diff
@@ -89,7 +97,14 @@
     ruleset_id = c.lastrowid
```
cooperq: Is there a reason to keep these lines of code around? If not they should be deleted. Also the sqlite cursor and the `c` variable should probably be removed.

cooperq:
lgtm other than my above comments. @jsha do you feel that this is thoroughly tested?

utils/make-sqlite.py
jsha: I kept the sqlite generation around because the validator still uses it. Eventually I'd like to break that dependency, but it will be a bigger, harder change, and I wanted to get this out for the next release so Mozilla will start accepting our releases again.

src/chrome/content/code/HTTPSRules.js
jsha: Good point, fixed.

jsha:
It's moderately tested. Would definitely appreciate if you would kick the tires a little bit too!

cooperq:
seems good to me!

PR #4113, opened 2016-02-09
Create Douban.xml

renyouguo:


fuglede:
The first target host is already covered in <code>Douban.com.xml</code>; perhaps you could merge yours with that one?

PR #4120, opened 2016-02-10
New ruleset for static-course-assets.s3.amazonaws.com

GigabyteProductions:


J0WI:
Hmm, there was a cause why we don't rewrite *.s3.awazonaws.com. May be @fuglede knows more?

fuglede:
I never looked at this one in particular, but <code>git log -p</code> reveals quite a few problematic hosts. http://static.via.me.s3.amazonaws.com/ is one example.

GigabyteProductions:
This one didn't seem problematic. I'm able to successfully browse Cisco's NetAcad resources that are located on this with this rule. Although, another problem I see for you guys is that you can't really test actually using the site anonymously, so I'm guessing this isn't going to get merged anyway.

fuglede:
We can probably live with that; but you're saying that it's not possible to cook up rewritable URLs that do not require authentication?

In any case, the URL should be added in  <code>rules/AmazonAWS.xml</code>.

GigabyteProductions:
It's not that the rewrite triggers a re-authentication; this website requires authentication from https://www.netacad.com, no matter what protocol you're connecting with. My point is the only people who will be able to test-browse the resources on this website are Cisco NetAcad students, which I thought might be a problem for getting this merged.

fuglede:
Yes, it's not ideal of course. It wouldn't be the first one of its kind though, and the "good" thing is that it also means that any breakage will affect correspondingly fewer people

PR #4122, opened 2016-02-10
[MPlayer] Update ruleset

jetomit:
Add new subdomains (bugzilla, ffmpeg, samples, trac) and remove comment.


J0WI:
Rule lgtm.

Please double check the www\d.mplayerhq.hu domains again and add some comments for mirrors that are broken with https.

jetomit:
Thanks, I added a comment for HTTP-only mirrors.

PR #4126, opened 2016-02-12
added ruleset for computerwerk.org

moonlies:
Wildcard ruleset with two testing urls.


J0WI:
It seems that wiki.computerwerk.org moved to computerwerk.org/wiki
You should also add a redirect for ^ to www.

Is there a reason to keep the wildcard rule when there is only one working domain?

moonlies:
Since we are not planing to publish websites without a valid certificate, we would not have to update the rulesets.

J0WI:
But this would still break the wiki subdomain and doesn't fix the insecure redirect from ^ to www.

J0WI:
@moonlies are you still working on this?
Please also fix the code style.

moonlies:
I will rewrite them and make another pull request someday.

PR #4129, opened 2016-02-13
Update Internet-Archive.xml

redranamber:


J0WI:
host times out from here, on http and https.

redranamber:
I didn't know how to give a good test domain. But if you borrow a book from openlibrary.org it downloads from here. However it won't download with the ruleset forcing https.

PR #4130, opened 2016-02-13
Create 101weiqi.com.xml

renyouguo:


J0WI:
Could you add a comment, that the cert doesn't match demo.101weiqi.com?
rule lgtm

PR #4131, opened 2016-02-13
new ruleset for letterboxd

phibo23:
https officially supported since a few days ago according to their feedback thread http://feedback.letterboxd.com/forums/136771-general/suggestions/2505918-login-registration-over-https

### DIFF
--- a/src/chrome/content/rules/Letterboxd.com.xml
+++ b/src/chrome/content/rules/Letterboxd.com.xml
@@ -2,12 +2,22 @@
 	<target host="letterboxd.com" />
 	<target host="www.letterboxd.com" />
 
-<!-- the following subdomains don't support https yet but are listed here for reference -->
+<!-- 
+	the following subdomains don't support https yet but are listed here for reference 
+-->
+
 <!-- 	<target host="news.letterboxd.com" /> -->
 <!-- 	<target host="embed.letterboxd.com" /> -->
 <!-- 	<target host="feedback.letterboxd.com" /> -->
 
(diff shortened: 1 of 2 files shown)


phibo23:
that file src/chrome/content/rulesets.json was created when running
```
test/firefox.sh --justrun
```
so I figured it should be added to gitignore

J0WI:
You should also check `*.ltrbxd.com`.
The are at least `a` and `s2`

J0WI:
Sorry, I closed accidentally.

phibo23:
ok, they have a wildcard certificate for *.ltrbxd.com (just as they do for letterboxd.com) but not all servers of their cdn (?) appear to be configured for https. I'll look into it and add anything I find to be working.

phibo23:
as far as I understand, all of the servers of their cdn support https but they don't have certificates for multi level subdomains yet. If you visit letterboxd.com without https, static files are served from a variety of cdn servers. But if you visit with https, only the three single level subdomains (a, s1 and s2) are used, so I guess it's their own decision to "bypass" the cdn as a sideeffect of using https.

src/chrome/content/rules/Letterboxd.com.xml
```diff
@@ -7,6 +7,15 @@
 <!-- 	<target host="embed.letterboxd.com" /> -->
 <!-- 	<target host="feedback.letterboxd.com" /> -->
 
+<!-- there is also ltrbxd.com which appears to be mainly used for a cdn. However not all subdomains work correctly with https. For example, every file on a.ltrbxd.com is also available on skyfall.a.ltrbxd.com, oblivion.a.ltrbxd.com etc. These servers can be accessed with https but browsers will display errors/warnings because there is only a wildcard certificate for *.ltrbxd.com and not one for each of the subdomains. -->
```
J0WI: Please use a linebreak here

J0WI:
Thanks!

PR #4140, opened 2016-02-13
add rule for Whatsbroadcast.com

numismatika:


J0WI:
I found some aliases:
api2
my2
www2
img.wbcsrv.com (mismatch/401 on www and ^)

PR #4143, opened 2016-02-13
add rule for TZ.de

numismatika:


J0WI:
There is some mixed content on articles, that breaks videos.
Mismatch on markt.tz.de

numismatika:
@J0WI do you have an example for such a broken video in an article?
The only ones i found so far come from snacktv and work.

J0WI:
I got mixed content from "http://www.video.oms.eu/ada/cloud/omsv_container_151.js", but it seems that most videos are included from YouTube or snacktv.

Could you add comments for mismatching subdomains?

numismatika:
i created  #4171 for this advertising company as well

PR #4149, opened 2016-02-15
Add HTW-Berlin.de ruleset

kollegah1338:


src/chrome/content/rules/HTW-Berlin.de.xml
```diff
@@ -0,0 +1,133 @@
...
+	<target host="inka.f4.htw-berlin.de" />
+	<target host="kd.htw-berlin.de" />
+	<target host="lsf.htw-berlin.de" />
+	<target host="moodle.htw-berlin.de" />
+	<target host="my.htw-berlin.de" />
+	<target host="oce-docworks.rz.htw-berlin.de" />
+	<target host="ox.htw-berlin.de" />
+	<target host="portal.rz.htw-berlin.de" />
+	<target host="secure.htw-berlin.de" />
+	<target host="sisis.rz.f4.htw-berlin.de" />
```
J0WI: host does not exist anymore

src/chrome/content/rules/HTW-Berlin.de.xml
```diff
@@ -0,0 +1,133 @@
...
+		- wiw-fernstudium.htw-berlin.de
+		- wiw-master.htw-berlin.de
+		- wm.htw-berlin.de
+		- wr-bachelor.htw-berlin.de
+		- wr-master.htw-berlin.de
+
+	Problematic subdomains (not working):
+		- bewitec.inka.f4.htw-berlin.de
+		- code.repo.inka.f4.htw-berlin.de
+		- demo.inka.f4.htw-berlin.de
```
J0WI: WFM

J0WI:
Incredible work!
This site has a huge structure, so it's quite impossible to test everything manually.
You could do something like default_off testing if you think this is a risky change, but personally I think it's quite stable.

kollegah1338:
@J0WI 
Fixed these 2 subdomains

This is quite huge, but this is a problem for all rules for university pages I guess. But as most subdomains just don't work, the ones that are enabled are the big main pages. These shouldn't really break anymore and most of them redirect to HTTPS by default anyway, so it should be safe to enable it.

I probably also missed a ton of subdomains that exist, but I hope that by having this as a base for further improvements more students might contribute later on.

J0WI:
Yes, rules for university pages tend to be huge and complex, but it's awesome when someone can invest the time in them.

PR #4152, opened 2016-02-15
Add rule for FreeCallsToUSA.co.uk

davecardwell:
Add rule for https://www.freecallstousa.co.uk/

PR #4153, opened 2016-02-15
Add rule for FreeCallsCanada.co.uk

davecardwell:
Add rule for https://www.freecallscanada.co.uk/

### DIFF
--- a/src/chrome/content/rules/FreeCallsCanada.co.uk.xml
+++ b/src/chrome/content/rules/FreeCallsCanada.co.uk.xml
@@ -1,0 +1,9 @@
+<ruleset name="FreeCallsCanada.co.uk">
+	<target host="freecallscanada.co.uk" />
+	<target host="www.freecallscanada.co.uk" />
+
+	<rule from="^http://freecallscanada\.co\.uk/" to="https://www.freecallscanada.co.uk/" />
+	<rule from="^http:" to="https:" />
+
+	<securecookie host="^(www\.)?freecallscanada\.co\.uk$" name=".+" />
+</ruleset>


src/chrome/content/rules/FreeCallsCanada.co.uk.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="FreeCallsCanada.co.uk">
+	<target host="freecallscanada.co.uk" />
+	<target host="www.freecallscanada.co.uk" />
+
+	<test url="http://freecallscanada.co.uk/" />
+	<test url="https://freecallscanada.co.uk/" />
+	<test url="http://www.freecallscanada.co.uk/" />
+
+	<rule from="^https?://freecallscanada\.co\.uk/" to="https://www.freecallscanada.co.uk/" />
```
J0WI: You should not touch https domains and forcing to www as long as this works server side.

davecardwell: @J0WI Thank you for checking. I have updated the regular expression and removed the redundant tests.

J0WI:
Still no need to force the redirect to www, because this already works server-side.
Same on #4152.

Btw: If those hosts are related, you could merge the files or add a comment for each other.

davecardwell:
@J0WI There are likely to be several more of these sites, so would the preferred method be to have a single “FreeCallsToX.xml” file that contained the rules for all of them?

J0WI:
If one company owns all those services, use the Company name.
If there is really a bunch of related hosts and the rule becomes much more complex, group the hosts in whatever makes sense and give a hint in comments to the related files.

davecardwell:
Rules moved to the existing localphone.com.xml in #4199

PR #4157, opened 2016-02-16
Vagrant additions

xaocon:
Added common vagrant files to the ignore list and added git to the of packages to install for debian based systems in case the dev system does not already have it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ tokenkeys.py*
 from-preloads/
 test_profile/
 .venv/
+Vagrantfile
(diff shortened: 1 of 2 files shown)


Hainish:
Looks good to me.  In the future we may want to standardize the `Vagrantfile` so we have a common dev setup, in which case we'll have to remove it from `.gitignore`

xaocon:
I considered setting up a Vagrant file for a moment and would be willing to do so but I thought something like that would probably need to be discussed by team members.

PR #4168, opened 2016-02-18
don't fail script when no translations files

aoaaxy:


jsha:
This shouldn't be necessary in general. Have you run `git submodule update --init` ?

aoaaxy:
Oops, I didn't. That explains everything. Thanks! :)

PR #4170, opened 2016-02-18
add rule for WiDSL

numismatika:


J0WI:
This breaks login page because of mixed content.

There are also some non-working hosts:
m.widsl.de
kunden.widsl.de
kunden.m.widsl.de
www\d+.widsl.de

numismatika:
indeed because the login formular includes http://kunden.widsl.de/index.php which can't be secured.
Looks like a mail is due to their customer service.
TLS config is also rather bad

J0WI:
You could work around this using `platform="mixedcontent"`.
Of course it would be even better if they fix it...

jeremyn:
It looks like the problems what were blocking this pull request have been fixed. @numismatika Are you still interested in working on this?

numismatika:
Sure, just need to make sure which sub domains are currently in use there-

jeremyn:
Please do this, and then I'll merge it:
- [ ] Add `target` for https://hotspot.widsl.de
- [ ] Remove [this special rewrite `rule`](https://github.com/numismatika/https-everywhere/blob/2d0ff116f9b8c5f75dc1463d916365e4a1e2fd86/src/chrome/content/rules/WiDSL.xml#L6)

jeremyn:
Thanks, merged.

PR #4172, opened 2016-02-18
Kalooga.com: add publishing.kaloo.ga

numismatika:
This adds the geo,img and publishing subdomain for kaloo.ga

### DIFF
--- a/src/chrome/content/rules/Kalooga.com.xml
+++ b/src/chrome/content/rules/Kalooga.com.xml
@@ -5,9 +5,16 @@
 <ruleset name="Kalooga.com (partial)">
 
 	<target host="publishing.kalooga.com" />
+	<target host="geo.kaloo.ga" />
+	<target host="img.kaloo.ga" />
+	<target host="publishing.kaloo.ga" />
 
+	<test url="http://publishing.kaloo.ga/wid/16513.js?loc=http%3A%2F%2Fwww.bz-berlin.de%2Fberlin%2Fber-zweite-anzeige-gegen-jochen-grossmann" />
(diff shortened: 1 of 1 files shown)


fuglede:
Would you be able to provide some non-trivial test URLs? Otherwise it looks like it will be a bit tricky to verify the function of this one.

numismatika:
<test url="http://publishing.kaloo.ga/wid/16513.js" />
	<test url="http://publishing.kaloo.ga/acct/4455.js" />
	<test url="http://geo.kaloo.ga/json/" />
	<test url="http://img.kaloo.ga/thumb?url=http%3A%2F%2Fwww.shz.de%2Fimg%2Fwirtschaft%2Fcrop11177506%2F8806138589-cv16_9-h335%2Furn-newsml-dpa-com-20090101-151111-99-05409-large-4-3.jpg&md5val=911f772734b61479440330f1140d13b4&key=00f8a403a132c924d4f37b5df7818f1454f668f8&size=192x140" />

the problem is the img subdomain which creates preview pictures and needs a parameter.

fuglede:
I see. Also, the first one returns a <code>HTTP/1.1 204 No Content</code> here. Otherwise LGTM; if you could put those in to the extent possible, that would be good (helps out on maintenance).

numismatika:
The first one also seems to require an url as parameter, the problem is  how do i encode those properly to make the test scripts happy?

fuglede:
I see. I believe that you should be able to just use any URL encoder such as [this one](http://meyerweb.com/eric/tools/dencoder/) to get rid of any special characters causing issues.

numismatika:
@fuglede  looks like Travis is happy now.

fuglede:
Great, thanks!

PR #4176, opened 2016-02-19
added bleachbit.org (fixed #4128)

### LINKED ISSUE #4128: bleachbit.org doesn't redirect to HTTPS
http://www.bleachbit.org doesn't redirect to https://www.bleachbit.org/

Not much else to say. :laughing:

J0WI:


fuglede:
The root domain is problematic (which I believe you already noticed?):

    $ curl https://bleachbit.org
    curl: (35) Unknown SSL protocol error in connection to bleachbit.org:443

J0WI:
Actually I planned to do that, yes.

PR #4179, opened 2016-02-19
Updated Web_Foundation.org.xml

J0WI:


fuglede:
The front page of https://labs.webfoundation.org/ looks rather broken here.

J0WI:
They load some images over JS :(

PR #4184, opened 2016-02-19
[Maps.me] New ruleset, [My.com] Rewritten

rugk:


J0WI:
Got 403 on fb.maps.me
The embedded Video on the main site is blocked.

It seems that every product/app/whatever have a subdomain on my.com. So you may consider to keep the wildcard and just adding test urls.

rugk:
Oh okay. How do you always get all this subdomains?

rugk:
> Got 403 on fb.maps.me

Yeah, but this seems to be okay. I think this is intended. (`fb` = `facebook`)

rugk:
> The embedded Video on the main site is blocked.

Which video?

J0WI:
> Yeah, but this seems to be okay. I think this is intended. (fb = facebook)

Yes, but it's an insecure redirect.

> Oh okay. How do you always get all this subdomains?

See https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

> Which video?

On the bottom of maps.me

rugk:
> Yes, but it's an insecure redirect.

Okay, but the Facebook ruleset should catch it. And I also think Facebook is even preloaded.

> See https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

Ah thanks. Need to look how to do that with DuckDuckGo.

> On the bottom of maps.me

Ah it is not there. That's why I don't see it. :laughing: 

/cc @mapsme @deathbaba
Could you fix this HTTP URL of the YouTube video, whcih is embedded on Maps.me? YouTube supports and uses HTTPS by default.

J0WI:
> Okay, but the Facebook ruleset should catch it.

Yes, but with plain http you could still hijack the redirect, that it points to something evil.

> Thanks. Need to look how to do that with DuckDuckGo.

Sadly DDG doesn't support wildcards on site parameters. I think I reported it somewhere long time ago.

rugk:
> Yes, but with plain http you could still hijack the redirect, that it points to something evil.

That's why it is in the rule... wait. It is not there. Why did not you said this? :smile: 
I puzzling what your comment means here and then it is simple another missing subdomain. I though something with the redirect would be wrong...

J0WI:
Sorry, your right. I lost the content of previous comments a bit. There is something wrong:
> Got 403 on fb.maps.me

So this should be commented or redirected directly to the target on Facebook.

rugk:
> So this should be commented or redirected directly to the target on Facebook.

Why? Especially why deactivate ruleset?
And if the Facebook account URL changes this is not good either.

But of course I can make the change.

deathbaba:
http redirect of fb.maps.me <http://fb.maps.me/> works correctly.

rugk:
Okay, so no slash?

rugk:
So...

deathbaba:
?

J0WI:
The slash in the from pattern is the important one, but LGTM anyway.

Issue #4072, opened 2016-02-02
HTTPS Everywhere breaks Lynda.com videos


tillulen:
HTTPS Everywhere version 2015.12.16 for Google Chrome (stable channel under Windows 8.1) breaks the playback of Lynda.com free videos.

Steps to reproduce:

1. Enable HTTPS Everywhere.
2. Open any freely accessible video on Lynda.com, for example [this one](http://www.lynda.com/After-Effects-tutorials/Welcome/372438/383542-4.html).
3. Click the Play button. (The big Play button titled "Watch this week's free episode" will actually play the latest video in the series. To play the linked video, click the tiny Play button to the left of the Welcome video, which is the first in the list on the left hand side. The problem should be reproducible with any of those videos.)

Expected result: the video would be played back. Actual result: there is a black rectangle instead of the video and no sound.

The video works fine if you disable HTTPS Everywhere.

PR #4193, opened 2016-02-20
Disable Lynda.com

funkydude:
Closes https://github.com/EFForg/https-everywhere/issues/4072

There's so many exclusions it's pretty much pointless to have this active.


J0WI:
You should comment why this is disabled.

funkydude:
@J0WI done

PR #4198, opened 2016-02-21
[GoogleAPIs] Add ct.googleapis.com

fuglede:
The 404s in the folder roots look a bit different over HTTP and HTTPS, but according to https://tools.ietf.org/html/rfc6962#section-4, all CT logs are to be served over HTTPS, so as far as the actual logs are concerned, this should be alright.


J0WI:
Is there any valid test url? The current one seems to be broken.

fuglede:
Hm, how so? It works here;

    $ diff <(curl -s "http://ct.googleapis.com/aviator/ct/v1/get-entries?start=0&end=1") <(curl -s "https://ct.googleapis.com/aviator/ct/v1/get-entries?start=0&end=1")

returns 0.

J0WI:
Hmm, now it works.

Issue #4196, opened 2016-02-21
Arstechnica rule modification


terrorist96:
Currently, the rule for Arstechnica.com will force https if the url ends with arstechnica.com/civis/ucp.php, but it won't if it ends with arstechnica.com/civis/viewtopic.php.
It should be forced for both instances. What modifications to the current xml file need to be done to achieve this?

terrorist96:
Would be nice if this could be added for the next version.

terrorist96:
https://arstechnica.com/information-technology/2017/01/ars-announces-https-by-default-finally/

PR #4201, opened 2016-02-22
Update Ars-Technica.xml

terrorist96:
Addresses #4196

### DIFF
--- a/src/chrome/content/rules/Ars-Technica.xml
+++ b/src/chrome/content/rules/Ars-Technica.xml
@@ -70,7 +70,7 @@
 	<target host="ars.to" />
 	<target host="arstechnica.com" />
 	<target host="*.arstechnica.com" />
-		<exclusion pattern="^http://(?:www\.)?arstechnica\.com/(?!ads/|civis/ucp\.php|favicon\.ico|public-shared/|services/incr\.php|subscriptions(?:$|[?/])|wp-content/)" />
+		<exclusion pattern="^http://(?:www\.)?arstechnica\.com/(?!ads/|civis/ucp\.php|civis/viewtopic\.php|favicon\.ico|public-shared/|services/incr\.php|subscriptions(?:$|[?/])|wp-content/)" />
 	<target host="*.arstechnica.net" />
 
 


terrorist96:
@J0WI what does `needs test coverage update` mean?

J0WI:
This will fail ruleset tests, see https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md

terrorist96:
Please tell me what to add/remove to fix it. Thanks.

J0WI:
You need to add test urls for every possible expression that matches the rule and at least three examples for wildcards.

terrorist96:
Even for all the existing rules?  How can I find test URLs? Something like this: `https://arstechnica.com/civis/viewtopic.php?f=2&t=1308189&p=30695775#p30695775`? 
And I don't know what you mean by wildcards.

J0WI:
Exactly. Note that the test urls should be as generic as possible and should not contain any errors (e.g. 404).

There is a wildcard target host on https://github.com/EFForg/https-everywhere/pull/4201/files#diff-aa9c3be14dc8df076fc11fd18d4adeabR72
You could also replace this with a list of subdomains as target hosts.

terrorist96:
Here's what I'm trying to accomplish: if I comment on an article from arstechnica.com and someone comments after me, I'll get an email with a link like this
http://arstechnica.com/civis/viewtopic.php?f=2&t=1310747&p=30849471&e=30849471
If I'm not already logged in, this will take me to a page to log in. Https isn't forced for this page and it should be. The default login page (accessed by clicking the login button from the site) ends in civis/ucp.php, which is already covered by the rule, but the link to login that redirects to a specific thread (as obtained from an email notification) ends in civis/viewtopic.php, which doesn't have https forced. So I can't give a generic version of the link; it has to point to a specific thread, like the example I just pasted for the article about the Tesla Model X. If you strip the stuff after viewtopic.php, it will load a 404 page.

terrorist96:
Actually, the generic http://arstechnica.com/civis/ needs https forced too.

terrorist96:
@fuglede I need help getting this merged.

terrorist96:
Not fixed in (Version: 2016.5.10)

J0WI:
Please read our guides carefully:

This ruleset does not comply with our current standards, so it needs to be rewritten.
This means:
- get rid of unneeded wildcards
- remove non-capturing groups
- test coverage
- update documentation

jeremyn:
@J0WI I get what you're saying but this is literally a one-line change, and `Ars-Technica.xml` is [already in the whitelist](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist.txt#L438). I think this can be merged after the whitelist hash is updated, although @terrorist96 has deleted their branch so we might need a new pull request. What do you think?

Hainish:
source repo deleted

PR #4205, opened 2016-02-22
[Motley-fool] breaks images, disable

pde:
Eg: http://www.fool.com/investing/general/2013/10/12/facebooks-incredible-growth-story-in-6-charts.aspx


cooperq:
lgtm

PR #4210, opened 2016-02-23
Fixes for Firefox Developer Edition (resolves #4195)

### LINKED ISSUE #4195: HTTPS Everywhere is broken on Firefox Developer Edition
I am using 64-bit Firefox Developer Edition 46.0a2 (2016-02-20).

I installed HTTPS Everywhere from AMO, not EFF.org.

I have version 5.1.3.

The SSL Observatory option doesn't work in the toolbar icon, and neither does the about option. On top of that, all of the settings in the observatory (accessed through the about:addons menu) are broken and don't work.

Is this related to changes in how Firefox handles addons, or is this a bug?

If anybody can help me confirm I am not the only one having these issues, that would be great.

Feedback appreciated.

Hainish:
1. Refactor js one-liner to allow ssl observatory component to load
2. HTTPSEverywhere object can no longer be refernced directly from
   toolbar xul

### DIFF
--- a/src/components/ssl-observatory.js
+++ b/src/components/ssl-observatory.js
@@ -313,7 +313,12 @@ SSLObservatory.prototype = {
     function toHexString(charCode) {
       return ("0" + charCode.toString(16)).slice(-2);
     }
-    return [toHexString(h.charCodeAt(i)) for (i in h)].join("").toUpperCase();
+
+    var hexArr = [];
+    for (i in h){
+      hexArr.push(toHexString(h.charCodeAt(i)));
+    }
+    return hexArr.join("").toUpperCase();
   },
 
   ourFingerprint: function(cert) {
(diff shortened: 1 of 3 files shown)


src/components/ssl-observatory.js
```diff
@@ -313,7 +313,12 @@ SSLObservatory.prototype = {
     function toHexString(charCode) {
       return ("0" + charCode.toString(16)).slice(-2);
     }
-    return [toHexString(h.charCodeAt(i)) for (i in h)].join("").toUpperCase();
+
+    var hexArr = [];
+    for (i in h){
+      hexArr.push(toHexString(h.charCodeAt(i)));
```
jsha: Have you tested this loop? Suggestion: run before and after the patch, look at submitted hex values from the logs for a known host (e.g. libssh.org), and make sure they are the same.

Hainish: Just checked - they are the same.

cooperq:
LGTM

jsha:
LGTM

PR #4221, opened 2016-02-25
Init new URLs with default-port of 443, not 80

### LINKED ISSUE #4220: HTTPS Everywhere breaks SVG <use> elements (e.g. YouTube embed buttons, forecast.io clip-path), in Firefox
STR: **(NOTE: these YouTube STR no longer trigger the bug, as of ~June 2016. Use forcecast.io STR further down instead.)**
1. Install Firefox Nightly from https://nightly.mozilla.org/
2. Install HTTPS Everywhere from https://www.eff.org/Https-everywhere and restart Firefox to complete installation.
3. Visit http://www.youtube.com/embed/C4qgAaxB_pc and click the video to start playing.

(NOTE: the URL you type or paste in needs to be HTTP, not HTTPS. That is important.)

ACTUAL RESULTS:
Controls (e.g. play button) are not visible.

EXPECTED RESULTS:
Controls should be visible.

I diagnosed this in https://bugzilla.mozilla.org/show_bug.cgi?id=1250770 -- the issue is here:
https://github.com/EFForg/https-everywhere/blob/fbfaac74e0d2e15227ce69421028c2b337db66aa/src/chrome/content/code/HTTPSRules.js#L163

HTTPS Everywhere is initializing a `nsIStandardURL` with a "default port" of 80, but it's almost-certainly an HTTPS URI (given that it's an upgraded URI that HTTPS Everywhere is producing), so 80 is almost-certainly-wrong.

The bug goes away if I replace 80 with 443 in that init() call.

NOTE: This bug will likely not reproduce in Firefox Beta 45 or Firefox Aurora 46, because those versions have a [band-aid](https://bugzilla.mozilla.org/show_bug.cgi?id=1249452) for a related Gecko bug, and that band-aid happens to work around this bug as well. So -- to test this, use Firefox Nightly 47.

dholbert:
This fixes #4220 and https://bugzilla.mozilla.org/show_bug.cgi?id=1250770 (breakage with buttons on embedded http YouTube videos).

We could also do something fancy like testing `newurl` (the string) to see if it starts with "https" vs. "http", and choosing the default port we pass based on that. However, since we are HTTPS Everywhere, we should be pretty sure that the new URI will be HTTPS, so using 443 as our default port seems like a pretty good bet.

### DIFF
--- a/src/chrome/content/code/HTTPSRules.js
+++ b/src/chrome/content/code/HTTPSRules.js
@@ -160,7 +160,7 @@ RuleSet.prototype = {
       return null;
     var newuri = Components.classes["@mozilla.org/network/standard-url;1"].
                  createInstance(CI.nsIStandardURL);
-    newuri.init(CI.nsIStandardURL.URLTYPE_STANDARD, 80,
+    newuri.init(CI.nsIStandardURL.URLTYPE_STANDARD, 443,
              newurl, uri.originCharset, null);
     newuri = newuri.QueryInterface(CI.nsIURI);
     return newuri;


dholbert:
Some background on this change:

So this "default port" field (which I'm changing from 80 to 443) isn't actually used for much -- Firefox doesn't *actually* end up making a HTTPS connection over port 80, despite the fact that HTTPS Everywhere is kind of asking it to do so here.

But, one place where this field *does* get exercised is in same-origin checking code, for the SVG `<use>` element. That code compares two nsStandardURL objects to see if they're equal (one for the document, and one for the expanded `<use xlink:href=...>` URL).  And if they don't match, the <use> element doesn't render anything. In this case, one of those URLs comes from the channel where the document was loaded, and it has the bogus "port 80" URL that HTTPSEverywhere is currently creating in this code -- and the other URL is generated based on the content we're *actually* looking at, which knows it's being served over port 443. So we get a bogus URL mismatch.

If we create the upgraded URL with a default-port of 443 in the first place, then we don't get that bogus URL mismatch, and the <use> content will load correctly as it should.

dholbert:
[Tagging @jsha who's kindly helped me get HTTPSEverywhere issues fixed in the past]
Jacob, could you get this merged if all looks well, and/or let me know if there's any process I should follow / anyone else I should tag to take a look?  Thanks!

J0WI:
Let me also ping @fuglede and @semenko

fuglede:
@J0WI: Can't claim to have ever thought about that before, nor to have any good overview of the repercussions, but from @dholbert's description, it certainly sounds reasonable to me.

dholbert:
For the record, Firefox itself actually had its own version of this bug, with HSTS (strict transport security) http-->https upgrades, which I fixed in https://bugzilla.mozilla.org/show_bug.cgi?id=1247733 . 

(That instance was a bit more subtle than this one, though -- there, we were directly modifying an HTTP nsStandardURL object to upgrade it to HTTPS, and it was left with a stale default port, which I needed a new API to modify. Here in this HTTPS Everywhere bug, though, it's a bit more straightforward -- we're creating a nsStandardURL from scratch, so we should just create it with the correct default port up-front.)

J0WI:
Let me also ping @Hainish

PR #4222, opened 2016-02-26
Marktplaats.nl rule full rewrite

fathorse:
More mature ruleset. Handles lots dead ends, the cookiewall and odd behavior on misc subdomains


J0WI:
@fathorse you should have a look at the failing tests.

fathorse:
Marktplaats recently started a major overhaul of their website, adding https to http downgrades on a fair bit of pages, making this rather elaborate ruleset problematic. Will replace it with a simplified version.

PR #4223, opened 2016-02-26
gsmmap.org ruleset

fathorse:
Simple ruleset for gsmmap.org


J0WI:
I get an error on https://www.gsmmap.org you may comment it and rewrite to ^.

Note that we have a much simpler syntax for direct rewrites: 
```
<rule from="^http:"
	to="https:"/>
```

J0WI:
@fathorse there is still the issue on www.

J0WI:
closed by #6575

PR #4225, opened 2016-02-26
Simple.nl ruleset

fathorse:
Simple ruleset for simpel.nl (heh).


fuglede:
Looks good to me; a few simplifications are possible: The securecookies could be reduced to

and the actual rules to

PR #4226, opened 2016-02-26
Songkick ruleset

fathorse:
Simple ruleset for songkick.com


fuglede:
Looks like the subdomains <code>help</code>, <code>blog</code>, and <code>devblog</code> could potentially also be coverable? The subdomain <code>support</code>, however, does not work.

PR #4230, opened 2016-02-27
Recreate Deepin related ruleset.

lqf96:
Including "*.deepin.com", "*.deepin.org" and "*.linuxdeepin.com" domains.

### DIFF
--- a/src/chrome/content/rules/Deepin.xml
+++ b/src/chrome/content/rules/Deepin.xml
@@ -1,6 +1,8 @@
 <ruleset name="Deepin (partial)">
     <!-- Deepin.com -->
     <target host="www.deepin.com" />
+    <!-- Redirects to www subdomain -->
+    <target host="deepin.com" />
     <!-- Redirects to *.linuxdeepin.com: -->
     <target host="packages.deepin.com" />
     <target host="cdimage.deepin.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Deepin.xml
```diff
@@ -0,0 +1,49 @@
...
+    <!-- Redirects to *.linuxdeepin.com -->
+    <rule from="^http://(packages|cdimage)\.deepin\.com/" to="https://$1.linuxdeepin.com/" />
+    <rule from="^http://(packages)\.deepin\.org/" to="https://$1.linuxdeepin.com/" />
+    <!-- Redirects directly to HTTPS -->
+    <rule from="^http:" to="https:" />
+    
+    <!--
+    Mismatch:
+    wiki.deepin.org
+    blog.deepin.org
```
J0WI: Seems to working now (https://github.com/EFForg/https-everywhere/pull/4230/files#diff-b460b3e9ba52aa834775ba7d1b087c00R11)

J0WI:
I just realized that there is also a img3.linuxdeepin.com.
Please also redirect deepin.org/com to www.deepin.org/com. This is because the server side redirect doesn't support https.

lqf96:
OK, I've updated all of that.

J0WI:
Thanks!

PR #4236, opened 2016-02-28
[Songkick] Update ruleset

fuglede:
This adds a few subdomains as well as comments about broken ones.


J0WI:
tourbox WFM

fuglede:
Yep, here as well.

PR #4240, opened 2016-02-29
Added SSL vpn host.

jsavoie:


fuglede:
The atlseccon.com rule looks good to me. However, http://vpn.usainteanne.ca times out here, so I do not fully see the value of redirecting that one.

jsavoie:
There is no daemon listening on port 80 on vpn.usainteanne.ca, but I want the redirect to prevent clients from visiting http://vpn.usainteanne.ca through mitm/sslstripping.

PR #4244, opened 2016-03-01
New and improved Rules for 1and1 brands

Mit0x2:
**Affected Rules:**
- 1and1.xml(added)
- 1and1-Internet.xml (deleted)
- 1and1.es.xml (deleted)
- 1u1s.de.xml (deleted)
- 1und1.de.xml (deleted)
- GMX.xml (edited)
- Mail.com.xml (edited)
- Profiseller.de.xml (added)
- United-Internet.xml (edited)
- Web.de.xml (edited)
- UI_CDN.com.xml (deleted)
- UI_CDN.net.xml (deleted)
- UI_CDN.xml (added)

**Who am I?**
My name is Timo and I'm working at 1and1. Since some rules for our brands were broken or incomplete, we decided to fix them so that your and our customers have a great experience with your Extension on our websites.

**What has been done?**
- All rulesets for 1and1 domains were merged into one single file, since they all belong to one Company, 1and1. The rules were simplified and corrected.
- The GMX rules have been simplified and all necessary TLD's have been added. Some of the TLD's listed before are redirected to the now listed ones, so having an extra configuration for them is unnecessary.
- The Web.de rules have been simplified and completed. Instead of trying to match all subdomain, the rule is now targeting all subdomains (*.web.de) and makes an exception were no SSL certificate is used by the server.
- The UICDN configurations were merged into one single file and modified in a manner, which makes them not doing anything. It's documented and the files were left in there to prevent the creation of a new rule file for those domains.

**Testing**
All of the rules checked in were checked by our Quality Assurance Team by browsing all of our websites. We weren't able to find any strange behaviors, so please tell us if anything feels strange to you.

**Contact**
If you have any questions to any of the modified rules, feel free to contact me here: timo.schulz@1und1.de

Best wishes from Germany,
Timo


loveithateit:
@Mit0x2 can confirm that you are absolutely and justifiably sure that when @jsha ever runs the fetch tests that you will, in every instance indefinitely, update these rulesets before he disables them. Remember that https-everywhere is not important enough relative to his time's worth for him to spend more than 5 minutes or so every so often on it.

Secondly, I'm obviously not as skilled at git as you are, as I find it much easier to view and use git { log diff show revert cherry-pick } and so on with files which have not been renamed. Please could you share your technique that makes those tools just as easy to use across renamed files.

src/chrome/content/rules/1and1-Internet.xml
```diff
@@ -1,77 +0,0 @@
...
-	ᵖ Plaintext reply
-	ʳ Refused
-
-
-	Problematic hosts in *1and1.com:
-
-		- about ᵐ
-		- blog ᵐ
-		- faq ᵐ
-		- status ᴬ
```
loveithateit: This comment is still true, and yet you rewrite it. We do not like to break things, generally speaking.

src/chrome/content/rules/1and1-Internet.xml
```diff
@@ -1,77 +0,0 @@
...
-	ᵈ Dropped
-	ᵖ Plaintext reply
-	ʳ Refused
-
-
-	Problematic hosts in *1and1.com:
-
-		- about ᵐ
-		- blog ᵐ
-		- faq ᵐ
```
loveithateit: This comment is also true, and yet you now rewrite it directly. Again, we generally do not like to break things.

loveithateit: This comment is also true, and yet you now rewrite it directly. Again, we generally do not like to break things.

loveithateit:
Actually, all the comments you removed and ignored appear to be true still. This leads one to suspect that perhaps assuming good faith or competence is not warranted in your case.

In general, it's nicer when contributors think it worth their time to read and think before trying to make changes.

Mit0x2:
Hi @loveithateit ,
first of all thank you for your time taking a look into this pull request.
You might have the wrong first impression of myself. I'm not the uber worker at 1and1 who thinks he's the one and only. I'm just a little working student and people like you, who invest there free time in a project like this, have my full respect! I'm surely not as skilled with git as you are, but let me try to explain myself.

I thought it would be a good idea to merge the rules, because it makes them (in my eyes) easier to maintain. Instead of searching throught thousands of rules (as the rules i merged hadn't even a pattern the filenames followed) to find that last single rule to change. I'm clearly getting your point, it's a pain to review changes like this, but if it's worth it?!

I've spend time, much time, by trying to make the best out of the little I had. I spoke to many collegues who manage all the domains we use here at 1and1, but as you might know, it isn't always that simple in big companies. 

I'm totally with you, that purging all the comments, isn't the way to go and I'll respect that for my next attempt contributing to this project.

I just wanted to clear things up, so we're not sitting on each end of the wire having the wrong impression of each other. And maybe you can give me a hint, wether i should merge rules or leave their names like they are?

Thanks in advance and for your time,
Timo

loveithateit:
@Mit0x2 usually one domain per file is the way to go. This strikes a balance between making sense for users (which rules cover which domains) usability (easy to tell which rule to disable in the drop-down when the ruleset causes problems without making it so that a silly number of rulesets must be disabled) minimizing the amount of fallout on fetch test failures, and ease of reasoning through the rulesets. If you have reason to believe that a subset of that domain is more quirky or may break sooner than the rest, more than one file can be appropriate.

>I thought it would be a good idea to merge the rules, because it makes them (in my eyes) easier to maintain. Instead of searching throught thousands of rules (as the rules i merged hadn't even a pattern the filenames followed) to find that last single rule to change.

grep $domain_name or whatever your preferred alternative to grep is. cat *xml in rulesets/ to speed up the first time (or run makexpi.sh).

>I'm clearly getting your point, it's a pain to review changes like this, but if it's worth it?!

You're clearly not getting my point.

The point is not that it's a pain to review. Look at git log the **previous commit splits and states that it intends to split** (if you don't have the decency to check even the previous change, then please don't try to make any changes). The point of not renaming is to preserve history.

>And maybe you can give me a hint, wether i should merge rules or leave their names like they are?

Please leave them as they are.

Mit0x2:
@loveithateit thanks again for your time and for clearing things up. I'm sorry for missunderstanding your point.
Best wishes from Germany,
Mit0

PR #4249, opened 2016-03-01
Update ruleset

kalpetros:
Set Greek-travel ruleset to disabled
by default due to problems it is causing

### DIFF
--- a/src/chrome/content/rules/Greek-travel.xml
+++ b/src/chrome/content/rules/Greek-travel.xml
@@ -1,11 +1,14 @@
 <!-- This ruleset is experimental. If you experience problems please
 open an issue at https://github.com/kargig/https-everywhere-greek-rules. -->
-<!-- Update 01/March/2016: Greek-travel set to disabled by default due to problems it is causing -->
-
-<ruleset name="Greek-travel (buggy)" default_off="breaks site">
+<!-- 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Greek-travel.xml
```diff
@@ -1,10 +1,11 @@
 <!-- This ruleset is experimental. If you experience problems please
-	 open an issue at https://github.com/kargig/https-everywhere-greek-rules -->
+open an issue at https://github.com/kargig/https-everywhere-greek-rules. -->
+<!-- Update 01/March/2016: Greek-travel set to disabled by default due to problems it is causing -->
 
-<ruleset name="Greek-travel">
-  <target host="*.aegeanair.com" />
+<ruleset name="Greek-travel (buggy)" default_off="breaks site">
```
J0WI: seems to be `http redircet`

J0WI:
@kalpetros could you update the cause?
I'll merge it then.

kalpetros:
@J0WI I updated my pull request. It seems that this ruleset triggers mixed content and breaks a big part of the site.

J0WI:
Please replace wildcard and regex or provide some test urls to pass our tests:

bardiharborow:
@J0WI Fixed by #4931.

PR #4250, opened 2016-03-02
Veloviewer: Add ruleset

klusark:


fuglede:
Thanks for the PR. It looks like the subdomain <code>cf</code> might also be coverable, whereas <code>blog</code> unfortunately is not.

PR #4218, opened 2016-02-24
[LongTail] Comment out invalid rules, closes #4216

funkydude:
Simply a quick review of all the domains in this rule to fix https://github.com/EFForg/https-everywhere/issues/4216 before next release.

PR #4257, opened 2016-03-03
Redirect fullrate.dk to www.fullrate.dk

fathorse:
https://fullrate.dk = invalid cert

### DIFF
--- a/src/chrome/content/rules/Fullrate.xml
+++ b/src/chrome/content/rules/Fullrate.xml
@@ -3,6 +3,6 @@
 	<target host="www.fullrate.dk" />
 	<target host="forum.fullrate.dk" />
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http://(www\.)?fullrate\.dk/" to="https://www.fullrate.dk/"/>
+	<rule from="^http://forum\.fullrate\.dk/" to="https://forum.fullrate.dk/"/>
 </ruleset>


funkydude:
You should do this instead, which will also pass the checks.
`<rule from="^http://fullrate\.dk/" to="https://www.fullrate.dk/"/>`
`<rule from="^http:" to="https:"/>`

funkydude:
@fuglede Why are you merging PRs that 1) need work and 2) Don't pass Travis, when there are PRs that aren't waiting on either of these conditions, but are waiting to be merged?

fuglede:
@funkydude: The original Travis test failed out, then was superceded by one which passed, which in turn was superceded by one [which failed due to an unrelated error](https://github.com/EFForg/https-everywhere/pull/4326#issuecomment-195384311).

funkydude:
@fuglede you didn't counter point 1. The rule pointlessly uses more memory than required, and your prioritization of merging a rule that needs work when there's 122 outstanding pull requests is odd to say the least.

fuglede:
Regarding the point about memory, are you referring to the usage of <code>(www\\.)?</code>? If there is something to be optimised there, that would be good to know. Currently, we have 11127 occurrences of <code>(?:www\\.)?</code>, 618 of <code>(?:www)?</code>, 6563 of <code>(www\\.)?</code>, and  <code>19</code>. If you are referring to the 40-50 chars that could be saved in the forum rule, feel free to submit a PR. Regarding my prioritisation, I do not see how I can match that one up with whatever ideals you might have in that respect, if those ideals do not include prioritising the fixing of rulesets that cause sites to be unavailable to users of the add-on; indeed, that would -- to me -- be odd to say the least.

funkydude:
I'm referring to my first comment in this PR. That change would use the already available RegEx instead of pointlessly creating new ones: `<rule from="^http:" to="https:"/>`

Regarding prioritization: It's pretty awful. For example, who knows how sites are broken by a rule resolved in https://github.com/EFForg/https-everywhere/pull/4218 waiting 18 days now for merging. You were even mentioned a week ago.

The PR handling in general needs worked on, it seems @J0WI needs a medal in this regard for the time he puts in.

PR #4260, opened 2016-03-03
Ruleset: Update senate.gov to exclude Sessions

Hainish:


src/chrome/content/rules/Senate.gov.xml
```diff
@@ -15,7 +15,10 @@
 
 		<!-- The rules subdomain is all 403 via TLS -->
 		<!-- As is the Senate Intelligence Committee -->
-		<exclusion pattern="^http://(?:www.)?(?:intelligence|rules)\.senate\.gov/" />
+		<exclusion pattern="^http://(?:www.)?intelligence\.senate\.gov/" />
+		<exclusion pattern="^http://(?:www.)?rules\.senate\.gov/" />
```
loveithateit: Is this change really necessary?

Hainish: It is better to specify explicit routes rather than conditional regex, as
the style guide specifies.

The other change is necessary to view Rep. Sessions' page.

RReverser: The style guide talks about "long strings of subdomains" as those can be expressed via `<target host />` instead. It doesn't apply to exclusion pattern, especially for two subdomains.

Hainish: The point is that clarity is the most important factor. Given the new
exclusion for Rep. Sessions, having one exclusion encompass multiple
subdomains and others have just one is ugly and irregular.

RReverser: Ah right, missed the sessions below. I agree that consistent styling is more important, then (although I suspect significant slowdowns if all the rules are rewritten this way).

Hainish:
This is ready for review again

terrorist96:
The submitted https request was not able to be completed at this time. 
Please retry your request using http. This may require disabling some browser based plug-ins.

That's the message I'm getting. Disabling the senate.gov rule and forcing http fixes it.
Can this be set to default=off? @J0WI

J0WI:
Disabling the whole ruleset would be suboptimal for all the subdomains.

terrorist96:
> Disabling the whole ruleset would be suboptimal for all the subdomains.

Okay..

terrorist96:
Still broken in Version: 2016.5.10

Hainish:
It wouldn't be difficult to enumerate all the subdomains for members of congress.  In fact, we can simply parse the YAML provided by https://github.com/unitedstates/contact-congress to generate this list.  Then we could determine the problematic domains using a fetch test.  I may put in the effort to do this test at some point.

In the meantime, I'm in favor of incremental changes.

terrorist96:
Members of Congress change all the time though.

J0WI:
@Hainish referring to our style guide you should prefer capturing groups `(www\.)?` over non-capturing `(?:www\.)?`. 
Could you also remove the redirect for `www.senate.gov` (see https://github.com/EFForg/https-everywhere/pull/4260#issuecomment-210921247).

I'll merge it that as a bugfix to open the way for further improvements (as proposed in https://github.com/EFForg/https-everywhere/pull/4260#issuecomment-219176480).

terrorist96:
Still broken in 2016.6.9.

J0WI:
@Hainish are you planning to continue here?

Hainish:
I've been leaving this for when I have more time to pursue the method described in https://github.com/EFForg/https-everywhere/pull/4260#issuecomment-219176480.  This is pretty low on my priorities list, though.  In the meantime, being that https://www.senate.gov/ results in a 200 response and yet gives a message stating you should use http instead, I'm inclined to disable this rule altogether.  Or perhaps rewrite the rule to *only* target subdomains (such as https://www.cruz.senate.gov/ and https://www.scott.senate.gov/ that @terrorist96 mentioned above) which we know support https.

jeremyn:
Jeff Sessions is now the U.S. Attorney General and not in the U.S. Senate, though http://www.sessions.senate.gov/public/ still exists.

Other comments:

* https://www.senate.gov doesn't have that weird "HTTP 200 but you should use the unencrypted URL anyway" error
* The new common failure case seems to be HTTPS > HTTP redirects, for example http://www.sessions.senate.gov/public/ and http://www.senate.gov/states/

@Hainish Is this PR basically a catchall for fixing the senate.gov ruleset? If so, can we update the PR title? Also if you check the "allow edits from maintainers" option, I might make some changes myself.

terrorist96:
Well it looks like senate.gov supports https now, so this issue is moot.

jeremyn:
The existing ruleset isn't great for multiple reasons and needs heavy work. Either way though, this PR might as well be closed.

jeremyn:
The `senate.gov` behavior is also hardcoded in [`test/firefox/test/test-httpse-installed.js`](https://github.com/EFForg/https-everywhere/blob/a65d0bade57f82e919c5a871bec66e3e669c72bc/test/firefox/test/test-httpse-installed.js#L24-L36), to check that http://www.senate.gov/robots.txt is rewritten to https://www.senate.gov/robots.txt. We probably shouldn't use `senate.gov` as a test case given that the ruleset is so weird. For example I would not have been too opposed to simply disabling the ruleset entirely due to it being weird.

Hainish:
It should be safe to redirect `www.senate.gov` to HTTPS, and exclude specific senators until we get a comprehensive list of HTTPS support, since all senators' sites are hosted on subdomains of `senate.gov`.  The CIO Council specifically lists `www.senate.gov` as supporting HTTPS, and enforcing it: https://pulse.cio.gov/https/domains/#q=senate.gov

The above https://github.com/EFForg/https-everywhere/blob/a65d0bade57f82e919c5a871bec66e3e669c72bc/test/firefox/test/test-httpse-installed.js#L24-L36 should be changed, however, to a site in our rulesets that supports HTTPS but does not enforce it server-side.  This was the case previously for `www.senate.gov`.

jeremyn:
@Hainish Regarding a blanket rewrite for `www.senate.gov`, in https://github.com/EFForg/https-everywhere/pull/4260#issuecomment-278767388 I pointed out that https://www.senate.gov/states/ redirected to plain HTTP. Now both http://www.senate.gov/states/ and https://www.senate.gov/states/ return an HTTP 403. I don't think we can count on anything at `senate.gov` to be stable. 

I'm not totally sure what `test-httpse-installed.js` does, but it should be changed to check either an explicit flag or to check a site under EFF or other very friendly control. Otherwise the site admins of `senate.gov` can affect HTTPS Everywhere test results by doing something weird to http://www.senate.gov/robots.txt.

Hainish:
@jeremyn The intention of this particular code segment in the tests is to ensure that when visiting a site that ordinarily does not redirect to HTTPS but has a ruleset, visiting this site with HTTPS Everywhere installed results in you landing on the HTTPS site.  A basic functional test.  In the past, we've relied on `libssh.org` to do this, since for a long time it was a perfectly suited.  Recently, they've moved to HTTPS by default via a redirect.

The problem is that all current EFF sites redirect to HTTPS by default, thus do not qualify for this test.  We could register a new domain and create a new rule, but that requires some initial setup work.  This is probably the right way to go.

There was a project in the works some years ago to actually spin up a web server locally, install root certificates within the test suite environment, install a test ruleset at runtime, and make sure redirects are performed properly for a site hosted locally.  Apparently we were using the `apache` top-level directory for this purpose.  This would provide a environment where the test is very tightly controlled, but it would also add time onto the test suite and seems to me like over-engineering the solution.

Hainish:
I've gone with the option in the 2nd paragraph in https://github.com/EFForg/https-everywhere/pull/4260#issuecomment-279106153, and opened a ticket to have this site created.  Once it's created, I'll have to create a new ruleset to ensure redirection happens.

jeremyn:
@Hainish How does this test fit into the larger testing picture? For example imagine the site admins at `senate.gov` decided to make http://www.senate.gov/robots.txt return a 404. Would all our Travis tests break? Local `test.sh`? The installed add-on itself?

Also the new domain can be a subdomain of `eff.org`, right? `blahblahblah.eff.org` should be good enough for this test.

Hainish:
@jeremyn our tests would only break if the URL that is loaded by a browser with HTTPS Everywhere when you load http://www.senate.gov/robots.txt is in fact HTTP rather than HTTPS.

Essentially, the test checks whether you are actually being redirected to HTTPS or not, but it's *intended* to check whether it's HTTPS Everywhere that redirects you.  This means that for http://www.senate.gov and http://www.libssh.org the tests pass but we're not actually checking for the right thing, because it's the site itself, and not HTTPS Everywhere, that's redirecting you.

A subdomain of `eff.org` wouldn't work, because `eff.org` is in the HSTS preload list, and this includes subdomains.  So we'll be redirected to HTTPS no matter what, even if HTTPS Everywhere isn't doing it's job.

jeremyn:
I think I understand what the test is testing for, but I don't understand where the test is used. Imagine the most malicious actor possible at `senate.gov`. Could they make all our Travis tests fail? Could they break the build process that pushes a new version to the web stores? Something worse?

I had not noticed that http://www.senate.gov actually redirects itself. My earlier concern was only theoretical.

`hsts-prune` didn't catch `eff.org` when you ran it before. Do you know why?

Another URL we could use is http://www.example.com, which supports both plain HTTP and HTTPS. I also doubt IANA will themselves ever do an HTTP > HTTPS redirect, though this is just a guess. We'd need to make a ruleset for it, but we probably should have one anyway. Of course this is also a security risk, we're just trusting IANA instead of the U.S. senate.

Hainish:
> I think I understand what the test is testing for, but I don't understand where the test is used. Imagine the most malicious actor possible at `senate.gov`. Could they make all our Travis tests fail? Could they break the build process that pushes a new version to the web stores? Something worse?

No.  We wouldn't have relied on any 3rd party capable of this.  The worst that happens is the one test starts failing for Firefox.  Which we'd see immediately, and we'd have to move to a different site for this test which fits the criteria.  Which wouldn't be hard.

> `hsts-prune` didn't catch `eff.org` when you ran it before. Do you know why?

It did.  See https://github.com/EFForg/https-everywhere/pull/8128/files

> Another URL we could use is http://www.example.com, which supports both plain HTTP and HTTPS. I also doubt IANA will themselves ever do an HTTP > HTTPS redirect, though this is just a guess. We'd need to make a ruleset for it, but we probably should have one anyway. Of course this is also a security risk, we're just trusting IANA instead of the U.S. Senate.

We've traditionally excluded example.com from the rulesets as a control group.  When you think about what having HTTPS on `example.com` actually gets you, it's not a whole lot.  There are no subpages, logins, PII, or sensitive information that is contained there.  And a passive adversary can already tell you're going to `example.com` just from your DNS request.  So I think we should keep this without a ruleset, and just register a domain that we can control for this purpose.

jeremyn:
Thanks for clarifying about the test.

I see now that `hsts-prune` removed the `eff.org` domains. I'm not sure why I made that comment about `hsts-prune`.

I disagree with your analysis of `example.com`. It's a [`top-100k` site](http://www.alexa.com/siteinfo/example.com) and just as vulnerable to MITM modification, JavaScript insertion etc as any other site. Depending on its visitor demographics it might be a special target, for example it might be a good vector for a state actor to run an exploit on the computers of technical users who speak English.

Hainish:
@jeremyn a state actor would be in a position to MITM *any* plain-old-http site, and could do so without actually explicitly knowing the domain you're going to beforehand.  If you want to be protected from *that*, you should be using the 'Block all unencrypted connections' feature.

Hainish:
> I would not have been too opposed to simply disabling the ruleset entirely due to it being weird.

A few maintainers have separately endorsed this plan.  I'm cutting a new release today, so I've made an expedited decision to [remove this ruleset](https://github.com/EFForg/https-everywhere/commit/31cbbd6dc9ce2c1cf473cbad4e252b809a7a90c5) for now.

jeremyn:
@Hainish I'm not sure what you're arguing for here with the `example.com` discussion. It's a high traffic site that allows plain HTTP connections without redirecting. These are our prime ruleset targets. Would you allow the ruleset if I submitted it?

Hainish:
A new contributor comes to the project, and follows the rules for `./make-trivial-rule example.com`.  There's a utility in actually *seeing* that this new ruleset is working when you build the extension and run it, where it wasn't before.  This has to be weighed against any advantage to actually providing a rule for `example.com`.  I'm not against it, but it's useful to have a control group.  Perhaps this should be a separate domain which we register, rather than example.com.

Hainish:
A follow-up to the last few discussion points: in https://github.com/EFForg/https-everywhere/pull/9095 I changed the `senate.gov` URL to another URL which EFF controls.

jeremyn:
I'm closing this PR since `Senate.gov.xml` was deleted later in https://github.com/EFForg/https-everywhere/commit/31cbbd6dc9ce2c1cf473cbad4e252b809a7a90c5.

PR #4266, opened 2016-03-04
[Gesetze-im-Internet.de] New ruleset

rugk:


fuglede:
The first target host appears not to work at all:

    $ curl http://gesetze-im-internet.de
    curl: (6) Could not resolve host: gesetze-im-internet.de

rugk:
Indeed this does not work with Chromium and curl. However in Firefox it works for very strange reasons, but in the network console it does also only show request going to `http://www.gesetze-im-internet.de`.

fuglede:
Yeah, that's a bit strange. It fails (as it should) in Tor Browser (Firefox/38.0 ESR), though.

rugk:
> Yeah, that's a bit strange

So you can reproduce this? Okay, that's enough for me to open a issue on bugzilla - even if it is just some kind of misbehaviour....
So here you go: https://bugzilla.mozilla.org/show_bug.cgi?id=1253865
Comment there if you think more information would be necessary.

fuglede:
I can reproduce it in <code>Mozilla/5.0 (X11; Linux x86_64; rv:44.0) Gecko/20100101 Firefox/44.0</code> and in <code>Mozilla/5.0 (X11; Linux x86_64; rv:38.0) Gecko/20100101 Firefox/38.0 Iceweasel/38.6.1</code>, but not in Tor Browser (which is based on the  same ESR as Iceweasel, so that's a bit curious).

PR #4267, opened 2016-03-05
Add new rulesets

Bisaloo:
New rulesets added:

- mozillazine-fr.org
- mk2.com


src/chrome/content/rules/mk2.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="MK2">
+  <target host="mk2.com" />
+  <target host="www.mk2.com" />
+  <target host="m.mk2.com" />
```
J0WI: Please also add api.mk2.com and comments for refused media.mk2.com and mismatched preprod.mk2.com

src/chrome/content/rules/mozillazine-fr.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="mozillazine-fr (partial)">
```
J0WI: For what reason is this exposed as partial?

src/chrome/content/rules/mozillazine-fr.xml
```diff
@@ -0,0 +1,11 @@
...
+
+	<target host="mozillazine-fr.org"/>
+	
+	<exclusion pattern="^http://(www\.)?cname.mozillazine-fr\.org" />
+	<test url="http://cname.mozillazine-fr.org"/>
```
J0WI: This is not necessary as long as the url is not a target host. But please add comments for non-working hosts.

J0WI:
Thanks for your contribution.
Please separate unrelated hosts in different PRs next time.

Bisaloo:
I made the requested changes. Thank you for your comments, let me know if there is something else I should change.

J0WI:
Yes, you should mention that `www.mozillazine-fr.org` is also mismatched.

Bisaloo:
Ok, it should be better now.

PR #4276, opened 2016-03-06
Create Mbnet.pt.xml

ekaris:


fuglede:
Thanks for the PR! A few issues in this one:

* The test URL is actually redundant; the target hosts will automatically be tested in a ruleset like this.
* The non-www host does not actually resolve, so it should not be covered in the ruleset (shameless plug: #3107 would have caught this);

    $ curl https://mbnet.pt
    curl: (6) Could not resolve host: mbnet.pt

While not related to the ruleset, it's kind of a curious collection of headers that they have decided to serve:

    $ curl -I https://www.mbnet.pt
    ...
    Strict-Transport-Security: max-age=15552000; includeSubDomains
    Strict-Transport-Security: max-age=31536000; includeSubdomains; preload

ekaris:
The duplicate HSTS headers are against the specs so I will report it to them.

fuglede:
Yeah, there's the duplication, but also the <code>preload</code> token is not all that useful when the root domain does not resolve.

PR #4278, opened 2016-03-06
Update GHacks.net.xml

ekaris:


fuglede:
The page looks a bit weird here with this ruleset; a number of pictures around the site are not rendered, such as these ones on the front page:

ghacksnotquitegood

ekaris:
They are in the process on adding HTTPS by default, I have been in talks with them, so it will be fixed soon

fuglede:
Sounds good; closing this until the issue is fixed then, and we'll just reopen it when it is.

PR #4279, opened 2016-03-06
Include reference to subdomain finder

fuglede:
Inspired by @J0WI's consistently high subdomain coverage, this includes a reference to a straightforward-to-use Python script for using search engines to discover subdomains. It's not perfect but in my limited testing, it does do a pretty good job.

### DIFF
--- a/ruleset-style.md
+++ b/ruleset-style.md
@@ -21,14 +21,15 @@ which subdomains are covered.
 If you know all subdomains of a given domain support HTTPS, go ahead and use a
 left-wildcard, along with a plain rewrite from "^http:" to "^https:". Make sure
 to add a bunch of test URLs for the more important subdomains. If you're not
-sure what subdomains might exist, you can iteratively use google queries and enumerate
+sure what subdomains might exist, you can iteratively use Google queries and enumerate
 the list of results like such:
 
 1. site:*.eff.org
 2. site:*.eff.org -site:www.eff.org
 3. site:*.eff.org -site:www.eff.org -site:ssd.eff.org
 
-... and so on.
+... and so on. A script for doing exactly this is [currently available on Google
+Code Archive](https://code.google.com/archive/p/gxfr/).
 
 If there are a handful of tricky subdomains, but most subdomains can handle the
 plain rewrite from "^http:" to "^https:", specify the rules for the tricky


funkydude:
@jsha Should be merged

J0WI:
Is there any chance to get this in our tests?

fuglede:
@J0WI: So you would do something like add an requirement that every subdomain is mentioned in one way or another in the ruleset? (With an appropriate whitelist for existing rulesets?)

I can see how that would be useful to the ruleset editor looking to add complete documentation, but I'm also thinking that this would raise the bar for new (and old) contributors (for instance in the case where a user modifies a heavy ruleset, maybe to fix a bad bug, and then is told to also look into a pile of possibly unrelated domains).

J0WI:
Yes, it's quite difficult to maintain existing rules without having a clear separation between pure bugfix and rewriting/enhancements, also on the release schedule.
I required this in the documentation for most changes now. But mostly simple bugfixes already fails on our own tests, because the ruleset contains tons of other broken parts.

For new rulesets it would be a huge help and might speed up the review process. I already have to use simple bash/vim magic that produces me a list of `-site:`'s.

Sadly Google doesn't accept huge queries, but for most cases it works (not sure how this script will do it).

fuglede:
Is your magic of a kind that lends itself well to sharing? It generally performs much better than the script in OP.

J0WI:
Sadly not. It's just a ugly mix of bash, vim and copy-paste. Most parts are manual anyway because of the inconsistent way we comment problematic hosts.
Writing a short Python script would be a better approach.

jeremyn:
The URL in the commit, https://code.google.com/archive/p/gxfr/ , refers to http://ptscripts.googlecode.com/svn/trunk/gxfr.py , and the second URL 404s. There are various forks of it on GitHub, but if we still want this, we should probably include in one of the HTTPS Everywhere repositories directly.

Also, looking at [one of the forks](https://github.com/leonteale/pentestpackage/blob/master/gxfr.py), I see the code is pretty rough. I hope we'd improve it before including it.

fuglede:
That's unfortunate; I'd also be wary about just including it as-is. If nothing else because it should be possible to make something which functions better, does not require anyone to think about licenses, and which might even be maintainable by someone involved with the project.

Closing this for now, as the contents of the PR are redundant, but I would applaud the inclusion of a new script in `utils`.

jeremyn:
I'll just note here: a subdomain-finding trick I've been using is to look at the developer tools/network tab when a domain loads and look for resources that are pulled from subdomains.

galeksandrp:
Sublist3r mentioned in #6912

jeremyn:
Looking at the code for each, Sublist3r looks better written.

jeremyn:
Sublist3r now outputs its subdomains sorted the way we prefer them, see https://github.com/aboul3la/Sublist3r/pull/38 .

Issue #4273, opened 2016-03-06
Does HTTPS Everywhere allow rules that redirect to a different domain?


VanguardLH:
I ran into a problem visiting http://www.virusbtn.com/.  That site switched to a new domain, which is https://www.virusbulletin.com/.  The first uses HTTP which uses a Location header to point their HTTPS domain.  So visiting their HTTP site attempts to redirect the visitor to their new but different domained HTTPS site.  

I looked in the rules in HTTPS Everywhere regarding the virusbtn.com site.  All they do is change the http:// protocol to https:// but the virusbtn.com remains the same.  I realize that it is considered a no-no to redirect to a different domain since that can be used as a covert tactic to get the visitor to a different site than they specified and the visitor may not notice the URL change in the address bar of their web browser.  However, I also know that sites will sometimes use a new domain to build a new version of their site.  When they are ready to go public with their new web site, they put the Location header in their old site's home page to push visitors to their new site.

Switching from HTTP to HTTPS can sometimes break a site, especially if the site does it which opens a window of vulnerability between the HTTP and HTTPS connects.  Redirecting to a different domain can trigger an alert (by an aware user) that they may not be where they thought they would go.  So I am wondering if HTTPS Everywhere allows its rules to not only change the protocol (from HTTP to HTTPS) but to also change to a different domain for, in this case, a new site to replace the old one and the new site implements HTTPS.

If redirecting to a different domain while changing the protocol is permitted in the rules then the rule for virusbtn.com is out of date.  It should redirect http://www.virusbtn.com/ to https://www.virusbulletin.com.

fuglede:
Answering the question first of all: Not only is that allowed, it's actually pretty common. The most common use-case is for either adding or removing the "www" part of a hostname for a poorly configured domain. Another common one is for redirecting CDN buckets to their generic hostnames in which case certificates often do not cover the hosts the CDN delivers content to.

In both of those cases, it comes with a bit of a downside though, though, as it may break the site it attempts to fix the instant the site manager fixes their site themselves, so it is often better to not "help" the site too much, if it can be avoided.

One example could be the rule

    <rule from="^http://cms\.kiva\.org\.s3\.amazonaws\.com/"
            to="https://s3-us-west-1.amazonaws.com/cms.kiva.org/" />

In your specific case, we actually have to update the ruleset, since it looks like the https://www.virusbtn.com is misconfigured and could cause breakage:

    $ curl https://virusbtn.com
[...]

PR #4280, opened 2016-03-06
[VirusBulletin] Add new domain

funkydude:
Resolves https://github.com/EFForg/https-everywhere/issues/4273

### DIFF
--- a/src/chrome/content/rules/Virus_Bulletin.xml
+++ b/src/chrome/content/rules/Virus_Bulletin.xml
@@ -10,7 +10,7 @@
 
 	<securecookie host=".+" name=".+" />
 
-	<rule from="^http://virusbtn\.com/" to="https://www.virusbtn.com/" /><!-- Cert only matches www. -->
+	<rule from="^https?://virusbtn\.com/" to="https://www.virusbtn.com/" /><!-- Cert only matches www. -->
 	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/Virus_Bulletin.xml
+++ b/src/chrome/content/rules/Virus_Bulletin.xml
@@ -1,39 +1,16 @@
-<!--
-	Cert doesn't match !www.
-
-
-	Insecure cookies are set for these hosts:
-
-		- www.virusbtn.com
-
--->
 <ruleset name="Virus Bulletin">
 
-	<!--	Direct rewrites:
-				-->
+	<!-- Old domain automatically forwards to virusbulletin.com -->
 	<target host="www.virusbtn.com" />
-
-	<!--	Complications:
-				-->
 	<target host="virusbtn.com" />
 
+	<!-- New domain -->
+	<target host="virusbulletin.com" />
+	<target host="www.virusbulletin.com" />
 
-	<!--	Not secured by server:
-					-->
-	<!--securecookie host="^www\.virusbtn\.com$" name="^vbtn_session$" /-->
+	<securecookie host=".+" name=".+" />
 
-	<securecookie host="^www\.virusbtn\.com$" name=".+" />
-
-
-	<!--	Redirect keeps path and args,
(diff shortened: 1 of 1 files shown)


fuglede:
As mentioned in my comment in #4273, https://www.virusbtn.com seems problematic enough that it should not be used as a target for the forwarding.

Under normal circumstances, I would just point to #3107 and say that that would have caught the issue, but that does not actually seem to be the case: running <code>fetch-test.sh rules/Virus_Bulletin.xml</code> reveals no issues at all. While completely off-topic for this PR, this is rather confusing to me; https-everywhere-checker is based on <code>pycurl</code>, which does not like the site at all:

	In [1]: import pycurl
	In [2]: c = pycurl.Curl()
	In [3]: c.setopt(c.URL, 'https://www.virusbtn.com')
	In [4]: c.perform()
	---------------------------------------------------------------------------
	error                                     Traceback (most recent call last)
	<ipython-input-4-ca1d06f79dfb> in <module>()
	----> 1 c.perform()

	error: (60, 'SSL certificate problem: unable to get local issuer certificate')

fuglede:
Ah, seeing the issue now; I hadn't noticed that <code>https-everywhere-checker</code> uses its own root certs, which for whatever reason have discrepancies from the ones that are shipped with Debian Jessie.

	In [1]: import pycurl
	In [2]: c = pycurl.Curl()
	In [3]: c.setopt(c.URL, 'https://www.virusbtn.com/')
	In [4]: c.setopt(c.CAPATH, '/home/fuglede/https-everywhere/test/rules/platform_certs/chromium')
	In [5]: c.perform()
	<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
	...

@jsha / @hiviah, this seems like a problem?

funkydude:
(I was about to blame your cURL version)

@fuglede Debian has had issues with certs since January, it works fine in other (newer) Linux distributions and in every browser. Here's an example: https://einstein.phys.uwm.edu/forum_thread.php?id=11768

fuglede:
Yeah, I came across https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708. I didn't look into the background myself, but from that it looks like the removal is entirely on purpose. And probably warranted:

    $ curl -s https://www.thawte.com/roots/thawte_Premium_Server_CA.pem | openssl x509 -noout -text | grep Public-Key
                Public-Key: (1024 bit)

Should we not expect such changes to diffuse into other stores?

funkydude:
@fuglede The certificates change (directly) isn't to blame, it's the use of an old OpenSSL. In the case of Ubuntu 14.04, they patched theirs. https://bugs.launchpad.net/debian/+source/ca-certificates/+bug/1551615

fuglede:
But as far as I can tell from https://www.ssllabs.com/ssltest/analyze.html?d=www.virusbtn.com, only 1 path is provided?

funkydude:
@fuglede I don't know what's taking Debian so long to resolve a "grave" issue when it's already been resolved for Ubuntu, but I don't see the reason to change this rule to workaround it. This way when they kill off the old domain, people will actually update their bookmarks.

hiviah:
@fuglede https-everywhere-checker used its own certs which should have been kept in sync with whatever Firefox has, plus transitive closure of intermediate certs for validating chains with missing intermediate certs. But the certs just got a few updates and haven't been kept up to date.

It might be better to use 'static_ca_path' option, point to copy of your distro's certs. IIRC c_rehash utility from openssl needs to be run in the dir beforehand, otherwise openssl won't be able to lookup the proper certs (maybe this have changed and you just point it at /etc/ssl/certs).

fuglede:
@funkydude: Agreed on that.

@hiviah: Apologies for going completely off-topic in this PR, but why patch up the broken chains? We would want those to give errors since they will when browsers discover them in the wild

PR #4281, opened 2016-03-06
Create Guru3d.com.xml

ekaris:


funkydude:
You should forward requests from guru3d.com to www.guru3d.com and add a cookie rule.

ekaris:
@J0WI Can you pull?

funkydude:
@ekaris Please fix your rule to also forward https queries: `from="^https?://guru3d\.com/"`

You're still missing a cookie rule, you should start adding these for complete coverage.

Also list the following as refused:
`capone.guru3d.com`
`downloads.guru3d.com`
`forums.guru3d.com`

J0WI:
I didn't get any cookies on www.guru3d.com.
The redirect for https shouldn't be necessary here, because they are no links pointing to this url (see discussion in #4303).
However, you also have to put ^guru3d.com in the list of refused hosts.

ekaris:
Good?

funkydude:
@J0WI you are probably blocking them, I see 5 cookies.
You have no way of knowing what or who links to this website

ekaris:
I only get 1 cookie and it's empty, besides using securecookie on all rules is a theme for a discussion on mailing list, can you pull @J0WI ?

funkydude:
@ekaris No, it isn't. Setting secure cookie on all rules would be stupid, hence why it's optional. It's to be used on a case by case basis where it doesn't break a website. In this situation the rule is so simple it can be safely used.

Also, stop demanding your PRs be merged, ask for review. They will be merged when they are ready, yours isn't. You need to fix your redirect for ^guru3d.com as I already stated.

J0WI:
I agree that this discussion should be somewhere else. I never required to secure cookies in the past and I won't require it here - nevertheless it would be nice to have.

But the redirect from ^ to www is a real blocker.

funkydude:
@J0WI Not trying to imply that a securecookie rule should be a requirement. However, as a frequent contributor, it is something he should consider adopting for complete coverage.

funkydude:
Only issue I see is you've accidentally got 4 w's in the cookie host. For future reference, you can just use `.+` in basic rules like these. Doing so will also save memory, as a new RegEx won't be created.

The test URL shouldn't be needed I don't think ( @J0WI ?). The script *should* handle it.

J0WI:
Thanks!

PR #4282, opened 2016-03-06
Add a dot

fuglede:
This adds a single period to the about window.


funkydude:
I'm not sure, I think this would complicate things.

fuglede:
@funkydude: Good thing we have Travis.

PR #4287, opened 2016-03-07
Additionals domains under cr.yp.to

MilanKral:


fuglede:
Thanks for the PR; it looks good to me, but could you update the header comment to reflect your change? Having the target hosts ordered alphabetically would also be nifty.  Feel free to add (and test) the below as well:

Some other subdomains that do not work:
* rump2007 (bad cert)
* rump2008 (bad cert)
* rump2009 (bad cert)
* rump2010 (bad cert)
* rump2011 (bad cert)
* snakeoil (connection denied)
* pairings.temporary (SSL error)

Some subdomains that could be covered:
* cubehash
* eecm
* pi
* untied

fuglede:
👍

PR #4290, opened 2016-03-07
add rule for Icony.com

numismatika:


J0WI:
I also found flirt and jira.
You should also redirect ^ to www and comment the mismatch.

J0WI:
@numismatika are you still working here?

numismatika:
it was already done but not pushed to github.

PR #4294, opened 2016-03-07
add rule for HFO Telecom

numismatika:


J0WI:
WFM:
rechnung.hfo-telecom.de
portal.hfo-telecom.de

PR #4296, opened 2016-03-07
Create G2a.com.xml

ekaris:


funkydude:
You're lacking `g2a.com` and a cookie rule.

ekaris:
@fuglede Can you clarify the need of a cookie rule? All my pulls where accepted without one.

funkydude:
You highlighted the wrong username.

The cookie rule sets the "secure" flag on all cookies to guarantee the browser transmits it over HTTPS. If you've not been using them, it's time to start.

ekaris:
@funkydude I know what securecookie does, It was intended, I just want a clarification from someone with commit access.

fuglede:
Let's see; at first glance, the <code>securecookie</code> will not do all that much, since the connection is secured before any cookies are sent insecurely. One place where I can see it making a difference is if one tries to access a subdomain, not covered by the ruleset, insecurely, in which case a cookie without the secure flag from the higher-level domain would be accessible when maybe it should not be.

I can't really come up with other examples than that, though, which in particular means that I don't really see a point with <code>name=".+" host=".+"</code> on wildcard-subdomain targets, but I could be missing something there.

funkydude:
>I know what securecookie does, It was intended, I just want a clarification from someone with commit access.

You asked to clarify the need for it, that's what I did?

@fuglede You're pretty spot on, but since it adds another layer of protection, there's no reason not to use it.

ekaris:
Most rules don't use it, why don’t pull this and discuss the problem on mailing list and if people want run a script on all rules to add the secure cookie flag to the ones that don’t have it?

fuglede:
Scripting it for other things than wildcard rulesets might be tricky due to the behavior of cookies across subdomains, but yep, that's for a separate discussion.

PR #4299, opened 2016-03-08
Remove Bothar.xml

RReverser:
Using lookahead with substitution causes `$1` to appear in the end URL:

```javascript
> 'http://bothar.ie/'.replace(new RegExp('^http://(?:www\.)?bothar\.ie/'), "https://$1bothar.ie/")
"https://$1bothar.ie/"
```

Same problem as older #2980.

### DIFF
--- a/src/chrome/content/rules/Bothar.xml
+++ b/src/chrome/content/rules/Bothar.xml
@@ -1,6 +1,0 @@
-<ruleset name="Bothar">
-  <target host="www.bothar.ie" />
-  <target host="bothar.ie" />
-
-  <rule from="^http:" to="https:"/>
-</ruleset>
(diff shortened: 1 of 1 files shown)


funkydude:
There is no need for one here, the basic `^http:` will work fine.

RReverser:
@funkydude Yeah, agreed.

jsha:
Actually, www.bothar.ie has an invalid cert. The intent of this rule appears to be redirecting www.bothar.ie to bothar.ie in addition to the HTTPS. Here's what I'd recommend:

```
<rule from="^http://www.bothar.ie" to="https://bothar.ie" />
```

funkydude:
@jsha they both have invalid certs. This rule should be off by default.

@RReverser I guess you didn't check the certs?

RReverser:
@funkydude No, as you saw in my initial commit, I was just fixing all the automatically found invalid regexps.

jsha:
Good point. Since this rule is totally broken and doesn't seem to have history worth preserving, I think you can outright delete it rather than set `default_off="..."`.

RReverser:
@jsha Sounds good. Updated PR.

PR #4300, opened 2016-03-08
Fix invalid RegExp in Fox_News.xml

RReverser:
Same problem as #4299 and older #2980.

### DIFF
--- a/src/chrome/content/rules/Fox_News.xml
+++ b/src/chrome/content/rules/Fox_News.xml
@@ -21,12 +21,6 @@
 	** 403, akamai
 
 
-	Problematic domains:
-
-		- www.fbnstatic.com *
-		- ureport.foxnews.com		(works; mismatched, CN: *.filemobile.com)
-
-
 	Fully covered domains:
 
 		- global.fncstatic.com
@@ -34,41 +28,18 @@
 -->
 <ruleset name="Fox News (partial)">
 
-	<target host="www.fbnstatic.com" />
 	<target host="global.fncstatic.com" />
 		<!--
 			- css:	503
 			- js: 404
-					-->
-		<exclusion pattern="^http://(?:www\.fbn|global\.fnc)static\.com/static/(?:.+/)?(?:j|cs)s/" />
-	<target host="*.foxbusiness.com" />
+		-->
+		<exclusion pattern="^http://global\.fncstatic\.com/static/(?:.+/)?(?:j|cs)s/" />
+	
 	<target host="*.foxnews.com" />
-	<target host="facebook.foxnewsinsider.com" />
+		<exclusion pattern="^http://(?:www\.|latino\.|magazine\.|nation\.|radio\.)?foxnews\.com/" />
+	
 	<target host="foxnews.ramp.com" />
 
-
-	<rule from="^http://(www\.)?fbnstatic\.com/"
-		to="https://a57.foxnews.com/$1static.com/" />
-
-	<rule from="^http://global\.fncstatic\.com/"
-		to="https://global.fncstatic.com/" />
-
-	<rule from="^http://(beta\.)?video\.foxbusiness\.com/"
-		to="https://$1video.foxbusiness.com/" />
-
-	<rule from="^http://(a57|ads|video\.(?:imag|insider|latino|ureport)|live|video|beta\.video(?:\.latino)?)\.foxnews\.com/"
-		to="https://$1.foxnews.com/" />
-
-	<rule from="^http://metrics\.foxnews\.com/"
-		to="https://foxnews.112.2o7.net/" />
-
-	<rule from="^http://uereport\.foxnews\.com/services/"
-		to="https://filemobile.com/services/" />
-
-	<rule from="^http://facebook\.foxnewsinsider\.com/"
-		to="https://facebook.foxnewsinsider.com/" />
-
-	<rule from="^http://foxnews\.ramp\.com/"
-		to="https://foxnews.ramp.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/Fox_News.xml
+++ b/src/chrome/content/rules/Fox_News.xml
@@ -47,7 +47,7 @@
 	<target host="foxnews.ramp.com" />
 
 
-	<rule from="^http://www\.fbnstatic\.com/"
+	<rule from="^http://(www\.)?fbnstatic\.com/"
 		to="https://a57.foxnews.com/$1static.com/" />
 
 	<rule from="^http://global\.fncstatic\.com/"


jsha:
LGTM, but tests are failing.

RReverser:
@jsha It's not tests - looking from logs, Travis just crashed yesterday for various PRs https://travis-ci.org/EFForg/https-everywhere/jobs/114529777

jsha:
I restarted the build after fixing the overall build brokenness, and it looks like this one is legitimately failing.

RReverser:
Ah, not enough tests... Thanks, I'll add some.

RReverser:
Lack of the tests aside, could you please take a look at the changes I did in the updated commit? I walked through the covered URLs and removed rules that covered non-(no-more-?)existent domains and those that were non-functioning from the beginning, so ruleset became scarily small.

RReverser:
@jsha This is still a problem (`http://www.fbnstatic.com/` is rewritten to `https://a57.foxnews.com/$1static.com/`).

jsha:
Hi @RReverser,

Sorry for the delay in reviewing! This change looks good. The failing tests say:

```
ERROR src/chrome/content/rules/Fox_News.xml: Not enough tests (0 vs 4) for <Exclusion pattern '^http://global\.fncstatic\.com/static/(?:.+/)?(?:j|cs)s/'>
ERROR src/chrome/content/rules/Fox_News.xml: Not enough tests (0 vs 6) for <Exclusion pattern '^http://(?:www\.|latino\.|magazine\.|nation\.|radio\.)?foxnews\.com/'>
```

What that's saying is that the tests want some examples of URLs that should be excluded. For instance, for the second one, it wants 6 tests of the form:

```
<test url="http://radio.foxnews.com/" >
```

For the former, if wants tests that match that particular regex. Since that regex can't "fix" a page that would otherwise have blocked mixed content, I would be okay with removing that global.fncstatic.com domain if it's not practical to find test cases for it.

jeremyn:
Targets with wildcards are discouraged. The ruleset should be rewritten to include `target`s for each specific subdomain.

@RReverser Are you still interested in working on this?

RReverser:
@jeremyn Why are they discouraged? How otherwise to enforce SSL for all subdomains? (especially when you don't know all of them)

I just saw this comment, but otherwise the rule is still broken, so yes, I guess I'm still interested.

jsha:
I would state it a little less broadly: Targets with wildcards are good if you have high confidence that you know about all important subdomains and have tested them all for HTTPS support. In most cases it's hard to know this. But when it's possible to know, wildcards are great.

jeremyn:
I consider the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) to discourage wildcards, but it does allow for them.

There are some big problems with wildcards. One is that they cover domains that aren't available publicly. Another is that they cover future domains that don't exist yet. Also, "high confidence that you know about all important subdomains and have tested them all for HTTPS support" is a fuzzy thing to test for or review. Fuzziness leads to arguments and wasted time.

There are some strategic uses for wildcards. [Here](https://github.com/EFForg/https-everywhere/blob/9698e64a2de7cf37509ab13ba9dcfd5bd4f84a95/src/chrome/content/rules/360.cn.xml#L98-L103) is an example of using a wildcard with a specific rule to cover a lot of repetitive subdomains without listing them out. Another situation I can imagine but don't have an example of would be a site owner with a lot of subdomains officially promising, in a public blog post or press release, to support and test HTTPS on all current and future pages.

In practice though, wildcards are either inappropriate because the site is small enough that you can list out the subdomains easily, or large enough that the wildcard breaks things.

jsha:
As the original author of the style guide, probably I was being too
ambiguous. Here are the two relevant bits, from different paragraphs:

> Avoid using the left-wildcard (<target host='*.example.com' />) unless
you intend to rewrite all or nearly all subdomains. Many rules today
specify a left-wildcard target, but the rewrite rules only rewrite an
explicit list of hostnames.

This was specifically a reaction to a then-common pattern of using broad
matching in the <target> tag, but then more narrow matching in regexes.
The goal of this paragraph is not to encourage or discourage using
left-wildcards, but to say that if your ruleset effectively targets only
certain exact hosts, it should reflect that in the <target> tags. That
was to allow better programmatic processing of the rulesets.

> If you know all subdomains of a given domain support HTTPS, go ahead
and use a left-wildcard,

This was where I intended to encourage appropriate use of wildcards.

You're correct that the rule is fuzzy. However, sometimes that's
reasonable. For instance, if you're concerned that there may be internal
subdomains that don't support HTTPS and are not accessible on the public
Internet, I think it's fine to go ahead with a wildcard ruleset until we
get a bug report from someone using one of those internal subdomains.

jeremyn:
I gave reasons why "you know all subdomains of a given domain support HTTPS" is essentially never, because of internal and future domains, so I interpret the style guide as discouraging them.

In practice I almost never see a valid use of wildcards. I assume it means the reviewer doesn't know how to search for subdomains or doesn't understand why that's important. My first HTTPS Everywhere PR (https://github.com/EFForg/https-everywhere/pull/4821) is an example of a bad wildcard. If a contributor wants to use wildcards then it is their responsibility to show they've checked current domains and to describe why a wildcard is more appropriate than specific targets given the problems wildcards have.

Waiting for the bug report is not a good solution. Turnaround time on issues and PRs is not fast. Also with internal problems there is no easy way to verify the problem or fix, and verifying is important to make sure it's not some sort of attack.

@jsha I've discussed the problems with vague documentation with @Hainish and I don't want to have that discussion here. Please feel free to talk to him or email me if you want more details.

RReverser:
Sorry I don't think I'll be able to continue on the requested changes, but if someone wants to take over this PR, please do (I just ticked the "allow edits from maintainers" checkbox).

jeremyn:
I'd like to do something with this PR but I'm not sure what.

@RReverser Are you willing to walk me through what's going on with this PR and/or make needed changes? (For example, https://foxnews.ramp.com doesn't exist anymore, and https://www.foxnews.com seems to work and shouldn't have an `exclusion`.) Otherwise, we're in a situation where neither the PR submitter nor the maintainers want to do anything with it, which suggests we should just close the PR entirely.

RReverser:
@jeremyn Well the original issue and fix from https://github.com/EFForg/https-everywhere/pull/4300/commits/fcc31b48758004fda4a5202de07354c3ee0a30b1 still stands as rule is still broken (due to absent group generating `$1` in the resulting URL), but since throughout review a one-line fix turned out into pretty much refactoring of entire Fox News ruleset, I already said that I don't have time to continue the effort.

The original issue still occurs though and present in the XML.

jeremyn:
Yes, this PR got sort of complicated. I'm okay with making the minimal changes needed for the bug and then adding this ruleset to the [whitelist](https://github.com/EFForg/https-everywhere/blob/d980b8b4c619459faf6c3f1e60805689bc38ab23/utils/ruleset-whitelist.csv) to avoid the problems with the tests.

Looking at just https://github.com/EFForg/https-everywhere/commit/fcc31b48758004fda4a5202de07354c3ee0a30b1 and the new `rule`:

```xml
	<rule from="^http://(www\.)?fbnstatic\.com/"
		to="https://a57.foxnews.com/$1static.com/" />
```

`fbnstatic.com` doesn't exist so we don't need the `?`. (Also there is no `target` for `fbnstatic.com`.) In fact it looks like the `rule` could be just:

```xml
	<rule from="^http://www\.fbnstatic\.com/"
		to="https://a57.foxnews.com/www.static.com/" />
```

except that https://a57.foxnews.com/ww.static.com/ returns a `404`. Do you have an example of a specific page with broken behavior (missing stylesheet etc) that is fixed by this change?

jeremyn:
I'm closing this PR due to lack of information about how to reproduce the problem this is supposed to fix.

RReverser:
@jeremyn I provided an information about the invalid rule, and it's still there. I have no idea what subdomains it would match as I'm not author of the rule or original website, but it does show up in our tests as failing each time due to advanced checks that detect `$1` without matching group.

The rule still needs to be either removed or fixed.

jeremyn:
Okay, I understand the problem better. The error you're fixing is that the original `rule` rewrites `http://www.fbnstatic.com` to `https://a57.foxnews.com/$1static.com/` with a literal `$1`, which is obviously wrong.

One URL for `www.fbnstatic.com` is http://www.fbnstatic.com/markets/earnings-calendar .

* With HTTPS Everywhere disabled, the (presumed correct) response is a `301` redirect to http://www.foxbusiness.com/category/markets/earnings.html .

* With HTTPS Everywhere enabled and using the original rule, the request is rewritten to https://a57.foxnews.com/$1static.com/markets/earnings-calendar and gets a `404` response.

* With HTTPS Everywhere enabled and using the new rule from https://github.com/EFForg/https-everywhere/commit/fcc31b48758004fda4a5202de07354c3ee0a30b1, the request is rewritten to https://a57.foxnews.com/www.static.com/markets/earnings-calendar and gets a `404` response.

So the effect of updating the rule for this URL is to change from one failure mode to another.

The ruleset is [disabled by default](https://github.com/RReverser/https-everywhere/blob/9bf49cbeb4696cdea76df23f685da1d11031f693/src/chrome/content/rules/Fox_News.xml#L39) now anyway so it's not causing any problems either way.

Let's just keep this closed and someone else can feel free to come along and fully update this ruleset if they want. What do you think?

RReverser:
> Let's just keep this closed and someone else can feel free to come along and fully update this ruleset if they want. What do you think?

That's fine by me as long as noone accidentally reenables this ruleset without fixing the rule.

PR #4303, opened 2016-03-08
Add ruleset for www.parkerpen.com

Liudvikam:


J0WI:
I also found origin.parkerpen.com and mgmt.parkerpen.com (mismatch).
Please also comment the mismatch on ^parkerpen.com.

src/chrome/content/rules/Parkerpen.com.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="parkerpen.com">
+	<target host="parkerpen.com" />
+	<target host="www.parkerpen.com" />
+	<target host="assets.parkerpen.com" />
+	
+	<rule from="^https?://(www\.)?parkerpen\.com/"
```
J0WI: We don't touch https

Liudvikam:
> I also found origin.parkerpen.com and mgmt.parkerpen.com (mismatch).

Comment added. Also, mgmt.parkerpen.com isn't actually a mismatch, so I added it to the ruleset.

> Please also comment the mismatch on ^parkerpen.com.

Done.

> We don't touch https

I did that because if someone visits https://parkerpen.com, they get a cert error. The redirect fixes that. Should I still remove it?

fuglede:
Quite a few rules actually use the <code>https?</code>trick,

    https-everywhere/rules$ grep 'from="^https?' * | wc -l
    79

but yeah, it can lead to some undesired and unnecessary maintenance.

J0WI:
I'll merge this now after retesting, but we should clean the `https?` rules up.

PR #4308, opened 2016-03-09
Add Not Always ruleset

VoidPhantom:


J0WI:
Could you also add (www.)notalwayshopeless.com?

VoidPhantom:
Is that the new one? Right now it seems to be an empty look-alike of Not Always Learning.

J0WI:
I don't know, I just found that domain and it seems to be related and working.

VoidPhantom:
Alright, so this now clearly states that Not Always Hopeless is a new Not Always website that is going to launch on March 18. I figured it’s OK to add it to this ruleset, so here goes.

J0WI:
Thanks!

PR #4312, opened 2016-03-09
[XNET.cz] Removed unnecessary redirection

Roman3349:


J0WI:
The following works for me:
email.xnet.cz

Could you document the mismatch on:
blog.xnet.cz
wiki.xnet.cz
www.xnet.cz

Refused
webftp.xnet.cz
redirect.xnet.cz
ads2.xnet.cz (ads1 does not exist)

The name of the ruleset is quite irritating, so you may also comment, that https is refused on (www.)kraxnet.cz.

Roman3349:
@J0WI Ruleset is renamed and I documented mitmatches.

fuglede:
Thanks!

PR #4318, opened 2016-03-09
added Blick.xml

J0WI:


funkydude:
`app.blickamabend.ch` causes breakage for me, mixed content

Could use a `"^\w+\.blick\.ch$"` cookie rule to only secure the subdomains that fully support HTTPS.

J0WI:
rebased.

J0WI:
Now this seems finally ready to review :)
cc @jeremyn

jeremyn:
Refused:
- [ ] https://gutscheine.blick.ch

Mixed content:
- [ ] https://f.blick.ch
- [ ] https://f1.blick.ch
- [ ] https://f2.blick.ch
- [ ] https://f3.blick.ch
- [ ] https://app.blickamabend.ch

Redirect:
- [ ] https://amp.blickamabend.ch

J0WI:
I didn't notice that you can use the CDNs as an alias, so I use a more restrictive redirect rule now.

jeremyn:
The exclusions in your latest commit baa027ad9b22c14e93b5387cc850c9e6539f8cca aren't necessary because no rule matches them. I would delete them, and move [this](https://github.com/J0WI/https-everywhere/blob/baa027ad9b22c14e93b5387cc850c9e6539f8cca/src/chrome/content/rules/Blick.xml#L47) somewhat confusing (IMO) comment to the top and change it to something like:

`Static content from these URLs work, but as an alias for ^(www.) they have mixed content issues.`

J0WI:
They are necessary to pass the tests, because the targets are used as test urls and Travis find no rule matching them. (see https://travis-ci.org/EFForg/https-everywhere/jobs/160697291#L293)

jeremyn:
Oh, I see, thanks. The exclusions are fine then.

Can you replace the inline comment with the one in my previous comment (or something like it)? I specifically don't like `CDN url`.

J0WI:
using your proposal now.

jeremyn:
Thanks, merged.

PR #4319, opened 2016-03-09
added cash.ch.xml

J0WI:


fuglede:
The subdomain beta has some issues; for instance, https://beta.cash.ch/news/top-news/preise-zentren-kommen-unter-druck-428487?autostart=1 is broken. Aso, some content from disqus is blocked; is that a problem?

If you want to, some parts of the www domain, could be secured, cf. e.g. https://www.cash.ch/services/img/news/235/111057_Credit_Suisse_New_York_0.jpg.

J0WI:
I thought about the static content on `www`, but I think it's not worth when they launch the new `beat`site soon.

The only problematic content is the video from simplex.tv. I 'll see if they fix this before the launch.

fuglede:
Makes sense; do you want to remove the problematic domain or close/reopen this PR then to keep the backlog cleaner?

J0WI:
I made an exclusion for now.

PR #4331, opened 2016-03-11
Fix failing ruleset tests.

jsha:
@loveithateit: The ruleset check was not broken, but was finding a real problem
with the ruleset: URLs like "http://marketing.yell.com///" don't get rewritten
by the rule you were adding tests to, because they get caught by the exclusion.
I've fixed the rule to only apply to the root URL, but please double check that
this matches your intention.

cc @cooperq for code review.


cooperq:
lgtm

PR #4335, opened 2016-03-12
Create Groupees.com.xml

ekaris:


fuglede:
It looks like the subdomains www and blog could also be covered in this one, doesn't it?

ekaris:
Done

fuglede:
Great, thanks!

Issue #3975, opened 2016-01-23
https://ctan.org just works


foobar13373:
https://ctan.org supports HTTPS.

PR #4336, opened 2016-03-12
[CTAN] New ruleset

fuglede:
This closes https://github.com/EFForg/https-everywhere/issues/3975.


J0WI:
There is also a mirrorS and a ftp, but they don't serve a valid cert.

fuglede:
Updated.

PR #4339, opened 2016-03-12
[VSHosting] Added ruleset

Roman3349:


J0WI:
Sadly, it seems that every customer gets his own subdomain. This results in a bunch of subdomains, that doesn't work with https.

Could you document this behaviour?

J0WI:
@Roman3349 are you still working on this?

Roman3349:
@J0WI Added comments about non-functional subdomains.

J0WI:
Great!

PR #4346, opened 2016-03-12
[EZ-Web-Hosting] Update ruleset

Roman3349:


J0WI:
church.ez-web-hosting.com also works for me.

Sadly, it seems that every user gets a domain on ez\d+.ez-web-hosting.com/~user. This results in many cert or mixed content failures. Could you document this behaviour?

Roman3349:
Done

J0WI:
Rule LGTM, but conflicts with previous change.

Roman3349:
@J0WI Done

Issue #4082, opened 2016-02-03
The green checkmark (showing the rule is enabled) is not visible


ghost:
There are two modes, apparently: 
1. Either the rule is enabled but no green checkmark can be seen, although the rule name is green(or sometimes gray depending on if it was re-enabled on a different webpage and you're looking at the webpage which was loaded with it disabled - but still no red checkmark on this webpage)
2. Or the rule is disabled and the red X mark appears, with rule name being gray.

Tested https-everywhere commit: 9c0aad1329a3a04d24d555b7c69e18d40333141f HTTPS-Everywhere 5.1.2  Last Updated: 02/03/2016

Expected: the enabled rules to have the green checkmark, but they are lacking it(for unknown reasons) as seen here:

![lack of green checkmark](https://cloud.githubusercontent.com/assets/17016209/12795886/82e16310-cabc-11e5-9116-8bb729719635.png)

Clicking one of them, makes it red:
![red x](https://cloud.githubusercontent.com/assets/17016209/12795887/82e6b248-cabc-11e5-92f8-81c5f1833b44.png)

Tested on firefox 44.0, here's `about:buildconfig`
```
Build platform
target
x86_64-pc-linux-gnu
Build tools
Compiler 	Version 	Compiler flags
x86_64-pc-linux-gnu-gcc 	5.3.0 	-Wall -Wempty-body -Wpointer-to-int-cast -Wsign-compare -Wtype-limits -Wno-unused -Wcast-align -pipe -march=native -g0 -fno-stack-protector -fPIC -std=gnu99 -fgnu89-inline -fno-strict-aliasing -fno-math-errno -pthread -pipe
x86_64-pc-linux-gnu-g++ 	5.3.0 	-Wall -Wempty-body -Woverloaded-virtual -Wsign-compare -Wwrite-strings -Wno-invalid-offsetof -Wcast-align -pipe -march=native -g0 -fno-stack-protector -fPIC -fno-exceptions -fno-strict-aliasing -fno-rtti -fno-exceptions -fno-math-errno -std=gnu++0x -pthread -pipe -DNDEBUG -DTRIMMED -freorder-blocks -Os -fomit-frame-pointer
Configure arguments

[...]

Issue #2618, opened 2015-08-20
Icons for rules missing in context menu


rugk:
![httpse_iconsmissing](https://cloud.githubusercontent.com/assets/11966684/9388261/a6da9af6-4764-11e5-99f6-baf3f7388d13.png)

I think there should be tick/red reload icons or something like this.

FYI the activate/deactivate also looks different and probably this could be the reason (although the 'activate/deactivate' look is intended as I assume):
![httpse_iconsmissing2](https://cloud.githubusercontent.com/assets/11966684/9388431/9d130336-4765-11e5-8bec-9dfce900d87e.png)

Add-on 5.0.10
Firefox 40.0.2
Windows 7

rugk:
The issue still appears in v5.1.1.

vgturtle127:
@rugk @fuglede @jsha 

No icons on HTTPS Everywhere 5.1.3 using 64-bit Firefox 44.0.2, 64-bit Firefox Developer Edition, Google Chrome 48, all on 64-bit Windows 10 Pro.

I never knew there were supposed to be icons! :cry:

gloomy-ghost:
Closing as obsolete since we have migrated to WebExtension.

PR #4349, opened 2016-03-12
Revert moot icon to its previous state of existence

fuglede:
'Moot' was removed, then partially reintroduced, in d28729967680a70d06f916176b8da1350c563227 and 3d1c38a76a359bb1580239cfb7cab75d393f26a5; along the way, its corresponding dropdown menu icon was removed even though it is actually still referenced. This simply reintroduces the icon, which in particular fixes https://github.com/EFForg/https-everywhere/issues/4082.


fuglede:
Looks like #2618 is also relevant here.

jsha:
LGTM, thanks! We do plan to re-remove 'moot' as a concept eventually, but good to fix this in the meantime.

fuglede:
I think you might have unintentionally closed this one instead of merging it? :)

jsha:
Whoops, sorry I missed this! Can you merge master into your branch so tests will pass?

fuglede:
No worries; updated.

PR #4355, opened 2016-03-13
[RFC-Editor] Update ruleset

Roman3349:


J0WI:
Could you also check ftp.rfc-editor.org?

Roman3349:
Added ftp.rfc-editor.org.

Issue #3934, opened 2016-01-19
MNGinteractive.com rule breaks stylesheet


terrorist96:
http://www.dailycamera.com/

Should this rule be set to default=off or can the issue be worked around?

terrorist96:
It also breaks the stylesheet on this site:
http://www.contracostatimes.com/

terrorist96:
This is still broken in Version: 2016.2.18

terrorist96:
Another example of a site broken due to this rule:
http://www.denverpost.com/

**Should this rule be set to default=off or can the issue be worked around?**

terrorist96:
www.mercurynews.com (should be fixed as well when new version of https everywhere is released)

PR #4362, opened 2016-03-13
Update MNGinteractive.com.xml

terrorist96:
Fixes https://github.com/EFForg/https-everywhere/issues/3934

### DIFF
--- a/src/chrome/content/rules/MNGinteractive.com.xml
+++ b/src/chrome/content/rules/MNGinteractive.com.xml
@@ -9,7 +9,7 @@
 		- extras	(redirects to http.../$, akamai)
 
 -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/MNGinteractive.com.xml
```diff
@@ -9,12 +9,12 @@
 		- extras	(redirects to http.../$, akamai)
 
 -->
-<ruleset name="MNGinteractive.com">
+<ruleset name="MNGinteractive.com" default_off="breaks layout">
```
J0WI: The issue is, that the host times out. So could you use `default_off="timeout"`?

Issue #3527, opened 2015-12-19
Pbsrc.com STILL breaks photobucket functionality


terrorist96:
This was reported as an issue in the 2015.8.13 version and is still an outstanding issue in 2015.12.16.

Pbsrc.com breaks flash copy action in the Email & IM, Direct, HTML, IMG fields:
http://s721.photobucket.com/user/terrorist96/media/ghgdsfghfd_zps0378f4b9.png.html

terrorist96:
This is still broken in Version: 2016.2.18

terrorist96:
Still broken in Version: 2016.3.23

terrorist96:
Still broken in Version: 2016.5.10

terrorist96:
Still broken in Version: 2016.5.10 :sob:

J0WI:
I would try the following fix:
```
<exclusion pattern="^http://pic2\.pbsrc\.com/flash/ZeroClipboardFV3\.swf">
```
This also requires to add a test url:
```
<test url="http://pic2.pbsrc.com/flash/ZeroClipboardFV3.swf">
```
Furthermore, the issue should be documented in the ruleset.

J0WI:
Reopen, see https://github.com/EFForg/https-everywhere/pull/4796#issuecomment-225250030

terrorist96:
Looks to be fixed now.

terrorist96:
Actually never mind. Still broken. See:
http://media.photobucket.com/user/ThomasLovesJeremy/media/tumblr_m2lvj8gtln1qhzv6oo1_500.gif.html?filters[term]=animated%20gifs&filters[primary]=images&filters[featured_media]=1748&filters[secondary]=videos&sort=1&o=2

terrorist96:
Can't reproduce anymore.

PR #4364, opened 2016-03-13
Update Pbsrc.com.xml

terrorist96:
Fixes https://github.com/EFForg/https-everywhere/issues/3527

### DIFF
--- a/src/chrome/content/rules/Ars-Technica.xml
+++ b/src/chrome/content/rules/Ars-Technica.xml
@@ -70,7 +70,7 @@
 	<target host="ars.to" />
 	<target host="arstechnica.com" />
 	<target host="*.arstechnica.com" />
-		<exclusion pattern="^http://(?:www\.)?arstechnica\.com/(?!ads/|civis/ucp\.php|favicon\.ico|public-shared/|services/incr\.php|subscriptions(?:$|[?/])|wp-content/)" />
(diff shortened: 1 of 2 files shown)


terrorist96:
@J0WI I have no idea what I'm doing. It got pulled over here. I give up.

src/chrome/content/rules/Ars-Technica.xml
```diff
@@ -100,5 +100,8 @@
 
 	<rule from="^http://(?:(?:media|origin)\.arstechnica\.com|(?:cdn|static)\.arstechnica\.net)/"
 		to="https://cdn.arstechnica.net/" />
+		
+	<rule from="^http://arstechnica.com/civis"
+		to="https://arstechnica.com/civis"
```
J0WI: Missing `/>`

terrorist96: Thanks. Fixed.

terrorist96:
Would be great if this could be merged.

J0WI:
This still doesn't pass our tests because of lacking test urls.

terrorist96:
I'm not a pro. I don't know where to find test URLs. Can you please help?

J0WI:
This contains some unrelated changes from Ars-Technica.xml.
I think without them you are just disabling a rule which doesn't require any tests.

Apart from that it would be great if you could just make a specific exclusion that we can still secure other sites that are not affected.

terrorist96:
Idk how to make the exclusion specific. Can you please type out what I need to put in the commit? I made a new PR
https://github.com/EFForg/https-everywhere/pull/4522

PR #4365, opened 2016-03-13
[Skat.dk] Add new subdomains and comments on broken ones

fuglede:


J0WI:
WFM:
inddrivelse.skat.dk
motorregister.skat.dk
export.skat.dk
extranet.skat.dk
pdcs.skat.dk

Refused:
(www.)tarif.skat.dk

fuglede:
Thanks @J0WI, updated it a bit. Do you have any non-trivial test URLs for demonstrating the purpose of export.skat.dk and inddrivelse.skat.dk?

And by the way, do you have a script for finding subdomains? The one I'm using (which is based on Google searches) seems to be a bit inferior to whatever you are doing.

J0WI:
Just use `site:inddrivelse.skat.dk` for search a valid test url.

Sadly Google limits the ability of scripting the searches, so I still search them manually.

fuglede:
Updated again. It must be possible to bypass whatever checks Google might have come up with. (Or find a search engine with no such restrictions.)

PR #4368, opened 2016-03-14
created rule for dji.com

schmiedecker:


fuglede:
Thanks for the PR. It looks like your git submodules are out of sync with those in the master branch; try running a <code>git submodule update</code>.

schmiedecker:
'done, has not changed anything? Could you explain the issue a bit? Thx ...

fuglede:
Sure thing: The HTTPS Everywhere repository hooks into a few other git repositories (for instance to get translations and tests), called [submodules](https://git-scm.com/docs/git-submodule), and the way they are kept synchronized is by having the working (main) repository keep track of the specific commit hashes of the subrepositories. Now, in your case, it seems that your local copy of the repository is pointing at the wrong versions of the subrepositories. This becomes evident when you look at the [files changed for this PR](https://github.com/EFForg/https-everywhere/pull/4368/files): if we just merged it, those two subrepositories would end up in incorrect versions.

J0WI:
This needs to be rebased.

jeremyn:
@schmiedecker Are you still interested in working on this?

schmiedecker:
Sure, but I fear my git-foo powers have not improved.

Would you be so kind to add the rules & close this pull request?

jeremyn:
@schmiedecker We're currently discussing if and when reviewers should directly modify pull requests, see https://github.com/EFForg/https-everywhere/issues/7242 . But that discussion is still open.

In the meantime, I'm not 100% sure on what triggers it, but when it happens to me, I can sync up (that is, fix) the subproject hashes by rerunning the `install-dev-dependencies.sh` install process. So I recommend resetting back to the current master branch, rerun `install-dev-dependencies.sh`, then add your ruleset. Hopefully the git diff will then just have the new ruleset without the subproject hash changes.

schmiedecker:
retry with new pull-request, #7358 and no modifications to any other files but the rule file for dji

Issue #4118, opened 2016-02-09
HTTP Nowhere prevents access to onion services


willscott:
This seems like an instance of cognitive dissonance between Tor Browser Bundle and HTTPS Everywhere. When "Block all HTTP Requests" is checked, the extension will block requests with the HTTP protocol, as might be hoped.  However, Tor onion services including the default 3g2upl4pq6kufc4m.onion duck duck go search engine included in Tor Browser Bundle, are generally accessed using HTTP due to the difficulty of acquiring SSL certificates for those addresses. While they use HTTP, they still receive end-to-end encryption in transport, and shouldn't necessarily be blocked blocked by an 'HTTP Nowhere' policy.

PR #4370, opened 2016-03-14
Allow requests to hidden services with HTTP Nowhere

fuglede:
This excludes .onion hostnames from the main HTTP Nowhere request block. It also updates the HTTP Nowhere label to reflect the change (but perhaps a better wording exists?).

The change is tested in Tor Browser 5.5.3 in which it does what is intended.

This fixes https://github.com/EFForg/https-everywhere/issues/4118

Note: Since this changes the name of a i18n string, a pull of the PR needs to be synced with the translations submodule.

### DIFF
--- a/src/chrome/locale/en/https-everywhere.dtd
+++ b/src/chrome/locale/en/https-everywhere.dtd
@@ -19,7 +19,7 @@
 <!ENTITY https-everywhere.menu.observatory "SSL Observatory Preferences">
 <!ENTITY https-everywhere.menu.globalEnable "Enable HTTPS Everywhere">
 <!ENTITY https-everywhere.menu.globalDisable "Disable HTTPS Everywhere">
-<!ENTITY https-everywhere.menu.blockHttpRequests "Block all HTTP requests">
+<!ENTITY https-everywhere.menu.blockHttpRequests "Block all unencrypted requests">
 <!ENTITY https-everywhere.menu.showCounter "Show Counter">
 <!ENTITY https-everywhere.menu.viewAllRules "View All Rules">
 
--- a/src/chrome/locale/en/https-everywhere.dtd
+++ b/src/chrome/locale/en/https-everywhere.dtd
@@ -19,7 +19,7 @@
 <!ENTITY https-everywhere.menu.observatory "SSL Observatory Preferences">
 <!ENTITY https-everywhere.menu.globalEnable "Enable HTTPS Everywhere">
 <!ENTITY https-everywhere.menu.globalDisable "Disable HTTPS Everywhere">
-<!ENTITY https-everywhere.menu.blockHttpRequests "Block all unencrypted requests">
+<!ENTITY https-everywhere.menu.blockUnencryptedRequests "Block all unencrypted requests">
 <!ENTITY https-everywhere.menu.showCounter "Show Counter">
 <!ENTITY https-everywhere.menu.viewAllRules "View All Rules">
 
--- a/src/chrome/content/code/HTTPS.js
+++ b/src/chrome/content/code/HTTPS.js
@@ -43,8 +43,8 @@ const HTTPS = {
     var isSTS = securityService.isSecureURI(
         CI.nsISiteSecurityService.HEADER_HSTS, channel.URI, 0);
     if (blob === null) {
-      // Abort insecure requests if HTTP Nowhere is on
-      if (httpNowhereEnabled && channel.URI.schemeIs("http") && !isSTS) {
+      // Abort insecure non-onion requests if HTTP Nowhere is on
+      if (httpNowhereEnabled && channel.URI.schemeIs("http") && !isSTS && !/\.onion$/.test(channel.URI.host)) {
         IOUtil.abort(channel);
       }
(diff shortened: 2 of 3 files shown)


src/chrome/locale/en/https-everywhere.dtd
```diff
@@ -19,7 +19,7 @@
 <!ENTITY https-everywhere.menu.observatory "SSL Observatory Preferences">
 <!ENTITY https-everywhere.menu.globalEnable "Enable HTTPS Everywhere">
 <!ENTITY https-everywhere.menu.globalDisable "Disable HTTPS Everywhere">
-<!ENTITY https-everywhere.menu.blockHttpRequests "Block all HTTP requests">
+<!ENTITY https-everywhere.menu.blockHttpRequests "Block all unencrypted requests">
```
jsha: I think if you want a new translation here you're going to need to create a new item.

fuglede: Hm, yeah, that would be a lot nicer.

Hainish:
I understand the motivation, but since `.onion` services [can now be signed by CAs](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.3.4.pdf), I think there is an understandable use case for users to *not* want to access `.onion` sites without a TLS certificate accompanying it.  It has been argued that the way `.onion` name resolution works substitutes the security assurances that TLS provides.  I would argue that this is not strongly guaranteed.  Currently, `.onion` addresses are generated using the weak `SHA-1` algorithm, providing only 80 bits of security.  One could generate a certificate that collides with a `.onion` address with non-negligible possibility.  The equivalent level of security in the x.509 chain gives [negative visual indicators in the latest Chrome](https://sha1-2017.badssl.com/), and support for these certificates will soon be deprecated.

These concerns are being [addressed by Tor](https://gitweb.torproject.org/torspec.git/tree/proposals/proposal-status.txt#n260), but the improved hidden service crypto hasn't been built out yet.

fuglede:
> One could generate a certificate that collides with a .onion address with non-negligible possibility.

This is a little bold-sounding to me: certainly, consensus these days is towards SHA-1 deprecation, but to my knowledge, no actual collisions have been described yet.

Now, correct me if I'm wrong, but recalling the CA/B [conditions for issuance](https://cabforum.org/wp-content/uploads/EV-V1_5_9.pdf#53) of certificates for onion services, what would keep our SHA-1 colliding attacker from obtaining a signature for their duped service on the same grounds as the legit one? Wouldn't taking that as a premise imply that we should also restrict access to TLS-secured onion services?

Hainish:
> Now, correct me if I'm wrong, but recalling the CA/B conditions for issuance of certificates for onion services, what would keep our SHA-1 colliding attacker from obtaining a signature for their duped service on the same grounds as the legit one?

Currenly `.onion` addresses are required to obtain Extended Validation certificates in order to be signed by a CA.  Since EV certs require validation of location and organization, it's not as easy as just generating a collision.

fuglede:
Afaik, the current conditions are exactly those linked above.

Hainish:
[Ballot 144 of the CA/B forum](https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/), which passed in February last year, states the requirement:

> A CA may issue an EV Certificate with .onion in the right-most label of the Domain Name provided that issuance complies with the requirements set forth in this Appendix

and

>  Effective May 1, 2015, each CA SHALL revoke all unexpired Certificates with an Internal Name using onion as the right-most label in an entry in the subjectAltName Extension or commonName field unless such Certificate was issued in accordance with Appendix F of the EV Guidelines.

fuglede:
Indeed; afaict this does not take into account the case of two legitimate claims to the same hash.

Also, if I recall correctly (but I didn't go source hunting), directory servers disallow duplicate hashes with differing public keys, so that it would never actually occur in practice (is that right?).

Hainish:
I'm okay with merging this, since it does give *some* cryptographic guarantees.  However we should keep in mind moving forward that these guarantees are using weaker ciphers than almost all signatures being used by CAs today.  In the future, as `SHA-1` attacks become easier and `SHA-1` is deprecated in the browser, we hope that Next-Generation Hidden Services will roll out and provide better security, but there's no guarantee of this.

fuglede:
But again, if directory servers explicitly disallow duplicate hash entries if public keys disagree, why is collision an issue?

Hainish:
Were you able to find whether this is true?  How is the canonical service established - is it the first that is advertised?

fuglede:
I couldn't find any proper documentation, unfortunately no, and I came across sources claiming that dupes would be rejected, and others claiming that the most recent entry wins. Someone should cook up a proper preimage attack and test it. ;)

Coming back to the other point though, assuming that one can take over the hostname, how would TLS certificates help when the host is taken over? As far as I can tell from the guideline, there is no protection from that happening at all (and indeed that appears to have been one of the concerns during that ballot).

In any case, something else we could do would be to have this be disjoint from HTTP Nowhere, offered as a separate option in Tor Browser.

PR #4371, opened 2016-03-15
Updated Meetup.xml

J0WI:


funkydude:
> `hep.meetup.com`

I think you mean `help`

> `<rule from="^http://meetup\.com/" to="https://www.meetup.com/" />`

You should also forward https requests so `from="^https?://meetup\.com/"`

Same goes for `from="^http://(photos|static)\d\.meetupstatic\.com/"`

J0WI:
Thanks for exposing the typo.

> You should also forward https requests so from="^https?://meetup\.com/"

Disagree, see discussion in #4303. There is no use case, except the page uses weird redirects or a domain is deprecated, but still linked on many third party sites.

funkydude:
Oh please, @fuglede contradicts himself on a daily basis. One day he says he doesn't see the point of cookie rules, the next day he is using them himself. He himself has redirected https rules previously.

I will personally opt to continue this "extra maintenance" (nonsense, he doesn't even provide any examples of any negatives) to give the user a complete experience. There is no way of knowing who might link said domain or type it manually.

J0WI:
Would be great to have this merged @fuglede @jer

fuglede:
Looks good to me; I could not find any issues. Would you be able to provide some examples/test URLs for the `static\d`/`photos\d` parts of the ruleset? It was not clear to me why the counter stops at 4.

J0WI:
@fuglede I've updated the regex and the test urls. There are only numbers 1 to 4.

fuglede:
Perfect, thanks! I'm a bit confused about the different roles that `photos` and `static` play on their site, but that's what it is.

PR #4379, opened 2016-03-16
[FitBit] Enable and update ruleset

funkydude:
[FitBit] Enable and update ruleset

@J0WI lay it on me :)

### DIFF
--- a/src/chrome/content/rules/FitBit.xml
+++ b/src/chrome/content/rules/FitBit.xml
@@ -1,6 +1,8 @@
 <!--
 	Forwards to HTTP (March 2016):
 		- staticcs.fitbit.com
+		- static1.fitbit.com
+		- static0.fitbit.com
 -->
 <ruleset name="FitBit">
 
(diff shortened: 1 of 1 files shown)


funkydude:
Out of curiosity. I nearly added staticcs.fitbit.com as it responds, but forwards to HTTP, which I assume would create a loop. Would the script pick up on this?

J0WI:
Do you have an example for the cache redirect?

The static domains are woriking for me, up to static9. As well as the following:
assets.fitbit.com
investor.fitbit.com
status.fitbit.com
strava.fitbit.com
contact.fitbit.com
ifttt.fitbit.com

Broken ones:
(www.)lp.fitbit.com
support.fitbit.com

funkydude:
> Do you have an example for the cache redirect?

What do you mean? This isn't something I added, but something I left in as it appears to work.

>The static domains are working for me, up to static9

I just tried again, they definitely do not work. They forward to an HTTP page. If you then try convert _that_ page to HTTPS, you'll suffer mixed content problems.

J0WI:
>> Do you have an example for the cache redirect?

> What do you mean? This isn't something I added, but something I left in as it appears to work. What example do you seek?

I found one and it works. CDN tend to use a different path than the original source.


>>The static domains are working for me, up to static9

>I just tried again, they definitely do not work. They forward to an HTTP page. If you then try convert that page to HTTPS, you'll suffer mixed content problems.

Yepp, there is a mixed content problem on $, but all files that are served from there are available over https.

funkydude:
@J0WI done, but I think I want to remove the cache rule, I'm not comfortable with it.

funkydude:
Remove the cache rule. Seems like a future source of potential problems, and probably isn't used on the actual website anyway.

PR #4311, opened 2016-03-09
Spring cleaning 2016

RReverser:
First and foremost – apologies for the amount of changes here. I can understand if you won't want to merge that many at once, but I believe that such clean up is long awaited and would benefit all of us.

---

This replaces all the rulesets that have only rules in the form of
```xml
<rule from="^http://(A\.|B\.|C\.|...)?domain\.com/" to="https://$1domain.com/" />
```
to match the [Ruleset Style Guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) by generating list of explicit target domains + trivial
```xml
<rule from="^http:" to="https:" />
```
rule which is also better optimizable by replacer engine.

Those changes were auto-generated and re-checked to be at least valid XML.

During the generation, following rule mistakes were automatically found:
```javascript
[ 'www.ao2.it'; ] 'ao2.it' '^http://(www\\.)?ao2\\.it/'
[ 'cityemail.com', 'www.cityemail.com'; ] 'citymail.com' '^http://(www\\.)?citymail\\.com/'
[ 'cityemail.com', 'www.cityemail.com'; ] 'www.citymail.com'; '^http://(www\\.)?citymail\\.com/'
[ 'claws-mail.org', 'www.claws-mail.org'; ] 'git.claws-mail.org' '^http://(git\\.|www\\.)?claws-mail\\.org/'
[ '*.hohndel.org' ] 'hohndel.org' '^http://(subsurface\\.|trac\\.)?hohndel\\.org/'
[ 'consortiumlibrary.org', 'www.consortiumlibrary.org'; ] 'consortiumlibrary.com' '^http://(www\\.)?consortiumlibrary\\.com/'
[ 'consortiumlibrary.org', 'www.consortiumlibrary.org'; ] 'www.consortiumlibrary.com'; '^http://(www\\.)?consortiumlibrary\\.com/'
[ 'eltatar.com', '*.eltatar.com' ] 'eltartar.com' '^http://(www\\.)?eltartar\\.com/'
[ 'eltatar.com', '*.eltatar.com' ] 'www.eltartar.com'; '^http://(www\\.)?eltartar\\.com/'
[ 'gleam.io' ] 'js.gleam.io' '^http://(js\\.)?gleam\\.io/'
[ 'marshut.net', 'www.marshut.net'; ] 'marsnet.net' '^http://(www\\.)?marsnet\\.net/'
[...]

PR #4381, opened 2016-03-16
Trivialize rules

Hainish:
This PR sources the script by @RReverser from #4311, adding it to utils/ and running it on the ruleset.

Since #4311 contained so many ruleset changes, we decided it would be best to audit the script to ensure proper behavior, and re-run it against the latest ruleset.  I have done so, and noticed no unusual behavior after manually checking dozens of ruleset changes.

### DIFF
--- a/utils/duplicate-whitelist.txt
+++ b/utils/duplicate-whitelist.txt
@@ -193,6 +193,7 @@ nandos.com
 mars.jpl.nasa.gov
 mars.nasa.gov
 nelonen.fi
+www.nelonen.fi
 *.netapp.com
 netnow.co
 www.netnow.co
@@ -284,6 +285,7 @@ vk.com
 vkontakte.ru
 vlc-addons.org
 vpncompare.co.uk
+www.vpncompare.co.uk
 help.walmart.com
 faq.web4all.fr
 members.webs.com
--- a/utils/trivialize-rules/.gitignore
+++ b/utils/trivialize-rules/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/utils/trivialize-rules/package.json
+++ b/utils/trivialize-rules/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "trivialize-rules",
+  "version": "1.0.0",
+  "main": "trivialize-rules.js",
+  "dependencies": {
+    "highland": "^2.7.1",
+    "progress": "^1.1.8",
+    "verbal-expressions": "^0.2.1",
+    "xml2js": "^0.4.16"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "start": "node --harmony_destructuring trivialize-rules.js",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "Ingvar Stepanyan",
+  "description": ""
+}
--- a/utils/trivialize-rules/trivialize-rules.js
+++ b/utils/trivialize-rules/trivialize-rules.js
@@ -1,0 +1,172 @@
+"use strict";
+const _ = require('highland');
+const fs = require('fs');
+const readDir = _.wrapCallback(fs.readdir);
+const readFile = _.wrapCallback(fs.readFile);
+const writeFile = _.wrapCallback(fs.writeFile);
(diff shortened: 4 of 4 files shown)


Hainish:
There are numerous rules that have duplicate targets after running the script, I'll have to clean this up before merging.

Hainish:
I've solved most of the validation problems by looking for duplicates only within the scope of a specific platform.  There were two remaining hosts after this change, which I've added to the duplicate-whitelist.txt

Hainish:
@jsha can you review this PR (since it includes more than just the changes by @RReverser) and let me know if it looks good?  If it does, I can run the script over the ruleset once more and it won't cause failures in the test script.

RReverser:
> Since #4311 contained so many ruleset changes, we decided it would be best to audit the script to ensure proper behavior, and re-run it against the latest ruleset. I have done so, and noticed no unusual behavior after manually checking dozens of ruleset changes.

Thanks! Although, as I said, I could do that myself as well. Would be nice to have that contribution bump on Github as an additional perk :)

> There are numerous rules that have duplicate targets after running the script, I'll have to clean this up before merging.

This is interesting, I thought this was handled by script.

utils/make-sqlite.py
```diff
@@ -37,7 +37,12 @@
 c.execute('''DROP TABLE IF EXISTS targets''')
 c.execute('''CREATE TABLE targets
              (host TEXT,
+              platform_id INTEGER,
               ruleset_id INTEGER)''')
+c.execute('''DROP TABLE IF EXISTS platforms''')
+c.execute('''CREATE TABLE platforms
```
jsha: It's weird to add new code that depends on SQLite when we are planning to remove SQLite altogether. Can you do this in-memory in trivial-validate.py instead?

jsha:
> Would be nice to have that contribution bump on Github profile as an additional perk

That should be fine - the main thing is that, given a pull request too large to review, we need some way to verify it. One way we could do it: Once this change is landed, you run the script and put up a PR. Bill runs the script is his repo too, pulls down your branch, and verifies no diffs. Assuming that's all good, Bill can consider the PR reviewed and merge it, which will give you credit for the commit.

Hainish:
@RReverser perhaps an easier way to do it is to use `--author` in git, and attribute the ruleset diff to you in a separate commit?
cc @jsha

utils/make-sqlite.py
Hainish: I was afraid you might say that :)
I figured I would just refactor this when the time came to get rid of `sqlite`, but I agree it adds complexity to something we're depricating. Refactoring it now.

Hainish:
@jsha ready for review again.  I've removed all sqlite references from `trivial-validator.sh` but I wasn't sure if the `sqlite` db was used elsewhere, so I kept it in `utils/make-sqlite.sh`

RReverser:
@Hainish Dunno if easier, never used that command so not sure how it exactly works. I assume you didn't change JS itself? If so, given that script is already reviewed, wouldn't be easier (and with the same result) if I would just update my PR after you merge your own changes separately?

Hainish:
This is PR is now dependent on https://github.com/EFForg/https-everywhere/pull/4387, which I've separated out for clarity.

Hainish:
@jsha rebased against master!

jsha:
Mostly LGTM. The one comment I had: We should check for target hosts that don't get rewritten by any rule, and therefore get deleted. Mostly those will be errors in the rules, but there's a small possibility that those hosts may be affected by a securecookie tag. Talked to @Hainish in person and he's checking.

RReverser:
Good point. Mostly I checked for the opposite as that was easy to do in scope of this PR, but that one would be easy as well.

Hainish:
@jsha I didn't find any target domains of this type.  I double-checked by manually adding `foo.com` as a target domain to one of the overwritten files, and that's the only domain that is returned.  I wonder if someone manually went through and removed targets not affected by rewrite rules at some point?

Hainish:
@RReverser feel free to update your PR with the results from the script, and I can compare to my own run & merge.

RReverser:
@Hainish Awesome, will do. Thanks.

PR #4384, opened 2016-03-16
Update Qt.io.xml

ekaris:
Please pull only Qt.io.xml commit


J0WI:
You need to use a separate branch for this commit.

PR #4385, opened 2016-03-16
Update RUB rules: Add Flexnow

PotcFdk:


J0WI:
You may be also interested in adding more subdomains. We have a guide [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

Since university sites are quite complex and it's very hard to get a 100% coverage, I'll merge this single change.

PR #4387, opened 2016-03-17
Remove sqlite

Hainish:


utils/trivial-validate.py
```diff
@@ -42,6 +40,44 @@ def fail(s):
...
+        if filename.endswith('/00README') or filename.endswith('/make-trivial-rule') or filename.endswith('/default.rulesets'):
+            continue
+        try:
+            tree = etree.parse(filename, xml_parser)
+        except Exception as oops:
+            print("%s failed XML validity: %s\n" % (filename, oops))
+            sys.exit(1)
+
+        try:
+            platform = xpath_ruleset(tree)[0].attrib["platform"]
```
jsha: You can select this more directly with `/ruleset/@pattern`. And double-check what the results are when not present: you might be able to get away without the try/except.

utils/trivial-validate.py
```diff
@@ -42,6 +40,44 @@ def fail(s):
 with open(thispath + '/duplicate-whitelist.txt') as duplicate_fh:
     duplicate_allowed_list = [x.rstrip('\n') for x in duplicate_fh.readlines()]
 
+filenames = glob.glob('src/chrome/content/rules/*')
+
+def generate_hosts_count():
+    # Generate a count of the number of times a host appears, and on which
+    # platforms
+    hosts = {}
```
jsha: I think you can make this whole function a lot simpler with:

```
c = collections.Counter()
c.update([(hostname, platform)])
for tuple, count in c.most_common():
  if count == 1:
    break
  fail(...)
```

Also, the counting in here should happen within the main loop through the rulesets rather than as a separate step.

Hainish:
Ready for review again @jsha

utils/trivial-validate.py
```diff
@@ -160,19 +158,25 @@ def nomes_all(where=sys.argv[1:]):
...
-xpath_ruleset_file = etree.XPath("/ruleset/@f")
+xpath_ruleset_platform = etree.XPath("/ruleset/@platform")
 xpath_host = etree.XPath("/ruleset/target/@host")
 xpath_from = etree.XPath("/ruleset/rule/@from")
 xpath_to = etree.XPath("/ruleset/rule/@to")
 
-conn = sqlite3.connect(args.db)
-c = conn.cursor()
-for row in c.execute('''SELECT contents from rulesets'''):
+c = Counter()
```
jsha: Please name the counter more descriptively.

utils/trivial-validate.py
```diff
@@ -160,19 +158,25 @@ def nomes_all(where=sys.argv[1:]):
...
+for filename in filenames:
+
```
jsha: Remove the blank space after the beginning of the for loop.

utils/trivial-validate.py
```diff
@@ -180,22 +184,30 @@ def nomes_all(where=sys.argv[1:]):
...
-    rf = xpath_ruleset_file(tree)[0]
     from_attrib = xpath_from(tree)
     to = xpath_to(tree)
     for test in tests:
         if not test(tree, rn, from_attrib=from_attrib, to=to):
             failure = 1
-            fail("%s failed test: %s" % (rf, test.__doc__))
+            fail("%s failed test: %s" % (filename, test.__doc__))
+
+    try:
```
jsha: I'd recommend instead avoiding the exception:

```
platforms = xpath_ruleset_platform(tree)
if len(platforms) == 0:
  platform = ''
else:
  platform = platforms[0]
```

PR #4392, opened 2016-03-17
Rename webchat.oftc.net to OFTC.net and add new hosts

mnordhoff:
* Rename **webchat.oftc.net** ruleset to **OFTC.net**.
* Add `oftc.net` and `www.oftc.net`.
* `https://oftc.net/` has a mismatched certificate for `www.oftc.net`, but serves an HTTP redirect to `www.oftc.net` anyway.

### DIFF
--- a/src/chrome/content/rules/OFTC.net.xml
+++ b/src/chrome/content/rules/OFTC.net.xml
@@ -1,9 +1,11 @@
 <!--
 	Problematic hosts:
 
-		- oftc.net *
+		- oftc.net ¹
+		- git.oftc.net ²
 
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/OFTC.net.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Problematic hosts:
+
+		- oftc.net *
```
J0WI: git.oftc.net does also serve a mismatched cert

mnordhoff: True. I think they mostly use GitHub now, but i've added a comment.

PR #4394, opened 2016-03-17
Fix rewriter.js, which went out of date.

jsha:
Also update it to always rewrite to HTTPS, rather than protocol-relative.

cc @Hainish for review.


Hainish:
lgtm, works for me

PR #4397, opened 2016-03-18
Include certain second-level booking.com subdomains

CoreDumpling:
Some (but not all) of these subdomains should be moved
to HTTPS such that the <securecookie> rule does not
block them from setting cookies at the first level.


J0WI:
Can you provide some test domains?

J0WI:
@CoreDumpling are you still working on this?

J0WI:
I just found `blog.suite.booking.com` which does not support https, so I'm closing this.

PR #4400, opened 2016-03-18
[Spring cleaning] Trivialize rules where possible.

RReverser:
**This is reopen of #4311, see the original PR for details and comments.**

---

This replaces all the rulesets that have only rules in the form of

<rule from="^http://(A\.|B\.|C\.|...)?domain\.com/" to="https://$1domain.com/" />

to match the Ruleset Style Guide by generating list of explicit target domains + trivial

rule which is also better optimizable by replacer engine.

Those changes were auto-generated and re-checked to be at least valid XML.


Hainish:
@RReverser, I'm getting the following diff when I run the script myself:

```
iff --git a/src/chrome/content/rules/Blockstream.com-falsemixed.xml b/src/chrome/content/rules/Blockstream.com-falsemixed.xml
deleted file mode 100644
index 19442ca..0000000
--- a/src/chrome/content/rules/Blockstream.com-falsemixed.xml
+++ /dev/null
@@ -1,18 +0,0 @@
-<!--
-Fetch error: http://blockstream.com/ => https://blockstream.com/: (28, 'Connection timed out after 10000 milliseconds')
-       For rules not causing false/broken MCB, see Blockstream.com.xml.
-
--->
-<ruleset name="Blockstream.com (false MCB)" platform="mixedcontent" default_off='failed ruleset test'>
-
-       <target host="blockstream.com" />
-       <target host="www.blockstream.com" />
-
-
-       <securecookie host="^\.blockstream\.com$" name=".+" />
-
-
-       <rule from="^http:" to="https:" />
-
-</ruleset>
```

RReverser:
That's weird, are you running on latest master?

Hainish:
Yes, I think your PR is missing 5583f958585391a048eb9d98be0a3c0dea8e51df, that's why I'm getting the diff... did you rebase from the latest master?

RReverser:
Yes, you can see in https://github.com/RReverser/https-everywhere/commits/spring-cleaning-2016 that my commit sits on top of current master. And `src/chrome/content/rules/Blockstream.com-falsemixed.xml` is actually currently not in the repo https://github.com/EFForg/https-everywhere/find/master, perhaps you have unclean local state?

RReverser:
Wait... I guess, found it, let me try smth.

RReverser:
Updated.

Hainish:
Great, looks good to me!

Hainish:
Hmm @RReverser I'm getting a lot of coverage errors in travis.  A lot of these are for exclusion patterns.

RReverser:
@Hainish Well, it's because script can't autogenerate new tests, while files are kinda "modified". I believe we need to whitelist them in https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist.txt as we did with @fuglede for https://github.com/EFForg/https-everywhere/pull/3017.

Hainish:
@jsha do you think this is the right approach?

Hainish:
I suspect we should slowly work through these and make sure the tests are updated for each one.  Probably a lot of work, but we may forget about them if we just whitelist.

RReverser:
@Hainish Hmm, do you really want to go through 2442 changed files just because they were auto-updated? :/

Hainish:
#4404 attempts to fix tests

@RReverser the coverage tests mark problems with much less than all 2442 of the changed files.

RReverser:
@Hainish @jsha I've updated the script to detect even more cases that can be trivialized, would you be open to accepting another round of cleanups? (with the same caveat that I'm not adding new tests for all the updated rulesets, as it's a batch operation that touched a thousand of files)

PR #4401, opened 2016-03-18
[WEDOS] Added ruleset

Roman3349:


fuglede:
This creates a redirect loop between http://www.wedos.com and https://www.wedos.com.

Roman3349:
Fixed

fuglede:
Great, thanks.

PR #4404, opened 2016-03-18
Add rules to ruleset coverage whitelist in order to fix tests (see #4400)

Hainish:
Ready for review @jsha


Hainish:
I'll have to sort these in a separate commit.  This should work for now, though

PR #4407, opened 2016-03-19
Create wifiwx.com.xml

renyouguo:


J0WI:
I also found html.wifiwx.com, event.wifiwx.com and www3.wifiwx.com.
Can you check if it's safe to redirect them or adding a comment why it's not.
There are some mixed content issues, but I can't say if they are relevant for the content of the site.

PR #4408, opened 2016-03-19
[Mall] Remove HSTS preloaded ruleset

Roman3349:


J0WI:
`Prefer dashes over underscores in filenames. Dashes are easier to type.`
And we got four of them here. Could you fix that too?

Note that the holding also supports https:
nrholding.com
www.nrholding.com
zbozi.nrholding.com

If this is related, you can add them here. Otherwise just update the link in the comment.

J0WI:
Aah, just found them on HSTS preload list:

```
  { "mall.cz", true },
  { "mall.hr", true },
  { "mall.hu", true },
  { "mall.pl", true },
  { "mall.sk", true },
[...]
  { "mimovrste.com", true },
```

So they need no longer a rule at all. :)

Roman3349:
I removed HSTS preloaded ruleset. And I am going to add new ruleset for nrholding.com.

PR #4411, opened 2016-03-19
[VZP.cz] Added new targets

Roman3349:


J0WI:
I get timeouts on webcall and edeska.
Could you add comments for them?

Roman3349:
@J0WI Done

PR #4416, opened 2016-03-20
Updated rules for somethingawful.com

rufoa:
SA has now deployed SSL on most parts of the site so here are the updated rules. Many thanks!

### DIFF
--- a/src/chrome/content/rules/Something-Awful.xml
+++ b/src/chrome/content/rules/Something-Awful.xml
@@ -1,4 +1,4 @@
-<ruleset name="Something Awful (partial)">
+<ruleset name="Something Awful">
 
 	<target host="somethingawful.com" />
 	<target host="www.somethingawful.com" />
@@ -19,9 +19,6 @@
 	<securecookie host="^forums\.somethingawful\.com$" name=".+" />
 	<securecookie host="^archives\.somethingawful\.com$" name=".+" />
 
-
-	<rule from="^http://somethingawful\.com/"
-		to="https://www.somethingawful.com/" />
 
 	<rule from="^http:"
 		to="https:" />
--- a/src/chrome/content/rules/Something-Awful.xml
+++ b/src/chrome/content/rules/Something-Awful.xml
@@ -1,21 +1,21 @@
-<!--
-	Nonfunctional subdomains:
-
-		- (www.)	(shows secure)
-		- fi
-		- forums	(times out)
-		- i
-
--->
 <ruleset name="Something Awful (partial)">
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Something-Awful.xml
```diff
@@ -1,21 +1,29 @@
...
+	<target host="videos.somethingawful.com" />
 
 
 	<securecookie host="^secure\.somethingawful\.com$" name=".+" />
+	<securecookie host="^forums\.somethingawful\.com$" name=".+" />
+	<securecookie host="^archives\.somethingawful\.com$" name=".+" />
+
 
+	<rule from="^http://somethingawful\.com/"
+		to="https://www.somethingawful.com/" />
```
J0WI: This shouldn't be necessary.

src/chrome/content/rules/Something-Awful.xml
```diff
@@ -1,21 +1,29 @@
-<!--
-	Nonfunctional subdomains:
-
-		- (www.)	(shows secure)
-		- fi
-		- forums	(times out)
-		- i
-
--->
 <ruleset name="Something Awful (partial)">
```
J0WI: Is this rule still partial? I don't think so.

src/chrome/content/rules/Something-Awful.xml
rufoa: Unfortunately it is as the bare domain forces a downgrade to http whereas www. works with https

src/chrome/content/rules/Something-Awful.xml
rufoa: Not any more - please rename

src/chrome/content/rules/Something-Awful.xml
J0WI: This doesn't matter.
^ will be redirected to https by us, the server redirects to unsecured www then and we will redirect to www over https again.

src/chrome/content/rules/Something-Awful.xml
J0WI: You have to rename this yourself in your branch.

src/chrome/content/rules/Something-Awful.xml
rufoa: Ah right, got it

src/chrome/content/rules/Something-Awful.xml
rufoa: Will do asap (and the other change)

rufoa:
@J0WI rules now fixed. Many thanks

PR #4422, opened 2016-03-20
[Truetrophies] Add subdomain static

fuglede:


ekaris:
The test URL shouldn’t be in HTTPS?

fuglede:
@ekaris: The logic is that it takes the test URL (http://static.truetrophies.com/images/banner/tt/gamerpage.jpg), tries to secure it (https://static.truetrophies.com/images/banner/tt/gamerpage.jpg) and compares the results.

PR #4424, opened 2016-03-20
Delete old ruleset databases in 'make' and 'make clean'

mnordhoff:
* `make clean` now deletes the old `rulesets.unvalidated.sqlite` and `rulesets.sqlite` files
* `makecrx.sh` and `makexpi.sh` now delete the old `rulesets.sqlite` file and super-old `default.rulesets` file. (I'm pedantic.)
* (A `makecrx.sh` comment that referred to SQLite has been updated.)

I think it's important for `make clean` to continue to remove the old files, for people who forget to clean before pulling or switching between branches. (Ahem.)

I also think it's useful for `makecrx.sh` and `makexpi.sh` to do it, but that might be more controversial. (And maybe should have been a separate pull request.)

I don't know about you folks, but i build HTTPS Everywhere from a dirty directory, and i was unwittingly building double-size XPIs with `default.rulesets` for *months* after the switch to SQLite. And i just caught myself doing the same thing with an old `rulesets.sqlite` file after last month's move to JSON.

### DIFF
--- a/makecrx.sh
+++ b/makecrx.sh
@@ -20,7 +20,7 @@
 # releases signed by EFF :/.  We should find a more elegant arrangement.
 
 cd $(dirname $0)
-RULESETS_JSON="$PWD/pkg/rulesets.json"
+RULESETS_JSON=pkg/rulesets.json
 
 if [ -n "$1" ]; then
   BRANCH=`git branch | head -n 1 | cut -d \  -f 2-`
@@ -39,7 +39,7 @@ echo "Building chrome version" $VERSION
 [ -e pkg/crx ] && rm -rf pkg/crx
 
 # Clean up obsolete ruleset databases, just in case they still exist.
-rm -f "$PWD/src/chrome/content/rules/default.rulesets" "$PWD/src/defaults/rulesets.sqlite"
+rm -f src/chrome/content/rules/default.rulesets src/defaults/rulesets.sqlite
 
 # Only generate the ruleset database if any rulesets have changed. Tried
 # implementing this with make, but make is very slow with 15k+ input files.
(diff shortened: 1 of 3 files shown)


J0WI:
Hmm, let me ping @jsha or @Hainish with that.

makecrx.sh
```diff
@@ -38,7 +38,10 @@ echo "Building chrome version" $VERSION
 [ -d pkg ] || mkdir -p pkg
 [ -e pkg/crx ] && rm -rf pkg/crx
 
-# Only generate the sqlite database if any rulesets have changed. Tried
+# Clean up obsolete ruleset databases, just in case they still exist.
+rm -f "$PWD/src/chrome/content/rules/default.rulesets" "$PWD/src/defaults/rulesets.sqlite"
```
jsha: Just to be paranoid, I would leave out the -f (I don't think it's needed), and would use `.` instead of `$PWD`, just in case `$PWD` is not set for any weird reason, so it doesn't have an unexpected affect. Same below.

mnordhoff: @jsha I used `-f` to stop it from throwing an error if the files already don't exist. But i could switch it to `|| true` or something; what do you think?

Good point about `$PWD`. It's also used when `RULESETS_JSON` is set, want to change that too?

jsha: Sure, sounds good. Good point about the `-f`. I'm okay leaving that in - I was just being excessively paranoid.

mnordhoff:
Done. I think it's spooky that `-f` does multiple things, including one so innocuous, but that's how `rm` rolls. :X

jsha:
LGTM

PR #4429, opened 2016-03-21
Updated NeoOffice.org.xml

J0WI:


jeremyn:
Since we don't include userguide.neooffice.org, shouldn't this still be `(partial)`?

J0WI:
I tend to use the `(partial)` flag only, when a major part of the website is not secured.
In can be irritating to see the flag also on the main site (www), which is fully secured.

On userguide. this rule won't show up, so a user won't notice why this is `(partial)`.

jeremyn:
Thanks for the explanation.

PR #4435, opened 2016-03-21
fixed #4008

### LINKED ISSUE #4008: english-heritage.org.uk Forced HTTP Redirect
www.english-heritage.org.uk seems to have added JS to force redirect back to http-only mode.

J0WI:


fuglede:
I didn't look into it in detail, but it looks like parts of this one might still be securable, cf. e.g. https://www.english-heritage.org.uk/favicon.ico.

J0WI:
Yes, static resources should work since the redirect is over JS.

PR #4437, opened 2016-03-21
Add rule for mnultimate.org

ishults:


J0WI:
Please also add the www subdomain as a target host an a comment, that the the server doesn't serve a mismatched cert for the dns subdomain.

Our code style recommends to use tabs.

ishults:
Oh ok, sorry.  Should be fixed now.

J0WI:
There is still a comment missing for the mismatch on `dns.mnultimate.org`.
Rule lgtm.

ishults:
Shoot sorry, missed that point the first time.  I believe I've fixed it?  I looked over the README and other commits and think I did it right, but please let me know if I still missed something.

J0WI:
Great!

ishults:
Thanks for your help and patience!

Issue #4432, opened 2016-03-21
redirect loop sometimes not detected


diracdeltas:
Ex: http://www.apple.com/apple-events/march-2016/ sends HTTPS-E into an infinite redirect loop in both Chrome and Firefox (and Brave). at least in Chrome, looks like the channel ID keeps increasing so HTTPS-E doesn't increment the redirect counter which is keyed by ID. i haven't looked into what Apple is doing to implement this annoying type of redirect.

diracdeltas:
Just fixed this in Brave by adding some code to the redirect loop detection heuristic which checks if a URL returns 307 more than 5 times in 100ms. https://github.com/brave/browser-laptop/commit/9b945c5286e3d5672adfd62d6b5c09f03b305b2d#diff-a4a57e1c5be3df294e71a726d2e444cbR189

J0WI:
```
<script type="text/javascript">
if (true && window.location.protocol == "https:") {
window.location.href = "http:" + window.location.href.substring(window.location.protocol.length);
}
</script>
```

no comment

diracdeltas:
Ugh, I didn't realize the Apple event was today. We would have done some releases. :(

PR #4443, opened 2016-03-21
[Apple] Add exclusion

funkydude:
Creates an exclusion for /apple-events/ which is currently broken
(constant reload loop).

Partially resolves https://github.com/EFForg/https-everywhere/issues/4432

This is technically already too late for the Apple event which was today.

### DIFF
--- a/src/chrome/content/rules/Apple.xml
+++ b/src/chrome/content/rules/Apple.xml
@@ -524,7 +524,7 @@
 
 		<!--	Redirects to http:
 						-->
-		<exclusion pattern="http://www\.apple\.com/(?:\w\w/)?shop(?:$|[?/])" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Apple.xml
```diff
@@ -534,6 +534,10 @@
 			<test url="http://www.apple.com/shop/payment_plan" />
 			<test url="http://www.apple.com/uk/shop/buy-mac/macbook/space-grey-512gb" />
 
+		<exclusion pattern="^http://www\.apple\.com/apple\-events/" />
```
J0WI: The backslash shouldn't be necessary.

Please document that the redirect is JS based.

funkydude:
@J0WI done

J0WI:
Thanks!

PR #4451, opened 2016-03-22
podseed - Content Delivery Network

MilanKral:


J0WI:
Thanks for adding this rule!
Please use tabs to follow our code style. The rule itself LGTM.

PR #4459, opened 2016-03-23
[Imageforensic] New ruleset

rugk:


J0WI:
Note: There is no Flattr button during mixed content blocking, but the site works.

PR #4461, opened 2016-03-24
Fix trivial validate

Hainish:
Ready for review @cooperq @jsha


Hainish:
A little more context... before this change, the script `utils/validate.sh` when run from the project root would just exit successfully when running https://github.com/EFForg/https-everywhere/blob/master/utils/validate.sh#L16
This is because `trivial-validate.sh`, when run, looks for all the xml files in a certain directory which *doesn't exist* when being run from the project root.

Unfortunately, all our post-commit hooks and testing scripts run `trivial-validate` in this way.  So in fact it was never really testing anything at all.

Hainish:
This also fixes one file that slipped by our test suite due to the validation tests not being run properly.

PR #4462, opened 2016-03-24
Add HSTS-based rewrites to the Node rewriter.

jsha:
Also, bring all dependencies up-to-date and remove the hacky workaround for
explosive regex in URIjs (seems to be fixed upstream now).

cc @Hainish @cooperq

### DIFF
--- a/rewriter/hsts-transform.js
+++ b/rewriter/hsts-transform.js
@@ -1,0 +1,26 @@
+// A tool to transform transport_security_state_static.json into one big trivial
+// ruleset on stdout.
+// To get transport_security_state_static.json:
+// curl https://chromium.googlesource.com/chromium/src/+/master/net/http/transport_security_state_static.json?format=TEXT \
+//  | base64 -d > transport_security_state_static.json
+var fs = require('fs');
+
+if (process.argv.length != 3) {
+  console.error('Provide path to transport_security_state_static.json as only argument.');
+  return;
+}
+var contents = fs.readFileSync(process.argv[2], 'utf8');
+contents = contents.split('\n').filter(line => !/^\s*\/\//.test(line)).join('\n');
+var secStateStatic = JSON.parse(contents);
+console.log('<ruleset name="hsts">')
+secStateStatic.entries.forEach(function(entry) {
+  if (entry.mode != "force-https") {
+    return;
+  }
+  console.log('  <target host="' + entry.name + '" />');
+  if (entry.include_subdomains) {
+    console.log('  <target host="*.' + entry.name + '" />');
+  }
+});
+console.log('  <rule from="^http:" to="https:" />');
+console.log('</ruleset>');
--- a/rewriter/rewriter.js
+++ b/rewriter/rewriter.js
@@ -24,7 +24,7 @@ var path = require("path"),
 
     URI = require("URIjs");
 
-var ruleSets = null;
+var ruleSets = new rules.RuleSets({});
 
 /**
  * For a given directory, recursively edit all files in it that match a filename
@@ -48,7 +48,6 @@ function processDir(dir) {
   .on('error', function (err) { console.error('fatal error', err); })
   .pipe(es.mapSync(function (entry) {
     var filename = path.join(dir, entry.path);
-    console.log("Rewriting " + filename);
     try {
       processFile(filename);
     } catch(e) {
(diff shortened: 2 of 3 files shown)


rewriter/hsts-transform.js
```diff
@@ -0,0 +1,26 @@
...
+contents = contents.split('\n').filter(line => !/^\s*\/\//.test(line)).join('\n');
+var secStateStatic = JSON.parse(contents);
+console.log('<ruleset name="hsts">')
+secStateStatic.entries.forEach(function(entry) {
+  if (entry.mode != "force-https") {
+    return;
+  }
+  console.log('  <target host="' + entry.name + '" />');
+  if (entry.include_subdomains) {
+    console.log('  <target host="*.' + entry.name + '" />');
```
Hainish: You'll probably want to make this `entry.name.trim()` because the upstream `transport_security_state_static.json` mistakenly contains spaces after some domains.

rewriter/rewriter.js
```diff
@@ -120,7 +107,16 @@ for (var i = 2; i < process.argv.length; i++) {
   }
 }
 
-loadRuleSets();
+var hstsFilename = path.join(__dirname, 'hsts.xml');
+try {
+  fs.statSync(hstsFilename);
+} catch (e) {
+  console.error('hsts.xml not found. Run hsts-transform.js.');
```
Hainish: Any reason not to just output to `hsts.xml` instead of `stdout` from `hsts-transform.js`?

fuglede:
Just a comment: Perhaps(?) it might also be of interest to include [those hosts which are pending publication in Chromium](https://hstspreload.appspot.com/pending); as of right now, there are 3405 such hosts, which is rather significant when compared to the ~11000 in the Chromium preload list.

Hainish:
@fuglede I suggest against this, since there is a quality-control process between the point of submission and inclusion, which verifies that the HSTS header is actually present on the remote HTTPS endpoint, that includeSubdomains=true and perhaps other checks as well.  Including the pending domains may present a vector for DoS, since one may submit a domain for which HTTPS is not offered at all.

fuglede:
@Hainish: Actually checks are carried out already on submission. To see this, try submitting fleetpal.io to https://hstspreload.appspot.com, and note that it does not appear on https://hstspreload.appspot.com/pending afterwards; in particular, this checks for includeSubdomains (and, as far as I can tell, every single one of the requirements specified on the submission page).

What exactly is done between the removal hosts from the pending lists and the addition to the live lists, I'm not 100% sure about; [the docs](https://docs.google.com/document/d/1LqpwT2aAekrWPtLui5GYdHSGlZNMNRYmPR14NXMRsQ4/edit?pref=2&pli=1#heading=h.oylpfoy9k544) mention "manual" reviews but not exactly what they entail.

fuglede:
The above comment is partly voided by their [new](https://twitter.com/lgarron/status/733103612235128832) [change to how the site works](https://groups.google.com/a/chromium.org/forum/#!msg/security-dev/4gVLcw99NU4/ajei4N8CCgAJ). In that, we have in particular the following:

> However, I think site owners will appreciate that the requirements are now all checked when the domain is submitted to hstspreload.appspot.com – there is no secondary review with additional checks, and you get feedback immediately when you enter the domain on the submission site.

On the other hand, it looks like this change will bring about [updates more often](https://twitter.com/lgarron/status/730195591389552641), so that the added value would be less, comparatively.

lgarron:
> The above comment is partly voided by their new change to how the site works.

Yep, all the requirements are checked at submission time now. However, *tons* of sites develop problems in the few weeks after submission. I would highly recommend scanning using `[hstspreload](https://github.com/chromium/hstspreload) batch`

> On the other hand, it looks like this change will bring about updates more often, so that the added value would be less, comparatively.

Note that this is another problem that isn't solved yet. Updates will continue to be throttled to ≈once per Chrome version until someone (probably me) ports https://github.com/chromium/hstspreload/blob/master/cmd/transport_security_state_static_generate/transport_security_state_static_generate.go to Python and adds it to the Chromium build process.

Hainish:
@jsha do you have time to look at the few comments I made? @J0WI marked this with the [Before Next Release](https://github.com/EFForg/https-everywhere/milestone/15) milestone, but I don't think it's vital for the next release.

jsha:
Yeah, I removed that milestone - I don't think this is super urgent.

J0WI:
I added the milestone here to get rid of the "Should we add preloaded HSTS rules or not"-dilemma.

cowlicks:
@jsha closing this since it's stale, feel free to re-open

PR #4463, opened 2016-03-24
Remove clicknupload.com. No longer working.

Balgden:
Clicknupload dropped support for https. Remove ruleset as it breaks the website.


J0WI:
Do you think this will be fixed in future? If so, you can just add the `default_off="mismatched"`flag.

Balgden:
I doubt it. They redirect https to http and it's been like this for a few weeks already.

J0WI:
Okay, thanks for clarify.

PR #4467, opened 2016-03-24
Update Wikimedia.xml

davidabian:


J0WI:
You need to add test urls for the exclusions. Please also use trailing slashes there.

davidabian:
Done.

Issue #4396, opened 2016-03-17
rewriter should also rewrite hsts preload links


Hainish:
We've been removing rules from our rulesets for hosts that are included in the HSTS preload list.  This is fine within the browser context, but it looks like the rewriter only sources from our rulesets and not the preload list.  We should fix this.

Most users' browsers now include and use the preload list, so this is less important than rewriting for rules that aren't in the list.  Nonetheless, it might be useful for users with older browsers or those that don't support the preload list.

fuglede:
Regarding the browser context, as you are probably already aware, it is worth noting that the various browsers currently have wildly different preload lists, so it may not be completely obvious where to draw the line: It takes a while before the entries from the chromium list makes it all the way to Firefox ESR (and currently, there seems to be a concensus that they are removed from HTTPS-E when they make it to  ESR), but on the other hand, Mozilla is actively removing stale entries from the preload list, which Google does not do with chromium, so one is not a subset of the other. On the third hand, it can even take a while for submitted sites to make it to the preload list; currently, [2093 sites](https://hstspreload.appspot.com/pending) have been submitted but not included.

Preloaded sites are submitted as HTTPS-E ruleset PRs every now and then; would it be worth it to decide on and document what can go into the ruleset database, just to have something to point new contributors to? (I know that I've been confused myself about the matter a couple of times already.)

Hainish:
Thanks for the feedback, @fuglede.  @jsha took the reins on this issue with https://github.com/EFForg/https-everywhere/pull/4462

jeremyn:
In issue #3069, which is a list of specific sites that need rules, there was a brief discussion about whether sites that force HTTPS should be in that list. @rugk linked to [this FAQ](https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?), "Why does HTTPS Everywhere include rules for sites like PayPal that already require HTTPS on all their pages?"

In this issue it sounds like the policy is that we shouldn't make a ruleset for a site if that ruleset just duplicates what the HSTS preload list does. If so, should someone update this FAQ?

rugk:
Well... the HSTS preload list is a special case. It is not mentioned in the FAQ.

jeremyn:
[...]

PR #4474, opened 2016-03-24
ProxyBay related cleanups

funkydude:
Removes an old, dead ruleset.
Splits one out.
Updates current one with some comments.

More work needed in a separate PR.


src/chrome/content/rules/ProxyBay.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	HSTS: max-age=15552000; includeSubDomains; preload || Not in the list yet
```
J0WI: It is: https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsSTSPreloadList.inc#5074

src/chrome/content/rules/ThePirateBay.xml
```diff
@@ -11,15 +11,14 @@
 	<target host="www.thepiratebay.org" />
 	<target host="pirates-forum.org" />
 	<target host="www.pirates-forum.org" />
-	<target host="kanyebay.com" />
-	<target host="www.kanyebay.com" />
```
J0WI: host is still up

funkydude:
@J0WI Resolved

J0WI:
preloaded sites don't need to be added to HTTPS Everywhere. You can safely remove the proxybay.la rules.

funkydude:
@J0WI there doesn't appear to be a clear consensus on that right now: https://github.com/EFForg/https-everywhere/issues/4396

On top of that, since this domain will likely change again soon, I do not wish to delete and re-create it for whatever time it takes for the new domain to be added to the list. It will also need forwarding from the old to the new domain. So, I'm leaving it in for this one in specific.

J0WI:
I think this will be fixed in #4462

jeremyn:
@funkydude @J0WI I noticed this because it's old and `top-1k`. What's blocking this?

J0WI:
@jeremyn the point is, that sites in the [HSTS preload](https://hstspreload.appspot.com/) list don't need to be in HTTPS Everywhere, because they are already forced by the browser. I think browsers doings this well than HTTPS Everywhere and it would shrink our extension.

Quoting @jsha from the mailing list:
> It's fine to remove an auto-generated HSTS rule, if:
> - Its hosts are now fully covered in the HSTS preload list.
> - The secure cookie rules are not necessary (e.g. the site secures all its cookies, *or* only sets cookies that are scoped exactly to the covered HSTS domain).

But we haven't cleaned up our rules yet (see https://github.com/EFForg/https-everywhere/issues/4396#issuecomment-240007647).

jeremyn:
@J0WI So does that mean the blocker is you want @funkydude to remove `ProxyBay.xml` from the pull request?

J0WI:
Either this or at least adding a comment, that we can easily clean up preloaded hosts if we want to do this.

J0WI:
I think I'll prefer #5013 which is more up to date and also includes more PirateBay related hosts.

J0WI:
closed by #5013

PR #4475, opened 2016-03-25
Update ruleset: BitMEX.com no longer giving missing location error up…

Hainish:
…on validation


J0WI:
There is also blog.bitmex.com and testnet.bitmex.com.

Hainish:
Thanks @J0WI, added.

PR #4478, opened 2016-03-25
Add d2jsp.org and subdomains

deadNightTiger:
They recently implemented HTTPS, so I thought to make some basic rules and submit them.


J0WI:
You don't need to create a new file for every host. As long as they are related or have the same TLD, you can merge them into one file.

The 4th level www subdomains doesn't exist, so you should not include them.

deadNightTiger:
Thank you for your comment, I updated this pull request accordingly.

J0WI:
Thank you!

PR #4482, opened 2016-03-25
More subdomains for Unizar.es.xml

davidabian:


J0WI:
Great, lgtm.
Just two additional hosts: `maplicada.unizar.es` is also supported. `titulaciones.unizar.es` seems to be quite popular, but https is refused there.

davidabian:
Done, @J0WI. Thanks again.

J0WI:
Please also document `titulaciones.unizar.es`

davidabian:
The problem is that there are too many subdomains like titulaciones which are not supporting HTTPS. Should we document all of them? :cry:

J0WI:
Yes, this is for our atlas and future maintenance.

davidabian:
Partially (but perhaps sufficiently) done.

J0WI:
Awesome, thanks!

PR #4486, opened 2016-03-26
add ruleset: shirazlug.ir

salmoh:


J0WI:
Could you also add `events.shirazlug.ir` and `qa.shirazlug.ir`?

PR #4490, opened 2016-03-26
Add new Udacity subdomains.

bgkonzman:
Udacity recently added some new subdomains. Patching them in.


J0WI:
Thanks for the update!

PR #4495, opened 2016-03-27
[AOL] Removed exclusion for nonexistent URL, added...

galenide:
…the notifications.mail.aol.com target host, and secured cookies for certain domains.

### DIFF
--- a/src/chrome/content/rules/AOL.xml
+++ b/src/chrome/content/rules/AOL.xml
@@ -115,13 +115,6 @@
 	ᵐ Cert only matches api.oscar.aol.com
 
 
-	Partially covered hosts in *aol.com:
-
-		- my.screenname *
-
-	* Logout excluded
-
-
 	These altnames don't exist:
 
 		- holiday.aol.com
@@ -166,6 +159,7 @@
 	<target host="mail.aol.com" />
 	<target host="api.mail.aol.com" />
 	<target host="cdn.mail.aol.com" />
+	<target host="notifications.mail.aol.com" />
 	<target host="rpc.mail.aol.com" />
 	<target host="rpc1.mail.aol.com" />
 
@@ -190,12 +184,6 @@
 
 	<target host="expapi.oscar.aol.com" />
 
-		<!--	Added by AOL, required for http logout.
-								-->
-		<exclusion pattern="^http://my\.screenname\.aol\.com/_cqr/logout/" />
-
-			<test url="http://my.screenname.aol.com/_cqr/logout/" />
-
 		<!--	Redirects to http:
 						-->
 		<!--exclusion pattern="^http://(heroes|m|netscape|on|toshiba|welcome|www)\.aol\.com/$" /-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/AOL.xml
```diff
@@ -190,12 +184,6 @@
 
 	<target host="expapi.oscar.aol.com" />
 
-		<!--	Added by AOL, required for http logout.
-								-->
-		<exclusion pattern="^http://my\.screenname\.aol\.com/_cqr/logout/" />
-
-			<test url="http://my.screenname.aol.com/_cqr/logout/" />
-
```
J0WI: I feel a bit uncomfortable with that, since it was added by the AOL team itself.

galenide: Right now, it appears that the URL http://my.screenname.aol.com/_cqr/logout/ leads to a 404 response error page with a message about "They page you are looking for might have been removed." This error page is encountered regardless of whether the URL is accessed via HTTP or HTTPS. It is possible that the URL was functional in the past and that the URLs for logging out of the AOL service have changed over time. (It might be possible for the URL to become functional in the future, but that does not seem to be necessarily likely.)

galenide:
@J0WI @fuglede Would it be possible to confer with AOL and find out if the exclusion for http://my.screenname.aol.com/_cqr/logout/ is still necessary? (From what one can tell, when a user is signed into AOL Mail, accessing the URL gives an error page and does not log out the user.) If the exclusion is no longer necessary, it would be possible to improve the ruleset by doing such things as securing all cookies for the my.screenname.aol.com domain.

J0WI:
Since @gleeworks is no longer active I think I'm okay with the fact that this domain seems to be useless now.
Let me just rerun the test before I merge.

J0WI:
Okay, now we have a mismatch. @galenide could you have a look?

src/chrome/content/rules/AOL.xml
gleeworks: As I understand it, the specified exclusion pattern defines just the root of URLs to be excluded.  If you do an HTTP Trace on a logout of www.aol.com, it will actually invoke http://my.screenname.aol.com/_cqr/logout/mcLogout.psp with some query parameters.  I don't work on the application at AOL anymore, but I did confirm with the developers that this is still needed.  Thanks!

gleeworks:
I just got email notification about this ticket yesterday from the mention of my userid.  I don't work on the application anymore, but I confirmed with the AOL developer that the exclusion pattern is still needed for logout.  As I remember correctly, the specified exclusion pattern defines just the root of the URL to be excluded, which should apply to the actual non-SSL Logout URL (e.g., http://my.screenname.aol.com/_cqr/logout/mcLogout.psp with some query parameters).  Thanks!

src/chrome/content/rules/AOL.xml
gleeworks: BTW, that is true too.  account.login.aol.com moved to i.aol.com.

J0WI:
Just pinging @galenide to make sure you are aware of https://github.com/EFForg/https-everywhere/pull/4495#issuecomment-222727249

jeremyn:
@galenide I'm not sure what the status is here, is this ready for review?

PR #4499, opened 2016-03-27
[i4wifi.cz] Added new ruleset

Roman3349:


J0WI:
`old.i4wifi.cz` does also exist, but times out on https.

Roman3349:
@J0WI Done

J0WI:
Thanks!

PR #4503, opened 2016-03-27
[Infogr.am] Add exclusion

### LINKED ISSUE #4502: Infogr.am (partial) breaks page element
http://www.timesrecordnews.com/news/education/gol-burg-votes-to-arm-teachers-2eb94c42-1f08-3e4e-e053-0100007fb1cf-373284441.html

Infogr.am (partial) needs to be deactivated in order to fix the infograph towards the end of the article.

fuglede:
This fixes #4502.

### DIFF
--- a/src/chrome/content/rules/Infogr.am.xml
+++ b/src/chrome/content/rules/Infogr.am.xml
@@ -36,9 +36,16 @@
 	<exclusion pattern="^http://e\.infogr\.am/js/embed\.js\?" />
 		<test url="http://e.infogr.am/js/embed.js?gm4" />
 		<test url="http://e.infogr.am/js/embed.js?" />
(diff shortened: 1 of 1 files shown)


terrorist96:
Shouldn't the rule be more broad? This rule is probably breaking similar elements on different sites too.

fuglede:
I wouldn't be surprised either, but without any further patterns in how these scripts are used by third parties, it is difficult to figure out what to exclude exactly, without simply excluding the entire domain; did you come across any other problematic cases that could help in understanding how the page is used?

terrorist96:
> did you come across any other problematic cases that could help in understanding how the page is used?
No, but I'll report it in this thread if I come across it again.

fuglede:
Thanks. If it turns out to be problematic in more cases, I'll throw in my vote to simply disable the domain.

PR #4504, opened 2016-03-28
[Mbl.is] Add exclusion to fix redirect loop and breakage

fuglede:
The entirety of the "fasteignir" subpage would appear to serve redirects to HTTP, so we exclude that to avoid some loops. In some (but apparently not quite all) cases, this would also break the search form on http://www.mbl.is/fasteignir/.

### DIFF
--- a/src/chrome/content/rules/Mbl.is.xml
+++ b/src/chrome/content/rules/Mbl.is.xml
@@ -5,10 +5,12 @@
 	Problematic subdomains:
 
 		- (www.) *
-		- m *		(works)
-		- static *	(works)
+		- icelandmonitor *
+		- m *
+		- media *
+		- static *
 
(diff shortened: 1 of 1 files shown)


J0WI:
The wildcard target will need some test urls for future tests.

fuglede:
In case the testing mechanism is changed, you mean? For now it runs fine.

J0WI:
I required them in the past as documented in our guides.

fuglede:
Yeah, it's also pretty strange that it didn't complain. Anyway, I added a few.

J0WI:
Rule lgtm.
Could you add comments for the refused secure connection on `icelandmonitor.mbl.is` and `media.mbl.is`?

fuglede:
Done.

Issue #4310, opened 2016-03-09
Add amzn.to to bit.ly


IDKwhattoputhere:
amzn.to is used by Bitly to shorten Amazon links. It seems like SSL is supported here. Example: https://amzn.to/1g3Uzdp

J0WI:
done in #4505

IDKwhattoputhere:
Much appreciated, guys.

PR #4505, opened 2016-03-28
Bitly HTTPS Support

jehiah:
I'm happy to finally open up a PR to update HTTPS-Everywhere to reflect new HTTPS support on the Bitly Platform to finally give proper resolution to #1309. I'll comment here w/ a link to a Bitly Blog post talking about HTTPS support  when that's available (soon!).

This PR adds 38k domains powered by Bitly to reflect complete HTTPS support (thanks LetsEncrypt!), and consolidates various rules that covered domains powered by Bitly. 

By putting these in a single file, it will be easier to update and keep this list in sync as domains get added to or removed from the Bitly platform.

This also closes outstanding Issue #4310, and provides a more complete resolution to #4099 #4097 #1350


J0WI:
Woa, that's great!
We have to make sure that no other domain is lost in the new rule. I think this should be possible on the top of the test we already have.
Let me ping @fuglede @semenko and @jsha for having a look here.

Btw: It would be even better if bitly use preloaded HSTS for it's domains. This would make these rules obsolete.

jehiah:
@J0WI thanks for the look. `HSTS` is totally on my radar, but there are some complications. Since we don't own the domains we can't definitively assert that we can use `includeSubdomains` which means `HSTS` preloading will require more hand holding and is further out timeframe wise (automated HSTS preloading requires `includeSubdomains` flag).

I did run a scripted audit to ensure that all the domains we power were removed from other spots they were listed, and i'm pretty sure existing tests will flag if i missed any of those. 

I've taken my best pass at updating the configs i touched to remove wildcards so they pass tests, but i'm totally open to any feedback on better paths to take for those.

J0WI:
Yes, Google still requires the `includeSubdomains` attribute for daily submits. But Google and Mozilla also have some exceptions for some services. You may have more luck with a request e.g. on [Bugzilla] (https://bugzilla.mozilla.org/) or their Mailing lists (e.g. [dev-security-policy](https://lists.mozilla.org/listinfo/dev-security-policy) or [dev-tech-crypto](https://lists.mozilla.org/listinfo/dev-tech-crypto)).
You could provide a static list yourself where they can grab periodically the entries from.

> I did run a scripted audit to ensure that all the domains we power were removed from other spots they were listed, and i'm pretty sure existing tests will flag if i missed any of those. 

Could you provide the script you used?
Than we can just review the script and compare the output.

jehiah:
I'm aware of those options for pursuing HSTS. thanks for listing them!

To check for duplicate targets I applied c1e69b73c0648507a7030e1c9ffa6e040b59b0f7 locally then went through the following steps. It's not perfect, but it highlighted some duplicates which are all addressed in this PR.

```bash
python utils/merge-rulesets.py 
cd src/chrome/content/rules
cat default.rulesets | grep '<target' > default.rulesets.targets
cat default.rulesets.targets | awk -F '"' '{print "\""$2"\""}' | sort | uniq > default.rulesets.targets.domains.sorted
cat default.rulesets.targets.domains.sorted ~/bitly_bsd_20160321.csv | sort | uniq -d
```

jsha:
@jehiah: This is so awesome, thanks! I'm really happy that Let's Encrypt has enabled Bitly to support HTTPS on these domains.

We have a check we run periodically to make sure the domains in HTTPS Everywhere actually exist, and respond to HTTPS, and don't redirect to a 404 or back to HTTPS. You can run this with `./fetch-test.sh /path/to/ruleset.xml`. Note that this approach may be fairly slow for 38k domains, so I wrote a faster, simpler scanner in Go: https://gist.github.com/jsha/5190452f160b0b424ccb1beb7bb22d7a (expects a list of hostnames on stdin).

It turned up a fair number of error cases. For instance, 148.li is in the ruleset but doesn't resolve. Some, like https://aha.to/, redirect to a connection refused domain. Can you tell us more about the criteria for inclusion in the list?

Also cc @Hainish to help out with review.

jehiah:
This dataset is domains configured on the Bitly platform that now have HTTPS support on our platform (ie: that we have a certificate for), and passed a validation check that they had DNS pointing to our platform relatively recently from when i pulled the snapshot. 

I expect some natural flux in domains coming and going from our platform. I'm not sure the perfect way to handle that except to update this list periodically. Already a few in this list don't point to the Bitly platform. (and it's worth noting that flux will also be a barrier in getting these domains HSTS preloaded)

Your Go script is following redirects, so you are seeing errors following to where the root of the domain redirects to. I'm assuming that wasn't your intent. We allow our customers to set where the root of their domain redirects to, and as you have discovered it seems some of those destinations could use some TLC. In Bitly's case, requesting `/robots.txt` is something that should always return a 200. Not withstanding that, HTTPS on these domains should all work, and be valid for the Bitlink links on these domains. (I've forked your gist to update it to not follow redirects)

fuglede:
Imagine if those sites were preloaded though; the preload list has about 11k domains in it, so adding 38k would be amazing (but maybe also a bit [too much](https://docs.google.com/document/d/1LqpwT2aAekrWPtLui5GYdHSGlZNMNRYmPR14NXMRsQ4/edit?pref=2&pli=1)).

jsha:
> I expect some natural flux in domains coming and going from our platform. I'm not sure the perfect way to handle that except to update this list periodically. Already a few in this list don't point to the Bitly platform.

Yep, this is a good point.

It was actually intentional that my Go script follows redirects, since that's the behavior of `./fetch-test.sh` too. For most cases that's correct - if the HTTPS version of a website redirects to a broken page, we don't want to send people to the HTTPS version.

However, Bitly's case is different: Most visits to these branded domains don't actually go to `/`, they go to some other path. So the behavior of `/` is not really relevant. We have a tag, `<test url="..." />` to provide more meaningful URLs. We could change this ruleset to include test URLs pointing at robots.txt for each domain, but that wouldn't disable the implicit fetch for `/`. The way we handle this for other sites is to add an exclusion for `/` (if it's broken, we don't want to rewrite it). However, that's not really right for Bitly - for most of these domains `/` should work.

I think the best we can do is, as you say, update this periodically. Fortunately the breakage when it is out of date is not very big. Can you commit to updating the list once a quarter, or write a script to update based on public data so we can include that script in the repo?

Hainish:
I'm currently auditing the 38k domains in this PR with the modified `go` script provided by @jehiah.  Where a domain does not resolve or `status_code != (30x || 200)`, I'll disable the entry.  I'm running this from a few different network perspectives to corroborate the results, and will use the intersection of the two when applying changes.

Hainish:
It looks like running this script against such a large list of domains is causing some friction with network system limits.  For instance, most of the domains after a certain point result in the following:

```
zza.fm: error Get https://zza.fm/: dial tcp 69.58.188.49:443: connect: cannot assign requested address
```

This is because there are too many connections with status `TIME_WAIT`, the network cxns become saturated and the system is unable to allocate more source ports.  Setting the following may help:

```
sysctl -w net.ipv4.tcp_timestamps=1
sysctl -w net.ipv4.tcp_tw_recycle=1
```

This script takes quite a while to run, though I haven't timed it, I'd estimate about 6 hours.  If I keep running into problems I may have to forgo the auditing / exclusion of problematic targets and just merge with an overly-broad set of target domains.

Hainish:
Looks like revising the script to initialize a custom `Transport` with `DisableKeepAlives: true` fixes the problem.  Re-running.

Hainish:
To summarize:

I ran @jsha's golang script that @jehiah and I modified against the list of Bitly branded domains on two different instances of AWS (located at `us-west-1` and `eu-central-1`), compared the results of the two runs, and disabled 2,711 problematic domains.

jehiah:
❤️  thanks @Hainish!  I'm excited to see this land, and to help move the internet towards 100% encryption.

I'll take all the learnings from this PR and try to make the next update in a few months painless.

gugu:
okay, now bit.ly clients is not allowed to switch to another service. Thanks a lot!

jsha:
@gugu: The most useful actions here, in order of preference, would be:

1) Ask your new shortlink provider to support HTTPS
2) Submit a patch to HTTPS Everywhere removing your short domain from the Bit.ly rewrite list.

PR #4506, opened 2016-03-28
Netflix update

numismatika:
Netflix has enabled TLS on most of their websites so i think it is time to re-enable the ruleset.
Sites without valid cert & timeouts still are off by default.


J0WI:
Anyone with a Netflix account to review this?
cc @fuglede @semenko @loveithateit

PR #4518, opened 2016-03-30
[Handelsregister.de] add handelsregister.de -> www.handelsregister.de

julianladisch:


J0WI:
We don't redirect non existent hosts. Could you remove ^handelsregister.de completely?
You may also add a comment that only the www prefix exist.

J0WI:
@julianladisch let me know if you are interested in fixing this or if I should take it.

PR #4522, opened 2016-03-31
Update Pbsrc.com.xml

### LINKED ISSUE #3527: Pbsrc.com STILL breaks photobucket functionality

terrorist96:
Fixes #3527

### DIFF
--- a/src/chrome/content/rules/Pbsrc.com.xml
+++ b/src/chrome/content/rules/Pbsrc.com.xml
@@ -50,7 +50,7 @@
 		- static2		(→ ostatic2.pbsrc.com)
 
 -->
-<ruleset name="Pbsrc.com (partial)">
+<ruleset name="Pbsrc.com (partial)" default_off="breaks flash copy action">
 
 	<target host="pbsrc.com" />
 	<target host="*.pbsrc.com" />
--- a/src/chrome/content/rules/Pbsrc.com.xml
+++ b/src/chrome/content/rules/Pbsrc.com.xml
@@ -50,8 +50,9 @@
 		- static2		(→ ostatic2.pbsrc.com)
 
 -->
-<ruleset name="Pbsrc.com (partial)" default_off="breaks flash copy action">
-
+<ruleset name="Pbsrc.com (partial)">
+	
+	<exclusion pattern="^http://*/flash/ZeroClipboardFV3.swf"
 	<target host="pbsrc.com" />
 	<target host="*.pbsrc.com" />
 


J0WI:
Could you check if you can solve the issue by excluding just a single file?
A good way to start is by looking at the loaded resources: https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor

If you see a CORS error in the console, you may able to fix it by excluding just the .xml source. Otherwise you may have to exclude JS and SWF resources.

terrorist96:
Here's a screenshot of the console with the Pbsrc.com rule active and inactive. What am I looking for?

J0WI:
Hmm, there are some mixed content issues.
Can you check if allowing the mixed content would work?

terrorist96:
There's no option in Chrome to 'load unsafe scripts', so I'm not sure how you want me to check if allowing mixed content works or not.

J0WI:
Just have a look at https://stackoverflow.com/questions/18321032/how-to-get-chrome-to-allow-mixed-content

terrorist96:
Bro, I just said, there's no shield icon in the address bar for me to click in order to load unsafe scripts. Can you please try reproducing this bug yourself so you can see what I'm talking about?
All I know is that disabling the Pbsrc rule fixes the issue. I don't understand why it's so difficult to disable this rule on photobucket.

J0WI:
Sorry I just realized, that the Flash file is not loaded if you have no Flash Player installed.
You should see something like `/flash/zeroclipboardFV3.swf` loaded in your network monitor. Can you try if an exclusion for that would fix it?

terrorist96:
I see zeroclipboardFV3.swf when I disable the Pbsrc rule. How would I test an exclusion? I think this would have been fixed a long time ago if someone more experienced would take a look, instead of this back and forth lol.

terrorist96:
How's that ^
???

J0WI:
You regex syntax is wrong, but I would prefer to avoid the wildcard anyway when possible.
You also have to add a test url for the exclusion.

terrorist96:
Well I don't know what more you want me to do. This is too much work trying to figure out what you want me to do. Someone else can fix it; don't care anymore.

terrorist96:
Still broken in Version: 2016.5.10

J0WI:
> Well I don't know what more you want me to do. This is too much work trying to figure out what you want me to do. Someone else can fix it; don't care anymore.

Okay I close this. I proposed a fix in #3527 if you like to give it a chance again.

PR #4528, opened 2016-04-01
Disabled outdated New Yorker rule

Hainish:


Hainish:
Ready for review @jsha

J0WI:
Some subdomains are working, so rewriting would be more worth than just disabling.

Hainish:
@J0WI have you observed these domains being loaded as the primary location or just as included resources?  If it is only included resources, there's no significant advantage to rewriting, since the resource inclusion code itself can be easily modified.

Currently, subscribers can't log in and hit a paywall to read articles.  I think it best to disable this rule temporarily, until we can more thoroughly go through and audit the subdomains.

PR #4532, opened 2016-04-03
[openblox.org] New ruleset

mark-otaris:
HTTPS is supported on all subdomains.

### DIFF
--- a/src/chrome/content/rules/OpenBlox.xml
+++ b/src/chrome/content/rules/OpenBlox.xml
@@ -1,0 +1,7 @@
+<!-- HTTPS supported on all subdomains (has HSTS with includeSubdomains) -->
+
+<ruleset name="OpenBlox">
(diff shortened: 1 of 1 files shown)


fuglede:
Thanks for the PR, @mark-otaris. A couple of comments:

* The ruleset does not actually cover openblox.org itself (that is, the root domain); you could add that one.
* The add-on testing suite reports errors because [all wildcard rules need valid tests](https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md); you could add a few of those for the subdomains you know of.

mark-otaris:
@fuglede I added test URLs for a few subdomains.

fuglede:
Looks good; thanks. Do you want to add openblox.org as a target host as well?

mark-otaris:
@fuglede Done.

fuglede:
👍

PR #4540, opened 2016-04-04
Update 163.com-mixedcontent.xml

renyouguo:


src/chrome/content/rules/163.com-mixedcontent.xml
```diff
@@ -6,13 +6,16 @@
...
+	<target host="chat.study.163.com" />
 	<target host="email.163.com" />
 	<target host="mail.163.com" />
 	<target host="i.money.163.com" />
 	<target host="piao.163.com" />
 	<target host="trip.163.com" />
-
-
+	
+	<target host="ct.stu.126.net" />
```
J0WI: why is this mixed content?

renyouguo: It comes from chat.study.163.com .
Should I open another PR for it?

J0WI: You can include it into 163.com.xml since this hosts servs no mixed content.

renyouguo:
#4755

PR #4543, opened 2016-04-04
Fix bug in Chromium, Firefox to limit FQDN

Hainish:


Hainish:
ready for review @jsha

PR #4547, opened 2016-04-05
Create Discordapp.com.xml

ekaris:


J0WI:
There are some subdomains:
support
status
blog
feedmach
merch
i18n

Since this host serves a HSTS header with the `includeSubdomains` attribute you can also safely use a wildcard target.

ekaris:
Done

J0WI:
wildcard targets requires test urls. You can just use the targets that are already covered by the wildcards as test urls.

ekaris:
I think is good

J0WI:
Yes, thanks!

PR #4553, opened 2016-04-05
[Nic.cz] Added new targets

Roman3349:


fuglede:
It's of course a problem that http://jsns.nic.cz/ redirects to a broken site on its own, but I don't think we should fix the site for them. That being said, and ignoring the bad certficate, it seems like the two sites do serve almost identically functioning content, with some trivial exceptions;

    $ diff <(curl -sk https://jsns.nic.cz) <(curl -s https://www.jsns.cz)
    167c167
    < <div id="breadcrumbs">&nbsp;</div><div id="cblock">
    ---
    > <div id="breadcrumbs"><a href="/cz/index/1/Jsns_cz.html">jsns.cz</a> &raquo; Aktuálně</div><div id="cblock">
    ....

And then a minor point: Could you order the targets alphabetically? Makes maintenance a bit easier. Otherwise LGTM.

J0WI:
Thanks!

PR #4554, opened 2016-04-05
[Knot-DNS.cz] Added new target

Roman3349:


J0WI:
Could you also document the invalid cert on rpm.knot-dns.cz?

Roman3349:
@J0WI Done

PR #4562, opened 2016-04-06
[News Limited] Remove wildcard rule

### LINKED ISSUE #4561: News Limited (partial) rule needs to be turned off for news.com.au
http://www.news.com.au/national/breaking-news/no-plans-to-change-gun-laws-keenan/news-story/914ebfe8e1e7caec2b7cef8ba9ef20b0

News Limited (partial) forces HTTPS, but Chrome will give the following warning:
`Your connection is not private
Attackers might be trying to steal your information from www.news.com.au (for example, passwords, messages, or credit cards). NET::ERR_CERT_COMMON_NAME_INVALID`

Ignoring the warning and proceeding to the site works, but then you also have to load unsafe scripts. Then everything looks fine except for the red crossed out https in the URL bar.

fuglede:
The previous change introduced a broken catch-all rule on accident. Removing it fixes #4561.

### DIFF
--- a/src/chrome/content/rules/News_Limited.xml
+++ b/src/chrome/content/rules/News_Limited.xml
@@ -31,10 +31,10 @@
 
 	Problematic subdomains:
 
-		- ^ ¹
+		- ^, www ¹
 		- connect ²
 
-	¹ Works, akamai
+	¹ Works, mismatched, akamai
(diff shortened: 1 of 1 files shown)


J0WI:
merging as simple bugfix for next release.

PR #4568, opened 2016-04-07
added qrpedia.org.xml

J0WI:


fuglede:
Already exists as <code>Qrpedia.org.xml</code>;

    ['src/chrome/content/rules/Qrpedia.org.xml', 'src/chrome/content/rules/qrpedia.org.xml'] failed case-insensitivity testing.

J0WI:
Oh, nice.

PR #4570, opened 2016-04-07
New ruleset for chip-chap.com

PabloCastellano:


fuglede:
This one could be simplified a bit, as the rules could be replaced by a single

PabloCastellano:
Thanks for the pointer @fuglede! I have modified it.

fuglede:
👍

PR #4571, opened 2016-04-07
Update MNGinteractive.com.xml

### LINKED ISSUE #3934: MNGinteractive.com rule breaks stylesheet

terrorist96:
Fixes #3934
@J0WI here it is


J0WI:
Thanks!

PR #4572, opened 2016-04-07
Exclude non-https protected #1663 (mixed content causing it not to load)

Hainish:


fuglede:
Travis complains about the lack of testing;

    ERROR src/chrome/content/rules/xkcd.xml: Not enough tests (0 vs 1) for <Exclusion pattern '^http://xkcd\.com/1663/'>

Otherwise LGTM.

Hainish:
Updated @fuglede

fuglede:
Come to think of it, don't you want to exclude <code>http://(www\\.|m\\.)?xkcd\\.com/1663/</code>?

Hainish:
Good catch @fuglede, fixed

PR #4574, opened 2016-04-08
[Roblox] Add some subdomains

mark-otaris:
The www and forum subdomains now support HTTPS, for all users and also for visitors not logged in.


J0WI:
WFM:
m.roblox.com
en.help.roblox.com
uk.roblox.com
web.roblox.com

Mismatch:
corp.roblox.com
blog.roblox.com
wiki.roblox.com
shop.roblox.com
bloxcon.roblox.com
devforum.roblox.com

Can you document the problematic domains by adding comments for them on the top of the rule?
Our style guide also requires tabs instead of spaces.

mark-otaris:
@J0WI 

Those work, as does de.roblox.com. I have added them after some tests and fixed the whitespace issue.

J0WI:
Could you add comments on the top of the rule for the mismatching ones?

mark-otaris:
@J0WI Done, and I've also added many other domains and subdomains.

src/chrome/content/rules/Roblox.xml
```diff
@@ -1,6 +1,92 @@
...
+	<target host="de.roblox.com" />
+	<target host="en.help.roblox.com" />
+	<target host="forum.roblox.com" />
+	<target host="m.roblox.com" />
+	<target host="uk.roblox.com" />
+	<target host="web.roblox.com" />
+	<target host="www.roblox.com" />
+	<target host="*.rbxcdn.com" />
+	<target host="shoproblox.com" />
+	<target host="*.shoproblox.com" />
```
J0WI: There are not enough test urls for this wildcard.

mark-otaris:
@J0WI Since there are no other subdomains of shoproblox.com (es, pt, de and such do not exist even though they're linked to), I just removed the wildcard.

mark-otaris:
@J0WI The checks fail because of "No route to host" errors, but it doesn't seem to me this is a problem with the ruleset.

J0WI:
Let me rerun the tests.
But at least in the case of `www.shoproblox.com` https doesn't work (I don't tested them all manually yet).

mark-otaris:
@J0WI The tests for shoproblox.com used to succeed and it supported HTTPS, but the shop was removed in the past week and now redirects to the catalog on roblox.com. Unfortunately this redirect doesn't support HTTPS, so I had to remove targets for shoproblox.com.

In addition, uk.roblox.com and de.roblox.com are now redirects too and they don't support HTTPS, so I had to remove them as well.

The errors left are all "No route to host" errors.

J0WI:
@fuglede do you have any idea why Travis can't connect?
Is this because of a stupid configured firewall?

fuglede:
No, no idea, unfortunately, @J0WI.

J0WI:
We might be blocked by a firewall.
@mark-otaris I propose to try adding it on `utils/ruleset-coverage-whitelist.txt`.

mark-otaris:
@J0WI What would be the format of that file?

J0WI:
sha256sum of the file content + filename

mark-otaris:
@J0WI It appears to still fail.

J0WI:
I hope @Hainish could have a look when he is back.

Hainish:
Re-running tests

Hainish:
Tested on my local, worked fine.  Removed the coverage exception, there was no reason it should be there.

Subsumed by https://github.com/EFForg/https-everywhere/commit/824e79a054e063aa5568b6fd460c96a052b132b5

PR #3696, opened 2015-12-30
Update Inspectlet.com.xml

renyouguo:

PR #4575, opened 2016-04-08
Update Inspectlet.com.xml

renyouguo:


J0WI:
Just some code style: Could you move the comments from #3696 with the paragraph `Problematic subdomains:`. So they match the style of other documented problems.

renyouguo:
@J0WI Sorry,my english is too pool to understand what should I do.

renyouguo:
@J0WI Is it just like : https://github.com/EFForg/https-everywhere/pull/4577/commits/45bb14c055250f3bc14d427f7080861c245501bc

J0WI:
Renaming this file makes no sense. This are then hosts that should be enabled by default.
You could create a desperate file for falsemixed `blog.inspectlet.com` if you want.

renyouguo:
@J0WI  Ok for blog.*

src/chrome/content/rules/Inspectlet.com.xml
```diff
@@ -1,34 +1,32 @@
...
 		- __insp_ref
 -->
 
-<ruleset name="Inspectlet.com">
+<ruleset name="Inspectlet">
+
+	<!--	Mixed Content-->
+	<target host="blog.inspectlet.com" />
 
+	<rule from="^http://blog\.inspectlet\.com/wp\-content/(.*)" to="https://blog.inspectlet.com/wp-content/$1" />
```
J0WI: `(.*)` and `$1` is not needed here, that's what the core of HTTPS Everywhere already does.
You also don't need to escape `-`. But this won't work anyway since all other request will be redirected by the generic rule on [L28](https://github.com/EFForg/https-everywhere/pull/4575/commits/7f90d46fdbce55afe5d16f4a1cf39bb0451e55cf#diff-627e1fdb4727ad6e4139e127c7a09bc1R28)

But you could use a "negative exclusion" like this:
`<exclusion pattern="^http://blog\.inspectlet\.com/(?!favicon\.ico|wp-content/|wp-includes/)" />`
This requires three valid test urls that are not excluded and three that are excluded.

Issue #4516, opened 2016-03-30
Break in firefox，ok in chrome


renyouguo:
For example：
http://v.qq.com/cover/b/btiwlze2ftoy54i.html

In firefox,we have to disable the rule "QQ.com(partial)" to see the video.However ,it is ok in chrome.

renyouguo:
#4577 has no business,it is my fault.

renyouguo:
This issue has been fixed by tencent(qq.com).

PR #4577, opened 2016-04-08
Delete Baidu.com-problematic.xml

renyouguo:
pan.baidu.com & yun.baidu.com support https well now.


renyouguo:
However,there is another bug:

Break in firefox，ok in chrome · Issue #4516 · EFForg/https-everywhere
https://github.com/EFForg/https-everywhere/issues/4516

J0WI:
See comments in #4584, this needs to be merged in a single PR.

renyouguo:
@J0WI You can merge this first,then #4584 will be ok.

J0WI:
That's a bad workflow, but I'll make an exception.
Please use a single commit when just transferring a host to another file next time.

PR #4592, opened 2016-04-09
Create Gtimg.com.xml

renyouguo:


fuglede:
Could you add a non-trivial test URL to this one? As the root of the target host returns an empty body, it is hard to verify that this one works otherwise.

renyouguo:
@fuglede https://ra.gtimg.com/web/default_fodders/300x250_1.swf

fuglede:
Thanks @renyouguo; could you also add it as a

    <test url="http://ra.gtimg.com/web/default_fodders/300x250_1.swf" />

That will help us catch any errors with the site that may occur in the future. A small note on what the domain is actually used for could also be helpful (even if is not 'required' per se).

J0WI:
This results in some mixed content errors (e.g. https://ra.gtimg.com/icast/1452425/1452425.htm).

There are also various subdomains:
wximg.gtimg.com
static.gtimg.com

Broken ones:
3366.gtimg.com
3glogo.gtimg.com
img2.gtimg.com
mat1.gtimg.com
a.map.gtimg.com

J0WI:
> A small note on what the domain is actually used for could also be helpful (even if is not 'required' per se).

renyouguo:
In China, we will never go to https://ra.gtimg.com/icast/1452425/1452425.htm .The same web? Never. They are Abandoned.
However,  http://ra.gtimg.com/web/default_fodders/300x250_1.swf will be seen in v.qq.com , which is in common use.
So, Could I enable ra.gtimg.com by default?

J0WI:
You can make exclusions for those sites who are serving mixed content issues. You need to find a clever regex for them and add some test urls.

src/chrome/content/rules/Gtimg.com.xml
```diff
@@ -8,20 +8,25 @@
...
-
         <!--    Test them at:
         https://wximg.gtimg.com/tmt/_events/promo/EyxiHkkq/html/index.html
         -->
         
+        <target host="ra.gtimg.com" />
+                <test url="http://ra.gtimg.com/sc/vqq/crystal-min.js" />
+                <test url="http://ra.gtimg.com/web/default_fodders/qqliveweb/300x250_1.swf?v=20130124" />
+
+        <exclusion pattern="^http:\/\/ra\.gtimg\.com\/\w+\/\d+\/\d+\.html?$" />
```
J0WI: Please document that you made this exclusion because of mixed content errors.
Slashes don't need to be escaped. Removing the backslashes does also improve readability.

renyouguo:
Check errors have nothing todo with my PR.

J0WI:
Great! Thanks for addressing all comments.

PR #4602, opened 2016-04-09
Create Steamspy.com.xml

ekaris:


J0WI:
This contains some mixed passive content from cdn.akamai.steamstatic.com, but I think that's okay.

PR #4603, opened 2016-04-09
[UserVoice] Big update (fixes #4514)

### LINKED ISSUE #4514: UserVoice unsubscribe links broken
Current ruleset breaks UserVoice unsubscribe links like this one `http://www6.uservoice.com/unsubscribe/u/17202/e6499023e9bdd9235601feed76495b/3570127` (Not a valid link for obvious reasons).

mnordhoff:
* Exclude `www6`. Fixes #4514, and also importantly, fixes the sign up link on their homepage.
* Add ^ and `www`
* Add tests. The hosts are just ones (previously) featured in the ruleset, those i saw on their homepage, and a few client wildcard client subdomains from Google. The list is likely incomplete, but at least it's an improvement.
* Don't exclude `blog` or `developer`
* Exclude `email` and `www6` from securecookie. (I don't know if it's necessary, but it seems likely.)
* Rename from "Uservoice (partial)" to "UserVoice"
* Remove special handling for `http://(www.)uservoice.com/images/`; they don't appear to exist.
* Style


J0WI:
Thanks a lot for the update!

PR #4606, opened 2016-04-10
Create msra.cn.xml

renyouguo:


renyouguo:
@J0WI 
Is it OK?

J0WI:
It's incomplete. There are much more subdomains to document.
We have a guide about how to do this [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

J0WI:
Just two more:
msra.cn
duilian.msra.cn

Also a little code style: you don't need to close and reopen the comment brackets between the two blocks.

Please add the `(partial)` keyword to the name of the ruleset since the sites lacks of https support a lot.

PR #4608, opened 2016-04-10
[ArXiv] Update ruleset

fuglede:
The entire site appears to work now. I've sent them a query regarding the possibility of re-enabling it by default.

### DIFF
--- a/src/chrome/content/rules/ArXiv.xml
+++ b/src/chrome/content/rules/ArXiv.xml
@@ -1,33 +1,21 @@
 <!--
 	For other Cornell University coverage, see Cornell.xml
 
-
-	CDN buckets:
-
-		- s3.amazonaws.com/arxiv.org/
-
-
 	Problematic subdomains:
 
 		- www	(mismatched)
 
 -->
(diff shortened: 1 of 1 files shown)


mhl10:
I can confirm that we enabled SSL site-wide very recently, so it would be OK to accept this pull request (which really we should have created, but you beat us to it ;-)). 

Thanks,
Martin
arXiv.org

fuglede:
Hi Martin; cool that you're here. Even with this pull request, the ruleset is still disabled by default (meaning that users will not actually benefit from the added security unless they happen to notice that the ruleset is available). Would you also be okay with the ruleset being enabled by default again?

mhl10:
Sorry, I overlooked that it was still disabled in the changeset. But yes, I'd be okay with it being enabled by default. Thanks for accelerating the process--it's been on my todo list!

fuglede:
That's great; thanks.

J0WI:
@fuglede could you document the mismatches on (de|fr|xxx)\.arxiv.org?

@mhl10 feel free to implement preloaded HSTS, that would make our rule obsolete :)

fuglede:
@J0WI: Done.

PR #4609, opened 2016-04-11
Add rulesets for Acessa and Xchan

andersbateva:
These two sites support HTTPS, but don't make redirection by default. They need rulesets.

### DIFF
--- a/Acessa.xml
+++ b/Acessa.xml
@@ -2,6 +2,11 @@
         <target host="acessa.com" />
         <target host="www.acessa.com" />
 
+        <exclusion pattern="usuarios.acessa.com" />
+        <exclusion pattern="webmail.acessa.com" />
+        <exclusion pattern="aol.acessa.com" />
+        <exclusion pattern="estevao.acessa.com" />
+
         <rule from="^http:"
                 to="https:" />
 </ruleset>
--- a/Acessa.xml
+++ b/Acessa.xml
@@ -1,0 +1,7 @@
+<ruleset name="Acessa">
+        <target host="acessa.com" />
+        <target host="www.acessa.com" />
+
+        <rule from="^http:"
+                to="https:" />
+</ruleset>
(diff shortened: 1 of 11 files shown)


J0WI:
Please also include other subdomains. We have a guide about how to find them:

andersbateva:
Left-wildcard? But It's not my objective to rewrite all addresses!

J0WI:
You don't need to use a wildcard. You can just search other subdomains as described in the guide and add them as additional targets. Please also document broken ones for future maintenance.

andersbateva:
Thanks, I've read the guide you linked, and learned how to discover more subdomains.
Xchan does not have subdomains at all. To Acessa, I added some exclusions. OK?

J0WI:
`<exclusion>` only apply on target hosts, so the current exclusions will never be used.
We use to add comments on the top of the ruleset to document hosts, that doesn't work with https at all. See e.g. [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ImmobilienScout24.xml) or an example.

I also got two additional hosts:
crcmg.acessa.com
discovirtual.acessa.com

andersbateva:
CRCMG is just a redirect, and DiscoVirtual does not works as HTTP nor HTTPS.

J0WI:
Could you add a comment for future maintenance?

andersbateva:
OK, I turned the exclusions into comments, and added CRCMG (that appears to be not a redirect anymore) and DiscoVirtual. OK?

J0WI:
Almost :)
Comments should be above the `<ruleset>` tag.

J0WI:
This PR now contains some unrelated hosts.
Please open new PRs for each of them.

andersbateva:
I don't know how. I tried creating new push requests, and they ended here!
Also, judging by the progress of this push request, I suspect none of them will be approved some day anyway...

J0WI:
Yes, you pushed all your changes in your master branch and now requested to merge your master in to ours.
This is a very bad practice. You should create a new branch for every change and request to merge your branches into our master.

This makes is easier and therefor also faster to review your changes.

andersbateva:
More than 2 months have passed.... There is something wrong with this project.

Issue #2308, opened 2015-07-23
Problems on heise.de


KommX:
If you follow this link
http://www.heise.de/foto/galerie/fullscreen/?sort=neueste&page=1
you end up on a broken page.
![heise_enabled](https://cloud.githubusercontent.com/assets/7859245/8846087/acf8cc2e-3124-11e5-83b4-29a5afc7fdde.png)

If you disable the "Heise.de (partial)" rule everything's fine.
![heise_disabled](https://cloud.githubusercontent.com/assets/7859245/8846260/c7de4176-3125-11e5-983c-defeffa23866.png)

numismatika:
Heise seems to be experimenting lately with TLS,  certain parts of their page get tls and some all of a sudden redirect to plain again.
#2262 fixes the news pages 
I'll experiment with the gallery soon, probably needs exceptions for a few files.

numismatika:
@KommX second commit in my pull-request seems to fix it.

PR #4610, opened 2016-04-11
Heise.de:

numismatika:
- everything below /security/ is no longer redirected to plain for testing reasons
- secure more images on the main domain
- simplify exclusion pattern
- more testurls

### DIFF
--- a/src/chrome/content/rules/Heise.de.xml
+++ b/src/chrome/content/rules/Heise.de.xml
@@ -25,6 +25,8 @@
 
 		- www.abo.heise.de
 
+	no securable content on ^, short-urls redirect to www on their own (can't be secured either)
+
 -->
 <ruleset name="Heise.de (partial)">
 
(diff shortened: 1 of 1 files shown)


J0WI:
@numismatika could you fix the test coverage?

numismatika:
I am currently having trouble figuring out what exactly is missing.
That is why i created https://github.com/EFForg/https-everywhere/issues/4622

J0WI:
I think you need to test everything with and without www. Can you also replace `(?:www\.)?` with `(www\.)?`?
Would be great to have this in our next release.

numismatika:
@J0WI even if i remove the parts without www the test fails.
^ as target host doesn't make sense anyway because it is only used for link shortening and hotlinking articles.

J0WI:
^ as target make sense to have a secure redirect to www (or what ever), but it requires to include all the test urls twice: with and without www.

numismatika:
the folders that can be secured, are never used with ^, so unless they enable tls on the complete page, it is useless as a target.

How do you explain this one:
https://github.com/numismatika/https-everywhere/blob/5767f7151c51828d9e9b48249b489094f649c5f5/src/chrome/content/rules/Heise.de.xml

https://travis-ci.org/numismatika/https-everywhere/jobs/130533876

These do not have ^ for test reasons, and it still seems to be off.

J0WI:
https://github.com/numismatika/https-everywhere/blob/5767f7151c51828d9e9b48249b489094f649c5f5/src/chrome/content/rules/Heise.de.xml#L75 should be www to have a test for the `avw-bin` pattern
Several patterns contains uncapped dots (e.g. `.html`). A dot would mean _any character_ in regex and requires more than just one test url.

You should also sort regex and test urls in the same alphabetical order, this makes it easier to find such issues.

J0WI:
For negative lookahead the tests also require test urls that don't match the expression.

J0WI:
> For negative lookahead the tests also require test urls that don't match the expression.

@numismatika we have a new release soon (#5045), would be great if you could add the missing test urls.

numismatika:
@J0WI sorry didn't manage to solve this one yet,
the correct order of targets, rules, exceptions and testurls seems to elude me right now.
my experimental branch either has tests failing, or the rules don't apply or both.

numismatika:
@J0WI  could you check the new approach?

src/chrome/content/rules/Heise.de.xml
```diff
@@ -39,74 +39,58 @@
 
 	<!--	Complications:
 				-->
-	<target host="heise.de" />
```
J0WI: Please document why you removed it or keep redirecting to www.

src/chrome/content/rules/Heise.de.xml
```diff
@@ -39,74 +39,58 @@
...
+			<test url="http://www.heise.de/security" />
+			<test url="http://www.heise.de/security/" />
 			<test url="http://www.heise.de/stil/heise-ui.css" />
 			<test url="http://www.heise.de/support/lib/teaser_linking.js" />
+			<test url="http://www.heise.de/software/screenshots/" />
+
+
 
 	<rule from="^http://m\.heise\.de/(js/|icons/|[ai]vw-bin/|security/icons/|stil/|fonts/)" to="https://m.heise.de/$1"/>
+		<exclusion pattern="^http://m\.heise\.de/$" />
```
J0WI: Your regex doesn't match the root path, so you should be able to remove it (same on www).

src/chrome/content/rules/Heise.de.xml
```diff
@@ -39,74 +39,58 @@
...
 	<rule from="^http://www\.shop\.heise\.de/"
 		to="https://shop.heise.de/" />
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http://www\.heise\.de/security"
+		to="https://www.heise.de/security/" />
+
+	<rule from="^http://www\.heise\.de/([ai]vw-bin/|favicon\.ico|foto/icons/|icons/|ix/images/|imgs/|js/|scale/|security/|software/screenshots/|stil/|support/lib/)" to="https://www.heise.de/$1"/>
+		<exclusion pattern="^http://www\.heise\.de/$" />
```
J0WI: The above exclusion is not needed.

For the below exclusions it would be nice, if you can find the issue with them to document. Referring to git log you added them yourself a year ago.

J0WI:
rule lgtm, failed test is not related.
See inline comments for some documentation enhancements.

numismatika:
The ^ was removed during debugging tests to find the root cause of failed tests. I'll verify if we still need it.
Exclusion for the top level folder was needed to make the tests pass although from a human perspective they are not necessary at first glance.

src/chrome/content/rules/Heise.de.xml
numismatika: @J0WI the PR should have mentioned #2308 which fixed the heise foto gallery

J0WI: Yes, please mention the issues here, that makes it much easier.

numismatika:
@J0WI as you can see in the last travis run at https://travis-ci.org/numismatika/https-everywhere/jobs/153832028

`<exclusion pattern="^http://m\.heise\.de/$" />`

is needed to avoid

> ERROR src/chrome/content/rules/Heise.de.xml: No rule or exclusion applies to test URL http://m.heise.de/

J0WI:
@numismatika thanks for clarify, I now understand what you meant.
To make the test passing could you readd the exclusion? I'll merge it then.

numismatika:
line is killed, travis is happy now.

J0WI:
Thanks!

PR #4611, opened 2016-04-11
New ruleset for nanopool.org

PabloCastellano:


J0WI:
Thanks for adding this.

Can you also add `eth.nanopool.org`?
Please also leave some comments about `exp.nanopool.org` and `exp1.nanopool.org`, that they time out on https.

fuglede:
Added this in #4712, only because this one had been idle for a while.

PR #4613, opened 2016-04-11
New ruleset for melpa.org

manandbytes:


J0WI:
There is a mismatch on www.melpa.org. Can you leave a comment about that and redirect www.melpa.org to melpa.org?

manandbytes:
Here you are.

PR #4616, opened 2016-04-12
Updated rulesets for Planet-Work and On-web

fvanniere:
I've fixed and updated the rulesets for our domains. It passes the checker.


jeremyn:
Thanks for the update! I can review this, but can you please fix this problem with the tests?

```
ERROR src/chrome/content/rules/On-web.fr.xml: No rule or exclusion applies to test URL http://planet-work.on-web.fr/
```

J0WI:
closed by #11239

PR #4619, opened 2016-04-12
[Linode] Add some hosts

mnordhoff:


J0WI:
There are also some problematic customer domains on `*.nodebalancer.linode.com` and `*.members.linode.com`. Could you document this as well?

mnordhoff:
Done.

I snuck in another couple hostnames. (One of which i had to revert, sigh.)

src/chrome/content/rules/Linode.xml
```diff
@@ -1,32 +1,67 @@
...
-		- library
-		- manager
-		- stats
-		- status	(→ linode.statuspage.io)
+		- apt ¹
+		- status ²
 
+	¹ "gnutls_handshake() warning: The server name sent was not recognized"
+	  SSL Labs says "Incorrect SNI alerts: apt.linode.com"
+	  It works in Chrome and Firefox, though.
```
J0WI: I think this one is a bug in our tests.

mnordhoff: @J0WI Yes and no. If you scroll down to the bottom of the [SSL Labs report](https://www.ssllabs.com/ssltest/analyze.html?d=apt.linode.com&s=97.107.139.216), you can see there **is** a genuine issue ("Incorrect SNI alerts"). (Even if i don't know what it means.) Chrome and Firefox ignore the problem, while GnuTLS treats it as fatal, but the test isn't _wrong_.

It's a super-unimportant host (an apt repository, normally not accessed in a browser at all), so i'm inclined to disable the host and merge this as is. Maybe file a bug to follow up in the future and see if the tests can be changed to ignore it.

mnordhoff: @J0WI It means the HTTPS vhost isn't configured with "ServerAlias apt.linode.com" (though it works because it's the default vhost and the cert is a wildcard), so the server sends back an `unrecognized_name` warning alert:

    $ openssl s_client -connect apt.linode.com:443 -servername apt.linode.com -state
    CONNECTED(00000003)
    SSL_connect:before/connect initialization
    SSL_connect:unknown state
    SSL3 alert read:warning:unrecognized name
    SSL_connect:SSLv3 read server hello A
    ...

According to the obsolete [RFC 4366 § 3.1](https://tools.ietf.org/html/rfc4366#section-3.1):

       If the server understood the client hello extension but does not
       recognize the server name, it SHOULD send an "unrecognized_name"
       alert (which MAY be fatal).

According to the current [RFC 6066 § 3](https://tools.ietf.org/html/rfc6066#section-3):

    If the server understood the ClientHello extension but
       does not recognize the server name, the server SHOULD take one of two
       actions: either abort the handshake by sending a fatal-level
       unrecognized_name(112) alert or continue the handshake.  It is NOT
       RECOMMENDED to send a warning-level unrecognized_name(112) alert,
       because the client's behavior in response to warning-level alerts is
       unpredictable.

Evidently, the tests are choosing to treat it as fatal, while Chrome and Firefox ignore it.

I came across a wget thread and patch about changing their GnuTLS code from the former to the latter:

https://lists.gnu.org/archive/html/bug-wget/2013-05/msg00033.html

I dunno if the test stuff can do the same.

mnordhoff: For what it's worth, Apache 2.2.30 and 2.4.10 no longer send the warning. ([Bug 56241](https://bz.apache.org/bugzilla/show_bug.cgi?id=56241)) https://apt.linode.com/ is running 2.2.22.

mnordhoff: @J0WI It works fine with the curl/GnuTLS/pycurl versions used by Ubuntu 15.10. It's not an LTS release, but if Travis were to be switched...

mnordhoff: Simple test:
    $ sudo aptitude install python-pycurl
    $ python
    import pycurl
    c = pycurl.Curl()
    c.setopt(c.URL, 'https://apt.linode.com/')
    c.perform()

mnordhoff: `fetch-test.sh` works fine with `apt.linode.com` under 15.10, too.

mnordhoff:
This pull request is good to go now thanks to #4683.

PR #4620, opened 2016-04-13
Create Steamgifts.com.xml

ekaris:


J0WI:
This domain already send a preloaded HSTS header, so there is no need for a rule.
Thank you anyway.

ekaris:
There have been merges of domains with HSTS in the past, what makes this one different?

J0WI:
I think they were merged unintentionally.
We discussed this on the mailing list and the conclusion was, that there is no need to have preloaded hosts in our rulesets. You can open a PR to remove them if you find any.

PR #4627, opened 2016-04-14
added Golfsuisse.ch.xml

J0WI:


jeremyn:
https://m.golfsuisse.ch returns a 404. This rule should be removed and `(partial)` added to the ruleset name.

J0WI:
fixed.

PR #4630, opened 2016-04-14
added Events.ch.xml

J0WI:


jeremyn:
I think we should include (www.)events.ch. The mixed content for https://events.ch/en/ is just some images.

J0WI:
done

PR #4631, opened 2016-04-14
added Glueckspost.ch.xml

J0WI:


J0WI:
@fuglede any idea why GnuTLS refuses to connect?

fuglede:
@J0WI: I don't have time to look into it at the moment, but from the surface, it looks like an instance https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708 (but <code>https-everywhere-checker</code> should catch those, so maybe it's something different).

jeremyn:
This looks fine to me. I'd merge it except @fuglede has already commented on it and I don't want to step on toes.

@fuglede Do you mind if I merge this in?

fuglede:
@jeremyn: Sure; it looks like the original issue is no longer present. And don't worry about stepping on me in the future (at least not if the comment is about something that does not need further fixing. :)).

PR #4636, opened 2016-04-14
101weiqi.com

J0WI:
Thanks to @renyouguo for the previous work on this rule.

### DIFF
--- a/src/chrome/content/rules/101weiqi.com.xml
+++ b/src/chrome/content/rules/101weiqi.com.xml
@@ -1,3 +1,6 @@
+<!--
+	Mismatch: demo.101weiqi.com
+-->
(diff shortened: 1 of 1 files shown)


renyouguo:
@J0WI It is not ok.
http://www.101weiqi.com/play/p/82481/#main
http://7j1yxi.com1.z0.glb.clouddn.com/file/headicon_128.png

J0WI:
What's the error there?
I can't see any errors. Do I have to be logged in to reproduce?

renyouguo:
Yes, I have logged.
In https, the robot can not go.
Or we can say, we cannot see what is the robot go.

renyouguo:
#4976 @J0WI

PR #4637, opened 2016-04-14
Senate of Poland

J0WI:
Thanks to @jwilk  for the previous work


jeremyn:
Both https://w.senat.gov.pl/ and http://w.senat.gov.pl/ don't connect for me, so that domain should be updated in the comment, but this looks good otherwise.

J0WI:
I also replaced the `_` char in the file name, as our style guide recommends.

jeremyn:
The comment for `ww.senat.gov.pl` was fine, you didn't need to remove it.

J0WI:
oh, sure.

jeremyn:
Thanks.

PR #4639, opened 2016-04-14
Google - Default redirection to HTTPS

MilanKral:
A default redirection rule is missing in Google.xml.
You can see the effect when opening
http://www.google.com/humans.txt

### DIFF
--- a/src/chrome/content/rules/Google.xml
+++ b/src/chrome/content/rules/Google.xml
@@ -431,4 +431,9 @@
 		<test url="http://www.google.ca/firefox/" />
 		<test url="http://www.google.com/firefox/" />
 		<test url="http://www.google.com.au/firefox/" />
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Google.xml
```diff
@@ -431,4 +431,9 @@
 		<test url="http://www.google.ca/firefox/" />
 		<test url="http://www.google.com/firefox/" />
 		<test url="http://www.google.com.au/firefox/" />
+
+	<!-- Default redirection to HTTPS -->
+	<rule from="^http:"
+		to="https:" />
```
J0WI: This would likely break a lot of other stuff, please only redirect humans.txt

J0WI:
Please open a new PR that only redirects `humans.txt`.

PR #4643, opened 2016-04-14
New ruleset: HKGov-ITF.xml

Riotism:


src/chrome/content/rules/HKGov-ITF.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Innovation and Technology Fund">
+
+	<target host="itf.gov.hk" />
+	<target host="www.itf.gov.hk" />
+	<target host="www3.itf.gov.hk" />
+	
+
+	<rule from="^http://(?:www\.)?itf\.gov\.hk/"
```
J0WI: the part in brackets is unnecessary, but please leave a comment that https on ^itf.go.hk is not working.

PR #4652, opened 2016-04-17
Schibsted.no

numismatika:
i added a couple of subdomains found on norwegian news-sites and those urls from the server alternative names that resolve to a valid ip.
Some internal or old hostnames that do not resolve here still are also found in the cert that matches most domains here.

### DIFF
--- a/src/chrome/content/rules/Schibsted.io.xml
+++ b/src/chrome/content/rules/Schibsted.io.xml
@@ -192,11 +192,11 @@
 	<target host="api.messaging.pre.schibsted.io" />
 	<target host="images.messaging.pre.schibsted.io" />
 	<target host="rkt.pre.schibsted.io" />
-	<target host="
+
 	<target host="embla.precollab.schibsted.io" />
 	<target host="www.embla.precollab.schibsted.io" />
 	<target host="api.sgd.precollab.schibsted.io" />
-	<target host="
+
 	<target host="presence.schibsted.io" />
 	<target host="best.privacy-dev.schibsted.io" />
 	<target host="bestest.privacy-dev.schibsted.io" />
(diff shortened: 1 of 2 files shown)


J0WI:
retest

J0WI:
Can you document how all this different hosts are related? You may also find a more generic name for the ruleset.

numismatika:
For now SAN they share the same cert, my norwegian is non existent so my guess is that it is the common service provider for these newspapers or a holding company.

J0WI:
Such holdings can be quite large, I would separate the files than and link them in comments.

numismatika:
I would have done a merge now and refinement later.
Breaking it into seperate files would require more time and a norwegian user to help.

numismatika:
For me mail.schibsted.no offers a valid outlook web interface.
I therefore suggest to overwrite the changes 2weeks ago that @Hainish  did

J0WI:
This branch has conflicts that must be resolved

J0WI:
Some hosts are already listed in other rulesets.

numismatika:
Probably right @J0WI , untangling the web of norwegian newssites takes a while.

jeremyn:
`Schibsted.no.xml` should be split up into rulesets organized around specific domains, as @J0WI (partially) said in https://github.com/EFForg/https-everywhere/pull/4652#issuecomment-235309309. Also, just spot checking, some of the targets are bad and should be reviewed by @numismatika before a maintainer looks more closely.

The targets should be restricted to just public-user-facing domains like `www`. We don't need to cover really technical domains like `stoneboard.dev.vgnext.schibsted.io`.

I'm not personally committing to review this, but I think these are minimal requirements before any maintainer is likely to look more closely.

numismatika:
I will split it by Domain.
Looks like wildcard dns is used as well by Schibsted so it needs some filtering indeed.

PR #4653, opened 2016-04-17
[Ahosting.cz] Added new ruleset

Roman3349:


J0WI:
Can you leave a comment, that there are problematic customer subdomains? (e.g. uvt.ahosting.cz, 1.ahosting.cz, octaviaclub2.ahosting.cz)

LGTM then.

Roman3349:
@J0WI Added

PR #4658, opened 2016-04-18
Add several rules

ott:


J0WI:
Please have a look at our style guide. We require some tests for wildcard hosts.
Please also open a separate PR for every new ruleset, as long as they are not related.

Hainish:
source repo deleted

PR #4662, opened 2016-04-20
Update Calendars.com.xml

suryapatt:
We are from Calendars.com development team. This is a fix for #3523  (Calendars.com completely broken). We have modified the rule set and also added more tests. We were successful in building and applying this fix locally.


suryapatt:
Can we get an idea of when will be this fix released for live ?

fuglede:
Not from here, but releases come pretty quickly these days; you could try poking @Hainish for details.

PR #4667, opened 2016-04-21
[zhihu] Add zhuanlan.zhihu.com

Arthur2e5:


Arthur2e5:
The CI error looks like a temporary network problem to me, since they are not consistent in their error domains. Please retest.

PR #4675, opened 2016-04-24
Add ImmobilienScout24.de

IDKwhattoputhere:


J0WI:
Please also include other subdomains. We have a guide about how to find them:

IDKwhattoputhere:
That's only how the rule should look like or am I missing something? I can't seem to find where it says how to find the subdomains.

J0WI:
```
If you're not sure what subdomains might exist, you can iteratively use google queries and enumerate the list of results like such:

    site:*.eff.org
    site:*.eff.org -site:www.eff.org
    site:*.eff.org -site:www.eff.org -site:ssd.eff.org

... and so on.
```
Please also document non working subdomains for future maintenance.

IDKwhattoputhere:
Not sure how I missed that. Anyway I made a change to the file. Any mistakes?

src/chrome/content/rules/ImmobilienScout24.xml
```diff
@@ -0,0 +1,35 @@
...
+	<target host="tracking.immobilienscout24.de" />
+	
+	<!-- not working:
+		blog.immobilienscout24.de
+		developer.immobilienscout24.de
+		forum.immobilienscout24.de
+		www.focus.immobilienscout24.de
+	-->
+
+	<rule from="^http://(?:www\.)?immobilienscout24\.de/"
```
J0WI: Please remove the `?:` and also add a comment for the mismatch on ^immobilienscout24.de
Comments should be on the top above the ruleset.

J0WI:
Thanks a lot!
I found an additional host that should be included here:
www.static-immobilienscout24.de (^ does not exist)
wwworig.static-immobilienscout24.de (mismatch)

IDKwhattoputhere:
I feel like that last change shouldn't cause an error. What now?

J0WI:
Travis hang some times, tests should pass if I would rerun them. But there are still a few points:
`www.static-immobilienscout24.de` itself can be safely rewritten (add it as target host). The comment for non-existent ^ is correct, but you should rephrase it (doesn't fit in the non-woking hosts list).

The comments should be realy on the top, means above the ruleset tag. It would be useful to have a small description why the non working ones are not working. You can also make categories.
See [other rulesets](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/cesky-hosting.cz.xml) for an example.

Just ping me if you have questions.

IDKwhattoputhere:
pong

J0WI:
Looks great now, thanks for contributing!

Issue #4645, opened 2016-04-14
Travis's curl is too old, causing spurious fetch-test.sh issues


mnordhoff:
Spun off from [comments](https://github.com/EFForg/https-everywhere/pull/4619#discussion-diff-59726811) on pull #4619. CC @J0WI.

Travis runs Ubuntu 14.04, which is 1.9 years old and apparently hopelessly out-of-date in Internet-years. ;-) Its (|lib|py)curl version is too old for modern features that we need, causing spurious `fetch-test.sh` failures.

### ECDSA

First and critically, it fails on ECDSA websites, such as an ever-growing number of CloudFlare clients. For example, the `Archive.is` ruleset updated in #4599, which [passed Travis a week ago](https://travis-ci.org/EFForg/https-everywhere/jobs/121946285), would [fail `fetch-test.sh` if proposed today](https://travis-ci.org/mnordhoff/https-everywhere/jobs/123192829):

    INFO Finished comparing http://archive.today/ -> https://archive.today/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://blog.archive.today/ -> https://blog.archive.today/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://www.archive.today/ -> https://www.archive.today/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://archive.is/ -> https://archive.is/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://www.archive.is/ -> https://www.archive.is/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://archive.li/ -> https://archive.li/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://blog.archive.li/ -> https://blog.archive.li/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    INFO Finished comparing http://www.archive.li/ -> https://www.archive.li/. Rulefile: src/chrome/content/rules/Archive.is.xml.
    ERROR src/chrome/content/rules/Archive.is.xml: Fetch error: http://archive.today/ => https://archive.today/: (35, 'gnutls_handshake() failed: Handshake failed')
[...]

PR #4683, opened 2016-04-26
Reinstall pycurl, making sure it uses OpenSSL instead of GnuTLS

Hainish:
(from https://github.com/EFForg/https-everywhere/issues/4645#issuecomment-214907617)

The problem seems to be not that we're using an outdated version of `curl` and `libcurl`, but that the version of `pycurl` that is installed by default in the travis build system uses `gnutls`, which apparently as @mnordhoff states does not support ECDSA keys.

I found that running the fetch test on https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Archive.is.xml on my local instance of Ubuntu 14.04 did not result in an error.  Thinking it curious, I noticed that the errors that we encounter are referencing `gnutls`, but that my own pycurl uses `openssl`.  I had both libraries installed, but it seems that pycurl prefers `openssl` when it is available, unless explicitly told not to.  When travis hits the `pip install -r test/rules/requirements.txt` step, it skips over pycurl because the build environment already has it installed:

```shell
$ pip install -r test/rules/requirements.txt

Requirement already satisfied (use --upgrade to upgrade): pycurl in /usr/lib/python2.7/dist-packages (from -r test/rules/requirements.txt (line 1))
```

This system pycurl in `/usr/lib/python2.7/dist-packages` must have had `gnutls` explicitly preferred.  We can fix this by following up the `pip install -r` commands with a `PYCURL_SSL_LIBRARY=openssl pip install --upgrade --force-reinstall --compile pycurl` command.  This seems to work, using `openssl` instead: https://travis-ci.org/Hainish/https-everywhere/jobs/125960621

I've opened a PR with the fix https://github.com/EFForg/https-everywhere/pull/4683

Fortunately, this is much quicker than building `curl` and `libcurl` on travis - in another branch I saw it adding ~2.5 mins to our test time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ virtualenv:
 install:
   - pip install -r test/rules/requirements.txt
   - pip install -r test/chromium/requirements.txt
+  - PYCURL_SSL_LIBRARY=openssl pip install --upgrade --force-reinstall --compile pycurl
 env:
   - FIREFOX_VERSION=firefox-latest
   - FIREFOX_VERSION=firefox-esr-latest
@@ -32,6 +33,6 @@ before_script:
   - export FIREFOX=$PWD/firefox/firefox
 script:
   - ./test.sh
-  - # test/travis-ruleset-fetch.sh
+  - test/travis-ruleset-fetch.sh
 sudo: required
 dist: trusty


Hainish:
Ready for review @jsha.  See comment at https://github.com/EFForg/https-everywhere/issues/4645#issuecomment-214907617

jsha:
LGTM. Pending update to PR description.

Hainish:
(from https://github.com/EFForg/https-everywhere/issues/4645#issuecomment-214907617)

The problem seems to be not that we're using an outdated version of `curl` and `libcurl`, but that the version of `pycurl` that is installed by default in the travis build system uses `gnutls`, which apparently as @mnordhoff states does not support ECDSA keys.

I found that running the fetch test on https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Archive.is.xml on my local instance of Ubuntu 14.04 did not result in an error.  Thinking it curious, I noticed that the errors that we encounter are referencing `gnutls`, but that my own pycurl uses `openssl`.  I had both libraries installed, but it seems that pycurl prefers `openssl` when it is available, unless explicitly told not to.  When travis hits the `pip install -r test/rules/requirements.txt` step, it skips over pycurl because the build environment already has it installed:

```shell
$ pip install -r test/rules/requirements.txt

Requirement already satisfied (use --upgrade to upgrade): pycurl in /usr/lib/python2.7/dist-packages (from -r test/rules/requirements.txt (line 1))
```

This system pycurl in `/usr/lib/python2.7/dist-packages` must have had `gnutls` explicitly preferred.  We can fix this by following up the `pip install -r` commands with a `PYCURL_SSL_LIBRARY=openssl pip install --upgrade --force-reinstall --compile pycurl` command.  This seems to work, using `openssl` instead: https://travis-ci.org/Hainish/https-everywhere/jobs/125960621

I've opened a PR with the fix https://github.com/EFForg/https-everywhere/pull/4683

Fortunately, this is much quicker than building `curl` and `libcurl` on travis - in another branch I saw it adding ~2.5 mins to our test time.

fuglede:
Does this also mean that we are in the clear regarding the potential falsely disabled rulesets from the previous database-wide run(s) -- I imagine that those used a proper <code>pycurl</code> setup?

Hainish:
@fuglede It looks like in `ubuntu` at least, `libcurl4-openssl-dev` and `libcurl4-gnutls-dev` are mutually exclusive, and either one is needed to `pip install pycurl`.  However our instructions in https://github.com/EFForg/https-everywhere/blob/master/install-dev-dependencies.sh install `libcurl4-openssl-dev` on developers machines.  So unless the developer running the manual fetch-test followed an atypical install process, I'd guess rulesets haven't been falsely disabled.  It's probably worth re-enabling any rules with `default_off='failed ruleset test'` and re-running the fetch test for all rules, though.

Hainish:
@fuglede I've created a new repo which is pretty much just a docker wrapper for doing a full fetch test.

This re-enables all rules marked as `default_off='failed ruleset test'` and then performs a fetch test on all rules using the `disable-broken-rulesets` config.  At the end of the run, it dumps a patch file into a directory that the host shares with the docker container, and exits.

I figure this will be useful for running on a server which itself doesn't have https-everywhere installed, but has docker.  Obviously you want to run this on a server, since the full fetch test will take a long time to run, and running on a workstation will result in intermittent connectivity issues.

https://github.com/EFForg/https-everywhere-full-fetch-test

fuglede:
Nifty! The only issue I could see with that is that approach it will not take care of the error messages inserted automatically in the XML comment during the initial run.

PR #4685, opened 2016-04-27
Add Egotastic.com

IDKwhattoputhere:
Mixed content warning for some advertising scripts. Is that a problem? Seems to me like security shouldn't be compromised for ads. What's the rule of thumb here?
Edit: Would it be better to add a platform="mixedcontent"?


J0WI:
There are some mixed ads, but the site works.

PR #4688, opened 2016-04-28
Fix rules for Red Hat

jimpriest:
* fix jobs - no longer goes to third party url
* fix investors url

### DIFF
--- a/src/chrome/content/rules/Red_Hat.xml
+++ b/src/chrome/content/rules/Red_Hat.xml
@@ -125,25 +125,14 @@
 	<target host="support.redhat.com" />
 	<target host="webmail.redhat.com" />
 	<target host="www.redhat.com" />
+	<target host="jobs.redhat.com" />
+	<target host="investors.redhat.com" />
 
 	<!--	Complications:      -->
 	<target host="www.cz.redhat.com" />
 	<target host="et.redhat.com" />
 
-		<exclusion pattern="^http://investors\.redhat\.com/(?!common/)" />
-
-			<!--	+ve:
-					-->
-			<test url="http://investors.redhat.com/corporate-governance.cfm" />
-			<test url="http://investors.redhat.com/results.cfm" />
-
-			<!--	-ve:
-					-->
-			<test url="http://investors.redhat.com/common/css/ndq-common.css" />
-
-
 	<securecookie host="." name="." />
-
 
 	<rule from="^http://www\.cz\.redhat\.com/"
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Red_Hat.xml
```diff
@@ -168,10 +160,10 @@
 		<test url="http://et.redhat.com/foo?bar" />
 
 	<rule from="^http://investors\.redhat\.com/"
-		to="https://investor.shareholder.com/" />
+		to="https://investors.redhat.com/" />
```
fuglede: This rule could just be removed completely to have it be caught by the catch-all at the bottom. The domain should also not be listed under "complications" anymore.

jimpriest: Thanks for the feedback! I cleaned up these and removed both jobs and investors from the 'complications' section.

src/chrome/content/rules/Red_Hat.xml
```diff
@@ -126,12 +126,9 @@
 	<target host="webmail.redhat.com" />
 	<target host="www.redhat.com" />
 
-	<!--	Complications:
-				-->
+	<!--	Complications:      -->
 	<target host="www.cz.redhat.com" />
 	<target host="et.redhat.com" />
-	<target host="investors.redhat.com" />
```
Hainish: Why remove this line?  Looks like investors.redhat.com has a valid cert

jsha:
Tests are failing:
```
ERROR src/chrome/content/rules/Red_Hat.xml: Not enough tests (0 vs 1) for <Rule from '^http://jobs\.redhat\.com/' to 'https://www.redhat.com/en/jobs'>
```

jimpriest:
Ack - I need to RTFM on this - thought we didn't need that one any longer. Let me fix.

jimpriest:
I tweaked again. I removed all the investor related tests - that was all outdated I believe.

I will try to find some time in the next day or two to review the docs on how this all supposed to work. I thought this was going to be a simple fix :)

src/chrome/content/rules/Red_Hat.xml
fuglede: The target (as well as the jobs.redhat.com one) is just moved to [here](https://github.com/EFForg/https-everywhere/pull/4688/files#diff-612abbdc45c9a25f4bf6477a15bf5e76R129), since it no longer adds complications.

PR #4695, opened 2016-04-30
[CJS-CDKeys.com] Splits ruleset due to false MCB; fixes #4498

### LINKED ISSUE #4498: cjs-cdkeys.com
Is currently broken, owner has been notified.

fuglede:
This modifies <code>trivial-validate.py</code> to no longer include the test for unescaped dots in the exclusion patterns; this test was introduced by @jsha in fb2d2bfd8ad0ca9ae169d3de283b58d0a3a1838e, which makes me think that it is probably there for a reason, though (no reason is given in the commit message specifically, but the same commit fixes up two patterns where wildcards were used (erroneously in one case)).

In this case, I ended up scratching my head for a good while because of it, as I used wildcards in an exclusion in order to fix #4498.

Edit: This no longer changes the test but only fixes the ruleset; see below.

### DIFF
--- a/src/chrome/content/rules/CJS-CDKeys.com-mixed.xml
+++ b/src/chrome/content/rules/CJS-CDKeys.com-mixed.xml
@@ -1,0 +1,12 @@
+<ruleset name="CJS-CDKeys.com (mixed content)" platform="mixedcontent">
+
+	<target host="cjs-cdkeys.com" />
+	<target host="www.cjs-cdkeys.com" />
+
+
+	<securecookie host="^\.cjs-cdkeys\.com$" name=".+" />
+
+
+	<rule from="^http:" to="https:" />
+
+</ruleset>
--- a/src/chrome/content/rules/CJS-CDKeys.com.xml
+++ b/src/chrome/content/rules/CJS-CDKeys.com.xml
@@ -1,11 +1,30 @@
-<ruleset name="CJS-CDKeys.com">
+<!--
+	Rule to secure static content. See CJS-CDKeys.com-mixed.xml for
+	a more general ruleset, including a rule that cause breakage in
+	mixed content blocking browsers.
+-->
+<ruleset name="CJS-CDKeys.com (partial)">
 
 	<target host="cjs-cdkeys.com" />
 	<target host="www.cjs-cdkeys.com" />
 
+	<!-- Exclude everything but stuff that is static enough. -->
+	<exclusion pattern="^(?![\w:/.-]+\.(css|gif|ico|jpg|js|png)$)" />
+		<!-- Matching exclusion -->
+		<test url="http://cjs-cdkeys.com/" />
+		<test url="http://www.cjs-cdkeys.com/" />
+		<test url="http://www.cjs-cdkeys.com/account.php" />
+		<test url="http://www.cjs-cdkeys.com/cart.php" />
+		<test url="http://www.cjs-cdkeys.com/wishlist.php" />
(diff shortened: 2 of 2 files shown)


J0WI:
Would it be possible to use something like `[\w/]+` instead of `.*`?

fuglede:
I did actually try that first; that produces a similar error in the current test setup..

J0WI:
Okay, @jsha should clarify what he did there...

jsha:
That's interesting. It was over a year ago so my memory was fuzzy, but I'm guessing it was a reaction to the one incorrect usage and the fact that most exclusion patterns seem to use more specific matching. What was the error you got with `[\w-/]+`?

I'm fine with removing that check.

jsha:
Though if we remove it, we should also delete the function.

fuglede:
@jsha: I should have probably been clearer on that, but in order to have the pattern match what I wanted, the character group also needs to include a dot. In that case, I get the following error, independently of whether or not the dot is escaped:

	$ bash test.sh
	+ '[' -n '' ']'
	+ git rev-parse
	++ git rev-parse --show-toplevel
	+ cd /home/fuglede/https-everywhere
	+ source utils/mktemp.sh
	+++ which mktemp
	++ MKTEMP=/bin/mktemp
	++ mktemp
	++ /bin/mktemp
	+ TMP=/tmp/tmp.r2mY0htbMF
	+ trap 'rm "$TMP"' EXIT
	+ '[' -d test/rules ']'
	+ '[' 0 -gt 0 ']'
	+ python2.7 test/rules/src/https_everywhere_checker/check_rules.py test/rules/coverage.checker.config
	ERROR Exception parsing src/chrome/content/rules/CJS-CDKeys.com.xml: bad character range at position 12

By the way, another thing we could do would be to disallow only dots that are not followed by <code>+</code>, <code>*</code>, or <code>{</code>. In those cases, it should be clear that the user did not forget to escape them on accident (although it's hard to rule out that there's also a use case for the single-dot wildcard).

J0WI:
A single dot could still be used as `.{1}`. But what is the benefit of using a dot instead of restrictive list (e.g. `[\wöäü\?\$\.%&#!~-]`)?
A dot also includes some characters that should not be used in urls (like `\s`).

fuglede:
In this case, the benefit was that using a character group containing a dot was also impossible (which I suppose is the real problem).

fuglede:
Ah, my bad! I had written <code>[\w:\/-.]</code> instead of <code>[\w:\/.-]</code>, which is what the parser was complaining about. I updated the PR to no longer disable the test but to simply fix the broken ruleset.

J0WI:
Escaping a normal shouldn't be necessary, isn't it?

fuglede:
@J0WI: A normal slash, you mean? If so, no, I don't think so.

J0WI:
Yes I meant a slash (corrected). Would you like to remove it then?
We just have a few left in all our rules.

fuglede:
Done (even if I don't completely see the benefit. :))

J0WI:
Readability and,of course, mainly [code style](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).
Thanks for changing!

jsha:
Excellent, thanks for tracking this down!

fuglede:
Came across this one again, and it turns out that the script is not actually working in all cases, cf. #7320. Having fixed it locally, it looks like there's unfortunately a ton of violations in the current ruleset database:

    $ python trivial-validate.py | grep failure | wc -l
    221

Issue #4206, opened 2016-02-22
blocking http://www.linksyssmartwifi.com 


sTeglo:
I have version 5.1.3 running inside Firefox 44.0.2 and it will not load the Linkssys Smart Wifi page.  This page is used to gain access to my home router from outside my network.  It's sort of big deal for anyone that is using the Linksys Smart Wifi stuff to manage their home router.  I realize that I'm not in the same league as most of you guys but it has good enough functionality for me and it would be great if you could fix this problem. Thanks, love your work!

funkydude:
What page are you referring to exactly? I don't have an account, so I can't log in, but simply browsing to https://www.linksyssmartwifi.com/ works fine for me. It displays the account log-in screen.

sTeglo:
thanks for taking interest in this issue.  I have uploaded a 90 second video showing you the details   http://1drv.ms/1QBUIrw  (note that you may need to set the resolution to see it clearly - attached screen grab showing the resolution control)

![set resolution on video playback](https://cloud.githubusercontent.com/assets/17417268/13269927/8499aa36-da3f-11e5-9e80-640382783e62.png)

funkydude:
Thanks a lot for the video. Unfortunately, I still can't reproduce this. I tried in Chrome on Windows and Firefox on Android, both work over HTTPS fine and both are updated.

I then tried accessing the HTTPS version over a proxy that allows scripts, which also worked.

As a final step I tested the website against the SSL Labs test, and both IPs returned results fine.

I *think* uninstalling an addon removes all related settings? Maybe give that a try.

stevenbenner:
@sTeglo Just for clarification, do you experience this problem at home? If not, are you on a network controlled by a Linksys router when this is happening?

stevenbenner:
@sTeglo @J0WI - Since pr #4699 has been merged, this issue should be fixed in the next HTTPS Everywhere release.

shallowham7:
Nice post..Linksys Smart Routers Installation and Setup  - Setup your linksys router.If you have any query related linksys router, feel free to contact us. We are available 24/7 to help our customers.visit us our site -[]( https://linksyssmartwifinet.com/)

shallowham7:
Nice post..Linksys Smart Routers Installation and Setup - Setup your linksys router.If you have any query related linksys router, feel free to contact us. We are available 24/7 to help our customers.visit us our site - **https://linksyssmartwifinet.com/**

shallowham7:
[...]

PR #4699, opened 2016-05-01
Remove linksyssmartwifi.com rules

stevenbenner:
This is a revert of commit 593e7cfa106cd147fa738510d1fb5c1403a984da.

### Summary

Unfortunately, forcing https is not compatible with the default settings on most Linksys EA and WRT series routers today. By default these routers have https access disabled on the router’s internal web server, which handles requests by users within the router’s network. The reason for this default is because the routers use self-signed certificates for https that cause browsers to show a big, scary warning.

When you are not behind a Linksys router the Linksys Smart Wi-Fi cloud will work over https. In fact it automatically redirects all http requests to https.

Merging this pull request will probably resolve GitHub issue #4206.

### Background

When you are on a network behind a Linksys Smart Wi-Fi router any computer using DHCP will use the router as its DNS server. The router’s DNS server will report the IP address of the linksyssmartwifi.com domain as the router’s own IP address. Thus sending any internal users directly to the web server on the router, instead of the web servers on the internet.

#### Preconditions
* Be on a network behind a Linksys EA or new WRT series router
* Have your computer set to use the router as your DNS server (default behavior in DHCP)
* Have a recent version of HTTPS Everywhere installed in your browser

#### Steps
1. Factory reset the router to default settings
2. Set up the router
3. Attempt to access http://linksyssmartwifi.com/ with HTTPS Everywhere enabled

#### Observed behavior
* Browser will fail to load the URL because the router’s web server is not listening to port 443

### Workaround
Users can work around this by enabling https access. But to do this they must either disable HTTPS Everywhere or navigate to the router’s IP address in their browser (http://192.168.1.1/ by default) to access the administration web application.

Once the user has access to the admin page they can enable https by completing the following steps:

1. Open the "Connectivity" applet by clicking that item in the left navigation
2. Navigate to the "Administration" tab
3. Check the “HTTPS” checkbox.
4. Click the “Ok” or “Apply” button

Once they have enabled https on the router users may access the web site via https, but must either bypass the self-signed certificate warning or add an exception to the browser for that URL.

CC: @loveithateit (original commit author) @sTeglo (original bug reporter) @funkydude (original bug commentator)


stevenbenner:
Info on build failures:

```
ERROR src/chrome/content/rules/Cisco.xml: Fetch error: http://csc-stage.cisco.com/ => https://csc-stage.cisco.com/: (6, 'Could not resolve host: csc-stage.cisco.com')

ERROR src/chrome/content/rules/Cisco.xml: Fetch error: http://solutionpartner.cisco.com/ => https://solutionpartner.cisco.com/: (60, 'SSL certificate problem: certificate has expired')
```

These are probably outdated rules/URLs. My change does not affect the current cisco.com rules.

Do I need to try to fix the current cisco ruleset before we can land this?

sTeglo:
Steven thank you for dogging this issue to the ground.

J0WI:
@stevenbenner since you touched the Cisco rule you should also fix the tests. You can also try to rebase your branch if this has been fixed right before.

stevenbenner:
I have submitted pr #4955 to remove the subdomain that is no longer available. And it looks like Cisco has fixed the expired certification on the other subdomain.

Once #4955 has landed I'll rebase this pull request, which should fix the build.

J0WI:
Let me rerun the tests

stevenbenner:
For some reason Travis was testing (and failing) on some random baidu stuff.

Rebased on master to avoid testing unrelated changes.

PR #4711, opened 2016-05-05
[Apple] Added checkrepair and mif subdomains

galenide:


src/chrome/content/rules/Apple.xml
```diff
@@ -78,6 +78,8 @@
 		- a1.phobos *
 		- a1721.phobos *
 		- a1980.phobos *
+		- service.ess		(self-signed cert in chain)
+		- service[12].ess	(ditto)
```
J0WI: you may merge this two comments, but please do not use `ditto`

src/chrome/content/rules/Apple.xml
```diff
@@ -103,27 +105,22 @@
 		- appleid
 		- appleseed
 		- appleseedcdn
-		- application
-		- as-images
 		- beta
 		- bugreport
 		- buyiphone
 		- buyiphone[1-4]
 		- certifications
```
J0WI: As long as they are listed as `<target host=` you can remove the `Fully covered hosts` block.

J0WI:
Thanks!

PR #4720, opened 2016-05-08
[Craigslist.ca] New Ruleset

lucaspetter:


fuglede:
Mostly looks good to me, but https://blog.craigslist.ca is mismatched (as for the craigslist.org case).

lucaspetter:
Thanks for noticing the blog.craigslist.ca subdomain. However, it looks like nothing is there, it mismatches over https and serves a 403 over http. Whereas blog.craigslist.org mismatches over https but serves the blog site over http. I also couldn't find any links to blog.craigslist.ca, all blog links throughout Craigslist's non-US domains point to blog.craigslist.org.

Should I exclude blog.craigslist.ca? Or redirect it to blog.craigslist.org? Or just leave it?

fuglede:
I'd say exclude it; if it's not used, then excluding it is not an issue. If they ever decide to use it, and we do not exclude it, then we will have a broken rule.

lucaspetter:
Okay, sounds like a good plan.

Issue #4578, opened 2016-04-08
https://www.explainxkcd.com


foobar13373:
https://www.explainxkcd.com works (in the first step redirects to HTTP, however, but it works via HTTPS if you manually enter HTTPS on every URL)

mnordhoff:
[I made a trivial ruleset](https://github.com/mnordhoff/https-everywhere/commit/42040b8e70b36569db58d5e7cfd2db7504bdc368), but we can't use it. The [sign up page](http://www.explainxkcd.com/wiki/index.php?title=Special:UserLogin&type=signup&returnto=Main+Page) uses an HTTP-only captcha*. You can't sign up over HTTPS.

* Recaptcha supports HTTPS, of course. But explain xkcd forces `<script src="http:...`.

fuglede:
I [notified](https://twitter.com/fuglede/status/728166658636926976) one of the site authors. @mnordhoff: You could mark the ruleset as <code>platform="mixedcontent"</code> to have something that's better than nothing.

mnordhoff:
I pushed a branch with `platform="mixedcontent"`.

https://github.com/mnordhoff/https-everywhere/tree/explainxkcd.com-platform
mnordhoff/https-everywhere@1a0ea268562121d5618ca997c147e323fa717bb8

I think it's better to wait a little while to see what the site author says instead of merging that now?

fuglede:
Yep, let us give them a couple of days.

fuglede:
Closing by #4724, but unfortunately the false mixed content blocking issue persists.

PR #4724, opened 2016-05-09
[Explainxkcd.com] New ruleset

rugk:


fuglede:
There's no PR available but actually @mnordhoff had worked on this one already and came across an issue; see https://github.com/EFForg/https-everywhere/issues/4578.

rugk:
Mhh, okay, for now I've added the mixed content.
However if the authors of the site would fix the issue, it would be a much better solution. Especially as it affects the register page where users enter passwords.

fuglede:
Definitely would. I didn't hear back from them, unfortunately.

PR #4726, opened 2016-05-09
[OpenBSD.org] Enable, add ^ lists www, remove https.

mnordhoff:
* Enable
* Add `^`, `lists` & `www`
* Remove `https` (no longer resolves)

Reported by someone on IRC.

They also mentioned that `cvsweb` may or may not be coming soon, but in any case it isn't here yet.


J0WI:
I think there are some more ftp mirrors, but those comments should be clear.

PR #4706, opened 2016-05-02
[Minecraft] Update ruleset

fuglede:
Both ^ and www now work. This also adds a comment about a broken subdomain.

This closes https://github.com/EFForg/https-everywhere/issues/4703

PR #4730, opened 2016-05-10
[Minecraft] Add subdomains

fuglede:
As a follow-up on https://github.com/EFForg/https-everywhere/pull/4706#issuecomment-217903213, this adds several subdomains as targets, adds comments on broken ones, and adds a separate ruleset for a subdomain subject to false mixed content blocking.


J0WI:
Awesome, thanks!

PR #4733, opened 2016-05-10
Add 1fichier.com and mirrors

IDKwhattoputhere:
Example link: `https://alterupload.com/?dkqfao9flj`
One of their domains (dfichiers.com) wasn't configured correctly (SSL_ERROR_BAD_CERT_DOMAIN) so I left it out.

### DIFF
--- a/src/chrome/content/rules/1fichier.xml
+++ b/src/chrome/content/rules/1fichier.xml
@@ -12,6 +12,10 @@
 	<target host="tenvoi.com" />
 	<target host="dl4free.com" />
 
+	<!-- not working:
+		dfichiers.com
+	-->
+
 	<rule from="^http:"
 		to="https:" />
 
--- a/src/chrome/content/rules/1fichier.xml
+++ b/src/chrome/content/rules/1fichier.xml
@@ -1,0 +1,18 @@
+<ruleset name="1fichier">
+
+	<target host="1fichier.com" />
+	<target host="*.1fichier.com" />
+	<target host="alterupload.com" />
+	<target host="cjoint.net" />
+	<target host="desfichiers.com" />
+	<target host="megadl.fr" />
+	<target host="mesfichiers.org" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the ruleset! Can you add some tests for the `*.1fichier.com` rule? The goal for rules with a left-hard wildcard is [three tests](https://github.com/EFForg/https-everywhere/blob/e8475a04c6116361f108d56158b5a05a767fb260/ruleset-testing.md). At least, please add a test for http://img.1fichier.com/ .

jeremyn:
Also, please add a comment at the top that all of the non-1fichier.com URLs are mirrors of https://1fichier.com/ .

IDKwhattoputhere:
Hard to use real example links for non-img.1fichier.com subdomains since the pages expire quickly.

jeremyn:
Test URLs should be http, not https. See [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md#example).

The tests just need to rewrite correctly, have no HTTPS errors, and not have different content than the corresponding HTTP page. They don't have to be interesting. It looks like three good test cases are `www`, `img`, and `st-1` (notice that `https://a.1fichier.com/` includes `https://st-1.1fichier.com/piwik.js` so `st-1` is probably a stable subdomain). You don't need anything after `.com/` in any of these cases.

Please squash everything into a single commit. Thanks!

jeremyn:
Thanks for the update. Merged.

By the way please don't worry about squashing. It's easier to review unsquashed updates.

IDKwhattoputhere:
> squashing

Sorry but what do you mean by that? Tried googling it which was a mistake.

jeremyn:
I meant rebasing to combine multiple git commits into one commit (see [here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)).

Issue #4718, opened 2016-05-08
Add britishairways.com


jbrowngithubuser:
The homepage for British Airways defaults to HTTP so leaks username/password (users can login on the top right), even though it supports HTTPS. Please add britishairways.com and www.britishairways.com

https://www.ssllabs.com/ssltest/analyze.html?d=britishairways.com
https://www.ssllabs.com/ssltest/analyze.html?d=www.britishairways.com

J0WI:
The current rule needs to be updated: https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/BA.xml

We have a nice guide about how to do that here: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

illicitonion:
This got merged already :)

PR #4734, opened 2016-05-10
rebased BA.xml, fixing #4718

J0WI:


J0WI:
Alexa rank ~177 in UK, marking for next release.

PR #4737, opened 2016-05-10
fixed #4621

### LINKED ISSUE #4621: https://socialblade.com/
This is likely the biggest site with Youtube rankings. Please create and add this rule:

http://socialblade.com/ -> https://socialblade.com/

J0WI:


fuglede:
I'm seeing mixed content issues on https://socialblade.com/blog/ and https://socialblade.com/chat/.

J0WI:
@fuglede can you give me a hint what's wrong with the test urls?

fuglede:
It looks like it wants three things that are hit by the exclusion. Note that http://socialblade.com/blog/ is not actually excluded due to the final slash in the pattern. If you remove the slash, you should see the test counter increase to 1. The tests currently added do not add to the count since none of them match.

fuglede:
Looks like they're working. 👍

However, even with the exclusion, there's still a link on the front page to the broken https://socialblade.com/blog/. One way around that would be to force a [downgrade](https://github.com/EFForg/https-everywhere/blob/master/utils/downgrade-whitelist.txt). Another one to [hope that they fix it](https://twitter.com/fuglede/status/730481620533293057).

J0WI:
Of course they have to fix their mixed content issues. =P

fuglede:
Just had a quick look again; it seems like `/blog/` now works more than before, doesn't it? If so, we do not need to worry about downgrading.

J0WI:
updated.

Issue #4488, opened 2016-03-26
Add clashdaily.com


IDKwhattoputhere:
I had an attempt at this but it breaks the front page of the site.

```
<ruleset name="ClashDaily">

	<target host="clashdaily.com" />
	<target host="www.clashdaily.com" />

	<rule from="^http:"
		to="https:" />

</ruleset>
```
Any help is appreciated.

J0WI:
Locks good so far. Could you add a comment for the mismatching hosts?
cd.clashdaily.com
news.clashdaily.com
opinion.clashdaily.com

Please open a PR with your changes.

PR #4740, opened 2016-05-11
Add ClashDaily.com

IDKwhattoputhere:
https://github.com/EFForg/https-everywhere/issues/4488#issuecomment-218298393

### DIFF
--- a/src/chrome/content/rules/ClashDaily.xml
+++ b/src/chrome/content/rules/ClashDaily.xml
@@ -1,0 +1,15 @@
+<ruleset name="ClashDaily">
+
+    <target host="clashdaily.com" />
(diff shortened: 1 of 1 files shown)


fuglede:
As far as I can see, the server redirects https://(www.)clashdaily.com/anything to http://(www.)clashdaily.com/anything. If that is the case, then this rule will do nothing but create a redirect loop, unfortunately.

IDKwhattoputhere:
I see. The site didn't do that when I opened the issue. Quite unfortunate that they changed it. I guess PR and issue can be closed then?

fuglede:
I guess so; unfortunate indeed.

PR #4744, opened 2016-05-11
[Google Blogger] enable HTTPS for all Google Blogger blogs

J0WI:
Google finally allows HTTPS on all Google Blogger blogs:
https://security.googleblog.com/2016/05/bringing-https-to-all-blogspot-domain.html

This redirects all known Blogger hosts to HTTPS, including all Blogger subressources.

There might be some user side mixed content, but this is trivial in comparison of the huge coverage of this ruleset. As described in the blogpost, Google is already investigating.

This won't work for custom domains of course.

ping @fuglede @Hainish for review


Hainish:
For certain rulesets that we flag as important, we should probably have some infrastructure in place to ensure that when we run a full fetch test which disables rulesests, e.g., via

    python test/rules/src/https_everywhere_checker/check_rules.py test/rules/disable-broken-rulesets.checker.config

the entire rulesets do not get disabled.  For these rulesets, we should use @jsha's the more granular `golang` script, forked here: https://gist.github.com/Hainish/d535e23266295aad45bf5ee6e233a3a3

We should consider whether we want to do this with a whitelist or some other mechanism.

jsha:
I don't recommend adopting the golang script. There's nothing to
recommend it above the Python script. It introduces a new language into
a codebase that already has JavaScript and Python (and Bash). And it
lack a lot of features of the Python script like understanding test tags
and exclusions.

If we're investing in new directions in improving the fetch testing,
they should be in JavaScript, either in the browser or via PhantomJS.

PR #4151, opened 2016-02-15
torrent related sites

galeksandrp:
- Exceptions
	- PimpMyuTorrent
	- DuckieTorrent
	- DuckieTV
- Deletes
	- torque.bittorrent.com

PR #4748, opened 2016-05-11
[Github-Pages.xml] added exclusions from #4151 and fixed #4372

### LINKED ISSUE #4372: GitHub Pages rule breaks site
http://ibotpeaches.github.io/Apktool/

Site will reload repeatedly when the GitHub Pages rule is active.

J0WI:


src/chrome/content/rules/Github-Pages.xml
```diff
@@ -5,6 +5,28 @@
...
+	<exclusion pattern="^http://ibotpeaches\.github\.io/Apktool/" />
+	<exclusion pattern="^http://schizoduckie\.github\.io/DuckieTorrent/" />
+	<exclusion pattern="^http://schizoduckie\.github\.io/PimpMyuTorrent/" />
+	<exclusion pattern="^http://leaflet\.github\.io/" />
+
+	<test url="http://duckietv.github.io/DuckieTV/" />
+	<test url="http://ibotpeaches.github.io/Apktool/" />
+	<test url="http://schizoduckie.github.io/DuckieTorrent/" />
+	<test url="http://schizoduckie.github.io/PimpMyuTorrent/" />
+	<test url="http://leaflet.github.io/Leaflet.markercluster/example/marker-clustering-realworld.388.html" />
```
danzel: The leaflet.github.io rule can be removed now, it has been corrected to work.

J0WI: When greping though the repo I can still find some hardcoded `http:` links, but the initial issue seems to be fixed.

J0WI:
ping @fuglede for review

src/chrome/content/rules/Github-Pages.xml
```diff
@@ -5,6 +5,26 @@
 
 	<target host="*.github.io" />
 
+
+	<!-- pages that require http -->
+	<exclusion pattern="^http://duckietv\.github\.io/DuckieTV/" />
```
fuglede: A couple of issues here:

* http://duckietv.github.io/DuckieTV (without the trailing slash) still redirects to the HTTPS version.
* I'm actually seeing some CORS issues on http://duckietv.github.io/DuckieTV/ that may or may not be related:

skaermbillede fra 2016-06-16 22 47 28

J0WI:
Added exclusion also without trailing slash.
I think the CORS error is not related, I can't see any difference with HTTPS Everywhere disabled.

fuglede:
Ruleset LGTM, but what is the idea behind taking the exceptions from #4151; that that one had unrelated changes at the time of the PR?

J0WI:
I didn't expect an update on #4151 in near time and picked the bug fixes here.
It doesn't matter which one gets merged as long as they are still about the same.

PR #4753, opened 2016-05-12
Simplified ruleset for marktplaats.nl

fathorse:
Simplified ruleset that only handles item pages and categories (searches are now downgraded to http). Due to marktplaats.nl undergoing major reconstruction I would argue that a couple of robust core rules is better than no ruleset at all.

### DIFF
--- a/src/chrome/content/rules/marktplaats.nl.xml
+++ b/src/chrome/content/rules/marktplaats.nl.xml
@@ -1,9 +1,13 @@
-<ruleset name="Marktplaats (buggy)" platform="mixedcontent" default_off="Mixed content block breaks advert editing and certain images">
-  <target host="*.marktplaats.nl" />
-  <target host="*.marktplaats.com" />
-  <target host="*.marktplaats.net" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the update! I can review this, but can you please fix these errors first?

```
ERROR src/chrome/content/rules/marktplaats.nl.xml: No rule or exclusion applies to test URL http://www.marktplaats.nl/
ERROR src/chrome/content/rules/marktplaats.nl.xml: Not enough tests (1 vs 3) for <Rule from '^http://www\.marktplaats\.nl/a/([^?]*)\.html.*' to 'https://www.marktplaats.nl/a/\g<1>.html'>
ERROR src/chrome/content/rules/marktplaats.nl.xml: Not enough tests (0 vs 1) for <Exclusion pattern 'http://www\.marktplaats\.nl/unicorn'>
```

Please let me know if you have questions on how to fix them. Thanks.

J0WI:
@fathorse ping

PR #4755, opened 2016-05-12
Update 163.com.xml

renyouguo:


J0WI:
Slow host, rerun test

renyouguo:
I have to say, 163 and 126 is one of the fastest sites in china.
Slow in other county?Maybe.

J0WI:
It's the connection from China to Travis (US?) that is slow.

renyouguo:
Yes,there is the Great Fire Wall, so you will get the different errors for each time.

J0WI:
It finally passed all tests :)

PR #4756, opened 2016-05-12
Update and rename NetEase.com.xml to NetEase.xml

renyouguo:


renyouguo:
The check error has nothing todo with this PR, why?

fuglede:
My guess would be that it is because you are requesting a pull to <code>release</code>. Could you request one to <code>master</code> instead? That being said, the error that it actually catches is super curious, as the domain in question appears to work fine to me.

renyouguo:
#4766

PR #4757, opened 2016-05-12
Update 163.com.xml

renyouguo:


J0WI:
Are 163 and 126 related?

J0WI:
Just saw an error from your last PR:
Could you move `chat.study.163.com` to the falsemixed?
But you can keep `ct.stu.126.net` here, as long as this host is related with 163 (You may find a more generic title for the rule then?).

J0WI:
Thanks for all the fixes (see also #4765).

renyouguo:
@J0WI You have merged this. However, I can only find 163.com.xml in https://github.com/EFForg/https-everywhere/find/master .
Why?

J0WI:
GitHub has some problems to list that many files, just search for it or use CLI:
https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/163.com.xml

renyouguo:
@J0WI Oh, before i @ you, I have searched and found 163.com.xml.
What I mean is that I have "Update and rename 163.com.xml to WangYi.xml", and you have merged it, but it is original.
https://github.com/EFForg/https-everywhere/pull/4757/commits/fe8aa05586cf361a9122c9ffc2c29beca5686a48

J0WI:
See https://github.com/EFForg/https-everywhere/commit/42e1f34fd66cc70a35de90e2e374bd4a89c65dd4 by @loveithateit

renyouguo:
@J0WI Thanks.

Issue #3997, opened 2016-01-25
Craigslist.org (partial) ruleset breaks craigslist.org redirection


anonsubmitter:
The Craigslist.org (partial) ruleset prevents craigslist.org from redirecting to [helsinki.craigslist.fi](http://helsinki.craigslist.fi/), which doesn't support HTTPS. The failed redirection leads to a "This Connection is Untrusted" warning page.

It might be worth looking into if this is happening on other local craigslist sites as well.

lucaspetter:
Craigslist has a [big list](https://www.craigslist.org/about/sites) of various regional sites, but the only ones that support https are those on .org, .ca, and .co.uk domains. This is also shown in the certificate, which is only valid for .org, .ca, and .co.uk domains.

When going to craigslist.org with HTTPS Everywhere on, I get this redirect sequence:

1. http://craigslist.org/, which gets rewritten by HTTPSE to
2. https://www.craigslist.org/, which redirects to
3. https://geo.craigslist.org/, which redirects to
4. //vancouver.craigslist.ca/, which is equivalent to
5. https://vancouver.craigslist.ca/

The protocol-relative redirect in number 4 above looks like the source of your "Untrusted Connection" problem. I'm guessing the [geo.craigslist.org](https://geo.craigslist.org/) server probably gives a protocol-relative redirect every time, even if the destination doesn't support https, as with [helsinki.craigslist.fi](http://helsinki.craigslist.fi/). If that's true, then this issue affects all non-.org/.ca/.co.uk regions. But since I'm in Canada, where Craigslist _does_ support https, I can't verify any of that.

**Could you (or anyone else outside the *.craigslist.org, .ca, and .co.uk regions) check whether https://geo.craigslist.org/ gives a protocol-relative redirect?** If so, then we'll know what the problem is for sure.

J0WI:
I can confirm this behavior.
We should implement a downgrade rule for geo.craigslist.org.

lucaspetter:
This issue is fixed in #4771

lucaspetter:
Travis says there's an error:
```
failure: non-whitelisted downgrade rule in Craigslist.org (partial) redirects to http.
failure: /home/travis/build/EFForg/https-everywhere/utils/../src/chrome/content/rules/Craigslist.org.xml failed test: Rule redirects to something other than https.
```

How do you whitelist a downgrade?

PR #4759, opened 2016-05-12
[Craigslist.co.uk] New Ruleset

lucaspetter:


J0WI:
Could you make a comment to see also [Craigslist.org.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Craigslist.org.xml)?
Please do the same for #4720.
On `Craigslist.org.xml` should be a comment for both all them.
`Microsoft.xml` or `GoogleServices.xml` should be good examples for what I mean.

You may have a look at #3997.

lucaspetter:
Okay, the "see also" comments are done. The one for Craigslist.ca.xml is in the separate pull request.

#3997 can be treated separately, I've added a comment there with some new info.

PR #4080, opened 2016-02-03
[Root.cz] New ruleset

oskar456:
Root.cz recently enabled HTTPS on their servers.

PR #4760, opened 2016-05-12
[DoesNotWork.eu] Added new target time.doesnotwork.eu

Roman3349:


Roman3349:
This PR has same problem in tests as PR #4080.

PR #4761, opened 2016-05-12
Modify url for checker to behave more like uri.href behaves in js

Hainish:
When running the `https_everywhere_checker` fetch test, I noticed that some rule exceptions were not being matched when they should.  For example, in `rules/Lenovo.xml` there is an exception for `http://social.lenovo.com/`, but this exception is not matched for the test url `http://social.lenovo.com` since it lacks the trailing `/`.  This results in the fetch test attempting to fetch `https://social.lenovo.com`, which is not the intended behavior (this fails).

In javascript, this is not an issue, since when the URL is applied to the `RuleSets`, it is first modified so that the trailing `/` is always present.  Likewise, in our `node.js` scripts creating a `URIjs` `URI` object does the same.

Here are a few examples that should result in the same output, but don't:

https://github.com/Hainish/https-everywhere/blob/fix-checker-example/test.js
https://github.com/Hainish/https-everywhere/blob/fix-checker-example/test/rules/src/https_everywhere_checker/test.py

This PR modifies the URL passed to `apply` in `rules.js` so that the trailing slash is always present.

### DIFF
--- a/test/rules/src/https_everywhere_checker/rules.py
+++ b/test/rules/src/https_everywhere_checker/rules.py
@@ -1,4 +1,5 @@
 import regex
+from urlparse import urlparse
 
 class Rule(object):
 	"""Represents one from->to rule element."""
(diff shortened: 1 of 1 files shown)


Hainish:
Ready for review @jsha @cooperq

jsha:
We should instead modify our ruleset validation to ensure that test URLs always have a slash after the hostname, and fix up any broken rulesets.

PR #4763, opened 2016-05-13
Test formatting

Hainish:
As per @jsha's suggestion [here](https://github.com/EFForg/https-everywhere/pull/4761#issuecomment-218911406), I've added a test that ensures rule test url paths begin with a /


Hainish:
Ready for review @jsha

jsha:
The rules changes look good, but why did you implement the slash check in the https_everywhere_checker rather than trivial-validate?

Hainish:
@jsha a similar test which checks the formatting of rules was added to the `https_everywhere_checker`, so I figured this to be an appropriate place to put it.  I'm pretty agnostic where it goes, though.

PR #4766, opened 2016-05-13
Update and rename NetEase.com.xml to NetEase.xml

renyouguo:


src/chrome/content/rules/NetEase.xml
```diff
@@ -0,0 +1,18 @@
+<ruleset name="NetEase">
+
+	<target host="nos.netease.com" />
+	<target host="rev.da.netease.com" />
+	
+	<target host="netease.im" />
+	<target host="app.netease.im" />
+	<target host="bbs.netease.im" />
```
J0WI: I got mixed content issues here

src/chrome/content/rules/NetEase.xml
```diff
@@ -0,0 +1,18 @@
+<ruleset name="NetEase">
+
+	<target host="nos.netease.com" />
```
J0WI: I can find a lot of other hosts here and please always cover/document ^netease.com/www.netease.com.

src/chrome/content/rules/NetEase.xml
```diff
@@ -0,0 +1,31 @@
...
+		- xqn
+		- www.netease.com
+
+	MCB:
+		- bbs.netease.im
+-->
+
+<ruleset name="NetEase">
+
+	<target host="netease.com" />
```
J0WI: Mismatch (see Travis)

PR #4769, opened 2016-05-13
New Rule Sebastian.Expert

SebastianWojciechowski:


J0WI:
Can you also document the mismatch on `www.sebastian.expert`?
If you are the owner of this host, you may be interested in [preloaded HSTS](https://hstspreload.appspot.com/]) which makes this rule obsolete.

SebastianWojciechowski:
hi, sorry for the delay. What mismatch with www.sebastian.expert? where is it?

J0WI:
It seems that is not yet preloaded, could you add the comment documenting the mismatch?
I'll merge it then.

jeremyn:
https://sebastian.expert is preloaded now. https://www.sebastian.expert still gets a mismatch, but we shouldn't add a rule just to document a mismatch.

This ruleset isn't needed now so I'm closing this pull request.

PR #4774, opened 2016-05-14
Update IdQQimg.com.xml

renyouguo:


J0WI:
Could you please comment that `idqqimg.com` and `www.idqqimg.com` doesn't exist?

PR #4778, opened 2016-05-14
Create geetest.xml

renyouguo:


src/chrome/content/rules/geetest.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="geetest">
+
+	<target host="geetest.com" />
+	<target host="api.geetest.com" />
+	<target host="front.geetest.com" />
+	<target host="user.geetest.com" />
+	<target host="static.geetest.com" />
```
J0WI: account.geetest.com
my.geetest.com
old.geetest.com
static.geetest.com

This rule requires the ` platform="mixedcontent"` because of all the mixed content issues.

J0WI:
This still misses the `account` subdomain.
I hope #4866 will be merged soon to fix the timeout false positives.

PR #4780, opened 2016-05-14
add mmbiz.qpic.cn

crackevil:


jeremyn:
Thanks for the ruleset, however https://mmbiz.qpic.cn uses an invalid certificate for the domain so we shouldn't rewrite to it. If you want, you can note this as a problematic domain with the reason in the comment at the top of the ruleset.

PR #4784, opened 2016-05-14
add rules for dxy.cn

crackevil:


jeremyn:
Thanks for the ruleset! This is a complicated set of domains.

Please add targets for `dq.dxy.cn`, `file.dxycdn.com`, `img.dxycdn.com`, and `res.dxycdn.com`.

I see a lot of subdomains like http://ophth.dxy.cn/ . Most of these are broken with HTTPS because of mixed content, but some are broken for other reasons like https://blog.dxy.cn/ which simply can't connect. Can you add a comment at the top of the ruleset mentioning the many subdomains along with their problems? See [here](https://github.com/EFForg/https-everywhere/search?utf8=%E2%9C%93&q=mixed+content&type=Code) for commenting examples. Please list all the domains individually if possible.

PR #4785, opened 2016-05-14
Create Pixiv.xml

renyouguo:


J0WI:
Please include or document [all known subdomains](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md):
```
If you're not sure what subdomains might exist, you can iteratively use google queries and enumerate the list of results like such:

    site:*.eff.org
    site:*.eff.org -site:www.eff.org
    site:*.eff.org -site:www.eff.org -site:ssd.eff.org

... and so on.
```

J0WI:
I found a few more:
embed.pixiv.net
mx.pixiv.net
genepixiv.pr.pixiv.net
kurofune.pr.pixiv.net
pool.pr.pixiv.net
help.pixiv.net
inside.pixiv.net
imgaz.pixiv.net

www.pximg.net
pixiv.co.jp
serv.ads.pixiv.org


If you have that much comments you should remove the xml overhead and place them on the top of the rule, grouped by errors (see [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/cesky-hosting.cz.xml) for an example).

J0WI:
Could you also check your other outstanding PRs and add subdomains and documentation there?
Just ping me and I'll have a look then.

src/chrome/content/rules/Pixiv.xml
```diff
@@ -0,0 +1,49 @@
...
+	<target host="i3.pixiv.net" />
+	<target host="i4.pixiv.net" />
+	<target host="img-comic.pixiv.net" />
+	<target host="img-novel.pixiv.net" />
+	<target host="novel.pixiv.net" />
+	<target host="sensei.pixiv.net" />
+	<target host="sketch.pixiv.net" />
+	<target host="source.pixiv.net" />
+	<target host="www.pixiv.net" />
+	<exclusion pattern="^http://www\.pixiv\.net/(?!favicon\.ico)" />
```
J0WI: Please document why

src/chrome/content/rules/Pixiv.xml
```diff
@@ -0,0 +1,49 @@
...
+	- kurofune.pr.pixiv.net
+	- m.pixiv.net
+	- mx.pixiv.net
+	- pool.pr.pixiv.net
+	- recruit.pixiv.net
+	- times.pixiv.net
+	- serv.ads.pixiv.org
+-->
+
+<ruleset name="test">
```
J0WI: please use a clear name, related to the hosts it redirects.

src/chrome/content/rules/Pixiv.xml
```diff
@@ -0,0 +1,49 @@
...
+
+	<target host="accounts.pixiv.net" />
+	<target host="comic.pixiv.net" />
+	<target host="factory.pixiv.net" />
+	<target host="i1.pixiv.net" />
+	<target host="i2.pixiv.net" />
+	<target host="i3.pixiv.net" />
+	<target host="i4.pixiv.net" />
+	<target host="img-comic.pixiv.net" />
+	<target host="img-novel.pixiv.net" />
```
J0WI: See also imgaz.pixiv.net

renyouguo: @J0WI 
At line4

src/chrome/content/rules/Pixiv.xml
```diff
@@ -0,0 +1,53 @@
...
+	<target host="source.pixiv.net" />
+	
+	<target host="d.pixiv.org" />
+	
+	<target host="pximg.net" />
+	<target host="www.pximg.net" />
+	
+	<!--	Redirect to http most of www.*	-->
+	<target host="www.pixiv.net" />
+	<exclusion pattern="^http://www\.pixiv\.net/(?!favicon\.ico|login\.php)" />
```
J0WI: This requires some test urls

J0WI:
@fuglede see https://travis-ci.org/EFForg/https-everywhere/jobs/133940052#L368
Do you know what this error means? Would it be a good thing to have a list of error codes that will be retried by the Python script?

fuglede:
@J0WI: The script uses `curl` to perform all HTTP requests, so its error messages will be those from `curl`; in this case [this one](https://stackoverflow.com/questions/1924434/curl-error-52-empty-reply-from-server).

src/chrome/content/rules/Pixiv.xml
```diff
@@ -0,0 +1,55 @@
...
+	<test url="http://www.pixiv.net/favicon.ico" />
+	<test url="http://www.pixiv.net/login.php" />
```
J0WI: You also need test urls that are excluded (not redirected).

renyouguo:
@J0WI Would you set the check auto reopen while 1 ok and 1 error ?
Five time, tired.

J0WI:
I think there is no way to do this, is there?
cc @Hainish

Hainish:
I know of no way to do this auto-retry when we have one failure and one success.

One way reduce the amount of false negatives would be to retry URLs that fail for whatever reason.  We could see if it succeeds after 1 or 2 retries.

Hainish:
https://github.com/EFForg/https-everywhere/pull/5015 should result in fewer false negatives

PR #4787, opened 2016-05-14
[JW_Platform.com.xml] fixed #4531

### LINKED ISSUE #4531: JW Platform.com (partial) breaks video
http://dailycaller.com/2016/03/30/jane-sanders-hillary-tries-to-get-votes-from-people-at-the-worst-time-in-their-life-video/

JW Platform.com (partial) breaks video

J0WI:
also added api.jwlatform.com


Hainish:
thanks @J0WI!

PR #4792, opened 2016-05-14
update Jabber.at.xml

J0WI:


fuglede:
A wildcard for the subdomains would also work here, given that the thing is already HSTS preloaded in some browsers.

J0WI:
Oh sure, we really need a test for that.

fuglede:
For catching preloadables, you mean? And if so, what should it do? In this case, preloadability has not yet descended all the way through to ESR, so current consensus appears to be that it should be treated on the same footing as all other rulesets.

What I could see being useful would be to have a setup similar to what we use on Wikipedia where general maintenance scripts take care of things that we like but probably don't want to check on every single change; I'm thinking stuff like killing completely preloaded target hosts (which I'm still not certain I'm personally in favor of), enforcing a particular ruleset style, running https-everywhere-checker, etc.

J0WI:
The test should check if a host sends the HSTS preload attribute and if the hosts is in the list.
The list gets uplifted to ESR and TOR, but the list on SSL Labs is not up to date (TOR lacks a stable branch where the list can be pulled from automatically).

So we both agree that we don't want to maintain HSTS rulesets. A script could generate rules automatically during the build process out of the list (so they don't need to be in the repo).
But I'm also fine if they would just be deleted. Because the preloaded list of browsers if has a much better performance than HTTPS Everywhere and is not affected by false mixed content blocking issues. It's also well maintained by browser vendors and webmasters.

fuglede:
But how would we check if a host is in all of the relevant lists (without running in to the same problem as SSL Labs)?

J0WI:
I think the list of Chrome and Firefox (if there are any differences) are enough, because all other are based on them.
I don't know if the list of Edge is public available.

J0WI:
#4462 looks interesting btw.

fuglede:
Based on them, but [as I see that you are already aware](https://github.com/ssllabs/ssllabs-scan/issues/317), only actually including them at some point in the (potentially far) future.

PR #4793, opened 2016-05-14
update IchKoche.at.xml

J0WI:


jeremyn:
Please add a target for `ichkoche.at`.

Also I notice that some of the cookies don't have the `isSecure` flag set, such as `https://www.ichkoche.at`-`xtrvn`. I can confirm that the ruleset is changing `isSecure` for some of the cookies though. Do you know why it's not getting all of them? Maybe a bug?

J0WI:
I'm trying a more generic regex now.

jeremyn:
Looks good, thanks.

PR #4794, opened 2016-05-14
update irc.lc.xml

J0WI:


fuglede:
This (and the original ruleset) causes mixed content blocking on all pages with the IRC client; for example

* https://dev.irc.lc/report
* https://irc.lc/report
* https://irc.lc/quakenet/diablo3

Moreover, the reference to inframonde.qwebirc.net is not securable.

J0WI:
I'll go with a simple `platform="mixedcontent"`.

fuglede:
That will not be sufficient, since inframonde.qwebirc.net is not secured by us.

J0WI:
Isn't this platform for users how don't care about mixed content warnings?

fuglede:
Hm, you might actually be right; I always used it in the more conservative sense of avoiding problems for browsers where [false mixed content blocking](https://bugzilla.mozilla.org/show_bug.cgi?id=878890) is an issue, but the [EFF page](https://www.eff.org/https-everywhere/rulesets#mixed-content) does not mention that issue:

> Some rulesets may trigger active mixed content (i.e. scripts loaded over HTTP instead of HTTPS). This type of mixed content is blocked in both Chrome and Firefox, before HTTPS Everywhere has a chance to rewrite the URLs to an HTTPS version. This generally breaks the site. However, the Tor Browser doesn't block mixed content, in order to allow HTTPS Everywhere to try and rewrite the URLs to an HTTPS version.

> To enable a rule only on platforms that allow mixed content (currently only the Tor Browser), you can add a platform="mixedcontent" attribute to the ruleset element.

To be honest, I also never understood why Tor Browser allows active mixed content in the first place; seems like a terrible practice on a tunnel where MITMing is more than likely to occur.

jeremyn:
[This comment](https://github.com/J0WI/https-everywhere/blob/25ab8c0030a8633c57dd81e0f451336b55ed4523/src/chrome/content/rules/irc.lc.xml#L3) should have `.org` instead of `.orgi`.

https://pad.irc.lc should be noted as `refused`.

Otherwise this looks fine to me. @fuglede what do you think?

J0WI:
pad.irc.lc is refused

fuglede:
@jeremyn: I think I managed to convince myself that @j0wi was right (but I didn't manage to convince myself that Tor Browser is not crazy in its defaults).

J0WI:
Without having confirmed this, it might be possible that the Tor Browser blocks mixed content scripts with NoScript.

fuglede:
Just tried it out with Tor Browser and NoScript, and the IRC client appears to load just fine.

PR #4801, opened 2016-05-15
Update and rename qiyi.com.xml to qiyi.xml

renyouguo:


J0WI:
Mixed content:
- vip
- www
- passport

Different content/404
- 2012
- labs
- list
- v
- yule

Refused:
- app
- so

Timeout:
- campus
- i



For the mixed content issues you need to add the `platform="mixedcontent"` attribute or you put them in a separate "falsemixed" file to have the other hosts still enabled for all (see other falsemixed files for an example).

J0WI:
Oh sorry, I mixed this with #4803

renyouguo:
All qiyi.com works for iqiyi.com .
Chinese sites cost long long long time to switch to https, and change them much time.
So, I do not think it is nessary to add mixedcontent or 404 web to iqiyi.com / qiyi.com .
They will change next time……

J0WI:
Ouch, that sounds horrible...
Could you make a short comment about that and also a reference to `iqiyi.com.xml`?
I think you should link to my comment above or copy the list in the rule that who ever is working on this rule next time can check the list.

PR #4803, opened 2016-05-15
Update iqiyi.com.xml

renyouguo:


J0WI:
I had to revert this:
`ERROR src/chrome/content/rules/iqiyi.com.xml: Not enough tests (1 vs 3) for <Exclusion pattern '^http://www\.iqiyi\.com/(?!favicon\.ico|player\/cupid\/common\/clear.swf)'>`

PR #4777, opened 2016-05-14
Update Qlogo.cn.xml

renyouguo:

PR #4805, opened 2016-05-15
add wx.qlogo.cn

crackevil:


renyouguo:
I have add this and others at #4777

PR #4807, opened 2016-05-15
[Lisonfan.com] Add ruleset

fuglede:
The site ships with a 301 redirect to HTTPS as well as the preload token but is not yet on any of the vendor preload lists. This closes https://github.com/EFForg/https-everywhere/issues/4804.

### DIFF
--- a/src/chrome/content/rules/Lisonfan.com.xml
+++ b/src/chrome/content/rules/Lisonfan.com.xml
@@ -1,0 +1,6 @@
+<ruleset name="Lisonfan.com">
+	<target host="lisonfan.com" />
+	<target host="www.lisonfan.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


J0WI:
Not sure if we should just wait until this host is preloaded by browsers.

fuglede:
These days, preload list updates are coming [quite slow](https://twitter.com/Scott_Helme/status/730148020646518786), and it looks like it can easily take several months for the new items to hit the Tor Browser preload list, so I think that it's a save bet that a new HTTPS Everywhere version is out before that happens.

J0WI:
Okay, this reasonable. But I still hope that we will find an easy solution for preloaded hosts.

fuglede:
Agreed. One thing we could do to have a consistent solution would be to simply batch-add all hosts that are either [preloaded in Chromium](https://chromium.googlesource.com/chromium/src/+/master/net/http/transport_security_state_static.json) or [pending preload](https://hstspreload.appspot.com/pending) but which have not yet seen the light in Tor Browser. Security-wise that might not add *that* much, but it would make our lives easier, and it might make using HTTP Nowhere a bit more pleasant.

Is https://gitweb.torproject.org/tor-browser.git/tree/security/manager/boot/src/nsSTSPreloadList.inc?h=tor-browser-38.6.1esr-6.0-1 the canonical place to look for an up-to-date Tor Browser HSTS preload list?

J0WI:
Yes, but as already exposed by Ivan there is no stable branch. You have to update to the current branch from time to time.
In [Firefox](https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsSTSPreloadList.inc) this is much easier, but unfortunately they switched to another format than JSON for performance reasons.
But this also means that Tor will got this new format in the next major release too.


See also:
#4792
https://github.com/ssllabs/ssllabs-scan/issues/317

fuglede:
Doing so would look [something like this](https://github.com/fuglede/https-everywhere/commit/c14915e3259e71aa8aab039b5f8c0127f34e43db) ...

J0WI:
This sounds great!
I would even include the wildcard for them, regardless of possible impact.

In #4462 they are all in one file called HSTS, but I think it doesn't make a difference as long as they are not in the repo and just added on build time.

PR #4808, opened 2016-05-15
Update imagecurl.com.xml

zero77:


J0WI:
Left wildcards requires at least 3 [test urls](https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md).
Rule LGTM.

zero77:
The tested https URLs I'm aware of currently are:
cdn.imagecurl.com
www.imagecurl.com
imagecurl.com
Does that meet the left wildcard requirements.

J0WI:
^imagecurl.com doesn't match the wildcard, but if you are sure that there are no more subdomains and that there will no unsecured subdomains in near future, you can workaround using a test domain like `http://www.imagecurl.com/foo` as the third one.

zero77:
yes they are all the subdomains I'm currently aware of although i don't know about the future.
Whats the time period be for it will be live within the add on.

PR #4812, opened 2016-05-16
added Otto.de.xml

J0WI:


jeremyn:
Please add `pxc.otto.de` and `ats.otto.de`.

PR #4813, opened 2016-05-16
added Gutefrage.net.xml

J0WI:


J0WI:
Regarding the stats on alexa and the fact, that this popular host doesn't use https by default, it's worth to have this in next release.

fuglede:
The exclusion pattern is non-functional since www.gutefrage.net is not among the target hosts; I think you intended to add that one as well?

Some other resources, besides /js/, that appear to be securable, are /images/, /login, /serviceapi, and /favicon.ico.

J0WI:
I skipped `serviceapi` becaue the host doesn't return a 2xx response.
It's odd that our tests require test urls that are not redirected.

fuglede:
Yeah, I'm not actually sure, what the logic behind that is (and I couldn't seem to find a description in the code after having had a brief look), but as you've probably also noted, they're not actually checked in the checker runs.

PR #4816, opened 2016-05-16
add rule for Teltarif.de

numismatika:
According to
www.teltarif.de/https-verschluesselung-sicherheit/news/63626.html
did the website introduce TLS for both mobile and desktop version of their site


J0WI:
Thanks for sharing this with us!

PR #4817, opened 2016-05-17
added advenacm.com.xml

CheapApples12:


J0WI:
Thanks for contributing!
I think this host is related to #4818, so could you please add a comment on both linking to each other?
You can see an example [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/GoogleBlogger.xml#L3).

Please also add the following hosts:
bootstrap.advenacm.com
repository.advenacm.com
support.advenacm.com

I can see some mixed content issues on `blog.advenacm.com`. I would be great if you could address them and include this host too. ([upgrade-insecure-requests](https://www.w3.org/TR/upgrade-insecure-requests/) is a quick way to mitigate.)
If this is not possible, you should add a comment about the issue, that a future maintainer is aware of it.

CheapApples12:
Hi J0WI,

That is correct, this host is closely related to #4818 and I have added a comment on both as you asked. I've also added the above hosts to #4817 as well as blog.advenacm.com which I was able to add due to mitigation of the mixed content issues after following the article you provided on upgrade-insecure-requests.

Thank's for allowing me to be apart of this amazing extension and hopefully, together we can make the internet a safer place!

J0WI:
Please see comments in #4818

J0WI:
@CheapApples12 would be great to have this small comments fixed, so I can merge both. :)

jeremyn:
@CheapApples12 Are you still interested in working on this?

CheapApples12:
@jeremyn We've changed our domain from advenam.com to advenacs.com. I'll reflect the changes in the pull request :)

CheapApples12:
Scratch that.. I'll make a new pull request :)

PR #4818, opened 2016-05-17
added advena.io.xml

CheapApples12:


src/chrome/content/rules/advena.io.xml
```diff
@@ -1,7 +1,15 @@
...
+	advena.io and advenacm.com are owned by the same company, but are separate
+	websites, no aliases. Advenacm.com's https-everywhere rule can be found at
+	https://github.com/EFForg/https-everywhere/pull/4817
+-->
+
 <ruleset name="Advena">
         <target host="advena.io" />
         <target host="www.advena.io" />
         <target host="my.advena.io" />
+        <target host="*.advena.io" />
```
J0WI: Wildcard targets requires at least 3 test urls and would make the two above obsolete.

src/chrome/content/rules/advena.io.xml
```diff
@@ -1,7 +1,15 @@
+<!--
+	Advena.io is the primary content serving domain of the Advena CM network.
+	advena.io and advenacm.com are owned by the same company, but are separate
+	websites, no aliases. Advenacm.com's https-everywhere rule can be found at
+	https://github.com/EFForg/https-everywhere/pull/4817
```
J0WI: Please use filenames instead of PR links (same on #4817).

Hainish:
source repo deleted

PR #4819, opened 2016-05-17
Create Bandisoft.xml

renyouguo:


src/chrome/content/rules/Bandisoft.xml
```diff
@@ -0,0 +1,22 @@
+<!--	Broken
+	- apps.
```
J0WI: Please clarify to 'refused'

src/chrome/content/rules/Bandisoft.xml
```diff
@@ -0,0 +1,22 @@
...
+	- blog.
```
J0WI: Please clarify to 'mismatch'

PR #4823, opened 2016-05-17
Update BuzzFeed.xml

J0WI:
[BuzzFeed](https://www.buzzfeed.com/jasonreich/buzzfeed-and-https) announced official HTTPS support so I updated the ruleset.


J0WI:
Note: They send preloaded HSTS header, but not on ^buzzfeed.com, with a too short duration and without subdomains.

jeremyn:
I've seen this on a bunch of rulesets, what does it mean? (for cookies) `Not secured by server`

J0WI:
Cookies should always be set with the HttpOnly and the Secure flag. This means that Cookies can only be read by the server over https. Otherwise Cookies can be read over an insecure connection or over JavaScript, which means that e.g. session can be hijacked.

jeremyn:
But what specifically does someone mean with `Not secured by server`? Why is this specially noted? For example we don't include `<!-- Allows HTTP connection -->` before every single `<target>` tag.

J0WI:
That's maybe to give a hint that insecure Cookies are send over https, which is a really bad practice.
I didn't make any thought of this, but this comment is likely equal to the presence of a securecookie rule.
We also had some discussions about the need of securecookie rules, because if all connections are rewritten to https anyway you can't SSLstrip it anyway.

jeremyn:
- [ ] Why are CDN buckets [noted](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L2) at all?
- [ ] `resources` is misspelled [here](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L57).
- [ ] I think this is a better test for [here](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L60): `<test url="http://s3-ec.buzzfed.com/static/" />`
- [ ] Can [this securecookie tag](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L65) be the more generic and preferred `<securecookie host=".+" name=".+" />` ?
- [ ] Can you add a comment for why [this rule](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L70-L71) exists? Also `s3-ec.buzzfed.com` has no DNS record.
- [ ] [This rule](https://github.com/J0WI/https-everywhere/blob/058666ed1076a45b0a6be171b14d5ff3ec25cb85/src/chrome/content/rules/BuzzFeed.xml#L72-L73) doesn't seem to get activated. Missing `target`?
- [ ] Missing `s.buzzfed.com`, `s3-ak.buzzfed.com`. `shop.buzzfeed.com`.
- [ ] Note `pglb.buzzfed.com` as mismatched.
- [ ] `ct.buzzfeed.com` is not mismatched
- [ ] Rewrite http://blog.buzzfeed.com to https://www.buzzfeed.com/blog (?).
- [ ] `get.buzzfeed.com` is refused, not timeout.

J0WI:
Thanks for reviewing. :)

> I think this is a better test for here: `<test url="http://s3-ec.buzzfed.com/static/" />`

This would not give a valid response

> get.buzzfeed.com is refused, not timeout.

Request still times out here.

jeremyn:
You removed `s3-ec.buzzfed.com` from the ruleset entirely ([1](https://github.com/EFForg/https-everywhere/pull/4823/commits/0621170f9c9fdbf192af2f0af799fd29c9bc0228#diff-ecc3edcf82bf190d211841fdfa794838L55), [2](https://github.com/EFForg/https-everywhere/pull/4823/commits/0621170f9c9fdbf192af2f0af799fd29c9bc0228#diff-ecc3edcf82bf190d211841fdfa794838L60)) so we don't need a test for it now.

I'm getting a timeout for https://get.buzzfeed.com now so that's fine.

jeremyn:
This should fix the tests:

- [ ] [This rule](https://github.com/J0WI/https-everywhere/blob/e70238ca1c3a75b14c7e050d6701f9c1915c563b/src/chrome/content/rules/BuzzFeed.xml#L61-L62) needs an extra escaping `\`.
- [ ] [This target](https://github.com/J0WI/https-everywhere/blob/e70238ca1c3a75b14c7e050d6701f9c1915c563b/src/chrome/content/rules/BuzzFeed.xml#L32) should be `shop.buzzfeed.com`, not `shop.buzzfed.com`.

PR #4827, opened 2016-05-18
Added ruleset for <collyers.ac.uk>

dylmye:
Ruleset containing comments. Tested on latest FF, Windows 10.


dylmye:
Tested as noted on the commit. Can remove the comments, obviously.

dylmye:
Note: seems like the Travis CI fails on local links (eg rocket.collyers.ac.uk) - I can confirm both that the cert is a wildcard for *.collyers.ac.uk, and that on the intranet https is available.

fuglede:
Non-internet-facing sites are not currently supported (the PR could be merged, but the ruleset would automatically be disabled in the next database-wide run).

One thing you could do to have the rewrite of the local subdomains work on your own machine would be to put it in the [Firefox profile folder](https://www.eff.org/https-everywhere/rulesets#testing) (if you use Firefox).

dylmye:
@fuglede please see updated commit

fuglede:
Thanks for the update! 👍

Issue #4770, opened 2016-05-13
Videos on golem.de don't work


gossi:
Hey guys,

just made a fresh firefox installation and added 4 addons:
- AdblockPlus
- Ghostery
- PrivacyBadger
- HTTPS everywhere

No further settings. Videos on golem.de don't work with default https-everywhere settings, see here: http://www.golem.de/news/traceroute-wann-ist-ein-nerd-ein-nerd-1605-120903.html

Hope you can fix that.

J0WI:
I can't reproduce this whits only HTTPS Everywhere installed.
Please check if the issue depends on a third party extension.

CheapApples12:
It's got to be a third-party extension causing the issues. I can't reproduce any errors and the only extensions I have are HTTPS Everywhere, Authy and Gyazo

gossi:
I have the following addons installed:

- Adblock Plus
- Ghostery
- HTTPS Everywhere
- Privacy Badger

I deactivated all but https everywhere. Restart firefox, a lot of ads but no video. What I did to fix it for me is locate my profile folder (`~/Library/Application Support/Firefox/Profiles/<myid>.default/HTTPSEverywhereUserRules` on OSX) and created the `Golem.de.xml` file in there with the following contents:

```xml
<ruleset name="Golem.de">

		<!--	Direct rewrites:
					-->
	<target host="cpxl.golem.de" />
	<target host="cpx.golem.de" />
	<target host="account.golem.de" />
	<target host="suche.golem.de"/>
	<!--<target host="video.golem.de" />-->
	<!-- whitepaperdb.golem.de expired-->
	<!-- TLS not supported on golem.de and ads.golem.de -->

	<rule from="^http:" to="https:"/>

</ruleset>
```

Which got me video.

fuglede:
I'm seeing the same issues; on the surface (plowing through the Firefox network log), it looks a lot like some of the [crossdomain.xml](https://github.com/EFForg/https-everywhere/issues?utf8=%E2%9C%93&q=crossdomain.xml) issues that we have had before, but I did not manage to find a nice solution. #4829 contains the same fix that you are using.

J0WI:
I just found out how to reproduce this:
- If Flash Player is _installed_, Golem will force Flash and break HTTPS.
- If Flash Player ist _not installed_, Golem will use HTML5 video and it works fine.

Not that click to play won't work, because Flash is still forced then.

fuglede:
Hah, good catch!

[...]

PR #4829, opened 2016-05-18
[Golem.de] Disable subdomain video

fuglede:
This fixes https://github.com/EFForg/https-everywhere/issues/4770. It would be nice of course to disable only part of the domain, but I couldn't find an easier way out of this one. In particular, the following attempts did not solve the problem:

* Exclude all js/swf/xml on the domain but leave the rest intact.
* Exclude everything but the "files/" subfolder (which contains
  the relevant videos).

### DIFF
--- a/src/chrome/content/rules/Golem.de.xml
+++ b/src/chrome/content/rules/Golem.de.xml
@@ -1,3 +1,15 @@
+<!--
+	Problematic domains:
+
+		- ^ ¹
+		- ads ¹
+		- video ³
+		- whitepaperdb ²
+
+	¹: Refused
+	²: Expired
+	³: Breaks video (cf. GH #4770).
+-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Golem.de.xml
```diff
@@ -1,14 +1,23 @@
...
+	²: Expired
+	³: Breaks video (cf. GH #4770).
+-->
 <ruleset name="Golem.de">
 
 		<!--	Direct rewrites:
 					-->
 	<target host="cpxl.golem.de" />
 	<target host="cpx.golem.de" />
 	<target host="account.golem.de" />
```
ridwande: <target host="video.golem.de" />
<target host="ads.golem.de" />

PR #4529, opened 2016-04-01
[Cloudfront] Attempt to fix rotten tomatoes video playback

funkydude:
Broken trailer (video) on http://www.rottentomatoes.com/m/the_revenant_2015/
The source `d3biamo577v4eu.cloudfront.net` seems to support HTTPS just fine but I believe rotten tomatos is trying to force it over HTTP (I couldn't locate the actual source), so I'm excluding this rule in an attempt to resolve it.

Did a search and this was actually reported last year: https://github.com/EFForg/https-everywhere/issues/2742

PR #4830, opened 2016-05-18
Remove redundant ruleset whitelist entries

fuglede:
This removes from the ruleset whitelist all those filenames that either

* no longer exist, or
* have no matching sha256sum,

which roughly cuts the whitelist in half.

The script used to achieve this is the following:

	#!/bin/bash

	# To be run from rules/

	WLIST=../../../../utils/ruleset-coverage-whitelist.txt
	WLISTFILES=$(cut -f3 -d " " $WLIST)

	for file in $WLISTFILES; do
	  # Remove those that no longer exist
	  if [ ! -f $file ]; then
	    sed -i "/ $file/d" $WLIST
	    echo "$file no longer exists"
	  # Remove those whose hashes no longer match
	  elif ! grep -q $(sha256sum $file) $WLIST; then
	    sed -i "/ $file/d" $WLIST
	    echo "$file no longer matches the whitelist hash"
	  fi
	done

I realize that this will miss those files that are duped in the whitelist, and where at least one of the dupes have the proper <code>sha256sum</code>, but those will eventually be killed off anyway.


J0WI:
Great!
This will also fix the outstanding issue to merge #4529.

@Hainish could you have a second look here? This shouldn't break anything.
Script need to be rerun before merge.

Hainish:
I'll look today

Hainish:
This is great, thanks for doing this @fuglede (I was actually planning on getting to this at some point, but you made my job easier :smile:)
Reviewing now.

Hainish:
@fuglede I would change this in a few ways.  Add the script itself to `utils` and ensure it runs from the `git root` as we do in `test.sh`, in the following way:

```shell
if [ -n "$GIT_DIR" ]
then
    # $GIT_DIR is set, so we're running as a hook.
    cd $GIT_DIR
else
    # Git command exists? Cool, let's CD to the right place.
    git rev-parse && cd "$(git rev-parse --show-toplevel)"
fi
```

Lastly, rebase from `master`.

fuglede:
Happy it's useful! Rebased, merged with the recent pile of ruleset changes, and removed the entries that are now redundant because of that one as well.

I did consider if it was worth adding a <code>utils/maintenance</code> folder; down the line you'll probably want something like that anyway (for doing database-wide checker runs, enforcing ruleset style, etc.)? For now, it's just in <code>utils</code>.

Hainish:
Great, verified results look the same on my local. Merging.

J0WI:
To prevent this in future we should make the tests fail when the hash of a listed file doesn't match anymore.

fuglede:
I don't see why; if it no longer matches, the whitelist entry should be functionally useless, and it would raise the bar for new contributors by a good amount.

J0WI:
Just as a hint that you should remove it from the list or update the hash sum (use case in #4529 or #4903).

fuglede:
Yeah, not a big fan; the task is so easy to do programmatically that it would also just be a bit of a waste of people's time to enforce it.

Hainish:
I agree with @fuglede, it's already sometimes quite a barrier to contribute a meaningful ruleset, we shouldn't make it harder (plus this can be safely run perennially, no need to continuously keep it up to date)

PR #4832, opened 2016-05-19
Tor observatory integration (resolves #2719)

### LINKED ISSUE #2719: Stop connecting to check.torproject.org on every browser startup
Or at least offer the possibility to disable this checkup and assume the browser isn't (or is) using tor.
This is phoning home and that doesn't make sense for a privacy extension.

Hainish:
This fixes a few separate but related issues with how we test proxy settings for the SSL observatory.

Commit https://github.com/EFForg/https-everywhere/commit/6de406768a56ad5dabda4b6f614c786d1a0805a1 resolves the edge case in https://github.com/EFForg/https-everywhere/issues/2719#issuecomment-219892166 where a user who has set a custom proxy will be making outbound requests to `check.torproject.org`.

The second issue, discovered while researching the first, is explained here: https://github.com/EFForg/https-everywhere/issues/2719#issuecomment-220182600.  Within the Tor Browser, we attempt to discover if the user can make outbound requests over Tor before the circuit is actually established.  This is technically a race condition, but since it will always take longer for the circuit to be established than to load the observatory code, this check always fails.  https://github.com/EFForg/https-everywhere/commit/c7df15106a049159ee5561cd0ee61c7efb82256b resolves the issue by only checking proxy settings [when the browser window and all its components have been loaded and initialized.](https://developer.mozilla.org/en-US/docs/Observer_Notifications)

### DIFF
--- a/src/components/ssl-observatory.js
+++ b/src/components/ssl-observatory.js
@@ -176,7 +176,7 @@ SSLObservatory.prototype = {
 
   findSubmissionTarget: function() {
     // Compute the URL that the Observatory will currently submit to
-    var host = this.prefs.getCharPref("extensions.https_everywhere._observatory.server_host");
+    var host = this.myGetCharPref("server_host");
     // Rebuild the regexp iff the host has changed
     if (host != this.submit_host) {
       this.submit_host = host;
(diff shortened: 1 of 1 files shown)


Hainish:
Ready for review @jsha

Hainish:
or @cooperq

cooperq:
lgtm

PR #4835, opened 2016-05-19
Create Dota2lounge.com.xml

ekaris:


fuglede:
The site blocks necessary mixed content from http://cdn.dota2lounge.com/css/bright.min.css?1456821822 (which can not be secured).

ekaris:
What should I do?

ekaris:
@J0WI What should I do?

J0WI:
You can rewrite `cdn.dota2lounge.com` to `dota2lounge.com` and test if this works in all cases.
Please also document this issue and also that `www.dota2lounge.com` does not exist (because usually it does).

ekaris:
Good?

src/chrome/content/rules/Dota2lounge.com.xml
```diff
@@ -1,15 +1,14 @@
 <!--
 	Non working:
 	www.dota2lounge.com
```
J0WI: Does not exist

src/chrome/content/rules/Dota2lounge.com.xml
```diff
@@ -1,15 +1,14 @@
...
-
 -->
 <ruleset name="Dota2lounge.com">
 	<target host="dota2lounge.com" />
 
-	<rule from="^http:" to="https:" />
-
 	<test url="http://cdn.dota2lounge.com/" />
```
J0WI: You  need to add this as a target host

ekaris:
Good?

fuglede:
This broke the site; the CSS will be blocked as mixed content even though we are securing it.

PR #4838, opened 2016-05-19
Create Wandoujia.xml

renyouguo:


src/chrome/content/rules/Wandoujia.xml
```diff
@@ -0,0 +1,34 @@
...
+	* uowechat.wandoujia.com
+
+	* t.wdjcdn.com
+	* img.wdjimg.com
+
+	* s.wdjimg.com
+	* ss.wdjimg.com
+-->
+
+<ruleset name="test">
```
J0WI: Please use a more specific name

src/chrome/content/rules/Wandoujia.xml
```diff
@@ -0,0 +1,34 @@
+<!--
+	Broken:
```
J0WI: Could you be more specific what breaks?

PR #4839, opened 2016-05-19
add hq.sinajs.cn

crackevil:


jeremyn:
There's another [sinajs.cn.xml ruleset](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/sinajs.cn.xml). Can you please modify this pull request to instead include `hq.sinajs.cn` in that file?

PR #4840, opened 2016-05-19
[Scienceathome.org] New ruleset

fuglede:


src/chrome/content/rules/Scienceathome.org.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	Note: All subdomains, other than the ones given as targets, appear
+              to redirect ^http://(.+)\.scienceathome\.org/ to
+                          https://$1.scienceathome.org/$1.
```
J0WI: I only found `new.` (which times out) and it redirects like `^http://new\.scienceathome\.org/ to https://www\.scienceathome.org/`.

fuglede: Right, what I meant here was that all

PR #4845, opened 2016-05-20
[Nyaa] Update rulesets

synthtech:
- [nyaa.se](https://www.nyaa.se) now supports https on all its subdomains
- nyaa.eu now redirects to nyaa.se

**Edit:** Added main domain, secured all cookies.

### DIFF
--- a/src/chrome/content/rules/Nyaa.eu.xml
+++ b/src/chrome/content/rules/Nyaa.eu.xml
@@ -1,23 +1,17 @@
 <!--
-Enable HTTPS on the nyaa.eu domain
+	nyaa.eu redirects to nyaa.se (See Nyaa.se.xml)
 
+	Self-signed, mismatch:
+		- cthuko.nyaa.eu
 -->
 <ruleset name="Nyaa.eu">
+	<target host="nyaa.eu" />
+	<target host="www.nyaa.eu" />
+	<target host="files.nyaa.eu" />
+	<target host="sukebei.nyaa.eu" />
 
-	<!--	Direct rewrites:
-				-->
-  <target host="nyaa.eu" />
-  <target host="www.nyaa.eu" />
-  <target host="sukebei.nyaa.eu" />
+	<!-- CloudFlare cookies: -->
+	<securecookie host="^\.nyaa\.eu$" name="^(?:__cfduid|cf_clearance)$" />
 
-	<!--	CloudFlare cookies:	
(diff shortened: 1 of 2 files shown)


J0WI:
This conflicts with some recent changes.

synthtech:
Well, I rebased it and kept the wildcards, though I have no idea why the build failed.

src/chrome/content/rules/Nyaa.eu.xml
```diff
@@ -1,23 +0,0 @@
-<!--
-Enable HTTPS on the nyaa.eu domain
-
```
J0WI: self-signed, mismatch: cthuko.nyaa.eu

J0WI:
Please keep the rule for `nyaa.eu`, because an access to nyaa.eu would be not secured then and there are also other content, that is not (yet) redirected.

Please add comments on both rules linking to each other and include the comment about the non-securable host on nyaa.eu.

The build should pass next time, the errors there has been fixed.

synthtech:
Ok, added the comments and readded the `.eu` rule.

Build seems to be failing because the `https://nyaa.se` to `https://www.nyaa.se` redirection hasn't been working the past couple of days.

J0WI:
The redirect still fails. You may add a redirect rule or comment it out.

synthtech:
Added a redirect rule.

src/chrome/content/rules/Nyaa.se.xml
```diff
@@ -10,15 +10,8 @@
...
-					-->
-	<!--securecookie host="^\.nyaa\.se$" name="^(?:__cfduid|cf_clearance)$" /-->
-
-	<securecookie host="^\.nyaa\.se$" name="^(?:__cfduid|cf_clearance)$" />
-
-
-	<rule from="^http:"
-		to="https:" />
-
+	<rule from="^http://nyaa\.se/" to="https://www.nyaa.se/" />
```
J0WI: Please also document this issue.

synthtech:
Added a comment explaining the issue and slightly modified the rule to redirect the https version of the main domain as well.

J0WI:
We do not redirect https, except a site uses malformed relative links.

synthtech:
Ok, reverted that edit.

J0WI:
lgtm now, Thanks for the additions!

PR #4852, opened 2016-05-20
Create iqdb

BlackHazel:


jeremyn:
Thank you for the ruleset! Please:

* Add a target for `3d.iqdb.org` .
* I see a lot of subdomains on https://iqdb.org/ , such as `(simple search)` http://anime-pictures.iqdb.org/ , that don't work because the certificate doesn't include that subdomain. Add a comment at the top of the ruleset mentioning that subdomains other than the two targets don't work. See [here](https://github.com/EFForg/https-everywhere/search?utf8=%E2%9C%93&q=problematic+domains&type=Code) for comment style examples.
* Squash your changes into one commit.

BlackHazel:
@jeremyn PR Updated. Hope the comment are good like that.

jeremyn:
That's an improvement but it's a little confusing to me. Can you please just replace [these lines](https://github.com/BlackHazel/https-everywhere/blob/df417321b019b30734a1ce859851027cb2a25e17/src/chrome/content/rules/iqdb.xml#L4-L9) with this, properly indented?

`- *.iqdb.org other than targets (domain mismatch)`

Also, please squash everything into one commit. Thanks again!

jeremyn:
Thanks!

PR #4853, opened 2016-05-20
Create voidtools.com.xml

renyouguo:


J0WI:
Could you also redirect `http://www.forum.voidtools.com/` to `https://www.voidtools.com/forum/`?
(You should double check if they are really the same.)

J0WI:
Of course you should comment that this is necessary because of the mismatch on `www.forum.voidtools.com`. Sorry that this was not clear.

renyouguo:
@J0WI What should I do? I cann't understand the check error.

src/chrome/content/rules/voidtools.com.xml
```diff
@@ -0,0 +1,16 @@
+<ruleset name="voidtools.com">
+
+	<!--	Mismatch on  www.forum.voidtools.com	-->
+
+	<target host="www.forum.voidtools.com"/>
+	<rule from="^http://www\.forum\.voidtools\.com/" to="https://www.voidtools.com/forum/" />
+		<test url="http://www.forum.voidtools.com/" />
+		
+`	<!--	Directly	-->
```
J0WI: Remove the grave accent here on col 1 to fix syntax errors.

PR #4857, opened 2016-05-20
Create criosweb.xml

renyouguo:


src/chrome/content/rules/criosweb.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="foobar2000">
```
fuglede: This looks like a typo.

src/chrome/content/rules/criosweb.xml
```diff
@@ -0,0 +1,10 @@
...
+
+	<target host="criosweb.ro"/>
+	<target host="infoeuri.criosweb.ro"/>
```
fuglede: This causes mixed content issues on https://infoeuri.criosweb.ro/blog/

src/chrome/content/rules/criosweb.xml
```diff
@@ -0,0 +1,10 @@
...
+	<target host="luci.criosweb.ro"/>
```
fuglede: Lots of blocked mixed content on https://luci.criosweb.ro/.

renyouguo:
@fuglede I have add 4 test rules. What can I do to fix the check error?

fuglede:
I'm not immediately sure, but perhaps it's because the [old catch-all rule](https://github.com/EFForg/https-everywhere/pull/4857/files#diff-aab10da1e4948ff8f4183fb2b3bfc342R6) was left in, so that the second rule is never actually applied, so that one would have to go.

I also noticed -- but that should have no influence on the test count -- that the test URLs do not actually point at existing resources. For instance, http://luci.criosweb.ro/wp-content/uploads/2015/11/workflow.png exists, but http://luci.criosweb.ro/wp-content/uploads/2015/11/workflow.png/ does not, so those would have to be fixed as well.

renyouguo:
@fuglede Thanks, it is ok now.

fuglede:
Thanks for the update; the test URLs still do not quite refer to existing content due to the trailing slashes: http://luci.criosweb.ro/wp-content/uploads/2015/11/board-config.png/ does not exist, but http://luci.criosweb.ro/wp-content/uploads/2015/11/board-config.png does.

renyouguo:
@fuglede Ok now.

fuglede:
Thanks for updating those; would you mind [squashing](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) the commits to keep the history a bit cleaner?

renyouguo:
#5147 
@fuglede https://travis-ci.org/EFForg/https-everywhere/jobs/137028240#L267
Would you please tell me how to fix this issue?

BTW, my 70+ PRs are waiting.
https://github.com/EFForg/https-everywhere/pulls/renyouguo

fuglede:
Hm, I think I need some more context to that one. It is unrelated to this PR, right?

renyouguo:
@fuglede Yes.

PR #4861, opened 2016-05-20
Create cdnmaster.xml

renyouguo:


J0WI:
I got a mismatch on `f.cdnmaster.com`, could you add a comment for this?

PR #4864, opened 2016-05-20
Create uptodown.xml

renyouguo:


J0WI:
There are much more subdomains. At least all language domains are working. See also utdstc.com.

renyouguo:
@J0WI I'm not sure , is wildcard ok?

J0WI:
All app domains seems to be okay, so a wildcard makes sense.
There is a mismatch on ^, blog. and gstatic. For blog you may need a downgrade, because the links from app pages to it are protocol relative.

J0WI:
Downgrade rules have to be listed in `utils/downgrade-whitelist.txt`

J0WI:
You can replace both redirects to https with a simple `<rule from="^http:" to="https:" />`, this will also avoid the current error.

J0WI:
(img|stc)\.utdstc\.com needs to be added as target host

J0WI:
Thanks!

PR #4865, opened 2016-05-20
Update Baidu.xml

renyouguo:


renyouguo:
Those time out check error site is so fast in china.

Issue #4810, opened 2016-05-16
Latency to Asian hosts


J0WI:
It turns out, that Travis has a huge latency to Asian hosts.
This results in frequent false positive timeouts during tests.

fuglede:
For what it's worth (which is not all that much), this is not a huge issue for the day-to-day Travis usage: the test only [compares with the current master branch](https://github.com/EFForg/https-everywhere/blob/master/test/travis-ruleset-fetch.sh#L23). This means that the fetch test is run for PRs but not for the merge commit, allowing us to ignore the results if needed, and after confirming that the false negatives are indeed false.

However, this of course has no influence on the periodic database-wide runs of the checker which will still give false negatives.

fuglede:
@J0WI: Will you keep an eye on whether or not this was sufficient to kill off the issues?

J0WI:
Sure. I'll reopen if I still run in this issue.

PR #4866, opened 2016-05-20
Increase ruleset testing timeout limits

fuglede:
This doubles the length of the timeout restrictions for all current https-everywhere-checker configs in an attempt to fix https://github.com/EFForg/https-everywhere/issues/4810.

Arguably, since Travis is the one causing that particular issue, it could suffice to change only http.checker.config, but increasing all of them is probably safer for database-wide runs anyway.


J0WI:
I wonder why we have the same config several times?

This change is trivial and LGTM, but Python and tool stuff should be reviewed by someone like @Hainish.

fuglede:
The configs are all slightly different (and used for different purposes).

Hainish:
@fuglede this will probably cause the full fetch test take quite a bit longer (see https://github.com/efforg/https-everywhere-full-fetch-test/), but if that means not mistakenly disabling rulesets which aren't actually broken, then that's a good thing.

Issue #4869, opened 2016-05-20
mturk.com mistake


xJMV:
Error made: `<rule from="^http://mturk\.com/" to="https://www.mtur.com/"/>`

Should be : `<rule from="^http://mturk\.com/" to="https://www.mturk.com/"/>`

PR #4873, opened 2016-05-21
[MTurk] Fix rule

fuglede:
This closes https://github.com/EFForg/https-everywhere/issues/4869. A bit scary that the checker lets that big a distance through.


J0WI:
`workersandbox.mturk.com` works and mismatch on `blog.mturk.com`. Just to have full coverage.

fuglede:
Done.

PR #4880, opened 2016-05-21
Add ruleset for thregr.org

wavexx:


J0WI:
There are also `sec.` and `e.`, but they serve a mismatching cert. Could you add a comment documenting this?

wavexx:
Well, they serve a completely unrelated cert under SNI, not really mismatching.
Do you really want to add comments for such cases? Adding comments for VH will get outdated quickly.

J0WI:
> Adding comments for VH will get outdated quickly.

Is this general or specific for this host?
In general such comments makes it easier in future to maintain the rules.

wavexx:
On Sun, Jul 24 2016, J0WI <notifications@github.com> wrote:
>> Adding comments for VH will get outdated quickly.
>
> Is this general or specific for this host?
> In general such comments makes it easier in future to maintain the rules.

Well, as the owner of thregr.org, I can say this is the only public
cert. I add-remove other hosts with SNI, but the certs are not for
public use (if you look, they use an independent cert root and do not
validate on purpose).

So even if I add some comments about the currently available certs, it
will become outdated in a few weeks. What's the point? It's like adding
comments for other currently available VH running on the same domain.

Since I'd have other rules for other domains I do not own, this is more
like a general question. I generally never scan for other available
certs when I add a rule unless resources are explicitly loaded from it.

J0WI:
Okay, thanks for explaining.

For domains you don't own I highly recommend to look for other subdomains, because you easily miss some important ones, that are not linked on the main site.

PR #4882, opened 2016-05-22
[WhatPulse.org] Add ruleset

synthtech:
Adds support for [whatpulse.org](https://whatpulse.org).

**Edit:** I'm not sure why the build failed, [developer.whatpulse.org](https://developer.whatpulse.org) appears to work just fine on my end.

**Edit 2:** Added `www` subdomain, added notes for `amcdn` and `direct` mismatched subdomains, secured cookies.

### DIFF
--- a/src/chrome/content/rules/WhatPulse.org.xml
+++ b/src/chrome/content/rules/WhatPulse.org.xml
@@ -1,0 +1,19 @@
+<!--
+	Mismatch:
+		- amcdn.whatpulse.org
+		- direct.whatpulse.org
+		- help.whatpulse.org
+		- www.whatpulse.org
+
+	Missing intermediate certificate:
+		- developer.whatpulse.org (https://github.com/EFForg/https-everywhere/pull/4882)
+
+-->
+<ruleset name="WhatPulse.org">
+	<target host="whatpulse.org" />
+	<target host="api.whatpulse.org" />
+
+	<securecookie host="^\.?whatpulse\.org$" name=".+" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


jeremyn:
Thanks for the ruleset!

https://www.whatpulse.org/ is broken because the certificate is only valid for whatpulse.org. Firefox reports this error. Chrome redirects with this message in the developer tools console: `Redirecting navigation www.whatpulse.org -> whatpulse.org because the server presented a certificate valid for whatpulse.org but not for www.whatpulse.org. To disable such redirects launch Chrome with the following flag: --force-fieldtrials=SSLCommonNameMismatchHandling/Disabled/`. Please remove this target and add the subdomain under `Mismatch` in the comment.

https://developer.whatpulse.org/ errors in Firefox because it's missing an intermediate certificate, which is the same problem the test has. It works in Chrome and Edge. @Hainish , what should we do here?

synthtech:
Ok, added the comment for `www` and left `developer` alone for now.

jeremyn:
Thanks.

We need to wait for feedback on the missing certificate for `developer` before I merge this.

Hainish:
@jeremyn we should leave `developer` out if it will cause problems for Firefox users.

Hainish:
Good catch!

jeremyn:
@Hainish Thanks for the guidance.

@synthtech Please delete the `developer` target and change the top comment to this or something similar:

```xml
<!--
	Mismatch:
		- amcdn.whatpulse.org
		- direct.whatpulse.org
		- help.whatpulse.org
		- www.whatpulse.org

	Missing intermediate certificate:
		- developer.whatpulse.org (https://github.com/EFForg/https-everywhere/pull/4882)
-->
```

synthtech:
Done

jeremyn:
Thanks! Merged.

FYI: I tried https://developer.whatpulse.org in Firefox and it works fine now. I updated recently from Firefox 48.0 to 48.0.1, so maybe 48.0.1 has the new intermediate certificate? Anyway, `./fetch-test.sh` still reports an error, so I guess the HTTPS Everywhere development certificates should be updated.

jeremyn:
@synthtech But just to be clear, you don't need to do anything further. Thanks again.

fuglede:
@jeremyn: I checked, and the chain is still broken here, so that that's not an issue with the HTTPS Everywhere certificate repository. Firefox caches intermediate certificates (dubbing them "software security devices" in the list of certificates), so maybe you picked up the missing certificate somewhere else? I generally find `test.sh --justrun` an easy way to play around with the default certificate store.

jeremyn:
I don't know how I could have picked up a missing certificate other than through the 48.0.1 update. I browse on Windows so I can't use `./test.sh` there. I can confirm it's still broken using `fetch-test.sh` in the Linux development environment.

jeremyn:
Also, Chrome no longer gives me the redirect message.

fuglede:
Seems like a curious setup: https://www.ssllabs.com/ssltest/analyze.html?d=developer.whatpulse.org reports two different paths to a CA, yet the 2-certificate chain I'm given through `openssl s_client -connect developer.whatpulse.org:443 -showcerts` appears to match neither.

jeremyn:
The intermediate certificate `StartCom Class 1 DV Server CA` in both cases is the same and I think that's what's missing. I suppose path 1 was added because of the `Weak or insecure signature, but no impact on root certificate` warning in path 2, though I'm not sure what that means.

If I try https://developer.whatpulse.org with an Android device. I get a warning in Chrome but it's fine in Firefox 48.0. The intermediate certificate must just be slowly making its way around the various stores.

PR #4883, opened 2016-05-22
[Panda Express] New ruleset

galenide:


galenide:
@J0WI @fuglede Are there issues with this ruleset that need to be addressed? If so, it would be useful to know what they are. Thanks.

J0WI:
Could you also document the issues? Rule LGTM.

Timeout:
pandaexpress.com (without www)
click.pandaexpress.com
pages.pandaexpress.com
view.pandaexpress.com

Mismatch:
m.pandaexpress.com

J0WI:
Thanks!

PR #4900, opened 2016-05-24
add new rules

Kernel-Error:
add new rules for my website


J0WI:
This host is already HSTS preloaded, so there is no need to include it in HTTPS-Everywhere.
Thanks anyway!

PR #4903, opened 2016-05-24
removed obsolete VK-ols.xml and add userapi.com.xml

J0WI:


jeremyn:
For `userapi.com.xml`, can we just do a left-hand wildcard `<target host="*.userapi.com" />`? I've tried some subdomains like https://foo.userapi.com and they seem to work.

J0WI:
I found some issues on content that is served there, e.g. on cs6183.userapi.com https does not work.

J0WI:
Firefox build is broken because of #6489.
Rule tests passed.

jeremyn:
https://cs[0-9]?.userapi.com all work. https://cs10.userapi.com is unable to connect. https://cs11.userapi.com is a mismatched certificate.

I'm not sure what to recommend. What do you think?

J0WI:
I'm not sure because I didn't found any test domains for them.

jeremyn:
I consider https://cs1.userapi.com/ itself a valid test domain, but let's be cautious and leave them out, especially for a high profile site like this.

PR #4904, opened 2016-05-24
[Expo2020] Add ruleset

synthtech:
**Edit:** Added missing `grant` subdomain and added comment explaining `www` redirect.

### DIFF
--- a/src/chrome/content/rules/Expo2020.xml
+++ b/src/chrome/content/rules/Expo2020.xml
@@ -1,0 +1,17 @@
+<!--
+	404:
+		- www.expo2020dubai.ae
+
+	Invalid:
+		- stage.expo2020dubai.ae
+
+	Mixed content:
+		- expo2020dubai.ae
+
+-->
+<ruleset name="Expo 2020 (partial)">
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the ruleset! Here are some problems I see:

* https://expo2020dubai.ae/ is broken because of mixed content, for example the video autoplays with HTTP but not with HTTPS. Either delete the target and move this to a `Mixed content` section in the comment and rename the ruleset to `Expo 2020 (partial)`, or add `platform="mixedcontent"` to the `<ruleset>` tag, in which case the rule will [usually be disabled](https://www.eff.org/https-everywhere/rulesets#mixed-content).

* https://www.expo2020dubai.ae/ 404s but is a valid HTTPS connection. I don't think we should rewrite the URL to change the site's behavior even if the site's behavior is to 404. Please delete the target and the rewrite rule.

Please be sure to squash all changes into one commit.

synthtech:
Added comments for `www` and the main domain.

jeremyn:
Thanks!

Issue #4762, opened 2016-05-12
Unable to test IPv6 hosts


J0WI:
Travis does not support IPv6. IPv6 only hosts will fail our tests.
See #4080 or #4760.

Travis refuses to support IPv6 in near future (see linked comments in https://github.com/EFForg/https-everywhere/pull/4080#issuecomment-218583651)

How do we handle that?
I think this get much more important for us in near future.
cc @Hainish @jsha

J0WI:
As a workaround we should think about a 6to4 tunnel:
https://wiki.debian.org/DebianIPv6#IPv6_6to4_Configuration

mnordhoff:
Looks like they don't support proto-41 traffic either. If that remains true, 6in4 and 6to4 are out:

https://stackoverflow.com/questions/31066205/how-to-allow-protocol-41-6in4-through-the-gce-firewall
https://cloud.google.com/compute/docs/networks-and-firewalls
https://cloud.google.com/compute/docs/protocol-forwarding/

Hainish:
I think a quick solution to this is to use a third-party ipv6 over 4 such as [miredo](http://www.remlab.net/miredo/).  This usually works OOTB, but still seems to be failing for my test travis run:

https://travis-ci.org/Hainish/https-everywhere/jobs/129889699#L648

Hainish:
I wonder if the error is occurring due to ipv6 kernel modules not being loaded?  I see the same error as the travis build on my local before installing `miredo`, but as soon as installing `miredo` the problem is resolved.

J0WI:
It should be enabled in newer versions of Ubuntu. You may check for `/proc/net/if_inet6` and the config in `/etc/sysctl.d/` anyway.
May be it also requires to restart the networking service?

Hainish:
It looks like within a docker container, we can get travis speaking ipv6 via `miredo`:

https://travis-ci.org/Hainish/test-travis-ipv6/builds/131532034
https://github.com/Hainish/test-travis-ipv6/tree/b0442d7a7903367f1f27c458d19d947b0dbd0a6b

This is less ideal than getting it to work natively in the travis build environment, since it would require restructuring our travis file quite a bit.  I'll continue trying to get native support working.

Hainish:
It seems as though travis doesn't want you poking around at ipv6 settings: https://travis-ci.org/Hainish/test-travis-ipv6/builds/131556865

Hainish:
[...]

PR #4905, opened 2016-05-24
Dockerize travis test

Hainish:
@cooperq ready for review


cooperq:
looks good to me! :+1:

fuglede:
Purely out of interest: What are the benefits/downsides of doing it through `docker`?

Hainish:
@fuglede for the history of this issue, see https://github.com/EFForg/https-everywhere/issues/4762  Even despite the longer test times, the advantage of being able to do IPv6 tests and having a locally-reproducible test environment is in my estimation worth the trade-off.  If we continue to see errors like the above, we might have to reevaluate that trade-off.

fuglede:
Makes sense; thanks!

PR #4908, opened 2016-05-25
[Public.Resource.Org] Enabled ruleset, added/removed domains, improve…

galenide:
…d cosmetics, documented unsupported domains


jeremyn:
Thanks for the ruleset! Please:

* note `yourhonor.org` as a `cert name mismatch` in the top comment
* squash everything into one commit

jeremyn:
Actually, don't worry about squashing the commits.

galenide:
When accessed as normal (i.e. without HTTPS Everywhere), it appears that `http://www.public.resource.org/` gives a 302 redirect to `https://public.resource.org/` instead. In addition, accessing `http://www.yeswescan.org/` as normal gives a 302 redirect to `https://www.yeswescan.org/`; however, an error then occurs because the certificate is valid for `yeswescan.org` (among other domains) but is not valid for the `www.yeswescan.org` domain. This raises the question as to whether it would be useful for the ruleset to explicitly redirect `http://www.public.resource.org/` to `https://public.resource.org/` and to redirect `http://www.yeswescan.org/` to `https://yeswescan.org/` also. (Even without these two explicit redirects, the ruleset should be an improvement over the current ruleset that appears to be disabled by default.)

jeremyn:
I don't get a certificate mismatch for https://www.wwlbd.org but otherwise your latest updates look good.

Thanks for noticing the redirect issue. A specific redirect to bypass broken HTTPS is fine. The [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) says not to do redirects if the starting HTTPS domain "serves HTTPS correctly", but here https://www.public.resource.org/ and https://www.yeswescan.org/ are broken, so we can redirect around them.

galenide:
In the ruleset, a number of "Unsupported domains" are listed. When the ruleset is _not_ enabled, the behavior of these domains is as follows:

`http://resource.org/` redirects to `https://public.resource.org/` (the cert is valid for public.resource.org)
`http://patent.resource.org/` redirects to `https://public.resource.org/` (the cert is valid for public.resource.org)
`http://www.public.resource.org/` redirects to `https://public.resource.org/` (the cert is valid for public.resource.org)
`http://web.resource.org/` displays content
`http://xml.resource.org/` redirects to `http://xml2rfc.ietf.org/`
`http://www.wwlbd.org/` redirects to `https://wwlbd.org/` (the cert is valid for wwlbd.org)
`http://www.yeswescan.org/` redirects to `https://www.yeswescan.org/` and triggers a cert name mismatch error
`http://yourhonor.org/` redirects to `https://yourhonor.org/` and triggers a cert name mismatch error
`http://www.yourhonor.org/` redirects to `https://www.yourhonor.org/` and triggers a cert name mismatch error

Right now, it appears that Travis CI testing has failed, with a "Unknown SSL protocol error" in the case of `law.resource.org`. Accessing `https://law.resource.org/` in a browser appears to work, though the process of loading the site seems to take longer than some of the other domains (such as `public.resource.org`.) There is also the question as to whether the specific HTTPS support at Public.Resource.Org has changed over time.

jeremyn:
I "bounced" the pull request to rerun the tests, which work now. Sometimes that helps.

Thanks for the analysis but it's not clear what you are asking. Are you wondering if you should add additional rewrite rules to avoid broken HTTPS like you did for http://www.yeswescan.org ? Sure. If http://example.com redirects to a working https://www.example.com by going through an HTTPS-misconfigured https://example.com , then you can add a rule to rewrite http://example.com directly to https://www.example.com .

HTTPS support at any site can change over time, we just need to make do with what we have now.

galenide:
As of now, the ruleset has a normal redirect ("^http:" to "https:") plus a specific redirect so that http://www.yeswescan.org/ is mapped to https://yeswescan.org/ (which works) instead of https://www.yeswescan.org/ (which gives a cert name mismatch.) At this point, it is does not appear that any additional redirects are needed. For the unsupported domains, other than http://www.yeswescan.org/, the domains either redirect to an HTTPS site that works (such as https://public.resource.org/) or to an HTTP site. Though there is an issue where https://yourhonor.org/ and https://www.yourhonor.org/ have cert name mismatches, it is not clear that anything can be done for those domains as of now.

Are there additional issues that need to be addressed in order for the ruleset to be accepted?

jeremyn:
It looks fine to me. Do you mind if I [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) due to all the commits?

galenide:
A squash-and-merge should be OK.

jeremyn:
Thanks again! Merged.

PR #4913, opened 2016-05-25
Update Microsoft.xml

renyouguo:


src/chrome/content/rules/Microsoft.xml
```diff
@@ -601,11 +580,14 @@
 		<!--	Needed for Windows 8+ Captive Portal Helper:
 			See https://github.com/EFForg/https-everywhere/issues/835
 						-->
-		<exclusion pattern="^http://go\.microsoft\.com/fwlink/\?LinkID=219472&amp;clcid=0x409$" />
+		<exclusion pattern="^http://go\.microsoft\.com/fwlink/\?LinkID=219472(.*)" />
```
J0WI: `.*` are not nice, but I think in this case it's okay.

PR #4917, opened 2016-05-25
Create Commandlinefu.com.xml

davidabian:


src/chrome/content/rules/Commandlinefu.com.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Commandlinefu.com">
+
+	<target host="www.commandlinefu.com" />
+
+	<rule from="^http://commandlinefu\.com/"
+		to="https://www.commandlinefu.com/" />
```
J0WI: This will never be used because `commandlinefu.com` is not a listed target host.
You can remove this lines anyway since both aliases support https.

davidabian: Done, @J0WI. Thanks.

PR #4920, opened 2016-05-26
Update Baidupcs.com.xml

renyouguo:


J0WI:
Please use tabs instead of spaces and document that `baidupcs.com` does not exist.

J0WI:
You still have tabs there....

PR #4922, opened 2016-05-26
Update Baidu.xml

renyouguo:


src/chrome/content/rules/Baidu.xml
```diff
@@ -230,11 +228,13 @@
 	<target host="baishi.baidu.com" />
 	<target host="bce.baidu.com" />
 	<target host="login.bce.baidu.com" />
-	<target host="bs.baidu.com" />
+	<!--	Broken:
```
J0WI: please use the more specific "expired"

renyouguo:
@J0WI Done.

J0WI:
Thanks!
If you change that much comments you should consider moving them all on the top for readability.

PR #4923, opened 2016-05-26
utils/trivial-validate.py Windows fix

galeksandrp:


J0WI:
I'm not familiar with Python, but I think there are generic solutions like `os.path`. We may have the same problem in other scripts.
Let me ping @Hainish with that.

Hainish:
@galeksandrp a cleaner approach to this would be to use the following:

```python
filename_split = filename.split(os.path.sep)
basename = filename_split[len(filename_split) - 1]
```

then see if basename is equivalent to any of those three filenames.  This is os-agnostic due to the use of `os.path.sep`

PR #4926, opened 2016-05-26
Create wipe.de.xml

IDKwhattoputhere:
Analytics service. Example can be found here: `http://www.dastelefonbuch.de/`


J0WI:
It seems that this host is used for a lot of customers (see `pub.oy1.oe.wipe.de`).
Could you document that?

J0WI:
Sorry, I forgot to mention that you should also comment that there is no host on `wipe.de` and `www.wipe.de`.

PR #4933, opened 2016-05-27
Fix #2068

### LINKED ISSUE #2068: Broken videos on 3news.co.nz
https://www.3news.co.nz/nznews/man-on-hunger-strike-over-weight-loss-surgery-2015070618

You get a Flash not found notice with the 3 News rule enabled.  If you just disable protection to get around this in Firefox you still get an error that the video cannot play.  The 3 News rule has to be disabled for player content to work properly.

galeksandrp:


J0WI:
Please also check:
upload.mediaworks.co.nz
test2.mediaworks.co.nz
test1.mediaworks.co.nz
test.mediaworks.co.nz

and add comments or redirects

src/chrome/content/rules/3_News.xml
```diff
@@ -33,10 +34,10 @@
 	<securecookie host="^(?:www\.)?3news\.co\.nz$" name=".+" />
 
 
-	<rule from="^http://(?:www\.)?3news\.co\.nz/"
-		to="https://www.3news.co.nz/" />
+	<rule from="^https?://(?:www\.)?3news\.co\.nz/"
```
J0WI: We try to avoid redirects from https, except it breaks something in the site (like wrong protocol relative redirects).

J0WI: If you touch this rule anyway, could you also remove the `?:`?

src/chrome/content/rules/MediaWorks.xml
```diff
@@ -35,7 +37,10 @@
 	<securecookie host="^(?:cdn|static)\.mediaworks\.co\.nz$" name=".+" />
 
 
-	<rule from="^http://(cdn|static)\.mediaworks\.co\.nz/"
-		to="https://$1.mediaworks.co.nz/" />
+	<rule from="^https?://cdn\.mediaworks\.co\.nz/"
```
J0WI: We try to avoid redirects from https, except it breaks something in the site (like wrong protocol relative redirects).

galeksandrp:
Added.

src/chrome/content/rules/3_News.xml
```diff
@@ -33,10 +23,7 @@
 	<securecookie host="^(?:www\.)?3news\.co\.nz$" name=".+" />
```
J0WI: Since www and ^ are removed, the securecookie makes no sense anymore.

galeksandrp:
Removed.

PR #4941, opened 2016-05-27
[1f0.de] Update ruleset

synthtech:


src/chrome/content/rules/1f0.de.xml
```diff
@@ -1,17 +1,11 @@
...
-	<target host="1f0.de"/>
-	<target host="www.1f0.de"/>
-
-	<!--	Cert isn't valid for !www
-		Pages redirect to !www		-->
-	<rule from="^http://(?:www\.)?1f0\.de/wp-content/"
-		to="https://www.1f0.de/wp-content/"/>
+<ruleset name="1f0.de">
+	<target host="1f0.de" />
+	<target host="files.1f0.de" />
```
J0WI: Please also add `git.1f0.de`, `www.1f0.de` and a comment for non-working `chaos.1f0.de`.

synthtech:
Should be fixed.

On a side note, should `www` always be added even if it redirects back to the main domain?

J0WI:
Thanks!

PR #4946, opened 2016-05-28
Create Tutorialspoint.xml

renyouguo:


src/chrome/content/rules/Tutorialspoint.xml
```diff
@@ -0,0 +1,15 @@
+<ruleset name="Tutorialspoint">
+
+	<!--	^tutorialspoint.com/ do not exist.	-->
```
J0WI: Secure connection refused on `dev.`

Please move these comments on the very top of the file.

src/chrome/content/rules/Tutorialspoint.xml
```diff
@@ -0,0 +1,15 @@
...
+	
+	<target host="www.tutorialspoint.com"/>
+
+	<!--	Mixed content	-->
+	<exclusion pattern="^http://www\.tutorialspoint\.com/(.+\.(html?|php))?$" />
+		<test url="http://www.tutorialspoint.com" />
+		<test url="http://www.tutorialspoint.com/index.htm" />
```
J0WI: Please also add an example for `.html`.

renyouguo:
@J0WI Done.

J0WI:
@renyouguo Thanks, but I think the change on the regex itself is not necessary, now you would need even more test urls.

renyouguo:
@J0WI 
/(.+\.(html?|php))? could cover /html ?
So I change the regex.

J0WI:
My comment was just about the lack of a test url that contains a `.html` file, not just `.htm`.
Sorry if that was not clear.

renyouguo:
@J0WI Sorry, I have found other mixed content links like http://www.tutorialspoint.com/html5 , so I have changed the regex total.

src/chrome/content/rules/Tutorialspoint.xml
```diff
@@ -0,0 +1,22 @@
...
+	Mismatch: ^tutorialspoint.com
+	
+	Refused: dev.tutorialspoint.com
+	
+	-->
+	
+	<target host="www.tutorialspoint.com"/>
+
+	<!--	Mixed content	-->
+	<exclusion pattern="^http://www\.tutorialspoint\.com/(?!.*\.(css|js|gif|ico|jpg|png))" />
```
J0WI: Would it be possible to use something like `[\w/-]+` instead of `.*` here?

src/chrome/content/rules/Tutorialspoint.xml
```diff
@@ -11,11 +11,14 @@
...
-		<test url="http://www.tutorialspoint.com/index.htm" />
-		<test url="http://www.tutorialspoint.com/tutor_connect/index.php" />
+	<exclusion pattern="^http://www\.tutorialspoint\.com/(?![\w/-]+\.(css|js|gif|ico|jpg|png|woff))" />
+		<test url="http://www.tutorialspoint.com/theme/css/style-min.css" />
+		<test url="http://www.tutorialspoint.com/theme/js/custom-min.js" />
+		<test url="http://www.tutorialspoint.com/images/loading-cg.gif" />
+		<test url="http://www.tutorialspoint.com/favicon.ico" />
+		<test url="http://www.tutorialspoint.com/images/apple_store.jpg" />
+		<test url="http://www.tutorialspoint.com/green/images/logo.png" />
+		<test url="http://www.tutorialspoint.com/theme/css/fonts/font-awesome/fonts/fontawesome-webfont.woff?v=4.1.0" />
```
J0WI: Please remove the `?v=foo` fragment.

PR #4960, opened 2016-05-29
Fix #4684

### LINKED ISSUE #4684: SigmaBeauty rule breaks Add to Bag functionality
http://www.sigmabeauty.com/spa-mat/p/bcm01

SigmaBeauty rule needs to be modified because it's breaking the Add to Bag functionality.

galeksandrp:


J0WI:
Could you also update a few other things here?
Our code style requires tabs and removing of non-capturing groups for readability. Forcing a redirect from ^ to www is not needed anyway since both support https.


There are also some more hosts:

Mismatch on:
blog.sigmabeauty.com
enews.sigmabeauty.com

WFM:
help.sigmabeauty.com

J0WI:
Please add comments on the top of the file for the mismatching ones.
This will make maintenance easier.

galeksandrp:
Added.

J0WI:
Thanks!

PR #4972, opened 2016-05-30
Create url.cn.xml

renyouguo:


src/chrome/content/rules/url.cn.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+Mismatch on other *.url.cn
```
J0WI: Please be a bit more verbose

J0WI:
Could you provide some test cases for this host?
I don't understand the behavior so far.

J0WI:
Uhm, okay.
Thanks for clarify.

PR #4973, opened 2016-05-30
Create Pythonclock.org.xml

renyouguo:


src/chrome/content/rules/Pythonclock.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Pythonclock.org">
+
+	<target host="pythonclock.org" />
```
J0WI: Please also add the `ẁww.` alias.

PR #4974, opened 2016-05-30
git.openwrt.org works

MilanKral:


J0WI:
Could you please also add `mirror2.openwrt.org` and a comment for `buildbot.openwrt.org`?
I think we should also remove the `(partial)` flag here since all major hosts are supported.

J0WI:
Thanks!

PR #4977, opened 2016-05-31
Create letvimg.xml

renyouguo:


J0WI:
Please document that (www.)letvimg.com don't exist.
Can you provide a test case for this?

J0WI:
This seems to be related with `letvcdn.xml` (pending on #4978).
Could you add a comment on both referring to each other? Please use the filename for this, not bug numbers.

PR #4978, opened 2016-05-31
Create letvcdn.xml

renyouguo:


J0WI:
Please document the behavior of www and ^.
Can you provide a test case where this is used?

PR #4983, opened 2016-05-31
Update Sina.xml

renyouguo:


J0WI:
I had to revert this:
`ERROR src/chrome/content/rules/Sina.xml: Not enough tests (1 vs 3) for <Exclusion pattern '^http://app\.sina\.com\.cn/(?!favorite\.ico|script/)'>`

PR #4987, opened 2016-06-01
[ColorOfChange] New ruleset

galenide:


jeremyn:
Thanks for the ruleset! Some problems I see:

* https://colorofchange.org/ now works so please remove the downgrade and the exclusion.
* http://represent.colorofchange.org/ redirects to https without the rule, and it has mixed content issues, so I'm okay with making a rule for `represent` that's active by default because we're not making it worse. So please move `represent` into the main ruleset and delete the mixed content ruleset.
* Change the securecookie rule to also catch `.colorofchange.org`.
* Squash everything into one commit.

jeremyn:
Actually, please don't squash the commits. I'm sorry for any confusion.

galenide:
@Hainish The ruleset has been updated as specified and has been resubmitted as part of a series of unsquashed commits. Please check this.

jeremyn:
@galenide Thanks for the changes. Can you please also:

- [ ] add targets for https://static.colorofchange.org , https://iam.colorofchange.org , and https://vote.colorofchange.org
- [ ] note https://stories.colorofchange.org in a top comment as a mixed content blocking problem

I'm sorry I missed these subdomains the first time around. By the way, you can see the ones besides `static` [here](https://www.google.com/search?q=site%3A*.colorofchange.org&ie=utf-8&oe=utf-8#q=site:*.colorofchange.org+-site:act.colorofchange.org+-site:orig.colorofchange.org+-site:represent.colorofchange.org+-site:secure.colorofchange.org+-site:www.colorofchange.org), and you can find `static` by looking `Developer tools > Network` when loading for example https://www.colorofchange.org/ .

Did you specifically want @Hainish to review this and not me?

galenide:
With regard to the overlooked subdomains, that's all right. As for reviewing the ruleset, I meant to specify **jeremyn** but accidentally specified **Hainish**. Even so, it does not really matter as to who specifically reviews the ruleset.

From what it appears, http://stories.colorofchange.org/ redirects to https://stories.colorofchange.org/ when the ruleset is not enabled, but for now, it may be OK for the ruleset to exclude stories.colorofchange.org and to document the subdomain as having mixed content blocking issues (which is what the ruleset does.)

jeremyn:
Oh, I see what you mean about `stories`, I didn't notice that it redirected already. If you want to make it a target and get rid of the top comment, that's fine. Otherwise this is ready to go.

By the way, because of all the commits, I want to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) if that's all right with you.

galenide:
From what I understand, a squash-and-merge should be OK.

jeremyn:
Great! Thanks again for your work.

PR #4988, opened 2016-06-01
[HangSeng.com] New ruleset

Marco129:
Add support for [Hang Seng Bank](https://www.hangseng.com)

### DIFF
--- a/src/chrome/content/rules/HangSengBank.xml
+++ b/src/chrome/content/rules/HangSengBank.xml
@@ -1,0 +1,13 @@
+<ruleset name="Hang Seng Bank">
+	<target host="www.hangseng.com" />
+	<target host="bank.hangseng.com" />
+
+	<rule from="^http://hangseng\.com/"
+		to="https://www.hangseng.com/" />
+
+	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/HangSengBank.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Hang Seng Bank">
+	<target host="www.hangseng.com" />
+	<target host="bank.hangseng.com" />
```
J0WI: ebusiness.hangseng.com seems to work
Please also add a comment for the mismatch on www.cn.hangseng.com

src/chrome/content/rules/HangSengBank.xml
```diff
@@ -0,0 +1,13 @@
...
+
+	<rule from="^http://hangseng\.com/"
```
J0WI: you also need to include hangseng.com as a target

J0WI:
There are some errors shown that aren't related with your changes (see#5252).
Rebasing your branch should fix this.

J0WI:
Thanks!

PR #4989, opened 2016-06-01
[teltarif.de] NEW RULESET 

teltarif:


J0WI:
Thanks for your contribution.
Could you please merge your changes into https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Teltarif.xml?

If you are evolved in the teltarif admin team you should have a look at [preloaded HSTS](https://hstspreload.appspot.com/) which would make this rule obsolete.

teltarif:
There are both the same ruleset: could you please just commit one of them as this ruleset doesn't seem to be live yet > Thanx!

teltarif:
Can you tell me when this teltarif ruleset will go live?
Please let me know if there is anything more from my side to do

J0WI:
It's live (see link above), but we haven't released a new version of HTTPS Everywhere yet.
You have some hosts that are not yet included in the file. Could you add them to the existing file?

teltarif:
yes, you are right, there were some errors in the original Teltarif.xml: I've corrected them, basically copied the ruleset from teltarif.de.xml to Teltarif.xml and deleted teltarif.de.xml so now there is Teltarif.sml left with the updated ruleset

teltarif:
Is everything ok with the last commit (deleted teltarif.de.xml, updated Teltarif.xml)? 
Can you tell me when the new version of HTTPS Everywhere will be released?

src/chrome/content/rules/Teltarif.xml
```diff
@@ -1,11 +1,10 @@
 <ruleset name="Teltarif">
-
-	<target host="teltarif.de"/>
-	<target host="m.teltarif.de"/>
-	<target host="mobil.teltarif.de"/>
-	<target host="www.mobil.teltarif.de"/>
```
J0WI: Why you deleted this host?

src/chrome/content/rules/Teltarif.xml
```diff
@@ -1,11 +1,10 @@
 <ruleset name="Teltarif">
-
-	<target host="teltarif.de"/>
```
J0WI: Why you deleted this host?

src/chrome/content/rules/Teltarif.xml
```diff
@@ -1,11 +1,10 @@
...
-	<target host="m.teltarif.de"/>
-	<target host="mobil.teltarif.de"/>
-	<target host="www.mobil.teltarif.de"/>
-	<target host="www.teltarif.de"/>
-
+        <target host="www.teltarif.de" />
+        <target host="m.teltarif.de" />
+        <target host="mobil.teltarif.de" />
+        <target host="www.0180.info" />
```
J0WI: Please also add `0180.info` (without www).

J0WI:
You should existing hosts that works.
We have a new release very soon (#5045), so if you hurry up we can merge it for the upcoming release.

teltarif:
OK, I did as you said: would be great to have that teltarif ruleset in the new release

Hainish:
You made it in to 5.1.10!

PR #4990, opened 2016-06-01
Add subdomains to existing rule, prettify naming.

marcoslater:


marcoslater:
Would make a new pull requests for the last few changes, but this is still left open, so may as well use it. 😄

J0WI:
Thanks four your contribution!
Could you please split this PR up that we only have one ruleset changed in each PR?

marcoslater:
Sure thing, let me close this one then in the meantime.

Issue #1123, opened 2015-02-26
Do not restart Firefox after installing this plugin


chhantyal:
As far as I know, Firefox already supports plugin install without restarting browser.

It would be great if this plugin also works that way, if there is no particular reason to restart.

Thanks.

reedy:
https://adblockplus.org/blog/how-many-hacks-does-it-take-to-make-your-extension-install-without-a-restart

https://developer.mozilla.org/en-US/Add-ons/Bootstrapped_extensions


The question is whether it's worth the time/effort to do so

diracdeltas:
I have a branch somewhere that started converting HTTPS Everywhere to a restartless extension. At the time the process wasn't well-documented, but perhaps it's changed. I think it's this branch: https://github.com/diracdeltas/https-everywhere/tree/fennec

Things weren't working IIRC but see src/bootstrap.js for a starting point if you're implementing this.

diracdeltas:
I can't unassign myself in Github anymore, so I'll just note that this task is free for anyone to take on.

isonmad:
I mean, for the record, the exact same guy also wrote that all those showstoppers are gone now and haven't applied since Firefox 8. https://adblockplus.org/blog/why-you-should-make-your-next-add-on-restartless

But this is gonna be fixed anyway with the planned switch to WebExtensions for firefox right? https://github.com/EFForg/https-everywhere/issues/7389#issuecomment-255483554

gloomy-ghost:
We will move to WebExtension which is restartless as described in #9958, and I don't think there will be further developments for the XPCOM version as it's already deprecated.

PR #4995, opened 2016-06-02
Fetch rulset via chrome url

chrmod:
In order to make https-everywhere a SystemAddon, it cannot have unpack true
flag in install.rdf. The flag was there, so ruleset.json can be references
with FileUtils API.

This PR, introduce alternative way of loading rulesets that are exposed as chrome resources. In this way an extension can be marked as unpack=false

### DIFF
--- a/src/chrome/content/code/HTTPSRules.js
+++ b/src/chrome/content/code/HTTPSRules.js
@@ -246,7 +246,30 @@ const RuleWriter = {
     return rv;
   },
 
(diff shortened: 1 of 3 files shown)


chrmod:
this pr should push https://github.com/EFForg/https-everywhere/issues/1123 one step forward

chrmod:
@pde please take a look

J0WI:
cc @Hainish

PR #4996, opened 2016-06-02
Create RPGGeek.com.xml

salty-horse:
This website is part of the BoardGameGeek family. The file is a copy of the VideoGameGeek file.

I reported this website in the original commit for BoardGameGeek, but got the domain wrong, so it wasn't maintained like VideoGameGeek.

### DIFF
--- a/src/chrome/content/rules/BoardGameGeek.com.xml
+++ b/src/chrome/content/rules/BoardGameGeek.com.xml
@@ -3,27 +3,13 @@
 
 		- Geekdo-static.com.xml
 		- VideoGameGeek.com.xml
-
-
-	Insecure cookies are set for these domains and hosts:
-
-		- boardgamegeek.com
-		- .boardgamegeek.com
-		- www.boardgamegeek.com
-
+		- RPGGeek.com.xml
 -->
 <ruleset name="BoardGameGeek.com">
 	<target host="boardgamegeek.com" />
 	<target host="www.boardgamegeek.com" />
 
-
-	<!--	Not secured by server:
-					-->
-	<!--securecookie host="^(?:www\.)?boardgamegeek\.com$" name="^geekproxy$" /-->
-	<!--securecookie host="^\.boardgamegeek\.com$" name="^(?:__qca|SessionID)$" /-->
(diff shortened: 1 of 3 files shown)


jeremyn:
Thanks for the ruleset! Some comments:

* Unless I'm missing something, you can delete [these](https://github.com/salty-horse/https-everywhere/blob/55a964204b83e96578844d5270ce1780705cddf2/src/chrome/content/rules/RPGGeek.com.xml#L5-L9) and [these other](https://github.com/salty-horse/https-everywhere/blob/55a964204b83e96578844d5270ce1780705cddf2/src/chrome/content/rules/RPGGeek.com.xml#L17-L20) comments.
* Please use multiple `<securecookie>` tags for simplicity in line with the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

Please squash all changes into one commit.

salty-horse:
I used BoardGameGeek.xml and VideoGameGeek.xml as a template.
If you want, I can edit all of them in the same patch.

I don't understand your suggestion re `<securecookie>`.  Are you referring to the guideline "Avoid regexes with long strings of subdomains"?
Should I split `(?:\.|www\.)?` into three(...?) separate instructions?

jeremyn:
I'm not sure what you mean by "edit all of them in the same patch". What you have here is fine. I just mean that extra small commits for my feedback should be squashed.

I did mean splitting into three instructions but I agree that might be too verbose in this simple case. If that feels weird to you, could you at least remove the non-capturing group marker `?:`? That's also in the style guide.

salty-horse:
This file is a carbon copy of the existing files BoardGameGeek.xml and VideoGameGeek.xml

Any changes I make here should logically be applied to the other files. I was asking if should modify these files in the same commit where I add the file for RPGGeek.

jeremyn:
Sure, that's fine. Also, never mind about squashing commits, sorry for any confusion. You can just make the changes however is most comfortable for you.

salty-horse:
Done!

jeremyn:
Thanks!

PR #5003, opened 2016-06-03
Create lofter.xml

renyouguo:


renyouguo:
@J0WI lofter is Wangyi's Tumblr, so there is no cert error.

src/chrome/content/rules/lofter.xml
```diff
@@ -0,0 +1,34 @@
+<!--
+
+Mismatch:
+	^lofter.com
+
+MCB:
+	*.lofter.com
```
J0WI: Could you document that every user get his own subdomain?

J0WI:
blocked by #5015

PR #5005, opened 2016-06-03
Create Yam.xml

renyouguo:


src/chrome/content/rules/Yam.xml
```diff
@@ -0,0 +1,52 @@
+<!--
+fail:
```
J0WI: please try to be more specific

J0WI:
Tests passed now.

PR #5006, opened 2016-06-03
Update Tumblr.xml

renyouguo:


src/chrome/content/rules/Tumblr.xml
```diff
@@ -131,62 +115,27 @@
 	<target host="assets.tumblr.com" />
 	<target host="secure.assets.tumblr.com" />
 	<target host="embed.tumblr.com" />
+	<target host="media.tumblr.com" />
 	<target host="*.media.tumblr.com" />
```
J0WI: Please add two more test urls for this wildcard.

src/chrome/content/rules/Tumblr.xml
```diff
@@ -122,7 +106,7 @@
 	are still vulnerable whenever they browse blogs while logged in. :(
 
 -->
-<ruleset name="Tumblr.com (partial)">
```
J0WI: This rule is still `partial` because every user has it's own subdomain which is redirected to http (please document this behavior on the top).

src/chrome/content/rules/Tumblr.xml
```diff
@@ -65,28 +66,17 @@
...
-
-		- data *
-		- media *
-		- 25.media *
-		- 37.media *
-
-	* Mismatched
-
-
 	Fully covered subdomains:
```
J0WI: You can remove this section since all are listed as target hosts below.

PR #5008, opened 2016-06-03
Exclude leaflet.github.io

micolous:
From the developers in Leaflet/Leaflet.markercluster#680, Leaflet's CDN doesn't have certificates configured correctly on HTTPS, so requests should not be redirected for that domain until such time they do.

I'm not 100% sure I've got the syntax correct -- I've copied this from another PR so hopefully this is correct.

### DIFF
--- a/src/chrome/content/rules/Github-Pages.xml
+++ b/src/chrome/content/rules/Github-Pages.xml
@@ -4,6 +4,9 @@
 <ruleset name="GitHub Pages">
 
 	<target host="*.github.io" />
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks!
You should use plain http for the test url that it matches the regex.
I merged this into my pending PR #4748 to avoid a merge conflict.

micolous:
Thanks, the affected site has now fixed the problem.  I see you've got the rollback sorted too, so all good!

PR #5010, opened 2016-06-03
Create PConline.xml

renyouguo:


src/chrome/content/rules/PConline.xml
```diff
@@ -0,0 +1,26 @@
+<!--
+	fail:
```
J0WI: please clarify why it fails.

PR #5012, opened 2016-06-03
Create Srcf.net.xml

sampablokuper:


jeremyn:
Thanks for the ruleset!

Please move the domains in the comment into a new comment in the top, like [these examples](https://github.com/EFForg/https-everywhere/search?utf8=%E2%9C%93&q=problematic+domains&type=Code). Also, please try to group the comment subdomains so that we know what the problem is with each of them, or else remove them. The comment should help explain for each subdomain why there is not already a rule for that subdomain.

PR #5013, opened 2016-06-03
TPB

galeksandrp:


J0WI:
I fear this needs to be updadated, because some of them may be down now.
Could you explain the difference between all the files? You should have a look at our existing ThePirateBay.xml

Please add some coments showing the relation between all the files, if there are any.

J0WI:
Note that you also need to mention these rules in ThePirateBay.xml

galeksandrp:
Added.

src/chrome/content/rules/ThePirateBay.xml
```diff
@@ -14,9 +15,20 @@
 	<target host="kanyebay.com" />
 	<target host="www.kanyebay.com" />
 
-	<!-- Should probably split the proxy stuff out to a separate module and cover all the proxies listed at thepiratebay-proxylist.org -->
+	<!-- Other The Pirate Bay rulesets:
```
J0WI: Please move this comment block on the top.

J0WI:
Thanks for the additions! There is a similar PR #4474, could you therefore adding the changes from there:
Removing Old_Pirate_Bay.org.xml
Adding tpb.run www.tpb.run.

Please remove the unrelated change in translations.

galeksandrp:
Added.

J0WI:
Thanks for all the additions!

PR #4919, opened 2016-05-26
Create Baidu_CN.xml

renyouguo:

PR #5015, opened 2016-06-04
test fetch error twice

Hainish:
When a fetch error occurs, try the URL once more before reporting an error


J0WI:
See also #4919

J0WI:
Would love to see this fixed:)

Hainish:
@cooperq can you review this?

renyouguo:
Would you support to auto recheck while check.1 and check.2 get differnet errors ?
In #5626 , I gets different errors each time, when it is ok all the time accuraly.

J0WI:
Adding High Priority, because it blocks a lot of PRs, especially Asian hosts.

PR #5029, opened 2016-06-04
Delete MyWOT-problematic.xml

renyouguo:


J0WI:
I also got filter.mywot.com (blocked fonts are trivial)

Issue #4997, opened 2016-06-02
HTTPS on Amazon


J0WI:
Amazon announced to finally support HTTPS. We should rebase our Amazon rules to check which sites can be secured now.

See also #1578 #3323 #3444 #4255 #4366

xiaoyinl:
HTTPS support on Amazon is still partial. For instance, https://www.amazon.com/J.K.-Rowling/e/B000AP9A6K/ redirects to http.

J0WI:
Seems to happen on all `/B00\w+/` pages -.-

ivan:
https://www.amazon.com/J.K.-Rowling/e/B000AP9A6K/ seems to stay on HTTPS now.

PR #5031, opened 2016-06-04
Update Amazon.xml

renyouguo:
askville has been cut down in 2013.


J0WI:
This need to be completely rebased anyway, see #4997.

PR #5032, opened 2016-06-04
Update Beijing_Qianfendian.xml

renyouguo:


src/chrome/content/rules/baifendian.xml
```diff
@@ -0,0 +1,36 @@
+<!--
+	Nonfunctional subdomains:
+
+		- passport	(no https)
+		- (www.)	(^ → www)
```
J0WI: WFM:
passport.

Refused:
en.

expired:
blog.
setup.

PR #5042, opened 2016-06-06
Create GeekPark.xml

renyouguo:


J0WI:
It seems that you just redirect CDNs?
So rules for those CDNs would make more sense, to avoid duplicates.

renyouguo:
@J0WI qiniu (seven cows) is an popular CDN in China. We can't list all the sites which are using it. No way to add rule for qiniu CDN.

renyouguo:
@Hainish @jeremyn

J0WI:
At least this rule should be marked as partial.
It's hard to say if the redirected CDNs will also affect other sites, so the name of the rule can be irritating.

renyouguo:
@J0WI Long time no see，miss you much.

J0WI:
I think I can live with test urls here.

PR #5044, opened 2016-06-06
[GitLab.io] New ruleset

lucaspetter:


src/chrome/content/rules/GitLab.io.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	For other GitLab coverage, see GitLab.com.xml.
+-->
+
+	<ruleset name="GitLab Pages">
+	<target host="gitlab.io"/>
+	<target host="*.gitlab.io"/>
+
+	<test url="http://www.gitlab.io/"/>
+	<test url="http://pages.gitlab.io/"/>
```
J0WI: Could you provide a third one?
Rule LGTM.

lucaspetter:
As far as I can tell, [pages.gitlab.io](https://pages.gitlab.io) and [https://www.gitlab.io](https://www.gitlab.io) are the only "official" subdomains of gitlab.io. Plenty of [other user-created subdomains](https://duckduckgo.com/?q=site%3Agitlab.io&ia=web) exist, but they're mostly personal blogs or testing sites, so I was hesitant to add test URLs for them. GitLab Pages is currently less than [6 months old](https://about.gitlab.com/2015/12/22/gitlab-8-3-released/), so it's still in the early stages of building up a stable userbase. If we add a user-made site as a test URL and that site later goes down, then so will this ruleset.

For comparison, the Git*Hub* Pages [ruleset](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Github-Pages.xml) has no test URLs at all.

What do you think, should we add a user-made subdomain as a test URL? If so, here are a few examples I found while [searching](https://duckduckgo.com/?q=site%3Agitlab.io&ia=web):
- [1](https://maxlaumeister.gitlab.io/)
- [2](https://pymeshnet.gitlab.io/)
- [3](https://paulfitz.gitlab.io/vermont/)

J0WI:
> For comparison, the GitHub Pages ruleset has no test URLs at all.
Yes, I fixed this in #4748

I used GitHub pages of popular companies there, but I think there are currently no such examples on GitLab. So feel free to add any subdomain, that seems to be working long term.

lucaspetter:
Now one of the builds is failing and Travis says "Chromium: HTTP to HTTPS redirection failed." The build was passing fine before. I don't understand why it's failing when a new test URL is the only change.

J0WI:
Was a temporary failure, test passed now.
Thanks for your contribution!

PR #5046, opened 2016-06-07
Added t411.in and t411.ch to T411 rule.

logut:
This website changed domain name
It used t411.me t411.io t411.in  and is now using t411.ch.
I excluded wiki.t411.ch because it has a redirection problem as said in #2988.


J0WI:
This is a dupe of #5750 which contains some more hosts.

PR #5059, opened 2016-06-08
Add rules for German Le Monde diplomatique

gandaro:
I have got one question before this pull request is merged. Regarding `securecookie`, the following is stated in *docs/rulesets.html*:

> A cookie whose domain attribute starts with a "." (the default, if not specified by Javascript) will be matched as if it was sent from a host name made by stripping the leading dot.

Does that mean that cookies like the following will be matched by `<securecookie host="^monde-diplomatique\.de$" name=".+" />`?

Screenshot of a cookie with domain ".monde-diplomatique.de" in Google Chrome's developer tools.


gandaro:
Well, obviously not, as I tested it. Now it should be ready for merging.

J0WI:
Thanks!

PR #5064, opened 2016-06-08
Added Sankaku Complex image boards

AndreasSE93:
Also added a outcommented exclusion rule for charts.sankakucomplex.com to point out that not all subdomains have https.
charts.sankakucomplex.com/niconico can be reached from the Videos menu in the navbar at the main page.

### DIFF
--- a/src/chrome/content/rules/sankakucomplex.com.xml
+++ b/src/chrome/content/rules/sankakucomplex.com.xml
@@ -1,3 +1,10 @@
+<!--
+
+	Nonfunctional domains:
+
+		- charts	(blank page)
+
+-->
 <ruleset name="Sankaku Complex.com">
 
 	<target host="sankakucomplex.com" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the update! I think I understand what the commented exclusion is for but it is confusing. Instead, please add a comment at the top ([examples](https://github.com/EFForg/https-everywhere/search?utf8=%E2%9C%93&q=problematic+domains)) that says `charts` is broken. Also please be sure to squash all changes into one commit.

AndreasSE93:
@jeremyn

> Also please be sure to squash all changes into one commit.

Do you mean in the future if I add multiple sites at once, or that I should make these updates that you requested and squash that commit with this commit and force push?

jeremyn:
I mean that instead of two commits, one for the original update and one to do the small comment change I asked for, you should squash them into one commit and force-push to your branch so that I only merge one commit.

Generally multiple commits are fine if they're for logically different things, but here you're modifying the first commit based on my feedback, so we just want one commit in the end.

jeremyn:
Actually, never mind about squashing. Sorry for the confusion.

PR #5065, opened 2016-06-09
Exclude non-default ports.

dbaxa:
Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/src/chrome/content/rules/Atlassian.xml
+++ b/src/chrome/content/rules/Atlassian.xml
@@ -86,13 +86,14 @@
 
 		<!--	Added by Atlassian:
 						-->
-		<exclusion pattern="^http://(?:.+\.(?:ams|dyn\.syd|inf|internal|sf|office)|go)\.atlassian\.com/" />
+		<exclusion pattern="^http://(?:.+\.(?:ams|dyn\.syd|inf|internal|sf|office)|go)\.atlassian\.com" />
 
 			<test url="http://x.ams.atlassian.com/" />
 			<test url="http://eg.ams.atlassian.com/" />
 			<test url="http://eg.dyn.syd.atlassian.com/" />
 			<test url="http://eg.inf.atlassian.com/" />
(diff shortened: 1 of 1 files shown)


dbaxa:
cc @Hainish

J0WI:
I don't think that removing the trailing slash is a good practice, @fuglede?

Note that your commits are not signed as [expected](https://help.github.com/articles/signing-commits-using-gpg/).

dbaxa:
@J0WI do commits need to be signed now?

J0WI:
No no, I's just a bit irritating because of the "Signed-off-by" part in the commit message.

fuglede:
@J0WI: I don't have any very strong opinions on the matter (but I don't quite see why it was removed here).  But yeah, now that particular exclusion excludes any connection to

     http://go.atlassian.com.some.unrelated.subdomain.atlassian.com

even if that is probably not intended. In any case, exclusions only come into effect for a ruleset if a target host matches.

dbaxa:
@fuglede it was removed because we want internal sites that run on ports other than 80 to not be sent to https:// . I was considering updating the regex to accept any port followed by `/` but when I looked I didn't see other examples of doing that but I did see rules that seemed to not end with `/`.

fuglede:
Thanks for the explanation @dbaxa; that certainly makes sense. Can you confirm that non-default ports were not excluded before? I'm wondering as I seem to remember that HTTPS-E did not work properly with non-default ports in the first place (so that excluding them should not actually be necessary).

dbaxa:
@fuglede I can confirm that with non-default ports in chrome HTTPS-E changes an access to `http://eg.internal.atlassian.com:81/` into `https://eg.internal.atlassian.com`.

fuglede:
Thanks for elaborating. This looks good to me then; we could avoid the super edgy edge cases, but the added complexity to the pattern is probably not worth it.

dbaxa:
@fuglede thank you for merging my pull request.

PR #5073, opened 2016-06-10
Update Mmstat.com.xml

renyouguo:


J0WI:
I also got ju.mmstat.com
Please also document the behavior for www and ^.

PR #5079, opened 2016-06-10
Create eweiqi.xml

renyouguo:


J0WI:
I also got wbb.eweiqi.com.
Is soft.eweiqi.com only available in china or does this host no longer exist?

PR #5082, opened 2016-06-10
PureFTPd.org and DNSCrypt.org

MilanKral:


src/chrome/content/rules/PureFTPd.org.xml
```diff
@@ -0,0 +1,20 @@
+<ruleset name="PureFTPd.org">
+	<target host="pureftpd.org" />
+	<target host="www.pureftpd.org" />
+
+	<target host="www.ucarp.org" />
```
J0WI: This host seems to be unrelated

src/chrome/content/rules/PureFTPd.org.xml
```diff
@@ -0,0 +1,20 @@
+<ruleset name="PureFTPd.org">
```
J0WI: Please also document mismatch on geocaching.pureftpd.org

J0WI:
@MilanKral would be great if we could finish this :)

J0WI:
Thanks!

src/chrome/content/rules/ucarp.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="ucarp.org">
+	<target host="ucarp.org" />
+	<target host="www.ucarp.org" />
+
+	<rule from="^http://(?:www\.)?ucarp\.org/"
```
J0WI: quoting from our style guide:
Prefer capturing groups (www\.)? over non-capturing (?:www\.)?.

J0WI:
Could you also add comments on the top, that other contributors are aware of the relation between the tree files?
See eg. Google or Amazon rules for an example. [[1](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Google.tld_Subdomains.xml#L3), [2](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/GoogleServices.xml#L5)]

src/chrome/content/rules/ucarp.org.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+    For other pureftpd.org subdomains, see PureFTPd.org.xml
```
J0WI: Please also add a comment on PureFTPd.org.xml (and DNSCrypt.org.xml if related)

J0WI:
Thanks for the corrections!

PR #5085, opened 2016-06-10
.onion and .i2p

galeksandrp:
Since Tor Browser include HTTPS Everywhere, we can add .onion rules.

Also tests for
- src/chrome/content/rules/Cyphdbyhiddenbhs.onion.xml
- src/chrome/content/rules/Blockchainbdgpzk.onion.xml
- src/chrome/content/rules/FacebookCoreWWWi.onion.xml
- src/chrome/content/rules/Y6xjgkgwj47us5ca.onion.xml

https://github.com/EFForg/https-everywhere/search?utf8=%E2%9C%93&q=onion+mixedcontent&type=Code


chris-barry:
Without something like a new "platform=tor" flag, this doesn't really make sense to bring in.

J0WI:
Nice suggestion, but you shouldn't mix this with ruleset changes.
Ping @Hainish @jsha for this new feature

jsha:
Hi @galeksandrp,

This is a cool idea, but it effectively would turn HTTPS Everywhere into a clearinghouse or authoritative source of information on which onion sites correspond to which real sites. We don't have the resources to a do a trustworthy job of maintaining such a list, so I'm afraid we can't do it.

J0WI:
@jsha I think HTTPS on .onion will do this anyway.
.onion certs are currently only available for some very well known hosts (e.g. Facebook). Some of them even put it together with other alt names to confirm the ownership.

chris-barry:
I've been brainstorming this for a while. Now seems like a good idea to
dump some thoughts!

There would have to be a standard way of verifying hosts (
https://example.com/onion.txt or similar) or a policy of only accepting
.onion's that have a certificate. This would limit it to, Facebook,
Blockchain.info, and The Intercept's SecureDrop instance. Manually
confirming sites doesn't work (I've tried, it's super annoying).

There would also have to be a UI change (similar to the block unsecured
http mode) that would allow users to opt in to onion sites as an
alternative. Along with this, an option like "tor=true" in the rulesets.

I am not sure how requested this feature is, but personally I think it
would be cool. Especially with second generation onion services coming up.
:)

On Jul 26, 2016 11:18, "J0WI" <notifications@github.com> wrote:

@jsha <https://github.com/jsha> I think HTTPS on .onion will do this anyway.
.onion certs are currently only available for some very well known hosts
(e.g. Facebook). Some of them even put it together with other alt names to
confirm the ownership.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/EFForg/https-everywhere/pull/5085#issuecomment-235301598>,
or mute the thread

fuglede:
There was a project -- based on HTTPS Everywhere -- which did exactly this: https://github.com/chris-barry/darkweb-everywhere

chris-barry:
That's me :)

fuglede:
Hah, cool. It's a really nifty idea. If a more-or-less standardized way of announcing hidden service hosts through HTTP/DNS, I imagine that we would quickly see directories of such announcements, and that the corresponding act of making a database for rewriting rules could be completely automated. automated.

chris-barry:
Yeah. Without some sort of way to have a clearnet address announce who they
are, I'm not sure how else it could scale (maybe an http X-onion?).
Verifying who owns what is kinda annoying :(

galeksandrp:
@chris-barry Could you please review rulesets (https://github.com/chris-barry/darkweb-everywhere/pull/47) and CI rulesets testing system (https://github.com/chris-barry/darkweb-everywhere/pull/50)?

libBletchley:
@fuglede @chris-barry :

> There was a project -- based on HTTPS Everywhere -- which did exactly this: https://github.com/chris-barry/darkweb-everywhere

That project is dead now b/c of a Firefox change, despite the usefulness it would have had to Chromium users.  So now what?  

@jsha 
> We don't have the resources to a do a trustworthy job of maintaining such a list, so I'm afraid we can't do it.

What if we accept the limitation that such a list is verifiable by a machine?  Certainly if an SSL cert mentions both the domain on the clearweb as well as the darkweb, then it's indisputably correct and also can be automated.  We could possibly also accept that if the HTTPS site on the clearweb references the `.onion`, then it's good enough.  In fact, the db could include a parameter like:
```
association_mechansim ::= (Site_Redirect | SSLcert | HTTPheader | HTTPcontent | manual)
```
and the user could control which verification mechanism satisfies their security policy.  Also remember that HTTPS-everywhere is just a hack for non-serious websites anyway.  Properly secured websites where security is important don't need HTTPS-everywhere rules.

BTW, it's interesting to notice what privacyinternational.org is doing.  If you go their their clearweb site, their server detects you are on Tor and automatically redirects to their `.onion` site.  It's the only website I've seen do that, which I think is quite slick.  It's the real solution -- but note that it's still only a partial solution because a DNS leak would have still occured if the browser is leaky.  However, HTTPS-everywhere _could_ have done that replacement so that even a leaky browser doesn't leak.  And HTTPS-everywhere could know for certain about the validity by running this:
```
$ torsocks curl --head https://www.privacyinternational.org 
HTTP/2 301 
server: nginx
date: Sat, 11 Nov 2017 11:41:22 GMT
content-type: text/html
content-length: 178
location: https://privacyintyqcroe.onion/
```

PR #5115, opened 2016-06-12
Create Worten.pt.xml

ekaris:


ekaris:
@J0WI good to go

J0WI:
This host serves HSTS header including all subdomains.
So non-working hosts are broken.

PR #5118, opened 2016-06-13
Pixiv.xml: workaround paths requested with CORS

Kovensky:
(At least) Chrome sends `Origin: null` with the rewritten request, which causes `Access-Control-Allow-Origin` to be omitted from the response, failing the CORS check.

Exclude URLs that are requested with CORS from the rules.

### DIFF
--- a/src/chrome/content/rules/Pixiv.xml
+++ b/src/chrome/content/rules/Pixiv.xml
@@ -38,6 +38,8 @@
 	<target host="sensei.pixiv.net" />
 	<target host="sketch.pixiv.net" />
 	<target host="source.pixiv.net" />
+	<!-- Exclude paths requested with CORS, would fail Access-Control-Allow-Origin check -->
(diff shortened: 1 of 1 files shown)


Kovensky:
The URLs having CORS issues are not public, so they cannot be accessed by the test robot.

Kovensky:
Firefox also fails to load due to a failed CORS check, but unlike Chrome, it doesn't actually log the failed request headers; but I suspect the problem is the same. Ironically, `fetch('http://source.pixiv.net/path/to/failed/resource.js', {mode: 'no-cors'})` without the exclusion rule actually sends the correct `Origin:` header on Firefox, returning a valid CORS response.

Kovensky:
Good news -- I managed to get the time to work around the rule's issues... by using https for CORS accesses to begin with. This exception is no longer needed.

The bug with (or at least undocumented behavior of) the extension URL rewriting clobbering Origin with "null" is still a potential problem, though.

PR #5131, opened 2016-06-16
Revert "Create Dota2lounge.com.xml"

fuglede:
Reverts EFForg/https-everywhere#4835 -- the ruleset causes mixed content issues:

skaermbillede fra 2016-06-16 22 03 41

Optimally, we would take that into account with a partial ruleset, but adding this for now as a quick fix.


J0WI:
Can we just set platform to mixedcontent?

fuglede:
@J0WI: Probably; I didn't go through the entire site at the time of [the comment](https://github.com/EFForg/https-everywhere/pull/4835#issuecomment-220374152), so this was just the quick fix.

fuglede:
Looks like that would work; added that to #5338 instead of here.

PR #5133, opened 2016-06-17
[OpenCCC] New ruleset

galenide:


src/chrome/content/rules/OpenCCC.xml
```diff
@@ -0,0 +1,26 @@
+<ruleset name="OpenCCC">
+
+	<target host="openccc.net" />
```
fuglede: https://openccc.net/ serves a 403, but http://openccc.net/ has content.

galenide: From what it appears, http://openccc.net/ generates a 302 redirect for https://www.openccc.net/uPortal/ At the current time, the ruleset has an exclusion that is intended to match http://openccc.net/ (when there is nothing after the last slash) with the intent that accessing the URL will lead to the redirect (or whatever other content might come along in the future) and not to a URL such as https://openccc.net/ that, as mentioned, serves a 403 error. Another option would be to explicitly redirect `^http://openccc\.net/?$` to `https://openccc.net/uPortal/`, though that might pose an issue if the OpenCCC site changes its pages in the future. Right now, there are no downgrades in the ruleset (such as `^https://openccc\.net/?$` redirected to `http://openccc.net/` ) in order to avoid making things more complicated.

fuglede: Ah, sorry, I had simply missed the exclusion. I agree that this approach is sensible.

PR #5136, opened 2016-06-18
[https-everywhere-rewriter] Rename URIjs to urijs

fuglede:
From `npm`:

    npm WARN deprecated URIjs@1.11.2: package renamed to "urijs" (lower-case), please update accordingly

### DIFF
--- a/rewriter/package.json
+++ b/rewriter/package.json
@@ -1,10 +1,10 @@
 {
-  "name" : "https-everywhere-rewriter",
-  "version" : "1.0.0",
-  "dependencies" : {
-    "xmldom" : "0.1.17",
-    "URIjs" : "1.11.2",
-    "readdirp" : "0.3.2",
-    "event-stream" : "3.0.20"
+  "name": "https-everywhere-rewriter",
+  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


jeremyn:
`npm` wants you to replace `URIjs` with `urijs`, so you also need to update `package.json`. From a fresh checkout, these commands will modify `package.json` for you:

```
$ npm uninstall --save URIjs
$ npm install --save urijs
```

I ran a small test with the rewriter using the new module, and it seems fine, but I don't know much about what the rewriter does. You should probably test it too.

fuglede:
Ah, thanks. Unfortunately, I'm not having much luck in terms of testing it these days, cf. #5137.

fuglede:
Disregard that, I managed to make it work, and indeed it appears to work as intended;

	https-everywhere/rewriter$ mkdir test
	https-everywhere/rewriter$ echo http://google.com > test/index.html
	https-everywhere/rewriter$ node rewriter.js test
	Loading rules...
	Rewriting test/index.html
	Found  http://google.com
	https-everywhere/rewriter$ cat test/index.html
	https://www.google.com/

PR #5149, opened 2016-06-19
Update and rename wifiwx.com.xml to WifiWX.com.xml

renyouguo:


src/chrome/content/rules/WifiWX.com.xml
```diff
@@ -0,0 +1,20 @@
...
+        Test:
+        http://app.wifiwx.com/download/android.php?ver=3.0
+        http://app.wifiwx.com/js/stat.js
+-->
+
+<ruleset name="WifiWX.com (partial)">
+        
+        <target host="wifiwx.com" />
+        <target host="event.wifiwx.com" />
+        <target host="html.wifiwx.com" />
```
J0WI: I get various 404 with this rule.
Same on hd. and fapi.

PR #5154, opened 2016-06-20
[freebsd.org] Update subdomains

youdly:
Add some previously non-functional subdomains.


J0WI:
WFM:
bugs.au.freebsd.org
www.bugs.au.freebsd.org

It seems that each community runs its own services. They all don't support HTTPS:
ftp.[lang].freebsd.org
mailman.[lang].freebsd.org
lists.[lang].freebsd.org
www.[lang].freebsd.org

Could you document this behavior?

J0WI:
Thanks!

PR #5157, opened 2016-06-20
Create Olx.pt.xml

ekaris:


src/chrome/content/rules/Olx.pt.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Mismatch:
+ help.olx.pt
+ 
+ Timeout:
+ www.olx.pt
+-->
+
+<ruleset name="Olx.pt">
+	<target host="olx.pt" />
```
J0WI: + yasukegroup.olx.pt

src/chrome/content/rules/Olx.pt.xml
```diff
@@ -0,0 +1,13 @@
+<!--
```
J0WI: Please use a consistent code style for comments.

ekaris:
Done

PR #5158, opened 2016-06-20
Update Jcloud.com.xml

renyouguo:


J0WI:
See also:
dataplus.jcloud.com
cpsc.jcloud.com (refused)
i.jcloud.com

PR #5159, opened 2016-06-20
Create Sfu.ca.xml

davidabian:


davidabian:
False negative?

J0WI:
retest

src/chrome/content/rules/Sfu.ca.xml
```diff
@@ -84,7 +84,6 @@
 	<target host="apps.code.sfu.ca" />
 	<target host="beediecommunity.sfu.ca" />
 	<target host="beedie.sfu.ca" />
-	<target host="blogs.sfu.ca" />
```
J0WI: you should add it as a comment

davidabian: Okay. I simply deleted it because that subdomain seems to have been "stopped", but it's true that blogs.sfu.ca is still reachable.

J0WI:
Thanks!

PR #5162, opened 2016-06-21
Add a HTTP->HTTPS redirect rule for hn.premii.com. Passes bash test.sh

aawc:


J0WI:
`reddit.premii.com` and `premii.com` also works.
There's an error on `www.premii.com`, could you document that? You can redirect it to non-www since they are aliases.

aawc:
Done. PTAL.

J0WI:
Thanks, LGTM!

PR #5164, opened 2016-06-21
alpinelinux.org

galeksandrp:


J0WI:
There are some more subdomains to redirect/document:
nl.alpinelinux.org
build.alpinelinux.org
distfiles.alpinelinux.org
dev.alpinelinux.org
patchwork.alpinelinux.org
git.alpinelinux.org
forum.alpinelinux.org
bugs.alpinelinux.org
pkgs.alpinelinux.org
lists.alpinelinux.org
wiki.alpinelinux.org

src/chrome/content/rules/Alpinelinux.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Alpinelinux.org">
+	<target host="alpinelinux.org" />
+	<target host="www.alpinelinux.org" />
+
+	<rule from="^http://www\.alpinelinux\.org/" to="https://alpinelinux.org/"/>
```
J0WI: We don't force redirects as long as everything works fine. It's up to the webmasters to do this.

galeksandrp:
Added.

PR #5171, opened 2016-06-21
rsdn.ru

galeksandrp:


jeremyn:
@galeksandrp This domain looks entirely gone. Please confirm and if that's right, close the issue, thanks.

galeksandrp:
They are moved to rsdn.org

galeksandrp:
They are moved to rsdn.org

jeremyn:
Thanks.

- [ ] Remove https://projects.rsdn.org and https://wiki.rsdn.org , they don't exist at all, even in DNS, so there's no need to mention them.
- [ ] https://rsdn.org and https://www.rsdn.org look like valid targets to me. I get a gray-lock-with-yellow-triangle in Firefox which means mixed content is not being blocked (see [here](https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox)). So, you can turn these into targets.

Because of the previous items:
- [ ] Delete the top comment
- [ ] Remove `(partial)` from the name

galeksandrp:
Main site have frames with http targets.

jeremyn:
Can you please be more specific about the problem?

galeksandrp:
Проекты > Nitra

Expected result: loading frame with http://confluence.jetbrains.com/display/Nitra
Actual result: mixed content.

https://rsdn.org/?http://confluence.jetbrains.com/display/Nitra
http://rsdn.org/?http://confluence.jetbrains.com/display/Nitra

jeremyn:
I see now, thanks. Can you add those URLs as examples to the comment like I wrote in the modified checklist [here](https://github.com/EFForg/https-everywhere/pull/5171#issuecomment-242920029)? Otherwise it might not be clear to someone looking at this ruleset how these domains are broken.

galeksandrp:
Added.

PR #5172, opened 2016-06-21
tldrlegal.com

galeksandrp:


J0WI:
Why is this rule partial?

J0WI:
Just oversaw community. which does now also mismatch.

galeksandrp:
Removed partial

PR #5173, opened 2016-06-21
kinopoisk.ru

galeksandrp:


J0WI:
Without the exclusion the test urls makes no longer sense.
Rule LGTM.

galeksandrp:
Done.

PR #5179, opened 2016-06-21
Update NetEase.xml

renyouguo:


J0WI:
This host seems to be currently down (502).

PR #5197, opened 2016-06-21
fixstars.com

galeksandrp:


J0WI:
I get redirected to http and there are also a few more subdomains (not supporting https).
Please check if this rule still make sense.

galeksandrp:
Improved.

J0WI:
The http redirect problem is not yet fixed.

galeksandrp:
Partial.

PR #5202, opened 2016-06-21
idms-linux.org

galeksandrp:


src/chrome/content/rules/idms-linux.org.xml
```diff
@@ -0,0 +1,7 @@
+<!--cdn. timed out-->
+<ruleset name="idms-linux.org">
+	<target host="idms-linux.org" />
+	<target host="www.idms-linux.org" />
+	<target host="wiki.idms-linux.org" />
```
J0WI: wiki redirects to http on other than /$
lists.idms-linux.org seems to work.

galeksandrp:
Added.

J0WI:
The http redirect problem is not yet fixed.

galeksandrp:
Moved to comments.

J0WI:
Now you also removed the comment for `cdn.idms-linux.org`

galeksandrp:
Added back.

J0WI:
Thanks!

PR #5210, opened 2016-06-21
minino.galpon.org

galeksandrp:


src/chrome/content/rules/galpon.org.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="minino.galpon.org">
+	<target host="minino.galpon.org" />
+	<target host="galpon.org" />
```
J0WI: please document the mismatch

src/chrome/content/rules/galpon.org.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="minino.galpon.org">
```
J0WI: You may rename to something like 'GALPon' to cover all included hosts

galeksandrp:
Documented.

PR #5211, opened 2016-06-21
neptuneos.com

galeksandrp:


src/chrome/content/rules/neptuneos.com.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="neptuneos.com">
+	<target host="neptuneos.com" />
+	<target host="www.neptuneos.com" />
```
J0WI: download.neptuneos.com

galeksandrp:
Added.

PR #5212, opened 2016-06-21
nethserver.org

galeksandrp:


src/chrome/content/rules/nethserver.org.xml
```diff
@@ -0,0 +1,12 @@
+<!--community. timed out
+wiki. mismatch
+docs. mismatch
+dev. refused
+mirror. mismatch-->
```
J0WI: see also mirror1.

Please document the mismatch on ^

galeksandrp:
Documented.

PR #5214, opened 2016-06-21
ourproject.org

galeksandrp:


J0WI:
Could you also check for subdomains here?

galeksandrp:
Added.

J0WI:
Great, thanks!

PR #5232, opened 2016-06-21
syllable.org

galeksandrp:


jeremyn:
Thanks. I didn't check all the non-English variants but I think all of these except https://forum.development.org have active mixed-content blocking. How do you want to handle that?

galeksandrp:
Ads only. We can prefer security over usability.

jeremyn:
We should definitely not break their ads.

galeksandrp:

jeremyn:
Thanks. Just to confirm, are you okay with squash-and-merge on this and other pull requests?

galeksandrp:
Yes.

jeremyn:
Thanks, merged.

PR #5234, opened 2016-06-21
sysresccd.org

galeksandrp:


jeremyn:
Thanks. https://sysresccd.org/ redirects to https://www.system-rescue-cd.org/ and there's a ruleset for that [here](https://github.com/EFForg/https-everywhere/blob/f163a0e59e3d65b4ef4860a33bf6e00ec6d92cce/src/chrome/content/rules/system-rescue-cd.org.xml). Please:

- [ ] Delete the sysresccd.org ruleset you added here
- [ ] Add targets for `sysresccd.org`, `www.sysresccd.org`, and `beta.sysresccd.org` to system-rescue-cd.org.xml
- [ ] Note `beta.system-rescue-cd.org` as having an invalid certificate in a top comment

galeksandrp:
Moved.

jeremyn:
Thanks. https://beta.sysresccd.org is actually okay and should have a target. If you watch the request in Firefox's [Network Monitor](https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor), you can see that https://beta.sysresccd.org works fine, and it's https://beta.system-rescue-cd.org/ with the bad certificate.

galeksandrp:
https://travis-ci.org/galeksandrp/https-everywhere/jobs/155594580

jeremyn:
Thanks @galeksandrp .

@Hainish Can you take a look at the [Travis failure](https://travis-ci.org/galeksandrp/https-everywhere/jobs/155594580) in the previous comment? The problem is that `fetch-test.sh` fails against

```xml
<target host="beta.sysresccd.org" />
```

but according to my browser, https://beta.sysresccd.org has a valid HTTPS configuration and 301-redirects to https://beta.system-rescue-cd.org/ , which has an invalid HTTPS configuration. So `fetch-test.sh` is incorrectly reporting an invalid HTTPS configuration for `beta.sysresccd.org` when actually it's the redirect target `beta.system-rescue-cd.org` with the problem.

fuglede:
From what I can tell, this is working as intended, as adding the relevant rule would cause breakage (even if the breakage is caused by a misconfiguration) without having a complementary downgrade rule. (The natural follow-up question is if the test is able to figure out what happens if a downgrade rule is present -- it probably won't, but in that case it has already done its job anyway.)

jeremyn:
@fuglede In this case the downgrade rule would be something like

```xml
<rule from="^https://beta\.sysresccd\.org/" to "http://beta.system-rescue-cd.org/" />
```

is that right?

fuglede:
I was actually thinking from https://beta.system-rescue-cd.org/ but both make sense. However, the thing is, if we add that one, we are effectively doing close to nothing in terms of security, but we risk end up in a situation where we make matters even worse if one day the site works properly, so I would opt to not do anything at all here, but just add a remark about the issue in the ruleset.

jeremyn:
@fuglede Thanks. I agree that we shouldn't do anything except note the problem.

@galeksandrp I've updated the to-do list in the earlier comment [here](https://github.com/EFForg/https-everywhere/pull/5234#issuecomment-242859670) based on the recent discussion.

galeksandrp:
Documented.

jeremyn:
Thanks, merged.

PR #5235, opened 2016-06-21
t2-project.org

galeksandrp:


src/chrome/content/rules/t2-project.org.xml
```diff
@@ -0,0 +1,7 @@
+<!--www2. nonexist-->
```
J0WI: Mismatch:
bugs.t2-project.org

Timeout:
torrent.t2-project.org

src/chrome/content/rules/t2-project.org.xml
```diff
@@ -0,0 +1,7 @@
...
+<ruleset name="t2-project.org">
+	<target host="t2-project.org" />
+	<target host="www.t2-project.org" />
+	<target host="dl.t2-project.org" />
```
J0WI: incomplete cert chain

galeksandrp:
Added.

PR #5245, opened 2016-06-21
univention.com

galeksandrp:


jeremyn:
Please rewrite this pull request to instead update [Univention.de.xml](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/Univention.de.xml).

galeksandrp:
Moved.

jeremyn:
Add targets:
- [ ] https://appcenter.software-univention.de
- [ ] https://docs.software-univention.de
- [ ] https://errata.software-univention.de
- [ ] https://ucc.software-univention.de
- [ ] https://ucc2.software-univention.de
- [ ] https://updates.software-univention.de
- [ ] https://forge.univention.org

Also:
- [ ] Change ruleset name to "Univention"
- [ ] Rewrite http://forge.univention.de to https://forge.univention.org
- [ ] Rewrite http://apt.univention.de to https://updates.software-univention.de
- [ ] Rewrite http://download.univention.de to https://updates.software-univention.de
- [ ] Change `securecookie` `host` to `.+`
- [ ] Delete all comments in the `ruleset` tag
- [ ] Delete [this part of the top comment](https://github.com/galeksandrp/https-everywhere/blob/cc12d9752b4e6803b9ef5aecb2965ca1270be010/src/chrome/content/rules/Univention.de.xml#L15-L20)
- [ ] Reorder targets and problematic subdomains comment by top-level domain (TLD) going left, so for example `.com` should be above `.de`, and `piwik.` should be directly above `www.piwik.`. Then move `^` and `www` to the top of their TLD group.

galeksandrp:
Nice work, @jeremyn!

jeremyn:
Thanks. Good catch on `ucc2`, but I notice now that `ucc` is also different content. I've updated the to-do list.

I see you sorted the targets. You put `software-univention` beneath `univention`, which is technically "wrong" but now I think it's better the way you have. Please be sure to sort the various problematic subdomains though.

galeksandrp:
Sorted.

jeremyn:
Thanks, but both the comment and the targets are still slightly unsorted. In both cases the top-level domains should go `com`, `de`, `org`. Also in the top please move `wiki.univention.de` directly under `sdb.univention.de`.

galeksandrp:
Sorted.

jeremyn:
Thanks! Merged.

PR #5250, opened 2016-06-21
workaround.ch

galeksandrp:


J0WI:
See also:
bitflu.workaround.ch
slackintosh.workaround.ch

src/chrome/content/rules/workaround.ch.xml
```diff
@@ -0,0 +1,8 @@
+<!-- bitflu. ERR_CONNECTION_REFUSED
```
J0WI: please use keyword refused

src/chrome/content/rules/workaround.ch.xml
```diff
@@ -0,0 +1,8 @@
...
+slackintosh. NET::ERR_CERT_AUTHORITY_INVALID -->
```
J0WI: please use keyword self-signed

galeksandrp:
Added.

PR #5255, opened 2016-06-22
smeserver.org

galeksandrp:


J0WI:
It seems that they moved to contribs.org

galeksandrp:
Added www. contribs.org ruleset #5822

J0WI:
You need to add www. as target here.

galeksandrp:
Added, thanks.

PR #5256, opened 2016-06-22
tfm.ro

galeksandrp:


src/chrome/content/rules/tfm.ro.xml
```diff
@@ -0,0 +1,11 @@
+<!--blog. mismatch
+cop. mismatch
+felicitari. mismatch
+portal. mismatch
+victor. mismatch
+download. mismatch
+stash. mismatch-->
+<ruleset name="tfm.ro">
+	<target host="tfm.ro" />
```
J0WI: Please also add `www.tfm.ro`

galeksandrp: Added.

PR #5258, opened 2016-06-22
webconverger.com

galeksandrp:


J0WI:
Could you please also document unsupported subdomains?

galeksandrp:
Documented.

J0WI:
See also:
archive.
attract.
cvt.
download.
eu.download.
prefs.
support.

galeksandrp:
Added.

PR #5270, opened 2016-06-22
playkey.net

galeksandrp:


src/chrome/content/rules/Playkey.net.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Playkey.net">
+	<target host="playkey.net" />
+	<target host="www.playkey.net" />
```
J0WI: Mismatch:
^
blog

Refused:
www

WFM:
api.playkey.net
support.playkey.net

galeksandrp:
Added.

PR #5273, opened 2016-06-22
Coverity.com

galeksandrp:


galeksandrp:
Hostname with platform '' shows up in 2 different rulesets

PR #5274, opened 2016-06-22
apiary.io

galeksandrp:


src/chrome/content/rules/apiary.io.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="apiary.io">
```
J0WI: Mismatch on docs.*.apiary.io (customer domains)

galeksandrp:
Documented.

PR #5279, opened 2016-06-22
david-dm.org

galeksandrp:


jeremyn:
With the target for https://www.david-dm.org and without the special rewrite rule, I get a simple certificate mismatch in Firefox, and using `fetch-test.sh` on Debian I get `(51, "SSL: no alternative certificate subject name matches target host name 'www.david-dm.org'")`.

Can you please simply change [that line](https://github.com/galeksandrp/https-everywhere/blob/af540a0101955fd9697ae0eef5f85e377f84ab1f/src/chrome/content/rules/david-dm.org.xml#L3) to this?

```xml
<!-- <target host="www.david-dm.org" /> --><!-- certificate mismatch -->
```

galeksandrp:
Done.

jeremyn:
[This](https://github.com/galeksandrp/https-everywhere/blob/a982dd3e040a39aac0bdc15267240bf4c3b95b3d/src/chrome/content/rules/david-dm.org.xml#L11) should be `<test>` not `<target>`.

galeksandrp:
Fixed, thanks!

PR #5281, opened 2016-06-22
freeboard.io

galeksandrp:


src/chrome/content/rules/freeboard.io.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="freeboard.io">
+	<target host="freeboard.io" />
+	<target host="www.freeboard.io" />
```
J0WI: mismatch on www

see also alpha.

galeksandrp:
Added.

PR #5282, opened 2016-06-22
hook.io

galeksandrp:


src/chrome/content/rules/hook.io.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="hook.io">
```
J0WI: Mismatch on geoip.hook.io

galeksandrp:
Added.

PR #5289, opened 2016-06-22
zenhub.io

galeksandrp:


src/chrome/content/rules/zenhub.io.xml
```diff
@@ -0,0 +1,4 @@
+<ruleset name="zenhub.io">
+	<target host="zenhub.io" />
```
J0WI: www.zenhub.io
api.zenhub.io
dashboard.zenhub.io
files.zenhub.io

galeksandrp:
Added.

PR #5297, opened 2016-06-23
cherry-pick Lshare.xml from #4609

J0WI:


jeremyn:
- [ ] The site doesn't identify itself as `L Share`. Change the ruleset name to `lsh.re` or `LinkShare`, and the filename to match.

J0WI:
done.

PR #5299, opened 2016-06-23
cherry-pick Telegram.wiki.xml from #4609

J0WI:


src/chrome/content/rules/Telegram.wiki.xml
```diff
@@ -0,0 +1,16 @@
...
+
+	Mismatch:
+		ip
+		data
+		www
+-->
+<ruleset name="Telegram Wiki">
+	<target host="telegram.wiki" />
+
+	<rule from="^http://www\.telegram\.wiki/"
```
fuglede: This rule will never go into effect as the host is not among the targets.

On an unrelated note, the title of the PR has a typo but that's what it is.

Otherwise LGTM.

J0WI:
I've added the missing target.

PR #5307, opened 2016-06-24
newsru.com

galeksandrp:


jeremyn:
Thanks for the ruleset!

My Firefox is unable to connect at all to https://newsru.com and https://www.newsru.com .

There are also a lot of subdomains as you can see [here](https://www.google.com/search?q=site%3Anewsru.com&ie=utf-8&oe=utf-8#q=site:*.newsru.com+-site:www.newsru.com). Can you go through this and add subdomains either as targets if they work or in a top comment if they don't?

galeksandrp:
Added.

jeremyn:
Thanks for adding all those subdomains. It's too bad more of them don't work.

Some minor things:

- [ ] I get a `timeout` for https://newsru.com instead of `refused`, but if you get `refused` then you can leave it, just let me know.
- [ ] `www.newsru.com` is in the top comment twice
- [ ] Can you add an example URL that's different for image.newsru.com ?

galeksandrp:
Yes, i get `refused`

jeremyn:
Thanks, merged.

PR #5315, opened 2016-06-24
rublacklist.net

galeksandrp:


src/chrome/content/rules/Rublacklist.net.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Rublacklist.net">
+	<target host="rublacklist.net" />
+	<target host="www.rublacklist.net" />
```
J0WI: reestr.rublacklist.net

Cloudflare error on talk.rublacklist.net

src/chrome/content/rules/Rublacklist.net.xml
```diff
@@ -0,0 +1,6 @@
...
+
+	<rule from="^http://(?:www\.)?rublacklist\.net/" to="https://rublacklist.net/" />
```
J0WI: www => ^ is not necessary, we let this up to the server.

PR #5316, opened 2016-06-24
sci-hub.cc

galeksandrp:


src/chrome/content/rules/Sci-hub.cc.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Sci-hub.cc">
+	<target host="sci-hub.cc" />
+	<target host="www.sci-hub.cc" />
+	<target host="sci-hub.ac" />
+	<target host="www.sci-hub.ac" />
+
+	<rule from="^http://(?:www\.)?sci-hub\.(?:cc|ac)/" to="https://sci-hub.ac/" />
```
J0WI: Please document why and remove non-capturing group.

J0WI:
It seems that several documents live on their own subdomain (where https is refused). Could you document this behavior?

galeksandrp:
Done.

J0WI:
I meant a generic hint, that the mirrors of the documents live on their own subdomain where https is refused, because there might be plenty of them.

galeksandrp:
Documented.

J0WI:
It seems that some aliases are currently down.

galeksandrp:
Documented.

PR #5323, opened 2016-06-25
tjournal.ru

galeksandrp:


src/chrome/content/rules/Snob.ru.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Snob.ru">
+	<target host="snob.ru" />
+	<target host="www.snob.ru" />
```
J0WI: stellaclar.snob.ru
shop.snob.ru

galeksandrp:
Added.

PR #5328, opened 2016-06-25
[xkcd] Downgrade https://c.xkcd.com/ to http

mnordhoff:
https://xkcd.com/ now includes a protocol-relative link to `//c.xkcd.com/random/comic/`; since https://c.xkcd.com/ has a [problematic certificate chain (among other issues)](https://www.ssllabs.com/ssltest/analyze.html?d=c.xkcd.com), it needs to be downgraded.

### DIFF
--- a/src/chrome/content/rules/xkcd.xml
+++ b/src/chrome/content/rules/xkcd.xml
@@ -56,7 +56,6 @@
 	<target host="xkcd.com" />
 	<target host="3d.xkcd.com" />
 	<!--
-	<target host="c.xkcd.com" />
 	<target host="c0.xkcd.com" />
 	<target host="c1.xkcd.com" />
 	<target host="c2.xkcd.com" />
@@ -79,6 +78,7 @@
 
 	<!--	Complications:
 				-->
+	<target host="c.xkcd.com" />
 	<target host="www.store.xkcd.com" />
 
 	<target host="xkcd.org" />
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/xkcd.xml
```diff
@@ -88,15 +88,22 @@
...
 
 	<rule from="^http://(www\.|m\.)?xkcd\.org/"
 		to="https://$1xkcd.com/" />
 
+	<rule from="^https://c\.xkcd\.com/"
+		to="http://c.xkcd.com/"
+		downgrade="1" />
+
+		<test url="https://c.xkcd.com/" />
+
```
Hainish: Why do an active downgrade rather than just excluding `c.xkcd.com` from the ruleset?  I'm getting a valid SHA-1 signed cert, which is problematic but better than plain http.

mnordhoff: @Hainish Because http://xkcd.com/ is rewritten to https://xkcd.com/ and that page includes a link to `//c.xkcd.com/random/comic/`, which is effectively forced to https unless there's an active downgrade.

The certificate is valid, but the [certificate chain isn't complete](https://www.ssllabs.com/ssltest/analyze.html?d=c.xkcd.com) so it won't work on all clients. (It uses a popular CA, so it will probably work on almost all of them, though.)

SSL Labs shows that the server is also vulnerable to an old TLS MITM attack, so the value of https is somewhat diminished anyway.

Hainish: @mnoordhoff: can you verify that the following is true:

It seems like we may be reducing the security of users in certain contexts here, but also making the site functional in other contexts where either (a) the intermediate certificate is not cached or (b) the browser disallows certain weaker TLS configurations

Hainish: If the above is true it seems like an acceptable tradeoff to me and I'm fine with merging

mnordhoff: @Hainish (a) Right. The site fails for clients that don't already have the intermediate cached. (It's a popular CA, so almost all clients probably already do.

(b) I don't think any current clients block the insecure TLS setup. It's functional, just not safe.

Downgrading reduces security for clients that have the intermediate and aren't being actively MITMed in a certain way, but... c'est la vie. Have to be functional for everyone.

PR #5335, opened 2016-06-26
Create WallstreetCN.xml

renyouguo:


src/chrome/content/rules/WallstreetCN.xml
```diff
@@ -0,0 +1,38 @@
...
+	- (markets.)?s.wallstcn.com
+	- static.wallstcn.com
+
+-->
+
+<ruleset name="WallstreetCN (partial)">
+
+	<!--	Complication	-->
+
+	<target host="*.wallstreetcn.com" />
```
J0WI: a wildcard would also include all mismatched hosts from above

J0WI:
I also found live. and m.

PR #5340, opened 2016-06-26
[UserScape] Redirect ^ to www

fuglede:
Currently, userscape.com serves 404s on all HTTPS requests. This takes care of the issue by redirecting all traffic to www.

### DIFF
--- a/src/chrome/content/rules/UserScape.xml
+++ b/src/chrome/content/rules/UserScape.xml
@@ -1,19 +1,22 @@
 <!--
 	Problematic subdomains:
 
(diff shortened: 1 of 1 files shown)


J0WI:
There are some more to document:

Mismatch:
www.ww.w.userscape.com
www.ww.userscape.com
ww.userscape.com
drop.userscape.com

404:
blog.userscape.com

fuglede:
Updated.

J0WI:
the weird www-combination subdomains seems to be deprecated, so go for it.

Issue #1421, opened 2015-04-10
Ooyala breaks video players


tancrackers:
Website:
http://www.mediaite.com/tv/geraldo-dashcam-video-gets-officer-slager-off-his-murder-charge/

When Ooyala is enabled on Https-everywhere, the flash player does not run.
When Ooyala is disabled, it runs.

PR #5345, opened 2016-06-26
Ooyala:

numismatika:
- add opf & player subdomain
- ^ serves a valid cert

### DIFF
--- a/src/chrome/content/rules/Ooyala.xml
+++ b/src/chrome/content/rules/Ooyala.xml
@@ -39,6 +39,8 @@
 	<target host="extras.ooyala.com" />
 	<target host="info.ooyala.com" />
 	<target host="l.ooyala.com" />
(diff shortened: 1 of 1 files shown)


J0WI:
Could you reconsider the mixed content problems?

numismatika:
You mean removing that platform-tag?

J0WI:
yes

numismatika:
I only checked one website where it is used so i can't safely say it is fine to remove that tag.
We should find more sites with ooyala and check there before we remove it.

J0WI:
The comment says it's just on www and only fonts and ads.
Those are both trivial I think.

numismatika:
Yea but the reference to #1421 worries me.

J0WI:
That's likely CORS, but I also don't want to break anything, so LGTM.
Thanks for updating!

PR #5347, opened 2016-06-26
add rule for SpeedBurger.com

numismatika:


J0WI:
wrong content is served on speed-burger.com over https (redirects on http).

numismatika:
^ is rewritten to www which works fine here.
Where exactly did you get a problem?

J0WI:
Could you document that?
Also `-` don't need to be escaped. Please also add trailing slashes to the urls.

J0WI:
Thanks!

PR #5351, opened 2016-06-26
Vox-Media:

numismatika:
- merge Vox-CDN into Vox-Media
- ea-cdn has a proper cert now
- add volume cdn-subdomain


J0WI:
Thanks!

PR #5358, opened 2016-06-27
Added vodkamethod.com

herrjemand:


J0WI:
As on you other PRs: It would be great if you would support preloaded HSTS. :)

herrjemand:
@J0WI Will sort it out in next week. Ta

PR #5360, opened 2016-06-27
Updated nieman.de

herrjemand:


J0WI:
Adding your domains to the preloaded HSTS list would make this rule obsolete: https://hstspreload.appspot.com/

You may also like to add jeman.de.

PR #5361, opened 2016-06-27
Update 163.com-mixedcontent.xml

renyouguo:


renyouguo:
@J0WI Done.

J0WI:
So you worked around MCB problems on all rules here? Then you should merge it with `163.com.xml`.
Or you rename this in something like `163.com-problematic.xml` to avoid having everything disabled when it's buggy, which happens very likely in such difficult rules.

src/chrome/content/rules/163.com-mixedcontent.xml
```diff
@@ -1,23 +1,125 @@
...
+	
+		<exclusion pattern="^http://1.163.com/$" />
+		
+		<rule from="^http://1\.163\.com/favicon\.ico/"
+			to="https://1.163.com/favicon.ico/" />
+			<test url="http://1.163.com/favicon.ico/" />
+		
+		<rule from="^http://1\.163\.com/(appguide|goods|hd|kpimission|user)/(.+)\.do"
+			to="https://1.163.com/$1/$2.do" />
+			<test url="http://1.163.com/appguide/getconfig.do" />
```
J0WI: You are including every *.do, so you also need to test more than one *.do for each path.

J0WI:
Please also update the reference in 163.com.xml

renyouguo:
@J0WI Done

PR #5363, opened 2016-06-27
ComputerBase remove rewrite Restrictions

numismatika:
As of today CB has enabled TLS on the whole site for all users, not just those that have a paid subscription.


J0WI:
Could you also add `computerbase.de` and `m.computerbase.de`?

J0WI:
Thanks!

PR #5392, opened 2016-06-29
mir-politika.ru

galeksandrp:


src/chrome/content/rules/mir-politika.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="mir-politika.ru">
+	<target host="mir-politika.ru" />
+	<target host="www.mir-politika.ru" />
```
J0WI: list.mir-politika.ru (mismatch)
mir-politika.com may be related.

galeksandrp:
Added.

PR #5402, opened 2016-06-29
kr-znamya.ru

galeksandrp:


src/chrome/content/rules/kr-znamya.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="kr-znamya.ru">
+	<target host="kr-znamya.ru" />
+	<target host="www.kr-znamya.ru" />
+	<rule from="^http://www\.kr-znamya\.ru/"
```
J0WI: Please document the mismatch on www.
Since this site sends a (not yet preloaded) HSTS header, you may also rewrite https://www to https://^.

PR #5405, opened 2016-06-29
kbpravda.ru

galeksandrp:


src/chrome/content/rules/kbpravda.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="kbpravda.ru">
+	<target host="kbpravda.ru" />
+	<target host="www.kbpravda.ru" />
```
J0WI: kbp.kbpravda.ru
www.kbp.kbpravda.ru

J0WI:
There is some mixed content. Please check if it breaks any functionality.

galeksandrp:
Nothing breaks.

PR #5411, opened 2016-06-29
poltava.to

galeksandrp:


src/chrome/content/rules/poltava.to.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="poltava.to">
+	<target host="poltava.to" />
+	<target host="www.poltava.to" />
```
J0WI: blog.poltava.to

J0WI:
These sites have mixed content issues.

galeksandrp:

PR #5416, opened 2016-06-29
sobytiya.info

galeksandrp:


J0WI:
Please also add doska.sobytiya.info

Please also document, that the server serves wrong content when using https on the following hosts:
afisha.sobytiya.info
en.sobytiya.info

J0WI:
Thanks!

PR #5423, opened 2016-06-29
u-f.ru

galeksandrp:


src/chrome/content/rules/u-f.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="u-f.ru">
+	<target host="u-f.ru" />
+	<target host="www.u-f.ru" />
```
J0WI: uf.u-f.ru

J0WI:
These sites have mixed content errors.

galeksandrp:

PR #5431, opened 2016-06-29
elec.ru

galeksandrp:


src/chrome/content/rules/elec.ru.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="elec.ru">
+	<target host="elec.ru" />
+	<target host="www.elec.ru" />
+	<target host="market.elec.ru" />
+	<target host="top.elec.ru" />
+	<target host="cdn.elec.ru" />
```
J0WI: static.elec.ru

Mismatch on:
advert.elec.ru
www.advert.elec.ru

J0WI:
Thanks!

PR #5433, opened 2016-06-29
kremlinpress.ru

galeksandrp:


J0WI:
cert expired recently.

J0WI:
This has now a valid cert.
But why is this rule partial?

galeksandrp:
Fixed.

PR #5439, opened 2016-06-29
rns.online

galeksandrp:


src/chrome/content/rules/rns.online.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="rns.online">
+	<target host="rns.online" />
+	<target host="www.rns.online" />
+	<target host="cdn.rns.online" />
+	<test url="http://cdn.rns.online" />
+	<test url="http://cdn.rns.online/" />
+	<test url="http://cdn.rns.online?" />
+	<test url="http://cdn.rns.online#" />
+	<test url="http://cdn.rns.online/?" />
+	<exclusion pattern="^http://cdn\.rns\.online/?(?:$|\?|#)" />
```
J0WI: A trailing slash is required.
Please also avoid non-capturing groups, as our code style recommends.

galeksandrp:
Improved.

J0WI:
There are still test urls without trailing slashes. That's wrong, because the `?` and `#` chars are not part of the TLD.
Please also remove the change in translations.

galeksandrp:
Improved.

src/chrome/content/rules/rns.online.xml
```diff
@@ -0,0 +1,16 @@
...
+	<target host="m.rns.online" />
+	<test url="http://cdn.rns.online/" />
+	<test url="http://cdn.rns.online/?" />
+	<test url="http://cdn.rns.online/?q" />
+	<test url="http://cdn.rns.online/?s" />
+	<test url="http://cdn.rns.online/#" />
+	<test url="http://cdn.rns.online/#q" />
+	<exclusion pattern="^http://cdn\.rns\.online/?$" />
+	<exclusion pattern="^http://cdn\.rns\.online/?\?" />
+	<exclusion pattern="^http://cdn\.rns\.online/?#" />
```
J0WI: `?` need to be escaped or removed. Exclusions without trailing slash are not allowed.

galeksandrp: Why?

J0WI: https://stackoverflow.com/questions/1617058/ok-to-skip-slash-before-query-string#answer-1617074

galeksandrp: Trailing slash forced ins exclusions.

J0WI:
Thanks!

PR #5440, opened 2016-06-29
trend.az

galeksandrp:


J0WI:
There is some non-trivial mixed content.

J0WI:
@galeksandrp would you like to fix this?

galeksandrp:
Ad and share button counter.

PR #5452, opened 2016-06-29
interfax.kz

galeksandrp:


src/chrome/content/rules/interfax.kz.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="interfax.kz">
+	<target host="interfax.kz" />
+	<target host="www.interfax.kz" />
+	<target host="press.interfax.kz" />
+	<target host="uploads.interfax.kz" />
```
J0WI: religion.interfax.kz
news.interfax.kz

PR #5455, opened 2016-06-29
prichernomorie.com.ua

galeksandrp:


src/chrome/content/rules/prichernomorie.com.ua.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="prichernomorie.com.ua">
+	<target host="prichernomorie.com.ua" />
+	<target host="www.prichernomorie.com.ua" />
+	<target host="en.prichernomorie.com.ua" />
```
J0WI: wrong content is served here.

J0WI:
This site has mixed content issues.

galeksandrp:

PR #5463, opened 2016-06-29
sakhalin.info

galeksandrp:


J0WI:
sakh.com seems related, can you check that?

src/chrome/content/rules/sakhalin.info.xml
```diff
@@ -0,0 +1,6 @@
+<!--9may. refused-->
```
J0WI: same on 70let.

galeksandrp:
Added.

J0WI:
Thanks!

PR #5473, opened 2016-06-29
euroradio.fm

galeksandrp:


fuglede:
Lots of (false) mixed content blocking on this one.

galeksandrp:

src/chrome/content/rules/euroradio.fm.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+m.euroradio.fm ¹
```
J0WI: ihwc2015.euroradio.fm

galeksandrp: Added.

PR #5478, opened 2016-06-29
7x7-journal.ru

galeksandrp:


src/chrome/content/rules/7x7-journal.ru.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="7x7-journal.ru">
+	<target host="7x7-journal.ru" />
+	<target host="www.7x7-journal.ru" />
+	<target host="static.7x7-journal.ru" />
+	<rule from="^http://www\.7x7-journal\.ru/"
```
J0WI: Please document mismatches on www. and old.

J0WI:
For future comimits: please try to format documented subdomains in a list like [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/FreeBSD.xml) for better readability.

PR #5479, opened 2016-06-29
autosport.com.ru

galeksandrp:


src/chrome/content/rules/autosport.com.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="autosport.com.ru">
+	<target host="autosport.com.ru" />
+	<target host="www.autosport.com.ru" />
+	<rule from="^http://www\.autosport\.com\.ru/"
```
J0WI: Please document the mismatch.

PR #5485, opened 2016-06-29
dp.ru

galeksandrp:


J0WI:
This site has mixed content issues.

J0WI:
@galeksandrp for all the mixed content issues:
The easiest way is to add `platform="mixedcontent"`. A more complex solution is to made specific exclusions and redirect just loaded resources.

galeksandrp:
Partial.

PR #5490, opened 2016-06-29
finam.ru

galeksandrp:


J0WI:
Can you document that none of the region subdomains support https (refused)?
old., partners. and bank. serve a mismatching cert, node. is refused, news.net.finam.ru may work.

On some sites, eg on shop., are some mixed active content issues.

Please move all comments on the top and sort the hosts alphabetically.

galeksandrp:
Added.

J0WI:
Thanks!

PR #5492, opened 2016-06-29
kodeks.ru

galeksandrp:


fuglede:
It looks like both https://kodeks.ru and https://www.kodeks.ru serve only a simple "It works!" message, and that static content, such as https://www.kodeks.ru/assets/images/banner/6_2016_k.png, can also not be secured, so unfortunately there's probably not much to gain here.

PR #5498, opened 2016-06-29
newtonew.com

galeksandrp:


src/chrome/content/rules/newtonew.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="newtonew.com">
+	<target host="newtonew.com" />
+	<target host="www.newtonew.com" />
+	<rule from="^http://www\.newtonew\.com/"
```
J0WI: Please document the mismatch on www

PR #5505, opened 2016-06-29
startsmile.ru

galeksandrp:


J0WI:
It seems that there are much more subdomains. Is there any risk to use a wildcard here?

galeksandrp:
Added.

PR #5509, opened 2016-06-29
afisha.ru

galeksandrp:


src/chrome/content/rules/afisha.ru.xml
```diff
@@ -0,0 +1,18 @@
...
+gorod. expired
+eda. mismatch
+lenovomotofest. expired
+mag. expired
+staytrue. expired
+email. timed out
+m. expired
+ios. mismatch
+email.newsletters. timed out
+vozduh. expired-->
```
J0WI: This seems not up to date. At least ^, www and m are working for me.

PR #5518, opened 2016-06-29
topwar.ru

galeksandrp:


jeremyn:
Thanks.

- [ ] https://forum.topwar.ru and https://cdn.topwar.ru work for me, you can uncomment them
- [ ] Note in a top comment that https://www.topwar.ru , https://mt.topwar.ru , and https://wf.topwar.ru have certificate mismatches

galeksandrp:
Added.

jeremyn:
https://forum.topwar.ru/ is still valid HTTPS, even though it currently redirects immediately to http.

src/chrome/content/rules/topwar.ru.xml
```diff
@@ -1,8 +1,17 @@
+<!--
+www.topwar.ru ¹
+forum.topwar.ru ⁶
```
galeksandrp: Yes.

jeremyn:
If the HTTPS is valid then why not include it as a target?

galeksandrp:
Redirect loop? Even if it mitigated after ~3loops by HTTSE, it is increase page load time.

jeremyn:
Oops, you're right. I didn't think about it looping. It should stay commented.

jeremyn:
Thanks, merged.

PR #5533, opened 2016-06-29
kavkaz-uzel.ru

galeksandrp:


J0WI:
This seems related to `kavkaz-uzel.eu.xml`

src/chrome/content/rules/kavkaz-uzel.ru.xml
```diff
@@ -1,4 +1,6 @@
 <!--
+Related: kavkaz-uzel.eu.xml
```
J0WI: The reference needs also to be added in `kavkaz-uzel.eu.xml`.

PR #5539, opened 2016-06-29
maam.ru

galeksandrp:


J0WI:
I also get various `\d{4}.maam.ru` hosts which are serving a mismatching cert. Could you document this behavior?

PR #5543, opened 2016-06-29
nordportal.ru

galeksandrp:


J0WI:
I also found doska.nordportal.ru and pressa.nordportal.ru, please document the mismatch there.
Please also check if the mixed content on this site breaks functionality.

J0WI:
What about the mixed content issues?

galeksandrp:
Single advert affected.

PR #5548, opened 2016-06-29
saint-petersburg.ru

galeksandrp:


src/chrome/content/rules/saint-petersburg.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="saint-petersburg.ru">
+	<target host="saint-petersburg.ru" />
```
J0WI: m.saint-petersburg.ru

J0WI:
There are some mixed content issues. Please check if they are breaking anything.
Please also document the issue on www.

J0WI:
What about mixed content here?

galeksandrp:
1 adv affected.

PR #5549, opened 2016-06-29
stadium.ru

galeksandrp:


src/chrome/content/rules/stadium.ru.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="stadium.ru">
+	<target host="stadium.ru" />
+	<target host="www.stadium.ru" />
+	<!-- <target host="archive.stadium.ru" /> --><!-- ERROR (35, schannel: SNI or certificate check failed: SEC_E_WRONG_PRINCIPAL (0x80090322) -    .) -->
```
J0WI: I also get a mismatch on new.stadium.ru

PR #5552, opened 2016-06-29
sibinfo.su

galeksandrp:


J0WI:
mixed content here seems to be trivial, so LGTM

PR #5554, opened 2016-06-29
tyumedia.ru

galeksandrp:


src/chrome/content/rules/tyumedia.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="tyumedia.ru">
+	<target host="tyumedia.ru" />
+	<target host="www.tyumedia.ru" />
+	<rule from="^http://www\.tyumedia\.ru/"
```
J0WI: Please document the mismatch on www

PR #5555, opened 2016-06-29
tverweek.com

galeksandrp:


src/chrome/content/rules/tverweek.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="tverweek.com">
+	<target host="tverweek.com" />
+	<target host="www.tverweek.com" />
+	<rule from="^http://www\.tverweek\.com/"
```
J0WI: Please document the mismatch on www

PR #5567, opened 2016-06-29
grad.ua

galeksandrp:


J0WI:
This site has mixed content issues and serves an invalid cert in main.grad.ua

src/chrome/content/rules/grad.ua.xml
```diff
@@ -0,0 +1,6 @@
+<!-- main. NET::ERR_CERT_AUTHORITY_INVALID -->
```
J0WI: please use human readable keywords like 'mismatched', 'refused', 'expired' or 'self-signed'.

PR #5573, opened 2016-06-29
pgtrk.ru

galeksandrp:


src/chrome/content/rules/pgtrk.ru.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="pgtrk.ru">
+	<target host="tv.pgtrk.ru" />
```
J0WI: tv.mobile.pgtrk.ru

J0WI:
Please also document the behavior of pgtrk.ru an www.pgtrk.ru

PR #5577, opened 2016-06-29
tvcom-tv.ru

galeksandrp:


src/chrome/content/rules/tvcom-tv.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="tvcom-tv.ru">
+	<target host="tvcom-tv.ru" />
+	<target host="www.tvcom-tv.ru" />
```
J0WI: The server redirects wrong here. This needs to be redirected to non-www.

J0WI:
Please also document the issue.

galeksandrp:
Added.

PR #5578, opened 2016-06-29
arigus-tv.ru

galeksandrp:


src/chrome/content/rules/arigus-tv.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="arigus-tv.ru">
+	<target host="arigus-tv.ru" />
+	<target host="www.arigus-tv.ru" />
```
J0WI: Mismatch on beta.arigus-tv.ru

galeksandrp:
Added.

PR #5579, opened 2016-06-29
topspb.tv

galeksandrp:


src/chrome/content/rules/topspb.tv.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="topspb.tv">
+	<target host="topspb.tv" />
+	<target host="www.topspb.tv" />
+	<target host="starts.topspb.tv" />
```
J0WI: server1.topspb.tv
server2.topspb.tv

PR #5582, opened 2016-06-29
tkgorod.ru

galeksandrp:


src/chrome/content/rules/tkgorod.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="tkgorod.ru">
+	<target host="tkgorod.ru" />
+	<target host="www.tkgorod.ru" />
+	<rule from="^http://www\.tkgorod\.ru/"
```
J0WI: redirecting www to ^ is not necessary.

PR #5583, opened 2016-06-29
ictv.ua

galeksandrp:


J0WI:
You may check for:
orator.ictv.ua
distalo.ictv.ua
dacha.ictv.ua
pogoda.ictv.ua
st1.ictv.ua
comic.ictv.ua

Please resort the targets and comments that they fit our recommend style guide.

galeksandrp:
Added.

PR #5584, opened 2016-06-29
Fixes and documentation for development on Vagrant with Windows 10 host/Fedora 24 Cloud guest

jeremyn:
I've been working on a development environment for HTTPS-Everywhere running with Vagrant. I'm on a Windows 10 host and setting up a [Fedora 24 Cloud](https://atlas.hashicorp.com/fedora/boxes/24-cloud-base) guest. Why not Ubuntu 16.04? Because the ubuntu/xenial64 box [has problems](https://bugs.launchpad.net/cloud-images/+bug/1565985). Why not Ubuntu 14.04? Because I need SMB support and that seems to require a [more recent kernel](https://github.com/mitchellh/vagrant/issues/6188#issuecomment-171396607).

My working environment is Git (Bash) for Windows 2.9.0.

This pull request fixes some documentation and adds missing dnf dependencies.

Here are various other issues I encountered that should be documented somewhere. I can write the documentation if you can tell me where to put it.

The current default folder sync method for this box is `rsync`, but Git Bash doesn't include rsync so that doesn't work. The usual `virtualbox` method doesn't work either because HTTPS-Everywhere uses symlinks and symlinks don't work correctly on Windows with the `virtualbox` method. `smb` (samba) is needed. This requires some preconfiguration, see [this Vagrant issue](https://github.com/mitchellh/vagrant/issues/7532) I submitted. Due to [this other Vagrant issue](https://github.com/mitchellh/vagrant/issues/7534), I started the box in admin PowerShell and then accessed it in Git Bash by directly sshing into it (`ssh -p 2222 vagrant@127.0.0.1`) instead of using `vagrant ssh`.

The default RAM allocated by the Fedora box is 1024MB. This is not enough to even compile lxml during the `install-dev-dependencies.sh` step, but 2048MB is enough for that. Later I kept getting `git push` timeouts due to the extensive pre-push testing even with 8192MB allocated. I ended up just unlinking the pre-push tests so I could successfully push at all.

The `pycurl` that pip installs doesn't work. You get this error in a Python interpreter:

```
>>> import pycurl
  File "<stdin>", line 1, in <module>
ImportError: pycurl: libcurl link-time ssl backend (nss) is different from compile-time ssl backend (none/other)
>>>
```

The fix is to first remove pycurl with `pip uninstall pycurl`, and then either

```bash
$ sudo dnf install python-pycurl
```

or

```bash
$ export PYCURL_SSL_LIBRARY=nss
$ pip install --user --no-allow-insecure --no-allow-external --compile pycurl
```

Chromium isn't available for Fedora so I installed Google Chrome by downloading the rpm from Google's site. I downloaded `chromedriver` from https://sites.google.com/a/chromium.org/chromedriver/downloads as described in [test/chromium/README.md](https://github.com/jeremyn/https-everywhere/blob/8712bf84858fd4d31b04011dc889767723ba85e8/test/chromium/README.md)

At this point `bash test.sh` and the pre-push hook should work. Running the tests in other ways might not work. In particular I ran into trouble trying to run `test/chromium/script.py` headless without 
[xvfb](https://github.com/EFForg/https-everywhere/blob/master/test/chromium.sh#L42). Instead I tried using `pyvirtualdisplay` as described [here](https://github.com/EFForg/https-everywhere/blob/master/test/chromium/script.py#L11) but eventually got hangs/failures that I couldn't get past. In case it helps you, for reference, I tried pip installing `pyvirtualdisplay` and dnf installing `python-xvfbwrapper` and `dbus-x11`. The dnf stuff might have been useless. The `pyvirtualdisplay` reference in `test/chromium/script.py` should probably either be improved or deleted entirely.

### DIFF
--- a/install-dev-dependencies.sh
+++ b/install-dev-dependencies.sh
@@ -22,7 +22,8 @@ elif type brew >/dev/null ; then
     echo '/usr/local/bin not found in $PATH, please add it.'
   fi
 elif type dnf >/dev/null ; then
-  sudo dnf install libxml2-devel python-devel libxslt-devel
+  sudo dnf install firefox gcc git libcurl-devel libxml2-devel libxslt-devel \
+    python-devel redhat-rpm-config xorg-x11-server-Xvfb
 fi
 
 # Get the addon SDK submodule and rule checker
--- a/test/chromium/README.md
+++ b/test/chromium/README.md
@@ -5,4 +5,4 @@
       - Install Selenium as a python package using ```pip install selenium```, or run install-dev-dependencies.sh and it will do the job
 - ChromeDriver
       - Run ```apt-get install chromium-chromedriver```, or run install-dev-dependencies.sh
-      - Or, manually download ChromeDriver from https://sites.google.com/a/chromium.org/chromedriver/downloads. Extract the executable to /usr/lib/chromium-browser/, so that the pasted executable's full path becomes /usr/lib/chromium-browser/chromedriver
+      - Or, manually download ChromeDriver from https://sites.google.com/a/chromium.org/chromedriver/downloads. Extract the executable to /usr/lib/chromium-browser/ (Ubuntu), /usr/lib/chromium/ (Debian), or /usr/bin/ (other) , so that the pasted executable's full path becomes /usr/lib/chromium-browser/chromedriver (for Ubuntu, otherwise substitute the directory as needed).
--- a/test/chromium/script.py
+++ b/test/chromium/script.py
@@ -45,7 +45,7 @@ except WebDriverException as e:
     error = e.__str__()
 
     if "executable needs to be in PATH" in e.__str__():
-        print "ChromeDriver isn't installed. Check test/chrome/README.md " \
+        print "ChromeDriver isn't installed. Check test/chromium/README.md " \
               "for instructions on how to install ChromeDriver"
 
         sys.exit(2)


jeremyn:
Also: the folder sync has to be not just `smb`, but also have `mfsymlinks` enabled:

```ruby
config.vm.synced_folder ".", "/vagrant", type: "smb", mount_options: ["mfsymlinks"]
```

`smb` makes existing symlinks work, but `mfsymlinks` lets us create new ones, which `install-dev-dependencies.sh` will try to do.

Hainish:
I'm still reviewing this, I'll finish early next week.  I'll be adding packages so that running `./install-dev-dependencies.sh` from a barebones Fedora 24 docker instance results in tests being able to be run.  This should not impede desktop Fedora installations from running `./install-dev-dependencies.sh`.

jeremyn:
You mention Docker instead of Vagrant, and also [in this comment on issue #5628](https://github.com/EFForg/https-everywhere/issues/5628#issuecomment-236034570) you advised me to use Docker for development. Vagrant seems like a better fit to me though, or maybe that's just because I know Vagrant better. For example, syncing files between the host and guest seems more straightforward with Vagrant.

Do you have any direct experience or notes that you could share in setting up a Docker development environment on Windows for HTTPS-Everywhere, similar to my above notes for Vagrant? I see https://github.com/EFForg/https-everywhere-docker-base but it's not clear to me from that how I would, for example, do a live Samba share so I could edit and git-push from the Windows host.

Hainish:
I don't want to prescribe any particular development environment.  I've used Vagrant in the past, and I tend to do development in Docker now because of the quick bootstrapping of containers and small memory footprint (plus I develop from within Qubes, so Vagrant on top of that would be double-layered VMs, which may be possible but [seems complicated](https://groups.google.com/forum/#!topic/qubes-users/utMyWi__IdI).  It's best if we have devs running similar development environments just so we can reproduce issues easier, but I don't have anything against using Vagrant for development per se.

For that reason, if we can support Docker without impacting support for Vagrant, this would be ideal.  I don't know what the best way to set up Docker on Windows is, but these instructions may help https://docs.docker.com/engine/installation/windows/

jeremyn:
Docker/Vagrant configuration is only crucial for Windows users because so much of the current setup assumes *nix tooling. But as far as I can tell, there is no supported development environment on Windows. I also assume Windows isn't normally considered by you or other core developers. For example, there's a symlink from `rules` to `src/chrome/content/rules` that's just a convenience for *nix users but an annoying problem for Windows users.

So the real question isn't Docker or Vagrant, but what development process is supported for contributors using Windows, particularly casual contributors. By support I don't mean answering questions so much as making an effort not to change anything that will break this supported process.

Hainish:
https://github.com/EFForg/https-everywhere/commit/76dece2b3a413d601641ba628ccf2e32c3ce239d ensures that `./install-dev-dependencies.sh` works with the minimal Fedora 24 on dockerhub, and should not break your Vagrant install.  Mind testing this out and confirming?

I think the overhead accompanied with supporting Windows 10 for development is not worth the benefit.  If Windows developers wish to contribute, I'm inclined to require either Ubuntu, Debian, or Fedora in a VM or docker.

jeremyn:
I understand about Windows 10 support. By the way, I tried to set things up using the new [Bash on Ubuntu on Windows](https://msdn.microsoft.com/en-us/commandline/wsl/about) functionality in the Windows 10 anniversary update. `fetch-test.sh` worked fine but `test.sh` did not, so it's not quite there yet. It uses Ubuntu 14.04, which is old, and the functionality is still officially beta. Once it improves, that might be the easiest way for people to contribute using Windows 10.

In Fedora 24 running in a VM, the new `install-dev-dependencies.sh` crashes [here](https://github.com/EFForg/https-everywhere/blob/8f077b02d1513d80ccff8d65ddcc2e0f7923a413/install-dev-dependencies.sh#L48):

```bash
    dbus-uuidgen > /var/lib/dbus/machine-id
```

with a permissions issue because writing to `/var/lib/dbus` requires root. You can fix the problem by changing that line to:

```bash
    $SUDO_SHIM -- sh -c 'dbus-uuidgen > /var/lib/dbus/machine-id'
```

following [this Stack Overflow answer](https://stackoverflow.com/a/5560480), but what problem are you solving here? Until this problem I had never heard of `machine-id`. From [the machine-id manpage](https://www.freedesktop.org/software/systemd/man/machine-id.html) it looks like `/etc/machine-id` is the canonical machine id: "The simple configuration file format of /etc/machine-id originates in the /var/lib/dbus/machine-id file introduced by D-Bus. In fact, this latter file might be a symlink to /etc/machine-id."

Otherwise the changes look good.

Hainish:
Dbus is a [system requirement for Firefox](https://www.mozilla.org/en-US/firefox/48.0/system-requirements/).  Without this line, the `test/firefox.sh` script fails in Fedora 24 docker with:

    process 3104: D-Bus library appears to be incorrectly set up; failed to read machine uuid: Failed to open "/var/lib/dbus/machine-id": No such file or directory

The fix can be found [here](http://www.torkwrench.com/2011/12/16/d-bus-library-appears-to-be-incorrectly-set-up-failed-to-read-machine-uuid-failed-to-open-varlibdbusmachine-id/).

This shouldn't affect anyone who already has a `machine-id` set due to the [conditional directly above the statement](https://github.com/EFForg/https-everywhere/blob/8f077b02d1513d80ccff8d65ddcc2e0f7923a413/install-dev-dependencies.sh#L47), but this fixes docker.

Thanks for the reviewing, I'll change and push.

Hainish:
Change made in 1ab9bfb15c90c15f805dc6bece243644077d3643

jeremyn:
I can't reproduce that error even missing `/var/lib/dbus/machine-id`, or even missing both that and `/etc/machine-id`, on either Fedora 24 Workstation or the more minimal Fedora 24 Cloud. I guess it's something needed for the Fedora 24 Docker box you're using.

Could you add a note to the script describing the purpose of that line? Otherwise it's hard to discover, and people might be uncomfortable with a script that writes to `/var/lib/dbus` for no obvious reason. Possible text:

```bash
  # This is needed for Firefox on some systems. See here for more information:
  # https://github.com/EFForg/https-everywhere/pull/5584#issuecomment-238655443
  if [ ! -f /var/lib/dbus/machine-id ]; then
    $SUDO_SHIM sh -c 'dbus-uuidgen > /var/lib/dbus/machine-id'
  fi
```

Hainish:
Done 9c78e8011023392e23de88cf7e1e753f4cdf9b37

jeremyn:
Just FYI to anyone reading, I've noticed that `fetch-test.sh` takes significantly longer, like 5+ times longer, to run when working over a Samba share (`/vagrant`) than it does when working directly in the VM's filesystem (`git clone /vagrant ~/https-everywhere`). Unfortunately I have no fix to suggest other than not running it out of the shared directory.

jeremyn:
Update: the Vagrant box `ubuntu/xenial64` works now. I've had some cipher issues with Fedora, even with #6215 , so I personally recommend Ubuntu for anyone who has that option.

PR #5596, opened 2016-07-01
Fix example rule in `ruleset-testing.md`.

jish:
The ruleset target was using `example.com` but the rule was using
`dezeen.com`.


fuglede:
LGTM

PR #5597, opened 2016-07-02
Create 115.com.xml

renyouguo:


src/chrome/content/rules/115.com.xml
```diff
@@ -0,0 +1,41 @@
+<!--
+
+	Mismatch:
+		tj.data
+
+	MCB:
+		help
+
+	Redirect to http:
```
J0WI: So why are they included in the list of targets?

src/chrome/content/rules/115.com.xml
```diff
@@ -0,0 +1,41 @@
...
+	<target host="m.115.com" />
+	<target host="my.115.com" />
+	<target host="msg.115.com" />
+	<target host="note.115.com" />
+	<target host="passport.115.com" />
+	<target host="pay.115.com" />
+	<target host="pc.115.com" />
+	<target host="q.115.com" />
+	<target host="static.115.com" />
+	<target host="thumb.115.com" />
```
J0WI: vip.115.com

src/chrome/content/rules/115.com.xml
```diff
@@ -0,0 +1,41 @@
...
+	Mismatch:
+		tj.data
+
+	MCB:
+		help
+
+	Redirect to http:
+		assets.115img.com
+		static.115img.com
+
```
J0WI: Got a timeout on tool.115.com

src/chrome/content/rules/115.com.xml
renyouguo: @J0WI So what should I do? Delete this note?

PR #5600, opened 2016-07-03
[Spiderweb Software] Redirect www.spiderwebsoftware.com -> https://spiderwebsoftware.com and added test URLs

galenide:


galenide:
@J0WI @fuglede From what it appears, the current Spiderweb Software ruleset has an issue wherein accessing a URL at the `www.spiderwebsoftware.com` domain (such as `http://www.spiderwebsoftware.com/products.html`) with the ruleset enabled generates an error because the certificate name is only valid for `spiderwebsoftware.com` and is not valid for the `www.spiderwebsoftware.com` domain. As such, applying this fix may be of importance for situations where the Spiderweb Software ruleset is enabled.

PR #5604, opened 2016-07-04
Autistici switched to Let's Encrypt

Devices:
No more self signed certificate.

### DIFF
--- a/src/chrome/content/rules/Autistici-Inventati.xml
+++ b/src/chrome/content/rules/Autistici-Inventati.xml
@@ -8,10 +8,6 @@
 	<securecookie host="^www\.autistici\.org$" name=".+" />
 	<securecookie host="^\.noblogs\.org$" name=".+" />
 
-
-	<!--	All seem identical, aside from ^inventati which redirects to
-		www.inventati. Avoid having to add four different exceptions.
-										-->
 	<rule from="^http:"
 		to="https:" />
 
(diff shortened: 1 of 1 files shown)


J0WI:
Please just use
```
 	<rule from="^http:"
to="https:" />
```
to simplify.

src/chrome/content/rules/Autistici-Inventati.xml
```diff
@@ -1,4 +1,4 @@
-<ruleset name="Autistici/inventati" default_off="self-signed">
+<ruleset name="Autistici/inventati">
 
 	<target host="autistici.org" />
 	<target host="*.autistici.org" />
```
J0WI: Wildcrard targets requires at least three test urls.
Use just www otherwise.

Devices:
Like this? I am not sure about your proposed rule change, since I just copied the current rules and only removed the default off thing.

src/chrome/content/rules/Autistici-Inventati.xml
```diff
@@ -12,10 +12,10 @@
...
 										-->
-	<rule from="^http://(?:www\.)?(?:autistic|inventat)i\.org/"
-		to="https://www.autistici.org/" />
+	<rule from="^http:"
+		to="https:" />
 
-	<rule from="^http://([^/:@\.]+)\.(autistici|inventati)\.org/"
-		to="https://$1.$2.org/" />
+	<rule from="^http:"
+		to="https:" />
```
J0WI: This block is only needed once.

Devices:
Don't know how I missed that.

J0WI:
Thanks!

PR #5615, opened 2016-07-07
[uv.es] Fixed ruleset

socram8888:
This pull request fixes the rules for www.uv.es, an Spanish university, and most of its subdomains.

Namely:
 - The name of the ruleset has been escaped. For some reason, the grave prevented HTTPS Everywhere from loading the file.
 - Adds a special rule for uv.es, redirecting to https://www.uv.es/
 - The subdomains of blogs.uv.es have been replaced by a single wildcard target. This can be done because the server has a wildcard certificate for *.blogs.uv.es.

### DIFF
--- a/src/chrome/content/rules/Uv.es.xml
+++ b/src/chrome/content/rules/Uv.es.xml
@@ -4,8 +4,6 @@
 
 	Not supporting HTTPS or using an invalid certificate:
 	
-		* uv.es
-	
 		* acceso.uv.es
 		* anglogermanica.uv.es
 		* aposta.uv.es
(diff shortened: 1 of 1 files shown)


socram8888:
Hm, it seems https://trobes.uv.es/ is causing the test to fail, yet it's working fine on my computer. Apparently the server isn't properly closing down the connection.

J0WI:
Could you document the removed hosts in the ruleset that they will not be added again in future?

socram8888:
Mmm... All the removed hosts have just been grouped to blogs.uv.es, they still exist.

J0WI:
Oh, sure. Thanks!

PR #5620, opened 2016-07-08
[IPython.org] Update ruleset

fuglede:


J0WI:
I get an invalid cert on `archive.ipython.org`.
Does the redirect for nbviewer matter? I think it's the same content.
You may add a hint for `Jupyter.org.xml`.

src/chrome/content/rules/IPython.org.xml
```diff
@@ -5,13 +5,18 @@
...
 
 	<target host="ipython.org" />
 	<target host="www.ipython.org" />
 	
+	<target host="nbviewer.ipython.org" />
+
+	<rule from="^http://nbviewer\.ipython\.org/"
+		to="https://nbviewer.jupyter.org/" />
+
 	<rule from="^http://(?:www\.)?ipython\.org/"
```
J0WI: non-capturing groups should be avoided by our style guide.

fuglede: They should, but if we want to get rid of them, we should script that rather than enforce it on a ruleset-by-ruleset basis.

fuglede:
I added some comments.

J0WI:
> I get an invalid cert on archive.ipython.org.

What about this one?

fuglede:
Thanks, missed that one.

J0WI:
Thanks for completing!

PR #5625, opened 2016-07-11
add rule for Technische Aufklärung

numismatika:
small mixed content warning about thirdparty stuff from google, but so far the player works, calendar works and i can't find anything that breaks.

### DIFF
--- a/src/chrome/content/rules/Technische_Aufklärung.xml
+++ b/src/chrome/content/rules/Technische_Aufklärung.xml
@@ -1,0 +1,11 @@
+<ruleset name="Technische Aufklärung">
+
+	<target host="technische-aufklaerung.de" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Technische_Aufklärung.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Technische Aufklärung">
+
+	<target host="technische-aufklaerung.de" />
+	<target host="www.technische-aufklaerung.de" />
+
+	<rule from="^http://www\.technische-aufklaerung\.de/"
```
J0WI: Please document the mismatch.

numismatika:
@J0WI  been there done that

J0WI:
Thanks!

PR #5626, opened 2016-07-12
Update and rename JD.com-mixedcontent.xml to JD.com-Problematic.xml

renyouguo:


renyouguo:
@J0WI @fuglede @loveithateit @Hainish 
Luckly, I pass the checks.
It gets different errors each time, when it is ok every time accuraly.
Would you support to auto recheck while check.1 and check.2 get differnet error?

J0WI:
@renyouguo you should ask this in #5015

renyouguo:
@J0WI Solve the check problem may cost much time. Would you please check this PR firstly?

J0WI:
I'm a bit scared of merging rules that will break our master branch again. I really hope that #5015 will be merged soon and we have more stable results than.

renyouguo:
I'm a lot scared of updating jd.com rule again……
There are 20 PRs by me waiting your check~

renyouguo:
Pass again? I'm shocked！

Hainish:
@J0WI if it is only the fetch tests failing and not the validation tests, it should not cause the master branch to fail.  This is because the fetch tests are run in a way that compares what has changed *from* master.  If nothing has changed in the rulesets, it will not run any fetch tests.  You can tell in the Travis results which is the fetch test - the line in the travis results:

    docker run --rm -ti -v $(pwd):/opt -e FIREFOX=/$FIREFOX/firefox/firefox --privileged httpse bash -c "service miredo start && test/travis-ruleset-fetch.sh"

It may cause some problems down the line when we do a full fetch test, but with #5015 merged hopefully it won't.

J0WI:
@Hainish btw. I think we could generally speed up the tests by doing more things parallel:
- check_rules.py can be run separately, it doesn't depend on a specific browser version.
- makecrx.sh can be separated, it actually runs in both Firefox versions but seems to not fetch anything.
- fetch-test.sh seems to run in context of the two Firefox versions, so it depends on makexpi.sh right? It would be very surprisingly to have different fetch results there (currently we had only false positives). So I wonder if we can just fetch once using curl and just test if the extension works in Firefox as we do with Chrome.

`</offtopic>`

Hainish:
Thanks for the suggestion @J0WI.  I was considering separating out the `.travis.yml` env vars.  Currently we have `FIREFOX=firefox-latest` and `FIREFOX=firefox-esr-latest`.  I was thinking something like this:

```yml
env:
  - TEST=browser FIREFOX=firefox-latest
  - TEST=browser FIREFOX=firefox-esr-latest
  - TEST=fetch
```

Something like the above would allow us to separate out the tests, and they would run in parallel.  The disadvantage of the above approach would be that for every PR, the `docker build` step (which in turn downloads the upstream `electronicfrontierfoundation/https-everywhere-docker-base` dockerhub image) would have to be run 3 times instead of 2.

fuglede:
Someone (@semenko maybe?) suggested running nothing but the fetch test when the diff from master contained only changes to rules, which to me seems like a reasonable approach, even if it goes ever so slightly against the spirit of integration testing.

Hainish:
@fuglede that makes sense.  Want to open a separate issue for this?

renyouguo:
@J0WI @fuglede Would this be merged?

src/chrome/content/rules/JD.com.xml
```diff
@@ -1,115 +1,58 @@
...
 	³ Redirects to www.jcloud.com
+	⁴ Dropped
 	⁵ Reset
 
 
 	Problematic hosts in *jd.com:
-
-		- app ¹
+		
+		- apismart (404)
```
Hainish: Maybe use the above ² notation?

Hainish:
@renyouguo just a small note. Fix, then I'm ready to merge.

Hainish:
Hey, if this keeps not passing, don't worry about it. I can take a look and merge anyway even without Travis if it passes on my local machine, no need for you to keep at it.  (It's only a one line change from before, after all!)

On August 12, 2016 3:43:38 AM PDT, renyouguo <notifications@github.com> wrote:
>Reopened #5626.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/EFForg/https-everywhere/pull/5626#event-753845150

-- 

renyouguo:
@Hainish Thanks！

PR #5638, opened 2016-07-13
added factorio.com.xml (fixed #5623)

### LINKED ISSUE #5623: Add ruleset for https://www.factorio.com/
http://www.factorio.com/ -> https://www.factorio.com/

J0WI:


jeremyn:
Note:
- [ ] The configuration for https://guide.factorio.com is strange. On Firefox it redirects to https://forums.factorio.com with no problem, but on Chrome is gives me a mismatch warning saying the certificate is only good for www.factorioforums.com . So it is a mismatch, but maybe you can update the comment to be more specific, like `mismatch (some browsers only?)`. What do you think?

Add targets:
- [ ] https://eu2.factorio.com
- [ ] https://factorioforums.com
- [ ] https://www.factorioforums.com

Note as mismatch in top comment:
- [ ] `*.factorioforums.com` (other than `www`, as far as I can tell)

jeremyn:
Do you want to just not mention the strange behavior I found with https://guide.factorio.com ?

All of `*.factorioforums.com` mismatch for me, so shouldn't they all be in the comment? Anyway if you think we should list them specifically, can you add `ww.w.factorioforums.com` to the comment too?

jeremyn:
@J0WI : status check?

J0WI:
I just added the comment about `ww.w.factorioforums.com`, but I was not able to reproduce your issue on `guide.factorio.com`. I get a mismatch with all clients there.

jeremyn:
I get a mismatch now in Firefox too. Maybe I had something cached before.

Merged.

PR #5639, opened 2016-07-13
rebased PC-BSD.xml (fixed #5608)

### LINKED ISSUE #5608: pcbsd.org and www.pcbsd.org don't work well with HTTPS
The rules for pcbsd.org also seem to match pcbsd.org and www.pcbsd.org. Now the problem is: when I visit the website over HTTPS, it cannot load images, stylesheets, etc. The page looks pretty horrible. Could we please remove these rules?

J0WI:


jeremyn:
- [ ] Move https://api.pcbsd.org from mismatch to target
- [ ] Note https://pbibuild.pcbsd.org and https://pbibuild64.pcbsd.org as mismatches
- [ ] We don't need [this insecure cookie comment](https://github.com/J0WI/https-everywhere/blob/9231c2375499cf572e3f365137a2b06aca67f971/src/chrome/content/rules/PC-BSD.xml#L23-L28), right?
- [ ] I think we can generalize the [securecookie](https://github.com/J0WI/https-everywhere/blob/9231c2375499cf572e3f365137a2b06aca67f971/src/chrome/content/rules/PC-BSD.xml#L50) tag to `<securecookie host=".+" name=".+" />`

J0WI:
I would keep the insecure cookie comment, because there is no drawback and dev may get aware of the issue.
I had to remove the generic securecookie rule, because ^ and www are not fully secureable. The other hosts aren't serving any cookies.

jeremyn:
Do you mean that some of the `^`/`www` cookies are not `isSecure` even with the generic `securecookie`? It doesn't break anything though, right?

J0WI:
If you set the secure flag on the cookies but you can't access the site over https (because of mixed content), the site is not able to read the cookies.

jeremyn:
I think I understand. Are we running into a `securecookie` problem here because `www.pcbsd.org` is only partially allowed?

Also, for the comment, the drawback of the comment is that we are repeating ourselves between it and the `securecookie` tag itself. But if you want to leave the comment in, that's fine.

J0WI:
> I think I understand. Are we running into a securecookie problem here because www.pcbsd.org is only partially allowed?

Exactly.

jeremyn:
Thanks, merged.

Issue #5117, opened 2016-06-12
dvd.netflix.com broken


ramp-1s:
Can't see my netflix dvd queue. Note that the netflix streaming website is not affected, this issue only affects the dvd queue web page.

aqthegreat:
Same here. I called netflix and after some trouleshooting, we determined it was HTTPS Everywhere. This started on 6/10 for me, which was after the most recent update. I actually am not able to load the "MemberHome" page or the Queue.

mnordhoff:
http://dvd.netflix.com/ is mostly if not all HTTP. I haven't investigated closely, but my bet is that the `<securecookie>` rule is confusing the website -- the HTTPS pages don't see your cookies and redirect you to the HTTPS login page; the HTTPS login page sees them and redirects you back to HTTP, causing a loop.

aqthegreat:
That's interesting. You're probably right. Chrome tries several times to load the page, then eventually the server provides the 502 message.

As a workaround (for anyone searching), I was able to uncheck "netflix.com (partial)" under the "Stable Cases" section of the dropdown icon. That seems to allow the website to be functional. You have to at least be able to load some part of the netflix website though before that option becomes available.

numismatika:
i think @mnordhoff  is right, the securecookie line still is too broad.
Commenting it out fixes the layout on the dvd subdomain.
We need to figure out which cookies can be secured without breaking the site.

dstout:
I am also experiencing issues with dvd.netflix.com. I login from www.netflix.com, and when I click the DVD link, it had been giving a Bad Gateway error. Or more recently, it just thinks I'm not logged in on that domain, and keeps returning to the login screen. Turn off the extension, and it works fine.

numismatika:
Does it work if you go to https://dvd.netflix.com ? That domain is now tls-enabled.

aeshildr:
Still an issue as of 7/17/16 . I disabled everything, cleared cookies, etc. Narrowed it down to this extension only.

numismatika:
I made #5642 for it.
It disables the securecookie line that is broken and it also rewrites dvd subdomain to tls.
So that should fix it either way.

J0WI:
closed by #5642

PR #5642, opened 2016-07-13
Netflix:

numismatika:
- disable the securecookies rule  as it breaks dvd.netflix.com


numismatika:
has to be verified as the subdomain is available over https as well now

numismatika:
#5117 could be fixed by it

J0WI:
Please add a comment to document the issue with Cookies.

J0WI:
rerun tests

J0WI:
Thanks for fixing this!

PR #5658, opened 2016-07-18
mkb.ru

galeksandrp:


J0WI:
See also:
vbo.mkb.ru
corporate.mkb.ru
mpi.mkb.ru (only on port 8443)

galeksandrp:
Added.

J0WI:
It seems that `voltage-pp-0000.mkb.ru` is gone.

galeksandrp:
Removed.

PR #5672, opened 2016-07-18
metcombank.ru

galeksandrp:


src/chrome/content/rules/Metcombank.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Metcombank.ru">
+	<target host="metcombank.ru" />
+	<target host="www.metcombank.ru" />
+	<target host="dbo.metcombank.ru" />
```
J0WI: See also:
cfo.metcombank.ru
cfotest.metcombank.ru

galeksandrp: Added.

PR #5677, opened 2016-07-18
akbars.ru

galeksandrp:


src/chrome/content/rules/Akbars.ru.xml
```diff
@@ -0,0 +1,13 @@
+<!--quickweb. timed out-->
+<ruleset name="Akbars.ru">
+	<target host="akbars.ru" />
+	<target host="www.akbars.ru" />
+	<target host="online.akbars.ru" />
```
J0WI: See also online2. and sort hosts.

galeksandrp: Added

galeksandrp: Added.

PR #5679, opened 2016-07-18
severgazbank.ru

galeksandrp:


src/chrome/content/rules/Severgazbank.ru.xml
```diff
@@ -0,0 +1,6 @@
+<!--(www.)? timed out-->
```
J0WI: WFM

J0WI: correction: cert chain is incomplete

J0WI:
The renaming of the ruleset breaks our tests.

PR #5682, opened 2016-07-18
vbrr.ru

galeksandrp:


src/chrome/content/rules/Vbrr.ru.xml
```diff
@@ -0,0 +1,7 @@
+<!--(www.)? self signed-->
```
J0WI: WFM, I think this needs an update.

PR #5684, opened 2016-07-18
zapsibkombank.ru

galeksandrp:


src/chrome/content/rules/Zapsibkombank.ru.xml
```diff
@@ -0,0 +1,8 @@
+<!--(www.)? timed out-->
+<ruleset name="Zapsibkombank.ru">
```
J0WI: + acs.zapsibkombank.ru

galeksandrp: Added.

PR #5688, opened 2016-07-18
pochtabank.ru

galeksandrp:


src/chrome/content/rules/Pochtabank.ru.xml
```diff
@@ -0,0 +1,10 @@
+<!--(www.)? protocol error-->
```
J0WI: WFM.
Please also search for other subdomains, there are a lot more supporting https.

PR #5692, opened 2016-07-18
rosbank.ru

galeksandrp:


src/chrome/content/rules/Rosbank.ru.xml
```diff
@@ -0,0 +1,9 @@
+<!--m. mismatch-->
+<ruleset name="Rosbank.ru">
+	<target host="rosbank.ru" />
+	<target host="www.rosbank.ru" />
+	<target host="ibank.rosbank.ru" />
+	<target host="getlogin.rosbank.ru" />
```
J0WI: See also:
ib.rosbank.ru
mbr.rosbank.ru
rostrade.rosbank.ru
rtl.rosbank.ru

galeksandrp: Added.

PR #5693, opened 2016-07-18
rosevrobank.ru

galeksandrp:


src/chrome/content/rules/Rosevrobank.ru.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Rosevrobank.ru">
+	<target host="rosevrobank.ru" />
+	<target host="www.rosevrobank.ru" />
+	<target host="ibanking.rosevrobank.ru" />
+	<target host="vb.rosevrobank.ru" />
```
J0WI: blog.rosevrobank.ru
rst.rosevrobank.ru
spb.rosevrobank.ru

src/chrome/content/rules/Rosevrobank.ru.xml
```diff
@@ -0,0 +1,11 @@
...
+	
+	<rule from="^http://rosevrobank\.ru/"
+		to="https://www.rosevrobank.ru/" />
```
J0WI: This is not necessary, the redirect works on the side of the server.

galeksandrp: Removed.

PR #5697, opened 2016-07-18
tatsotsbank.ru

galeksandrp:


src/chrome/content/rules/Tatsotsbank.ru.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Tatsotsbank.ru">
+	<target host="tatsotsbank.ru" />
+	<target host="www.tatsotsbank.ru" />
+	
+	<rule from="^http://www\.tatsotsbank\.ru/"
+		to="https://tatsotsbank.ru/" />
```
J0WI: www works for me, so please remove this rule.

PR #5708, opened 2016-07-18
payonline.ru

galeksandrp:


src/chrome/content/rules/Payonline.ru.xml
```diff
@@ -0,0 +1,11 @@
+<!--www. 404-->
+<ruleset name="Payonline.ru">
+	<target host="payonline.ru" />
+	<target host="www.payonline.ru" />
+	<target host="secure.payonlinesystem.com" />
```
J0WI: Mismatch on `(www.)?payonlinesystem.com` and `smasol.payonline.ru`

galeksandrp:
Added.

Issue #5727, opened 2016-07-21
filecrypt.cc rule breaks click'n'load functionality


SimJoSt:
Filecrypt contains containers of links. The site contains the click'n'load functionality to automate transfering the containing links of the container to a download manager.
As this feature does not have https compatibility built in, it's broken through the use of https-everywhere.
Moreover the error message doesn't contain any information to what caused the problem and it took me some time to figure it out.

In the file https://github.com/EFForg/https-everywhere/blob/c2f8c67/src/chrome/content/rules/filecrypt.cc.xml an exclusion needs to be added.
Unfortunately I am not able to write it myself.

An example for a link, the click'n'load feature opens in a little popup, looks like this: http://filecrypt.cc/helper.html?1469099854883
The number is random and changes every time.

I will help anywhere I can.

Bisaloo:
@SimJoSt, from what I understand, this was fixed by your edits in #5728. Is that right? If so, can you close this please?

SimJoSt:
I didn't run into the problem again, so of course and thank you for the reminder.

PR #5728, opened 2016-07-21
filecrypt.cc: restore click'n'load functionality #5727

SimJoSt:
After the .html there always is a `?` followed by a string of number. Like this: `?1469099854883`
I don't know if the added exclusion pattern covers that, nor am I able to test it.


J0WI:
Thanks for offering this fix!
But your exclusion also requires a test url, see https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md

SimJoSt:
I tried to update it accordingly.
Hope it works, like travis checked :-)

PR #5733, opened 2016-07-21
Fix icons in CRX

galeksandrp:


J0WI:
@Hainish is this fixing the problem we had in the last roll out?

J0WI:
@galeksandrp could you add a description what this PR is fixing?

galeksandrp:
1. `git clone https://github.com/EFForg/https-everywhere.git && cd https-everywhere`
2. `./makecrx.sh`
3. image

J0WI:
I was not able to reproduce this on the new `webextensions` branch, but let's ping @Hainish again to ask him.

galeksandrp:
It is reproducible only on Chrome, since git clone does not create symlinks on Windows.

makecrx.sh
```diff
@@ -57,6 +57,10 @@ sed -e "s/VERSION/$VERSION/g" chromium/updates-master.xml > chromium/updates.xml
 mkdir -p pkg/crx/rules
 cd pkg/crx
 rsync -aL ../../chromium/ ./
```
Hainish: This `rsync` command should suffice, since the `-L` flag copies symlinks.

PR #5736, opened 2016-07-21
hashicorp.com

galeksandrp:


src/chrome/content/rules/HashiCorp.com.xml
```diff
@@ -12,6 +12,9 @@
 
 	<target host="hashicorp.com" />
 	<target host="www.hashicorp.com" />
+	<target host="atlas.hashicorp.com" />
+	<target host="shop.hashicorp.com" />
+	<target host="status.hashicorp.com" />
```
J0WI: releases.hashicorp.com
gems.hashicorp.com
checkpoint.hashicorp.com
scada.hashicorp.com
license.hashicorp.com

J0WI:
@galeksandrp please have a look at our guide, where we documented how to find subdomains: https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

J0WI:
Note that this host is already HSTS preloaded, so this ruleset might be obsolete.

PR #5743, opened 2016-07-21
vulkan-casino.com

galeksandrp:


J0WI:
How are all these hosts related?
It seems that they don't belong all to the same portal.

galeksandrp:
Seems yes, they only sharing same trademark.

Should I split it to multiple files?

J0WI:
For them who belongs to the same content/portal I'm okay with one file. Other may be separated and linked with comments if they are related.

J0WI:
@galeksandrp are you interested in continuing here?

PR #5749, opened 2016-07-22
revolutiontt.me

galeksandrp:


J0WI:
I get a 403 on the root of this host.
Can you provide a valid test url?

src/chrome/content/rules/Revolutiontt.me.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Revolutiontt.me">
+	<target host="revolutiontt.me" />
+	<target host="www.revolutiontt.me" />
```
J0WI: img.revolutiontt.me

galeksandrp:
Only this urls is publicily available:
https://revolutiontt.me/login.php
https://revolutiontt.me/recover.php
https://revolutiontt.me/irchelp.php
https://m.revolutiontt.me/index.php

J0WI:
Therefore `m.revolutiontt.me` should be added too.
But I get a 403 on all these urls.

J0WI:
Worked with other IPs.

PR #5750, opened 2016-07-22
t411.ch

galeksandrp:


src/chrome/content/rules/T411.ch.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="T411.ch">
+	<target host="t411.ch" />
+	<target host="www.t411.ch" />
+	<target host="wiki.t411.ch" />
+	<target host="forum.t411.ch" />
```
J0WI: api.t411.ch

Refused:
irc.t411.ch

galeksandrp:
Added.

J0WI:
Hosts that do not exist shouldn't be touched.

Please also use tabs instead of spaces to indent lines and avoid non-capturing grous in regex, as our code style requires.

PR #5752, opened 2016-07-22
bitsnoop.com

galeksandrp:


src/chrome/content/rules/Bitsnoop.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Bitsnoop.com">
+	<target host="bitsnoop.com" />
+	<target host="www.bitsnoop.com" />
```
J0WI: Refused:
blog.bitsnoop.com

galeksandrp:
Added.

PR #5759, opened 2016-07-22
status.io

galeksandrp:


src/chrome/content/rules/Status.io.xml
```diff
@@ -0,0 +1,11 @@
+<!--developers. timed out-->
```
J0WI: It seems that some customers get their own subdomain, which are not secureable. Can you document that?

galeksandrp:
Documented.

J0WI:
Thanks!

PR #5761, opened 2016-07-22
statushub.io

galeksandrp:


src/chrome/content/rules/Statushub.io.xml
```diff
@@ -0,0 +1,20 @@
+<!--blog. protocol error
+mozillastatus. mismatch
```
J0WI: customer get their own subdomain, which are not securable.
can you document this behavior instead of just list them?

Note that you need to keep the separate comment for the blog. subdomain.

galeksandrp:
Secured.

J0WI:
Thanks for adding the redirects, that's even better.

PR #5762, opened 2016-07-22
slack.com

galeksandrp:


J0WI:
This host is HSTS preloaded by browsers anyway, so it don't need to be in HTTPS Everywhere.
But you can secure the following hosts:
```
a.slack-edge.com
downloads.slack-edge.com
platform.slack-edge.com
(www and ^ do not exist)

slackhq.com
www.slackhq.com

slackatwork.com
www.slackatwork.com
```

galeksandrp:
Done.

J0WI:
Thanks!

J0WI:
Just realized that you missed `downloads.slack-edge.com`.
See https://slack.com/downloads for usage (bins live there).

galeksandrp:
Added, thanks!

PR #5768, opened 2016-07-22
librato.com

galeksandrp:


src/chrome/content/rules/Librato.com.xml
```diff
@@ -0,0 +1,9 @@
+<!--blog. mismatch-->
+<!--support.metrics. refused-->
+<ruleset name="Librato.com">
+	<target host="librato.com" />
+	<target host="www.librato.com" />
+	<target host="metrics.librato.com" />
```
J0WI: status.librato.com

Mismatch:
info.librato.com

PR #5769, opened 2016-07-22
campaignmonitor.com

galeksandrp:


src/chrome/content/rules/Campaignmonitor.com.xml
```diff
@@ -0,0 +1,9 @@
+<!--i3. wrong redirect-->
+<!--i1. wrong redirect-->
+<ruleset name="Campaignmonitor.com">
+	<target host="campaignmonitor.com" />
+	<target host="www.campaignmonitor.com" />
+	<target host="help.campaignmonitor.com" />
```
J0WI: hello.campaignmonitor.com
status.campaignmonitor.com
templates.campaignmonitor.com

J0WI:
Thanks!
It would be helpful if you could ping me after you addressed my comments. Otherwise I'm not getting notified.

PR #5770, opened 2016-07-22
avito.ru

galeksandrp:


J0WI:
Please also add `promo.avito.ru`

PR #5775, opened 2016-07-23
Added community and developers subdomains, improved naming.

marcoslater:


src/chrome/content/rules/getmondo.co.uk.xml
```diff
@@ -1,7 +1,9 @@
-<ruleset name="get Mondo.co.uk">
+<ruleset name="Get Mondo">
 
 	<target host="getmondo.co.uk" />
 	<target host="www.getmondo.co.uk" />
+	<target host="community.getmondo.co.uk" />
+	<target host="developers.getmondo.co.uk" />
```
J0WI: Could you also add devslack.getmondo.co.uk?

J0WI:
Note that serves preloaded HSTS header, but is not yet preloaded.

marcoslater:
@J0WI Good spot, added. 

I've also noticed they serve HSTS preload, but aren't on the list. Perhaps forgot?

J0WI:
They are still pending, so I'll merge this now.
We may remove the rule in future when they are on the list for sure.

PR #5788, opened 2016-07-24
Create kf5.com.xml

renyouguo:


src/chrome/content/rules/kf5.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	kf5.com is an open site, so there are lots of *.kf5.com sites, which are too much to list all of them.
```
J0WI: It seems that they all support https, so a wildcard with exclusions would make sense here.
The current targets can be kept as test urls.

src/chrome/content/rules/kf5.com.xml
```diff
@@ -8,17 +8,25 @@
...
-	<target host="assetscdn.kf5.com" />
-	<target host="blog.kf5.com" />
-	<target host="developer.kf5.com" />
-	<target host="qiniu.kf5.com" />
-	<target host="support.kf5.com" />
-	<target host="www.kf5.com" />
+	
+	<target host="*.kf5.com" />
+
+	<exclusion pattern="^http://(agents|d)\.kf5\.com/(.*)" />
```
J0WI: The `(.*)` on the end is not needed, the pattern will match anyway and is much simpler.

PR #5798, opened 2016-07-24
Tello.xml ruleset for tello.com and www.tello.com

theshoals:


J0WI:
Please also add a hint for the mismatching cert on blog.tello.com.

theshoals:
Thanks for the review.  I updated the pull request with a comment about blog.tello.com.

PR #4781, opened 2016-05-14
Update Yahoo_Japan.xml

renyouguo:

PR #5799, opened 2016-07-25
Update Yahoo_Japan.xml

luixxiul:
Include the top level domain only

Addresses #4781

On subdomains it won't work as described on L13.


J0WI:
Thanks!

PR #5800, opened 2016-07-25
Create Beetman.net.xml

BeetMan:
Add one website: https://beetman.net/


J0WI:
Please use tabs to follow our style guide.
There is also a host that refuses https (ghost.beetman.net), could you add a comment for that?

If this is your own site you should also have a look at: https://hstspreload.appspot.com/
This is built in browsers and would make this rule obsolete.

jeremyn:
@BeetMan Are you still interested in working on this?

PR #5806, opened 2016-07-25
[turisticke-znamky.cz] ruleset

MikkCZ:


fuglede:
Some of the pages that you land on by clicking the flags at the top of the main site will suffer from mixed content blocking, but it's nothing that breaks site function in any major way.

PR #5812, opened 2016-07-27
Create 0xacab.org rules.

just1602:


src/chrome/content/rules/0xacab.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="0xacab.org">
+
+	<target host="0xacab.org" />
+	<target host="www.0xacab.org" />
+	<target host="about.0xacab.org" />
```
J0WI: flock.0xacab.org

J0WI:
Thanks for adding them!

PR #5816, opened 2016-07-27
Added several doamins to Cloudtrax.

marcoslater:
Hopefully this works now.


marcoslater:
Looks like that fixed the issues. Seems that files.cloudtrax.com had some weird redirect issues, regardless of protocol, at least on /.

J0WI:
Please add a comment for the issue on `files.` that it won't be added accidentally in future.
Since you edit this anyway, could you sort the added subdomains alphabetically?

J0WI:
Thanks!

PR #5819, opened 2016-07-28
Updated Yahoo_Japan.xml

luixxiul:
- Removed chiebukuro from search subdomains because http://chiebukuro.search.yahoo.co.jp is redirected to http://chiebukuro.yahoo.co.jp
- Added chiebukuro.yahoo.co.jp
- Added detail.chiebukuro.yahoo.co.jp

### DIFF
--- a/src/chrome/content/rules/Yahoo_Japan.xml
+++ b/src/chrome/content/rules/Yahoo_Japan.xml
@@ -26,7 +26,6 @@
 			- chiebukuro subdomains:
 
 				- answer *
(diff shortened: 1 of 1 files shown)


J0WI:
But the timout on chiebukuro.search.yahoo.co.jp is still true

luixxiul:
Yes it is, correcting

luixxiul:
@J0WI how about now?

J0WI:
LGTM, thanks!

PR #5820, opened 2016-07-28
Update rakuten.co.jp.xml

luixxiul:


luixxiul:
@J0WI Should I include test urls?

luixxiul:
WIP based on http://www.rakuten.co.jp/sitemap/

luixxiul:
Move to #5820

PR #5829, opened 2016-07-28
icontem.com

galeksandrp:


J0WI:
Could you add the partial keyword to then rule name?

galeksandrp:
Done.

PR #5843, opened 2016-07-28
cctld.ru

galeksandrp:


src/chrome/content/rules/Cctld.ru.xml
```diff
@@ -0,0 +1,14 @@
+<!--meeting. mismatch
+meeting2015. mismatch
+safety. self signed
```
J0WI: test. self signed

galeksandrp:
Added.

J0WI:
Thanks!

PR #5845, opened 2016-07-28
Add OneBookShelf.xml

jeremyn:


J0WI:
I would expect onebookshelf.com in this ruleset.
Can you also comment on it?

jeremyn:
I've reorganized the top comment a little and added https://onebookshelf.com and https://www.onebookshelf.com under `Refused`.

PR #5865, opened 2016-07-30
relcom.host

galeksandrp:


src/chrome/content/rules/Relcom.host.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Relcom.host">
```
J0WI: see also (*.)relcomhost.ru

galeksandrp:
Added.

PR #5875, opened 2016-07-31
Create tanx.com.xml

renyouguo:


renyouguo:
@Hainish

Hainish:
For these, I keep getting redirected to `https://err.taobao.com/error1.html`.  Are there some routes for which resources are loaded from these domains?  Can you provide these in `<test>` tags?

renyouguo:
@Hainish I can't find the url which will not redirect to err.taobao.com for ets.tanx.com and pass.tanx.com.
However, they have https and are working well.

Hainish:
I feel kind of ambiguous about merging this.  Sure, it's a secure site, but we don't have any real examples of how this site is being used?

All other things being equal, I would favor including as many small sites as possible that support HTTPS, and the only criteria is that they function as expected and you lose no functionality from removing the ability to access the site over plain HTTP.  But for this site, I'm not sure we can claim that, since we have no real example of how the site is used.  We also face the additional scalability burden as explained in this issue:

https://github.com/EFForg/https-everywhere/issues/6322#issuecomment-240574166

I'm closing this for now, but open to reconsidering if we can find how this domain is actually used and get good test urls.

renyouguo:
@Hainish tanx.com belongs alimama.com , same with the biggest site: taobao.com
You can see their main site by tanx.com and mu.tanx.com , example in : https://mu.tanx.com/m/refer_sdk.htm
It is useful and common in China , as routes.

Issue #4075, opened 2016-02-03
archive.is not available via https in Russia


volth:
archive.is not available via https in Russia https://en.wikipedia.org/wiki/Archive.is#Worldwide_availability
Please consider either making the rewrite rules geo-aware or stop rewriting "archive.is".

stevenatwork:
>In Russia you can access only the HTTP version of the website; the HTTPS version is blocked on the state firewall.

https://en.wikipedia.org/wiki/Archive.is#Worldwide_availability

In my opinion this is a **terrible reason** to stop rewriting to HTTPS.

jsha:
Thanks for reporting! I wouldn't have put it quite so strongly, but we won't disable rulesets because of local censorship, nor make HTTPS Everywhere geo-aware. It would go against EFF's anti-censorship stance.

PR #5886, opened 2016-07-31
[Archive.is] Update ruleset

### LINKED ISSUE #5805: Archive.is redirecting HTTPS to HTTP
It seems archive.is has started redirecting HTTPS requests to HTTP (with an annoying 10-second delay). For example: https://archive.is/2VUZ (though I can't consistently reproduce this 100% of the time). Not sure how to disable a rule (just comment it out, or something else?) so I'm just opening an issue.

fuglede:
After having had a couple of different issues with this ruleset, cf. in particular https://github.com/EFForg/https-everywhere/issues/5805 and https://github.com/EFForg/https-everywhere/issues/4075.

I prompted the webmaster for what would be a proper solution moving forwards. On their request, this avoids redirecting anything to https://archive.is/ but ensures that the relevant connections are still upgrading by using archive.fo instead. Indeed, they considered disabling the option to use secure connections to archive.is but were happy with still providing the option in this slightly round-about fashion.

This closes #5805.

### DIFF
--- a/src/chrome/content/rules/Archive.is.xml
+++ b/src/chrome/content/rules/Archive.is.xml
@@ -10,11 +10,16 @@
 	<target host="archive.today" />
 	<target host="blog.archive.today" />
 	<target host="www.archive.today" />
+	<target host="archive.fo" />
+	<target host="www.archive.fo" />
 	<target host="archive.is" />
 	<target host="www.archive.is" />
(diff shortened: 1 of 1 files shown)


J0WI:
Can you comment that in the rule that we don't forget this in future?

fuglede:
Sure.

PR #5903, opened 2016-08-01
spark-com.ru

galeksandrp:


src/chrome/content/rules/spark-com.ru.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+spark-com.ru mismatch
+www.spark-com.ru mismatch
+speedtest.spark-com.ru mismatch
+go2ethernet.spark-com.ru mismatch
```
J0WI: idea.spark-com.ru
noaccess.spark-com.ru

PR #5911, opened 2016-08-01
akado-ural.ru

galeksandrp:


src/chrome/content/rules/akado-ural.ru.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+akado-ural.ru timed out
+www.akado-ural.ru timed out
```
J0WI: They WFM, please double check.

PR #5919, opened 2016-08-01
goodline.info

galeksandrp:


src/chrome/content/rules/goodline.info.xml
```diff
@@ -0,0 +1,32 @@
...
+yurga.goodline.info different content
+zelenogorskiy.goodline.info different content
+2016.goodline.info different content
+b2b.goodline.info different content
+-->
+<ruleset name="goodline.info">
+	<target host="goodline.info" />
+	<target host="www.goodline.info" />
+	<target host="lk.goodline.info" />
+	<target host="old-lk.goodline.info" />
```
J0WI: live.goodline.info

lgtm otherwise.

galeksandrp:
Added.

src/chrome/content/rules/goodline.info.xml
galeksandrp: Added.

PR #5923, opened 2016-08-01
zelenaya.net

galeksandrp:


src/chrome/content/rules/zelenaya.net.xml
```diff
@@ -0,0 +1,37 @@
+<!--
+efremov.zelenaya.net self signed
+tambov.zelenaya.net self signed
+elets.zelenaya.net self signed
```
J0WI: Please double check for cert chain issues.

galeksandrp: Fixed.

J0WI: Nope, they are still treated as self-signed. They just had chain issues, but now they expired recently. Let's see if they get fixed.

J0WI: @galeksandrp they are still treated as self-signed

PR #5924, opened 2016-08-01
airbites.net.ua

galeksandrp:


src/chrome/content/rules/airbites.net.ua.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+www.ipv6.airbites.net.ua nonexist
```
J0WI: old.airbites.net.ua
stb.airbites.net.ua
ts.airbites.net.ua

J0WI: You don't need to lists hosts that do not exist or that do not support HTTP at all.

PR #5941, opened 2016-08-01
fregat.net

galeksandrp:


src/chrome/content/rules/fregat.net.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+fregat.net nonexist
+www.fregat.net timed out
+home.fregat.net timed out
+noc.fregat.net mismatch
+stream.fregat.net timed out
```
J0WI: Could you update this list?

galeksandrp: Updated.

PR #5947, opened 2016-08-01
obit.ru

galeksandrp:


src/chrome/content/rules/obit.ru.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+obit.ru mismatch
+www.obit.ru mismatch
+en.obit.ru mismatch
+lg.obit.ru mismatch
+ftp.obit.ru expired
```
J0WI: Please double check, some of them WFM.
You may also find more subdomains.

galeksandrp: Added.

PR #5956, opened 2016-08-01
mosnet.ru

galeksandrp:


J0WI:
Could you also search for subdomains here?

src/chrome/content/rules/mosnet.ru.xml
```diff
@@ -1,6 +1,25 @@
...
+orange.mosnet.ru ³
+privateoffice.mosnet.ru ⁴
+raven.mosnet.ru ²
+speedtest.mosnet.ru ²
+speedtestlocal.mosnet.ru ²
+
+
+² refused
+³ timed out
+⁴ self signed
```
J0WI: Please double check, I think some are only chain issues.

galeksandrp: But we [don't accept domains with chain issues](https://github.com/EFForg/https-everywhere/pull/3279#issuecomment-160349032), isn't it?

galeksandrp: Fixed.

PR #5960, opened 2016-08-01
nashnet.ua

galeksandrp:


src/chrome/content/rules/nashnet.ua.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+forum.nashnet.ua different content
```
J0WI: WFM

See also `billing.nashnet.ua`

galeksandrp: Added.

PR #5966, opened 2016-08-01
balnet.ru

galeksandrp:


src/chrome/content/rules/balnet.ru.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="balnet.ru (partial)">
+	<target host="balnet.ru" />
+	<target host="www.balnet.ru" />
+	<target host="billing.balnet.ru" />
```
J0WI: game.balnet.ru
ns1.balnet.ru
test.balnet.ru

You may remove the partial flag.

galeksandrp: Added.

PR #5975, opened 2016-08-01
Update OpenX coverage

luixxiul:
- Included jp-u.openx.net
- Added test urls


luixxiul:
Please make sure the Travis error was caused by ``http://bid.openx.net``, which is actually a 503 Service Unavailable error.

luixxiul:
Let me know if the domain should be commented out or removed.

luixxiul:
``https://i-cdn.openx.com`` should work now.

PR #5977, opened 2016-08-01
Im-apps.net

luixxiul:
Found on http://goo.ne.jp like #5976 and http://qiita.com

### DIFF
--- a/src/chrome/content/rules/im-apps.net.xml
+++ b/src/chrome/content/rules/im-apps.net.xml
@@ -1,0 +1,14 @@
+<ruleset name="im-apps.net">
+
+	<target host="atm.im-apps.net" />
+	<target host="atr.im-apps.net" />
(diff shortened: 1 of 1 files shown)


luixxiul:
@jeremyn please review, thanks

jeremyn:
@luixxiul Thank you for your ruleset contributions but please don't ping me specifically on these.. There are almost 900 open pull requests and we should get through the older ones first.

luixxiul:
@jeremyn oh ok I see, I just felt people here thought my PRs were so trivial that they were not worth reviewing.

jeremyn:
@luixxiul Not at all! It's great that you're contributing. Please keep submitting as many rulesets as you like. There's just a huge number to review. I actually just started on the ruleset review team last week as a volunteer, so with an extra person things should move forward more quickly. Thanks for your patience.

PR #5978, opened 2016-08-01
Added logly.co.jp

luixxiul:
Found on http://goo.ne.jp like #5976


luixxiul:
@Hainish please review, thanks

PR #5989, opened 2016-08-01
karelia.pro

galeksandrp:


J0WI:
They have now a lot of certs from Let's Encrypt, so this needs an update.

galeksandrp:
Updated.

PR #5992, opened 2016-08-01
westlan.ru

galeksandrp:


src/chrome/content/rules/westlan.ru.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+westlan.ru mismatch
+www.westlan.ru mismatch
```
J0WI: bonus.westlan.ru refused

galeksandrp:
Added.

PR #6009, opened 2016-08-01
lan.ua

galeksandrp:


src/chrome/content/rules/lan.ua.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+lan.ua mismatch
+www.lan.ua mismatch
+mail.lan.ua mismatch
+fun.lan.ua mismatch
```
J0WI: Please double check, except from mail they are WFM

galeksandrp: Added.

PR #6017, opened 2016-08-01
forceline.net

galeksandrp:


J0WI:
Please also include other subdomains.

src/chrome/content/rules/forceline.net.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+forceline.net different content
+www.forceline.net different content
```
J0WI: Please double check if this has been fixed.

galeksandrp: Added.

PR #6019, opened 2016-08-01
linkline.ru

galeksandrp:
do not merge


src/chrome/content/rules/linkline.ru.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+linkline.ru self signed
+www.linkline.ru self signed
```
J0WI: also on speed.

galeksandrp:
Added.

PR #6021, opened 2016-08-01
taiget.ru

galeksandrp:


src/chrome/content/rules/taiget.ru.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+sec.taiget.ru non-2xx http code
```
J0WI: Mismatch on notice. and test.

src/chrome/content/rules/taiget.ru.xml
```diff
@@ -0,0 +1,10 @@
...
+-->
+<ruleset name="taiget.ru">
+	<target host="taiget.ru" />
+	<target host="www.taiget.ru" />
+	<target host="vpn.taiget.ru" />
```
J0WI: eirc. also works

galeksandrp:
Added.

PR #6028, opened 2016-08-01
indikom.ru

galeksandrp:


src/chrome/content/rules/indikom.ru.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+indikom.ru mismatch
+www.indikom.ru mismatch
```
J0WI: They WFM now

See also `b.indikom.ru`

galeksandrp: Added.

PR #6031, opened 2016-08-01
telenet.tomsk.ru

galeksandrp:


J0WI:
Can we make this rule for *.tomsk.ru?

PR #6033, opened 2016-08-01
ilca.ru

galeksandrp:


src/chrome/content/rules/ilca.ru.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+ilca.ru mismatch
+www.ilca.ru mismatch
```
J0WI: support. => mismatch

src/chrome/content/rules/ilca.ru.xml
```diff
@@ -0,0 +1,10 @@
...
+bill.ilca.ru different content
+-->
+<ruleset name="ilca.ru (partial)">
+	<target host="forum.ilca.ru" />
```
J0WI: mixed conetnet

galeksandrp:
Wrong content on http://bill.ilca.ru. Fixed.

J0WI:
Can't confirm. https redirects me to /user with login mask, plain http shows webpage.

galeksandrp:
Looks like (www.)?-identical webpage on plain http is misconfiguration, because (www.)? website contains links only to https://bill.ilca.ru, which i think is for credintals-protected billing, not for exact copy of main site.

J0WI:
This seems reasonable to me.

PR #6044, opened 2016-08-02
Added deqwas.net

luixxiul:
Found on http://shop.gnavi.co.jp/


luixxiul:
On some sites I notice URLs with random subdomains such as ``http://kafcs.deqwas.net/CookieSyncService/Sync.aspx``, so I think might be better to include them with a wildcard.

PR #6046, opened 2016-08-02
Added searchteria.co.jp

luixxiul:
See #6045

### DIFF
--- a/src/chrome/content/rules/searchteria.co.jp.xml
+++ b/src/chrome/content/rules/searchteria.co.jp.xml
@@ -1,0 +1,22 @@
+<!--
+
+	Called from http://j.gmodmp.jp/js/d.js
+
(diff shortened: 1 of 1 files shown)


fuglede:
You don't really need the test URL in cases like these; one is automatically inferred from the target host. As it's set up right now, the test is actually [run twice](https://travis-ci.org/EFForg/https-everywhere/jobs/149292462#L1572).

luixxiul:
@fuglede ok done

luixxiul:
@fuglede sometimes the test is run twice, sometimes once, why?

fuglede:
That sounds weird. Do you have the link to an example where it's run twice where it shouldn't be?

luixxiul:
Sorry, maybe I misunderstood something. No problem.

PR #6048, opened 2016-08-02
version6.ru

galeksandrp:


src/chrome/content/rules/version6.ru.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+www.version6.ru nonexist
+-->
+<ruleset name="version6.ru">
+	<target host="version6.ru" />
+	<target host="www.version6.ru" />
+	
+	<rule from="^http://www\.version6\.ru/"
```
J0WI: You shouldn't redirect hosts with no DNS entire.

galeksandrp:
Deleted.

PR #6049, opened 2016-08-02
6to4.ru

galeksandrp:


src/chrome/content/rules/6to4.ru.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+www.6to4.ru nonexist
+-->
+<ruleset name="6to4.ru">
+	<target host="6to4.ru" />
+	<target host="www.6to4.ru" />
+	
+	<rule from="^http://www\.6to4\.ru/"
```
J0WI: If no DNS entire exist you shouldn't redirect anything.

galeksandrp:
Deleted.

PR #6050, opened 2016-08-02
Exception for streamfarm.net-rules on erf.de

strch:
Without this exception videos cannot be playend on erf.de (see example links in commit message)

### DIFF
--- a/src/chrome/content/rules/Streamfarm.net.xml
+++ b/src/chrome/content/rules/Streamfarm.net.xml
@@ -6,6 +6,11 @@
 -->
 
   <exclusion pattern="^http://c22033\-ls\.i\.core\.cdn\.streamfarm\.net/" />
+
+<!-- 
+	breaks erf.de
+-->
+  <exclusion pattern="^http(s)?://(www\.)?erf\.de" />
   
   <test url="http://c1000305-o.p.core.cdn.streamfarm.net/" />
   <test url="http://c13009-l.l.core.cdn.streamfarm.net/" />


fuglede:
Hi @strch and thanks for the contribution; a couple of remarks:

* The addition of the exclusion will not actually do anything. Exclusions work as follows: Upon connection to a given host, the add-on checks if any of the target hosts match the connection. If they do, then the ruleset is checked for exclusions, and if none are found, then the connection is upgraded. In this case, `erf.de` is not covered by any of the target hosts, so the exclusion will never trigger.

* I can not actually reproduce the issue. I tried one of the URLs from your commit message, https://www.erf.de/fernsehen/mediathek/schlafschaf-tv/ein-neuanfang-unterm-regenbogen/6710-137, and the video displays fine here.  A couple of immediate thoughts to follow up: Are you using Flash? If you at the browser console (F12 in Firefox), or the network tab next to the console, do you see any errors or warnings regarding the non-loading video?

jeremyn:
@strch Are you still interested in working on this?

strch:
Sorry for answering so late and thanks fuglede for your analysis and clarifying comment.

Indeed I had some javascript errors but they were due to other addons/extensions (like ublock) interfering.

Hence the issue had to be resolved by the site owner of erf.de. And just today they relaunched, making the videos working again even for me ;-)

So nothing left to be done, jeremyn. And thanks again, fuglede for your efforts!

jeremyn:
@strch Thanks for reporting back. Just to be sure, are you okay with closing this pull request?

strch:
@jeremyn Yes the pull request can (and should) be closed.

PR #6051, opened 2016-08-03
Add hosts-file.net

WilliamFeely:
Recently hpHosts announced they started supporting HTTPS.  However, automatic redirect of the main domain (with and without www) got disabled due to issues with certain third-party applications.
Since redirecting on interactive web browsers should be safe, I am proposing the addition of hosts-file.net and www.hosts-file.net to HTTPS Everywhere.

### DIFF
--- a/src/chrome/content/rules/hosts-file.net.xml
+++ b/src/chrome/content/rules/hosts-file.net.xml
@@ -1,6 +1,7 @@
 <ruleset name="hpHosts">
         <target host="hosts-file.net" />
         <target host="www.hosts-file.net" />
+        <target host="forum.hosts-file.net" />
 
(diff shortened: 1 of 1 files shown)


J0WI:
Could you please also add `forum.hosts-file.net`?

WilliamFeely:
I thought I didn't need to because the forum subdomain already redirects, but I can do so.

J0WI:
As long as it's not preloaded HSTS we still need to secure it, because an attacker could hijack the initial request.

WilliamFeely:
Thanks for the explanation.  I just added the forum subdomain.

J0WI:
Thanks!

PR #6053, opened 2016-08-03
counterpath.com

galeksandrp:


src/chrome/content/rules/counterpath.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+counterpath.com redirect
+www.counterpath.com redirect
+-->
+<ruleset name="counterpath.com (partial)">
+	<target host="jira.counterpath.com" />
+	<target host="secure.counterpath.com" />
+	<target host="support.counterpath.com" />
+	<target host="blog.counterpath.com" />
```
fuglede: This host has some mixed content issues. The "most popular" post, https://blog.counterpath.com/794/bria-ipad-edition-demo-with-video-at-broadsoft-connections, blocks content from YouTube. Moreover, fonts are blocked but that's not a critical issue.

galeksandrp:
Deleted.

fuglede:
Thanks!

In this case, you might be able to still secure some amount of the content on the site, such as all static content, but of course narrowing that down is a lot more work.

PR #6054, opened 2016-08-03
Create fir.im.xml

renyouguo:


renyouguo:
@Hainish

Hainish:
Tests are failing with cert error https://travis-ci.org/EFForg/https-everywhere/jobs/153571989

renyouguo:
@Hainish Could this be merged now?

src/chrome/content/rules/fir.im.xml
```diff
@@ -0,0 +1,26 @@
...
+
+-->
+
+<ruleset name="fir.im">
+
+	<target host="fir.im" />
+	<target host="account.fir.im" />
+	<target host="api.fir.im" />
+	<target host="blog.fir.im" />
+	<target host="club.fir.im" />
```
Hainish: For this host, I'm getting image requests timeouts on the HTTPS endpoint - the HTTP endpoint loads quickly.

renyouguo:
@Hainish Ok now?

jeremyn:
@renyouguo Thanks for the ruleset. Please do this:

Add targets: 
- [ ] https://club.fir.im (I'm not getting `MCB` here, so you can remove it from the comment)
- [ ] https://ga.fir.im
- [ ] https://usercenter.fir.im

Also:
- [ ] Change `Mismatch` to `Invalid cert` in the top comment, because some of the certificates fail for multiple reasons and it's simpler to label it this way
- [ ] Move [these `www` targets](https://github.com/renyouguo/https-everywhere/blob/6cd1e4d4688034809b674debf03cd7b4d7362c9a/src/chrome/content/rules/fir.im.xml#L20-L21) directly under the `^fir.im` target [here](https://github.com/renyouguo/https-everywhere/blob/6cd1e4d4688034809b674debf03cd7b4d7362c9a/src/chrome/content/rules/fir.im.xml#L13)
- [ ] Remove all comments about `firimg.qiniudn.com`, there is no need to mention that
- [ ] Move [the `qbox.me` target](https://github.com/renyouguo/https-everywhere/blob/6cd1e4d4688034809b674debf03cd7b4d7362c9a/src/chrome/content/rules/fir.im.xml#L23) to [`Qbox.me.xml`](https://github.com/EFForg/https-everywhere/blob/e332b8853769586cde09d79917e5cd9d31e7fb6a/src/chrome/content/rules/Qbox.me.xml)
- [ ] Delete [the extra `qbox.me` test](https://github.com/renyouguo/https-everywhere/blob/6cd1e4d4688034809b674debf03cd7b4d7362c9a/src/chrome/content/rules/fir.im.xml#L24)

jeremyn:
I've updated the checklist through the latest commit 523f1c2a2fb6f5d7aa30f8cca7a44adf09304118.

I'm still not getting MCB on https://club.fir.im and I'm not sure what the comment `( https://club.fir.im/ use https://firimg.fir.im/ )` means.

Also, did you decide to not do the `qbox.me` change?

renyouguo:
https://club.fir.im : 
sp160910_231737
http://club.fir.im : 
sp160910_231851

qbox.me : 
#6829

jeremyn:
Thanks for the clarification. Mixed content blocking is what's described at https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox , where the browser blocks HTTP resources from being loaded on an HTTPS page. That's not happening for https://club.fir.im , so it's not MCB as we usually define it. Instead, the site uses leading `//` in the `src` for these images, but the `https://` version of these images has some problem such as an invalid certificate, so the images don't load.

I think that if the only thing that doesn't load are forum avatar images, then we should ignore that problem and include the site as a target. What do you think?

I've also merged the `qbox.me` pull request #6829 .

renyouguo:
I think we should disable it before the site fix this issue.
I have changed the comment.

jeremyn:
That's fine, thanks. Merged.

PR #6059, opened 2016-08-03
Update Hatena (problematic domains)

luixxiul:


fuglede:
I don't quite understand this one: It looks like https://cdn.api.b.hatena.ne.jp and http://cdn.api.b.hatena.ne.jp serve the same content (a 301 redirect), and that this host could actually be added as a target host.

luixxiul:
@fuglede ok updated.

src/chrome/content/rules/Hatena.xml
```diff
@@ -55,7 +55,7 @@
 
 	<target host="*.hatena.ne.jp" />
 	<target host="*.st-hatena.com" />
-
+	<target host="cdn.api.b.hatena.ne.jp" />
```
fuglede: Thanks for the update! It looks like I was a bit too fast in my previous comment: This target host is not actually necessary, since it is covered by the wildcard target in line 56; would you mind removing it again? Only the rule that you added is necessary.

luixxiul:
@fuglede how about now?

fuglede:
LGTM, thanks for the updates.

PR #6061, opened 2016-08-03
[WIP] Added gnavi.co.jp and gnst.jp

luixxiul:


luixxiul:
Travis says that "Test URL test succeeded" but I get the error in the job. I'm not quite sure why. Will you please check it? Thanks. /cc: @fuglede

fuglede:
The Travis setup runs two different shell scripts: One which focusses on ensuring that the redirects behave as expected, running through all target hosts and test URLs, and one which does basically everything else (including launching the browsers to see that nothing is broken). In this case, the former test works as expected, but the other one fails.

Now, it looks like the test failure is just due to a temporary failure within Travis, so let me just rerun it and see what happens.

PR #6070, opened 2016-08-04
Added liginc.co.jp

luixxiul:


jeremyn:
Thanks for the ruleset!

`creative.liginc.co.jp` is broken even if I disable mixed content protection.

luixxiul:
Thanks for reviewing.
````
http://creative.liginc.co.jp/wp-content/themes/ligworks/fonts/ligiconfont.woff?h7yjwt' from origin 'http://creative.liginc.co.jp' has been blocked by CORS policy
````
So I removed it.

jeremyn:
The domain in the comment should be `creative`, not `creatives`.

luixxiul:
Updated.

PR #6071, opened 2016-08-04
Added advision-adnw.jp

luixxiul:


fuglede:
For the less obvious of these, it can be helpful to add a test URL illustrating a non-404 response and the purpose of the domain. Just something like http://100540.advision-adnw.jp/t/001/797/a1001797.js to show that the domain hosts actual content.

PR #5119, opened 2016-06-13
[WhiteHouse.gov.xml] remove already preloaded

nextPrime:
https://whitehouse.gov has been preloaded in Chrome's HSTS preload list for quite a long time. So there is no need to keep this ruleset in HTTPS Everywhere. (Related: #2163, #3454)

[1] https://hstspreload.appspot.com/?domain=whitehouse.gov

PR #6075, opened 2016-08-05
[OPM.gov.xml] remove a redundant rule

nextPrime:
https://www.leadership.opm.gov/ no longer uses mismatched certificate, and it can successfully redirect to https://leadership.opm.gov/ by the server.

[1] https://www.ssllabs.com/ssltest/analyze.html?d=www.leadership.opm.gov&hideResults=on


nextPrime:
Related: #5119. https://www.servicesonline.opm.gov/ returns 403 whenever it is accessed by a non-browser client. It detects this by checking if both `User-Agent` and `Connection: Keep-Alive` request headers exist. Could we change our tester to mimic a browser's request?

Hainish:
lgtm

Issue #6079, opened 2016-08-06
Build `firefox-latest` broken


fuglede:
In all builds using `FIREFOX=firefox-latest`, the following errors occur, while the `firefox-esr-latest` appears to work as intended:

	+ echo 'running tests'

	running tests

	+ '[' -n /firefox-latest/firefox/firefox ']'

	+ xvfb-run cfx test -b /firefox-latest/firefox/firefox --profiledir=/tmp/tmp.xigEYiSDRl --verbose

	Using binary at '/firefox-latest/firefox/firefox'.

	Using profile at '/tmp/tmp.xigEYiSDRl'.

	libGL error: No matching fbConfigs or visuals found

	libGL error: failed to load driver: swrast

	process 211: D-Bus library appears to be incorrectly set up; failed to read machine uuid: UUID file '/etc/machine-id' should contain a hex string of length 32, not length 0, with no other text

	See the manual page for dbus-uuidgen to correct this issue.

	process 211: D-Bus library appears to be incorrectly set up; failed to read machine uuid: UUID file '/etc/machine-id' should contain a hex string of length 32, not length 0, with no other text

	See the manual page for dbus-uuidgen to correct this issue.

	[211] WARNING: Client failed to connect to the D-BUS daemon:

	/usr/bin/dbus-launch terminated abnormally without any error message: 'glib warning', file /builds/slave/m-rel-l64-00000000000000000000/build/src/toolkit/xre/nsSigHandlers.cpp, line 142

	(firefox:211): GConf-WARNING **: Client failed to connect to the D-BUS daemon:

	/usr/bin/dbus-launch terminated abnormally without any error message

	process 211: D-Bus library appears to be incorrectly set up; failed to read machine uuid: UUID file '/etc/machine-id' should contain a hex string of length 32, not length 0, with no other text

	See the manual page for dbus-uuidgen to correct this issue.

	[211] WARNING: Client failed to connect to the D-BUS daemon:

	/usr/bin/dbus-launch terminated abnormally without any error message: 'glib warning', file /builds/slave/m-rel-l64-00000000000000000000/build/src/toolkit/xre/nsSigHandlers.cpp, line 142

	(firefox:211): GConf-WARNING **: Client failed to connect to the D-BUS daemon:

	/usr/bin/dbus-launch terminated abnormally without any error message

[...]

PR #6082, opened 2016-08-06
Added maist.jp

luixxiul:


luixxiul:
This one also ``Test URL test succeeded.``

fuglede:
Regarding the comment: yep, the build error is unrelated (#6079).

PR #6085, opened 2016-08-06
Added apps-cms.jp

luixxiul:


fuglede:
This one is problematic: By default in both Firefox and Chrome, a lot of active content (including http://apps-cms.jp/style.css and http://app-cms.jp/lp/jquery.js) is blocked by the rule, meaning that the site layout is broken. You can either [mark the ruleset as mixedcontent](https://www.eff.org/https-everywhere/rulesets#mixed-content), or secure those parts that can be secured without breaking the layout.

luixxiul:
ok, updated.

fuglede:
Thanks for the update! For the record, if you want to see the difference locally, you can compare the result of viewing it in Firefox by changing the value of `security.mixed_content.block_active_content`in `about:config`.

PR #6088, opened 2016-08-06
Make trivial rule for particle.io

jish:
Particle uses quite a few subdomains and most appear to support HTTPS.
`community` in particular, needs a redirect rule as it does not
redirect to HTTPS automatically.

### DIFF
--- a/src/chrome/content/rules/Particle.io.xml
+++ b/src/chrome/content/rules/Particle.io.xml
@@ -1,0 +1,11 @@
+<ruleset name="Particle.io">
+	<target host="particle.io" />
+	<target host="www.particle.io" />
+
+	<target host="build.particle.io" />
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks for adding this!

Could you add comments on the top, that blog.particle.io and status.particle.io are serving a mismatching cert?
projects.particle.io also works with https.

jish:
I looked through some of the other files to get a sense of the comment style. I added this comment to the top of the file. If that is not correct, or if you have a guide I can look at, then I can adjust the comments.

```
<!--

	Problematic domains:

		- blog.particle.io	(mismatched)
		- status.particle.io	(mismatched)

-->
```

I also added a `<target>` tag for `projects.particle.io`.

J0WI:
Thanks!

PR #6093, opened 2016-08-07
Update AddThis.com

luixxiul:


J0WI:
Could you also add `status.addthis.com` and a comment for mismatching `go.addthis.com`?

luixxiul:
@J0WI OK, I updated it.

PR #6100, opened 2016-08-07
Added adlantis.jp

luixxiul:
@fuglede, @J0WI: please let me know if ``static.adlantis.jp`` should be added. I will rebase commits based on your suggestion.

### DIFF
--- a/src/chrome/content/rules/adlantis.jp.xml
+++ b/src/chrome/content/rules/adlantis.jp.xml
@@ -1,0 +1,24 @@
+<!--
+
+	Nonfunctional domains:
+
+		- click	(times out)
+		- tn	(connection closed)
+
+	Problematic subdomains:
+
+		- blog	(cert invalid)
+		- status	(cert invalid)
+
+-->
+<ruleset name="adlantis.jp">
+
+	<target host="static.adlantis.jp" />
+	<target host="www.adlantis.jp" />
+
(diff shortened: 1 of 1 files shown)


fuglede:
If you believe that `static.adlantis.jp` serves identical content on HTTP and HTTPS (apart from perhaps what is served on `/`), then do add it.

It could be useful to have some examples of content that is served there. From a quick look at it, all I could find was http://static.adlantis.jp/javascripts/AdLantisLoader.js?20090519 and https://static.adlantis.jp/javascripts/AdLantisLoader.js?20090519 (which are identical according to `diff`).

Is there any particular reason you believe that adding it could be problematic?

luixxiul:
@fuglede I noticed akamai servers were excluded from the targets, so I just thought other CDN traffics should be excluded as well.

fuglede:
In other rulesets, you mean? I think what happened with Akamai was that their CDN buckets proved to be a bit tricky to cover in a maintainable manner, as they turned out to change -- at least at one point of time -- their backends in ways that would break HTTPS Everywhere for all rulesets that tried to secure their connections.

Unless something similar could happen here, securing the CDN host should not be an issue. But it's great that you took the time to look into it.

luixxiul:
@fuglede OK, I rebased them.

fuglede:
Super, thanks.

PR #6108, opened 2016-08-07
Added fam-ad.com

luixxiul:


jeremyn:
Thanks for the ruleset!

`fam-ad.com` redirects to `admin.fam-ad.com` which you've correctly excluded as broken. Since the ruleset is so incomplete, can you please change the name to `fam-ad.com (partial)`?

luixxiul:
OK, done.

PR #6128, opened 2016-08-08
Updated marketo.com

luixxiul:


luixxiul:
@fuglede will you please let me know how to fix the error?

fuglede:
There seems to be a couple of unrelated errors. Let's look at them one at a time.

    ERROR src/chrome/content/rules/Marketo.xml: Non-2xx HTTP code: http://pages2.marketo.com/rs/marketob2/images/logo_marketo_new.png (200) => https://na-ab13.marketo.com/marketob2/images/logo_marketo_new.png (403)

I don't see why the one above is happening. Perhaps due to temporary issues at the time of testing.

    ERROR src/chrome/content/rules/Marketo.xml: Fetch error: http://launchpoint.marketo.com/ => https://launchpoint.marketo.com/: Too many redirects while fetching 'https://launchpoint.marketo.com/'

It you've also noticed yourself, the `launchpoint` subdomain is problematic as it redirects HTTPS connections to HTTP. The fetch test records this issue by counting the number of redirects that occur. It looks like the redirects occur for all requests to that domain, so that all we can do is to remove it from [the rule](https://github.com/luixxiul/https-everywhere/blob/b4d3d9672c9b122cde78770b9bf491af9325632d/src/chrome/content/rules/Marketo.xml#L147).

If you fix the latter issue, the former issue might fix itself (note that you might have to rebase the commits on top of `master`, as the testing framework has recently changed, but maybe you can get through without).

src/chrome/content/rules/Marketo.xml
```diff
@@ -132,7 +132,7 @@
 		<test url="http://pages2.marketo.com//?" />
 
 	<rule from="^http://pages2\.marketo\.com/(?=css/|images/|js/|rs/)"
-		to="https://na-abk.marketo.com/" />
+		to="https://pages2.marketo.com/" />
```
luixxiul: This solves the error in fetch test.

luixxiul:
@fuglede thanks, I updated the rule.

fuglede:
Great, thanks for the update.

PR #6131, opened 2016-08-08
Added prtimes.jp

luixxiul:


jeremyn:
Thanks for the ruleset! Please:

* Change the name to `prtimes.jp (partial)`
* Change the tests to `http://prtimes.jp/common/`, `http://prtimes.jp/i/` and `http://prtimes.jp/prn/`

luixxiul:
Thanks, but I think changing the test URLs to these will fail because of 403 error.

jeremyn:
The simpler test URLs pass when I run `fetch-test.sh` locally.

luixxiul:
clipboard01

jeremyn:
I see what you mean. That's all right though, because the important thing for the test is that http and https both return the same thing. even if that's an http 4xx error. If http://prtimes.jp/common/ is 200 and https://prtimes.jp/common/ is 403, that's bad, but if they are both 403, that's fine. Generally a test URL should be as simple as possible.

luixxiul:
Please make sure in other rulesets I am specifying resources based on the feedback of @Hainish here: https://github.com/EFForg/https-everywhere/pull/6219#issuecomment-239307637

luixxiul:
where ``f1.zenclerk.com`` returns the 403 error just like http://prtimes.jp/common/ fyi

jeremyn:
@Hainish Can you take a look at this? The question is which of these is a preferable test for

```xml
	<rule from="^http://prtimes\.jp/(common/|i/|prn/)"
		to="https://prtimes.jp/$1" />
```

Either

```xml
<test url="http://prtimes.jp/common/v4/html/js/single-post/main.js" />
```

or

```xml
<test url="http://prtimes.jp/common/" />
```

I prefer the second one, but @luixxiul points out that your comment [here](https://github.com/EFForg/https-everywhere/pull/6219#issuecomment-239307637) suggests you prefer the first one.

Hainish:
Either will do, but why not both?  The more test coverage the better.

jeremyn:
@Hainish If the tests are effectively redundant then then we shouldn't have both. I prefer the simpler one because it's the minimal example to test what's being tested, and also because it's (probably) less likely to be changed by the site owner. But, in the future I won't push back on contributors to simplify their tests.

Merged. Thanks @luixxiul.

PR #6138, opened 2016-08-08
Added yicha.jp

luixxiul:


luixxiul:
@fuglede would you please tell me how to make travis-ci test pass in this case?

PR #6140, opened 2016-08-08
Added nex8.net

luixxiul:


src/chrome/content/rules/nex8.net.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+
+	Nonfunctional subdomains:
+
+		- blog	(times out)
+
+-->
+
+<ruleset name="nex8.net">
+
```
fuglede: Could you add the bare host nex8.net as a target? Otherwise LGTM.

fuglede:
@hainish: There's probably a good reason, but since I'm not entirely sure what it is: why were these seemingly fine PRs closed? Something to do with the branches being deleted?

PR #6150, opened 2016-08-09
Create HDFCBank.xml

niket-play:
Ruleset for HDFC Bank, India.


J0WI:
Please have a look at our guides:

Wildcard rules need at least 3 valid test urls. You should also search for more subdomains to make sure, that this rule break no other subdomains. The best would be to have each subdomain as target or test url.

J0WI:
I'll prefer #6804. Thanks for this PR anyway!

PR #6154, opened 2016-08-09
Added app-liv.jp

luixxiul:


luixxiul:
@fuglede I am not sure how to exclude only the top level eg. http://android.app-liv.jp/

J0WI:
Can you document the http redirect on ^ and www.?

PR #6155, opened 2016-08-09
Line

luixxiul:


src/chrome/content/rules/Line.me.xml
```diff
@@ -0,0 +1,60 @@
...
+		- play	(invalid cert)
+
+	Mixedcontent:
+
+		- fs
+
+-->
+<ruleset name="Line.me">
+
+	<target host="*.game.line.me" />
```
J0WI: Please provide at least 3 test domains for this wildcard.

src/chrome/content/rules/Line.me.xml
```diff
@@ -0,0 +1,60 @@
+<!--
+
+	Fot more Line coverage, see Naver.jp.xml
+
+	Nonfunctional subdomains:
+
+		- ^	(redirected to http)
+		- at	(redirected to http)
```
J0WI: also on www

PR #6162, opened 2016-08-09
Gmo internet

luixxiul:


luixxiul:
@fuglede the travis error about http://point.gmo.jp/ is only applied to the top level. Would you please tell me how to fix the error?

PR #6164, opened 2016-08-09
Added .editorconfig

### LINKED ISSUE #5837: .editorconfig
Let's set up our own :-)
http://editorconfig.org/

luixxiul:
Closes #5837

cc: @fuglede, @J0WI, @Hainish

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[src/chrome/content/rules/*.xml]
+indent_style = tab
+indent_size = 4


Hainish:
Neat!

luixxiul:
Should ``indent-size`` be 8? /cc: @Hainish

Hainish:
I think 4 is fine for the `xml` indentation

luixxiul:
It seems to me that the tab size became smaller because of the setting from [this](https://github.com/EFForg/https-everywhere/pull/6217/files) to [this](https://github.com/EFForg/https-everywhere/pull/5976/files). Let me know if it is not good.

Hainish:
That probably depends on your editors previous setting.  One thing we could do is leave it unspecified, and that way indentation will be tabs but we won't be prescriptive in how that's displayed in the editors of our contributors.

luixxiul:
OK got it, thanks.

Hainish:
Thank you!

PR #6175, opened 2016-08-09
gismeteo.by

galeksandrp:


J0WI:
It seems that they are all related:

#6175
#6176
#6177
#6178
#6179

Could you link them with a comment on the top if this is true?

J0WI:
Also note that there is already Gismeteo.ua.xml, you should mention that too.

All rules lgtm.

galeksandrp:
Linked.

PR #6181, opened 2016-08-10
Add hungryhowies.com

Caboosey:


src/chrome/content/rules/hungryhowies.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="hungryhowies.com">
+        <target host="hungryhowies.com" />
+        <target host="www.hungryhowies.com" />
```
J0WI: Please also add the following targets:
franchising.hungryhowies.com
home.hungryhowies.com

Caboosey: Added.

src/chrome/content/rules/hungryhowies.com.xml
```diff
@@ -0,0 +1,9 @@
...
+        <target host="franchising.hungryhowies.com />
+        <target host="home.hungryhowies.com />
```
J0WI: There is a missing `"` resulting in a syntax error.

J0WI:
Thanks!

PR #6187, opened 2016-08-10
Added GradeScope.com

rohitjha:


src/chrome/content/rules/Gradescope.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Gradescope.com">
+	<target host="gradescope.com" />
+	<target host="www.gradescope.com" />
```
J0WI: Please also add `blog.gradescope.com`

src/chrome/content/rules/Gradescope.com.xml
```diff
@@ -0,0 +1,7 @@
...
+
+	<rule from="^http://gradescope\.com/" to="https://gradescope.com/" />
+	<rule from="^http://www\.gradescope\.com/" to="https://www.gradescope.com/" />
```
J0WI: Please just use `<rule from="^http:" to="https:" />`

J0WI:
Thanks!

PR #6188, opened 2016-08-10
Create luosimao.com.xml

renyouguo:


renyouguo:
@Hainish

Hainish:
Again, `<test>` URLs are helpful here.  All I see is a 404 when I try to access `client.luosimao.com`

J0WI:
Yes, test urls would be helpful for every host that does not return 200, especially CDNs like s.luosimao.com.

rule itself lgtm.

J0WI:
You need to exclude ^my.luosimao.com/$ for Travis.

J0WI:
Thanks!

PR #6191, opened 2016-08-10
Added Peerstudio.org

rohitjha:


src/chrome/content/rules/Peerstudio.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Peerstudio.org">
+	<target host="peerstudio.org" />
+	<target host="www.peerstudio.org" />
+
+	<rule from="^http://peerstudio\.org/" to="https://peerstudio.org/" />
+	<rule from="^http://www\.peerstudio\.org/" to="https://www.peerstudio.org/" />
```
J0WI: Please just use `<rule from="^http:" to="https:" />`

PR #6184, opened 2016-08-10
Added rule for academicintegrity.ucsd.edu in UCSD.edu.xml

rohitjha:
Added rule for academicintegrity.ucsd.edu as it works fine with both [HTTP](http://academicintegrity.ucsd.edu) and [HTTPS](https://academicintegrity.ucsd.edu).

PR #6192, opened 2016-08-10
Added podcast.ucsd.edu

rohitjha:


J0WI:
See https://github.com/EFForg/https-everywhere/pull/6184#issuecomment-241251059

PR #6216, opened 2016-08-11
Added A8.net

luixxiul:


J0WI:
rule LGTM, but why is it partial?

luixxiul:
Because this is an affiliate service site. There would be other subdomains on which it serves resources like beacons (see the list of the observed domains https://www.virustotal.com/en/domain/a8.net/information/). Still, if we call it incomplete, which is normal, rather than partial, I will remove it from the name.

J0WI:
`(partial)` is used when not all resources on a page can be secured (e.g. just static content). But we have no strong guideline about that.

luixxiul:
Updated with including other domains.

J0WI:
Thanks!

PR #6219, opened 2016-08-11
Added zenclerk.com

luixxiul:


Hainish:
@luixxiul test url looks like blank HTML.  I'm reluctant to include a domain unless we can see a clear utility for it, otherwise it's just extra kbytes sitting on our users machines.

luixxiul:
ok I updated with a new example.

PR #6226, opened 2016-08-12
[a4apphack] Update CDN bucket (old one still works)

mnordhoff:


J0WI:
@mnordhoff could you explain why we need this change (because the old one still works)?

mnordhoff:
@J0WI *Need* is a strong word. :P It may have slightly improved performance (old bucket is in the US, new bucket is a CDN).

At this point the subdomain no longer seems to resolve, so i suppose this pull request can be closed. :P (Or superseded by a different one to remove the rule entirely.)

J0WI:
Thanks, I just created #9179

PR #6237, opened 2016-08-12
Fix #5804 - rules for landsend.com.

### LINKED ISSUE #5804: Broken ruleset for Lands' End hides product listings
The ruleset for Lands' End prevents search results or product listing from displaying.

To reproduce:

1. Go to landsend.com
2. Click any category of product.

tomwys:


Hainish:
Looks like the rules for this ruleset only apply to `landsend.com`, `www.landsend.com`, and `business.landsend.com`.  The targets include `landsend.com` and `*.landsend.com`.  Better to delete the wildcard target and add the targets `www` and `business` so they match the rules.  I also see that `www.fr.landsend.com`, `loadtest01.landsend.com`, and `ocsex.landsend.com` all have https available.

Issue #4671, opened 2016-04-22
Error in "Visa"-rule


KommX:
Visa.xml breaks the following link:
http://www.visa.com/atmlocator

jeremyn:
Pull request #6244 should resolve this issue.

jeremyn:
Pull request #6244 has been merged so I'm closing this issue, but @KommX please feel free to reopen this issue and add a comment if the issue isn't fixed for you.

PR #6244, opened 2016-08-13
Add Visa.xml atmlocator exclusion (issue #4671)

jeremyn:
This should resolve issue #4671.

### DIFF
--- a/src/chrome/content/rules/Visa.xml
+++ b/src/chrome/content/rules/Visa.xml
@@ -41,13 +41,22 @@
 
 	<rule from="^http://corporate\.visa\.com/_(cs|image)s/"
 		to="https://corporate.visa.com/_$1s/" />
+	<test url="http://corporate.visa.com/_css/" />
+	<test url="http://corporate.visa.com/_images/" />
 
 	<rule from="^http://usa\.visa\.com/(css|ext|img|merchants)/"
 		to="https://usa.visa.com/$1/" />
+	<test url="http://usa.visa.com/css/" />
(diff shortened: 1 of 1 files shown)


J0WI:
Could you also add a comment about the issue?

src/chrome/content/rules/Visa.xml
```diff
@@ -31,6 +31,10 @@
 	<!--securecookie host="^usa\.visa\.com$" name="^visaAnonCookie$" /-->
 
 
+	<exclusion pattern="^http://(?:www.)?visa\.com/atmlocator" />
```
J0WI: Prefer capturing groups (www\.)? over non-capturing (?:www\.)?.

jeremyn:
I've changed the exclusion to use a capturing group, thanks.

I'm not sure what you mean about a comment.

J0WI:
Just a comment like "page returns 404 error on https, see bug 4671"

jeremyn:
I added a comment. The issue doesn't have extra information so I don't think linking to it is useful, and it's referenced in the first commit message anyway. If you feel strongly about linking to the issue in the ruleset, I'll modify the comment, just let me know.

J0WI:
Rule lgtm, but this rule needs test coverage update.

jeremyn:
I added tests but after checking around on the site, I think the ruleset needs to be reviewed generally.

J0WI:
Of course this is just a specific bugfix without fully rebasing it, but it doesn't make the rule worse :)

Issue #4975, opened 2016-05-30
Remove duplicated ruleset whitelist entries


J0WI:
Just discovered that the list contains some duplicated entries that are not removed by the script in #4830. So the script should contain some `sort | uniq` magic.

fuglede:
I did consider having the script remove the dupes (cf. the remark at the end of https://github.com/EFForg/https-everywhere/pull/4830#issue-155598754) but honestly didn't bother, as they are comparatively few in number. That being said, it is a very natural addition to [utils/ruleset-coverage-whitelist-cleanup.sh](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist-cleanup.sh).

jeremyn:
Pull request #6246 should resolve this issue.

PR #6246, opened 2016-08-13
Update whitelist cleanup script (issue #4975)

jeremyn:
This should resolve issue #4975. Just to note, this updated script removes 227 lines from `ruleset-coverage-whitelist.txt` compared to the original script removing 90.


Hainish:
lgtm

Issue #6240, opened 2016-08-12
AttributeError or TypeError after running fetch-test.sh


jeremyn:
I regularly get an error after running `fetch-test.sh`. I've included the output of `./fetch-test.sh rules/EFF.xml` from Fedora 24 and then from Ubuntu 14.04:

Fedora 24 / Python 2.7.12:
```
INFO Finished comparing http://www.eff.org/ -> https://www.eff.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://action.eff.org/ -> https://action.eff.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://supporters.eff.org/ -> https://supporters.eff.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://eff.org/ -> https://eff.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://defendinnovation.org/ -> https://defendinnovation.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://www.defendinnovation.org/ -> https://www.defendinnovation.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://globalchokepoints.org/ -> https://globalchokepoints.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://www.globalchokepoints.org/ -> https://www.globalchokepoints.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://httpsnow.org/ -> https://httpsnow.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://www.httpsnow.org/ -> https://www.httpsnow.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://ripmixmake.org/ -> https://ripmixmake.org/. Rulefile: rules/EFF.xml.
INFO Finished comparing http://www.ripmixmake.org/ -> https://www.ripmixmake.org/. Rulefile: rules/EFF.xml.
INFO Finished in 48.53 seconds. Loaded rulesets: 1, URL pairs: 12.
Exception in thread Thread-1 (most likely raised during interpreter shutdown):
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 804, in __bootstrap_inner
  File "test/rules/src/https_everywhere_checker/check_rules.py", line 91, in run
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'exception'
```

Ubuntu 14.04 / Python 2.7.6:
```
[...]

PR #6247, opened 2016-08-13
Break loop in thread run function (issue #6240)

jeremyn:
This pull request should resolve issue #6240.

@Hainish, can you review this please?

### DIFF
--- a/test/rules/src/https_everywhere_checker/check_rules.py
+++ b/test/rules/src/https_everywhere_checker/check_rules.py
@@ -83,12 +83,11 @@ class UrlComparisonThread(threading.Thread):
 		threading.Thread.__init__(self)
 
 	def run(self):
-		while True:
-			try:
-				self.processTask(self.taskQueue.get())
-				self.taskQueue.task_done()
-			except Exception, e:
-				logging.exception(e)
+		try:
+			self.processTask(self.taskQueue.get())
+			self.taskQueue.task_done()
+		except Exception, e:
+			logging.exception(e)
 
 	def processTask(self, task):
 		problems = []


jeremyn:
@fuglede @Hainish @J0WI Can someone please review this? The problem is annoying.

Hainish:
This change will make it so that if the number of files being checked exceeds the number of threads allowed, the process never completes.

For the fetch test, the number of threads is specified in https://github.com/EFForg/https-everywhere/blob/master/test/rules/http.checker.config#L23 as `40`.  To reproduce the above:

1. Change the number of threads in https://github.com/EFForg/https-everywhere/blob/master/test/rules/http.checker.config#L23 to `1`
2. Modify > 1 ruleset
3. Run the fetch test

It will never exit.

This is because without the loop, the UrlComparisonThread only checks exactly one changed ruleset (specified as ComparisonTasks in `taskQueue`) before exiting.  If the number of threads allowed exceeds the number of rulesets changed, there will not be enough threads to process the `taskQueue`.  Thus, https://github.com/EFForg/https-everywhere/blob/master/test/rules/src/https_everywhere_checker/check_rules.py#L433 never completes.

jeremyn:
That makes sense. My normal use case is running `fetch-test.sh` against one ruleset so I didn't notice the problem.

Regarding your improved fix in #6692: it looks like originally, the threads were started before the taskQueues were filled, so `while True` was a "lazy" way of making sure everything put into the queues would be completed. Because you now properly check that the queue is empty, you can't start the threads until after the queues have actually been filled. Do I understand the fix correctly?

Hainish:
That's almost right.  Here:

> Because you now properly check that the queue is empty, you can't start the threads until after the queues have actually been filled.

I would phrase this differently - it is not that we can't start the threads until the queues have been filled *because* we check that the queue is empty.  It's that, procedurally, we first fill the queue, and only then run the `UrlComparisonThread` which begins processing the queue, exiting when the queue is empty.

If we changed the conditional without moving the block below, all the `UrlComparisonThreads` would immediately exit, because the queue would be empty, because we wouldn't have filled it with tasks yet.

jeremyn:
Thanks @Hainish .

Closing pull request.

PR #6248, opened 2016-08-14
Added MERY.jp

luixxiul:


src/chrome/content/rules/MERY.jp.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+
+	Problematic subdomains:
+
+		- assets	(akamai)
+		- bravo	(akamai)
+		- cashmere	(akamai)
+		- content	(akamai)
+		- galena	(akamai)
```
J0WI: Why are they problematic?
They all WFM.

Please also add www. (redirects to plain ^)

luixxiul: Don't we need to remove traffics through Akamai? https://github.com/EFForg/https-everywhere/pull/6100#issuecomment-238074839
I'm confused here :-/

luixxiul: Or should I just rename it nonfunctional?

J0WI: I think it's just a problem when you redirect something like `images.foo.bar` to `eu1337.cdn.com/foo-images`, but as long as content live on their own domain it souldn't be a problem.

luixxiul: I see, later I will review my other PRs which include akamai traffics.

fuglede: Ah, indeed, and even then, redirecting to another domain can be completely fine; it's just that it turned out to break a lot of rulesets in the case of Akamai at one point in time. Apologies for any confusion.

luixxiul: I see, no problem. Maybe we'll have to take some time to enable the domains which have been excluded due to the case ;-)

luixxiul:
@J0WI OK, I updated the ruleset.

J0WI:
Thanks!

J0WI:
Could you also add/document www.?

J0WI:
Oh and could you also include some test urls for the CDN hosts? Then we will be more likely aware if they change something that breaks our rule.

luixxiul:
@J0WI OK, done.

PR #6273, opened 2016-08-14
Added pejs.net

luixxiul:


Foorack:
@luixxiul Out of curiosity, what method do you use when generating new rules?
Are you doing it manually by hand or have you made some tool to automate the file-generation process?

luixxiul:
By hand at all. If you go to some news sites you'll get a bunch of unencrypted 3rd party resources.

PR #6277, opened 2016-08-14
Add Braveinfo.net

Foorack:


Foorack:
Accidentally added multiple sites. Redoing.

PR #6282, opened 2016-08-14
Add 9to5mac.com

Foorack:


Foorack:
I have now lean't it should be Added instead of Add. Will fix in future commits.

J0WI:
Thanks for adding this!
Could you also add `academy.9to5mac.com` as target host?
Looks good to me then.

Foorack:
@J0WI Fixed.

J0WI:
Thanks!

PR #6284, opened 2016-08-14
Added 9to5toys.com

Foorack:


src/chrome/content/rules/9to5toys.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="9to5toys.com">
+	<target host="9to5toys.com" />
+	<target host="www.9to5toys.com" />
```
J0WI: + specials.9to5toys.com

Foorack:
@J0WI Fixed. Added specials.9to5toys.com

PR #6288, opened 2016-08-14
BorgensTidende updates

numismatika:
- convert to current styleguide
- secure subfolders that don't redirect to plain


J0WI:
This host has a lot of more subdomains. Could you also look for them?

src/chrome/content/rules/BergensTidende.xml
```diff
@@ -55,25 +115,43 @@
...
+	<target host="mediahash2.bt.no" />
+	<target host="mediahash3.bt.no" />
+	<target host="mobil.bt.no" />
+	<target host="multimedia.bt.no" />
+	<target host="nyttig.bt.no" />
+	<target host="old.bt.no" />
+	<target host="reise.bt.no" />
+	<target host="statisk.bt.no" />
+
+	<rule from="^http://statisk\.bt\.no/" to="https://statisk.bt.no/" />
```
J0WI: This is already covered by the generic redirect rule below.

src/chrome/content/rules/BergensTidende.xml
```diff
@@ -55,25 +115,43 @@
...
+
+	<rule from="^http://www\.bt\.no/(external/|public/|resources/|skins/)" to="https://www.bt.no/$1" />
```
J0WI: Everything else will also be redirected by the generic redirect rule below.
Why do you use this redirect here? Also the reason for the exclusion on `/$` is not clear to me.

J0WI: Ah, but an exclusion for `http://www.bt.no/tv/` is required.

numismatika: @J0WI if the top folder redirects to plain but subfolders can be secured, you need an exclusion to make the test script happy.

J0WI: I was not able to reproduce that. Can you confirm?

numismatika: I will revisit all these norwegian news pages when i got time.
Might become a lot simpler if i extrapolate correctly from the aftenposten website.

J0WI:
This does not pass our tests.

numismatika:
@J0WI now they pass

src/chrome/content/rules/BergensTidende.xml
```diff
@@ -1,79 +1,181 @@
 <!--
 	Bergens Tidende
 
-	For rules causing false/broken MCB, see BT.no-falsemixed.xml.
```
J0WI: See also `BT.no-falsemixed.xml`

src/chrome/content/rules/BergensTidende.xml
```diff
@@ -1,79 +1,181 @@
...
-
+	Cert chain issue:
+		- scenario.bt.no
+
+	Cert expired:
+		- nyhetsbrev.bt.no
+		- pust.bt.no
+
+	Cert mismatch:
+	- a.bt.no
```
J0WI: Please use a consistent amount of tabs to indent the comments.

src/chrome/content/rules/BergensTidende.xml
numismatika: removed

J0WI:
Thanks!

PR #6296, opened 2016-08-15
Update Lifehacker.com.xml

jeremyn:
This pull request partially resolves issue #3069.

### DIFF
--- a/src/chrome/content/rules/Lifehacker.com.xml
+++ b/src/chrome/content/rules/Lifehacker.com.xml
@@ -1,21 +1,22 @@
 <!--
 	For other Gawker coverage, see Gawker.xml.
-
-
-	CN: *.a.ssl.fastly.net
-
 -->
-<ruleset name="Lifehacker.com" default_off="mismatched">
-
+<ruleset name="Lifehacker.com">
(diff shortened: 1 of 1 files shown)


J0WI:
It seems that every user gets his own subdomain on this network. So we should document this and use a wildcard.
Our code style also requires tabs instead of spaces.

jeremyn:
I've made updates to the ruleset and squashed it into one commit.

I'm having trouble with `securecookie`. Note it needs to catch all of `lifehacker.com`, `.lifehacker.com`, `skillet.lifehacker.com`, `.skillet.lifehacker.com` etc. With the very general configuration in my pending change, on https://lifehacker.com it's missing `lifehacker.com > __k_iut`. I've also tried to be more specific:

```xml
	<securecookie host="^\.?lifehacker.com$" name=".+" />
	<securecookie host="^\.?\w+\.lifehacker.com$" name=".+" />
```

but it's still missing `__k_iut`. Do you have any suggestions?

J0WI:
No idea about the cookie issue. Is this a blocker?
Rule lgtm.

jeremyn:
Not a blocker for me, go ahead and merge, thanks.

PR #6301, opened 2016-08-15
Added dnsdumpster.com

Foorack:


J0WI:
This host seems to be down.

Foorack:
The host seems to be back up. My guess (from the type of service they are providing) is that someone is ddosing them. It showed server errors earlier today.

J0WI:
Merging, since the host seems to use https anyway.

PR #6318, opened 2016-08-15
PCWelt

numismatika:
- disable PCWelt rule for mobile & www subdomains, which redirect to plain
- add bilder subdomain


J0WI:
I think we can remove the mixed content platform then?
Could you also remove the wildcard and use bilder. as target host instead? This would make the whole rule easier to read. Also, comments should be on the top.

See also:
magazin.pcwelt.de (WFM)
software.pcwelt.de (http redirect)
browsercheck.pcwelt.de (http redirect)
www.browsercheck.pcwelt.de (mismatch)
newsletter.pcwelt.de (mismatch)
w3.pcwelt.de (mismatch)
shirtshop.pcwelt.de (mismatch)

J0WI:
This does not pass our tests.

numismatika:
@J0WI   checks are green now

PR #6320, opened 2016-08-15
UbuntuUsers: remove wildcards

numismatika:


J0WI:
Could you also document the mismatch on `staticwiki.ubuntuusers.de`?
The rules lgtm.

numismatika:
Sure, but that page is redirect to a totally different website. `http://shop.neuseeland-haus.de/`
I don't think it is used for any Ubuntu content at all

J0WI:
Thanks!

PR #6321, opened 2016-08-15
add rule for Bayern.de

numismatika:


rid007:
It's doesn't work properly with current HTTPS Everywhere browser plugin

numismatika:
@rid007  what exactly is wrong? The map service seems to be fine

rid007:
There is **no ruleset** with the latest version and also in **HTTPS Everywhere Atlas**. [https://www.eff.org/https-everywhere/atlas/?bayern](url)
There is a ruleset for statistik.bayern.de not from `www.bayern.de`
The main problem is they only the HTTPS cert. for `www.bayern.de` not for `bayern.de`.

src/chrome/content/rules/Bayern.de.xml
```diff
@@ -0,0 +1,27 @@
+<!-- cert mismatch on:
+	- www.ldbv.bayern.de
+	- www.gdi.bayern.de/
```
J0WI: At least `^` should also be documented.

src/chrome/content/rules/Bayern.de.xml
```diff
@@ -0,0 +1,27 @@
...
+	<target host="www.justiz.bayern.de"/>
+	<target host="www.pki.bayern.de"/>
+	<target host="rvr-prx.bayern.de"/>
+	<target host="intergeo10.bayernwolke.de"/>
+	<target host="intergeo11.bayernwolke.de"/>
+	<target host="intergeo12.bayernwolke.de"/>
+	<target host="intergeo13.bayernwolke.de"/>
+	<target host="intergeo14.bayernwolke.de"/>
+	<target host="intergeo15.bayernwolke.de"/>
+	<target host="intergeo16.bayernwolke.de"/>
```
J0WI: Could you try to add more subdomains?

src/chrome/content/rules/Bayern.de.xml
```diff
@@ -0,0 +1,27 @@
+<!-- cert mismatch on:
+	- www.ldbv.bayern.de
+	- www.gdi.bayern.de/
+
+	intergeo domains outside of the 10-16 range do not support TLS
+-->
+<ruleset name="Bayern.de">
```
J0WI: Please mark the rule as `partial` (depending on how many securable/non-working domains you will find).

J0WI:
Please merge `StatistikamtBayern.xml` into this rule.

jeremyn:
@numismatika Are you still interested in working on this?

numismatika:
It is just a matter of proper splitting, otherwise we will either get the rule disabled  if we go for wildcards or  never be able to merge it because of the large amount of subdomains.

jeremyn:
I recommend making this a `(partial)` ruleset and just including some of the most important subdomains. 1000 is way, way too many. Aim for more like 20-30.

PR #6334, opened 2016-08-16
Added adtechjp.com

luixxiul:


J0WI:
Please comment that `adtechjp.com` and `www.adtechjp.com` are refused.

PR #6335, opened 2016-08-16
Added wimg.jp.xml

luixxiul:


src/chrome/content/rules/wimg.jp.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="wimg.jp">
```
J0WI: Mismatch on ^wimg.jp

src/chrome/content/rules/wimg.jp.xml
```diff
@@ -0,0 +1,12 @@
...
+
+	<target host="i7.wimg.jp" />
+
+	<test url="http://i7.wimg.jp/profile/d290tm/20160727235827965_80.jpg" />
+
+	<securecookie host="." name="." />
```
J0WI: Please use `<securecookie host=".+" name=".+" />`

PR #6338, opened 2016-08-16
Added albennet.com

Foorack:


src/chrome/content/rules/Albennet.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+
+	Nonfunctional domains:
+
+		- alb (ERR_SSL_PINNED_KEY_NOT_IN_CERT_CHAIN)
```
J0WI: works for me

Foorack:
@J0WI The alb domain seems to work now for me well however this is one of my earlier pulls which I have lost write access to. (I deleted my fork and reforked without knowing this would happen) Have to recreate the pull.

Foorack:
New pull https://github.com/EFForg/https-everywhere/pull/6872

PR #6345, opened 2016-08-16
Added lhorn.de

Foorack:


src/chrome/content/rules/Lhorn.de.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Lhorn.de">
+	<target host="lhorn.de" />
+	<target host="www.lhorn.de" />
+
+	<rule from="^http://(?:www\.)?lhorn\.de/" to="https://www.lhorn.de/"/>
```
J0WI: please document the mismatch and remove the non-capturing group

J0WI:
See also `klaus.lhorn.de`

PR #6346, opened 2016-08-16
Added kdo.de

Foorack:


J0WI:
There are some more subdomains, would you like to add them too?

PR #6353, opened 2016-08-17
Added geokretymap.org

Rotzbua:


src/chrome/content/rules/geokretymap.org.xml
```diff
@@ -0,0 +1,8 @@
+<!--
+	Main page support https
+-->
+<ruleset name="geokretymap.org">
+	<target host="geokretymap.org" />
```
J0WI: api.geokretymap.org

Mismatch on www.

J0WI:
Thanks for adding this.
You may merge this into geokrety.org.xml or adding a comment on both files showing the relation.

Rotzbua:
@J0WI rebased and updated. I think 2 separate files are better. I added a comment, that both files are related.

src/chrome/content/rules/geokrety.org.xml
```diff
@@ -1,7 +1,12 @@
 <!--
 	Main pages support https
+	
+	Related page: geokretymap.org
```
J0WI: Please use the filename

Rotzbua:
@J0WI updated comment

J0WI:
thanks!

PR #6361, opened 2016-08-17
Added dy.fi

theel0ja:


J0WI:
Thanks for adding this rule. Sadly the certificate of this host expired recently.
When they fixed their cert, please also have a look at our [guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) about finding the subdomains of a host.

theel0ja:
Now they have Let's Encrypt enabled, so they have now a valid cert.

J0WI:
It seems that customers get their own subdomain, but without HTTPS support. Could you add a comment about this behavior?

theel0ja:
@J0WI Yes, they will not have HTTPS, but the main domain does (login pages etc)

J0WI:
@theel0ja
Sure. Please document this behavior with a comment in the ruleset for future maintenance.
The rule itself lgtm.

theel0ja:
@J0WI Is `<!-- Customers get their own subdomain, but they don't have SSL -->` good?

J0WI:
That's fine for me.

theel0ja:
Fixed.

PR #6374, opened 2016-08-17
githost.io

galeksandrp:


J0WI:
There are more subdomains and all are redirecting to https, send HSTS header, but have a expired cert.

galeksandrp:
Added.

PR #6378, opened 2016-08-17
styleci.io

galeksandrp:


J0WI:
This host is already in HSTS preload lists, so no need to add.
Thanks anyway!

PR #6385, opened 2016-08-17
finalbuilder.com

galeksandrp:


src/chrome/content/rules/finalbuilder.com.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+wiki.finalbuilder.com timed out
```
J0WI: timeout on ststic.
mismatch on help.

galeksandrp:
Added.

PR #6387, opened 2016-08-17
aspone.cz

galeksandrp:


J0WI:
@galeksandrp Please note that we use to set errors by type. This makes it much easier to read.

E.g.:
```
Mismatching hosts:
    - www.foo.example.com
    - cdn[1-3].example.com

HTTPS refused:
    - insecure.example.com
```
or alphabetical and with legend
```
    - www.foo.example.com °
    - cdn[1-3].example.com °
    - insecure.example.com *

° mismatch
* refused
```

PR #6389, opened 2016-08-17
go.cd

galeksandrp:


src/chrome/content/rules/go.cd.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="go.cd">
+	<target host="go.cd" />
+	<target host="www.go.cd" />
+	<target host="developer.go.cd" />
+	<target host="docs.go.cd" />
+	<target host="api.go.cd" />
```
J0WI: build.go.cd

Could you also sort them?

galeksandrp:
Added.

PR #6403, opened 2016-08-17
cloud66.com

galeksandrp:


Hainish:
@galeksandrp it does indeed look like http://cloud66.com has no https endpoint.  It also looks like `www` and `app` redirect to the http endpoint, in which case there would be no point in including them in this ruleset.  Finally, it looks like `birdseye` and `help` have mixed content warnings.

galeksandrp:
Wrong redirect. Fixed.

PR #6409, opened 2016-08-17
continuousdelivery.com

galeksandrp:


J0WI:
I also found `lapm.continuousdelivery.com`.
Why is this rule flagged as partial?

galeksandrp:
Added.

PR #6417, opened 2016-08-17
ponyfoo.com

galeksandrp:


src/chrome/content/rules/ponyfoo.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="ponyfoo.com">
+	<target host="ponyfoo.com" />
+	<target host="www.ponyfoo.com" />
```
J0WI: analytics.ponyfoo.com

galeksandrp:
Added.

PR #6418, opened 2016-08-17
risingstack.com

galeksandrp:


J0WI:
Please also add `resources.risingstack.com` and `quiz.risingstack.com`.
There is a mismatch on oneshot.risingstack.com

There are also some empty sites that say `This site is under development`, but I'm not sure if it's worth to comment about them.

galeksandrp:
Added.

PR #6427, opened 2016-08-18
Added chenhaot.com

Foorack:


src/chrome/content/rules/Chenhaot.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Chenhaot.com">
+	<target host="chenhaot.com" />
+	<target host="www.chenhaot.com" />
+
+	<rule from="^http://(?:www\.)?chenhaot\.com/" to="https://chenhaot.com/"/>
```
J0WI: www. works fine for me.

Foorack:
Again, this seems to be one of my older pulls which I cannot edit anymore. Will recreate pull with fix. Sorry for the trouble.

Foorack:
Updated version https://github.com/EFForg/https-everywhere/pull/6873.

PR #6435, opened 2016-08-18
Added images-dot.com

luixxiul:


src/chrome/content/rules/images-dot.com.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+
+	For more Asahi Shimbun coverage, see Asahi_Shimbun.xml
```
J0WI: Please also add a comment there.

J0WI:
Could you add a comment, that www does not exist?

PR #6437, opened 2016-08-18
spnet.ru

galeksandrp:


src/chrome/content/rules/spnet.ru.xml
```diff
@@ -1,24 +1,257 @@
...
+	<target host="pbx.spnet.ru" />
 	<target host="pop3.spnet.ru" />
 	<target host="pop3s.spnet.ru" />
 	<target host="smtp.spnet.ru" />
 	<target host="wiki.spnet.ru" />
+	<target host="zabbix.spnet.ru" />
+
+	<target host="www.spnet.ru" />
+
+	<rule from="^http://www\.spnet\.ru/"
```
J0WI: Please add also a comment about the issue

src/chrome/content/rules/spnet.ru.xml
```diff
@@ -1,24 +1,257 @@
...
+82.spnet.ru ¹
+846.spnet.ru ¹
+85.spnet.ru ¹
+89-222-212-0.spnet.ru ¹
+89-222-212-1.spnet.ru ¹
+89-222-212-2.spnet.ru ¹
+89-222-212-4.spnet.ru ¹
+89-222-212-5.spnet.ru ¹
+89-222-212-6.spnet.ru ¹
+89-222-212-7.spnet.ru ¹
```
J0WI: Well, you don't need to list all the reverse IP hosts. ;)

galeksandrp: For the luck, they are using wildcard domain *.spnet.ru (in addition to any-other-static-domains.spnet.ru), which cert is for hosting.spnet.ru only.

src/chrome/content/rules/spnet.ru.xml
J0WI: @galeksandrp I'm still missing this

galeksandrp: This is included in `*.spnet.ru`

PR #6449, opened 2016-08-18
Added ADResult.jp

luixxiul:


J0WI:
I don't think that I wildcard make sense here. 
Tghe only host that I found was `viewtest.adresult.jp` and that serves a mismatching cert too.

PR #6455, opened 2016-08-18
Updated Cint.com

luixxiul:


J0WI:
There is a mismatch on webmail. and various resources from \d+.cds.cint.com, but I'm not sure if it's safe to secure them all.

luixxiul:
Because not only ``\d+`` but also other arbitrary characters are included here, I mentioned and commented it out as a note.

PR #6457, opened 2016-08-18
Added WOVN.io

luixxiul:


src/chrome/content/rules/WOVN.io.xml
```diff
@@ -0,0 +1,21 @@
...
+		- t	(refused)
+
+	Fully covered subdomains:
+
+		- ^
+
+-->
+<ruleset name="WOVN.io">
+
+	<target host="j.wovn.io" />
```
J0WI: ^ and www can also be added as target hosts

luixxiul:
@J0WI OK, I updated.

PR #6460, opened 2016-08-18
Added R-oo.jp

luixxiul:


src/chrome/content/rules/R-oo.jp.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+
+	Nonfunctional subdomains:
+
+		- www
```
J0WI: www. works for me, ^ doesn't.

PR #6476, opened 2016-08-19
Updated chango.com

luixxiul:


J0WI:
Could you also check:
publisher.chango.com
radiology.chango.com
searchspots.chango.com
dev.chango.com
cw-m.d.chango.com
pdf.chango.com

PR #6493, opened 2016-08-20
Create 23zw.com.xml

renyouguo:


J0WI:
Please also add ^.
I get mixed content on all pages from `www.baidu.com/js/opensug.js`, can you confirm that this does not break anything?

PR #6498, opened 2016-08-20
Move Fnac.pt.xml into the rules directory

mnordhoff:
`Fnact.pt.xml` was in `src/chrome/`, not `src/chrome/content/rules/`.


fuglede:
Good catch!

PR #6520, opened 2016-08-21
Create boxun.com.xml

renyouguo:


src/chrome/content/rules/boxun.com.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Unable to visit:
+		- en
```
J0WI: seems to work outside of China, but has mixed content issues.

lqf96:
Maybe we don't need to care about China since this site seems to be blocked by GFW.

PR #6526, opened 2016-08-21
add rule for Epsilon Telecom

numismatika:


J0WI:
Please see also shop.epsilon-telecom.de and web.epsilon-telecom.de

J0WI:
Thanks!

PR #6470, opened 2016-08-19
Updated taboola.com

luixxiul:
Removed akamai traffics

PR #6540, opened 2016-08-22
Updatad taboola.com

luixxiul:
#6470

### DIFF
--- a/src/chrome/content/rules/Taboola.xml
+++ b/src/chrome/content/rules/Taboola.xml
@@ -10,14 +10,9 @@
 
 	Nonfunctional subdomains:
 
-
 		- blog	(refused)
 		- mb	(times out)
-		- cdn-yjp *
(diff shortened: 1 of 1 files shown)


luixxiul:
As the redirection of ``signup.taboola.com`` sometimes causes an 400 error and the the redirection seems to be configured properly by their side, I commented it out.

J0WI:
I was able to reproduce both, a 200 and 403 with https://signup.taboola.com/styles/assets/favicon.png and curl.
It turns out, that they use Round-robin DNS and one of their backend servers doesn't handle the request well.

luixxiul:
Would you please tell me whether I should stop commenting it out? /cc; @Hainish and @fuglede

J0WI:
Since this host seems quite frequent used, I would go for the exclusion, but you should update the comment to be more accurate.

Issue #6475, opened 2016-08-19
No HTTPS for www.porvoo.fi


nemobis:
www.porvoo.fi has no server configured for HTTPS. Should the rule be deleted? https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/porvoo.fi.xml

fuglede:
Thanks for the report! It should either be deleted or disabled (by adding `default_off="connection refused"` to the `ruleset` XML tag) -- disabling it can be useful if the failure is temporary as the ruleset history becomes more clearly preserved. Otherwise deletion is fine.

fuglede:
Just went ahead and submitted the fix as #6552.

PR #6552, opened 2016-08-22
[Porvoo.fi] Disable ruleset

fuglede:
This fixes https://github.com/EFForg/https-everywhere/issues/6475


J0WI:
They would support https on kartta.porvoo.fi and wilma.porvoo.fi

fuglede:
Updated.

J0WI:
Thanks :)

PR #5827, opened 2016-07-28
Updated rakuten.co.jp

luixxiul:
- Added information of nonfunctional domains, problematic domains, etc.
- Added domains linked from www.rakuten.co.jp/sitemap/
- Added domains on www.infoseek.co.jp

Closes #5824

PR #6554, opened 2016-08-22
Added rakuten-static.com

luixxiul:
#5827


luixxiul:
Closint this as merging this solely may break ruleset testing due to [this line](https://github.com/EFForg/https-everywhere/pull/5827/files#diff-26ed1b519bef0e9172c56c2156ca0641L4) in Rakuten.co.jp.xml. The commit above has been cherry-picked and pushed to ``luixxiul:Rakuten.co.jp`` in #5827.

PR #6562, opened 2016-08-23
Update NetEase.xml

renyouguo:


fuglede:
This domain (and actually also https://rev.da.netease.com) has [chain issues](https://www.ssllabs.com/ssltest/analyze.html?d=wr.da.netease.com&latest);

    $ curl -I https://rev.da.netease.com/

skaermbillede fra 2016-08-23 21 21 49

In fact, the problematic signee is RapidSSL SHA256 CA - G3 which is also causing problems in #6102, cf. #6569.

fuglede:
Thanks for the update!

PR #6565, opened 2016-08-23
Goldenline.pl

patryk:


J0WI:
Instead of a wildcard target, exclusions and test urls you should use a restrictive list of supported targets and documenting non-working hosts on the top using comment tags.

patryk:
@J0WI Changed.

J0WI:
Thanks!

PR #6610, opened 2016-08-26
Delete 23zw.com.xml

renyouguo:
The host owner has removed https……


J0WI:
We still have no clear guide line if such rules should be deleted or disabled.
What do you mean? @fuglede @jeremyn

jeremyn:
I think we should delete. It's simpler. Also we wouldn't create a rule for a site that doesn't have HTTPS, so why keep a rule around for a site without it?

fuglede:
How are we doing at keeping inactive rulesets out of memory? If that is no longer an issue, I would be all for keeping the rules, to provide history, and to let new contributors immediately know if the ruleset that they are currently working on has seen work earlier. Otherwise, I have no strong opinion on it.

Hainish:
Disabled rules are still kept in memory, and in fact still "enableable" from the browser.  For instance, direct your browser at http://static.atv.hu/, which is part of https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ATV.hu.xml.  You'll still see you can manually enable the rule from the drop-down menu.

So disabling a rule is useful if, for instance, some site has broken its HTTPS implementation, but may fix it in the future.  It lets users turn it back on once they do fix it, before we make a new release.  Also, if a site is only partially broken, disabling is preferable.  We don't want to impede functionality for users, but if some users are okay with partial site functionality and want the security regardless, they can manually enable t hat rule.  For sites that look like they've disappeared completely, we probably want to completely delete the ruleset, so that it isn't taking up space in memory.

renyouguo:
BTW, should we delete the sites that force https by server?
For instance:
ImQuQu.com.xml
https://imququ.com/post/moving-to-https-asap.html

jeremyn:
@renyouguo Sites that redirect from HTTP to HTTPS are still valid for HTTPS Everywhere, see [here](https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?).

jeremyn:
@fuglede Contributors making a new ruleset should already be searching the repository in GitHub first to make sure there isn't already an existing ruleset or pending pull request that does what they want. I admit we don't document this well.

fuglede:
@jeremym: That's my point: They will have no chance to find that existing ruleset if it has been deleted.

jeremyn:
@fuglede I mean searching the repository in GitHub, not just the code. For example, if you search the repository for `pan.baidu.com`, you get [this](https://github.com/EFForg/https-everywhere/search?q=pan.baidu.com&type=Issues&utf8=%E2%9C%93), including a result for pull request #4577 that was a pure deletion.

fuglede:
@jeremyn: Fair enough, I suppose that would catch some of them. I'm not sure people actually do that; at least we have plenty of dupes among issue reports. A slightly more esoteric point would be that this has no effect on those rulesets that have not been produced through GitHub PRs and issues; for example, there are 5134 closed and open PRs, but 21493 rulesets, so there must be a non-trivial mismatch.

renyouguo:
Hi, there are 10+ PRs waiting your check.

renyouguo:
@J0WI @jeremyn @fuglede @Hainish What's your decision?

jeremyn:
If we're voting, I vote to delete it.

fuglede:
@hainish: What I was wondering was more if we had been able to reduce the memory usage of the disabled ones to something below what we require from the active ones, cf. e.g. the remark in https://github.com/EFForg/https-everywhere/issues/1775#issuecomment-105989630 (and the general discussion in that thread).

Hainish:
@fuglede I don't think we have reduced memory usage for inactive/disabled rules.

When a user navigates to a target of a disabled ruleset, we currently present the user with the option to re-enable the ruleset.  In order for this to work, we have to at least have the ruleset list of targets loaded into memory.  One optimization could be that upon clicking to re-enable, all previously extraneous information that is now required (secure cookies, exceptions, the rules themselves) can be retrieved from the disk.

PR #6612, opened 2016-08-26
Remove rules which redirect to a SSL certificate problem

luciancor:
In our data we have more broken certificates which we will send in a different PR if this one goes fine.

cc: @diracdeltas


src/chrome/content/rules/Clear.com.xml
```diff
@@ -1,13 +0,0 @@
-<!--
-	For other Sprint coverage, see Sprint.com.xml.
```
J0WI: Please also remove the reference there.

src/chrome/content/rules/Lolagrove.com.xml
```diff
@@ -29,9 +27,6 @@
 	<securecookie host="^v3\.lolagrove\.com$" name=".+" />
 
 
-	<rule from="^http://lolagrove\.com/"
-		to="https://www.lolagrove.com/" />
-
```
J0WI: You should add a comment about the mismatch on the top, that it get not readded.

src/chrome/content/rules/Pingupod.de.xml
```diff
@@ -3,7 +3,6 @@
 	<!--	Direct rewrites:
 				-->
 	<target host="pingupod.de" />
-	<target host="www.pingupod.de" />
```
J0WI: This should be documented and redirected to non-www.
There is also a mismatch on mgp.pingupod.de and diaspora.pingupod.de can be addded as target to get redirected.

src/chrome/content/rules/Redditmade.com.xml
```diff
@@ -1,17 +0,0 @@
-<!--
-	For other Reddit coverage, see Reddit.xml.
```
J0WI: Please also remove the reference there.

J0WI:
Thanks for these fixes!
Please do not mix changes on completely unrelated files in one PR, this makes the review process easier and we can merge faster.

src/chrome/content/rules/Clear.com.xml
luciancor: @J0WI done

src/chrome/content/rules/Pingupod.de.xml
luciancor: @J0WI  added the documentation but didn't manage to get the redirects pass the tests.
I tried something like:

	<test url="http://www.pingupod.de/" />
	<test url="http://mgp.pingupod.de/" />
	<test url="http://diaspora.pingupod.de/" />

	<rule from="^http://(www|mgp|diaspora)\.pingupod\.de/"
		to="https://pingupod.de/" />

and got:

ERROR src/chrome/content/rules/Pingupod.de.xml: Not enough tests (0 vs 3) for <Rule from '^http://(www|mgp|diaspora)\.pingupod\.de/' to 'https://pingupod.de/'>

am I doing something wrong?

src/chrome/content/rules/Redditmade.com.xml
luciancor: @J0WI done

luciancor:
@J0WI thanks for your suggestions. We will  create parallel  PRs for the following fixes

src/chrome/content/rules/Pingupod.de.xml
J0WI: Use `<target host=" " />` instead of test urls to cover a whole host.
`<rule from="^http://(www|mgp|diaspora)\.pingupod\.de/"
    to="https://pingupod.de/" />` is not necessary because `<rule from="^http:"
 		to="https:" />` redirects all targets.

With this simplified syntax test urls are not required.

luciancor: @J0WI got it - can you please take another look?

J0WI: note that mgp.pingupod.de serves a mismatching cert (see also Travis logs).

luciancor:
@J0WI fixed the last Travis error. Please take another look and let me know if everything is fine now

J0WI:
`mgp.pingupod.de` is now not included at all, but should be documented as mismatching.
Sorry for the Misunderstanding.

luciancor:
@J0WI one more try :)
This time I see a Travis error but its in a file which I did not touch. Please take a look,

J0WI:
Yes, that's okay.

luciancor:
Thanks @J0WI ! We will add start creating the other PRs now. We have around 200 more broken certificates.

J0WI:
@luciancor just out of curiosity: Who is "we" and how did you detect the broken certs?

luciancor:
@J0WI we are a german startup called CLIQZ. We build a secure browser with integrated search. You can read more on [our website](https://cliqz.com/) or [techcrunch](https://techcrunch.com/2016/08/23/mozilla-invests-in-germanys-cliqz-a-private-search-browser/). 
Building the search engine we detected a long list of broken certificates. By intersecting our list with your rules we get around common 200 domains. 
On top of these 200 broken certificates we could contribute with quite an important number of new rules.  Would you be interested in having a discussion about it?

J0WI:
@luciancor if you have a script detecting the broken rules, would it be worth great to share this with us?

luciancor:
@J0WI unfortunately we do not have such a script. We discover all the broken HTTPS certificates as part as our crawling process and I simply made an intersection between that dataset and your rules.

PR #6621, opened 2016-08-27
Create imgmini.eastday.com.xml

renyouguo:


fuglede:
Looks good to me, but could you rename the ruleset and file to `eastday.com`? (Unless there's a good reason for specifying the subdomain.)

renyouguo:
@fuglede No match for other *eastday.com sites.

fuglede:
Still, then it would be better to add a comment in `Eastday.com.xml` with notes on the errors that appear in the other eastday.com subdomains. This makes it a bit easier to find the right ruleset file if changes need to be imposed later on.

fuglede:
Thanks. 👍

PR #6636, opened 2016-08-28
[Zdrojak.cz] Remove HSTS preloaded ruleset

Roman3349:
https://hstspreload.appspot.com/?domain=zdrojak.cz
Signed-off-by: Roman3349 <ondracek.roman@centrum.cz>

### DIFF
--- a/src/chrome/content/rules/Zdrojak.cz.xml
+++ b/src/chrome/content/rules/Zdrojak.cz.xml
@@ -1,7 +1,0 @@
-<ruleset name="Zdroják.cz">
-    <target host="zdrojak.cz" />
-    <target host="www.zdrojak.cz" />
-
-    <rule from="^http:"
-            to="https:" />
-</ruleset>


fuglede:
The site is not yet preloaded in Firefox ESR according to https://www.ssllabs.com/ssltest/analyze.html?d=zdrojak.cz&s=2001%3a1ab0%3af001%3aface%3a0%3a0%3a0%3a11&latest, so we should keep it around a bit longer.

Roman3349:
On site https://dev.ssllabs.com/ssltest/analyze.html?d=zdrojak.cz&s=2001%3A1ab0%3Af001%3Aface%3A0%3A0%3A0%3A11 is written that this page is preloaded in all browsers.

fuglede:
That's odd. Do you know what the difference between `dev` and `www` is, and what could cause the discrepancy (and which one to trust)? cc @ivanr.

ivanr:
I think it may be related to caching, as preload lists are downloaded periodically. The main site now also shows the host preloaded. IIRC, SSL Labs will check the "master" branches (I originally configured these URLs: https://github.com/ssllabs/research/wiki/Preload-Lists) so it will take some further time for the preloaded sites to appear in mainstream browser versions.

fuglede:
I see; thanks for popping by, @ivanr. I just reran it on [the main site](https://www.ssllabs.com/ssltest/analyze.html?d=zdrojak.cz&s=2001%3a1ab0%3af001%3aface%3a0%3a0%3a0%3a11&latest), and it still displays it as non-preloaded; the host also not appears in [the list from the wiki page](https://gitweb.torproject.org/tor-browser.git/plain/security/manager/boot/src/nsSTSPreloadList.inc?h=tor-browser-38.2.1esr-5.0-2).

ivanr:
IIRC, Tor doesn't have a master branch, so SSL Labs follows one particular version, which means that updates are not picked up. The URL can only be updated in new SSL Labs releases. I think SSL Labs should focus only on what's in Chromium because, once there, a hostname will eventually propagate everywhere. But there's definitely no easy way to determine when a hostname goes preloaded mainstream.

fuglede:
That's good to know; thanks again. For our purposes this is of course a bit unfortunate (but not the end of the world).

J0WI:
Tor's preload list is based on (or even equal to) Firefox.
I [verified](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.0-1#n12751) that this host is included, so lgtm.

PR #6646, opened 2016-08-29
psiphon.ca

galeksandrp:


src/chrome/content/rules/psiphon.ca.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="psiphon.ca">
+	<target host="psiphon.ca" />
+	<target host="www.psiphon.ca" />
```
J0WI: blog.psiphon.ca

galeksandrp: Added.

J0WI:
Thanks!

PR #6648, opened 2016-08-29
gp-digital.org

galeksandrp:


src/chrome/content/rules/gp-digital.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="gp-digital.org">
+	<target host="gp-digital.org" />
+	<target host="www.gp-digital.org" />
```
J0WI: comment.gp-digital.org

galeksandrp:
Added.

PR #6661, opened 2016-08-29
smarthead.ru

galeksandrp:


J0WI:
Could you please also add `client.smarthead.ru` and a comment about the mismatch on `support.smarthead.ru`?

galeksandrp:
Added.

PR #6664, opened 2016-08-29
ony.ru

galeksandrp:


src/chrome/content/rules/ony.ru.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+www.ony.ru ¹
```
J0WI: look.ony.ru ¹
www.ony.ru ¹
mdg.ony.ru ⁴

galeksandrp: Added.

PR #6670, opened 2016-08-29
media5.com

galeksandrp:


src/chrome/content/rules/media5.com.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+www.more.media5.com ²
```
J0WI: mismatch on m5.media5.com

galeksandrp:
Added.

PR #6676, opened 2016-08-29
mosaicmedia.ru

galeksandrp:


src/chrome/content/rules/mosaicmedia.ru.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+inshopper.mosaicmedia.ru ¹
+smmdesk.mosaicmedia.ru ²
+
+
+¹ mismatch
```
J0WI: also on en.mosaicmedia.ru

PR #6678, opened 2016-08-30
[Fri-Gate.org] Remove duplicate ruleset

synthtech:
Filename conflicts with 9c6f0fa.


Hainish:
This wasn't caught by tests due to an issue fixed in https://github.com/EFForg/https-everywhere/pull/6679

PR #6682, opened 2016-08-30
[University at Buffalo] Disable several subdomains

fuglede:
Ran a quick sweep with the checker; these are the errors it caught. This closes https://trac.torproject.org/projects/tor/ticket/16345.

### DIFF
--- a/src/chrome/content/rules/University_at_Buffalo.xml
+++ b/src/chrome/content/rules/University_at_Buffalo.xml
@@ -13,6 +13,7 @@
 		- (www.)?admissions ¹
 		- (www.)?academy ²
 		- advising ²
+		- www.alumni ⁵
 		- (www.)?apb ³
 		- www.cambi ¹
 		- www.chem ¹
@@ -29,7 +30,8 @@
 		- gsa ⁵
 		- (www.)?health ¹
 		- helpdesk ¹
-		- (www.)?giving  ⁵
+		- (www.)?hr ⁵
+		- (www.)?giving ⁵
 		- www.grad ¹
 		- www.ise ⁶
 		- (www.)?itpolicies ¹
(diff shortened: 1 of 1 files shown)


jeremyn:
Being extremely petty here, but the site redirects http://www.alumni.buffalo.edu to http://www.buffalo.edu/alumni , not to http://alumni.buffalo.edu , so strictly speaking I don't think the rewrite rule shouldn't go to http://alumni.buffalo.edu . But if you want to keep it to https://alumni.buffalo.edu that's fine, just let me know.

What's the [checker](https://github.com/EFForg/https-everywhere/tree/master/test/rules/src/https_everywhere_checker) for, exactly? I looked around a little but didn't see any documentation.

fuglede:
I did actually consider what you suggest to save the extra redirect but decided against it: If one day they change their URLs, the redirect might be updated, which could lead to unexpected behaviour from the rule; for this reason, we'll generally try to avoid following redirects, if it is possible without sacrificing security. That being said, it's not something I have very strong opinions on (in this case at least).

The checker library is what `fetch-test.sh` uses, so that's where you'll typically go hunting if the Travis fetch test gives buggy output. Besides the obviously useful Travis test, you can also use it to run ruleset database-wide runs to find all currently buggy rulesets (and AFAIR, that's what it was introduced for in the first place).

jeremyn:
The site does a lot of redirect hops, from http://www.alumni.buffalo.edu to http://www.buffalo.edu/alumni to http://www.buffalo.edu/content/www/alumni/ and finally to http://www.buffalo.edu/alumni.html . So http://www.buffalo.edu/alumni might not be that stable since it's only the first hop. On the other hand the problem with your rewrite is that the site doesn't follow that path at all, you're just making it up based on the URLs http://www.alumni.buffalo.edu and http://alumni.buffalo.edu being very similar. So neither choice is great. Let's just keep what you have. Merged.

Thanks for clarifying about the checker. I usually refer to `fetch-test.sh` by name. Running it against all rulesets would take a really long time.

PR #6683, opened 2016-08-30
Fix BusinessInsider.xml video

jeremyn:
This should fix issue #5014 regarding broken video on businessinsider.com (example [here](http://www.businessinsider.com/how-plucking-nose-hairs-can-cause-lethal-infections-2016-6)). Commits 9ea45fbf452bbc49e6f97810e2b2631c9f353ff0 and 1e0802e635d9038d51320bd5b63d8b5725a0c48f fix these errors I get with `fetch-test.sh`:

```
ERROR rules/BusinessInsider.xml: Fetch error: http://www.businessinsider.com/account => https://www.businessinsider.com/account: Too many redirects while fetching 'https://www.businessinsider.com/account'
ERROR rules/BusinessInsider.xml: Fetch error: http://www.businessinsider.com/register => https://www.businessinsider.com/register: Too many redirects while fetching 'https://www.businessinsider.com/register'
ERROR rules/BusinessInsider.xml: Fetch error: http://jobs.businessinsider.com/c/ => https://businessinsider.jobamatic.com/c/: (7, 'Failed to connect to businessinsider.jobamatic.com port 443: Connection refused')
ERROR rules/BusinessInsider.xml: Fetch error: http://insider.businessinsider.com/ => https://insider.businessinsider.com/: (6, 'Could not resolve host: insider.businessinsider.com')
ERROR rules/BusinessInsider.xml: Fetch error: http://oascentral.businessinsider.com/ => https://oascentral.businessinsider.com/: (6, 'Could not resolve host: oascentral.businessinsider.com')
```

Issue #5601, opened 2016-07-03
Travis-CI testing problem if a test URL has a domain name with no trailing slash


galenide:
From what it appears, Travis-CI testing may fail if a test URL does not have a trailing slash after the domain (i.e., `<test url="http://example.com/" />` would work but `<test url="http://example.com" />` would fail.) At the current time, the  [ruleset style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) shows test URLs with trailing slashes but does not mention any requirement as far as trailing slashes in test URLs.

When recently submitting changes to a ruleset, the [testing output](https://travis-ci.org/EFForg/https-everywhere/jobs/141954088) included the following:

```
Ruleset database has changed. Testing test URLs in all changed rulesets.

INFO Finished comparing http://www.spiderwebsoftware.com -> https://www.spiderwebsoftware.com. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://spidweb.com/ -> https://spiderwebsoftware.com/. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://www.spidweb.com -> https://www.spidweb.com. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://spiderwebsoftware.com/ -> https://spiderwebsoftware.com/. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://www.spiderwebsoftware.com/ -> https://spiderwebsoftware.com/. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://spidweb.com/ -> https://spiderwebsoftware.com/. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

INFO Finished comparing http://www.spidweb.com/ -> https://spiderwebsoftware.com/. Rulefile: src/chrome/content/rules/Spiderweb-Software.xml.

ERROR src/chrome/content/rules/Spiderweb-Software.xml: Fetch error: http://www.spiderwebsoftware.com => https://www.spiderwebsoftware.com: (51, "SSL: certificate subject name (spiderwebsoftware.com) does not match target host name 'www.spiderwebsoftware.com'")

[...]

PR #6684, opened 2016-08-30
[Lenovo] Add exclusion

fuglede:
This closes https://trac.torproject.org/projects/tor/ticket/18055


jeremyn:
What's the purpose of `utils/ruleset-coverage-whitelist.txt`? I don't know what that file is for or when it should be changed.

fuglede:
@jeremyn: When the ruleset test requirements were introduced, the requirements would not be satisfied for a large part of the existing ruleset collection. Thus, a whitelist was introduced to allow the test to run. Each entry consists of the name of the whitelisted ruleset together with its `sha256sum`, and if a ruleset is included, it will not be included in the test. As we continue to update the database, the length of the whitelist will continue to decrease, but in this case, I'll not have the time to produce the thorough update required to remove it from the whitelist, so I simply updated the hash. You'll probably see that happen every now and then.

Now, the obvious question is why the test ran anyway ...

jeremyn:
Oh, I see. I just had a hassle with test failures in pull request #6683, though not as many as you have here. If I had added `BusinessInsider.xml` to the whitelist, I could have avoided the extra rewrites. Of course that would not have been good to do especially since `BusinessInsider.xml` isn't in the whitelist already.

jeremyn:
By the way, I looked at this and couldn't figure it out. My vague idea is that something is wrong in Travis.

fuglede:
I found the explanation; see #6689.

jeremyn:
Why did you remove the trailing slash from 

```xml
<exclusion pattern="^http://www\.lenovo\.com/(?:lenovorecovery|support|training)" />
```

and the related tests? Also, you can remove the `?:` for simplicity.

fuglede:
Fixed the non-capturing groups. I removed the slash to ensure that the pattern would match both http://www.lenovo.com/lenovorecovery and http://www.lenovo.com/lenovorecovery/ (and similarly for the others; note that https://www.lenovo.com/training is currently broken).

jeremyn:
Thanks for updating the non-capturing groups.

I see what you mean about the trailing slash. In #5601 we're discussing some ambiguity about trailing slashes. It'd be great if you could contribute to that discussion.

Issue #5266, opened 2016-06-22
Please add 'vip.jr.jd.com'


LisonFan:
Thanks

J0WI:
Thanks for your suggestion.

We are always happy to receive pull requests ;)
https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md
https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md

fuglede:
Closed by #6687.

PR #6687, opened 2016-08-31
Add vip.jr.jd.com to JD.com.xml (issue #5266)

jeremyn:
This should resolve issue #5266.

### DIFF
--- a/src/chrome/content/rules/JD.com.xml
+++ b/src/chrome/content/rules/JD.com.xml
@@ -258,6 +258,7 @@
 	<target host="vc.jd.com" />
 	<target host="vcp.jd.com" />
 	<target host="vip.jd.com" />
+	<target host="vip.jr.jd.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/JD.com.xml
```diff
@@ -258,6 +258,7 @@
 	<target host="vc.jd.com" />
 	<target host="vcp.jd.com" />
 	<target host="vip.jd.com" />
+	<target host="vip.jr.jd.com" />
```
fuglede: Just a minor comment: Could you move this to [L202](https://github.com/jeremyn/https-everywhere/blob/fab6402ccd50b3735e8475eb0e60663dd4b09bf7/src/chrome/content/rules/JD.com.xml#L201) instead (so that the ruleset is ordered alphabetically after subdomain order)?

jeremyn:
Reordered.

Issue #6569, opened 2016-08-23
Test CA repository does not reflect Firefox state


fuglede:
PR #6102 gives an example of a host which whose SSL configuration will cause breakage in Firefox (due to an invalid chain) but which passes the fetch test.

cc @Hainish.

@jeremyn, @j0wi: You may want to be aware of this when reviewing; I just came across a second false positive.

J0WI:
When you visited a site with the same cert chain before, the browser caches the chain and doesn't show any errors.
As far as I know we integrated some cert chains in our test infrastructure that tests don't fail. We may reconsider this.

J0WI:
They are located in "test/rules/platform_certs/", but they might only be used on rules with platform CAcert or incomplete (do we have a platform for that?) specified.

jeremyn:
This is a useful link for anyone trying to understand this problem: https://wiki.mozilla.org/CA:FAQ .

fuglede:
@j0wi: The ones in `test/rules/platform_certs/default` are exactly the ones that `fetch-test.sh` and friends use, so those are the ones that would appear to be out of sync.

jeremyn:
Considering @J0WI 's "We may reconsider this.":

My rule of thumb has been that a rewrite works if the https and http version of a page do the same thing: return the same content, have the same 4xx error, whatever. If a site is misconfigured to not correctly fetch an intermediate certificate, the two versions don't do the same thing and the rule is bad. So by this rule of thumb, I don't think we should include prefetched intermediate certificates in our tests.

Hainish:
Looking into this now

Hainish:
I wonder how this list of intermediates was generated in the first place.

The [Mozilla CA Certificate Store](https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/) lists this [source file with all of the included root certificates](https://hg.mozilla.org/mozilla-central/raw-file/tip/security/nss/lib/ckfw/builtins/certdata.txt).  We could possibly parse this file and generate a new `test/rules/platform_certs/defaults` that is more current.

@jsha, do you know how this was generated in the first place?

fuglede:
Pinging @hiviah as he [added them in the first place](https://github.com/hiviah/https-everywhere-checker/commit/53e2048dfc27309e2085a1120d122581235465a4#diff-8990ddd6ec42bfc8d7da8b90be9c1d9b). Having an automated fetch-and-parser sounds like it would be great.

jsha:
[...]

PR #6697, opened 2016-09-01
add dogechain.info

J0WI:


fuglede:
Site serves a broken certificate chain.

    $ curl -I https://dogechain.info

Probably another false test positive due to #6569.

J0WI:
I just realized, that we may rather define a platform for this problematic than just disabling all this rules.

fuglede:
Not a bad idea. There's quite a few of them by now.

    https-everywhere/rules$ grep 'chain' * | wc -l
    767

jeremyn:
Note: this is one of the URLS requested in issue #3069.

fuglede:
For the record, I wrote them about the missing intermediate.

jeremyn:
- [ ] I don't get the incomplete certificate chain problem, so you can remove `default_off`.
- [ ] Note https://ws.dogechain.info as `Secure connection failed`.

J0WI:
There is no HTTP connection to ws.dogechain.info possible at all, so I think it's not worth to mention.

jeremyn:
That's fine. Merged. I've already removed the intermediate commit messages.

PR #6695, opened 2016-09-01
add paranoidsecurity.nl

J0WI:
see #3069

PR #6707, opened 2016-09-01
update torrents.to to torrents.me

J0WI:


jeremyn:
- [ ] Add a general `<securecookie host=".+" name=".+" />` tag

J0WI:
I don't think that we should require securecookie rules.
In this case, they are secured server side anyway and we rewrite all requests to https.

jeremyn:
I don't feel strongly about it here, so that's fine. Merged.

Can you think of any situation where a rule reviewer should ask a rule writer to add a `securecookie` tag, though? If this pull request is a bad example, then I can't think of a good one.

jeremyn:
For example, what's the difference between this pull request, and pull request #6695?

J0WI:
In #6695 the server send insecure cookies (however, I think you can't exploit it there).

I think the only scenario where the securecookie flag is really important is a site, that can't be fully secured and uses cookies in a sensible way (e.g. login).

jeremyn:
By "insecure cookies" do you mean it sends cookies with `isSecure` `false`? In this `torrents.me` pull request though, the `_ga` cookie is `isSecure` `false`, and adding a `securecookie` changes that to `true`.

J0WI:
This sounds like a tracking cookie, but I can't see this here.

PR #6713, opened 2016-09-01
reenable usrportage.de.xml

J0WI:


jeremyn:
- [ ] The only blocked mixed content is a GitHub follow button, so I don't think this needs to be `platform="mixedcontent"`,

J0WI:
platform removed.

jeremyn:
What does `secured by us.` mean in the comment?

J0WI:
We have a `Ghbtns.com.xml` rule, so even if the user allows mixed content it will be secured.

jeremyn:
Okay, thanks. Merged.

PR #5265, opened 2016-06-22
fnal.gov

galeksandrp:

PR #6722, opened 2016-09-01
Update Udistrital.xml

J0WI:
reenabled
added more hosts

### DIFF
--- a/src/chrome/content/rules/Udistrital.xml
+++ b/src/chrome/content/rules/Udistrital.xml
@@ -44,6 +44,7 @@
 	<target host="www.udistrital.edu.co" />
 	<target host="ceri.udistrital.edu.co" />
 	<target host="condor.udistrital.edu.co" />
+	<target host="fciencias.udistrital.edu.co" />
 	<target host="oas.udistrital.edu.co" />
 	<target host="portalws.udistrital.edu.co" />
 	<target host="sgral.udistrital.edu.co" />
--- a/src/chrome/content/rules/Udistrital.xml
+++ b/src/chrome/content/rules/Udistrital.xml
@@ -1,7 +1,11 @@
 <!--
 	404 on deep links:
 		- autoevaluacion.udistrital.edu.co
+		- comunidad.udistrital.edu.co
+		- gemini.udistrital.edu.co
 		- ilud.udistrital.edu.co
+		- laud.udistrital.edu.co
+		- mail.udistrital.edu.co
 
 	HTTP response (serives run on port 8080):
 		- espambienteydesarrollo.udistrital.edu.co
@@ -11,6 +15,7 @@
 	Refused:
 		- die.udistrital.edu.co
 		- doctoradoingenieria.udistrital.edu.co
+		- ingenieria1.udistrital.edu.co
 
 	Timeout:
 		- biblioteca.udistrital.edu.co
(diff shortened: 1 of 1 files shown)


jeremyn:
These behave differently between HTTP and HTTPS. Is that what `HTTP response` means?
- [ ] https://autoevaluacion.udistrital.edu.co (move from `404`)
- [ ] https://ilud.udistrital.edu.co (move from `404`)
- [ ] https://espambienteydesarrollo.udistrital.edu.co
- [ ] https://fasab.udistrital.edu.co
- [ ] https://maestriabosques.udistrital.edu.co

This is marked `Mixed content` but the differences seem minor enough that I think it should be a target:
- [ ] https://rita.udistrital.edu.co

These aren't `Invalid cert` for me, but instead have valid HTTPS and redirect to domains that don't exist:
- [ ] https://condor.udistrital.edu.co
- [ ] https://oas.udistrital.edu.co

Also:
- [ ] there are several other subdomains that aren't included, see [here](https://www.google.com/search?q=site%3A*.udistrital.edu.co+-site%3Aautoevaluacion.udistrital.edu.co+-site%3Ailud.udistrital.edu.co+-site%3Aespambienteydesarrollo.udistrital.edu.co+-site%3Afasab.udistrital.edu.co+-site%3Amaestriabosques.udistrital.edu.co++-site%3Adie.udistrital.edu.co+-site%3Adoctoradoingenieria.udistrital.edu.co++-site%3Abiblioteca.udistrital.edu.co+-site%3Acelsius.udistrital.edu.co+-site%3Acidc.udistrital.edu.co+-site%3Arevistas.udistrital.edu.co+-site%3Arita.udistrital.edu.co+-site%3Acondor.udistrital.edu.co+-site%3Aherbario.udistrital.edu.co+-site%3Aidexud.udistrital.edu.co+-site%3Aoas.udistrital.edu.co+-site%3Awww.udistrital.edu.co+-site%3Aportalws.udistrital.edu.co+-site%3Asgral.udistrital.edu.co+&ie=utf-8&oe=utf-8)

J0WI:
I clarified the comments a bit, but I decided to not trying to include all subdomains. I tried this in the past on university sites, but they tend to be much more complex then normal hosts, where full coverage is a reachable target.
Also the Google trick will not work for so much hosts, because the length of the search query is limited.
But let me know what you think about.

For the mixed content it seems to break the language menu, which is an important functionality.

jeremyn:
For https://rita.udistrital.edu.co, I only see Spanish and English language flags with and without mixed content blocking. Is it different for you?

I added some more subdomains and an updated search query that goes to the maximum length Google allows. Most of them don't work with HTTPS anyway, and since we're labeling the ruleset `partial`, I'm okay with leaving the rest off if you want. But if you actually click out in the updated search query, Google only finds about 300 significantly distinct results, not the hundreds of thousands it reports at first. So it wouldn't be that bad to go through them. I went through that [here](https://github.com/EFForg/https-everywhere/pull/5265#issuecomment-242941309) in another pull request.

J0WI:
The flags are present, but they don't work because the scripts from Google Translator get blocked.

jeremyn:
I see what you mean about https://rita.udistrital.edu.co , that's fine.

Since we might be leaving some off, please confirm when you're done (though please add https://fciencias.udistrital.edu.co since it's in the checklist and the search query).

J0WI:
added

jeremyn:
Are you done adding stuff?

J0WI:
I think so. Primary reason for this PR was to get rid of all `platform="firefox"`.

jeremyn:
I was surprised to see a `platform` other than `mixedcontent`. Somewhere else I saw one for `cacert`, which I was even less sure about.

Is `mixedcontent` the only platform we should use for new rules?

J0WI:
cacert and mixedcontent are both okay, we might have also s platform for incomplete cert chains in future (https://github.com/EFForg/https-everywhere/pull/6697#issuecomment-244210302).

jeremyn:
When is `cacert` enabled? I see it's allowed [here](https://github.com/EFForg/https-everywhere/blob/f45195d18399bbda5c5e0c850098d59be281720c/utils/relaxng.xml#L15) but I'm not sure where it's used ([here](https://github.com/EFForg/https-everywhere/blob/5f907648a5e51d169b358fd5f91a88cad03c59ed/chromium/rules.js#L187-L188), maybe?).

PR #6729, opened 2016-09-02
glx-dock.org

galeksandrp:


src/chrome/content/rules/glx-dock.org.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+
```
J0WI: themes.glx-dock.org (missmaching, expired, self-signed)

galeksandrp:
Added.

PR #6730, opened 2016-09-02
vhffs.org

galeksandrp:


src/chrome/content/rules/vhffs.org.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+
```
J0WI: projects.vhffs.org panel.vhffs.org both time out.

galeksandrp:
Added.

Issue #5874, opened 2016-07-30
Extension is broken/corrupted and won't repair on Chrome


faried:
I see the problem both on Mac OS 10.11.6 with Chrome 54.0.2810.2 and on Ubuntu 14.04.  The fact that it "broke" and I didn't notice it immediately seems like comp.risks fodder.

<img width="746" alt="screen shot 2016-07-30 at 10 36 23" src="https://cloud.githubusercontent.com/assets/13652/17272678/1973fe70-56b6-11e6-8bfd-753b16c62d6f.png">

auyongcheemeng:
Ditto on the the problem on Chrome 54.0.2810.2 (64 bit) on Win10.

Hainish:
I can reproduce this, looks like it's fairly platform-independent, happening with Chrome 54 across the board. @semenko any idea what may be going on here?

semenko:
Hm, IIRC that failure results when a local checksum/signature of the extension contents don't match the signature of what was submitted to the Chrome app store. (The intent is to prevent malware from modifying extension scripts.)

I've yet to see this personally -- and I don't have a great idea of what could be causing it.

If this weren't platform independent, I'd guess some client-side antivirus was rewriting / removing some of our package data [e.g. for a "suspicious" url].

Hainish:
Thanks @semenko.  Looks like this is happening with our Privacy Badger extension on Chrome 54 as well.  This may be some extra contents that are packaged but not checked on either the client-side or Chrome Web Store side.  I'll have to consider how to debug this, though - seems tricky, and I don't see any debugging output.  I wonder if we have contacts at CWS that can help?

semenko:
Very odd -- maybe there's a strange character encoding / odd file making it's way in, and that's hitting an edge case in Chrome?

I don't see this in 54.0.2816.0 (Ubuntu). And I don't see any similar bugs at https://crbug.com/

Hainish:
cc @cooperq - is there a similar issue for Privacy Badger you can link to?

faried:
I'm not sure what caused it, but it seems fixed now.  Maybe a transient issue with Chrome dev?  I'm on 54.0.2816.4 on OS X and 54.0.2816.0 on Linux now.

I ran into the Privacy Badger issue on Linux at the same time, but not on OS X.

Hainish:
@faried I'm still seeing it on Ubuntu w/ Chrome 54.0.2816.0 (64-bit)

[...]

PR #6731, opened 2016-09-02
getbootstrap.com

galeksandrp:
Fix https://trac.torproject.org/projects/tor/ticket/19396

### DIFF
--- a/src/chrome/content/rules/Getbootstrap.com.xml
+++ b/src/chrome/content/rules/Getbootstrap.com.xml
@@ -8,6 +8,11 @@
     <target host="blog.getbootstrap.com" />
     <target host="themes.getbootstrap.com" />
     <target host="v4-alpha.getbootstrap.com" />
+    
+    <!-- Fix https://trac.torproject.org/projects/tor/ticket/19396 -->
+    <exclusion pattern="^http://themes\.getbootstrap\.com/products/dashboard" />
+    
+    <test url="http://themes.getbootstrap.com/products/dashboard" />
 
     <securecookie host="^(www\.|expo\.|blog\.|themes\.|v4-alpha\.)?getbootstrap\.com$" name=".+" />
 
--- a/src/chrome/content/rules/Getbootstrap.com.xml
+++ b/src/chrome/content/rules/Getbootstrap.com.xml
@@ -14,7 +14,7 @@
     
     <test url="http://themes.getbootstrap.com/products/dashboard" />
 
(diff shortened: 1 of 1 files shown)


J0WI:
Please also leave a comment in the rule about the issue there.

galeksandrp:
Added.

fuglede:
The site appears to work fine for me, and in particular I did not notice any redirect loops. Are there any specific steps to reproduce?

jeremyn:
@fuglede The torproject bug labels it specifically for Chrome. I can't reproduce it on Firefox, and I can't test it on Chrome because of the #5874 .

jeremyn:
Now that there's a fix/workaround for #5874 , I can confirm that https://themes.getbootstrap.com/products/dashboard loops in Chrome but not Firefox.

jeremyn:
- [ ] Update the `securecookie` [here](https://github.com/galeksandrp/https-everywhere/blob/c2f1dc875be3b0cfb8a67dd4d6996729fb6893c0/src/chrome/content/rules/Getbootstrap.com.xml#L17) to remove `themes\.|`. Otherwise what can happen is that a secure cookie is set at, for example, https://themes.getbootstrap.com but then can't be read at http://themes.getbootstrap.com/products/dashboard .

galeksandrp:
Updated.

J0WI:
I wonder why this works in Firefox, because of the
```
  <!-- sad hack because github pages doesn't support https -->
  <script>
     if (window.location.protocol == "https:") window.location.protocol = "http:"
  </script>
```
Console shows
```
NS_ERROR_UNEXPECTED:   dashboard:6
	<anonymous> https://themes.getbootstrap.com/products/dashboard:6:48
```

jeremyn:
@J0WI I didn't dig into too much but it looks like this is a thing with Firefox, see [here](https://stackoverflow.com/questions/4723213/detect-http-or-https-then-force-https-in-javascript/10036029#10036029) and [here](https://stackoverflow.com/questions/34218769/set-window-location-protocol-in-different-browsers) for example.

fuglede:
I contacted the webmasters regarding the issue.

fuglede:
And here's their response:

> We appreciate you letting us know – it's an unfortunate necessity. Basically the way Shopify works it required us to force just the product pages to be insecure to iFrame our content in without issue
in some situations. It wasn't our favorite either, but was the only surefire way to solve the issue without basically getting Shopify to accept a feature request. That said, every other page is
secure, so we figured we could deal with the inconvenience.

PR #6733, opened 2016-09-02
aldeparty.eu

galeksandrp:


fuglede:
I came across mixed content issues on https://www.aldeparty.eu/en/about/history and https://www.aldeparty.eu/en/media/publications.

galeksandrp:
Ohh, that's not only fonts, added `platform="mixedcontent"`

PR #6736, opened 2016-09-02
Add ghbtns.com

theel0ja:


fuglede:
From Travis:

    failure: /opt/utils/../src/chrome/content/rules/ghbtns.com.xml failed test: The 'from' rule contains unescaped period in regular expression.  Try escaping it with a backslash.

On a separate note, you could combine your two rules into one, using

```xml
<rule from="^http://(www\.)?ghbtns\.com/" to="https://ghbtns.com/" />
```

fuglede:
In fact, now that I look at the site, `<rule from="^http:" to="https:" />` would also be just fine.

fuglede:
Looks like we have a new error now:

    failure: Hostname ghbtns.com with platform '' shows up in 2 different rulesets.

So, it looks like the hosts are already covered in `Ghbtns.com.xml` (note the uppercase G). So, thanks of course for the ruleset, but I'll go ahead and close this one, as we already have it covered.

Issue #4227, opened 2016-02-26
Breaks some pages of WDR (Westdeutscher Rundfunk) 


Septa9:
Firefox 43.0.4
Https Everywhere 5.1.4

Filter:
Westdeutscher Rundfunk

Some pages, i.e. https://www.wdr.de/programmvorschau/, are using resources from http://www1.wdr.de/resources/, which FF refuses to load with "Blocked loading mixed active content"

funkydude:
The problem is that www1.* doesn't support HTTPS. There's 2 solutions to this.

1) Gather a list of all the broken pages and put them in the ruleset as exclusions

2) Disable the rule for wdr.de

I would lean towards the latter.

Septa9:
Thanks for your quick reply.

>2) Disable the rule for wdr.de
>I would lean towards the latter.

Yeah, that's what I'm currently doing. Lazy as I am, I just was hoping that someone else might already have collected an exception list.

numismatika:
www1 is using a valid cert but that server redirects to plain.

-  blog.wdr.de
- containerids-id1.wdr.de
- containerids-id2.wdr.de
- containerids-id3.wdr.de
- containerids-id4.wdr.de
- containerids.wdr.de
- deviceids-meda-id1.wdr.de
- deviceids-meda-id2.wdr.de
- deviceids-meda-id3.wdr.de
- deviceids-meda-id4.wdr.de
- deviceids-meda.wdr.de
- deviceids-medp-id1.wdr.de
- deviceids-medp-id2.wdr.de
- deviceids-medp-id3.wdr.de
- deviceids-medp-id4.wdr.de
- deviceids-medp.wdr.de
- deviceids-meds-id1.wdr.de
- deviceids-meds-id2.wdr.de
- deviceids-meds-id3.wdr.de
- deviceids-meds-id4.wdr.de
- deviceids-meds.wdr.de
- deviceids-medt-id1.wdr.de
- deviceids-medt-id2.wdr.de
- deviceids-medt-id3.wdr.de
- deviceids-medt-id4.wdr.de
- deviceids-medt.wdr.de
- exporte.wdr.de
- freundeskreis.einslive.de
- static.freundeskreis.einslive.de
- wdr.de
- www.einsfestival.de
- www.einslive.de
- www.einslivekrone.de
- www.funkhauseuropa.de
- www.lindenstrasse.de
- www.planet-wissen.de
- www.sportschau.de
- www.wdr.de
[...]

Issue #2577, opened 2015-08-16
Videos don't play on sportschau.de


KommX:
The rule "Westdeutscher Rundfunk" includes the domain sportschau.de

Unfortunately the embedded videos don't play if the site is redirected to https.
Test case:
https://www.sportschau.de/doping/geheimsache-doping-102.html

numismatika:
They play for me, there is just a small note for mixed display content from
http://adaptiv.wdr.de
display content isn't blocked by default, just active.
Did you change that setting?

KommX:
@numismatika 
I can reproduce the behavior you described with Chrome Version 44.0.2403.155 (64-bit) and HTTPS Everywhere 2015.8.13.
But with the dev Version 46.0.2478.0 dev (64-bit) I have the problem described in my initial post. Perhaps  Chrome changed something with respect to mixed content?!

KommX:
The same (or at least related) issue exists for this site:
https://tv.sport1.de/player/player.php?id=s225420

It's fine with the stable version of Chrome and HTTPS Everywhere.
But with the dev version of Chrome the videos don't play.

Perhaps it's not yet an httpsE issue but maybe it will be in the future?!

numismatika:
It might be that chrome dev is stepping up security and blocking passive mixed content now as well.
One would need to check for related possible changes there.
Right now it can't be fixed with a https-everywhere rule , because Akamai doesn't serve certificates that are valid for event02-lh.streaming.sport1.de

KommX:
Now I have a test case with the current stable version of Chrome (Version 45.0.2454.99 (64-bit)):
1. Go to https://www.sportschau.de/fussball/bundesliga/videolewandowskieinfachtorgeschossen100.html
2. Try to get the video to play.

numismatika:
``
Blocked loading mixed active content "http://www1.sportschau.de/sportschau_specials/mm_ticker/css/radiolivestream.css"

Blocked loading mixed active content "http://containerids.wdr.de/container.js?761708" ``

Clear case of https://bugzilla.mozilla.org/show_bug.cgi?id=878890

You could temporarily override the active content blocking for the site in firefox or less secure:
security.mixed_content.block_active_content = false in about:config.

KommX:
The bug you mention is a Firefox bug. Does the same bug exist in Chrome?

numismatika:
I think this was https://code.google.com/p/chromium/issues/detail?id=122548

KommX:
[...]

PR #6738, opened 2016-09-02
WDR: add wdr4 and www1 subdomains

numismatika:
rest of the hosts from the SAN of the main certificate are omitted for now
There are small mixed content warnings but as far as i can tell the site doesn't break with the missing css.
Videos seem to work fine too.

### DIFF
--- a/src/chrome/content/rules/Wdr.de.xml
+++ b/src/chrome/content/rules/Wdr.de.xml
@@ -1,22 +1,80 @@
 <!--
+	No valid Cert for:
 
-	Problematic domains:
+		- adaptiv.wdr.de
+		- podcast.wdr.de
+		- print.wdr.de
+		- reportage.wdr.de
+		- www.einslivekrone.de
+		- www.kiraka.wdr.de
+		- www.kirche.wdr.de
+		- www.lindenstrasse.de
+		- www.wdrmaus.de
+		- ^wdr.de
 
-		- www1 ¹
+	certificatechain broken on:
+		- webmail.wdr.de
+		- freundeskreis.einslive.de
 
-	¹ Redirect to HTTP
+	Non-resolvable host:
+		- static.freundeskreis.einslive.de
+
+	Redirect to plain http on:
+		- digit.wdr.de
+		- www.einsfestival.de
+		- www.planet-wissen.de
+
+	Time-out on;
+		- klangkiste.wdr.de
+		- mobiltext.wdr.de
+		- neuneinhalb.wdr.de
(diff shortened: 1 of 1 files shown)


J0WI:
Could you also search for subdomains?

numismatika:
There are a lot of domains in the cert for the mainpage, will take a while to see if they are even used on the web.

J0WI:
You can [use Google](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) to search for subdomains, even those that have no or a different cert.

J0WI:
Does this also address #2577 and #4227?

src/chrome/content/rules/Wdr.de.xml
```diff
@@ -1,22 +1,79 @@
...
+		- ^wdr.de
 
-		- www1 ¹
+	certificatechain broken on:
+		- webmail.wdr.de
+		- freundeskreis.einslive.de
 
-	¹ Redirect to HTTP
+	Non-resolvable host:
+		- static.freundeskreis.einslive.de
```
J0WI: If a host is not resolvable (=> does not exist) you don't need to list it.

src/chrome/content/rules/Wdr.de.xml
```diff
@@ -1,22 +1,79 @@
...
+
+	Redirect to plain on:
```
J0WI: `plain` is not accurate, please use 'http' or 'plain http'.

src/chrome/content/rules/Wdr.de.xml
numismatika: It is part of the SAN part of the certificate for www.wdr.de so for the sake of completeness i noted that

src/chrome/content/rules/Wdr.de.xml
numismatika: will be changed

numismatika:
#2577 seems fine, videos on the sportschau page play. The live-stream placeholder loop is fine too.
Haven't found a real livestream yet for testing.
#4227 still is an issue. CSS&JS are blocked as mixed-content and FF ignores our rewrite (open FF bug)

J0WI:
>  #4227 still is an issue. CSS&JS are blocked as mixed-content and FF ignores our rewrite (open FF bug)

So this rule needs either exclusions or spitted into a `platform="mixedcontent"` part.

numismatika:
add `platform="mixedcontent"` to finally get this merged.
WDR takes too much time and not many things on the page are broken

J0WI:
trigger Travis again

Issue #6680, opened 2016-08-30
Pornhub videos not playing


Sanugon:
I can visit the site with no problems but as soon as I click on a video it just doesn't play. It opens the video but instead of playing it gives a black screen where the video should be,

I disabled all of my extensions and tested the videos again and they worked. So I systematically turned on all my extensions one by one till I got to https-everywhere and the problems started again. I then turned off all my extensions again and tested it solely with https-everywhere and it is definitely the culprit. For now I just disabled it for that site.

It's not a major issue but it could potentially cause problems on other video sites which doesn't use https.

Specs:

Google Chrome Version 52.0.2743.116 m (64-bit)
HTTPS Everywhere 2016.8.24

numismatika:
@Sanugon Which rules are active on the site? Does the console report any mixed content issues?

Sanugon:
@numismatika 

Stable Rules

Force encrypted connections to these websites:

PornHub.com (partial)
TrafficJunky.net (partial)
Google APIs
Doublepimp.com (partial)

numismatika:
@Sanugon i can't reproduce that issue here. Are you using flash for the videos by chance?

Sanugon:
@numismatika Funny enough the problem started again. I followed the same procedure as before and now the same thing is happening with Ad Block Plus. It is only ABP that is giving the issue; it still works fine with Ad Block enabled.

I'm starting to wonder if it has more to do with www.pornhub.com than the actual extensions.

Oh and the only flash I use is what comes pre-installed with Chrome. I do use a flash blocker, although the results are the same regardless if it's enabled or not.

Issue #5838, opened 2016-07-28
Squash and merge


Hainish:
@J0WI I think we should squash and merge by default, especially for ruleset changes.  This will make the git history a bit cleaner, especially since PRs are tending to add additional commits just to get the tests to re-run.

https://help.github.com/articles/about-pull-request-merge-squashing/

Hainish:
cc @fuglede

luixxiul:
like this: https://github.com/EFForg/https-everywhere/pull/5788

Hainish:
Ah right, so it looks like this is already being done for the most part.

luixxiul:
Hi, I usually squash my commits by myself locally before/after I create a PR. On #5788 commits are not squashed at all

Hainish:
I think it's fine if contribs don't squash their PRs, as long as the maintainers do it upon merge.  The end result is the same, but with our range of contributors it's more difficult to enforce on their end.

luixxiul:
@Hainish the reason why I do locally is that I tend to forget to pull the squashed commit before making any change on the local repo and mess it up quite badly. It is OK if you do not forget to do that.

fuglede:
Noted; I've typically asked contributors to do it themselves for long-winded PRs with more than, say, 3-4 trivial commits, but I'm fine with just doing it upon inclusion. Any hard limits for what is acceptable?

Hainish:
@fuglede no universal rule can be made here.  If greater granularity is required, it may make sense to break it out into more than a few commits.  But I mostly envision this happening for features and code changes, and less for ruleset changes.  I could see a PR that, say, included multiple types of sites across a few different rulesets having multiple commits that may make sense.  But in that case it's probably more appropriate to have those changes in multiple PRs and not just one.

J0WI:
I like to have unsquashed commits in PR to see what changed since the last feedback.
Squash before merge seems good to me and I'll do this in future.

Hainish:
Thanks!

fuglede:
Could you activate the GitHub feature to do this automatically for the repo? I never tried it but it looks like it will make the process quite a bit easier.

Hainish:
It looks like in the admin UI, I don't have the option to make it default.  I only have the option to uncheck either of the merging methods, which we don't want to do.

fuglede:
Ah, I thought it wasn't even on: I can't choose it here:

[...]

PR #6739, opened 2016-09-02
PornHub: add cdn-d-vid-public, ht & services subdomain

numismatika:
Found a couple more subdomains while checking #6680

### DIFF
--- a/src/chrome/content/rules/PornHub.xml
+++ b/src/chrome/content/rules/PornHub.xml
@@ -53,7 +53,10 @@
 	<target host="cdn-d-img.pornhub.com" />
 	<target host="cdn-d-ss.pornhub.com" />
 	<target host="cdn-d-static.pornhub.com" />
+	<target host="cdn-d-vid-public.pornhub.com" />
+	<target host="ht.pornhub.com" />
 	<target host="rt.pornhub.com" />
+	<target host="services.pornhub.com" />
 	<target host="www.pornhub.com" />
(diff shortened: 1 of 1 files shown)


jeremyn:
- [ ] https://ht.pornhub.com is broken from mixed content, for example the `Learn More` button under `Industry Leading Payouts` doesn't expand. We shouldn't include this as a target. Can you move it under `Mixed content:` in the top comment?

numismatika:
You are right. Google-API is blocked for mixed-content before we can rewrite it.

jeremyn:
Not to rush you just but so the status is clear: the next step is for you to make the requested change, unless you have questions for me.

jeremyn:
Looks good, thanks.

Last thing: we like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) commits, which means we rewrite commits in your name. Is that all right with you for this and any other pull requests you submit?

numismatika:
I used to sqash my commits in a local branch before submitting the PR anyway.
Do how you want to do it. As long as everyone gets better rules i am fine with it.

jeremyn:
Thanks. I've squash-and-merged.

I don't want contributors to squash their own commits. Unsquashed commits are easier to review. Instead, contributors should let us handle it. 

We're discussing what we should do in issue #5838 . Please feel free to add to that discussion. The current process is to squash-and-merge like I described.

PR #6753, opened 2016-09-04
Update Local.ch.xml

J0WI:


J0WI:
No idea why test failed here.

jeremyn:
You have [two targets](https://github.com/J0WI/https-everywhere/blob/43d0eaab6a8c29b370d23d99fe7079ebfdba1fc4/src/chrome/content/rules/Local.ch.xml#L16-L17) for `update.local.ch`.

J0WI:
Ah right, thanks!
I didn't find the `failure` between all the `warnings` in the logs.

J0WI:
See #6754 for the unrelated failure here.

jeremyn:
By the way, for a while Travis wasn't checking the whitelist. @fuglede recently did #6689 that added that check back in, which is I guess is why that unrelated `SmartBear.xml` problem is suddenly appearing now.

jeremyn:
Oh I see, you just merged the new `smartbear.com.xml` in. That's probably the real reason.

jeremyn:
Add target:
- [ ] https://auth.local.ch
- [ ] https://mobile.local.ch
- [ ] https://localsearch.ch
- [ ] https://www.localsearch.ch
- [ ] https://werbung.localsearch.ch
- [ ] https://bin.staticlocal.ch
- [ ] https://mobile.staticlocal.ch

Maybe add target:
- [ ] https://areyou.localsearch.ch (the web console shows mixed content for `jquery.form.js`, which might be serious, but I don't see what's broken. There's a form at the bottom of the page, maybe that doesn't work?)

Remove target and note as mixed content:
- [ ] https://blog.local.ch (footer is gone)

Note as certificate mismatch:
- [ ] https://auth.glb.local.ch

J0WI:
I added the subdomains, thanks!
For the mixed content issues I informed them, let see what happens.

I already have a separate branch for localsearch.ch, but it depends on this one, so I'll push it later.

jeremyn:
Okay, that's fine with `localsearch.ch`.

So you want to merge with https://blog.local.ch as a target and hope they fix it? That's fine.

You added a misspelling which I added to the to-do list above. Instead of `no content` though, maybe instead say something like `Connection reset`? `No content` made me expect a blank response.

PR #6755, opened 2016-09-04
fixed unescaped dot in regex

J0WI:


J0WI:
This currently breaks builds.
ping @fuglede @jeremyn

PR #6757, opened 2016-09-05
Update V2ex.co.xml

renyouguo:


jeremyn:
Thanks. Please:

Add target:
- [ ] https://i.v2ex.co

Also:
- [ ] What is [this `captive` comment](https://github.com/renyouguo/https-everywhere/blob/7edb250e05ba7df9f0dc3e0c8421ce88f7177c4f/src/chrome/content/rules/V2ex.co.xml#L4) for?

jeremyn:
I've updated the checklist through commit a96ab55fbbffe1b25fb3df19a6103b3a8d688723.

I think I understand the comment now. It doesn't matter for the rule though, right? We don't normally make comments about what each target is. Can you just delete the comment?

jeremyn:
Thanks, merged.

PR #6773, opened 2016-09-06
add Autoscout24.ch.xml

J0WI:


jeremyn:
- [ ] Note https://cms.autoscout24.ch as `timeout`

J0WI:
added.

PR #6782, opened 2016-09-06
add rule for Everydayfeminism.com

Mikaela:
I wasn't able to run the tests as the depedency installing script doesn't understand Arch Linux or pacman package manager, but this rule seems to work without breaking anything when added to Firefox custom rule directory. I use some other extensions though (PrivacyBadger could affect the tests), but I hope everything is done properly and your tests will raise the issues if there are any before any harm happens.

### DIFF
--- a/src/chrome/content/rules/Everydayfeminism.com.xml
+++ b/src/chrome/content/rules/Everydayfeminism.com.xml
@@ -1,0 +1,6 @@
+<ruleset name="Everydayfeminism.com">
+	<target host="everydayfeminism.com" />
+	<target host="www.everydayfeminism.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


jeremyn:
Thanks, this looks fine. For development we support Debian/Ubuntu, RedHat/Fedora, and Mac. If you want to add Arch support, that would be welcome and would probably not be too difficult to do.

We like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) commits, which means rewriting your commit in your name. Is that all right?

Mikaela:
> If you want to add Arch support, that would be welcome and would probably not be too difficult to do.

I fear it's outside my abilities/resources for now.

> We like to squash-and-merge commits, which means rewriting your commit in your name. Is that all right?

It's OK as long as you are aware that it erases the GPG signature and that is fine for you.

jeremyn:
That's fine, thanks. Merged.

Mikaela:
> Why would a rule work with Firefox desktop (50.0b5), but not with Firefox Beta for Android (50.0b4)? Both browsers say to be using HTTPS Everywhere 5.2.5 and I see the rule in question <https://github.com/EFForg/https-everywhere/blob/5.2.5/src/chrome/content/rules/Everydayfeminism.com.xml> under the tag 5.2.5. The site also doesn't have separate mobile site that I can see (I don't know anywhere deeper to look than the address bar though).

I wondered this on Sunday at #https-everywhere on irc.oftc.net which seems quite silent. I hope you do read at least GitHub if not IRC or there was some event preventing IRC from seeing.

jeremyn:
I don't follow HTTPS Everywhere IRC channels. The best place to report problems is GitHub.

I can confirm the problem you're seeing, that is, http://everydayfeminism.com doesn't get rewritten to HTTPS on Firefox for Android even with HTTPS Everywhere 5.2.5, but it does get rewritten in desktop Firefox. I don't know why. Could you please submit a new issue for this problem?

Mikaela:
Moved to https://github.com/EFForg/https-everywhere/issues/7317 .

PR #6790, opened 2016-09-07
Update certificates to match latest Firefox (fixes #6569)

### LINKED ISSUE #6569: Test CA repository does not reflect Firefox state

Hainish:
Export of the latest certs as of Firefox 48.0.2


Hainish:
This is ready for review @jsha, @cooperq 

FYI: I renamed several of the Firefox default crt export filenames so that they are compatible with NTFS filesystem naming requirements.

Testing from this seems to be consistent with what we see from a fresh firefox profile (see https://github.com/EFForg/https-everywhere/issues/6569#issuecomment-245142529 and the several comments below it)

cooperq:
lgtm

fuglede:
Excellent!

PR #6804, opened 2016-09-07
Added trivial rule for Hdfcbank.com

mananaBanana:
Trivial rule for HDFC Bank (banking website)

### DIFF
--- a/src/chrome/content/rules/Hdfcbank.com.xml
+++ b/src/chrome/content/rules/Hdfcbank.com.xml
@@ -1,10 +1,13 @@
 <ruleset name="Hdfcbank.com">
 	<target host="hdfcbank.com" />
-	<target host="*.hdfcbank.com" />
+	<target host="*.smartbuy.hdfcbank.com" />
+	<target host="www.hdfcbank.com" />
 
-	<test url="http://www.hdfcbank.com/" />
 	<test url="http://offers.smartbuy.hdfcbank.com/" />
-	<test url="http://m.hdfcbank.com/" />
+	<test url="http://mobileoffers.smartbuy.hdfcbank.com/" />
+	<test url="http://mobileoffers.smartbuy.hdfcbank.com/recharge" />
+	<test url="http://www.hdfcbank.com/" />
+	<test url="http://www.hdfcbank.com/smartemi" />
 
 	<rule from="^http:" to="https:" />
 </ruleset>
--- a/src/chrome/content/rules/Hdfcbank.com.xml
+++ b/src/chrome/content/rules/Hdfcbank.com.xml
@@ -1,8 +1,10 @@
-<!-- Problematic subdomains:
+<!-- 
+	Problematic subdomains:
 		
 		- ideabank.hdfcbank.com (No valid https page)
 		- mailer.hdfcbank.com (https version redirects to http://www.hdfcbank.com. Not adding rule as it would render this subdomain unuseable)
 		- testmailer.hdfcbank.com (https version redirects to http://www.hdfcbank.com)	
+		- https://mobilityrpmcc.hdfcbank.com/ (Incomplete certificate chain)
 
 	HTTPS Only domains:
 		- https://getprepaidcard.hdfcbank.com/
@@ -18,7 +20,6 @@
 		- https://enetbanking.hdfcbank.com/
 		- https://punjabadds.hdfcbank.com/
 		- https://mofpi.hdfcbank.com/
-		- https://mobilityrpmcc.hdfcbank.com/
 		- https://los.hdfcbank.com/
 		- https://dncdedupe.hdfcbank.com/
 		- https://onlineccredemption.hdfcbank.com/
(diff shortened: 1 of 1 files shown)


J0WI:
This is a dupe of https://github.com/EFForg/https-everywhere/pull/6150#issuecomment-241206722 with the same issue as mentioned there.
Please have a look at our guides about finding subdomains.

J0WI:
A wildcard target is not possible here, because some subdomains are broken with https.
E.g. `ideabank.hdfcbank.com` and `mailer.hdfcbank.com` show different content on https.

Please list all supported hosts as targets and comment unsupported ones.

mananaBanana:
How do I figure out which subdomains are causing the problem? Will I have to manually visit and check?

J0WI:
Yes. Sadly there is no way to detect malfunctions like CSP, CORS or mixed content issues.

mananaBanana:
Ok. Is there any way to list all the subdomains of a site? Using a search engine, perhaps? Then I could go about checking all of them and rewrite the ruleset accordingly.
Thanks.

J0WI:
Yes, we have a guide about searching subdomains on Google [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

mananaBanana:
Another question: What about pages that redirect to an http:// site?
Example: https://www.hdfcbank.com/smartemi redirects to  http://www.hdfcbank.com/personal/products/cards/loan-on-credit-card/smartemi. However an https version also exists - https://www.hdfcbank.com/personal/products/cards/loan-on-credit-card/smartemi

All in all, this site has some really messy redirects and lots of mixed content. I am new to all of this and would like it if there is someone to guide me through this.

J0WI:
If a whole host does not support https because of redirects or cert issues we use to add a comment on the top of the file about the cause.

If some paths having major (means breaking functionality) mixed content issues or redirect to http, you can made exclusions. Note that every pattern in the exclusion requires a [test url](https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md).

When there are to much exclusions, you can also be more restrictive in the redirect (e.g. [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/opengg.me.xml)) rule or comment the host out. If the whole host has mixed content issues, you can specify a platform that will disable the ruleset (e.g. [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/infolan.by.xml)).


But your described scenario is much easier to mitigate. If a redirect or a link points to a http url, HTTPS-Everywhere will upgrade it to https. It's only a problem, when the site forces http.

Just ping me if you have more questions.

mananaBanana:
Reduced the ruleset to only the urls with valid https pages. I haven't added pages which have only the https:// page valid.

J0WI:
Please have a look on our [FAQ](https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?) why it's important to add also hosts that already redirect to https.

Could you also document unsupported hosts on the top for future maintenance?

mananaBanana:
Sure, I'll add the comments. 
How do I add https sites? Because `<rule from="^http:" to="https:" />` would be invalid for them, right?

J0WI:
Can you give me an example?

mananaBanana:
Like how: https://getprepaidcard.hdfcbank.com/ is a valid https site but the http version of that site http://getprepaidcard.hdfcbank.com/ is not accessible. So, how do I add these https-only sites to the ruleset?

J0WI:
Since an attacker could still inject a fake response over port 80, I would add them anyway. But leave a comment for future maintenance about it.
I just verified that our fetch tests will pass.

mananaBanana:
Covered https sites as well. Is the comment format okay?

J0WI:
Please see also:
aps.hdfcbank.com (https://aps.hdfcbank.com/EntlWeb/)
corpssl.hdfcbank.com (https://corpssl.hdfcbank.com/EnetSSL/)
divpro.hdfcbank.com (https://divpro.hdfcbank.com/corporateWeb/)
dncdedupe.hdfcbank.com (https://dncdedupe.hdfcbank.com/dnc/)
enetbanking.hdfcbank.com (https://enetbanking.hdfcbank.com/corporate/CorporateLogin.html)
los.hdfcbank.com (https://los.hdfcbank.com/LOSHDFC/CheckLogin.los)
m.hdfcbank.com
mofpi.hdfcbank.com (https://mofpi.hdfcbank.com/EnetMVC/)
onlineccredemption.hdfcbank.com (https://onlineccredemption.hdfcbank.com/Prima_CustomerPortal/frmCustomer.aspx)
punjabadds.hdfcbank.com (https://punjabadds.hdfcbank.com/punjabprocess/login.aspx)
remit.hdfcbank.com
sme.hdfcbank.com (https://sme.hdfcbank.com/LoanFloWeb/Index.jsp)
usedcar.hdfcbank.com

J0WI:
It seems that mobilityrpmcc.hdfcbank.com has an incomplete certificate chain. LGTM otherwise.

J0WI:
Thank you!

PR #6806, opened 2016-09-07
Update thecustomizewindows

AbhishekGhosh:
Let's see.

### DIFF
--- a/src/chrome/content/rules/thecustomizewindows.com.xml
+++ b/src/chrome/content/rules/thecustomizewindows.com.xml
@@ -1,38 +1,20 @@
-<ruleset name="The Customize Windows.com">
-
-	<!--	Direct rewrites:
-				-->
-<target host="thecustomizewindows.com" />
-<target host="www.thecustomizewindows.com" />
-
-	<!--	Complications:
-					-->
-<target host="thecustomizewindows.net" />
-<target host="www.thecustomizewindows.net" />
-
-		<!-- needed for debugging and dev works related to HSTS -->
-
-		<!--exclusion pattern="^http://cloud\.thecustomizewindows\.com/"/-->
-
-			<!--test url="http://cloud.thecustomizewindows.com/" /-->
-
-<!-- Akamai CDN -->
-
-<securecookie host="^a248\.e\.akamai\.net$" name=".+"/>
-
-<!-- allow trusted services -->
-
-<securecookie host="^cdn\.goroost\.com$" name=".+"/>
-<securecookie host="^www\.googletagservices\.com$" name=".+"/>
-<securecookie host="^tpc\.googlesyndication\.com$" name=".+"/>
-<securecookie host="^googleads\.g\.doubleclick\.net$" name=".+"/>
-
-
-<!-- forcing higher security for currently unused dot net domain -->
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/thecustomizewindows.com.xml
```diff
@@ -1,38 +1,20 @@
...
+	<target host="thecustomizewindows.net" />
+	<target host="www.thecustomizewindows.net" />
+	
+	<securecookie host="^thecustomizewindows-3470\.e\.kxcdn\.com$" name=".+"/>
+	<securecookie host="^cdn\.goroost\.com$" name=".+"/>
+	<securecookie host="^www\.googletagservices\.com$" name=".+"/>
+	<securecookie host="^tpc\.googlesyndication\.com$" name=".+"/>
+	<securecookie host="^googleads\.g\.doubleclick\.net$" name=".+"/>
+	
+	<rule from="^http://(?:www\.)?thecustomizewindows\.net/" to="https://thecustomizewindows.com/"/>
```
J0WI: Use `(www\.)?` instead on `(?:www\.)?`
Please also document why this redirect is needed.

src/chrome/content/rules/thecustomizewindows.com.xml
```diff
@@ -1,38 +1,20 @@
...
+	<target host="www.thecustomizewindows.com" />
+	
+	<target host="thecustomizewindows.net" />
+	<target host="www.thecustomizewindows.net" />
+	
+	<securecookie host="^thecustomizewindows-3470\.e\.kxcdn\.com$" name=".+"/>
+	<securecookie host="^cdn\.goroost\.com$" name=".+"/>
+	<securecookie host="^www\.googletagservices\.com$" name=".+"/>
+	<securecookie host="^tpc\.googlesyndication\.com$" name=".+"/>
+	<securecookie host="^googleads\.g\.doubleclick\.net$" name=".+"/>
```
J0WI: Please do not set securecookie rules that don't depend on any target host in this ruleset.

AbhishekGhosh: Thanks for the comment, `securecookie` rules will be removed on next update.

jeremyn:
@AbhishekGhosh Are you still interested in working on this?

AbhishekGhosh:
@jeremyn Thanks for the comment. Yes, it is important for me as the site is itself for new developers (so that I can use as example). But as it is listed on HSTS Preload list so it got pruned later. If someone adds it with minimal ruleset for main domain thecustomizewindows.com, I will be glad. As it was removed as community decision, it will look harsh if I again send pull request.

jeremyn:
It looks like `thecustomizewindows.com` is HSTS preloaded, as you say, so we don't want to add a ruleset for it. `thecustomizewindows.net` is a technically valid target but it just redirects to `thecustomizewindows.com` and it's not worth having a ruleset just for that. As @J0WI wrote, the `securecookie`s aren't appropriate since they are for different domains. So there's nothing to do here.

I'm not sure what you mean in your comment. What minimal ruleset do you mean? Also I'm not sure what you mean by something looking "harsh".

AbhishekGhosh:
@jeremyn Yes, you are correct. Nothing to do in this case.

jeremyn:
All right, thanks @AbhishekGhosh . I'm closing this PR.

PR #6810, opened 2016-09-08
egais.ru

galeksandrp:


src/chrome/content/rules/egais.ru.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+www.egais.ru nonexist
+wiki.egais.ru ³
+
+
+³ timed out
+-->
+<ruleset name="egais.ru">
+	<target host="egais.ru" />
+	<target host="www.egais.ru" />
```
J0WI: Please remove nonexist target.

galeksandrp: Removed.

PR #6814, opened 2016-09-08
boomstarter.ru

galeksandrp:


J0WI:
Tests are passing even with empty target. Same reason here (`Core code changes have been made.`) @Hainish

galeksandrp:
Rebased.

PR #6817, opened 2016-09-08
updated Ok.ru.xml

J0WI:


jeremyn:
- [ ] Add target for https://mobile.ok.ru
- [ ] Sort [these ok.ru subdomains](https://github.com/J0WI/https-everywhere/blob/1cef27edc762f747d47d255e3630dd70d4ff0826/src/chrome/content/rules/Ok.ru.xml#L9-L14)
- [ ] [This Wikipedia article](https://en.wikipedia.org/wiki/Odnoklassniki) says Ok.ru is part of the Mail.ru Group, and https://ok.ru makes calls to `mail.ru` subdomains. I think this ruleset and [Mail.ru.xml](https://github.com/EFForg/https-everywhere/blob/f45195d18399bbda5c5e0c850098d59be281720c/src/chrome/content/rules/Mail.ru.xml) should point at each other.

J0WI:
done (had to exclude some failing hoss on mail.ru).

jeremyn:
None of the seven timeout subdomains actually time out for me, so they should be left in. I can confirm https://lotro.mail.ru is refused though.

J0WI:
Let's try with them included.

PR #6818, opened 2016-09-08
add Welt.de.xml

J0WI:


jeremyn:
Mark as `Different HTTP/HTTPS content`:
- [ ] https://static.apps.welt.de (sometimes? you may want to investigate this more)

Mark as `Incomplete certificate chain`:
- [ ] https://classiqs.welt.de
- [ ] https://stepstone.welt.de

Mark as `Mismatch`:
- [ ] https://apps.welt.de
- [ ] https://iqtest.apps.welt.de
- [ ] https://bundle.welt.de
- [ ] https://liveticker.welt.de
- [ ] https://m.liveticker.welt.de
- [ ] https://tv.welt.de

Mark as `Refused`:
- [ ] https://freie.welt.de
- [ ] https://luxusautos.welt.de
- [ ] https://schmid.welt.de

Add target:
- [ ] https://welt.de
- [ ] https://www.welt.de/(img|static)/ (and/or add some exclusion, whatever is needed)
- [ ] https://mobil.welt.de
- [ ] https://mobile.welt.de
- [ ] https://epaper.apps.welt.de
- [ ] https://co.la.welt.de
- [ ] https://judge.la.welt.de
- [ ] https://lo.la.welt.de
- [ ] https://loader.la.welt.de
- [ ] https://schrotty.la.welt.de

J0WI:
It seems that they recently relaunched their seite and they use https now anyway on their main site.
I updated the rule and included your feedback.

I have no issues with `static.apps.welt.de` and a mismatch on `classiqs.welt.de`.

jeremyn:
https://classiqs.welt.de has a variety of problems for me on a new Firefox profile:

```
classiqs.welt.de uses an invalid security certificate.

The certificate is not trusted because the issuer certificate is unknown.
The server might not be sending the appropriate intermediate certificates.
An additional root certificate may need to be imported.
The certificate is only valid for the following names:
  www.classiqs.com, classiqs.com
The certificate expired on Tuesday, November 24, 2015 11:50. The current time is Monday, September 19, 2016 11:31.

Error code: SEC_ERROR_UNKNOWN_ISSUER
```

`Mismatch` is one of the problems so if you want to call it that, that's fine.

jeremyn:
For `static.apps.welt.de`: these two sets of URLs give different responses:

http://static.apps.welt.de
https://static.apps.welt.de

http://static.apps.welt.de/lesestueck/2015/Postkarten
https://static.apps.welt.de/lesestueck/2015/Postkarten

J0WI:
Thanks for the example on `static.apps.welt.de`. I found some paths that need to be excluded with that information.
All 404 responses looks different on `static.apps.welt.de` over https, but I think that's trivial.

jeremyn:
I didn't check the exclusions but I assume they're fine.

This is extremely minor but the `static.apps.welt.de` exclusion should start with a `^`. Also, while you're in there, `different` is misspelled in the previous line.

Otherwise this is ready to go.

J0WI:
done.

J0WI:
Just let @Hainish that the alexa labels has not been updated here. I think it's because `www.weld.de` was not in the target list in the first commit.

jeremyn:
Thanks @J0WI , merged.

Hainish:
@J0WI yes, that would cause the label not to be applied.  In these cases, we'd have to review all subsequent commits since the initial one to determine if a label should be applied.  This is tricky but probably doable, yet I suspect it isn't worth the effort.

PR #6823, opened 2016-09-09
Rename Criteo.xml to Criteo.com.xml

renyouguo:
There is criteo.net


jeremyn:
It seems like criteo.net and criteo.com are related, is that right? http://peering.criteo.net refers to an email address `@criteo.com`. If these domains are closely related, then both `criteo.net` and `criteo.com` should be handled in the same ruleset.

renyouguo:
#6830

jeremyn:
Thanks.

PR #6825, opened 2016-09-09
Create guoshipartners.com.xml

renyouguo:


jeremyn:
Thanks for the ruleset. Please:

- [ ] Change [this comment](https://github.com/renyouguo/https-everywhere/blob/5215c0ebe03f040e5a96ed7b7d55cfb3bdbd010a/src/chrome/content/rules/guoshipartners.com.xml#L3) to just `www` instead of `^(www.)`, because `^` is not a valid domain name (`guoshipartners.com` is not in public DNS).
- [ ] Remove [this `ad-specs` test](https://github.com/renyouguo/https-everywhere/blob/5215c0ebe03f040e5a96ed7b7d55cfb3bdbd010a/src/chrome/content/rules/guoshipartners.com.xml#L11-L12), because the automatic test is enough.

jeremyn:
Thanks, merged.

PR #6828, opened 2016-09-09
Make trivial rule for ibisworld.com.

jish:


fuglede:
Looks like Travis caught an issue with this one:

    ERROR src/chrome/content/rules/Ibisworld.com.xml: Fetch error: http://ibisworld.com/ => https://ibisworld.com/: (51, "SSL: no alternative certificate subject name matches target host name 'ibisworld.com'")

If you could redirect http://ibisworld.com to https://www.ibisworld.com, that would take care of that.

A different issue I came across: The site https://www.ibisworld.com/about/faq.aspx has a reference to https://www.ibisworld.com/media/, which does not work.

jish:
> "If you could redirect http://ibisworld.com to https://www.ibisworld.com, that would take care of that."

I have redirected http://ibisworld.com to https://www.ibisworld.com

> "The site https://www.ibisworld.com/about/faq.aspx has a reference to https://www.ibisworld.com/media/, which does not work."

How should I resolve this issue?

src/chrome/content/rules/Ibisworld.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Ibisworld.com">
+	<target host="ibisworld.com" />
+	<rule from="^http://ibisworld.com" to="https://www.ibisworld.com" />
+
```
gloomy-ghost: I would suggest to add exclusion as follows:
```
<exclusion pattern="^http://www\.ibisworld\.com/media/" />
<test url="http://www.ibisworld.com/media/" />
````

jish:
I fixed the test failures in Travis CI and added the `<exclusion>` tag.

src/chrome/content/rules/Ibisworld.com.xml
fuglede: @gloomy-ghost: Unfortunately, as far as I can see, this would not be quite sufficient in fixing the issue, as the faulty reference is not due to a HTTPS Everywhere and will persist even with an exclusion. One way around it would be to add a downgrade rule for that particular URL, changing HTTPS requests to HTTP (but be aware that such downgrades come with problems of their own; in particular, if HTTPS will be forced by t he site on that URL at some point in the future, we end up in a situation, where we are *removing* security rather than adding it).

PR #6831, opened 2016-09-10
Create Ubuntu-CN.xml

renyouguo:


jeremyn:
Thanks. Please:

- [ ] Add the rest of [`*.ubuntu.com.cn`](https://www.google.com/search?q=site%3A*.ubuntu.com.cn&ie=utf-8&oe=utf-8#q=site:*.ubuntu.com.cn+-site:forum.ubuntu.com.cn) and [`*.ubuntu.org.cn`](https://www.google.com/search?q=site%3A*.ubuntu.org.cn&ie=utf-8&oe=utf-8#q=site:*.ubuntu.org.cn+-site:forum.ubuntu.org.cn) subdomains
- [ ] Rename the ruleset to `Ubuntu-CN`
- [ ] Rename the file to `Ubuntu-CN.xml`
- [ ] Add comments to both [`Ubuntu.xml`](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/Ubuntu.xml) and this ruleset so they refer to each other

jeremyn:
Thanks. I've updated the checklist through commit 09ae9834ce6ec11ca40623807c074da1bd623d17 and added some more items.

renyouguo:
@jeremyn 
wiki
www

jeremyn:
The HTTPS version though , https://wiki.ubuntu.org.cn , goes to `forum` if you accept the insecure connection.

jeremyn:
So this would normally be a `Different content`-type problem and we would mark the domains as bad. But I think we can call this a misconfiguration and use your original approach of redirecting to `.com.cn`. I've crossed out those items in the checklist.

jeremyn:
Can you please move the `forum` targets back with the other targets? Those don't need to be moved.

jeremyn:
Thanks, merged.

PR #6839, opened 2016-09-11
Update Qbox.me.xml

renyouguo:


jeremyn:
Thanks, merged.

Going forward, if you're making `qbox.me` changes that are related to changes in another ruleset, you can do the changes to both files in the same pull request. For example you could have done the change in pull request #6829 under pull request #6054.

PR #6842, opened 2016-09-11
Update Mozilla.com.tw.xml

renyouguo:


src/chrome/content/rules/Mozilla.com.tw.xml
```diff
@@ -1,13 +1,19 @@
 <!--
 	For other Mozilla coverage, see Mozilla.xml.
-
+	
+	Invalid certificate:
```
J0WI: They all have a incomplete cert chain, including fxosphone.

renyouguo: @J0WI These couse check errors.

J0WI: Yes, the comment is just not accurate. They have a valid cert, but serve an incomplete cert chain.

J0WI:
Thanks!

Issue #6816, opened 2016-09-08
Release 5.2.5 forthcoming


Hainish:
cc @jsha @cooperq @pde @J0WI @fuglede @semenko @loveithateit @diracdeltas @jeremyn 

To be made this afternoon most likely.

This addresses #5874 and updates rulesets.

Hainish:
This is blocking on updates from Google.

jeremyn:
Are we waiting for Google to fix their "reactivate extension" logic before we push a new version and, hopefully, reactivate everyone's extension?

J0WI:
Is the ruleset base already freezed or will current commits still hit the 5.2.5 release?

Hainish:
@J0WI new commits will make it in.

In my discussions with Chromium, there are two layers of possible fixes:

1. We wait for them to push a security fix to Chromium core
2. We apply a fix that the Chromium team gives us, which will require a new release to be made

Option 2 is generally preferable, but Chromium has not yet provided us this fix, and it may or may not actually work.  It's not clear to me that it will work, since I don't have an abundance of knowledge about how the particulars of Extension Content Verification work.

I'm not completely against releasing a new version before they push a fix if necessary, but all other things being equal I'd rather not if it is not necessary.

Hainish:
This will be deployed this week.  Chromium has failed to give us a way forward with option 2.

J0WI:
#6920 might be also worth having a look at.

Hainish:
This has been deployed

PR #6844, opened 2016-09-11
Updated ruleset coverage whitelist

Foorack:


J0WI:
I think this should be done anyway before #6816
cc @Hainish

Hainish:
I've added a step to our internal release checklist to run `./utils/ruleset-coverage-whitelist-cleanup.sh` as part of the release process.

This actually has little to do with the actual release we're making for users, but it's good to do periodically anyway and upon release is as good a time as any.

PR #6858, opened 2016-09-12
BME.hu: redirect vik.wiki alias to wiki.sch.bme.hu

TK-999:
vik.wiki is an alias for wiki.sch.bme.hu, which has an SSL certificate.

This change redirects all vik.wiki requests to wiki.sch.bme.hu (vik.wiki itself doesn't support HTTPS due to cert mismatch)


TK-999:
I temporarily commented out simonyi.bme.hu in the file - certificate expired 6 days ago, causing tests to fail.

PR #6862, opened 2016-09-12
update pandora

redranamber:
I checked pandora with a vpn and the region block works now. So I rewrote the rule to be more complete.

PR #6867, opened 2016-09-13
Update Pandora.xml

redranamber:
Fixed from #6862

### DIFF
--- a/src/chrome/content/rules/Pandora.xml
+++ b/src/chrome/content/rules/Pandora.xml
@@ -9,6 +9,6 @@
 	<test url="http://submit.pandora.com/"/>
 	<securecookie host=".+" name=".+" />
 	<rule from="^http://(amp\.|help\.|mediaserver-cont-\w{3}-\d{1}-v4v6\.|stats\.|submit\.)?pandora\.com" to="https://$1pandora.com"/>
-	<rule from="http://*.p-cdn.com" to= "https://*p-cdn.com"
+	<rule from="http://*.p-cdn.com" to= "https://*.p-cdn.com"
  </ruleset>
 
--- a/src/chrome/content/rules/Pandora.xml
+++ b/src/chrome/content/rules/Pandora.xml
@@ -1,14 +1,14 @@
 <ruleset name="Pandora">
 	<target host="pandora.com"/>
 	<target host="*.pandora.com"/>
-	<exclusion pattern="^http://advertising\.pandora\.com/"/>
-	<exclusion pattern="^http://investor\.pandora\.com/"/>
-	<test url="http://advertising.pandora.com/"/>
+	<target host="*.p-cdn.com"/>
 	<test url="http://amp.pandora.com/"/>
 	<test url="http://help.pandora.com/"/>
-	<test url="http://investor.pandora.com/"/>
+	<test url="http://mediaserver-cont-ch1-1-v4v6.pandora.com/images/public/amz/6/0/8/2/800032806_500W_500H.jpg"/>
 	<test url="http://stats.pandora.com/"/>
 	<test url="http://submit.pandora.com/"/>
 	<securecookie host=".+" name=".+" />
-	<rule from="^http:" to="https:"/>
-</ruleset>
+	<rule from="^http://(amp\.|help\.|mediaserver-cont-\w{3}-\d{1}-v4v6\.|stats\.|submit\.)?pandora\.com" to="https://$1pandora.com"/>
+	<rule from="http://*.p-cdn.com" to= "https://*p-cdn.com"
+ </ruleset
+
--- a/src/chrome/content/rules/Pandora.xml
+++ b/src/chrome/content/rules/Pandora.xml
@@ -1,8 +1,10 @@
 <ruleset name="Pandora">
 	<target host="pandora.com"/>
 	<target host="www.pandora.com"/>
-	<target host="stats.pandora.com/"/>
-	<target host="submit.pandora.com/"/>
+	<target host="amp.pandora.com"/>
+	<target host="help.pandora.com"/>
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the ruleset. Some initial comments:

* We strongly prefer that you list subdomains as targets instead of using a wildcard.
* Please indent the ruleset according to our [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).
* Any list of URLs should be sorted alphabetically starting with the top-level domain and reading left, with `^` and `www` moved to the top. So for example: `^`, `www`, `amp`, `stats`.

redranamber:
Fixed the style but the wildcard is necessary because they store images on unique subdomains.

jeremyn:
Can you give some examples?

redranamber:
https://mediaserver-cont-ch1-1-v4v6.pandora.com/images/public/amz/6/0/8/2/800032806_500W_500H.jpg
https://mediaserver-cont-sv5-3-v4v6.pandora.com/images/public/rovi/albumart/2/2/2/7/093624987222_500W_500H.jpg

jeremyn:
I guessed at the URL scheme and found these URLs by hand using DNS lookup:

https://mediaserver-cont-ch1-1-v4v6.pandora.com
https://mediaserver-cont-ch1-2-v4v6.pandora.com
https://mediaserver-cont-sv3-1-v4v6.pandora.com
https://mediaserver-cont-sv3-2-v4v6.pandora.com
https://mediaserver-cont-sv3-3-v4v6.pandora.com
https://mediaserver-cont-sv5-1-v4v6.pandora.com
https://mediaserver-cont-sv5-2-v4v6.pandora.com
https://mediaserver-cont-sv5-3-v4v6.pandora.com

@redranamber Can you casually browse a little in Pandora and see if you find any `mediaserver` URL that isn't in this list?

@fuglede @J0WI Can either of you think of a way to effectively add a target for `mediaserver-cont-*-v4v6.pandora.com` ?

redranamber:
I also see some coming from https://mediaserver-cont-dc6-2-v4v6.pandora.com and many are also coming from *.p-cdn.com such as http://cont-2.p-cdn.com/images/public/amz/9/2/3/8/781676658329_500W_500H.jpg which appears to be a Pandora domain.

redranamber:
I did a search and found it is http://cont-2.p-cdn.com.ipaddress.com/

jeremyn:
`mediaserver-cont-dc6-2-v4v6.pandora.com` and `cont-2.p-cdn.com` both seem to have guessable number schemes that you could list specifically.

When we list targets specifically, we can check each of them and confirm they're good. If we use wildcards though, then we might break functionality for subdomains we don't know about. Especially for a high traffic site like Pandora, we want to be careful.

redranamber:
I also see
http://t3-2.p-cdn.com/
http://cont-1.p-cdn.com/
http://t2-1.p-cdn.com/
http://t2-2.p-cdn.com/

J0WI:
> @fuglede @J0WI Can either of you think of a way to effectively add a target for mediaserver-cont-*-v4v6.pandora.com ?

No, but you can use a wildcard target and regex with all working hosts in the from to rule. But this also requires test urls.

jeremyn:
@J0WI That's a good idea, thanks.

@redranamber So what that means is that instead of

```xml
<rule from="^http:" to="https:"/>
```

you should do something like

```xml
<rule from="^http://(advertising\.|amp\.|mediaserver-cont-.+-v4v6\.)?pandora\.com" to="https://$1pandora.com"/>
```

Of course you need to add the other subdomains and test it, but that's the general idea.

fuglede:
The unescaped dot will not pass the tests, but you could use either `\d` or just the relevant character classes.

jeremyn:
@fuglede You mean the dot here: `.+`? You probably know, but that's not a period, but rather a regex wildcard.

fuglede:
Yep, but they are not currently allowed; https://github.com/EFForg/https-everywhere/blob/master/utils/trivial-validate.py#L77 (we had a brief discussion on this recently that I might be able to dig up if necessary.)

jeremyn:
I see, thanks. In that case, replace the second rule in my earlier comment with this:

```xml
<rule from="^http://(advertising\.|amp\.|mediaserver-cont-\w{3}-\d{1}-v4v6\.)?pandora\.com" to="https://$1pandora.com"/>
```

redranamber:
It seems the p-cdn.com resources need to be secured also to prevent them from being blocked as mixed content. So this includes an attempt at securing it.

jeremyn:
By the way, recent issue #6875 is a good example of why wildcards are risky.

redranamber:
I think this is a case where it's more likely to have an issue caused by not using a wildcard. I keep finding new subdomains while testing and some of them are being blocked automatically by Firefox as mixed content. Another I just saw is http://audio-ch1-t3-1-v4v6.pandora.com/

terrorist96:
#3069

redranamber:
I think for now the rule should just include the domains I haven't seen any problems with. So i have cut down the rule to just those.

jeremyn:
I think we should really try to include domains like `mediaserver` and `p-cdn.com`. Where are you getting mixed content blocking (MCB)? You shouldn't get MCB loading a `.jpg` for example.

I don't have a Pandora account so I can't help you search for these domains myself, sorry. In fact for that reason I'm probably not the best reviewer for this. @fuglede @J0WI Do either of you have a Pandora account you can use, and if so can you take over here?

fuglede:
> @fuglede @J0WI Do either of you have a Pandora account you can use, and if so can you take over here?

Unfortunately not, no.

Hainish:
I have an account.  LGTM

Issue #6863, opened 2016-09-12
How involved should the ruleset review process be?


jeremyn:
As a ruleset reviewer, I'm finding that most pull requests for "institutional" sites are usually very incomplete. My usual experience is that I do the [Google search trick](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) with the included domains and find sometimes dozens of missed sites, which I then write up as items to be completed. For example:
1. [`fnal.gov`](https://github.com/EFForg/https-everywhere/pull/5265#issuecomment-242941309), government research lab
2. [`migros.ch`](https://github.com/EFForg/https-everywhere/pull/6780#issuecomment-246150693), retailer
3. [`th-nuernberg.de`](https://github.com/EFForg/https-everywhere/pull/6857#issuecomment-246513409), school
4. [`udn.com`](https://github.com/EFForg/https-everywhere/pull/6824#issuecomment-246125911), news site

This has several problems. It takes time. There's no guarantee the contributor will even do the changes, which means my time might have been wasted. I imagine it's an unpleasant surprise to the contributor to discover all this extra work they need to do. For example in (3), the contributor was solving an issue someone else reported that was simply "Add https://www.th-nuernberg.de ", and now I'm asking them to add domains like https://bwc188113.bw.fh-nuernberg.de . Also, in (4), most of the domains to add were HTTPS failures anyway.

Ideally we want to encourage contributors to add as many domains as possible, especially if it's a new ruleset or they are doing a major update. But my current strategy seems too extreme. I think going forward what I'll do is, for new rulesets or major updates, if search trick reveals only a small number of missing domains, I'll note them specifically. If it's a lot of missing domains, I'll ask them to please add more and give them the search to get started. On the second review I'll just work with what they have and not ask them to add any more. And for pull requests that are either targeted fixes or adding a small number of domains, I'll just review what they've given.

@fuglede @Hainish @J0WI What do you all think?

J0WI:
I agree that we should not be that strict for bug fixes or small additions.
For new or rebased rules I think we should try to have full coverage, because we are HTTPS *Everywhere*, so we won't secure users only on `^` and `www.` urls. It also helps us to find bugs, especially for wildcard rules.

[...]

PR #6868, opened 2016-09-13
improve make-trivial-rule

J0WI:
improve make-trivial-rule duplicate heuristic
add comment block for nonfunctional hosts

### DIFF
--- a/src/chrome/content/rules/make-trivial-rule
+++ b/src/chrome/content/rules/make-trivial-rule
@@ -34,6 +34,15 @@ fi
 
 lower=$(echo "$1" | tr A-Z a-z)
 escaped=$(echo "$lower" | sed 's/\./\\./g' )
+
+duplicate=$(grep -EHin '<target\s+host="(w+\.)?'$escaped'"\s*/>' *.xml)
+if [ -n "$duplicate" ]
+then
+  echo "error: $lower already exists:" >&2
+  echo "$duplicate" >&2
+  exit 1
+fi
+
 if [ "$#" -gt 1 ] ; then
     to="$2"
 else
@@ -41,6 +50,15 @@ else
 fi
 
 cat > "$dest" <<END
+<!--
+	Nonfunctional hosts in *.$lower:
+
+	ʰ http redirect
+	ᵐ certificate mismatch
+	ʳ connection refused
+	ˢ self-signed certificate
+	ᵗ timeout on https
+-->
 <ruleset name="$capitalized">
 	<target host="$lower" />
 	<target host="www.$lower" />
--- a/src/chrome/content/rules/make-trivial-rule
+++ b/src/chrome/content/rules/make-trivial-rule
@@ -53,11 +53,11 @@ cat > "$dest" <<END
 <!--
 	Nonfunctional hosts in *.$lower:
 
-	ʰ http redirect
-	ᵐ certificate mismatch
-	ʳ connection refused
-	ˢ self-signed certificate
-	ᵗ timeout on https
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
 -->
 <ruleset name="$capitalized">
 	<target host="$lower" />


J0WI:
This is a really useful tool to create simple rules, I think we should also recommend this in our guides.

We may also give a hint in this script to manually search for more subdomains. Or even providing the search link?

I decided to use the letters from the superscript minuscule Unicode table for common errors. This hopefully brings more consistent and order in our comments.
cc @loveithateit if I missed something there.

jeremyn:
@J0WI I think it's great you did this work and I definitely like having a template top comment.

I don't like the superscript pattern because it's harder to read, and Unicode can be annoying. I'd much rather just use `(h)`, `(m)` etc.

When you specifically list categories, contributors who encounter a problem that's not listed like `incomplete certificate chain` or some bizarre, rare HTTPS configuration problem have to figure out what to do. You could add an `other problem` category. Also I've been thinking that we should just roll `self-signed`, `mismatch` etc into a single `invalid certificate` category. We really don't need that specific breakdown and it's annoying to have to use multiple categories.

Should we ask contributors who submit a new ruleset with no problematic subdomains to delete the top comment?

Also, when I make new rulesets, I just copy an existing ruleset and edit it. I haven't used `make-trivial-rule` in a while. Just throwing that out there.

J0WI:
> I don't like the superscript pattern because it's harder to read, and Unicode can be annoying. I'd much rather just use (h), (m) etc.

Personally I prefer Unicode symbols for readability. We also already have quite a few rules using them:
```
$ grep " ᵐ$" * | wc -l
1072
$ grep " m$" * | wc -l
59
```
The initial idea of the comments template was, that I was not able to type these characters on my keyboard and it was annoying to search them. So I wasn't using them.
I hope this will make it easier.

> When you specifically list categories, contributors who encounter a problem that's not listed like incomplete certificate chain or some bizarre, rare HTTPS configuration problem have to figure out what to do. You could add an other problem category. Also I've been thinking that we should just roll self-signed, mismatch etc into a single invalid certificate category. We really don't need that specific breakdown and it's annoying to have to use multiple categories.

The best practice for mixed content or cert chain issues would be to use a platform.
For uncommon issues contributors will have to write an extra comment/exclusion anyway.

> Should we ask contributors who submit a new ruleset with no problematic subdomains to delete the top comment?

I thought this is self-explaining, but we could easily do that.

jeremyn:
I know we have a bunch of rulesets with this Unicode pattern, I've seen them and don't like it.

The Unicode is just plain hard to read for my eyeballs, even with a large monitor. Also I feel that almost anyone would agree that 

```
Refused:
    a.example.com
    c.example.com

Self-signed:
    b.example.com

Timeout:
    d.example.com
    e.example.com
```

Is easier to write and review than:

```
    a.example.com ʳ
    b.example.com ˢ
    c.example.com ʳ
    d.example.com ᵗ
    e.example.com ᵗ

ʳ connection refused
ˢ self-signed certificate
ᵗ timeout on https
```

Also I really don't want to have to walk a contributor through using Unicode footnotes to comment something like [`migros.xml`](https://github.com/EFForg/https-everywhere/blob/49bd7f60e9d573ebf0a2d0d9016ba9f5c2edb175/src/chrome/content/rules/migros.xml). What would that look like?

J0WI:
The sort-by-category solution is nice for small rulesets, but when you have many commented subdomains you have to sort each category separately.

When you have all comments of a host in a list, you can easy add new ones and then just sort the whole list. Therefore it's also easier to find a host in the list, because they are only sorted in alphabetical order and not spitted into categories.

Maybe you have more than one domain in one ruleset, then you need to sort it in three times in the first example and only two times in the second:

```
Refused:
    a.example.com
    c.example.com

    a.example.org
    c.example.org

Self-signed:
    b.example.com

    b.example.org
```
vs
```
    a.example.com ʳ
    b.example.com ˢ
    c.example.com ʳ

    a.example.org ʳ
    b.example.org ˢ
    c.example.org ʳ

ʳ connection refused
ˢ self-signed certificate
```


In the category list you also can't note multiple issues, e.g.:
```
	bad.example.com ⁴ʰᵐ

	⁴ 404/diffrent content
	ʰ http redirect
	ᵐ certificate mismatch
```
(Means bad.example.com serves a mismatching cert, if you accept that it will redirect the wrong location and to http)


But let see how others think about it.
cc @fuglede @Hainish

Hainish:
I think there's a middle-ground here.  I agree with the readability of non-superscripted characters, as well as the niceties of having one master list and the ability to tag multiple times for a given host.  These are not mutually exclusive.  Take the following:

```
    a.example.com (r)
    b.example.com (s)
    c.example.com (r)
    d.example.com (t)
    e.example.com (t)
    f.example.com (4,s)

r: connection refused
s: self-signed certificate
t: timeout on https
4: 40X response
```

Look good?

J0WI:
These are some more characters to type, but I'm also fine with that.

jeremyn:
If we're going to use annotations then I agree regular ASCII is better than Unicode superscripts.

Does `(4, s)` actually come up in practice? Who would discover that?

I think the exhaustive list of categories ought to be this, with whatever names we find appropriate:

* Site doesn't load (includes `refused`, `timeout`, `secure connection failed`)
* Site loads, invalid certificate (includes `expired`, `self-signed`, `mismatch`, `incomplete certificate chain`)
* Site loads, bad mixed content blocking
* Site loads, HTTPS -> HTTP redirect loop

That's it. Possibly `incomplete certificate chain` should be split from `invalid certificate` because an incomplete chain can require specific testing to see.

Also see the discussion [here](https://github.com/EFForg/https-everywhere/issues/6863#issuecomment-248326491).

Hainish:
@jeremyn I think we should settle on some set of standard categories and note them in https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md.  But I don't think we should prematurely limit ourselves to a predefined set and prevent ourselves from adding more.  Who knows what situations will come up.  For instance, as SHA-1 certs become deprecated, if the [StartSSL SHA-1 heisenbug](https://sslmate.com/blog/post/chrome_cached_sha1_chains) presents a problem for many of the subdomains on a specific ruleset, we'll want to note that.  So I don't see any reason not to create a new category for it.

jeremyn:
If we're going to pre-list categories in the trivial rule template, the list of categories should minimal, obvious, and complete. `make-trivial-rule` is for new contributors who aren't sure what to do and are nervous about breaking things. For obscure problems we can walk them through how to document it but that should be unnecessary in most cases.

J0WI:
I plan to add a CONTRIBUTING.md file when this one is done.

Hainish:
With `CONTRIBUTING.md` (which I think is a great idea) perhaps we can include the canonical, full list there, and have some common subset included in `make-trivial-rule`.

J0WI:
I just updated this without Unicode chars.

Hainish:
lgtm!

PR #6243, opened 2016-08-13
Remove broken ruleset. Fix #5161.

tomwys:

PR #6869, opened 2016-09-13
Update whitelist for FierceMarkets.xml

jeremyn:
This updates the whitelist following the change in pull request #6243 .

@fuglede @Hainish @J0WI Please review this soon because this is breaking Travis.


J0WI:
I wonder why the tests passed in #6243 .

jeremyn:
I'm not totally sure but there were recent changes to how the whitelist is handled in Travis, see pull request #6689 . If you were relying on a test ran last month then the test result may have been outdated.

fuglede:
Sounds weird; do you have a reference for a broken build? https://travis-ci.org/EFForg/https-everywhere/builds looks mostly green.

jeremyn:
@fuglede https://travis-ci.org/EFForg/https-everywhere/builds/159734794

fuglede:
Thanks for the reference. #6689 only concerned fetch testing, so that *should* not be relevant here.

In fact, the job from #6243 correctly caught the failure (see https://travis-ci.org/EFForg/https-everywhere/jobs/152052034) but the `docker` run gave the wrong return value, leading to the false green; I could imagine that this would be fixed now that the test has been split into six different ones -- I guess we'll find out pretty quickly, but maybe @Hainish will also be able to tell immediately?

PR #6877, opened 2016-09-14
[cnzz] Ditch cnzz script sites, fix #5602

### LINKED ISSUE #5602: Must disable Cnzz.com or Bilibili.com's video will get error #2048

Artoria2e5:
These sites generate scripts that refer to other CNZZ hosts with nonfunctional HTTPS when visited through HTTPS.

### DIFF
--- a/src/chrome/content/rules/Cnzz.com.xml
+++ b/src/chrome/content/rules/Cnzz.com.xml
@@ -3,6 +3,9 @@
 	- cnzz.com
 	- mail.cnzz.com
 	- www.cnzz.com
+	- s[0-9]+.cnzz.com (#5602)
+	- c.cnzz.com (#5602)
+	- w.cnzz.com (#5602)
 	
 	Redirects to http:
 	- adm.cnzz.com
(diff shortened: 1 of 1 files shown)


J0WI:
Can you give some examples of such hosts?

Referring to Travis `db.cnzz.com` does not exist anymore, please remove that target.

Artoria2e5:
@J0WI See #5062.

The page http://www.bilibili.com/video/av6050274/ has a player that requires some cookie from cnzz (some sort of tracker) to work, and it references the script http://s11.cnzz.com/stat.php?id=2724999&web_id=2724999. The script will load another tracking element which finally injects the cookie. For this particular `id`, the script returns `hzs11.cnzz.com`, a server with some sort of blackholed HTTPS (time out).

When visited through HTTPS, s11.cnzz.com trys to be clever by showing you `https://` in the string constant soup instead of `http://`. That unfortunately take you to the black hole.

I tested a few other domains (c, w, s1, s12, s123...) with the path `/stat.php?id=2724999&web_id=2724999`, all giving the same sort of js chowder that takes you to `hzs11.cnzz.com`, with the same HTTPS glitch.

J0WI:
I was not able to reproduce the issue described in #5602, but my system may block the tracking scripts entirely.

J0WI:
I just tested this on a fresh Firefox profile and the video still works for me.
Note that I accepted the "默认使用Html5播放器" (default html5 player) pop-up.

Artoria2e5:
@J0WI The HTML5 player is kinda new, so they are generally less likely to do stupid things like "fail without tracking cookie" in it. Unfortunately certain videos and features are only available in the Flash-based players.

J0WI:
Let's see how @jeremyn thinks about this, since he was involved in #5602.

jeremyn:
I can't reproduce the error now in either Firefox or Chrome, which is different than what I reported in the issue [here](https://github.com/EFForg/https-everywhere/issues/5602#issuecomment-243614021). Both browsers are using Flash, not HTML5.

gloomy-ghost:
#5602 has already been closed.

PR #6871, opened 2016-09-14
Fix ITunes.com.xml HTTPS Mismatch

weems:
Closing #6870

PR #6878, opened 2016-09-14
Fix how 'Location' is parsed from header text

jeremyn:
This fixes the test problem identified in pull request #6871 , where [this code](https://github.com/EFForg/https-everywhere/blob/a37a897b014360893646a8699715bc5e58c011d0/test/rules/src/https_everywhere_checker/http_client.py#L382-L389) apparently (?) never finds the `Location` header, for example:

```
(Pdb) self._headerRe
regex.Regex('(?P<name>\\S+?): (?P<value>.*?)\\r\\n', flags=regex.V0)
(Pdb) headerStr
'HTTP/1.1 301 Moved Permanently\nServer: Apache\nDate: Tue Jun  1 12:48:03 PDT 1999 PDT\nReferer: http://itunes.com/\nLocation: http://www.apple.com/itunes/?cid=OAS-US-DOMAINS-itunes.com\nContent-type: text/html\nContent-length: 311\n\n'
(Pdb) headers = self._headerRe.findall(headerStr)
(Pdb) headers
[]
(Pdb)
```

Note that if in fact `location` has been incorrectly `False`-y for a while, then it's possible the code immediately after the `raise HTTPFetcherError` line has not been run recently or ever, and this fix may reveal new bugs in it.

Pinging @fuglede @Hainish @J0WI for review.

### DIFF
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -378,9 +378,9 @@ class HTTPFetcher(object):
 				raise HTTPFetcherError("Pycurl fetch failed for '%s'" % newUrl)
 			elif httpCode in (301, 302, 303, 307):
 				location = None
-				for piece in headerStr.lower().split('\n'):
-					if piece.startswith('location:'):
-						location = ':'.join(piece.split(':')[1:]).strip()
+				for piece in headerStr.split('\n'):
+					if piece.lower().startswith('location:'):
+						location = piece[len('location:'):].strip()
 				if location is None:
 					raise HTTPFetcherError("Redirect for '%s' missing location header" % newUrl)
 				
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -175,8 +175,6 @@ class HTTPFetcher(object):
 	"""Fetches HTTP(S) pages via PyCURL. CA certificates can be configured.
 	"""
 	
-	_headerRe = regex.compile(r"(?P<name>\S+?): (?P<value>.*?)\r\n")
-	
 	def __init__(self, platform, certPlatforms, fetchOptions, ruleTrie=None):
 		"""Create fetcher that validates certificates using selected
 		platform.
@@ -379,13 +377,11 @@ class HTTPFetcher(object):
 			if httpCode == 0:
 				raise HTTPFetcherError("Pycurl fetch failed for '%s'" % newUrl)
 			elif httpCode in (301, 302, 303, 307):
-				# Parse out the headers and extract location, case-insensitively.
-				# If there are multiple location headers, pick the last one.
-				headers = dict()
-				for k, v in self._headerRe.findall(headerStr):
-					headers[k.lower()] = v
-				location = headers.get('location')
-				if not location:
+				location = None
+				for piece in headerStr.split('\n'):
+					if piece.startswith('Location:'):
+						location = ''.join(piece.split(': ')[1:])
+				if location is None:
 					raise HTTPFetcherError("Redirect for '%s' missing Location" % newUrl)
 				
 				location = self.absolutizeUrl(newUrl, location)
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -378,11 +378,11 @@ class HTTPFetcher(object):
 				raise HTTPFetcherError("Pycurl fetch failed for '%s'" % newUrl)
 			elif httpCode in (301, 302, 303, 307):
 				location = None
-				for piece in headerStr.split('\n'):
-					if piece.startswith('Location:'):
+				for piece in headerStr.lower().split('\n'):
+					if piece.startswith('location:'):
 						location = ''.join(piece.split(': ')[1:])
 				if location is None:
-					raise HTTPFetcherError("Redirect for '%s' missing Location" % newUrl)
+					raise HTTPFetcherError("Redirect for '%s' missing location header" % newUrl)
 				
 				location = self.absolutizeUrl(newUrl, location)
 				logging.debug("Following redirect %s => %s", newUrl, location)
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -380,7 +380,7 @@ class HTTPFetcher(object):
 				location = None
 				for piece in headerStr.lower().split('\n'):
 					if piece.startswith('location:'):
-						location = ''.join(piece.split(': ')[1:])
+						location = ':'.join(piece.split(':')[1:]).strip()
 				if location is None:
 					raise HTTPFetcherError("Redirect for '%s' missing location header" % newUrl)
 				


test/rules/src/https_everywhere_checker/http_client.py
```diff
@@ -379,13 +377,11 @@ def fetchHtml(self, url):
...
-				# Parse out the headers and extract location, case-insensitively.
-				# If there are multiple location headers, pick the last one.
-				headers = dict()
-				for k, v in self._headerRe.findall(headerStr):
-					headers[k.lower()] = v
-				location = headers.get('location')
-				if not location:
+				location = None
+				for piece in headerStr.split('\n'):
+					if piece.startswith('Location:'):
```
Hainish: We should be careful here.  Although the `Location` header is normally

Hainish: We should be careful here.  Although the `Location` header is usually capitalized in this way, it is not always.  See https://www.shodan.io/search?query=location%3A

jeremyn: You're right, thanks. Also [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2) says header field names are case-insensitive. This should be easy to fix.

test/rules/src/https_everywhere_checker/http_client.py
```diff
@@ -379,14 +377,12 @@ def fetchHtml(self, url):
...
-				headers = dict()
-				for k, v in self._headerRe.findall(headerStr):
-					headers[k.lower()] = v
-				location = headers.get('location')
-				if not location:
-					raise HTTPFetcherError("Redirect for '%s' missing Location" % newUrl)
+				location = None
+				for piece in headerStr.lower().split('\n'):
+					if piece.startswith('location:'):
+						location = ''.join(piece.split(': ')[1:])
```
Hainish: For consistency, you might want to make the argument for the `startswith` call the same as the argument for the `split` call.  This way, if the web server returns an improperly formatted header (lacking the space) it won't cause an error or unintended behavior.

test/rules/src/https_everywhere_checker/http_client.py
```diff
@@ -379,14 +377,12 @@ def fetchHtml(self, url):
 			if httpCode == 0:
 				raise HTTPFetcherError("Pycurl fetch failed for '%s'" % newUrl)
 			elif httpCode in (301, 302, 303, 307):
-				# Parse out the headers and extract location, case-insensitively.
-				# If there are multiple location headers, pick the last one.
```
Hainish: I'm curious, why remove the above comments?

jeremyn: I distrust comments in general. The code here is six lines long, it doesn't need two lines of comments.

test/rules/src/https_everywhere_checker/http_client.py
jeremyn: Good point again, I'll change it. Actually though it would be properly formatted. From [the same RFC](https://tools.ietf.org/html/rfc7230#section-3.2) (bold mine):

> Each header field consists of a case-insensitive field name followed by a colon (":"), **optional** leading whitespace, the field value, and **optional** trailing whitespace.

Hainish: Ah, interesting - I've never seen headers lacking the leading whitespace.  In that case you might want to change just line 383 to be:

```python
location = ''.join(piece.split(':')[1:]).strip()
```

test/rules/src/https_everywhere_checker/http_client.py
Hainish: Fair enough.  I differ in opinion on comments, but I don't think it matters in this case.

test/rules/src/https_everywhere_checker/http_client.py
jeremyn: `''.join` instead of `':'.join` destroys the `:`. This wasn't noticeable with the much less likely `:<space>`, but very noticeable with `:` as in `http://`. (I made that mistake too.)

test/rules/src/https_everywhere_checker/http_client.py
```diff
@@ -379,14 +377,12 @@ def fetchHtml(self, url):
...
-				headers = dict()
-				for k, v in self._headerRe.findall(headerStr):
-					headers[k.lower()] = v
-				location = headers.get('location')
-				if not location:
-					raise HTTPFetcherError("Redirect for '%s' missing Location" % newUrl)
+				location = None
+				for piece in headerStr.lower().split('\n'):
+					if piece.startswith('location:'):
+						location = ':'.join(piece.split(':')[1:]).strip()
```
Hainish: I just spotted the source of another possible problem.  Here, the location will always be a lowercased version of the URL, due to line `381`.  URLs should be treated as case-sensitive, though.

> URLs in general are case-sensitive (with the exception of machine names). There may be URLs, or parts of URLs, where case doesn't matter, but identifying these may not be easy. Users should always consider that URLs are case-sensitive.
Ref: https://www.w3.org/TR/WD-html40-970708/htmlweb.html

jeremyn: Right again, in fact that comes up in the https://itunes.com example because the `Location` there is http://www.apple.com/itunes/?cid=OAS-US-DOMAINS-itunes.com . I'll fix it tomorrow.

Hainish:
Great, lgtm

fuglede:
Does this mean that this is actually a bug in curl in general?

jeremyn:
No, I think the bug was entirely on our side because we failed to parse out the `Location` header. What would the curl bug be?

fuglede:
I was just thinking about https://github.com/EFForg/https-everywhere/pull/6871#issuecomment-247150128.

jeremyn:
Oh. Of course @J0WI has to confirm but I assume they meant that they saw `Location` with curl and so confirmed that the bug was in our stuff.

fuglede:
Ah, could be. I'm just getting errors altogether;

    $ curl -I https://www.itunes.com
    curl: (51) SSL: no alternative certificate subject name matches target host name 'www.itunes.com'
    $ curl -I https://itunes.com
    curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104

jeremyn:
https://www.itunes.com has an invalid certificate, that's what #6871 is trying to fix.

I get something similar with `curl -I https://itunes.com`, no idea what that means. `curl -v https://itunes.com` shows the response headers though, including `Location`.

fuglede:
Looks like it's an iTunes thing: https://github.com/curl/curl/issues/1020#issuecomment-247793526.

jay:
I don't know if this helps but when I did a GET on it through Fiddler debugging proxy the server failed a compliance check:

> X-HTTPPROTOCOL-VIOLATION: [ProtocolViolation] The Server did not return properly formatted HTTP Headers. HTTP headers should be terminated with CRLFCRLF. These were terminated with LFLF.

Hainish:
@jeremyn I get no `Location` header with `curl -v`

```bash
(.venv)user@personal:~$ curl -I https://www.itunes.com
curl: (51) SSL: no alternative certificate subject name matches target host name 'www.itunes.com'
(.venv)user@personal:~$ curl -v https://www.itunes.com
* Rebuilt URL to: https://www.itunes.com/
* Hostname was NOT found in DNS cache
*   Trying 17.172.224.35...
* Connected to www.itunes.com (17.172.224.35) port 443 (#0)
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / AES128-GCM-SHA256
* Server certificate:
* 	 subject: 1.3.6.1.4.1.311.60.2.1.3=US; 1.3.6.1.4.1.311.60.2.1.2=California; businessCategory=Private Organization; serialNumber=C0806592; C=US; postalCode=95014; ST=California; L=Cupertino; street=1 Infinite Loop; O=Apple Inc.; OU=Internet Services; CN=itunes.com
* 	 start date: 2015-08-19 00:00:00 GMT
* 	 expire date: 2017-08-19 23:59:59 GMT
* 	 subjectAltName does not match www.itunes.com
* SSL: no alternative certificate subject name matches target host name 'www.itunes.com'
* Closing connection 0
* SSLv3, TLS alert, Client hello (1):
curl: (51) SSL: no alternative certificate subject name matches target host name 'www.itunes.com'
```

FYI, the error indicates that this host is using [SAN](https://en.wikipedia.org/wiki/Subject_Alternative_Name) to indicate which FQDNs the certificate is valid for, but the names listed in the x509 (itunes.com) didn't match the host name you specified (www.itunes.com).  You can verify that with this command:

    echo 1 | openssl s_client -connect www.itunes.com:443 | openssl x509 -noout -text

Looking at the `x509v3 Subject Alternative Name` field, we only get `itunes.com` and not `www.itunes.com`

jeremyn:
@Hainish https://itunes.com is what redirects and has the `Location` header, not https://www.itunes.com .

Hainish:
Ah.  Yeah in that case the SSL error see is strange.  For other https hosts that redirect, `curl -I` doesn't give me that error:

```
(.venv)user@personal:~$ curl -I https://eff.org
Server: nginx
Date: Tue, 20 Sep 2016 00:25:37 GMT
Content-Type: text/html
Content-Length: 178
Connection: keep-alive
Location: https://www.eff.org/
Strict-Transport-Security: max-age=31536000; includeSubdomains
X-Content-Type-Options: nosniff
```

PR #6879, opened 2016-09-14
Remove trailing slashes from Videa.hu.xml rule

jeremyn:
This should fix a problem introduced by pull request #6860 .

@fuglede @Hainish @J0WI This is breaking Travis, see [here](https://travis-ci.org/EFForg/https-everywhere/jobs/160021213).

PR #6881, opened 2016-09-15
Add utils/validate.sh to rules test

Hainish:
This should fix the issue discovered in https://github.com/EFForg/https-everywhere/pull/6879

cc @cooperq


test/travis.sh
```diff
@@ -56,6 +56,7 @@ if [ "$RULESETS_CHANGED" ]; then
     docker_build
     docker run --rm -ti -v $(pwd):/opt httpse python utils/trivial-validate.py
     docker run --rm -ti -v $(pwd):/opt httpse python utils/ruleset_filenames_validate.py
+    docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/validate.sh"
```
fuglede: `utils/validate.sh` runs `trivial-validate.py` which is also run two lines above, so there is some redundancy here.

Hainish: Nice catch, fixing now

PR #6886, opened 2016-09-15
Added wired.com

Foorack:
https://www.wired.com/2016/09/now-encrypting-wired-com/


jeremyn:
Wow, good catch! Please:

Mark as `Refused`:
- [ ] https://feeds.wired.com

Add target:
- [ ] https://assets.wired.com

Also:
- [ ] Add a generic `<securecookie host=".+" name=".+" />` tag (see [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md)) between the `target`s and `rule`. These are good to have when a site lets users log in and everything supports HTTPS.

jeremyn:
Good find on the old `Wired.xml` file. I've updated the checklist with some more items related to that.

Foorack:
@jeremyn All done.

jeremyn:
Thanks, merged.

PR #6895, opened 2016-09-16
I18n tor

Hainish:
This makes it so that en_US is not hardcoded, so that builds like https://people.torproject.org/~linus/builds/tbb-nightly-hardened-2016-09-15/tor-browser-linux64-tbb-nightly-hardened_ALL.tar.xz work with our tor testing script


Hainish:
Ready for review @cooperq

PR #6900, opened 2016-09-17
[Energinet.dk] Fix various issues

fuglede:
The existing ruleset caused a redirect loop due to the inclusion on the following script on all content containing pages:

```html
	<script type="text/javascript">
	{
	  var url=location.href;
	  if(url.indexOf("https://") > -1)
	  {
	    var newurl=url.replace("https://","http://");
	    location.href=newurl;
	  }
	}
	</script>
```

This fixes that by exclusion of those pages.

Moreover, clicking the "English"/"Danish" link, you'd end up on a page for editing the webpage contents, which is great in itself.

### DIFF
--- a/src/chrome/content/rules/Energinet.dk.xml
+++ b/src/chrome/content/rules/Energinet.dk.xml
@@ -1,6 +1,34 @@
-<ruleset name="Energinet.net">
+<!--
+	Problematic domains:
+
+		- ^ ¹
+
+	¹: Mismatch
+-->
+<ruleset name="Energinet.dk">
+
 	<target host="energinet.dk" />
 	<target host="www.energinet.dk" />
-	<securecookie host="^www\.energinet\.dk" name=".*"/>
-	<rule from="^http://(?:www\.)?energinet\.dk/" to="https://www.energinet.dk/" />
+
+	<!-- Causes redirect loops -->
+	<exclusion pattern="^http://(www\.)?energinet\.dk/(DA|EN)/" />
+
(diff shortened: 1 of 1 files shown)


jeremyn:
I see the site editing thing, ouch.

- [ ] `<exclusion pattern="^http://(www\.)?energinet\.dk/(DA|EN)/" />` is really broad since it affects almost all the URLs a user would actually explicitly visit. So the ruleset should `(partial)`.

fuglede:
Right, but it still secures all the content that is not visited explicitly, including a lot of static content. Regarding adding the `(partial)`, I don't think that that's something that has been done very systematically, and in particular it is not always clear what exactly it means.

jeremyn:
I agree the broad exclusion is fine.

In another issue or pull request (I don't know which one), while discussing `(partial)` with @J0WI , I agreed (?) with them that `(partial)` is appropriate when you're on a site, a ruleset is active, but HTTPS isn't happening in some obvious way. In this case a user is at `www.energinet.dk`, sees the `Energinet.dk` ruleset active, but the URL in the address bar is `http://`. So `(partial)` should be used here, if we use it anywhere.

jeremyn:
To put it another way: a user might be confused that a ruleset is active but the URL they're on is `http://`, in which case `(partial)` reassures the user that that behavior is expected.

fuglede:
Updated it to include "(partial)"

jeremyn:
Great, thanks. Merged.

Issue #6791, opened 2016-09-07
infogr.am (partial) breaks graph


terrorist96:
http://www.standard.net/Police-Fire/2016/08/21/Utah-concealed-weapons-permits-now-numbering-663K-see-surge-in-popularity

terrorist96:
http://www.theage.com.au/interactive/2016/gun-city/day1.html

terrorist96:
@fuglede 
https://github.com/EFForg/https-everywhere/pull/4503#issuecomment-202535289

> Thanks. If it turns out to be problematic in more cases, I'll throw in my vote to simply disable the domain.

fuglede:
Removed the subdomain from the ruleset in #6903.

PR #6903, opened 2016-09-17
[Infogr.am] Disable subdomain e.infogr.am

fuglede:
This particular subdomain has caused issues a number of places, and the most recent ones do not suggest an easy fix. This closes https://github.com/EFForg/https-everywhere/issues/6791.

### DIFF
--- a/src/chrome/content/rules/Infogr.am.xml
+++ b/src/chrome/content/rules/Infogr.am.xml
@@ -22,6 +22,14 @@
 	<target host="blog.infogr.am" />
 	<target host="www.infogr.am" />
 
+
+	<!--	Not secured by server:
+					-->
+	<!--securecookie host="^\.infogr\.am$" name="^ig_session$" /-->
+
+	<securecookie host="^(?!e\.infogr\.am$)." name="." />
+
+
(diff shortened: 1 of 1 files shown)


jeremyn:
- [ ] Did you mean to remove the `securecookie`? The `host` was for everything except `e.infogr.am`.

fuglede:
Good catch; I was a bit too fast there.

jeremyn:
You can change it to just `<securecookie host=".+" name=".+" />`, of course there is no need to refer to `e.infogr.am` any more.

fuglede:
Looks like you're right. I was wondering if this was there to somehow prevent infogr.am from settings cookies on the subdomain specifically, but that's not possible anyway. Updated.

jeremyn:
Do you mean the problem of setting a cookie on `.infogr.am` that now can't be read on `e.infogr.am`? The only way we can fix that is to not use `securecookie` at all. Both options are equally good/bad.

jeremyn:
Should I just merge this, then?

fuglede:
Yep, no further changes from here.

jeremyn:
Done.

PR #6904, opened 2016-09-18
Update Qbox.me.xml

renyouguo:


src/chrome/content/rules/GeekPark.xml
```diff
@@ -1,10 +1,10 @@
 <!--
-
-	dn-geekpark-new.qbox.me belongs www.geekpark.net , which mismatch https.
-
 	Warn：https://github.com/EFForg/https-everywhere/pull/5042#issuecomment-234782358
-
+	
+	Related:
+		Qbox.me.xml
```
J0WI: Please also add a reference to GeekPark.xml on Qbox.me.xml

src/chrome/content/rules/Qbox.me.xml
```diff
@@ -1,6 +1,9 @@
 <ruleset name="qbox.me">
 
-	<!--	MCB:	-->
+<!--
+	Related : Qbox.me.xml
```
J0WI: should be `GeekPark.xml`

PR #6889, opened 2016-09-16
Update Tencent_CN.xml

ivysrono:

PR #6905, opened 2016-09-18
Update Cloudfront.xml

renyouguo:


jeremyn:
It looks like http://aww.bugsplatsoftware.com/Blog/ makes calls to http://d22iv91mzdv9l4.cloudfront.net , which doesn't exist in DNS. Is the point of this pull request to fix their page for them by rewriting http://d22iv91mzdv9l4.cloudfront.net to https://www.bugsplat.com ? If so, that is not something we should be doing in HTTPS Everywhere.

renyouguo:
I think it is http to https...
Why?
Does #6889 same?

renyouguo:
No, iwgc.cn is not CDN.

jeremyn:
In that case, what we're discussing here may not apply to pull request #6889.

renyouguo:
All the *.bugsplatsoftware.com show:
`Error establishing a database connection`
Thank you, @jeremyn

PR #6908, opened 2016-09-18
[KeyDrive] Remove redundant ruleset and update another one

fuglede:
In particular, this closes https://trac.torproject.org/projects/tor/ticket/19252.


jeremyn:
Side question to you, @Hainish , and @J0WI : I think we should create a new label for issues and pull requests that have also been reported at https://trac.torproject.org . I consider these to have slightly higher priority. What do you all think?

jeremyn:
Note as `Refused`:
- [ ] https://domains.snapnames.com

Note as `Timeout`:
- [ ] https://domainauctions.moniker.com
- [ ] https://help.moniker.com

Add target:
- [ ] https://login.moniker.com
- [ ] https://maintenance.snapnames.com

Hainish:
@jeremyn for issues reported on trac, if we decide it is a high priority we should clone the issue here and label it as "high priority" on a case-by-case basis.  Not every issue on Tor's trac will be high priority, so there's no reason to label them as such.  We should however always link back to the trac issue.

fuglede:
Agreed on all of those; updated.

jeremyn:
@Hainish I don't mean we should label all trac.torproject.org problems as "High Priority". I'm suggesting that we label all such problems as something like "Tor Project" so that reviewers such as myself who give slightly higher priority to things affecting Tor can search for those issues if they want to. It would function similar to the Alexa labels.

Hainish:
@jeremyn many tickets from the Tor trac are simple ruleset issues.  It seems silly to add a label like `Tor Project` to these and prioritize them just because they came from the Tor trac.

Ex: 
https://trac.torproject.org/projects/tor/ticket/15561
https://trac.torproject.org/projects/tor/ticket/15490

Hainish:
It's also unclear to me how automated labels would work here since it is not clear to me that Tors trac has an API we can query.

Note that I'm not against having a label like `Tor Browser` for issues that are pertinent.  I'm just against automatically assigning this label to *any* issue that comes from Tor trac.

jeremyn:
There are about 100 issues and 300 pull requests that are completely unlabeled. We are, if anything, significantly underusing labels.

jeremyn:
Also, I don't think we should do automated `Tor trac` labels. I mentioned the Alexa labels only as examples of labels that provide optional information to those who are interested in it.

Hainish:
How does labelling any issue that comes from Tor trac as `Tor Browser`, even if it's a ruleset change, make anything clearer?  We need to manually move issues over in order for us to apply labels anyway, and if we're already doing that, taking the few seconds to discern if it's a real issue with `Tor Browser` rather than applying it regardless of the issue contents constitutes minimal overhead.

jeremyn:
I'm thinking more about pull requests than issues. Suppose this exact pull request had been submitted months ago, and not by @fuglede but instead by a one-time contributor. Without a label it would be almost invisible in the backlog. (And yes, this would have required a reviewer to have previously spotted the issue and marked it. We need to do better about labeling generally.)

Hainish:
@jeremyn I see what you're saying - as I said previously I think it's fine to create a new label like `Tor Browser` and apply it to pertinent issues.  But a ruleset PR shouldn't be labelled `Tor Browser` or even `Tor` just because it fixes an issue that was listed on the Tor trac.

jeremyn:
I understand your concern: saying an issue is associated with Tor might be misleading for a generic ruleset update. 

All the same I feel there should be a label-based way to identify issues "like this one", though what that means is not well-defined. Maybe what would be more helpful is to have a label that we use for pull requests that fix issues (trac or GitHub) to distinguish them from pull requests that add new rulesets or enhancements. This seems extra important for pull requests that fix Tor trac issues since those issues are not otherwise discoverable in GitHub.

Hainish:
@J0WI @fuglede what do you think about the above idea?  Would this help your workflow?

J0WI:
I think we already have the 'ruleset-bug', 'new-ruleset' and 'ruleset-enhancement' labels?
Of course it would be great if they get added automatically. I used to add them after having a first look on the changes, but since GitHub implemented the review feature there might be better ways to sort them out of my 'To-Do list'.

fuglede:
Probably not really. I think most GH/trac issue fixing PRs note that they fix issues in their titles or descriptions (and if they didn't, then we also would not be able to put a tag on it, so we should be good in either case).

PR #6919, opened 2016-09-20
Fix rules test by copying rulesets.json in extension make scripts rat…

Hainish:
…her than in validate.sh


Hainish:
Merging this in myself since the `rules` test is currently broken and this is a relatively simple fix.

Issue #4241, opened 2016-02-29
In Firefox Nightly, "NS_ERROR_XPC_NOT_ENOUGH_ARGS: Not enough arguments [nsICookieManager2.remove] HTTPS.js:164:0"


dholbert:
Not sure about exact STR, but in Firefox Nightly 47 with HTTPS Everywhere 5.1.4, I saw this error in my Browser Console (ctrl shift J):

> NS_ERROR_XPC_NOT_ENOUGH_ARGS: Not enough arguments [nsICookieManager2.remove] HTTPS.js:164:0

This points to the following source code in HTTPS Everywhere's HTTPS.js source file:

```
  handleInsecureCookie: function(c) {
      [...]
      cookieManager.remove(c.host, c.name, c.path, false);
```

The error happens because the `remove` function-signature was modified recently, here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1245184
http://hg.mozilla.org/mozilla-central/diff/ab3a305b3e41/netwerk/cookie/nsICookieManager.idl#l47

A new parameter "aOriginAttributes" has been added.

dholbert:
(CCing @bakulf who made this change in Gecko.)

bakulf:
the new method wants an additional param. Change that line to:

cookieManager.remove(c.host, c.name, c.path, c.originAttributes, false);

dholbert:
(Of course, that change on its own would make HTTPS Everywhere stop working in earlier versions of Firefox. So really, I suspect they might want to do this in a try{}catch{}, or something else to allow for both versions of the method.)

Vednier:
Right now Nightly (Firefox 50) throws a lot of errors like this in console
“nsICookieManager.remove()” is changed. Update your code and pass the correct originAttributes. Read more on MDN: https://developer.mozilla.org/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManagerHTTPS.js:164:6
“nsICookieManager2.add()” is changed. Update your code and pass the correct originAttributes. Read more on MDN: https://developer.mozilla.org/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager2HTTPS.js:165:6

Vednier:
> the new method wants an additional param. Change that line to:
cookieManager.remove(c.host, c.name, c.path, c.originAttributes, false);

No, this doesnt work, changing that in HTTPS.js only cause this error to appear.
[...]

PR #6920, opened 2016-09-20
nsICookieManager and OriginAttributes: fix issue #4241

bakulf:


jeremyn:
@Hainish Can you review this one, please?

Hainish:
Reviewing now

Hainish:
@bakulf currently a large portion of our users are on Tor Browser, the current version of which is derived from Firefox ESR (45.4.0esr).  Tor Browser >= 6.0 is using Firefox ESR >= 45.1.1esr.  It looks like the changes [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1259169) have not been backported to ESR, is that correct?

Hainish:
I believe this PR would break for our TB users.

bakulf:
You are right. If we want to be compatible with ESR, we cannot land this patch.
In this case, I suggest to keep the code as it is.
Alternatively we could add checks as: if ("originAttributes" in cookie) { cm.add(..., {});

PR #6922, opened 2016-09-20
removed Tony.xyz.xml

J0WI:
Host redirects to tonywebster.com which is HSTS preloaded.


jeremyn:
Also https://tony.xyz and https://www.tony.xyz time out.

Merged.

PR #6925, opened 2016-09-20
remove Truecrypt.xml

J0WI:
Truecrypt is dead.


jeremyn:
Well, http://www.truecrypt.org redirects to http://truecrypt.sourceforge.net , but HTTPS fails so we wouldn't have a ruleset for it anyway.

Merged.

PR #4178, opened 2016-02-19
Expand #4165

J0WI:
Thanks to @mathiasbynens for the previous commits

PR #6938, opened 2016-09-20
update Opera rules

J0WI:
This should also close my previous PR (#4178)


J0WI:
@jeremyn @gloomy-ghost I would really like to get this merged before I have to solve conflicts again. Could you have a look?

J0WI:
#10882

J0WI:
@jeremyn @Hainish let's merge this before they change something again.

jeremyn:
`Opera.xml`:

- [ ] What is the [`(404)`](https://github.com/J0WI/https-everywhere/blob/a890c9c56c5169d4c66cedb56062152fdae39c65/src/chrome/content/rules/Opera.xml#L74) next to https://in.opera.com in the top comment about? I just get a simple refused/dropped for that URL.
- [ ] What does [`Tests times out`](https://github.com/J0WI/https-everywhere/blob/a890c9c56c5169d4c66cedb56062152fdae39c65/src/chrome/content/rules/Opera.xml#L141-L143) for `bugfiles.operasoft.com` mean? I get a `303` to https://bugs.opera.com/secure/attachment/ which then gets a `400`.
- [ ] All of the [`/home test`s](https://github.com/J0WI/https-everywhere/blob/a890c9c56c5169d4c66cedb56062152fdae39c65/src/chrome/content/rules/Opera.xml#L201-L213) like http://ar.opera.com/home redirect to a `404`. I'm not really sure what we're testing here but that's not right. Maybe we can replace `/home` with `/index.htm` like in the later `test`s?
- [ ] We don't need both `?foo` and `?bar` `test`s [here](https://github.com/J0WI/https-everywhere/blob/a890c9c56c5169d4c66cedb56062152fdae39c65/src/chrome/content/rules/Opera.xml#L218-L230). The purpose of these tests is already confusing enough. Please remove all the `?bar` `test`s.

`Operamediaworks.com.xml`:

- [ ] Note https://blog.operamediaworks.com as `Time out`

`Operasoftware.com.xml`:

(This is fine, I'm just noting it here for completeness.)

`operacdn.com.xml`:

- [ ] I don't think we should redirect `download2` to `download3` because this assumes they are always identical and we don't know that for sure. On the other hand the last modified date for the `ftp/` directory at http://download3.operacdn.com is `06-Jun-2007 14:47` but at the HTTPS version https://download3.operacdn.com it's `08-Mar-2016 13:47`, so maybe we can't trust that HTTP and HTTPS are the same either. (I still think we should redirect 1, 3, and 4 though.)
- [ ] I know you didn't change it but the test URL http://www-static.operacdn.com/static-heap/f1/f1e88ec9b434964e7835dde0c9d36eb4889284c5/winphone-store-en.png ([here](https://github.com/J0WI/https-everywhere/blob/a890c9c56c5169d4c66cedb56062152fdae39c65/src/chrome/content/rules/operacdn.com.xml#L16)) returns a `404` so it's not testing anything more than the implicit `test` for the `target` does. So we can remove it, or replace the `test` URL if you can find one.

Also:

- [ ] We should standardize the ruleset `name`s and filenames. I propose that the filenames should be `Opera.com.xml`, `OperaMediaworks.com.xml`, `OperaSoftware.com.xml`, and `OperaCDN.com.xml`. The ruleset `name`s should be the same as the filenames but without the `.xml` (`Opera.com` etc). What do you think?

J0WI:
> Please remove all the ?bar tests.

They are required because of the complex regex. Otherwise we don't have enough test urls.

> I don't think we should redirect download2 to download3 because this assumes they are always identical and we don't know that for sure.

I compared them and they are very likely identical. I really want to keep all the downoads hosts, otherwise the software is downloaded over plain http. I added some test urls for them.

> We should standardize the ruleset names and filenames.

We have no guideline about this and your proposal doesn't follow any rule either (camel case?). So I'm just using the domain name now.

jeremyn:
We can keep the `?bar` `test`s to meet testing requirements. That `rule` has an unfortunately complicated regex.

I'm okay with keeping the `download2` to `download3` redirect if you really want to keep it.

I know we don't standardize ruleset names/filenames across the project but we should make some effort to standardize them at least within a group of related files like "stuff related to Opera". If you prefer a single capital letter at the beginning and the rest lowercase, that's fine with me. Please change the `ruleset` `name` in `Opera.com.xml` to `Opera.com (partial)`.

Please add a comment when you're done making changes and this is ready for me to review again.

J0WI:
Done.

jeremyn:
Merged, thanks.

J0WI:
Thanks!

PR #6939, opened 2016-09-20
[Migros] separate and add hosts

J0WI:


J0WI:
@jeremyn (pinging you, since you did the last review for 'migros'):
I have about 8 more files, but I can't open parallel PR's without conflicts in the 'Other rulesets' section of migros.xml.
How would you prefer to review them?

jeremyn:
Do whatever works best for you.

J0WI:
Commited everything into this branch.

jeremyn:
`Chocolatfrey.ch.xml`:
- [ ] Are [these comments](https://github.com/J0WI/https-everywhere/blob/444d7d6264b4fe032a0465619dac1040cf8c3ce5/src/chrome/content/rules/Chocolatfrey.ch.xml#L8-L9) a mistake?

`Doitgarden.ch.xml`:
- [ ] https://doitgarden.ch is both a mismatch and an HTTP redirect

`Micasa.ch.xml`:
- [ ] Try to capture more of http://www.micasa.ch, something like `(js|medias|static)` etc

`Migrol.ch.xml`:
- [ ] Add `securecookie` for `b2b.migrol.ch`
- [ ] Add target for http://websvc.migrol.ch

`Sportxx.ch.xml`:
- [ ] https://sportxx.ch is both a mismatch and an HTTP redirect
- [ ] Try to capture more of http://www.sportxx.ch, something like `(de/mconnect|dynaTraceMonitor)` etc

`migros.xml`:
- [ ] `melectronics.ch` is included in `migros.xml` but `vacando.ch` has its own file even though the behavior is the same, and `migipedia.ch` etc. Should these domains in `migros.xml` get their own files?
- [ ] Filename casing should be consistent. Capitalize the first letter of `migros.xml`, `hotelplan.xml`, and `migros-bank.xml` including in comments in every file
- [ ] If you move enough domains out of `Migros.xml`, then rename it to `Migros.ch.xml`
- [ ] Re-sort `Other Migros rulesets` after changing casing

J0WI:
chocolatfrey.com is part of chocolatfrey.ch, that's fine.
I addressed your feedback and moved more hosts from migros.xml to their own file. I didn't rename it to Migros.ch.xml because it has still not only *.migros.ch hosts in it. Namely there are the shopping sites, the host that redirects to community.migros.ch and the login server, that is used by all sites.

jeremyn:
[These](https://github.com/J0WI/https-everywhere/blob/a08fca055581786a30c090d67af7aa3ad88971c2/src/chrome/content/rules/Doitgarden.ch.xml#L5-L6) doitgarden errors are backwards. It's https://doitgarden.ch that has both problems, `www` is not mismatched.

For `micasa` and `sportxx`, negative exclusions are [not good](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md), see also https://github.com/EFForg/https-everywhere/issues/6985 . Can you rewrite these with positive rules?

`ressources` should be `resources` (appears twice).

I get `Secure Connection Failed` for https://archiv.migros-kulturprozent.ch , not a redirect.

There is both `M-fanshop.ch.xml` and `M-Fanshop.ch.xml`. For consistency I think all filenames in this pull request should have only the first letter capitalized, which also means renaming `Migros-Bank.xml` to `Migros-bank.xml`.

J0WI:
> For micasa and sportxx, negative exclusions are not good, see also #6985 . Can you rewrite these with positive rules?

I thought this is the current practice (see ?!wp-content exclusions). Otherwise I need a complex regex to rewrite static resources and then a second regex rule for the trivial cdn redirects.

jeremyn:
The style guide says to avoid negative lookaheads and calls out negative lookaheads in exclusions as particularly confusing. So yes, you'd have something like these two rules:

```xml
<rule from="^http://www\.micasa\.ch/(.+)\.(css|jpg|js|json|png|svg|woff)" to="^https://www.micasa.ch/$1.$2" />
<rule from="^http://(cdn[1-4])\.micasa\.ch/" to="^https://$1.micasa.ch/" />
```

jeremyn:
Also, I've unchecked the casing item in the checklist to make sure that problem doesn't get lost, particularly referring to my comment about it in https://github.com/EFForg/https-everywhere/pull/6939#issuecomment-251990028

J0WI:
done.

jeremyn:
I think you can get rid of the `exclusion`s in the Micasa and Sportxx rulesets, because they are not needed with the targeted `rule`s, and also you can then make the `test` indentation line up.

You did a different filename casing change than I asked for, for example you changed `Migros-ferien.ch.xml` to `Migros-Ferien.ch.xml` instead of `Migros-Bank.xml` to `Migros-bank.xml`. Do you feel strongly that the names should have multiple capital letters? If you do that's fine, you can leave it, but definitely `M-fanshop.ch.xml`/`M-Fanshop.ch.xml` should be consistent with themselves and the other filenames.

J0WI:
The exclusion for `$` is needed, because no rule matches and Travis use the targets as test urls.

I choose capital letters to match the site name.

jeremyn:
That's fine. Merged.

PR #6940, opened 2016-09-20
[obi.xml] add obi.ch

J0WI:


jeremyn:
https://obi.ch and https://www.obi.ch redirect forever. You could update [the comment](https://github.com/J0WI/https-everywhere/blob/b5fb683a61ec5f84930dfa382749cccd21655aaf/src/chrome/content/rules/obi.xml#L7) but we can't add them as simple targets.

J0WI:
I can confirm redirect issues for `obi.ch`, but `www.obi.ch` works fine for me and is https by default.

jeremyn:
https://www.obi.ch works fine for me now. Maybe I had something cached from https://obi.ch that was breaking it.

Also I noticed another subdomain https://ot.obi.ch in `Firefox Dev Tools > Network` that looks like a valid target.

PR #6947, opened 2016-09-21
update X2Go.org.xml 

luciancor:


J0WI:
This rule is default off because of cert issues anyway, so this change is not really helpful.
Also the cert is valid until 01.04.2034, but it's not trusted and mismatched.

luciancor:
clear - closing this

PR #6949, opened 2016-09-21
update Constant-Contact.xml 

luciancor:


J0WI:
They issued a new cert on Fri, 16 Sep 2016 12:41:02 UTC, so everything works fine for me. Can you confirm?

luciancor:
@J0WI yes - it seems fixed

PR #6950, opened 2016-09-21
remove flattr.net

luciancor:


J0WI:
This works as expected for me, why has it been removed?

jeremyn:
Thanks for the update.

https://static.flattr.net has valid HTTPS so I don't think `flattr.net` should be removed entirely.

Did Flattr document that they are shutting down their `.net` domains?

luciancor:
thanks for the update guys! 
so should we only remove : 	<target host="flattr.net" />	and <target host="www.flattr.net" />
in such cases? those have invalid certificates

J0WI:
```
	<rule from="^http://(?:www\.)?flattr\.net/"
		to="https://flattr.com/" />
```
They get redirected to .com, so everything works fine here.

luciancor:
got it - closing this. Thanks for the feedback!

PR #6951, opened 2016-09-21
update GitCafe.xml 

luciancor:


J0WI:
No targets left here. I also can't connect to this host, so I would recommend to remove it.

J0WI:
@luciancor would you like to update your PR to remove this rule completely?

luciancor:
@J0WI done!

PR #6952, opened 2016-09-21
update Hackable-devices.org.xml 

luciancor:


J0WI:
Same here: no targets left, please remove or default_off.
There are also wiki and list subdomains, but they also don't support https, so removing is fine.

luciancor:
@J0WI thx for feedback! updated this PR - please take a look

J0WI:
Thanks!

PR #6954, opened 2016-09-21
Update NHS.xml

tamoor-cliqz:


gloomy-ghost:
Maybe [L609](https://github.com/EFForg/https-everywhere/pull/6954/files#diff-5e1b0fde28a4b13114aca7562d4da3f5R609) could be changed to 
```
<rule from="^http://kentsexualhealth\.nhs\.uk/"
 		to="https://www.kent.gov.uk/sexualhealth/" />
```

tamoor-cliqz:
Agreed! we will update the PR

J0WI:
ping @tamoor-cliqz
cc @luciancor

luciancor:
@gloomy-ghost not sure if I understood correctly your comment - can you please check the last commit?

gloomy-ghost:
The target for kentsexualhealth.nhs.uk shouldn't be commented as we still need to rewrite the requests.

Also it looks like kentsexualhealth.nhs.uk is now redirecting to http://www.kent.gov.uk/social-care-and-health/health/sexual-health, can you update it?

luciancor:
@gloomy-ghost still not convinced that I understood correctly  but please take another look at my last commit.

src/chrome/content/rules/NHS.xml
```diff
@@ -607,8 +606,8 @@
 	<!--	Redirect drops forward slash
 		and path, but not args:
 						-->
-	<rule from="^http://kentsexualhealth\.nhs\.uk/"
-		to="https://www.kent.gov.uk/sexualhealth" />
+	<rule from="^http://kentsexualhealth\.nhs\.uk/[^?]*"
```
J0WI: Do you mean `^http://kentsexualhealth\.nhs\.uk/$`?

jeremyn:
Without HTTPS Everywhere, http://kentsexualhealth.nhs.uk directly redirects to http://www.kent.gov.uk/sexualhealth dropping the path (e.g. `/home.htm`) and keeping the query string, which I guess is what the comment above the `rule` means by "args".

With HTTPS Everywhere and the [original, existing rule](https://github.com/EFForg/https-everywhere/blob/7e4193f45aeb4633e79cdfa56abe1178bf746b39/src/chrome/content/rules/NHS.xml#L638-L639), http://kentsexualhealth.nhs.uk is rewritten to https://www.kent.gov.uk/sexualhealth dropping the path and keeping the query string.

So it looks like the original, existing HTTPS Everywhere rule is doing the correct thing and we don't need to change it, which means we don't need this PR.

@gloomy-ghost @J0WI Can one of you please double-check this? If I'm right, then we can close this PR.

J0WI:
I can confirm this behavior. So this is kind of a snapping redirect issue.
Another problematic I just discovered is that [Kent.gov.uk.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Kent.gov.uk.xml) is disabled.

jeremyn:
@J0WI So can we close this PR?

J0WI:
I think so.

jeremyn:
Thanks.

PR #6960, opened 2016-09-21
Update Gtimg.com.xml

renyouguo:


jeremyn:
Thanks. Please:

- [ ] Remove the [`Directly`](https://github.com/renyouguo/https-everywhere/blob/f8f290ae7ab6bda667049122c25d2e6db626cf3c/src/chrome/content/rules/Gtimg.com.xml#L62) comment, because we don't need it
- [ ] Move the new targets between `img2` and `wximg`
- [ ] Optional -- Move the existing `mat1` target and tests between the new `map` targets and `wximg`

jeremyn:
Thanks, merged.

PR #6962, opened 2016-09-21
Updated fronter.com rule to reflect full (rather than partial) HTTPS support

yitzhaq:
Every subdomain under fronter.com now supports (and redirects to) HTTPS, so I've removed some conditionals, and replaced them with a wildcard entry. At the same time I've added some explicit test URLs to ensure coverage of what's behind the wildcard, and also removed conditionals for securecookie.

The root of the bottom two test URLs only return 403, but I left them in in case you wanted >3. If testing against a 403 is no good (the redirect still works fine, though), feel free to get rid of them.

I've tried following the style guide, and hope the result is to your satisfaction.

There is one remaining issue, as also noted (presumably) by the rule checker, namely webfronter.us.fronter.com. While this does listen on HTTPS, the returned certificate is mismatched (it matches *.fronter.com but not *.us.fronter.com). I'm not entirely sure how to best handle nor express this in the rule - suggestions are most welcome.

This subdomain is only used to handle a (temporary) redirect from webfronter.us.fronter.com/$1 to webfronter.com/$1, and the target domain _does_ support HTTPS with a matched certificate (although it does not itself redirect visitors to HTTPS; I'll submit a separate rule for that domain).

I expect webfronter.us.fronter.com to become deprecated in a few months time.

FWIW I also fixed lists.fronter.com earlier this week, which now has a new, valid certificate. I have therefore also removed the comment concerning that.

Thanks.

### DIFF
--- a/src/chrome/content/rules/fronter.com.xml
+++ b/src/chrome/content/rules/fronter.com.xml
@@ -5,32 +5,19 @@
 
 	ʳ Refused
 
-
-	Problematic hosts in *fronter.com:
-
-		- lists ᵉ ˢ
-
-	ᵉ Expired
-	ˢ Self-signed
-
 -->
-<ruleset name="Fronter.com (partial)">
+<ruleset name="fronter.com">
 
 	<target host="fronter.com" />
-	<target host="brasil.fronter.com" />
-	<target host="cdn.fronter.com" />
-	<target host="eng.fronter.com" />
-	<target host="help.fronter.com" />
-	<target host="nor.fronter.com" />
-	<target host="login.fronter.com" />
-	<target host="pol.fronter.com" />
-	<target host="swe.fronter.com" />
-	<target host="www.fronter.com" />
+	<target host="*.fronter.com" />
 
+	<test url="http://login.fronter.com/" />
+	<test url="http://bucket.fronter.com/" />
+	<test url="http://api.fronter.com/" />
+	<test url="http://cdn.fronter.com/" />
+	<test url="http://notifications.fronter.com/" />
 
-	<securecookie host="^\." name="^(?:__cfduid|cf_clearance)$" />
-	<securecookie host="^\w" name="." />
-
+	<securecookie host=".+" name=".+" />
 
 	<rule from="^http:"
 		to="https:" />
--- a/src/chrome/content/rules/webfronter.com.xml
+++ b/src/chrome/content/rules/webfronter.com.xml
@@ -1,0 +1,6 @@
+<ruleset name="webfronter.com">
+	<target host="webfronter.com" />
+	<target host="www.webfronter.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


jeremyn:
Thanks for the ruleset. I haven't reviewed it closely yet, but wildcard `target`s are discouraged by the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md). Please rewrite `fronter.com.xml` to use explicit `target`s. The `Sublist3r` domain finder tool, also described in the style guide, shows no more than about 60 domains, which is a lot but still what I consider a reasonable number to list explicitly.

yitzhaq:
Sorry for taking forever to follow up on this. I am no longer affiliated with this domain, but after having worked on it for several years, and investing a lot of time into ensuring everything is encrypted everywhere, I can say with some authority that there are zero subdomains without HTTPS, nor will any such come into being.

As such, it is my personal opinion that this should be an acceptable exception to the style guide, considering any explicit list is bound to go out of date at some point in the future, and as such might miss records that do in fact support HTTPS.

Thanks for your consideration.

jeremyn:
@yitzhaq Are you still interested in working on this pull request?

yitzhaq:
@jeremyn As mentioned in my previous comment from September, I don't really see that it requires more work. I'd be happy to see it through to merging, however, I do still disagree with making an explicit list, bound to go out of date, when I have personal knowledge that the domain has no unencrypted subdomains, now or in the future.

If you are not interested in merging it in its current form, @kozejonaz is still involved with the domain, and might be willing to spend time on it.

jeremyn:
@yitzhaq For various reasons including security and maintainability we avoid relying on personal knowledge from contributors regarding rulesets and domains.

The technical way to commit to "there are zero subdomains without HTTPS, nor will any such come into being" is to add the top-level domain to the HSTS preload list at https://hstspreload.org. We exclude these domains from our rulesets. So, the only reason to make a ruleset for HTTPS Everywhere is if a domain is not in the HSTS preload list and is, therefore, not quite ready to make that commitment for whatever reason.

In any case, if you want to get someone else to work on the ruleset for this site, then please ask them to do so in a new pull request, and close this one.

yitzhaq:
I agree with the principle you describe. However;

Switching this domain and associated resources over to HTTPS was a gradual process, over a period of several years, one service and subdomain at a time. The aforementioned knowledge stems from myself having been the person to painstakingly upgrade each and every single part of this domain to achieve HTTPS Everywhere. I stand by the statement that the fronter.com domain is fully encrypted, through and through.

HSTS preload inclusion is still high on the want list, but it is currently ineligible because there are still certain redirects from fronter.com (on HTTPS) to HTTP on other domains. That is to say, all HTTP requests against any part of fronter.com will get rewritten to HTTPS, but the objectionable requests in question redirect (with HSTS) from an HTTPS fronter.com endpoint to an external HTTP endpoint on a third party domain. In short; The bad (unencrypted) parts exist outside the realm of this ruleset, yet in the context of preload inclusion, it is sufficient grounds for inadmissibility. This, to my mind, would constitute the "whatever reason" you refer to.

It is my understanding that your project's focus is contained within each individual domain ruleset. As such, I don't believe the last few remaining objections against preload inclusion should also need to be objections against wildcard inclusion in HTTPS Everywhere.

So I do still uphold, given the present state of facts, until the domain is eligible also for preload inclusion, that my proposed changes would improve upon the current state of this particular domain ruleset, even if it does happen to make an exception to the style guide. As you say, once preload inclusion is achieved, the ruleset can be removed altogether.
But until that point, I believe this is the best approach, with full coverage and better test URLs.

yitzhaq:
A final, semi-related note: The HSTS headers already include Preload, in preparation for future inclusion. This has (with some undeniable futility) been the case for well over a year. I still expect that, once currently blocking external factors evolve, preload inclusion will be actively pursued.

jeremyn:
@yitzhaq I agree with you that given some domain where all subdomains support HTTPS and any future subdomains will too, that a wildcard is better for that domain. However in practice we have no way of recognizing these domains for certain, so as I said for security and maintainability we prefer not to use wildcards. PR https://github.com/EFForg/https-everywhere/pull/4300, starting at https://github.com/EFForg/https-everywhere/pull/4300#issuecomment-274556441, has a related discussion about wildcards you might find interesting.

Anyway wildcard use is well-established by now and there is nothing particularly unique about this domain to justify using a wildcard here, though I know you feel differently. Though I respect your position, I don't really want to spend more time discussing it.

zoracon:
Closing due to no revisions made. Can reopen if interest for revision is indicated

yitzhaq:
As previously mentioned I am no longer affiliated with fronter.com, and by now I no longer know anyone who is. I stand by my previous comments, still believe the proposed update is an improvement on the current ruleset for this domain, and still believe the use of wildcard under the circumstances specific to this domain is fully justified, in spite of what the style guide may state in general terms.

I have no interest in doing any further work on this, and would discourage anyone to follow the proposed steps to do so, as any result gained from it would be inferior to the result gained from the ruleset I originally proposed. In summary, any ruleset for this domain that does not make use of a wildcard is bound to be either incomplete or likely to become quickly outdated, and would thus be a mostly wasted effort with a lesser gain.

On a personal note, I found this experience discouraging, and will think twice before contributing to the project again, which I frankly think is a shame. I fully believe in the underlying spirit and idea of the project, and have profound respect for the EFF, but disagree with this custodianship.

zoracon:
@yitzhaq Our current ruleset maintainers are working hard and I am currently trying to assist them the best way I can with looking through old PRs since I came on to the project. I realize there may have been some disconnect with feedback and I assure you the goal is not to discourage people from contributing. I take note and responsibility for any negativity you may have experienced. 

We do have updated docs on ruleset parameters and I hope they guide better process in the future.

PR #6963, opened 2016-09-21
Heise.de:remove all exceptions

numismatika:
As it looks like the rewrites to plain http are gone all over the page, at least for now.
There are still traces of mixed-content left so we need to make sure that this is a) permanently enabled and b) neither ads nor mixed content breaks the page


J0WI:
Sounds great! So far there is no official statement of them, as they did for `/security/`, but let's see.
Do not forgot to remove the `(partial)` flag and there are currently some failing tests.

numismatika:
There are subdomains like tarifrechner or rl that are not covered yet, so the partial flag is still valid.
They probably are still experimenting as i have seen lets encrypt certificates on some domains.
I doubt they will make an aannouncement until mixed content issues are gone.

src/chrome/content/rules/Heise.de.xml
```diff
@@ -78,23 +63,10 @@
...
 			<test url="http://www.heise.de/video/latest_vids.xml" />
 
-
-	<rule from="^http://m\.heise\.de/(js/|icons/|[ai]vw-bin/|security/icons/|stil/|fonts/)" to="https://m.heise.de/$1"/>
-		<exclusion pattern="^http://m\.heise\.de/$" />
-
-	<rule from="^http://(abo|piwik\.shop|prophet|shop)\.heise\.de/" to="https://$1.heise.de/"/>
-
-	<rule from="^http://www\.shop\.heise\.de/"
+	<rule from="^http://(www\.)?heiseshop\.de/"
```
J0WI: Please document this issue.

src/chrome/content/rules/Heise.de.xml
```diff
@@ -1,46 +1,31 @@
...
-	<target host="piwik.shop.heise.de" />
+	<target host="stat.heiseshop.de" />
+	<target host="www.heise-gruppe.de" />
+	<target host="www.heise-medien.de" />
 	<target host="www.heise.de" />
-
-	<!--	Complications:
-				-->
+	<target host="www.heiseshop.de" />
 	<target host="www.shop.heise.de" />
```
J0WI: Please also group the targets by host, not only alphabetically.

J0WI:
See also:
gutscheine.heise.de
jobs.heise.de (mismatch)

I think as long as the major part of the host works, it's safe to remove the partial flag.

src/chrome/content/rules/Heise.de.xml
J0WI: @numismatika rule lgtm, could you solve this one before I merge?

numismatika: I reordered the target hosts.

J0WI:
For ct there are also:
ho.ct.de
blog.ct.de
radio.ct.de
www.radio.ct.de
ctnotwin15.ct.de

Issue #6875, opened 2016-09-14
learn.vmware.com doesn't deliver a valid certificate


KommX:
see title

we shouldn't redirect to a certificate warning

jeremyn:
Thanks for the report.

[Here](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/VMware.xml#L71) is the target. The fix is to add an exclusion for `learn` [here](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/VMware.xml#L73) and a corresponding test between [these two lines](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/VMware.xml#L78-L79).

This issue is a good example of the problem with wildcards.

PR #6971, opened 2016-09-22
Update VMware.xml

gloomy-ghost:
Add exclusion for learn
Fix: #6875

### DIFF
--- a/src/chrome/content/rules/VMware.xml
+++ b/src/chrome/content/rules/VMware.xml
@@ -31,13 +31,13 @@
 
 		- ^ ¹
 		- info ²
-		- store ³
-		- learn ⁴
+		- learn ³
+		- store ⁴
 
 	¹ Refused
 	² Eloqua
-	³ Digital River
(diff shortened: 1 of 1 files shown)


gloomy-ghost:
There is a bunch of subdomains... Google gives 400 when I use
 `site:vmware.com -site:buildweb.eng.vmware.com -site:code.vmware.com -site:cuny.horizon.vmware.com ....`

It seems to be inevitable for losing some subdomains if we want to remove wildcard. Please let me know if temporary losing of subdomains is acceptable.

jeremyn:
I suggest keeping this pull request minimal to fix the reported problem with `learn`. If you want to do a bigger cleanup for `VMware.xml` -- please feel free! -- then do that in a separate pull request.

For this pull request:

- [ ] Please change where `learn` appears in each case so that it's alphabetical. You can renumber the footnotes in the comment to keep them as 1/2/3/4 if you want.

jeremyn:
Thanks, merged.

PR #6973, opened 2016-09-22
Update SSLlabs.xml

gloomy-ghost:
Add SSL Server Tests under CAs subdomains
  Securecookie for all


jeremyn:
Thanks. Please:

- [ ] Alphabetize the targets. The general rule is to alphabetize from the top-level domain at the right, reading left, and then move `^` and `www` to the top of their group. For example:

```
example.com
www.example.com
a.example.com
www.a.example.com
b.a.example.com
b.example.com
example.net
www.example.net
a.example.net
```

The update looks good otherwise.

jeremyn:
Thanks, merged.

PR #6977, opened 2016-09-22
Remove roles for Socialbakers company website

jakubboucek:
Remove it because no more needed. Socialbakers are now automatically redirect to https on all supported (sub)domains. Not redirected domains can't be forced through rule for some issues (mixed content, service without https, etc.) - when we fix that, it will be immediately redirected on server.


jeremyn:
Rulesets for sites that already redirect to HTTPS are valid for HTTPS Everywhere, see [here in the FAQ](https://www.eff.org/https-everywhere/faq#faq-Why-does-HTTPS-Everywhere-include-rules-for-sites-like-PayPal-that-already-require-HTTPS-on-all-their-pages?). The only current exception is if all the domains in the rulesets are in the [HSTS preload list](https://chromium.googlesource.com/chromium/src/net/+/master/http/transport_security_state_static.json), which `socialbakers.com` is not.

So, ordinarily we should not delete the ruleset. However in this case the rule is already `default_off` and `platform="mixedcontent"` and breaks our [our style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) in more than one way, so I'm okay with deleting it. But for example if someone else created a new valid ruleset for `socialbakers.com`, we would accept it.

We like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) commits, which means rewriting commits in your name. Is that all right with you for this and any other pull requests you submit to us?

jakubboucek:
OK, understand.

I will try to fix that. I don't know how to define exceptions in wildcard rule. For example: Is exception on domain `hello.socialbakers.com` that has work on https, but all another domains `*.socialbakers.com` is in.

jeremyn:
Does that mean you want to rewrite the rule to improve it, instead of deleting it?

If so, then don't use wildcards. Instead, make a target for each working subdomain, and make a note for each subdomain that doesn't work. Your ruleset should look something like [this](https://github.com/EFForg/https-everywhere/blob/03e4d70af7cdaee89acc803f6d65c46ab2d0d103/src/chrome/content/rules/Juno.com.xml), except of course with your own subdomains.

jakubboucek:
Yes, i want to update ruleset to currently valid project state. 

Is unable to manually define all used subdomain, because lot of content are server on dynamic non-predicable subdomains (testing runtime, private content, customers content). All of that is fully https. 

Please wait, I will consult it with our DevOps.

jeremyn:
@jakubboucek What's the status for this PR on your end? I'd like to either merge it or close it.

jeremyn:
I'm closing this pull request due to lack of response. Please feel free to create a new pull request when you've decided how you want to move forward.

PR #6987, opened 2016-09-24
Add MetaEbene.me 

numismatika:
small passive mixed-content warning because audio-files are loaded from the meta-subdomain that has no valid cert, but that is minor.

### DIFF
--- a/src/chrome/content/rules/MetaEbene.xml
+++ b/src/chrome/content/rules/MetaEbene.xml
@@ -1,0 +1,57 @@
+<!--
+	Invalid certificate:
+		- feeds.metaebene.me
+		- docs.podlove.org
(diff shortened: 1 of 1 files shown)


jeremyn:
How are these different domains related to each other?

numismatika:
They belong to the same person that runs a lot of podcasts and the community behind them and are hosted non the same server as  well as it looks like.

numismatika:
the meta subdomain that serves the podcasts now uses TLS too, so far the site looks working with a minor mixed content warning.
Should we merge this now?

numismatika:
Mixed-Content is gone as far as i can see.
I would say this can safely be enabled for the next release

PR #6989, opened 2016-09-24
rm rage4.com.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=8746), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#8236), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n9105)


jeremyn:
The preload lists include `rage4.com` but not `www.rage4.com`, so this rule is still partially valid.

jeremyn:
@J0WI ?

jeremyn:
@J0WI It looks like I commented right when you merged, so I guess we just missed each other.

Out of @gloomy-ghost 's pull requests here, #6995 , and #6997 , all of them seem incorrect because not all of the domains in the rulesets are in the lists. What do you think?

J0WI:
@jeremyn the preload list has two values.
old format:
```
class nsSTSPreload
{
  public:
    const char *mHost;
    const bool mIncludeSubdomains;
};
```
new format:
```
struct nsSTSPreload
{
  const uint32_t mHostIndex : 31;
  const uint32_t mIncludeSubdomains : 1;
};
```

So the host itself is preloaded and all it's subdomains.

jeremyn:
Oh okay, thanks.

PR #6996, opened 2016-09-24
Update GoogleTranslate.xml

gloomy-ghost:
securecookie for all
remove translate.googleapis.com*

\* Preloaded
[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=322), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#10327), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n11430)

### DIFF
--- a/src/chrome/content/rules/Google.tld_Subdomains.xml
+++ b/src/chrome/content/rules/Google.tld_Subdomains.xml
@@ -55,16 +55,19 @@
 	<target host="picasaweb.google.com.*" />
 	<target host="scholar.google.com.*" />
 
-	<securecookie host="^(?:accounts|adwords|m|picasaweb)\.google\.[\w.]{2,6}$" name=".+" />
+	<target host="translate.google.*" />
+	<target host="translate.google.co.*" />
+	<target host="translate.google.com.*" />
+	<securecookie host="^(accounts|adwords|m|picasaweb|translate)\.google\.[\w.]{2,6}$" name=".+" />
 
 	<!-- path that 404 on https but 301 on http -->
-	<rule from="^http://news\.google\.((?:com?\.)?\w{2,3})/archivesearch"
+	<rule from="^http://news\.google\.((com?\.)?\w{2,3})/archivesearch"
 			to="https://news.google.$1/news/advanced_news_search" />
 		<test url="http://news.google.ca/archivesearch" />
 		<test url="http://news.google.com/archivesearch" />
 		<test url="http://news.google.com.au/archivesearch" />
 
-	<rule from="^http://(accounts|adwords|finance|groups|id|lh\d|m|news|picasaweb|scholar)\.google\.((?:com?\.)?\w{2,3})/"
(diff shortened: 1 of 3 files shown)


jeremyn:
Thanks. Please:

- [ ] Add `translate.google.*`, `translate.google.co.*`, and `translate.google.com.*` to [`Google.tld_Subdomains.xml`](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/Google.tld_Subdomains.xml) along with whatever tests, cookie updates etc are needed
- [ ] Delete `GoogleTranslate.xml`
- [ ] Add a reference in [`GoogleServices.xml`](https://github.com/EFForg/https-everywhere/blob/f45195d18399bbda5c5e0c850098d59be281720c/src/chrome/content/rules/GoogleServices.xml) to `Google.tld_Subdomains.xml`

jeremyn:
Please move the various `translate` targets to within the appropriate TLD group, instead of bunching them at the end.

jeremyn:
Thanks. Please re-add the non-capturing `?:` markers to the `rule` tags. Note that the `rule` tags use groups so it might not be safe to remove `?:`.

gloomy-ghost:
Sorry, I don't get you. Are there any differences? `(?:)` is in the second group of `rule`, it only matters when the rewriting pattern contains `$3`.

jeremyn:
Honestly these regexes are complicated, I don't really understand how nested capturing groups work and I don't want to mess with it on high-traffic sites like Google.

gloomy-ghost:
Just had a test on it:

with `(?:)`	redirect to `https://$1.google.$2/`:	https://translate.google.com.kh/
without `(?:)`	redirect to `https://$1.google.$2/`:	https://translate.google.com.kh/

with `(?:)`	redirect to `https://$1.google.$3/`:	https://translate.google./ *
without `(?:)`	redirect to `https://$1.google.$3/`:	https://translate.google.com./

\* In this case we don't really have group 3 because `(?:)` doesn't count as a group, to show the relationship clearer there is another test, add a group on `\w{2,3}`:

(redirect to `https://$1.google.$3/`)

`^http://(translate)\.google\.((?:com?\.)?(\w{2,3}))`	https://translate.google.kh/
`$1` is `translate`, `$2` is `com.kh`, `$3` is `kh`

`^http://(translate)\.google\.((com?\.)?(\w{2,3}))`	https://translate.google.com./
`$1` is `translate`, `$2` is `com.kh`, `$3` is 'com.', `$4` is `kh`

jeremyn:
All right, you've forced me to learn something, thanks. [This StackOverflow Q&A](https://stackoverflow.com/questions/1313934/how-are-nested-capturing-groups-numbered-in-regular-expressions) is consistent with your results, which is that capturing groups are numbered from left to right in order of their opening parentheses. So, nested capturing groups will be `$3`, `$4` etc which doesn't matter for us. Also it seems okay in a browser, and passes the tests.

Merged.

PR #6997, opened 2016-09-24
rm With_Google.com.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=323), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#11154), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n12338)


jeremyn:
The preload lists only cover `withgoogle.com`, but not `*.withgoogle.com`, at least not `rework.withgoogle.com`, so this rule is at least partially valid.

J0WI:
This host sends no preload header an refers to GoogleServices.xml

jeremyn:
See pull request #7081 for a discussion on the `preload` header. I agree the `GoogleServices.xml` reference should be removed.

jeremyn:
See also issue #7126 .

jeremyn:
https://withgoogle.com has no `Strict-Transport-Security` header at all.

PR #7005, opened 2016-09-24
rm CDT.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=2418), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#1666), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n1841)


src/chrome/content/rules/CDT.xml
```diff
@@ -1,22 +0,0 @@
-<!--
-	Other Center for Democracy & Technology rulesets:
-
-		- Not_Without_a_Warrant.com.xml
-		- We_Need_to_Know.info.xml
```
J0WI: Please also update the references there.

PR #7006, opened 2016-09-24
add rule for Augur.io

numismatika:


src/chrome/content/rules/Augur.io.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Augur.io">
+
```
gloomy-ghost: Please also add `augur.io`, and move `www.augur.io` directly under it

src/chrome/content/rules/Augur.io.xml
```diff
@@ -0,0 +1,13 @@
...
+<target host="api.augur.io" />
+<target host="auspice.augur.io" />
+<target host="cdn.augur.io" />
+<target host="request.augur.io" />
+<target host="www.augur.io" />
+
```
gloomy-ghost: Add a securecookie tag here

You may need to adjust the indentation

PR #7009, opened 2016-09-24
rm Medium.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=811), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#6406), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n7082)


jeremyn:
I haven't thoroughly checked the HSTS status of `medium.com`, but this ruleset is the "center" ruleset that has references to various other Medium.com-related rulesets.

Can you check [these rulesets](https://github.com/EFForg/https-everywhere/blob/69c44f612773494091ca6be3c4f3a6ed571f9327/src/chrome/content/rules/Medium.xml#L19-L26) and see how many of them we need to worry about? It looks like some don't exist, and maybe others are also HSTS preloaded.

gloomy-ghost:
`growth-supply.xml` doesn't work anymore, should I disable it or delete it?

I haven't checked the others, but Firefox's list doesn't include any of them.

jeremyn:
I agree `medium.com` is covered by HSTS preloading and so we can delete the ruleset.

I'm okay with losing the Medium references, because these other sites just seem to be blogs or something that happen to be hosted by medium.com.

- [ ] I'm not getting a mismatch for https://www.500ish.com so you can remove the comment and special rule there.
- [ ] http://blog.growth.supply redirects to https://blog.growth.supply anyway, so our ruleset doesn't make the problem worse, so don't disable it.

Looks good otherwise.

gloomy-ghost:
However they are failing the test:

`https://www.500ish.com/`: [L225](https://travis-ci.org/EFForg/https-everywhere/jobs/163602231#L225)
`https://blog.growth.supply/`: [L207](https://travis-ci.org/EFForg/https-everywhere/jobs/163602231#L207)

jeremyn:
I've updated the checklist based on your last comment.

`growth-supply.xml` is entirely broken, let's just delete it.

jeremyn:
Thanks, merged.

PR #7015, opened 2016-09-24
rm Romab.com.xml

gloomy-ghost:
romab.com is preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=625), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#8551), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n9453)

romab.se is not in DNS record


J0WI:
But www.romab.se is and it's not preloaded.

J0WI:
Please document why this redirect to a different TLD is needed and that .com is already preloaded.

src/chrome/content/rules/Romab.se.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	(www.)?romab.se: Mismatched
+-->
+<ruleset name="romab.se">
+	<target host="romab.se" />
+	<target host="www.romab.se" />
+
+	<rule from="^http://(?:www\.)?romab\.se/"
```
J0WI: Please avoid non-capturing groups.

PR #7027, opened 2016-09-25
rm Konklone.com.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=948), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc?redirect_type=single#5603), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n6184)


J0WI:
For some reason this is not in the current list in the mozilla-central branch.

jeremyn:
The domain should be in [`mozilla-release`](https://dxr.mozilla.org/mozilla-release/source/security/manager/ssl/nsSTSPreloadList.inc), not `aurora` or `central`, because `release` is what normal users are actually using. We don't want to delete rulesets until the preload lists in the released stable browsers have the equivalent behavior, for example see https://github.com/EFForg/https-everywhere/pull/7055#issuecomment-249657438 .

J0WI:
It's back on all branches now.

PR #7044, opened 2016-09-25
rm Robtex.net.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=3128), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#8528), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n9427)


src/chrome/content/rules/Robtex.net.xml
```diff
@@ -1,14 +0,0 @@
-<!--
-	For other Robtex coverage, see RobTex.xml.
```
J0WI: Please also update the references there.

PR #7068, opened 2016-09-25
rm Peercraft.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1172), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#7637), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n8448)


src/chrome/content/rules/Peercraft.xml
```diff
@@ -1,16 +0,0 @@
-<!--
-	Other Peercraft rulesets:
-
-		- Net-Safe.xml
-		- Ntacdn1.net.xml
```
J0WI: Please also update the references there.

PR #7070, opened 2016-09-25
rm thecustomizewindows.com.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1307), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#9992), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n11053)


J0WI:
Only thecustomizewindows.com is preloaded.

src/chrome/content/rules/thecustomizewindows.net.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="TheCustomizeWindows.net">
+	<target host="thecustomizewindows.net" />
+	<target host="www.thecustomizewindows.net" />
+
+	<securecookie host=".+" name=".+"/>
+
+	<rule from="^http://(?:www\.)?thecustomizewindows\.net/"
```
J0WI: Please replace non-capturin group (`(?:www\.)?`) with `(www\.)?`.
Could you also add a comment why this redirect for .net to .com is needed and that .com is already preloaded? This makes it easier for future maintenance.

J0WI:
Thanks!

PR #7072, opened 2016-09-25
rm Fedora_People.org.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1323), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#3428), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n3789)


src/chrome/content/rules/Fedora_People.org.xml
```diff
@@ -1,17 +0,0 @@
-<!--
-	For other Fedora Project coverage, see Fedora.xml.
```
J0WI: Please also update the references there.

PR #7053, opened 2016-09-25
rm OpenSSL.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=14752), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#7395), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n8183)

PR #7081, opened 2016-09-25
rm Mathias_Bynens.be.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1064), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#6298), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n6965)


J0WI:
This host does not send the preload header (anymore).

jeremyn:
It still sends `Strict-Transport-Security:"max-age=31536000; includeSubDomains"` and is in the lists so it's still effectively preloaded. It's possible the site admin removed the `preload` header after the site was included in the lists, for whatever reason.

J0WI:
From https://hstspreload.appspot.com/:
> Chrome has not yet removed any domains from the preload list for failing to keep up the requirements after submission, but there are plans to do so in the future.

So if a site stops sending the preload header it might get removed in future.

jeremyn:
That is a good point, but I think it's too restrictive to say that we can't delete a ruleset because its HSTS preload status might change in the future.

jeremyn:
Here's another problem, this one from #7053 : https://wiki.openssl.org has a complete header `Strict-Transport-Security:"max-age=31536000; includeSubDomains; preload"` header, but its redirect target https://wiki.openssl.org/index.php/Main_Page doesn't even have a `Strict-Transport-Security` header.

J0WI:
As @ScottHelme analyzed in his [blog post](https://scotthelme.co.uk/death-by-copy-paste/) some admins just copied the config from somewhere else and got added to the preload list without knowing the consequences.
So I think we should make sure, that they are in the current preload list (e.g. Tor), as well as in the preload list of future versions (e.g. mozilla central) and that they serve a valid HSTS preload header.

jeremyn:
I've created issue #7126 to discuss this.

jeremyn:
@J0WI I'm going to copy your last comment into the issue so I can respond to it there.

jeremyn:
I now agree with @J0WI that leaving out `preload` from the `Strict-Transport-Security` header, as in https://mathiasbynens.be , is enough reason to reject a deletion. Closing.

PR #7086, opened 2016-09-25
Update Nos-oignons.net.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1528), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#7198), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n7967)


J0WI:
Only nos-oignons.net is preloaded.

J0WI:
Please document why this redirect to a different TLD is needed and that .net is already preloaded.

PR #7094, opened 2016-09-25
rm Blockchain.info.xml

gloomy-ghost:
Preloaded

[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1641), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#1259), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n1387)


jeremyn:
Please update [`Blockchainbdgpzk.onion.xml`](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/Blockchainbdgpzk.onion.xml). Looks good otherwise.

gloomy-ghost:
`platform="mixedcontent"` won't skip the test.

jeremyn:
I'm not sure what to do here.

@fuglede @J0WI A `.onion` ruleset has been changed and is failing Travis. How do we make the test pass?

J0WI:
I wonder that we even have .onion hosts in our ruleset.
It was discussed in https://github.com/EFForg/https-everywhere/pull/5085#issuecomment-235299467 that we should not redirect to a hidden service, but it this case it would only redirect a hidden service to https. Maybe @jsha can clarify his statement for this case?

jsha:
I think a "simple" rule for an onion hostname is fine. I just don't want to become a registry for mapping DNS names to Onion Service names, because we don't have a good way to validate that mapping.

However, it probably makes sense to add a new platform for Tor Browser specifically, and skip such rulesets in the tests, since onion hostnames won't resolve. as a stopgap here I think it would be reasonable to just whitelist the hash of the updated ruleset.

jeremyn:
Thanks @J0WI and @jsha .

@gloomy-ghost Following https://github.com/EFForg/https-everywhere/pull/7094#issuecomment-249715042 , please generate a `sha256sum` for the modified `Blockchainbdgpzk.onion.xml` and add it in the appropriate sorted location in [`utils/ruleset-coverage-whitelist.txt`](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist.txt).

J0WI:
@jsha the changes in would #5085 also patch our test to fetch to use tor, so testing is possible.

gloomy-ghost:
Well, I found some removed rulesets are still in the whitelist:
```
keybase.io Keybase.io.xml
changetip.com ChangeTip.com.xml
phpmyadmin.net PhpMyAdmin.net.xml
weblate.org Weblate.org.xml
```

jeremyn:
@gloomy-ghost Don't worry about removed rulesets or duplicates in the whitelist. They're harmless, and there's a [cleanup script](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist-cleanup.sh) that someone runs every once in a while.

gloomy-ghost:
However it is still failing…

J0WI:
Adding it to the whitelist does not help in this case. The whitelist is just for those rules, that have not enough test urls.

We need to patch our tests first before we can accept .onion rules.

jeremyn:
I'm getting a sha256 hash of `34c60fd37894f38c9b90375586055c39d11ed8fe2340606587ac2bcce29718c7` for the modified `Blockchainbdgpzk.onion.xml`, not `6deb04f50f7fdef16b0f2c8732f6c8043413383a8bbda21df5f0ff912b1990cb` which is what was added to the whitelist.

gloomy-ghost:
```
$ sha256sum Blockchainbdgpzk.onion.xml
6deb04f50f7fdef16b0f2c8732f6c8043413383a8bbda21df5f0ff912b1990cb *Blockchainbdgpzk.onion.xml
```
Quite strange

jeremyn:
Thanks, merged.

PR #6941, opened 2016-09-21
Update dianping.com.xml —— fix MCB

renyouguo:

PR #7123, opened 2016-09-25
Update geetest.xml —— fix page broken

renyouguo:


renyouguo:
@J0WI Please check the #6941 #6956

PR #7138, opened 2016-09-27
Create Maxon.net.xml

gloomy-ghost:


jeremyn:
Thanks. Please:

Add target:
- [ ] https://fcws.maxon.net
- [ ] https://landingpage.maxon.net
- [ ] https://r18.maxon.net
- [ ] https://register.maxon.net

Also:
- [ ] Remove the general `securecookie`, Sublist3r shows too many sites that we're missing (but which I don't think are even worth noting as problems).

jeremyn:
Thanks, merged.

PR #7139, opened 2016-09-27
Create Btcc.com.xml

gloomy-ghost:


jeremyn:
You're doing too many things in`BTCChina.com.xml`. For example, you redirect http://m.btcchina.com to https://mobile.btcchina.com , but https://m.btcchina.com works fine and the site redirects to https://mobile.btcchina.com on its own. Similarly with the btcc.com redirects.

jeremyn:
Thanks. Please:

`Btcc.com.xml`:

Add `target`:
- [ ] https://data.btcc.com
- [ ] https://mint.btcc.com
- [ ] https://pro-data.btcc.com

Also:
- [ ] Remove the `static` test.
- [ ] Remove the general `securecookie`.

`BTCChina.com.xml`:

Add `target`:
- [ ] https://api.btcchina.com
- [ ] https://com.btcchinacdn.com

Also:
- [ ] Remove the `target` and `rule` for https://cn.btcchina.com . If I allow the exception to get past the certificate mismatch, https://cn.btcchina.com does not redirect itself to https://api.btcc.com , so we shouldn't do that here.
- [ ] Same thing for https://forwards.btcchina.com .
- [ ] Remove the general `securecookie`.
- [ ] Add the full domain to the items in the top comment, so for example say `cn.btcchina.com` instead of just `cn`.

jeremyn:
As a general comment, there's no need to add an extra test for a target like `static.btcc.com` just because the root URL https://static.btcc.com looks weird.

jeremyn:
Thanks, merged.

Issue #7126, opened 2016-09-25
Discussion for guidelines on removing HSTS preload rulesets


jeremyn:
@gloomy-ghost has submitted several pull requests deleting rulesets that are in the HSTS preload lists: [Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1). ( @gloomy-ghost has helpfully included links to these rulesets in their pull requests.) @J0WI and I have separately handled these and it turns out we disagree on when a ruleset should be deleted. We have mainly discussed this in pull request #7081 , but I'll repeat the discussion below.

@J0WI rejects rulesets for deletion if the `Strict-Transport-Security` header does not contain the `preload` header that is required in the [HSTS preload submission requirements](https://hstspreload.appspot.com/), even if the domain is in the preload lists. I assume that the `preload` header was in the lists when the domain was submitted to the lists and was later removed for whatever reason. This turns to be a common situation. My own opinion is that if the domain is in the preload lists and it has a `Strict-Transport-Security` header, then we can delete the ruleset even if `preload` is missing. @J0WI correctly points out that according to the submission requirements,

> Chrome has not yet removed any domains from the preload list for failing to keep up the requirements after submission, but there are plans to do so in the future.

I think it is too restrictive to not delete a ruleset in the preload lists just because it might be removed from the lists later.

Another problem that I found in pull request #7053 : https://wiki.openssl.org has a complete `Strict-Transport-Security:"max-age=31536000; includeSubDomains; preload"` header, but its redirect target https://wiki.openssl.org/index.php/Main_Page doesn't even have a `Strict-Transport-Security` header. Should we delete a ruleset if some of the responses have the header but some don't? If so, how many URLs do we have to check? What if only `preload` is missing?

Pinging @fuglede and @Hainish for their input.

jeremyn:
@J0WI wrote [here](https://github.com/EFForg/https-everywhere/pull/7081#issuecomment-249434402):

> As @ScottHelme analyzed in his [blog post](https://scotthelme.co.uk/death-by-copy-paste/) some admins just copied the config from somewhere else and got added to the preload list without knowing the consequences.
[...]

PR #7141, opened 2016-09-27
rm Jim_Shaver.net.xml

gloomy-ghost:
Preloaded
[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1900), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#5159), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n5710)

HTTP: True
HTTP Redirect: True

HTTP Header: None
HTTPS Header: Strict-Transport-Security: max-age=31536000; includeSubDomains; preload


jeremyn:
According to [RFC 6797](https://tools.ietf.org/html/rfc6797), compliant hosts aren't supposed to send STS headers over insecure HTTP (section 7.2), and compliant browsers are supposed to ignore STS headers received over insecure HTTP (section 8.1). See [this Security.SE question](https://security.stackexchange.com/questions/84714/why-does-rfc6797-say-an-hsts-host-must-not-include-the-sts-header-field-in-http). For preload purposes, the header over plain HTTP is irrelevant and you don't need to mention it.

Also, although so far it's just my opinion ( https://github.com/EFForg/https-everywhere/issues/7126#issuecomment-249889749 ), I think we can accept deletions even if there is no plain HTTP version of the site, so you don't need to mention that either.

gloomy-ghost:
Well, I have never tried to read a RFC completely due to my poor English vocab...
So I'm not going to provide comments that regarding HTTP in future PRs.

Moreover, can you merge this PR if there is no modifying needed? Thanks

jeremyn:
Thanks for the deletion, merged.

PR #7143, opened 2016-09-27
rm Get_Fedora.org.xml

gloomy-ghost:
Preloaded
[Chromium](https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json?amp&sq=package:chromium&l=1956), [Firefox](https://dxr.mozilla.org/mozilla-aurora/source/security/manager/ssl/nsSTSPreloadList.inc#3934), [Tor](https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.4.0esr-6.5-1#n4337)

HTTP: True
HTTP Redirect: True

HTTP Header: None
HTTPS Header: Strict-Transport-Security: max-age=15768000; includeSubDomains; preload


jeremyn:
Thanks. Please remove the reference to this ruleset in [`Red_Hat.xml`](https://github.com/EFForg/https-everywhere/blob/90536e78269bf9b06bdd7c5e489cd47abd258640/src/chrome/content/rules/Red_Hat.xml). Looks good otherwise.

jeremyn:
For your changes in `Red_Hat.xml`: 

- [ ] https://et.redhat.com is `refused`, not `incomplete certificate chain`. The test only reports it as `incomplete certificate chain` because of the rewrite from `et.` to `community.` and `community.` reports the incomplete chain.
- [ ] For https://community.redhat.com , I get `(60, 'SSL certificate problem: unable to get local issuer certificate')` with `fetch-test.sh` on the command-line, which confirms what Travis gets. But, I can't reproduce the problem in Firefox, even with a new profile. Are you able to reproduce this problem in the browser?

gloomy-ghost:
FYI: [SSLLabs](https://www.ssllabs.com/ssltest/analyze.html?d=community.redhat.com) reports that the missing intermediate is `DigiCert SHA2 Extended Validation Server CA`, which I can find in Firefox even after refresh its profile.

jeremyn:
Thanks, merged.

jeremyn:
With a new profile in Firefox and checking in `Options > Certificates > View Certificates`, `DigiCert SHA2 Extended Validation Server CA` is listed with `Software Security Device`, which is typical for fetched intermediates, instead of `Builtin Object Token`. So I agree something is strange with this intermediate, which is why I merged the pull request.

Is Firefox automatically fetching a few certificates even for a new profile? @fuglede @J0WI What do you think?

J0WI:
`firefox -no-remote -profile $(mktemp -d) https://community.redhat.com/` fails for me.
However if you refresh a few seconds later it works. I think that's because Mozilla also uses DigiCert and after Firefox checked for stuff (like Updates, the Safe Browsing list or Telemetry) Firefox has cached the chain.

jeremyn:
I can reproduce the problem now, how tricky. Good catch.

PR #7146, opened 2016-09-27
[Bitbucket.org] status's certificate is good now; fix name capitalization

mnordhoff:
* `status.bitbucket.org`'s certificate is no longer mismatched; remove workaround

* Change the `<ruleset name>` from "BitBucket.org (partial)" to "Bitbucket.org (partial)". I didn't rename the files.

### DIFF
--- a/src/chrome/content/rules/BitBucket.xml
+++ b/src/chrome/content/rules/BitBucket.xml
@@ -9,20 +9,12 @@
 		- s3.amazonaws.com/bitbucket-assetroot/
 		- dwz7u9t8u8usb.cloudfront.net
 
-		- bitbucket.statuspage.io
-
-			- status
-
 
 	Problematic subdomains:
 
 		- portix	(works; mismatched, CN: bitbucket.org)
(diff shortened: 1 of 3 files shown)


jeremyn:
Thanks. I agree with the name change, but you should also rename the file for consistency. Specifically:

- [ ] Rename `BitBucket.xml`
- [ ] Rename `BitBucket-mismatches.xml`
- [ ] Update the reference in each file that points to the other file
- [ ] Change the ruleset `name` in `Bitbucket-mismatches.xml`

Also, we like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) pull requests, which means rewriting pull requests in your name. Is that all right with you for this and any other pull requests you submit to us?

mnordhoff:
Good point. Done.

I renamed it to "Bitbucket" (without ".org") since it includes a couple rules for "bitbucket.com". (I suppose splitting off another ruleset is another option.)

I don't mind squashing. :-)

PR #7153, opened 2016-09-28
projectatomic.io

galeksandrp:


src/chrome/content/rules/projectatomic.io.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+projectatomic.io ¹
+ask.projectatomic.io ⁴
+lists.projectatomic.io ⁴
```
J0WI: lists works for me

galeksandrp: Added.

galeksandrp: incomplete certificate chain

PR #7155, opened 2016-09-28
urldecode.org

galeksandrp:


src/chrome/content/rules/urldecode.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="urldecode.org">
+	<target host="urldecode.org" />
+	<target host="www.urldecode.org" />
```
J0WI: urlencode.org
www.urlencode.org

galeksandrp: Added.

PR #7171, opened 2016-09-28
japlusu.com

galeksandrp:


src/chrome/content/rules/japlusu.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="japlusu.com">
+	<target host="japlusu.com" />
+	<target host="www.japlusu.com" />
```
J0WI: mismatch on m.japlusu.com

galeksandrp: Added.

PR #7178, opened 2016-09-28
scotthel.me

galeksandrp:


jeremyn:
Thanks, but http://scotthel.me is already in the HSTS preload list so we don't need a ruleset for it. See the discussion in https://github.com/EFForg/https-everywhere/issues/7126 . Closing.

PR #7190, opened 2016-09-28
jsbin.com

galeksandrp:


jeremyn:
To whoever reviews this: there are multiple contributors, so please use a regular merge instead of a squash-and-merge.

PR #7201, opened 2016-09-28
eternallybored.org

galeksandrp:


src/chrome/content/rules/eternallybored.org.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+www.eternallybored.org ¹
```
J0WI: tanja.eternallybored.org

galeksandrp: Added.

PR #7217, opened 2016-09-29
yougetsignal.com

galeksandrp:


src/chrome/content/rules/yougetsignal.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="yougetsignal.com">
+	<target host="yougetsignal.com" />
+	<target host="www.yougetsignal.com" />
```
J0WI: ports.yougetsignal.com WFM

galeksandrp: Added.

PR #7227, opened 2016-09-30
[pwSafe] New ruleset

galenide:


J0WI:
Is this related to #7226? If so, please add a hint in the comments of both rulesets.

galenide:
The site at https://www.pwsafe.org/ (in pull request #7226) appears to be mainly about the free and open source [Password Safe](https://en.wikipedia.org/wiki/Password_Safe) software. The https://pwsafe.info/ site (in this pull request) is about pwSafe, a clone of Password Safe that is for the Mac OS X and iOS platforms. The pwSafe site appears to be managed by [App77](https://app77.com/); the Password Safe site [mentions](https://pwsafe.org/relatedprojects.shtml) pwSafe under "related projects" but it is likely that App77 is separate from the developers who manage Password Safe itself. Would it be worth adding comments to both rulesets indicating in each that there is another ruleset with additional Password Safe coverage?

J0WI:
Since the sites are not maintained by the same company/community I'm fine with this.

PR #7229, opened 2016-09-30
Update Qpic.cn.xml

ivysrono:


J0WI:
qidian.qpic.cn may also work.

J0WI:
Thanks!

PR #7238, opened 2016-09-30
ayyo.ru

galeksandrp:


J0WI:
There are some more subdomains, could you add them too?

galeksandrp:
Added.

J0WI:
I also got:
mag.ayyo.ru
s.ayyo.ru

galeksandrp:
Added, thanks.

PR #7241, opened 2016-09-30
qrator.net

galeksandrp:


src/chrome/content/rules/qrator.net.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="qrator.net">
+	<target host="qrator.net" />
+	<target host="www.qrator.net" />
+	<target host="client.qrator.net" />
```
J0WI: radar.qrator.net WFM
static.qrator.net times out

galeksandrp: Added.

PR #7245, opened 2016-09-30
asexuality.org

gmskosua:


jeremyn:
Thanks for your pull request. Your ruleset needs some revision according to the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md), for example:

* Wildcard URLs like `*.asexuality.org` are discouraged. You can use the Google search technique described in the style guide to find more domains.
* Wildcard `securecookie` tags should be `.+`/`.+`, not `.*`/`.*` (but see below)

Also, I notice one of the domains found through Google is https://ru.asexuality.org , which has an invalid certificate. This should be listed in a comment at the top as a `Mismatch`. See [here](https://github.com/EFForg/https-everywhere/blob/03e4d70af7cdaee89acc803f6d65c46ab2d0d103/src/chrome/content/rules/Juno.com.xml) for an example. Also, because https://ru.asexuality.org isn't covered, you shouldn't use a wildcard `securecookie` tag (see the style guide).

gmskosua:
> Wildcard URLs like *.asexuality.org are discouraged. You can use the Google search technique described in the style guide to find more domains.

I removed the wildcard and added a www target. The only subdomains I found are ru.asexuality.org and www.asexuality.org.

> Wildcard securecookie tags should be .+/.+

Done

> This should be listed in a comment at the top as a Mismatch.

Done

> you shouldn't use a wildcard securecookie tag

Done

jeremyn:
Thanks. Some more things:

* Your `test` is unnecessary because an implicit test is created for the `www` `target`, so please delete your `test`.
* You don't need to specify the leading dot in `securecookie`, see [here](https://www.eff.org/https-everywhere/rulesets#secure-cookies), however...
* This could happen: `www` sets a cookie for `.` that is marked HTTPS-only. If the user then visits `ru`, which is HTTP, that cookie can't be read and site functionality can break. The safest thing to do is to only add a `securecookie` for `www`, so please do that. This isn't a hard-and-fast rule though, you'll see it done differently depending on the ruleset.

Finally, we like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) pull requests, which means rewriting commits in your name. Are you okay with that for this and any other pull requests you submit to us?

gmskosua:
Thank you for your help.

> please delete your test.

Done

> You don't need to specify the leading dot in securecookie, see here, however...
> The safest thing to do is to only add a securecookie for www, so please do that.

Done

> we like to squash-and-merge pull requests, which means rewriting commits in your name.

Done

jeremyn:
Thanks, merged.

PR #7255, opened 2016-10-01
add CONTRIBUTING.md

J0WI:
See also #6868
cc @jsha @Hainish @jeremyn @fuglede @cooperq

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,11 @@
-# Contributing to HTTPS Everywhere
+# Contributing rulesets to HTTPS Everywhere
 
-If you want to create new rules to submit to us, we expect them to be in the src/chrome/content/rules directory. That directory also contains a useful script, make-trivial-rule, to create a simple rule for a specified domain. There is also a script called trivial-validate.py, to check all the pending rules for several common errors and oversights. For example, if you wanted to make a rule for the example.com domain, you could run:
+If you want to create new rules to submit to us, we expect them to be in the `src/chrome/content/rules` directory. That directory also contains a useful script, `make-trivial-rule`, to create a simple rule for a specified domain. There is also a script called `utils/trivial-validate.py`, to check all the pending rules for several common errors and oversights. For example, if you wanted to make a rule for the `example.com` domain, you could run:
 ```
 cd src/chrome/content/rules
 bash ./make-trivial-rule example.com
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# Contributing to HTTPS Everywhere
```
fuglede: Since this is only about contributions of rulesets, perhaps that should be reflected in the title/naming?

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
...
+
+If you want to create new rules to submit to us, we expect them to be in the src/chrome/content/rules directory. That directory also contains a useful script, make-trivial-rule, to create a simple rule for a specified domain. There is also a script called trivial-validate.py, to check all the pending rules for several common errors and oversights. For example, if you wanted to make a rule for the example.com domain, you could run:
```
fuglede: Style-wise, I would add some `'s to format the names of folders and files here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
...
+```
+cd src/chrome/content/rules
+bash ./make-trivial-rule example.com
+```
+This would create Example.com.xml, which you could then take a look at and edit based on your knowledge of any specific URLs at example.com that do or don't work in HTTPS. Please have a look at our [Ruleset Style Guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) where you can find useful tips about finding more subdomains. Our goal is to have as many subdomains covered as we can find.
```
fuglede: > Our goal is to have as many subdomains covered as we can find.

As I've mentioned before, I generally applaud the thoroughness that goes into finding all of those, but it's time-consuming and I don't think that it should not be prioritized over work that creates greater value, so I would hesitate to add it as an explicit "goal", but I acknowledge that opinions on this may vary.

jeremyn: I think most people who read "Our goal is to have as many subdomains covered as we can find." will understand that as a suggestion that need to do a little searching, not that they need to be super aggressive.

CONTRIBUTING.md
Hainish: I would note that `trivial-validate.py` is located within the `utils/` path just to be clear.

CONTRIBUTING.md
Hainish: I agree with @jeremyn here.  I also think that having the `CONTRIBUTING.md` file err on the side of thoroughness is a good thing, because I expect that contributors tend to fall short of the level of exhaustiveness we'd prefer generally.*  Having this language is probably a good thing so that the bar is a little higher, and it's better that contribs fall slightly short of a high bar than a low one.

* There's probably a number of reasons for this - either they're using automated scripts, or they are new to the project and aren't aware of our preferential procedure of subdomain inclusion.

J0WI:
I addressed most of the feedback, but I was not able to find a better language. I'm also not a native speaker.
My proposal is to add this file and maybe someone will improve it later or we will see, that more things need to be documented for new contributors.

PR #7259, opened 2016-10-01
fix #6978

### LINKED ISSUE #6978: File a rule request change here? - dslreports.com
I'd like to add to the ruleset for dslreports.com
https://www.eff.org/https-everywhere/atlas/domains/dslreports.com.html

an exception to NOT re-write the URL www.dslreports.com/speedtest
(with or without arguments), to https.

The reason is that users are explicitly given the option of switching to http for a non-encrypted speed test, this is used frequently by Firefox users as Firefox has a current performance issue with https over 100 megabit.
Popping up a message to ask for disabling of https everywhere is undesirable versus modifying the ruleset to whitelist this url so it all happens automatically. The attempt to switch to http can be identified further by matching query arguments ?httpsok=0

J0WI:


J0WI:
Is this a false positive of the `?` in the exclusion? I don't see any failure in the test urls.

jeremyn:
Travis says "Not enough tests" for the exclusion [here](https://travis-ci.org/EFForg/https-everywhere/jobs/164328724#L221).

jeremyn:
The person who reported the issue just wants us to whitelist http://www.dslreports.com/speedtest , not http://www.dslreports.com/speedtest?httpsok=0 . You could just do that and then remove the two `httpsok` tests. I don't think exclusions for URLs with query strings are a good idea anyway.

J0WI:
dslreports.com/speedtest works fine over HTTPS, but they ask you if you want to use plain HTTP. If you click on that link you get pointed to dslreports.com/speedtest?httpsok=0 which is broken with HTTPS Everywhere.

jeremyn:
According to the person who wrote #6978 , the site doesn't work well over HTTPS on Firefox:

> The reason is that users are explicitly given the option of switching to http for a non-encrypted speed test, this is used frequently by Firefox users as Firefox has a current performance issue with https over 100 megabit.

Also on Firefox when I visit the page, I get an OK/Cancel prompt:

> Firefox is not performing well with high speed testing over https.

> Switch now to non-encrypted test page?

Ve2mrx:
This should be excluded as well: [Puma 6 chipset performance test](http://www.dslreports.com/tools/puma6)

The site redirects it to HTTP, and the plugin reloads it in HTTPS. So, infinite loop.

J0WI:
@Ve2mrx could you open an issue for this?

Ve2mrx:
See here: [https://github.com/EFForg/https-everywhere/issues/8488](https://github.com/EFForg/https-everywhere/issues/8488)

PR #7263, opened 2016-10-02
Add tumba.ch.xml

xable:


src/chrome/content/rules/tumba.ch.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="tumba.ch">
+        <target host="tumba.ch" />
+        <target host="www.tumba.ch" />
```
J0WI: Please also add:
api.tumba.ch
les.tumba.ch
u.tumba.ch

J0WI:
Thanks!

PR #7267, opened 2016-10-02
Add rule for alpnames.com

PotcFdk:


J0WI:
Could you add commetns for the mismatching certs on r.promo.alpnames.com and ski.alpnames.com?

Issue #4405, opened 2016-03-19
Add iciciprulife.com


sagarun:
Please add iciciprulife.com. This is a bank/investment account provider.

PR #7268, opened 2016-10-02
Add rule for Iciciprulife.com

PotcFdk:


jeremyn:
@PotcFdk Are you still interested in working on this?

PotcFdk:
@jeremyn What do you mean? Is anything wrong with the current state of the PR?

jeremyn:
I haven't really looked yet, I just want to make sure that if I do look through it and provide feedback, you're available to update the PR.

PotcFdk:
@jeremyn Sure, I will.

jeremyn:
Before I do a more thorough review, please add more subdomains to the ruleset, either as working `target`s or as problems noted at the top. I can find over 30 subdomains using the Sublist3r tool described in the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md). Not all of them should necessarily be included, but that's just to give you a general idea of the number. If you don't want to use Sublist3r, you can use the Google search technique described in the style guide.

Also, in all cases where there is a list of domains, please sort them in alphabetical order starting from the top level domain at the right reading left, moving `^` and `www` to the top of their group. For example:
```
example.com
www.example.com
a.example.com
www.a.example.com
b.a.example.com
b.example.com
example.net
www.example.net
a.example.net
```

PotcFdk:
@jeremyn Added more domains, found using Sublist3r, in 386f78c

jeremyn:
Please make the following changes, and feel free to ask if you want me to clarify the reason for any of these:

- [ ] Change the top comment ([these lines](https://github.com/PotcFdk/https-everywhere/blob/386f78c749f10825efe76d3c647e5293a2713222/src/chrome/content/rules/Iciciprulife.com.xml#L1-L14)) to this:
```xml
<!--
	Invalid certificate:
		iciciprulife.com
		blog.iciciprulife.com (incomplete certificate chain)
		www.careers.iciciprulife.com
		qliksense.iciciprulife.com
		uat-aem.iciciprulife.com

	No working URL known:
		atlas.iciciprulife.com
		clickpss.iciciprulife.com

--!>
```

Also:
- [ ] Add `target` in the correctly sorted location:
```xml
	<target host="healthclaims.icicprulife.com" />
		<test url="http://healthclaims.iciciprulife.com/BCSUIEARPHASE2/js/functions.js" />
```
- [ ] Add `target` in the correctly sorted location:
```xml
	<target host="qlikview.iciciprulife.com" />
		<test url="http://qlikview.iciciprulife.com/qlikview/global.css" />
		<exclusion pattern="http://qlikview\.iciciprulife\.com/$`
```
- [ ] Remove `target` for https://atlas.iciciprulife.com
- [ ] Remove `target` for https://clickpss.iciciprulife.com
- [ ] Add `test` for `http://retailer.iciciprulife.com/ATLAS/jsp/common/login/login.jsp` indented under `retailer` `target`
- [ ] Add `test` for `http://webservices.iciciprulife.com/webservices/ServiceManager?operation.view=dataProcessor` indented under `webservices` `target`
- [ ] Delete the `exclusion` for `hrapps`
- [ ] Delete one of the tests for `hrapps`

PotcFdk:
Done all but the ones that remove atlas and clickpss. Can you elaborate on why we need to get rid of those?

Also:
* Why do you want healthclaims added? It has errored when I initially tried it out, and it still does, as you can see in the CI log.
* Why are the newly added tests required for the retailer and webservices targets?

jeremyn:
Thanks, I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7268/commits/4792621aca56010d49220995eb4a127c18ce5fd0 and modified it slightly to fix the `qlikview` XML and to add an `exclusion` to `healthclaims`.

I want to remove `atlas` and `clickpss` because I have a personal rule-of-thumb that I don't like to add `target`s if I can't find at least one working URL for them. That's why I asked you to add them to the new top comment as `No working URL known`. See issue https://github.com/EFForg/https-everywhere/issues/7662 for the related discussion. The root URL for `atlas` returns a `403` and for `clickpss` returns a `302` redirect to a `404`. The new tests for `retailer` and `webservices` are there for a similar reason, the root URL for both returns a `404`.

`healthclaims` seems to work other than the root URL, which causes a problem for the automatic test generated for the `target`. The new `exclusion` in the checklist should avoid the problem with the automatic test. The `exclusion` for `qlikview` is there for a similar reason.

jeremyn:
For this line [here](https://github.com/PotcFdk/https-everywhere/blob/0998c47a5fcb26763840d1eb0ff539ccf5bed55f/src/chrome/content/rules/Iciciprulife.com.xml#L9), instead of naming me specifically, how about changing it to this?

```xml
	No working URL known (see https://github.com/EFForg/https-everywhere/pull/7268#issuecomment-267633534 )
```

PotcFdk:
Rewrote the commit. What about the test for the exception? I'm a bit confused by it.

jeremyn:
The `healthclaims` target is incorrectly for `icicprulife`, not `iciciprulife`. It's missing an `i` in the middle. Can you try fixing the `target` and removing the `exclusion`? We may not need an `exclusion` for a `503`.

PotcFdk:
Apparently we do, it's a 404?

jeremyn:
That's right, sorry. It's https://qlikview.iciciprulife.com that gets a `503`. Please add back the `exclusion` for `healthclaims` (and you can leave in the `exclusion` for `qlikview`).

jeremyn:
Okay, this looks ready, thanks. Do you mind if I delete your commit messages when I squash-and-merge?

PotcFdk:
I think just using the commit message of e636abc would be best.
Make sure to keep the `resolves #4405` part.

jeremyn:
Great! Merged.

PR #7269, opened 2016-10-03
Add rule for o2.pl

PotcFdk:


J0WI:
Please have a look at our [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) to add more subdomains.

PotcFdk:
@J0WI what exactly do you mean? I've tried to find more subdomains, but I wasn't able to find more subdomains that worked with HTTPS.

J0WI:
We use to comment non-working host on the top.

PR #7273, opened 2016-10-03
add chromium paths for opensuse

julianladisch:


jeremyn:
I haven't tried doing HTTPS Everywhere development on OpenSUSE. I had some trouble with Fedora (see https://github.com/EFForg/https-everywhere/pull/5584 ) and even now it's not quite as friendly as Debian/Ubuntu.

Aside from the changes in this pull request, how good is OpenSUSE HTTPS Everywhere dev support otherwise? The problem with adding only partial OpenSUSE support is that new HTTPS Everywhere contributors might see that, think OpenSUSE is fully supported, and then run into various problems.

cowlicks:
The chromium.sh part was fixed in #12326

The other part, listing the SUSE path, should have been fixed by #12624 

In both cases, I used `which` to find the path to the bin from the executable's name.

PR #7274, opened 2016-10-03
[FAZ.net] stellenmarkt, boersenlexikon, boersenspiel, faz-community

julianladisch:


src/chrome/content/rules/FAZ.xml
```diff
@@ -35,30 +51,32 @@
 	Timeout:
 		- financeforum.fazjob.net
 		- immobilien.faz.net
+
+	Timeout at HTTP Port 80:
+		- filebox.faz.de
 -->
 <ruleset name="FAZ (partial)" platform="mixedcontent">
```
J0WI: Can you confirm, that this is still `mixedcontent`?

PR #7276, opened 2016-10-04
Add rule for alpharatio.cc

bvhr:


jeremyn:
Thanks for your contribution. We don't like wildcard targets, instead please create a target for each subdomain. Also we automatically create a test for each target, so for example [this test](https://github.com/bvhr/https-everywhere/blob/9756f933916cd7cf9e3a865e898da297dbf5507d/src/chrome/content/rules/alpharatio.cc.xml#L5) is unnecessary. Please see [our style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md).

bvhr:
Sorry, I shouldn't have referred to older rules as a guide rather than the style guide. Hopefully it's correct now.

jeremyn:
Thanks, the new ruleset looks fine.

We like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/) pull requests, which means we rewrite pull requests in your name. Is that all right with you for this and any other pull requests you submit to us?

bvhr:
No worries, that's fine with me.

jeremyn:
Thanks, merged.

PR #7284, opened 2016-10-07
Update Eset.xml

volatile-Daniel:
We need it for proper work of our security tool

### DIFF
--- a/src/chrome/content/rules/Eset.xml
+++ b/src/chrome/content/rules/Eset.xml
@@ -54,7 +54,7 @@
 
 	<!--	Direct rewrites:
 				-->
-	<target host="eset.com" />
+	<!--<target host="eset.com" /> page redirected on server side, it only break ESET Secuce Browser -->
 	<target host="anti-theft.eset.com" />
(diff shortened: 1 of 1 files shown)


volatile-Daniel:
Mentioned exclusion pattern address is processed internally and AFAIK there is no public url, which is able to cover this test.

jeremyn:
https://eset.com/BPPRedirector looks valid to me. What is the specific problem you're trying to fix? Maybe we can think of another way to handle it.

Hainish:
@volatile-Daniel can you change the comment to be "page redirected on server side, and this breaks ESET Secure Browser"

Other than that, looks good to me

src/chrome/content/rules/Eset.xml
```diff
@@ -54,7 +54,7 @@
 
 	<!--	Direct rewrites:
 				-->
-	<target host="eset.com" />
+	<!--<target host="eset.com" /> page redirected on server side, and this breaks ESET Secuce Browser -->
```
Hainish: "Secure" not "Secuce", right?

PR #7289, opened 2016-10-08
[Woolyss.com] Add ruleset

synthtech:
**Note:** HTTPS works on this page (https://tools.woolyss.com/html5-audio-video-tester/) but redirects to a 403 when accessing https://tools.woolyss.com


J0WI:
You can make an exclusion for `http://tools.woolyss.com/` and add `https://tools.woolyss.com/html5-audio-video-tester/` as a test url.

synthtech:
Ah, I forgot that was a thing. Added the exclusion and test.

J0WI:
Thanks!

PR #7302, opened 2016-10-09
[Strikingly.com] Add several exclusions to fix login and dashboard

fuglede:
In particular, this fixes https://github.com/EFForg/https-everywhere/issues/6865. Note that, unfortunately, a login is required to reproduce the issues.

### DIFF
--- a/src/chrome/content/rules/Strikingly.com.xml
+++ b/src/chrome/content/rules/Strikingly.com.xml
@@ -13,38 +13,59 @@
 
 	Problematic hosts in *strikingly.com:
 
+		- assets **
 		- support *
+		- www **
 
 	* Mismatched
+	** Breaks login and dashboard; see https://github.com/EFForg/https-everywhere/issues/6865
 
 
 	Fully covered hosts in *strikingly.com:
 
-		- (www.)?
-		- assets
+		- ^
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Strikingly.com.xml
```diff
@@ -13,38 +13,59 @@
...
 	Fully covered hosts in *strikingly.com:
 
-		- (www.)?
-		- assets
+		- ^
 		- b
 
 
-	Insecure cookies are set for these hosts:
+	Insecure unsecurable cookies are set for these hosts:
```
Hainish: "Insecure unsecurable"?

fuglede:
>"Insecure unsecurable"?

Insecure cookies which can not be secured (as that breaks the login functionality). "Unsecurable" could also work.

Hainish:
For a new login, http://www.strikingly.com/s#/domains behaves differently with HTTPSE disabled compared to enabled with this edit.

Above, if you change "insecure unsecurable" to just "unsecurable" I think that's much clearer.

Hainish:
Merged in 25ff27133ba9b74ee13e029bd802b71b2ebc7746

PR #7314, opened 2016-10-11
Create ncar.cc.xml

ivysrono:


src/chrome/content/rules/ncar.cc.xml
```diff
@@ -0,0 +1,21 @@
+<!--
+	MCB:
+		bbs.ncar.cc
+-->
+
+<ruleset name="ncar.cc">
+
+	<target host="bbs.ncar.cc" />
+	<exclusion pattern="^http://bbs\.ncar\.cc/$" />
+	<exclusion pattern="^http://bbs\.ncar\.cc/.+\.(html|php)" />
```
fuglede: From the looks of it, the tests are returning a false positive here, as they should have caught the unescaped dot in the exclusion pattern, cf. #7320.

ivysrono: @fuglede How to fix it？

fuglede: @ivysrono: That particular issue might mean that we will have to let some of these kinds of exclusions throught, but for now, just replace the unescaped dot by a suitable character class, such as `[\w/-]`, or whatever might be relevant for the pattern.

J0WI:
There is also `beiyong.ncar.cc`

J0WI:
Thanks!

Issue #7277, opened 2016-10-04
Using Sublist3r to find subdomains for rulesets


jeremyn:
We've had discussions recently on automating ruleset work, including using a tool to find subdomains, for example see https://github.com/EFForg/https-everywhere/pull/4279 and https://github.com/EFForg/https-everywhere/issues/6912 .

One tool we've discussed is https://github.com/aboul3la/Sublist3r . I recently submitted, and the Sublist3r author @aboul3la merged, a [pull request](https://github.com/aboul3la/Sublist3r/pull/38) to make Sublist3r produce subdomains in the sort order that we prefer. @aboul3la [likes HTTPS Everywhere](https://github.com/aboul3la/Sublist3r/pull/38#issuecomment-251431856). 

My question here is whether we want to officially recommend Sublist3r in our documentation, pull requests, and issues for contributors to find new subdomains. "Officially recommend" is the difference between "One tool you might try using is Sublist3r" and "We recommend Sublist3r".

I don't mean to suggest that we _require_ contributors to use Sublist3r.

Thoughts? Pinging @fuglede @Hainish @J0WI but anyone should feel free to comment.

jeremyn:
If we want to recommend Sublist3r, then we will need to discuss how to configure it. Below is an example.

In PR https://github.com/EFForg/https-everywhere/pull/6888 we want to investigate https://nav.gov.hu , which is for a governmental department in Hungary. By default Sublist3r runs against not just regular search engines but DNS and security sites as well. This generates a lot of sites [1] that are not appropriate targets, such as https://cms-test-b.ekaer.nav.gov.hu and https://mx.ekaer.nav.gov.hu . On the other hand, if we limit the search to just Baidu, Google, and Bing by [modifying the script](https://github.com/aboul3la/Sublist3r/blob/cab552215e2dbaddc8287db91446655c3feaae60/sublist3r.py#L1052), we get a shorter list [2] but it's missing the obvious `www.nav.gov.hu`.

<details><summary>[1]</summary>
>[-] Enumerating subdomains now for nav.gov.hu
[-] Searching now in Baidu..
[-] Searching now in Yahoo..
[-] Searching now in Google..
[-] Searching now in Bing..
[-] Searching now in Ask..
[-] Searching now in Netcraft..
[-] Searching now in DNSdumpster..
[-] Searching now in Virustotal..
[-] Searching now in ThreatCrowd..
[-] Searching now in SSL Certificates..
[...]

PR #7323, opened 2016-10-12
Enable for www.montereybayaquarium.org

illicitonion:
The cert is only valid for www., so exclude the bare host.

### DIFF
--- a/src/chrome/content/rules/Monterey_Bay_Aquarium.xml
+++ b/src/chrome/content/rules/Monterey_Bay_Aquarium.xml
@@ -1,13 +1,7 @@
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://montereybayaquarium.org/ => https://montereybayaquarium.org/: (51, "SSL: no alternative certificate subject name matches target host name 'montereybayaquarium.org'")
-	!www: cert only matches www
+<ruleset name="Monterey Bay Aquarium">
 
--->
-<ruleset name="Monterey Bay Aquarium" default_off='failed ruleset test'>
-
-	<target host="montereybayaquarium.org" />
 	<target host="www.montereybayaquarium.org" />
+	<exclusion pattern="^http://montereybayaquarium.org/.*" /> <!-- Cert only matches www -->
 
 
 	<securecookie host="^www\.montereybayaquarium\.org$" name=".+" />
--- a/src/chrome/content/rules/Monterey_Bay_Aquarium.xml
+++ b/src/chrome/content/rules/Monterey_Bay_Aquarium.xml
@@ -1,12 +1,7 @@
 <ruleset name="Monterey Bay Aquarium">
-
 	<target host="www.montereybayaquarium.org" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for the ruleset. If there is no `target` for a domain like `montereybayaquarium.org`, then it won't be matched and you don't need to add an `exclusion` for it.

jeremyn:
Thanks. Please make the following changes. The ruleset should look something like [this one](https://github.com/jeremyn/https-everywhere/blob/245f2f8812d56a7c16d4982bf59b97252cfac0b1/src/chrome/content/rules/Juno.com.xml) when you're done.

- [ ] https://montereybayaquarium.org
- [ ] https://tmp.montereybayaquarium.org

Mark as `Time out`:
- [ ] https://static.montereybayaquarium.org
- [ ] https://storage.montereybayaquarium.org

Add `target`:
- [ ] https://affiliate.montereybayaquarium.org
- [ ] https://m.montereybayaquarium.org
- [ ] https://mobile.montereybayaquarium.org
- [ ] https://newsroom.montereybayaquarium.org
- [ ] https://tickets.montereybayaquarium.org

Also:
- [ ] Delete [this 'test'](https://github.com/illicitonion/https-everywhere/blob/9969a6a0c35aaea02e5ee12b196fec267f41e948/src/chrome/content/rules/Monterey_Bay_Aquarium.xml#L8), because there is an implicit test created for the `target` already. See [the style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) for more information.
- [ ] Change [the `securecookie` `host`](https://github.com/illicitonion/https-everywhere/blob/9969a6a0c35aaea02e5ee12b196fec267f41e948/src/chrome/content/rules/Monterey_Bay_Aquarium.xml#L4) to `^(www|affiliate|m|mobile|newsroom|tickets)\.montereybayaquarium\.org`

Finally, we like to [squash and merge](https://help.github.com/articles/about-pull-request-merges/) pull requests, which means we combine and rewrite pull requests in your name. Is that all right with you for this and any other pull requests you submit to us?

illicitonion:
Thanks for taking the time :) I think everything's done now.

Happy to be squashed and merged.

For future reference, how did you generate that list of domains?

jeremyn:
Thanks, I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7323/commits/6edf1dc84236aab0a3c0fa15f80f4e44d7c6a4d6 . The `securecookie` `host` should have a trailing `$`, which was not in my original checklist, so it's my fault you didn't have it. I've updated the checklist accordingly.

Also we normally only have e.g. `tmp.montereybayaquarium.org` in the top comment, not `https://tmp.montereybayaquarium.org` so please change that. I include `https://` to make it easy to click on from the GitHub comment but I understand how that might be confusing.

You can find those domains using the [Google search trick](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md), [Sublist3r](https://github.com/EFForg/https-everywhere/issues/7277), and knowing that if `m.` is a domain, then `mobile.` might be too.

jeremyn:
Thanks, merged.

Issue #7319, opened 2016-10-11
write.amnestyusa.org uses a self-signed cert


austin987:
http://write.amnestyusa.org/signup/ redirects to https://write.amnestyusa.org/signup/

That site currently uses a self signed certificate:
write.amnestyusa.org uses an invalid security certificate. The certificate is not trusted because it is self-signed. The certificate is only valid for selfsigned.cloudwaysapps.com Error code: SEC_ERROR_UNKNOWN_ISSUER

I erroneously contacted Amnesty about it, so they are aware, but consider the site content not important enough for SSL. I've asked them to reconsider, but in any case, for now, it's broken.

fuglede:
Hm, thanks for the report, but `Amnesty_USA.xml` contains no rules for write.amnestyusa.org, and it never has, so HTTPS Everywhere should not be rewriting that request. Do you perhaps have a custom ruleset, or could a different add-in be causing the redirect?

austin987:
Hi @fuglede,

No, I don't have any custom rulesets. I see this in Firefox and Chromium, where I only have uBlock Origin. Disabling uBlock makes no difference.

The original URL (from an email) was http://www.e-activist.com/ea-action/broadcast.record.message.click.do?ea.url.id=742999&ea.campaigner.email=4DvyaN3Tko8gYHTECbwHV5ku6OPFCMj%2F&ea.campaigner.id=MHk1zxzJpriZkArzVWMSmA==&ea_broadcast_target_id=0

which, if I have HTTPS-E disabled, goes to http://write.amnestyusa.org/signup/. Perhaps some redirect magic is at play?

jeremyn:
There's an [`e-activist.com`](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/rules/E-activist.com.xml) rule that rewrites the URL in the previous comment from `http` to `https`. I guess the people who run e-activist.com have it set up so that if their URL is https, the target URL is https, and the same for http.

fuglede:
Great, thanks for the added details; a fix is included in #7324.

schutzsmith:
Thank you for bringing this to our attention @austin987! 

We're implementing a wildcard SSL for all of our properties in the next week.

PR #7324, opened 2016-10-12
[E-activist.com] Add exclusion

fuglede:
This fixes https://github.com/EFForg/https-everywhere/issues/7319

### DIFF
--- a/src/chrome/content/rules/E-activist.com.xml
+++ b/src/chrome/content/rules/E-activist.com.xml
@@ -11,7 +11,7 @@
 	<!-- Fixes https://github.com/EFForg/https-everywhere/issues/7319 -->
 	<exclusion pattern="^http://www\.e-activist\.com/ea-action/" />
 
-		<test url="http://www.e-activist.com/ea-action/broadcast.record.message.click.do?ea.url.id=742999&amp;ea.campaigner.email=4DvyaN3Tko8gYHTECbwHV5ku6OPFCMj%2F&amp;ea.campaigner.id=MHk1zxzJpriZkArzVWMSmA==&amp;ea_broadcast_target_id=0" />
+		<test url="http://www.e-activist.com/ea-action/broadcast.record.message.click.do?ea.url.id=742999" />
 
 
 	<!--	Not secured by server:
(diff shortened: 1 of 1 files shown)


jeremyn:
The test URL http://www.e-activist.com/ea-action/broadcast.record.message.click.do?ea.url.id=742999&amp;ea.campaigner.email=4DvyaN3Tko8gYHTECbwHV5ku6OPFCMj%2F&amp;ea.campaigner.id=MHk1zxzJpriZkArzVWMSmA==&amp;ea_broadcast_target_id=0 , which you got from https://github.com/EFForg/https-everywhere/issues/7319#issuecomment-253039002 , looks like it has personal identification stuff in it . Can you substitute a more generic URL from [Google](https://www.google.com/#q=site:www.e-activist.com%2Fea-action)?

jeremyn:
I agree with the test URL change.

Do you think we should remove the general `securecookie` given this new exclusion?

fuglede:
On 12 October 2016 19:10:10 CEST, Jeremy Nation <notifications@github.com> wrote:
>I agree with the test URL change.
>
>Do you think we should remove the general `securecookie` given this new
>exclusion?
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/EFForg/https-everywhere/pull/7324#issuecomment-253276132

I did consider it briefly, but it looks like it should be safe? If the URLs matching the pattern all come from email newsletters, I think it should be, at least.

jeremyn:
The failure case is if the user already has a secure `www.e-activist.com` cookie set in their browser, and the email URL is supposed to read it, but can't.

fuglede:
Or for `e-activist.com` (which actually happens, even though the site serves nothing but redirects); it does seem unlikely to play any role here though, doesn't it?

jeremyn:
Yeah, it would also be a problem if the `http` exclusion URL tried to read a secure `^e-activist.com` cookie.

Following the discussion in https://github.com/EFForg/https-everywhere/issues/2907 , I'm seeing fewer good cases for a wildcard `securecookie`. I would get rid of it for everything but the tiniest, simplest sites. But for this issue we can do whatever you prefer.

fuglede:
Having no particularly strong opinion on the matter, I went ahead and removed the `securecookie`.

jeremyn:
Thanks, merged.

PR #7326, opened 2016-10-12
Disable CJFE ruleset.

jsha:
Per bad cert, plus request by site owner:
https://trac.torproject.org/projects/tor/ticket/20253#comment:3

### DIFF
--- a/src/chrome/content/rules/CJFE.org.xml
+++ b/src/chrome/content/rules/CJFE.org.xml
@@ -14,7 +14,7 @@
 	* Secured by us
 
 -->
-<ruleset name="CJFE.org">
+<ruleset name="CJFE.org" default_off="invalid cert">
 
 	<!--	Direct rewrites:
 				-->


fuglede:
LGTM, but as a side note: it doesn't actually matter much for this particular one since it's broken on its own (well, except for https://snowdenarchive.cjfe.org/ which appears to work), and it looks like they didn't explicitly request the ruleset to be disabled, but this is the first case I've come across, where it's not actually obvious that they are in fact the webmaster of the site.

jsha:
> the first case I've come across, where it's not actually obvious that they are in fact the webmaster of the site.

Interesting! Are you saying that mainly because the Trac ticket was filed by an anonymous person? I agree that we don't have solid proof, though in general it sounds legit. Fortunately most "site owner" request for disablement are accompanied by ready proof of the ruleset's brokenness, so we haven't had to develop a more formal verification method yet.

CJFE:
I'm quite definitely the webmaster of the site. I'm unclear how this ruleset change takes effect. The code is changed and pushed to the git repository, does my chrome extension automatically update from that repository? Still turning up an error when I visit the site with HTTPS everywhere on, actually, worse now because the browser bar is automatically dropping the www. prefix which causes the page to not even be found. 

Thanks for the help there. This has been a persistent problem for our organization for years I am told, ever since we switched off of an old site (likely wordpress) which had a cert to a CMS/Hosting combined service (nationbuilder) which only offers a site-wide cert for its enterprise ($1000/month) customers. We still have the certs on the Snowden archives (hosted externally) and our payment pages (only cert provided by the CMS service) afaik.

Can you confirm that the problem was with the ruleset still listing us as accepting site-wide certs?

jsha:
@CJFE Thanks for the extra confirmation! Once a ruleset change is pushed to the master branch of the repository, it will be rolled into the next release, which happens approximately every 2-4 weeks. So expect to see this change go out with the next release, on that timeline. You can follow @HTTPSEverywhere on Twitter for updates.

> Can you confirm that the problem was with the ruleset still listing us as accepting site-wide certs?

More specifically, the problem was that the ruleset listed your site as having a valid HTTPS configuration, when it no longer does, as you describe above. Specifically if you visit https://www.cjfe.org/ in Chrome, and click on Advanced, you will see that the certificate is for `a248.e.akamai.net`, which is a mismatch with the actual name. This is a common problem for sites hosted with Akamai that don't specifically request a certificate. Note that Akamai does offer Let's Encrypt certificates now (though I'm not sure on pricing). I would recommend filing a support ticket with NationBuilder for them to offer free HTTPS using Let's Encrypt certificates, so you can get your site working with HTTPS again.

If you'd like to chat more about enabling HTTPS for your site, I'd be happy to talk about it! Feel free to email me - contact information at https://www.eff.org/about/staff/jacob-hoffman-andrews.

fuglede:
>Interesting! Are you saying that mainly because the Trac ticket was
>filed by an anonymous person?

Yep, exactly.

> I agree that we don't have solid proof,
>though in general it sounds legit. Fortunately most "site owner"
>request for disablement are accompanied by ready proof of the ruleset's
>brokenness, so we haven't had to develop a more formal verification
>method yet.

They are also so few in number that we might not even need a method for a while, and it's possible to handle on a case-by-case basis if necessary.

PR #7331, opened 2016-10-13
[Fansub.co] Add ruleset

synthtech:


src/chrome/content/rules/Fansub.co.xml
```diff
@@ -0,0 +1,15 @@
...
+
+	Refused:
+		- iseng.fansub.co
+		- oyatsu.fansub.co
+-->
+<ruleset name="Fansub.co">
+	<target host="fansub.co" />
+	<target host="www.fansub.co" />
+
+	<rule from="^http://(www\.)?fansub\.co/" to="https://fansub.co/" />
```
J0WI: You don't need to force non-www, this works fine server side.

synthtech: Fixed

J0WI:
Thanks!

Issue #7322, opened 2016-10-12
Always connects to check.tpo when Torbutton is installed, even if Observatory has been disabled


rustybird:
If the Torbutton Firefox extension is installed, HTTPS Everywhere will connect to `https://check.torproject.org` to test if SSL Observatory submissions can be torified. But it will do so even when the Observatory has been disabled, which is the Tor Browser default.

Looks like the only way to avoid this connection is by setting the following two prefs (to [disable the `tor_safe` variable](https://github.com/EFForg/https-everywhere/blob/ea5703d84a96067c4e1e0ed7cf0ec983fe404686/src/components/ssl-observatory.js#L879) and [hence the Tor check](https://github.com/EFForg/https-everywhere/blob/ea5703d84a96067c4e1e0ed7cf0ec983fe404686/src/components/ssl-observatory.js#L781)):

```
user_pref("extensions.https_everywhere._observatory.use_custom_proxy", true);
user_pref("extensions.https_everywhere._observatory.use_tor_proxy", false);
```

Hainish:
#7322 will fix this particular issue.

PR #7342, opened 2016-10-14
Do not use Tor Browser proxy settings directly, let TB handle network…

Hainish:
… connections

This resolves the issue in https://trac.torproject.org/projects/tor/ticket/20195 where the SSL Observatory proxy checking code and submissions were bypassing domain isolation.  That code was a relic from the TorButton days.

Now, check.torproject.org is no longer accessed when we're using Tor Browser, we assume successful Tor access.  In this case, we let TB transparently proxy for us, instead of accessing the Tor Browser proxy settings directly.

### DIFF
--- a/src/components/ssl-observatory.js
+++ b/src/components/ssl-observatory.js
@@ -778,7 +778,16 @@ SSLObservatory.prototype = {
      */
     this.proxy_test_successful = null;
 
-    if (this.getProxySettings().tor_safe == false) {
+    var proxy_settings = this.getProxySettings();
+    // if proxy_settings is false, we're using tor browser for sure
+    // if tor_safe is false, the user has specified use_custom_proxy
+    // in either case, don't issue request to tor check url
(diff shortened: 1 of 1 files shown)


Hainish:
@cooperq ready for review

jeremyn:
@Hainish Is this related to issue https://github.com/EFForg/https-everywhere/issues/7322 ?

Hainish:
@jeremyn this is related and will fix the particular issue of Tor Browser connecting to `check.torproject.org` when the observatory is disabled.  If a user has the Tor daemon installed separately and is not using Tor Browser, the check will still be made over the Tor socks port even if the observatory is disabled.

This check is necessary to determine whether the "Check certificates using Tor for anonymity" radio button should be disabled or not.  An optimization would be to only perform the check once "Use the observatory" is checked (or at startup if it is already checked) instead of always checking at startup.

cooperq:
LGTM

Issue #7312, opened 2016-10-10
Recode.net (partial) breaks images


terrorist96:
http://www.recode.net/2015/12/21/11621610/is-your-messaging-app-encrypted

`Recode.net (partial)` breaks images

jeremyn:
Please clarify what's broken.

terrorist96:
Images under "TRENDING" don't load.

jeremyn:
What does "don't load" mean for you, exactly?

terrorist96:
![1](https://cloud.githubusercontent.com/assets/12802056/19403881/103f27cc-9238-11e6-9f14-9b089c8b3cd4.PNG)
vs.
![2](https://cloud.githubusercontent.com/assets/12802056/19403880/103d0d66-9238-11e6-81bc-e105096937c7.PNG)

mnordhoff:
I filed #7381 about this also affecting Racked.com. The Polygon.com ruleset was fixed a few months ago. It sounds like all Vox properties need to be checked for an `optimally_sized_images` endpoint.

J0WI:
WFM. They finally use HTTPS now.

PR #7344, opened 2016-10-14
Update Recode.net.xml

jeremyn:
This fixes https://github.com/EFForg/https-everywhere/issues/7312, removes non-capturing group markers in the exclusion, and removes some comments. For reference, #7312 seems to be caused by a CORS problem related to:

http://www.recode.net/services/optimally_sized_images?imgkeys=51313827:standard:1:268x268:jpg,51313831:standard:1:268x268:jpg,51314537:standard:1:268x268:jpg,51329485:standard:1:268x268:jpg,51331885:standard:1:268x268:jpg,51358883:*:1:235x176:jpg,51359343:*:1:235x176:jpg,51359435:*:1:235x176:jpg,51360153:*:1:235x176:jpg&asset_keys=

### DIFF
--- a/src/chrome/content/rules/Recode.net.xml
+++ b/src/chrome/content/rules/Recode.net.xml
@@ -1,40 +1,13 @@
 <!--
 	For other Vox Media coverage, see Vox.com.xml.
 
-
-	Insecure cookies are set for these hosts: ᶜ
-
-		- www.recode.net
-
-	ᶜ See https://owasp.org/index.php/SecureFlag
-
 -->
 <ruleset name="Recode.net (partial)">
 
 	<target host="recode.net" />
 	<target host="www.recode.net" />
 
-		<!--	Redirects to http:
-						-->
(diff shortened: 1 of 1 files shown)


J0WI:
Could you leave some comments why the exclusion is needed?

jeremyn:
I don't understand them myself, the exclusion comments are weird. The old comments construct a different exclusion than it had, for example what happened to `authors` and the nested negative lookaheads? What do `Avoid potential XHR problems` and `Lessen non-Tor distinguishability` mean? What is going on with `style`?

J0WI:
I would just keep them.

jeremyn:
I feel somewhat strongly that we should delete outdated or misleading comments when we can.

J0WI:
Sure, but we still have them in the exclusion pattern, so they are not outdated.

jeremyn:
The comments don't exactly match the exclusion pattern so they should be updated, but I don't know what the comments are talking about so I'm not comfortable updating them. So, instead I've deleted them.

jeremyn:
I'm closing this as a duplicate of either pull request #9583 or #9700 .

PR #7345, opened 2016-10-14
[JW_Platform.com.xml] exclusion to fix #5086

### LINKED ISSUE #5086: JW Platform.com (partial) breaks video (on Chrome)
http://dailycaller.com/2016/03/30/jane-sanders-hillary-tries-to-get-votes-from-people-at-the-worst-time-in-their-life-video/

JW Platform.com (partial) breaks video

Another example: http://www.wnd.com/2016/04/trump-erupts-as-cruz-sweeps-colorado-without-votes/
breaks both videos. First video is also broken by Google API. #4545
#4531

@J0WI

J0WI:


jeremyn:
I can't reproduce the original video problem from issue #5086 on either http://www.realclearpolitics.com/video/2016/07/11/trump_i_am_the_law_and_order_candidate.html or https://deadspin.com/lebron-carmelo-wade-cp3-open-espys-with-statement-on-1783626133. Do we still need this fix?

Issue #7313, opened 2016-10-11
dreamhost.com and the possibility of new subdomains over time


galenide:
With the dreamhost.com domain, it appears that there are a number of subdomains under which third party sites are hosted, such as http://mug.dreamhost.com/, http://www.otismaxwell.dreamhost.com/, http://ovaltine.dreamhost.com/, http://crookedtimber1.dreamhost.com/, and http://apache2-xenon.shasta.dreamhost.com/. In addition, it appears that these subdomains for third party sites do not necessarily support HTTPS. As of this writing, the DreamHost ruleset has a wildcard rule for redirecting all subdomains by default to HTTPS, along with an exclusion list for subdomains that do not support HTTPS. From what it would seem, if it is possible over time for subdomains to be added under dreamhost.com for the hosting of third party sites, and with the subdomains not supporting HTTPS, then it may well be more useful for the DreamHost ruleset to have an explicit list of domains that should be redirected to HTTPS, and to only redirect domains on the list.

mnordhoff:
Hostnames like "apache2-xenon.shasta.dreamhost.com" are for a specific Apache 2 instance (xenon) on a specific shared web server (shasta). I'm not sure breaking internal infrastructure that shouldn't really be accessed by the public or at all is much of a problem.

(I'm not sure what all of those hostnames are, though.)

PR #7352, opened 2016-10-15
[DreamHost] Added exclusion for links subdomain, added rule for blog...

galenide:
...subdomain, removed exclusion for transferapproval subdomain, expanded observed subdomains list

### DIFF
--- a/src/chrome/content/rules/DreamHost.xml
+++ b/src/chrome/content/rules/DreamHost.xml
@@ -17,6 +17,16 @@
 		- dreamhoststatus.com			(refused)
 		- www.dreamhoststatus.com		(cert mismatch)
 
+	Note: It appears that the dreamhost.com domain has a number of subdomains
+	under which third party sites are hosted, such as http://mug.dreamhost.com/,
+	http://crookedtimber1.dreamhost.com/, and http://ovaltine.dreamhost.com/. In
+	addition, it appears that these subdomains for third party sites do not
+	necessarily support HTTPS. Given the possibility that the dreamhost.com
+	domain may have subdomains for third party hosted sites added or removed over
+	time, it is advisable for the DreamHost ruleset to have an explicit list of
+	target domains that are redirected to HTTPS instead of a wildcard
+	*.dreamhost.com rule combined with exclusions for nonfunctional domains.
+
 
 	Mixed content:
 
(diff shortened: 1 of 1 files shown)


galenide:
@jeremyn From what it appears, the dreamhost.com domain has a number of subdomains which do not support HTTPS (such as links and whoisweb) and for which the current DreamHost ruleset redirects to HTTPS. As such, this revised version of the ruleset may be of some importance as far as reducing the chance of users encountering errors due to HTTPS Everywhere.

J0WI:
This seems easy to convert from a wildcard rule to a list of targets.
@galenide would you like to do this? This simplifies the rule a lot.

J0WI:
See also #7313

src/chrome/content/rules/DreamHost.xml
```diff
@@ -28,39 +28,38 @@
...
 
 
-	<securecookie host="^(?:.*\.)?dreamhost\.com$" name=".+" />
+	<test url="http://media-dh-com.objects.dreamhost.com/mp4/player.swf" />
+
+
+	<securecookie host="^(\.?(abuse|atmail|discussion|gifts|media|panel|roundcube|signup|transferapproval|webftp|webmail|www))\.dreamhost\.com$"
+			name=".+" />
+	<securecookie host="(^|\.)objects\.dreamhost.com$"
+			name=".+" />
```
J0WI: You can also use `<securecookie host=".+"	name=".+" />` to match all targets.

src/chrome/content/rules/DreamHost.xml
```diff
@@ -28,39 +28,38 @@
...
+	<target host="www.dreamhost.com" />
+	<target host="abuse.dreamhost.com" />
+	<target host="atmail.dreamhost.com" />
+	<target host="blog.dreamhost.com" />
+	<target host="discussion.dreamhost.com" />
+	<target host="gifts.dreamhost.com" />
+	<target host="help.dreamhost.com" />
+	<target host="media.dreamhost.com" />
+	<target host="objects.dreamhost.com" />
+	<target host="*.objects.dreamhost.com" />
```
J0WI: Wildcard Hosts need at least 3 test urls.

J0WI:
Thanks!

PR #7374, opened 2016-10-19
[Zammad] New rulesets

rugk:


J0WI:
@rugk can you add comments on both rulesets showing their relation?

Issue #7381, opened 2016-10-20
Racked.com breaks images due to CORS issues


mnordhoff:
The [`Racked.com` ruleset](https://github.com/EFForg/https-everywhere/blob/88f83c2e0e877ca288ad25f6f258f3c70ad2c6bb/src/chrome/content/rules/Racked.com.xml) breaks all the images on (some?) articles due to CORS issues with XMLHttpRequest and `www.racked.com`.

URL: http://www.racked.com/2016/10/5/13165868/donald-trump-china-factories-neckties

Error: `XMLHttpRequest cannot load https://www.racked.com/chorus_page/optimally_sized_images?imgkeys=51193119%3A*%3A18%3A1757x975%3Awebp&asset_keys=7219403%3A440%3A440%2C7219513%3A1317%3A1317%2C7219619%3A440%3A440%2C7219669%3A1317%3A1317. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://www.racked.com' is therefore not allowed access.`

Do you think it would be better to add an exclusion or disable the ruleset entirely? The ruleset is of literally no use if the main page URL redirects to HTTP, the image thing is excluded, and all of the other elements are loaded from third-party domains, most of which are probably covered by other rulesets. Other pages may be in less dire shape, though.

mnordhoff:
#4007 and #7312 are about the same issue on other Vox properties.

PR #7383, opened 2016-10-20
[Racked.com] Default off (CORS issues). racked.com isn't problematic.

mnordhoff:
- Default off (CORS issues). #7381.
- `racked.com` is no longer problematic.

So it's good news and bad news. But mostly bad news.

As i wrote in #7381, it would be possible to add an exclusion instead of turning off the ruleset entirely, but it would make the ruleset entirely useless (at least on some pages).


J0WI:
The only securable resource would be a CSS which is loaded by HTTPS anyway. All other resources are covered by other Vox Media rulesets, so I'm fine with disabling it.

PR #7385, opened 2016-10-20
Create dpd.co.uk.xml

### LINKED ISSUE #7359: Add dpd.co.uk
Valid:
careers.dpd.co.uk
dpd.co.uk
drivers.dpd.co.uk
www.dpd.co.uk

HTTPS bugs:
help.dpd.co.uk
m.dpd.co.uk
qa-www.dpd.co.uk
sales.dpd.co.uk
www.careers.dpd.co.uk

gloomy-ghost:
Close #7359

### DIFF
--- a/src/chrome/content/rules/dpd.co.uk.xml
+++ b/src/chrome/content/rules/dpd.co.uk.xml
@@ -1,0 +1,33 @@
+<!--
+	Mismatch:
+		- www.careers.dpd.co.uk
+		- help.dpd.co.uk
+		- help-sae.dpd.co.uk
+		- m.dpd.co.uk
+		- sales.dpd.co.uk
+
+	Timeout:
+		- jobs.dpd.co.uk
+		- qa-www.dpd.co.uk
+
+	Insecure cookie:
+		- dpd.co.uk:
+			_ga, _gat, optimizelyBuckets, optimizelyEndUserId, optimizelyPendingLogEvents, optimizelySegments
+		- caeers.co.uk:
+			ASP.NET_SessionId, CultureSystemCodeID, CultureSystemLanguageID
+		- optimizely.com:
+			cdn
+-->
+<ruleset name="dpd.co.uk">
+	<target host="dpd.co.uk" />
+	<target host="www.dpd.co.uk" />
+	<target host="careers.dpd.co.uk" />
+		<target host="www.careers.dpd.co.uk" />
+		<rule from="^http://www\.careers\.dpd\.co\.uk/"
+			to="https://careers.dpd.co.uk/" />
+	<target host="drivers.dpd.co.uk" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
--- a/src/chrome/content/rules/dpd.co.uk.xml
+++ b/src/chrome/content/rules/dpd.co.uk.xml
@@ -5,18 +5,6 @@
 		- help-sae.dpd.co.uk
 		- m.dpd.co.uk
 		- sales.dpd.co.uk
-
-	Timeout:
-		- jobs.dpd.co.uk
-		- qa-www.dpd.co.uk
-
-	Insecure cookie:
-		- dpd.co.uk:
-			_ga, _gat, optimizelyBuckets, optimizelyEndUserId, optimizelyPendingLogEvents, optimizelySegments
-		- caeers.co.uk:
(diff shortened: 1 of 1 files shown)


jeremyn:
Please:

- [ ] Delete the [`Timeout`](https://github.com/gloomy-ghost/https-everywhere/blob/74feb36e68b47dd75804344239b4722348b2a989/src/chrome/content/rules/dpd.co.uk.xml#L9-L12) comment, because these time out even with unencrypted HTTP
- [ ] Delete the [`Insecure cookie`](https://github.com/gloomy-ghost/https-everywhere/blob/74feb36e68b47dd75804344239b4722348b2a989/src/chrome/content/rules/dpd.co.uk.xml#L13-L19) comment
- [ ] Either delete the `securecookie` or specifically list the subdomains it covers aside from `dpd.co.uk`.

This is the second recent pull request from you that you've had the `Insecure cookie` comment where I've asked you to delete it. I've seen these in other rulesets and I generally don't like them, but do you think we need them?

gloomy-ghost:
The `insecure cookie` is for `securecookie` coverages...

`<securecookie host=".+" name=".+" />` doesn't break anything here. I added these comments since you were asking `specifically list the subdomains` (maybe I misunderstood?)

jeremyn:
Sorry, I wasn't clear. A problem with `.+`/`.+` is this: imagine a site with `a.example.com` and `b.example.com`, where `a` supports HTTPS and `b` doesn't. `a` can set a cookie for `example.com` that `b` needs to read, but if HTTPS Everywhere has `.+`/`.+`, that cookie might be marked secure only, and now `b` can't read it and maybe some important functionality breaks. See the discussion in https://github.com/EFForg/https-everywhere/issues/2907 for more information.

So, I don't like `.+`/`.+` unless most or all subdomains are covered without any `exception`s. The fix then is to either get rid of `securecookie` or to specifically list the subdomains in the `host` like you see in, for example, [`Microsoft.xml`](https://github.com/EFForg/https-everywhere/blob/fc3e2b0c0e9f567dea056108f72d2e317e437ff6/src/chrome/content/rules/Microsoft.xml#L614).

gloomy-ghost:
I'm more confused after read #2909... 

1. if `login.example.com` set a session cookie for its root and `www.example.com` doesn't have https, we have to send cookie in http anyway.
2. If `login.example.com` set a option cookie for itself (e.g. `lan=en`), it doesn't really matter it's secured or not since all `login.` requests are rewritten.
3. Furthermore, we can't even tell if there is another subdomain (e.g. `verify.login.example.com`) needs read cookie under `login.`

So when is the right time to use `securecookie`? Webmasters won't be happy if the answer is tracking cookie under top domain 😆

jeremyn:
I agree, `securecookie` has limited use. https://github.com/EFForg/https-everywhere/issues/2907#issuecomment-245365536 is a good explanation of what problems it solves.

Hainish:
@gloomy-ghost sorry for the off-topic message, could I contact you in private?  I don't have an email address listed for you so I have to use an open issue.  Could you email me at bill@eff.org?

gloomy-ghost:
>  It's possible for a network attacker to trigger an HTTP request to nonexistent.example.com, which may send cookies.

I think that the middle man doesn't need to trigger any http requests since unsecured cookies are sent with the first http request (either `example.com` or `subdomain.example.com`). If there is a subdomain which doesn't support http, the cookies leak anyway. (Technically a cookie is set for its top domain means they are required in other subdomains. When we secure a cookie, users won't be able to log in if it's a session cookie, and webmasters won't have a accurate report if it's a tracking cookie.)

jeremyn:
The ruleset is good, merged.

The `nonexistent.example.com` comment was in response to the idea that `securecookie` isn't needed if the all of the (real) subdomains of `example.com` are secured.

Here is an attack: `example.com` and all active subdomains only support HTTPS. `example.com` sets a top-level cookie that is not secure-only. A hostile attacker with man-in-the-middle access wants to see a user's `example.com` cookie. Because `example.com` and subdomains only supports HTTPS, all real requests to `example.com` and subdomains are encrypted and the cookie is safe. So, the attacker waits until the user makes an unencrypted HTTP request to any site in the world, and then injects a request for `nonexistent.example.com/fake.css` into the unencrypted HTTP response. Assuming the browser can resolve `nonexistent.example.com`, it makes a request to it, sending the unencrypted `example.com` cookie, and now the attacker can see it.

gloomy-ghost:
Well, I get it now.

@Hainish I've sent you an email.

PR #7401, opened 2016-10-21
Update Pingupod.de.xml

tamoor-cliqz:


J0WI:
Cert has been renewed.

PR #7405, opened 2016-10-21
Update lezhin.com.xml

SaschaNaz:
Some hosts now has valid certifications, social.lezhin.com now supports HSTS, and there is a new host.


J0WI:
Let's trigger Travis again since test didn't run here.

src/chrome/content/rules/lezhin.com.xml
```diff
@@ -1,19 +1,19 @@
...
 	link.lezhin.com
 	-->
 	
 	<target host="lezhin.com" />
 	<target host="www.lezhin.com" />
 	<target host="m.lezhin.com" />
+	<target host="desk.lezhin.com" />
+	<target host="cdn.lezhin.com" />
+	<target host="acdn.lezhin.com" />
+	<target host="ncdn.lezhin.com" />
```
J0WI: ggtoon.lezhin.com
social.lezhin.com

It seems that there are also mirrors like:
\d+cdn.lezhin.com
\d+www.lezhin.com

SaschaNaz: 1. I think social.lezhin.com has HSTS header, should it still be added?
2. What does `\d+` mean?

SaschaNaz: I'm also curious how you are getting additional hosts.

J0WI: Yes, as long as hosts are not HSTS preloaded we include them in the rules to secure the initial requests. Also some users (including Tor Browsers) don't cache HSTS.
See https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md for a guide about finding subdomains.

J0WI:
You can't use regex that way in a target. I would propose to just document this behavior since this mirrors are not really in use.

SaschaNaz:
I tried adding some rules for them, do you want me to replace them with comments?

PR #7417, opened 2016-10-23
[ACLU of Florida] New ruleset

galenide:


jeremyn:
Please:

- [ ] Add a reference to this ruleset pointing to [`ACLU.xml`](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/rules/ACLU.xml), and from `ACLU.xml` to this ruleset ([example](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/rules/ACLU_of_Massachusetts.xml#L1-L6))
- [ ] Add target for https://northcentral.aclufl.org
- [ ] Remove `(^|\.)` from the `securecookie` tags, see [here](https://www.eff.org/https-everywhere/rulesets#secure-cookies)
- [ ] I understand you split the `securecookie` hosts because of the earlier target splits, but for simplicity, let's combine them into one `securecookie` tag
- [ ] I don't like the special `media` rules. Where these pdfs are linked from?

Is squash-and-merge okay?

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7417/commits/c5fadcd3dc8c7f5efbabf5b0aa406febb7b6ac8c .

galenide:
The PDFs that are specified in special rules were likely found via one or more Google searches. I do not know if they are linked to from any ACLU sites.

On the issue of adding a reference to this ruleset in `ACLU.xml`, pull request #7498 is relevant.

For this pull request, squash and merge should be okay.

jeremyn:
I see what you mean, but I don't like having file-specific rules. It looks like a rewrite from `media.aclufl.org` to `media.aclufl.org/media` should be fine. What do you think?

galenide:
The rule for `media.aclufl.org` with the regexp `^http://media\.aclufl\.org/(media(/|$))?` is intended to work as follows. The specific URLs http://media.aclufl.org/media and http://media.aclufl.org/media/ in addition to URLs that start with `http://media.aclufl.org/media/` essentially need to have the `http://` changed to `https://` instead. Other URLs under the media.aclufl.org domain need to have the prefix `http://media.aclufl.org/` changed to `https://media.aclufl.org/media/` instead.

When the ruleset is not enabled, a number of URLs under media.aclufl.org behave differently under HTTP versus HTTPS. For example, https://media.aclufl.org/media/2015Annual/2015-ACLU-Annual-Report-Web-Version.pdf works correctly, but http://media.aclufl.org/media/2015Annual/2015-ACLU-Annual-Report-Web-Version.pdf leads to a 302 redirect to https://media.aclufl.org/media/media/2015Annual/2015-ACLU-Annual-Report-Web-Version.pdf which leads to a 302 redirect to https://aclufl.org/wp-signup.php?new=media. (The resulting page states that "Registration has been disabled.") http://media.aclufl.org/ACLU%20Complaint%20Form%20-%20Spanish.pdf leads to a 302 redirect to 
https://media.aclufl.org/media/ACLU%20Complaint%20Form%20-%20Spanish.pdf which works correctly, but https://media.aclufl.org/ACLU%20Complaint%20Form%20-%20Spanish.pdf leads to a 302 redirect to https://aclufl.org/wp-signup.php?new=media instead.

It is possible that at the server end, HTTPS URLs under media.aclufl.org are expected to start with `https://media.aclufl.org/media/` (but not `https://media.aclufl.org/media/media/`) and that for HTTP URLs, the server does a redirect where `http://media.aclufl.org/` is changed to `https://media.aclufl.org/media/` instead, even if it makes the URL erroneous. A Google search for `site:media.aclufl.org` (including the use of the option to repeat the search and include omitted results) produces a number of search results for which all of the links are HTTPS URLs except for three.

jeremyn:
I agree with using

```xml
	<rule from="^http://media\.aclufl\.org/(media(/|$))?"
		to="https://media.aclufl.org/media/" />
```

Merged.

PR #7418, opened 2016-10-23
Create NZQA.govt.nz.xml

gloomy-ghost:


jeremyn:
Please:

- [ ] Clarify the `404` error like `404 (except for root, which redirects)` or similar
- [ ] Alphabetize the error groups like `404`, `Mismatch`, `Refused`, `Timeout`
- [ ] Move [these tests](https://github.com/gloomy-ghost/https-everywhere/blob/b6a538b7afecea4ecdab112038877d8c97a9573f/src/chrome/content/rules/NZQA.govt.nz.xml#L19-L20) and [these other tests](https://github.com/gloomy-ghost/https-everywhere/blob/b6a538b7afecea4ecdab112038877d8c97a9573f/src/chrome/content/rules/NZQA.govt.nz.xml#L22-L23) directly below [the rule they are testing](https://github.com/gloomy-ghost/https-everywhere/blob/b6a538b7afecea4ecdab112038877d8c97a9573f/src/chrome/content/rules/NZQA.govt.nz.xml#L25-L26), for readability.
- [ ] Combine [the `securecookie` tags](https://github.com/gloomy-ghost/https-everywhere/blob/b6a538b7afecea4ecdab112038877d8c97a9573f/src/chrome/content/rules/NZQA.govt.nz.xml#L38-L40) into `^(api-store|careers|secure)\.nzqa\.govt\.nz$`

gloomy-ghost:
I was sorting errors in their orders of happening:

Timeout (no reply)
Refused (port is not being used)
Reset (received RST)
Protocol error, Invalid cert (TCP established, TLS handshaking)
Non-2xx (HTTP error)
MCB (received page)

I'll push another commit if you think that alphabetizing is better.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7418/commits/44ef2e6ed6a0926c30cfbe4e87e32c22606d4846 .

That's a neat idea about sorting in order of the errors happening, I can see what you did that. A problem though is, that's too complicated to explain to contributors over and over. What really matters is that the order isn't simply random. So I think we should just do alphabetical ordering. Does that sound all right?

You can get rid of `\.?` in the new `securecookie`. The leading dot is assumed, see [here](https://www.eff.org/https-everywhere/rulesets#secure-cookies).

I also added a new item to the checklist for `?:` I understand why you added it but I think we should keep regexes as simple as possible.

jeremyn:
Looks good, merged.

PR #7419, opened 2016-10-23
[amazon.in.xml] fix #6157

### LINKED ISSUE #6157: https-everywhere causing issues on Amazon India website while updating zipcode.
On the Amazon India website selecting a new delivery location (entering a new pin code (zipcode) and clicking update) on the product page throws an error message !!
bug-1292467

1.Go to http://amazon.in on any Laptop or PC.
2.Search and select any product to buy.
3.Click "**Select a delivery address**" below "**Deliver to**:" label on the RHS.
4.A pop form is displayed with the option of signing in or just enter the PINCODE (Zipcode).
5.Enter any valid 6 digits PINCODE in India (Example : 560078, 440018)
6.Click update
**Actual response**: An error message is displayed: : "Please enter a valid 6 digit PINCODE !!

**Expected Response**:
The form should update the delivery address and display the delivery time for the entered new address (PINCODE).

This issue is not happening without the https-everywhere addon.
Firefox version : 48

J0WI:


jeremyn:
@J0WI Aside from the Travis problems: why is the `exclusion` for all of `/(gp|xa)` instead of just http://www.amazon.in/gp/delivery/ajax/address-change.html that was specifically in the issue?

J0WI:
There are some general issues with AJAX/XHR on amazon.in

J0WI:
This finally passes all tests.

J0WI:
@jeremyn could you have a look again here?

jeremyn:
- [ ] Add `test`s for the [first `exclusion`](https://github.com/J0WI/https-everywhere/blob/43ed8a739361b3374ce98590db97b894db16a009/src/chrome/content/rules/amazon.in.xml#L32) for the new `b` and `s`.
- [ ] Should the [second `exclusion`](https://github.com/J0WI/https-everywhere/blob/43ed8a739361b3374ce98590db97b894db16a009/src/chrome/content/rules/amazon.in.xml#L52) use `(www\.)?` instead of `www`?
- [ ] There should be a more specific explanation somewhere for the extra `exclusion`s that I asked about in https://github.com/EFForg/https-everywhere/pull/7419#issuecomment-279163042. I can't verify the specific need for those `exclusion`s, and maintainers in the future won't be able to tell if the `exclusion`s can be safely removed.

J0WI:
done

jeremyn:
I'm getting `404` errors when I try these `test` URLs in the browser:

http://www.amazon.in/xiomi-redmi-mi-mobiles/b?ie=UTF8&amp;node=6294306031
http://www.amazon.in/xa/dealcontent/v2/GetDealStatus
http://www.amazon.in/xa/dealcontent/v2/GetDeals

Also, the related `^` versions of the last two URLs get `301`, then `404`.

Can we use different URLs?

J0WI:
Out test infrastructure can only make GET requests which don't work for the Amazon API. I've added a comment about it.

jeremyn:
So you're saying that the test URLs like http://www.amazon.in/xa/dealcontent/v2/GetDealStatus will get a `404` in our tests, but still pass, and a `404` when testing in the browser's address bar, because they only work correctly with POST, is that right? That's fine.

Why did you add the `Select` URLs? They do the same thing the `Get` URLs do, so we haven't gained anything.

J0WI:
> So you're saying that the test URLs like http://www.amazon.in/xa/dealcontent/v2/GetDealStatus will get a 404 in our tests, but still pass, and a 404 when testing in the browser's address bar, because they only work correctly with POST, is that right?

Exactly.

> Why did you add the Select URLs? They do the same thing the Get URLs do, so we haven't gained anything.

We also haven't lost anything. ;)

jeremyn:
I know some people disagree but my opinion on test cases is that each test case should be there for a specific reason. It's not a big deal though.

PR #7423, opened 2016-10-24
Add questionablecontent.net

jayofdoom:
Also had to add an entry for an ad server they use, p.adsymptotic.com.

Issue #3069

### DIFF
--- a/src/chrome/content/rules/Questionablecontent.net.xml
+++ b/src/chrome/content/rules/Questionablecontent.net.xml
@@ -1,0 +1,10 @@
+<!--
+	No nonfunctional hosts in *.questionablecontent.net:
+-->
+<ruleset name="Questionablecontent.net">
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Questionablecontent.net.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+	No nonfunctional hosts in *.questionablecontent.net:
+-->
+<ruleset name="Questionablecontent.net">
+	<target host="questionablecontent.net" />
+	<target host="www.questionablecontent.net" />
```
J0WI: Please also add `forums.questionablecontent.net`

jayofdoom:
Thanks for the review; changes made as requested.

J0WI:
Thanks!

PR #7435, opened 2016-10-25
In HTTP Nowhere mode, attempt HTTPS before block.

### LINKED ISSUE #2115: 'Block all HTTP requests' should attempt HTTPS
Lately I've been using the 'Block all HTTP requests' feature in Chrome, and for quite a few smaller sites (e.g., personal blogs) I find myself manually changing the URL to `https://`. It works for a surprising amount of sites.

Wouldn't it make sense to attempt to switch a HTTP URL to HTTPS, even if there's no defined rule already? Since the site is already blocked at this point, I can't see any downside aside from a few extra milliseconds by trying HTTPS as a last-ditch effort.

Another benefit is it's possible that some SysAdmins will see a bunch of failed connections over 443/TCP in their firewall logs and hopefully get the message that their users want HTTPS.

My apologizes if this has already been brought up, I did glance at the issue tracker and didn't see it mentioned.

jsha:
This applies to both main frame resources and subresource loads, though in
practice only image subresources get a chance to be rewritten, due to mixed
content blocking.

Loop detection still works, so pages that redirect back to HTTP will get caught
by that and eventually get the "cancelled by extension" message.

Fixes #2115

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -241,8 +241,6 @@ function onBeforeRequest(details) {
   }
 
   var potentiallyApplicable = all_rules.potentiallyApplicableRulesets(uri.hostname);
-  // If no rulesets could apply, let's get out of here!
-  if (potentiallyApplicable.size === 0) { return {cancel: shouldCancel}; }
 
   if (redirectCounter[details.requestId] >= 8) {
     log(NOTE, "Redirect counter hit for " + canonical_url);
@@ -282,6 +280,15 @@ function onBeforeRequest(details) {
   }
 
   if (httpNowhereOn) {
+    // If loading a main frame, try the HTTPS version as an alternative to
+    // failing.
+    if (shouldCancel) {
+      if (!newuristr) {
+        return {redirectUrl: canonical_url.replace(/^http:/, "https:")};
+      } else {
+        return {redirectUrl: newuristr.replace(/^http:/, "https:")};
+      }
+    }
     if (newuristr && newuristr.substring(0, 5) === "http:") {
       // Abort early if we're about to redirect to HTTP in HTTP Nowhere mode
       return {cancel: true};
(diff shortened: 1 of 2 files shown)


Hainish:
Reviewed WebExtensions code, lgtm.  Added XPCOM code, @jsha can you review?

Hainish:
@cooperq do you have time to review this with me on Monday?

chromium/background.js
```diff
@@ -282,6 +280,15 @@ function onBeforeRequest(details) {
...
+    // If loading a main frame, try the HTTPS version as an alternative to
+    // failing.
+    if (shouldCancel) {
+      if (!newuristr) {
+        return {redirectUrl: canonical_url.replace(/^http:/, "https:")};
+      } else {
+        return {redirectUrl: newuristr.replace(/^http:/, "https:")};
+      }
+    }
     if (newuristr && newuristr.substring(0, 5) === "http:") {
```
cooperq: This if clause is redundant and should probably be removed.

jsha: It's actually not: the previous clause (`if (shouldCancel)`) covers the case where the request started with an HTTP request. This clause (`if (newuristr && newuristr.substring(0, 5) === "http:")`) covers the case where we started on an HTTPS page and got redirected to HTTP (by 301 or by a downgrade rule).

cooperq: Makes sense. I rescind my comments then.

gloomy-ghost:
Hey guys, @ivysrono just talked to me about [UpgradeMixedContent](https://addons.mozilla.org/zh-CN/android/addon/upgrademixedcontent/). It's an extension that inserts `upgrade-insecure-requests` to every https pages so browser will rewrite http requests before start mixed content blocking.

The flaw is inserting such a header let browser also rewrite display contents while it's not blocked by default. If the display contents aren't available for HTTPS, it fails. It's quite aggressive to do so for normal users who think http images don't really hurt, but it might be a good way for http-nowhere mode. Any thoughts?

jsha:
Could you open a fresh issue for this?

Hainish:
@gloomy-ghost tag me when you open the issue please

PR #7445, opened 2016-10-25
Add cpdb.co ruleset

VoidPhantom:


gloomy-ghost:
Please add targets:
m.cpdb.co
staging.cpdb.co

Note as `Invalid certificate` in a top comment:
ms.cpdb.co

jeremyn:
@VoidPhantom Are you still interested in working on this?

VoidPhantom:
Sorry for the long absence. :frowning_face:

Turns out, there are a lot more subdomains here than I originally thought, and an entirely new second-level domain (cpdp.co). While I added most of them to the ruleset, there are a few weird ones that I’m not sure what to do with: for www.cpdp.co, m.cpdp.co, and explore.cpdp.co, plain HTTP requests redirect to https://beta.cpdp.co, whereas HTTPS requests redirect to https://cpdb.co.

jeremyn:
I agree that the behavior for some of these subdomains is a little unexpected. I think you should make two rulesets, like this:

- [ ] `CPDB.co.xml`
```xml
<!--
	See also CPDP.co.xml

	Different HTTP/HTTPS content:
		explore.cpdb.co
		ms.cpdb.co

-->
<ruleset name="CPDB.co">
	<target host="cpdb.co" />
	<target host="www.cpdb.co" />
	<target host="beta.cpdb.co" />
	<target host="m.cpdb.co" />
	<target host="staging.cpdb.co" />

	<rule from="^http:"
		to="https:" />
</ruleset>
```

- [ ] `CPDP.co.xml`
```xml
<!--
	See also CPDB.co.xml

	Different HTTP/HTTPS content:
		www.cpdp.co
		explore.cpdp.co
		m.cpdp.co

-->
<ruleset name="CPDP.co">
	<target host="cpdp.co" />
	<target host="beta.cpdp.co" />
	<target host="mb.cpdp.co" />

	<rule from="^http:"
		to="https:" />
</ruleset>
```

What do you think?

VoidPhantom:
So, if I understand correctly, you’re suggesting that I:
1. Not include rules for subdomains that redirect to different destinations over HTTP and HTTPS. I’m okay with that.
2. Get rid of rules for subdomains which, when requested over HTTP, redirect to HTTPS on some other subdomain, but when requested over HTTPS, either refuse connection or serve an invalid certificate. Why is that, exactly? What is the problem with such rules?
3. Now that I’ve done 2, split the ruleset into two, one for each second-level domain. While I don’t mind that, what is the general rule for when it’s okay to put different second-level domains in the same ruleset? (I previously added a ruleset with multiple such domains in #4308, and that seemed to be fine…)

jeremyn:
Complicated rewrite rules like

```xml
<rule from="^http://explore\.cpdb\.co/" to="https://beta.cpdp.co/" />
```

are hard to test. It might be that the root URL of `example.com` redirects to `example.org` but a sub URL like `example.com/a/b/c.jpg` doesn't.

Also these rewrite `rule`s aren't really critical to the site. They look more like mistakes/bad configuration that the site owner might fix.

There are some concerns with cross-domain rewrites in particular, such as `example.com` to `example.org` instead of `a.example.com` to `b.example.com`. There's a current discussion in https://github.com/EFForg/https-everywhere/pull/13865 about that.

One-ruleset-per-domain is tidier. It allows for better ruleset names (see https://github.com/EFForg/https-everywhere/pull/12829#issuecomment-338031036, "Citizens Police Data Project" is the sort of name I want to avoid). Also some of the concerns in the link to https://github.com/EFForg/https-everywhere/pull/13865 in the previous paragraph apply here. I would ask you to modify your ruleset from https://github.com/EFForg/https-everywhere/pull/4308 if I were reviewing it today. (These things evolve -- I myself wrote https://github.com/EFForg/https-everywhere/blob/66e972e622210bc76e97fee9fb5b71e744a24c4f/src/chrome/content/rules/OneBookShelf.xml which I would reject today if I were reviewing it.)

VoidPhantom:
OK, done.

jeremyn:
This is a new checklist that replaces the previous two, based on what you have as of https://github.com/EFForg/https-everywhere/pull/7445/commits/20d170880ba505d0944050ff778874da47bb2110. Also my previous checklist had some outright errors, for example I'm not sure why I marked https://explore.cpdb.co as `Different HTTP/HTTPS content` instead of `Invalid certificate`, and I'm sorry if that was confusing. (However `explore.cpdb.co` should actually be removed, see below.)

These root domains have wildcard DNS entries, which you can find using `dig *.cpdb.co` and `dig *.cpdp.co`. This means any domain will match, for example https://example.cpdp.co which I just made up. However we don't want to include any random domain in the list unless it is something users are likely to find. The "remove entirely" items below are items that seem to be covered by the wildcard DNS entry and which don't exist in Google searches. (See [here](https://github.com/EFForg/https-everywhere/blob/45cf7c85151609862175b09c8ee682d204846465/CONTRIBUTING.md#enumerating-subdomains) for Google searching instructions to find subdomains.)

`CPDB.co.xml`: 

- [ ] Remove https://www.m.cpdb.co entirely

`CPDP.co.xml`:

Remove entirely:
- [ ] https://www.explore.cpdp.co
- [ ] https://www.m.cpdp.co
- [ ] https://explore.cpdp.co
- [ ] https://m.cpdp.co

Add `target`:
- [ ] https://mb.cpdp.co

VoidPhantom:
Wow, I didn’t know there was such a thing as wildcard DNS entries. I’ve done the modifications you requested.

A few things:
1. `cpdp.co` seems to not resolve now. Shall I remove the corresponding rule?
2. `www.cpdp.co` seems to redirect to `https://beta.cpdb.co/` both over plain HTTP and HTTPS now. Shall I add a rule for `www.cpdp.co`?

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7445/commits/1c7be8ed6831674b82987e94a906c9ff58010505. I've added two new items based on your previous comment. Also, just to be clear, `www.cpdp.co` should be added as a `target` above `beta.cpdp.co` (following our preferred sorting order), and it does not need a special `rule`.

VoidPhantom:
Right, when I said “rule”, I meant “target”; sorry about that. I’ll update the ruleset soon.

jeremyn:
Okay, this looks ready to merge. Are you done with your updates?

VoidPhantom:
I am; you can merge.

jeremyn:
Thanks, merged.

PR #7448, opened 2016-10-26
Delete qiyi.com.xml

ivysrono:


src/chrome/content/rules/iQiyi.com.xml
```diff
@@ -59,9 +58,8 @@
 	ˢ Secured by us, see https://www.paulirish.com/2010/the-protocol-relative-url/
 -->
 
-<ruleset name="iQiyi.com (partial)" platform="mixedcontent">
+<ruleset name="iQiyi.com (partial)">
 
-	<target host="passport.iqiyi.com" />
```
J0WI: What is the reason for removing this?

ivysrono: MCB
There is comments already.

PR #7470, opened 2016-10-30
Create ChinaW3C.org.xml

ivysrono:


gloomy-ghost:
The issue should be `incomplete certificate chain`

P.S. This site has lots of vulnerabilities... It's maintained by BUAA, maybe send a report to CNCERT?

jeremyn:
I don't think we should add new rulesets that are `default_off` because of an incomplete certificate chain. Can we close this?

PR #7471, opened 2016-10-30
Added rule for cryptopals.com

PanosSakkos:


J0WI:
@PanosSakkos the served cert does not match `www.cryptopals.com`, that's why our tests are failing.
You can add a redirect from `www.cryptopals.com` to `cryptopals.com` with a comment about the mismatch.

PanosSakkos:
@J0WI good point!

J0WI:
Could you add a comment about the mismatch?

J0WI:
closed by #9408

PR #7475, opened 2016-10-31
Added rule for bbk.bund.de.

Jochen-A-Fuerbacher:


J0WI:
We already have some rules containing `bund.de`. Maybe you can document the relationship between them or merge the files?

Jochen-A-Fuerbacher:
Merging those files wouldn't be a good idea, in my opinion. Most of them handle some websites of federal offices of Germany, which doesn't have anything in common.

I could put a comment to the top of this (bbk.bund.de) file, explaining that there are other files for the bund.de domain. Would that be ok?

J0WI:
Yes, this would be helpful. We already have such references in big rulesets like Amazon.xml or Google.xml.

src/chrome/content/rules/bbk.bund.de.xml
```diff
@@ -0,0 +1,15 @@
...
+-->
+<ruleset name="Bundesamt für Bevölkerungsschutz und Katastrophenhilfe">
+	<!-- Federal Office of Civil protection and Disaster Assistance. -->
+
+	<target host="bbk.bund.de" />
+	<target host="www.bbk.bund.de" />
+	<target host="austausch.bbk.bund.de" />
+	<target host="warnung.bund.de" />
+	
+	<rule from="^http://bbk\.bund\.de/" to="https://www.bbk.bund.de/"/>
```
J0WI: Please document the mismatch.

src/chrome/content/rules/bbk.bund.de.xml
```diff
@@ -1,4 +1,11 @@
 <!--
+	For other bund.de coverages, see:
+ 	- Federal-Office-for-Information-Security.xml
+	- Bundesnachrichtendienst.xml
+	- Bundesministerium_des_Innern.xml
+	- Federal_Office_of_Administration.xml
+	- Verwaltung_Online.xml
```
J0WI: We use to make a reference from the referenced files back to this one. See examples in `Google.xml` or `Amazon.xml`.

J0WI:
In such cases we use to define one 'main ruleset' where all references are listed and then just refer to the list in that main rule. Otherwise we have to maintain the list of references in all related rulesets. See examples in `Google.xml` or `Amazon.xml` (and the referred rules there).
I don't know if there is one ruleset of federal offices that could be defined as 'main rule'?

Some of the related rules do not fit our current requirements and break our tests. Could you just fix the errors there? Maybe they need to be rewritten, but this is likely a separate PR.

Jochen-A-Fuerbacher:
The `Verwaltung_Online.xml` ruleset contains the main `bund.de` domain. So I put the list there.

I will look at the rules that break the tests and open an other PR for that.

Jochen-A-Fuerbacher:
Fixed rules to pass tests in PR #8625.

Jochen-A-Fuerbacher:
Is there anything still to do here?

J0WI:
I would like to have all errors fixed before merging. #8626 needs to be reviewed by somebody else. There are still some errors in the fetch test, but they can be fixed here, when this branch gets rebased.

J0WI:
@Jochen-A-Fuerbacher all dependencies are merged. Could you please rebase your changes and make sure, that all tests are passing?

Jochen-A-Fuerbacher:
There are now two subdomains in `Federal_Office_of_Administration.xml` that break tests. See #8853.

PR #7483, opened 2016-11-01
Add ipip.net.xml

gloomy-ghost:


jeremyn:
Please:

Remove:
- [ ] https://scan.security.ipip.net , because it's refused for unencrypted HTTP too.
- [ ] https://api.ipip.net , because it 404s for unencrypted HTTP too.
- [ ] https://api2.ipip.net , same reason as `api`.

jeremyn:
Thanks, merged.

gloomy-ghost:
`LOVEAPP_SESSID` under `.ipip.net` doesn't seem to be secured, even the ruleset uses `.+`/`.+`

jeremyn:
Sometimes a cookie that seems should be covered under `securecookie` isn't made secure because of internal checking, see [here](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/code/HTTPSRules.js#L730-L742). I don't know if that's the case for the particular cookie you mentioned, but I don't consider that an automatic problem or bug.

gloomy-ghost:
Well, it tests securecookie applicability by checking `http://ipip.net/$random` in the ruleset, but `ipip.net` is not even in DNS...

jeremyn:
It sounds like a bug for cases when the root domain like `ipip.net` doesn't exist but subdomains set cookies for it.

PR #7281, opened 2016-10-06
Update Akamai.xml

gloomy-ghost:
Add a248.e.akamai.net
  Close #6007

PR #7484, opened 2016-11-01
Create qiniu.com.xml

gloomy-ghost:


ivysrono:
https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Qbox.me.xml

ivysrono:
kb/support Unable to connect in Tor.

gloomy-ghost:
It's just general time out

jeremyn:
I disagree with the wildcards for `*.qbox.me` and `*.qnssl.com` for the usual reasons.

gloomy-ghost:
It's a https CDN domain, likes a248.e.akamai.net

#7281 is still pending though.

jeremyn:
Why does that matter for it being a wildcard?

gloomy-ghost:
There are about half a million subdomains according to the number of Qiniu's customers. They all work for file distribution, and https://dn-xxx.qbox.com isn't a page that a regular user will be visiting.

PR #5242, opened 2016-06-21
ulteo.com

galeksandrp:

PR #7492, opened 2016-11-02
Add Ulteo.com.xml

jeremyn:
This is a new version of pull request #5242 , submitted due to lack of recent progress on the older pull request.

### DIFF
--- a/src/chrome/content/rules/Ulteo.com.xml
+++ b/src/chrome/content/rules/Ulteo.com.xml
@@ -3,15 +3,12 @@
 		blog.ulteo.com
 
 	Refused:
-		spoon.ulteo.com
-
-	Time out:
 		ulteo.com
 		archive.ulteo.com
-		demo.ulteo.com
 		doc.ulteo.com
 		i18n.ulteo.com
 		ladle.ulteo.com
+		spoon.ulteo.com
 
 -->
 <ruleset name="Ulteo.com">
--- a/src/chrome/content/rules/Ulteo.com.xml
+++ b/src/chrome/content/rules/Ulteo.com.xml
@@ -1,0 +1,26 @@
+<!--
+	Invalid certificate:
+		blog.ulteo.com
+
+	Refused:
+		spoon.ulteo.com
+
+	Time out:
+		ulteo.com
+		archive.ulteo.com
+		demo.ulteo.com
+		doc.ulteo.com
+		i18n.ulteo.com
+		ladle.ulteo.com
+
+-->
+<ruleset name="Ulteo.com">
+	<target host="www.ulteo.com" />
+	<target host="code.ulteo.com" />
+	<target host="docs.ulteo.com" />
+	<target host="project.ulteo.com" />
+	<target host="servicedesk.ulteo.com" />
+	<target host="support.ulteo.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


jeremyn:
The tests are getting errors like

> ERROR src/chrome/content/rules/Ulteo.com.xml: Fetch error: http://www.ulteo.com/ => https://www.ulteo.com/: (35, 'Unknown SSL protocol error in connection to www.ulteo.com:443 ')

According to Firefox 49.0.2, https://www.ulteo.com is encrypted with `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, 128 bit keys, TLS 1.2` . According to https://wiki.openssl.org/index.php/Manual:Ciphers(1) , `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` has the OpenSSL cipher name `ECDHE-RSA-AES128-GCM-SHA256`, which we already include in the test cipherlist [here](https://github.com/EFForg/https-everywhere/blob/ec05004c0ed2fe3306913f638a59f3ae44116b1b/test/rules/src/https_everywhere_checker/http_client.py#L52). So I don't know why this cipher isn't being recognized.

@fuglede @Hainish @J0WI Do you have any ideas?

J0WI:
No, I didn't found any issue with that host.

Hainish:
It looks like the problem is that the ciphers 'ECDHE-RSA-RC4-SHA' and 'RC4-SHA' are preferred before newer suites in our list.  These are sent along in the Client Hello TLS packet, and this server then selects one of them rather than any of the newer ciphers.  Once either of these ciphers are chosen, there is an error - the server is unable to complete the transaction.  So the server chooses ciphers that it doesn't actually support.

RC4 is deprecated, we should probably remove any RC4 ciphers from our list altogether.  The best thing to do would be to look at the Client Hello packet from each of the browsers we support, and choose the intersection of all ciphers they support.

Hainish:
The ciphers have been updated to that of Firefox 51 in https://github.com/EFForg/https-everywhere/commit/169999c1c98a8fea83d74948b4ed4f845340b9ef

Rerunning tests now.

jeremyn:
Let me recheck the domains to make sure this is up to date before we merge it.

jeremyn:
@Hainish I understand why the tests failed when I bounced it this time, but it looks pretty bad. All I changed was the top comment.

jeremyn:
@Hainish: another test is failing too, from #6697: https://travis-ci.org/EFForg/https-everywhere/jobs/204393004

Hainish:
@jeremyn looks like there's a TLS failure when trying to connect to the dockerhub server.  I'm rerunning the test now

Hainish:
This looks to be a problem with particularly the Travis build environment fetching dockerhub images - https://github.com/efforg/certbot is having the same problem, apparently.

Probably the best thing we can do for now is wait for them to fix it.  Since it's affecting anyone who uses docker within Travis, it will probably get fixed quickly.

J0WI:
Maybe we could only use docker as a fallback (if a AAAA record exist or the cipherlist is to new)?
This would avoid this kind of problems and speed up the test for most cases.

Hainish:
We have to rely on the travis build environment in either case.  Having a reliable way to reproduce tests a-la docker is far preferable to moving more of where the tests happen into the default travis build environment, with its idiosyncrasies and quirks.

jeremyn:
My Vagrant dev environment at https://github.com/jeremyn/https-everywhere-dev-environment is working out pretty well for me. I can make a more streamlined version of it if we want to use it in our automated testing somehow, just let me know.

PR #7498, opened 2016-11-02
[ACLU.org] Expand coverage, expand list of other ACLU rulesets

galenide:


jeremyn:
Please:

Add `target`:
- [ ] https://bullhorn.aclu.org
- [ ] https://mail.aclu.org
- [ ] https://survey.aclu.org
- [ ] https://login-msol-dc.vtv.aclu.org
- [ ] https://login-msol-ny1.vtv.aclu.org

Also:
- [ ] Add references to `ACLU-of-Georgia.xml`, `ACLU-of-California.xml`, and `ACLU-of-San-Diego.xml` in the top, for your recent pull requests.
- [ ] Delete [these comments](https://github.com/galenide/https-everywhere/blob/67b000a97783f9c3670af037d715cd095651ff94/src/chrome/content/rules/ACLU.xml#L22-L41), because I don't think they really help anything.
- [ ] Add `.aclu.org` to the subdomains [here](https://github.com/galenide/https-everywhere/blob/67b000a97783f9c3670af037d715cd095651ff94/src/chrome/content/rules/ACLU.xml#L44-L46), because that makes it easier to copy-and-paste for manual testing.
- [ ] You can just say `redirects to HTTP` [here](https://github.com/galenide/https-everywhere/blob/67b000a97783f9c3670af037d715cd095651ff94/src/chrome/content/rules/ACLU.xml#L45), unless there is some content breakage that isn't caused by the redirect.

jeremyn:
@galenide I think it's great you're submitting a bunch of new pull requests, however can you please try to take care of the comments on some of the pull requests you've already created? I personally don't like to build up a large backlog of open pull requests with any one contributor.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/7498/commits/f35757fa1f2aabb8529c521ff444414571be1caf .

galenide:
The purpose of including `.aclu.org` in the list of problematic subdomains seems unclear. In particular, it appears that `^aclu.org` supports HTTPS in that https://aclu.org/ redirects to https://www.aclu.org/. If `.aclu.org` was included in the list, what would be the given reason (i.e. `(cert mismatch)` or `(time out)`)?

When `privacyquiz.aclu.org` is included in the ruleset as a target host, accessing http://privacyquiz.aclu.org/ leads to a page that is broken in that the first quiz question and its choices are not displayed and the "risk assessment" indicator is blank. With such a page, users can not proceed with the quiz.

jeremyn:
> The purpose of including .aclu.org in the list of problematic subdomains seems unclear. In particular, it appears that ^aclu.org supports HTTPS in that https://aclu.org/ redirects to https://www.aclu.org/. If .aclu.org was included in the list, what would be the given reason (i.e. (cert mismatch) or (time out))?

I meant append `.aclu.org` to the subdomain names, so they look like

```
		- metrics.aclu.org (cert mismatch)
		- privacyquiz.aclu.org (redirects to HTTP & breaks content)
		- scorecard.aclu.org (cert mismatch)
```

This makes it easier for me to copy and paste these domains into a browser for testing.

> When privacyquiz.aclu.org is included in the ruleset as a target host, accessing http://privacyquiz.aclu.org/ leads to a page that is broken in that the first quiz question and its choices are not displayed and the "risk assessment" indicator is blank. With such a page, users can not proceed with the quiz.

I see. We're looking at a broken quiz after the redirect loop. In this case let's just say `(redirects to HTTP)` and hope that if they fix the redirect and we add the subdomain as a `target`, the content will also look good. (The problem with a vague `breaks content` comment is that other people looking at this ruleset might not be sure what content is broken, even if the quiz looks fine, so they might be excessively cautious adding the target.)

jeremyn:
Thanks, merged.

PR #7501, opened 2016-11-03
Add additional Flexnow rule for ruhr-uni-bochum.de

PotcFdk:
The system uses the `fn2.flexnow.ruhr-uni-bochum.de` domain after logging in.


PotcFdk:
That's weird, it works in Firefox, where it uses the certificate `1A:45:C4:39:95:E4:51` (SHA1 Fingerprint=`A8:E4:AD:02:D3:1E:39:4A:73:A9:A0:5E:E5:FE:2B:1A:4B:B8:F9:54`)
subjectAltNames `www.rz.ruhr-uni-bochum.de` and `www.itsb.ruhr-uni-bochum.de`.

Validates like this:
certificate chain

J0WI:
rerun tests

J0WI:
Seems fixed, Thanks!

PR #7507, opened 2016-11-05
[BallotReady] New ruleset

galenide:


jeremyn:
Please:
- [ ] Delete one of the two empty lines on each side of the `securecookie` tag
- [ ] Mark `ballotready.org` in a top comment as `Time out`
- [ ] Is squash-and-merge okay here?

By the way, why aren't you comfortable with a blanket squash-and-merge approval? Maybe we can address some concern you have. Also note that as far as I can tell, I'm the only reviewer who asks ahead of time, so you should assume that the other reviewers will squash-and-merge without asking.

galenide:
As far as squash-and-merge, it sometimes happens where a pull request for a new branch is opened and multiple commits are added to the branch over time in order to resolve issues with the content of the branch (most likely a ruleset.) In such a case, it is useful to be able to add commits to the branch in the normal manner even if the branch is eventually merged into the master branch via squash and merge. I am guessing that this would likely not be an issue.

In addition, it may happen where a pull request for a new branch with improvements for a ruleset is opened and the improvements are merged into the master branch and the pull request is closed. In such a case, if issues subsequently arise with the same ruleset at a later time, it is convenient to create another new branch with the improvements for the ruleset and to open a new pull request to facilitate merging the improvements (or, possibly, adjusting the improvements before they are merged.) I am guessing that this would most likely not be an issue even if the improvements for a ruleset were merged via squash-and-merge.

galenide:
In that case, it should be okay to do squash-and-merge for this pull request and future pull requests as well.

jeremyn:
Okay, thanks. Merged.

PR #7508, opened 2016-11-05
Update Youku.com.xml

ivysrono:


jeremyn:
Please:
- [ ] For `login` I'm getting `403, HTTPS only`, not `Timeout`. Can you confirm that and then change the comment if you agree?
- [ ] Next to `- i` under `MCB`, add `example: https://i.youku.com/i/UNTI4NDU4OTgw`
- [ ] Add a `test` under https://static.youku.com :

```xml
<test url="http://static.youku.com/v/js/blank.html" />
```

jeremyn:
I don't want to start linking to Travis results as a normal thing. Also I just tried again and I'm either getting `403` for both, or `403` just for HTTPS, depending on the URL. So let's change [these two lines](https://github.com/ivysrono/https-everywhere/blob/cfb3cb9e1c5c22dc8c82c678e6a40887ea5aa03e/src/chrome/content/rules/Youku.com.xml#L46-L47) to:

```xml
	403, HTTPS only:
		- login ( example: https://login.youku.com/user/signup?user=login )
```

Are you getting a timeout in the browser?

jeremyn:
Thanks, merged.

PR #7512, opened 2016-11-06
Create akixi.com.xml

gloomy-ghost:
This ruleset is generated by a script... Hope it works.

### DIFF
--- a/src/chrome/content/rules/akixi.com.xml
+++ b/src/chrome/content/rules/akixi.com.xml
@@ -1,0 +1,45 @@
+<!--
+	Refused:
+		- www.akixi.com
+
+	Timeout:
+		- eip.akixi.com
+		- nimbus.akixi.com
+		- stratus.akixi.com
+		- svn.akixi.com
+
+	Invalid Certificate:
+		- primus.akixi.com
+-->
+<ruleset name="akixi.com">
+	<target host="alto.akixi.com" />
+	<target host="broadconnect.akixi.com" />
+	<target host="cirrus.akixi.com" />
+	<target host="cumulus.akixi.com" />
+	<target host="epik.akixi.com" />
+	<target host="gamma.akixi.com" />
+	<target host="horizon.akixi.com" />
+	<target host="howdens.akixi.com" />
+	<target host="icore.akixi.com" />
+	<target host="inclarity.akixi.com" />
+	<target host="intech.akixi.com" />
+	<target host="masergy.akixi.com" />
+	<target host="mcmtelecom.akixi.com" />
+	<target host="momentum.akixi.com" />
+	<target host="nexogy.akixi.com" />
+	<target host="nimans.akixi.com" />
+	<target host="onecloud.akixi.com" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Please:

- [ ] Mark https://akixi.com under `Refused`
- [ ] Alphabetize the error types in the top comment
- [ ] Many of the `target`s have logins, which are especially appropriate for `securecookie`s. Do you want to add a specific `securecookie` to cover all the matching subdomains? The host would be `^(alto|broadconnect|cirrus|cumulus|epik|gamma|horizon|howdens|icore|inclarity|intech|masergy|mcmtelecom|momentum|nexogy|nimans|onecloud|onecloudnet|pressone|redcentric|redgap|saicom|stage2networks|support|tvf|vibe|voicenet|vonage)\.akixi\.com$` .

gloomy-ghost:
Such a long `securecookie`...

Also I'm getting `Reset` for both `^` and `www` now, not sure what's going on.

jeremyn:
I know, the `securecookie` is silly. It would be a nice enhancement to add an attribute or special `host` value to `securecookie` that means "every `target` without any `exclusion`s".

`^` and `www` are `Refused` for me now.

Looks good, merged.

gloomy-ghost:
Does `^\w` mean "every `target` except `^`"?

`.` generally doesn't match with `\w`, but it looks like the leading dot is ignored in https-ew. 

I've seen a great number of `^\w` usages in rulesets, however the style guide only mentioned that `.+`, `.*` and `.` are functionally equivalent.

jeremyn:
In a `securecookie`, `^` probably means "must start at the beginning". According to https://www.eff.org/https-everywhere/rulesets#secure-cookies , hosts in cookies with leading dots are matched as if the dot were stripped, so the fact that `\w` excludes a dot doesn't matter. I guess that `^\w` is an old way of saying `.*` before the project standardized on `.*`.

By the way, I like to use https://regex101.com for regex experimenting.

gloomy-ghost:
Thanks. Would it be okay to replace all of them to `.+` in one pull request?

jeremyn:
You mean in all 20000+ rulesets?

gloomy-ghost:
Yeah, we can do it both side and use diff to confirm that nothing goes
wrong.

jeremyn:
I would rather not. That's too many rulesets to review manually. Why do you want to do that?

gloomy-ghost:
It's confusing to have different patterns even they are functionally equivalent.

Also the replacing is not that hard. I used python in #7892 because I'm not good at bash, maybe we can use awk and compare the difference (There were some `UnicodeDecodeError`s while python were trying to read files)

gloomy-ghost:
163.com.xml currently uses `^\w`:

image

After change to `.+`:

image

...

jeremyn:
I don't know why `.+` encrypts the `__utmz` cookie and `^\w` doesn't, but given that, we should be even more careful before doing a mass search-and-replace.

PR #7517, opened 2016-11-07
Calacademy.org make trivial rule.

jish:
Also added:

* store.calacademy.org
* ticketing.calacademy.org

### DIFF
--- a/src/chrome/content/rules/Calacademy.org.xml
+++ b/src/chrome/content/rules/Calacademy.org.xml
@@ -1,6 +1,39 @@
 <!--
 	Nonfunctional hosts in *.calacademy.org:
 
+	Incomplete certificate chain:
+
+		calacademy.org
+		www.calacademy.org
+
+	Invalid certificate:
+
+		calendar.calacademy.org
+		hotels.calacademy.org
+		libcat.calacademy.org
+		pages.calacademy.org
+		research.calacademy.org
+		sacportal.calacademy.org
+		m.stg.calacademy.org
+
+	Timeout:
+
+		archdrop.calacademy.org (t)
+		casp.calacademy.org (t)
+		chat.calacademy.org (t)
+		collections.calacademy.org (t)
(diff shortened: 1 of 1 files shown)


gloomy-ghost:
Thanks. 

Please remove targets and note as `incomplete certificate chain`:
calacademy.org
www.calacademy.org

Note as `Invalid certificate`:
calendar.calacademy.org
hotels.calacademy.org

jish:
I've added and removed the domains listed.

I'm not sure how to use the legend at the bottom of the comment, so I looked through some of the other commits and found one letter notations at the end of the line wrapped in parenthesis, so I followed that convention.

    subdomain.example.org (t)

Let me know if anything looks off.

gloomy-ghost:
There are some different styles of comments, you can have a look at #6868

I personally prefer listing subdomains in categories. It's ok if you want to list them in order and mark their status one by one, but it's wired to use both styles in one ruleset.

gloomy-ghost:
I've updated https://github.com/EFForg/https-everywhere/pull/7517#issuecomment-260094044

We can't secure `legacy` after remove `www`, I'm sorry for that.

jish:
I removed `legacy.calacademy.org` and added as a comment.

I'm sorry for all of the back and forth, the tests do not run properly on my machine. When I run `bash test.sh` a new instance of Firefox opens up and just sits there and does nothing forever. Both my Terminal and the new Firefox instance will both just hang forever.

gloomy-ghost:
I'm using Windows, so I usually leave the testing to Travis CI. But did you run `install-dev-dependencies.sh` before testing?

The ruleset is good, thanks.

jish:
> "I'm using Windows, so I usually leave the testing to Travis CI."

I'm using Mac OS, but I guess Travis CI will have to be my strategy as well.

> "did you run install-dev-dependencies.sh before testing?"

Yes, I did.

PR #7518, opened 2016-11-08
Excluded qa.sockets.stackexchange.com from Stack Exchange ruleset

### LINKED ISSUE #3884: Stack Exchange (partial) breaks Firefox Web Sockets to qa.sockets.stackexchange.com
Basically all the web-socket related features like live updates are broken on Stack Exchange websites when "Stack Exchange (partial)" is enabled, and the following appears in the Firefox console:

> Firefox can't establish a connection to the server at wss://qa.sockets.stackexchange.com/.

Disabling "Stack Exchange (partial)" makes it work again, at the expense of not having `http` rewritten to `https`. I've checked, and the web sockets do behave the same using `http` or `https` URL's with the rule disabled.

You have to be logged in to see this issue. Then just visit any question [like this one](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454) and see the error in the console.

I've also looked over [the rules](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Stack-Exchange.xml), but nothing jumps out as being the issue.

braiam:
This host should not be downgraded to http, since it's a hop
to enable websockets. Close #3884

---
This change is [Reviewable](https://reviewable.io/reviews/efforg/https-everywhere/7518)

### DIFF
--- a/src/chrome/content/rules/Stack-Exchange.xml
+++ b/src/chrome/content/rules/Stack-Exchange.xml
@@ -54,7 +54,11 @@
 		<test url="https://meta.unix.stackexchange.com/" />
 		<test url="https://meta.opendata.stackexchange.com/" />
 		<test url="https://blog.gaming.stackexchange.com/" />
-		
+
+		<!-- but this one should not -->
+		<exclusion pattern="^https://qa\.sockets\.stackexchange\.com/" />
+		<test url="https://qa.sockets.stackexchange.com/" />
+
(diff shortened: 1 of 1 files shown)


AlexanderOMara:
Is there any hope for this getting merged? This is a pretty annoying bug for those that use this site and extension?

J0WI:
Thanks!

PR #7519, opened 2016-11-08
Create So.com.xml

gloomy-ghost:


jeremyn:
Mark as `403`:
- [ ] https://p.ssl.so.com

- [ ] https://maps.so.com

Mark as `Time out`:
- [ ] https://images.so.com
- [ ] https://j.news.so.com
- [ ] https://qcaptcha.so.com
- [ ] https://zhishu.so.com

Add `target`:
- [ ] https://m.image.so.com

Also:
- [ ] Just mark https://t.so.com as `Mixed content`, with no extra comments
- [ ] Change https://api.app.m.so.com from `403` to `Mixed content`. The root URL gets a 403 with both (unencrypted) HTTP and HTTPS. The specific pages I checked from Google have different mixed content blocking problems.
- [ ] For https://tran.news.so.com I get a 403 for both (unencrypted) HTTP and HTTPS. I can't find anything other than the root URL so I don't know if this is the same problem as https://api.app.m.so.com . I think `tran.news.so.com` should be a `target`.

gloomy-ghost:
dev.yunpan.360.cn redirects to https itself, 404 in https though.

jeremyn:
@gloomy-ghost Can you please take a look at this one?

gloomy-ghost:
http://dev.yunpan.360.cn/ (200)
```html
<html>  
<meta http-equiv="refresh" content="0;url=https://dev.yunpan.360.cn/">  
</html> 
```

https://dev.yunpan.360.cn/ (404)

jeremyn:
Are you saying the problem with the `test` should be ignored because the plain HTTP version of the page already does a meta-refresh to the broken HTTPS version of the page?

gloomy-ghost:
There are still some results in search engines, so maybe it just became broken some days ago. I can't tell if it's temporarily or permanently. What would you prefer?

jeremyn:
I'm not sure what you are asking me.

gloomy-ghost:
If it's a temporary error and will be fixed soon, there is no point to comment it and re-enable it days after.

If it's abandoned, I'm not sure do we even need to keep it in comments.

J0WI:
@gloomy-ghost what is the state of this PR?

src/chrome/content/rules/360.cn.xml
```diff
@@ -92,7 +87,11 @@
 	<target host="shouji.360.cn" />
 	<target host="t.360.cn" />
 	<target host="t321.tianji.360.cn" />
+	<target host="wangzhan.360.cn" />
 	<target host="yunpan.360.cn" />
+	<target host="*.yunpan.360.cn" />
+		<test url="http://c2.yunpan.360.cn/" />
+		<test url="http://c12.yunpan.360.cn/" />
```
J0WI: Can you provide one more test url here?

gloomy-ghost: Any numbers from 1 to 13 IIRC. They are shutting down this service shortly, so if there are any issues regarding the yunpan.360.cn subdomain I might just remove them altogether.

J0WI: It's just because we require 3 in our style guide

jeremyn:
It doesn't really matter now, but I've updated the now-completed checklist in https://github.com/EFForg/https-everywhere/pull/7519#issuecomment-260222637 including modifying some of the requested changes.

@J0WI Thanks for taking this one over.

PR #7527, opened 2016-11-10
Create Thewarehouse.co.nz.xml

gloomy-ghost:


jeremyn:
Add `target`:
- [ ] https://thewarehouse.co.nz

Mark as `No working URL known` unless you have a working URL, see https://github.com/EFForg/https-everywhere/issues/7662 :
- [ ] https://image.emails.thewarehouse.co.nz (and remove target)

Also:
- [ ] Switch the order of `Timeout` and `Invalid Certificate` so they are alphabetical

gloomy-ghost:
Would `/robots.txt` be a "working" URL?

jeremyn:
Yes, good catch. Can you add it as a `test` under the `rule`?

jeremyn:
Thanks. I've also added an item to the checklist about `securecookie`, I'm sorry I didn't mention that before.

gloomy-ghost:
Actually, I'm still confused about the difference of `.+`/`.+` and a long list that covers every subdomain in `host`...

jeremyn:
The problem is that `a.example.com` can set a cookie for `.example.com` that `b.example.com` might want to read. If `a` works with HTTPS and `b` doesn't, and a `.+/.+` `securecookie` forces `a` to make an `.example.com` cookie secure, then `b` won't be able to read it.

A specific example here is that both https://www.thewarehouse.co.nz and http://curtains.thewarehouse.co.nz can interact with a cookie named `__cq_uuid` for host `.thewarehouse.co.nz`, with the cookie value being a specific `$UUID`. `www` is a `target`, so it becomes https, but `curtains` gets a `Timeout` with https, so it has to be plain http. When you visit `www`, the value `$UUID` is set. When you visit `curtains`, your browser won't send this cookie to the server, because it's not allowed to send it via plain http because of the secure flag. However the server might expect to be able to know what `$UUID` is, but now it doesn't, so something might go wrong. 

Imagine that `__cq_uuid` is a login cookie and `$UUID` represents a user's identity. In this case, the user might experience the above problem as: they log in on `www`, but then seem to be logged out when they visit `curtains`.

We can solve this problem by restricting `securecookie` to only force `www.thewarehouse.co.nz` cookies to be secure, without doing anything with `.thewarehouse.co.nz` cookies.

Does this help?

jeremyn:
Thanks for making the changes, merged.

gloomy-ghost:
The thing is `curtains.` is kind of separated from `^`. They don't use same account, not even share shopping cart. Secure cookies under `^` may affect their statistics report though.

IMO, the login cookie is the most sensitive cookie that we are trying to secure, and it is a big step of security if we are able to secure it.

jeremyn:
To be clear, I made up the login behavior just to have an example. According to https://cookiepedia.co.uk/cookies/__cq_uuid , that's actually a tracking cookie.

This analysis

> The thing is `curtains.` is kind of separated from `^`. They don't use same account, not even share shopping cart.

requires that we pay attention to the relationship between every pair of HTTPS and non-HTTPS pages for a domain. That is not realistic, so I err on the side of not breaking things and discourage `.+/.+`.

Login cookies are important to protect but they are also usually important to the site's functionality, so we can't just break them. Also it is very possible that the cookie is itself encrypted and therefore safe to transmit over plain HTTP, in which case the user isn't gaining much security in exchange for breaking the site's functionality.

PR #7545, opened 2016-11-12
Update easyicon.net.xml

gloomy-ghost:
MCB Fixed

### DIFF
--- a/src/chrome/content/rules/easyicon.net.xml
+++ b/src/chrome/content/rules/easyicon.net.xml
@@ -7,17 +7,11 @@
 		ann
 		download
 -->
-
 <ruleset name="easyicon.net">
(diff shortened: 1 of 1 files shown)


jeremyn:
- [ ] https://www.easyicon.net/vip/ has mixed content blocking for jQuery. I don't know what actual functionality is broken but I don't like breaking jQuery. I think we should keep that `exclusion`. What do you think?

gloomy-ghost:
Well... They fixed https://www.easyicon.net/thank.php which was in my feedback, but left all others behind.

I'll update this thread after these pages are fixed.

gloomy-ghost:
It should be fixed by now, can you check again?

jeremyn:
Looks good, merged.

PR #7547, opened 2016-11-13
Update WordPress.xml

gloomy-ghost:
It looks a bit crazy, but they are using wildcard DNS records and certificates, and I don't see any problems with straightly rewrite.


jeremyn:
There could be mixed content issues on particular blogs even with wildcard certificates.

gloomy-ghost:
HSTS Header appears on every subdomains, we are just preventing SSL
Stripping during the first request.

jeremyn:
> HSTS Header appears on every subdomains

How do you know?

gloomy-ghost:
```python
>>> 'Strict-Transport-Security: max-age=15552000' in str(
    urllib.request.urlopen(
        'http://{0}.wordpress.com/'.format(
            str(random.randint(0,9223372036854775807))
        )
    ).info()
)
True
```

jeremyn:
Do you mean you're looking at whether the STS header is being returned from URLs like http://691242834217739860.wordpress.com ?

gloomy-ghost:
It doesn't have to be an integer, but they do send HSTS whether the subdomain is actually existing or not.

J0WI:
Wordpress announced HTTPS support long time ago and they do it very well so far.
So I would say go for it!

J0WI:
@gloomy-ghost could you add at least 3 vaild test urls for each wildcard?

J0WI:
@gloomy-ghost would be great to have this in the next release

src/chrome/content/rules/WordPress.xml
```diff
@@ -6,15 +6,27 @@
 <ruleset name="WordPress">
 	<target host="wordpress.com" />
 	<target host="*.wordpress.com" />
+		<test url="http://www.wordpress.com/" />
+		<test url="http://en.wordpress.com/" />
+		<test url="http://zh-cn.wordpress.com/" />
 
 	<target host="wordpress.org" />
 	<target host="*.wordpress.org" />
+		<test url="http://apps.wordpress.com/" />
```
J0WI: This needs to be a .org host

gloomy-ghost: oops, sorry

PR #7551, opened 2016-11-13
Update Ubuntu-CN.xml

ivysrono:


jeremyn:
Please:

- [ ] [These lines](https://github.com/ivysrono/https-everywhere/blob/1d07df8f60a807e1fe4b46d8e5dab26cf83db8ca/src/chrome/content/rules/Ubuntu-CN.xml#L32-L35) are okay, but I think it's a little clearer to rewrite them like this:

```xml
	<exclusion pattern="^http://linux\.ubuntu\.(com|org)\.cn/$" />
		<!-- Not excluded -->
		<test url="http://linux.ubuntu.com.cn/news/392498.html" />
		<test url="http://linux.ubuntu.org.cn/news/392498.html" />
```

What do you think? If you agree, please change it.

jeremyn:
Thanks, merged.

PR #7565, opened 2016-11-15
[Community Savings Credit Union] New ruleset

lucaspetter:


gloomy-ghost:
Thanks, Please:

- [ ] Add target `webmail.comsavings.com`

lucaspetter:
Okay, new target added.

gloomy-ghost:
Thanks, merged.

Separated commits make review easier because I can use the checklist with diff. This simple one is ok but I would have to re-review if there was a big ruleset, so please just leave the previous commits when a reviewer has begun reviewing.

PR #7576, opened 2016-11-15
[North Peace Savings Credit Union] New ruleset

lucaspetter:


lucaspetter:
`eft.npscu.ca` times out over http but not over https, causing the Travis test to fail. How should this be handled?

J0WI:
For me it also times out on https.

lucaspetter:
Thanks. I'm seeing it timeout on https now too. I've removed `eft.npscu.ca`.

PR #7578, opened 2016-11-15
[Tangerine Bank] New ruleset

lucaspetter:


src/chrome/content/rules/Tangerine-Bank.xml
```diff
@@ -0,0 +1,33 @@
...
+	<target host="m.tangerine.ca" />
+	<target host="ofx.tangerine.ca" />
+	<target host="tangerine.com" />
+	<target host="www.tangerine.com" />
+
+	<test url="https://tangerine.com/" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^https?://(www\.)?tangerine\.com/" to="https://$1tangerine.ca/" />
```
J0WI: We don't touch the https version

lucaspetter: I initally added that rule because `tangerine.com` times out over https. But it's probably not needed, so I'm changing it to an http-to-https redirect.

J0WI:
Thanks!

lucaspetter:
Thanks!

PR #7580, opened 2016-11-15
[Vancity] New ruleset

lucaspetter:


src/chrome/content/rules/Vancity.xml
```diff
@@ -0,0 +1,38 @@
...
+	<target host="myreloadable.ca" />
+	<target host="www.myreloadable.ca" />
+
+	<test url="https://vancityinsurance.com/" />
+	<test url="https://myreloadable.ca/" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^https?://vancityinsurance\.com/" to="https://www.vancityinsurance.com/" />
+	<rule from="^https?://myreloadable\.ca/" to="https://www.myreloadable.ca/" />
```
J0WI: We don't touch https connections

lucaspetter: I initially added those rules because `vancityinsurance.com` and `myreloadable.ca` both have valid certs only for `www.`. Since we can't do an https-to-https redirect, I'm removing the rules and targets.

J0WI: You could also just change `https?` to `http`, but I'm also fine with this.

lucaspetter: Thanks. That's a better way to do it, I'll open a new PR to fix that.

PR #6921, opened 2016-09-20
Update QQ.xml QQ.com-mixedcontent.xml

gloomy-ghost:
Remove commented target in QQ.xml that also exists in mixedcontent
Enable connect, imgcache, jsqmt in QQ.xml
Move kf, pay to mixedcontent
service has been suspended

PR #7586, opened 2016-11-15
Update Qlogo.cn.xml

gloomy-ghost:


jeremyn:
I'm getting `HTTP 400` for the URLs I checked in `Qlogo.cn.xml`. I'm not sure what to do about that.

I'm also not sure what to do about `Tencent_CN.xml` if it requires me to have an exit point in the GFW to test.

gloomy-ghost:
Do you mean all URLs? They are CDN subdomains and most of them have blank roots.

Could you try to use Tor and set `CN` as exit node? I guess `CN` nodes are mostly honeypots but they should be okay for testing as long as they don't drop 443.

gloomy-ghost:
I've added testing URLs for them.

Also would it be okay if @ivysrono could provide a screen recording to prove that these subdomains are available in https in RPC?

ivysrono:
dldir1
dp3
mmbiz
wx

jeremyn:
@gloomy-ghost I'm having trouble with Tor. I added `ExitNodes {CN}` to my torrc, but it seems to hang halfway through `Connecting to the Tor network`.

Also, I'm not going to sign off on a pull request based on a screenshot, meaning no disrespect to @ivysrono .

@Hainish : do you at the EFF have any tools available to test something from within the Chinese GFW?

gloomy-ghost:
There is a public proxy in China: 1.82.216.134:80

jeremyn:
I'm not comfortable using a random public proxy. I'm sorry if it seems like I'm being difficult.

gloomy-ghost:
It's all right, but I'm not sure what are you worrying about public proxies
since we are just testing https availability.

jeremyn:
It's not a good precedent. An attacker could add a malicious ruleset by saying, "Here's my ruleset, and by the way, you need to test it through this specific proxy", where the attacker controls the proxy and manipulates the response. We should be able to check rulesets using our own independent tools and settings.

gloomy-ghost:
An attacker can't manipulate response of https request unless he have a
valid certificate. If he had one, it would be unnecessary to push a
malicious ruleset to here (and I don't think a ruleset that only
redirecting http to https can be malicious).

In this case, I'm not asking you to use an specific proxy. You can search
"Chinese proxy” and choose whichever is working, or find a Chinese VPS to
test the ruleset. I understand you want to have a testing environment that
can be trusted, but https is designed for untrusted network isn't it?

jeremyn:
The whole problem here is that for some reason we get different results from exit points inside the GFW than we do from outside it, so the exit point does matter. I guess whatever technique is being used by the GFW could also be used by one malicious attacker managing one exit point.

gloomy-ghost:
For subdomains in Tencent_CN, we are getting `Mismatch` or `Timeout`
outside of China.

`Mismatch` is because Tencent has own CDN network in China (which is its
main market), but not overseas. I think it's a country-based DNS result
that you get a non-https CDN when you are not in China.

`Timeout` is not easy to determine. Maybe Tencent dropped oversea
connections, or GFW dropped them. My point is attackers with MITM access
can't "manipulate" a https response. All they can do is monitor encrypted
data or drop/reset connections, which has no difference with a Cisco
enterprise router.

jeremyn:
So to properly test these domains, I have to proxy DNS, not just HTTP(S) traffic. I have to trust that both the proxy's DNS and web traffic match what a regular user inside the GFW sees, and if either DNS or web traffic differs, then I as a maintainer might be misled.

jeremyn:
For "web traffic", I understand that HTTPS traffic can't be modified, however it's possible the person who controls the proxy might be able to route traffic outside the GFW, or not, depending on some criteria.

gloomy-ghost:
To clarify my opinion, there is at least one official server supporting https when the domain works with a proxy. Network hijacking is everywhere, but if it only routes traffic, it's more likely a router.

> I have to trust that both the proxy's DNS and web traffic match what a regular user inside the GFW sees

As I said, hijacking is everywhere and even Tencent can't say their https sites have 100% availability in China. CMCC is different with GWBN, CN2 is different with CHINANET. There are always some risks that https doesn't work, and we can't test them one by one. As a project rarely has official ruleset, we have to take this kind of risks.

jeremyn:
The way to guard against having the review manipulated is to use independent tools. In this case though I have to configure my environment a certain way and use one of a limited number of access points. I don't see how I personally could feel comfortable enough with a test in this situation to sign off on it. I've unassigned myself from this pull request.

I hope that @Hainish can come up with some approach. We need to figure out a general process for handling these inside-the-GFW issues.

ivysrono:
@gloomy-ghost 我看他们暂时解决不了GFW的测试问题，干脆绕开，进一步拆分，把没问题的先合并了。

gloomy-ghost:
@ivysrono 现在好歹绑着Alexa排名高的域名，再拆就真没人处理拆出来的了…

jeremyn:
@gloomy-ghost Here's an idea: since `Tencent_CN.xml` is the part with the difficult-to-test only-in-China problem, can you move it out of this pull request, maybe back into #6921 ?

gloomy-ghost:
@jeremyn The problem is [Travis-CI](https://travis-ci.org/EFForg/https-everywhere/jobs/176201241) won't pass since some overlaps between two rulesets.

jeremyn:
You mean `t8.qlogo.cn`? Can you remove only just that URL from the `Tencent_CN.xml` ruleset but otherwise leave it alone?

gloomy-ghost:
Well, I used `cherry-pick` for reducing conflicts, but I will do it if it makes the progress quicker.

jeremyn:
Please just only delete [this line](https://github.com/EFForg/https-everywhere/pull/7586/commits/42ee2852a8fb147d6fee7d60f272c1c2faaab79f#diff-6e197fc21dd0ab26e8e3a9a7951a78bdL15) and make no other changes to `Tencent_CN.xml`, not even cosmetic changes.

jeremyn:
I'm getting `400`s for the `test`s associated with these `target`s:
- [ ] https://mmbiz.qlogo.cn
- [ ] https://shp.qlogo.cn
- [ ] `https://q[1-4]?.qlogo.cn`
- [ ] https://qzapp.qlogo.cn

gloomy-ghost:
All of them are working for me.

PS: For `q[1-4]?.`, you need to replace `&#x26;`

jeremyn:
Okay, the URLs work now, thanks. I've updated the checklist. To clarify: by "replace `&#x26;`" you mean replacement it with `&`, so the URL should be

http://q.qlogo.cn/g?b=qq&k=Osib95YbszciblkflbYwNQgw&s=140

Merged.

Hainish:
I've reached out to some colleagues particularly well positioned to know about testing within the GFW.  Hopefully they will have some tips on the best strategy here.

PR #7595, opened 2016-11-17
Move BlueShore-Financial.xml to its proper place

smw-koops:


jeremyn:
Oops, good catch! Thanks.

PR #7599, opened 2016-11-18
[ACLU of New Mexico] New ruleset

galenide:
In the ruleset, it is mentioned that `^aclu-nm.org` doesn't redirect properly. More specifically, `https://aclu-nm.org/` redirects to `https://www.aclu-nm.org`, which works. At the same time, other URLs do not redirect properly. For example, `https://aclu-nm.org/about/` redirects to `https://www.aclu-nm.orgabout` and `https://aclu-nm.org/resources/know-your-rights/` redirects to `https://www.aclu-nm.orgresources/know-your-rights/`. The redirect from `^http://aclu-nm\.org/` to `https://www.aclu-nm.org/` is intended as a workaround for this issue.

### DIFF
--- a/src/chrome/content/rules/ACLU-of-New-Mexico.xml
+++ b/src/chrome/content/rules/ACLU-of-New-Mexico.xml
@@ -4,7 +4,7 @@
 
 
 	Problematic domains:
-	- ^aclu-nm.org		(doesn't redirect properly)
+	- ^aclu-nm.org		(broken redirect, example: https://aclu-nm.org/about > https://www.aclu-nm.orgabout/)
 
 -->
 <ruleset name="ACLU of New Mexico">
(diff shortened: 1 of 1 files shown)


jeremyn:
I've emailed ACLU-NM about their site behavior. In the meantime, please add a reference from `ACLU.xml` to this ruleset.

galenide:
With regard to this ruleset being referenced in `ACLU.xml`, pull request #7498 is relevant.

jeremyn:
> pull request #7498 is relevant.

I agree, we can handle it in that pull request.

jeremyn:
I haven't heard from ACLU of New Mexico so let's just go ahead with this.

- [ ] Please clarify how `^` is broken. I recommend changing [this line](https://github.com/galenide/https-everywhere/blob/094b82434f3f57582076aad37f8b1ead6e4a8b1c/src/chrome/content/rules/ACLU-of-New-Mexico.xml#L7) to

```xml
	- ^aclu-nm.org		broken redirect, example: https://aclu-nm.org/about > https://www.aclu-nm.orgabout/
```

jeremyn:
Thanks, merged.

PR #7601, opened 2016-11-18
[ACLU of Montana] New ruleset

galenide:


jeremyn:
Please:

- [ ] Remove the `Problematic domains` comment [here](https://github.com/galenide/https-everywhere/blob/caa5adb66422cbe26b51deb6276564d7f984df17/src/chrome/content/rules/ACLU-of-Montana.xml#L5-L8). It looks like all of `*.aclumontana.org` other than `www` is recognized in DNS but fails HTTPS, so we don't need to mention specific failures.

jeremyn:
Thanks, merged.

PR #7629, opened 2016-11-18
[ACLU of Missouri] New ruleset

galenide:


jeremyn:
This looks fine, but

- [ ] Can you please change [this entry in `ACLU.xml`](https://github.com/EFForg/https-everywhere/blob/10895935f2d34c3ec258a0d9e200309a925632c6/src/chrome/content/rules/ACLU.xml#L13) from `ACLU-of-Missouri` to `ACLU-of-Missouri.xml`? It looks like we missed that the first time.

galenide:
Please see pull request #7729

jeremyn:
Thanks, merged.

By the way, it would have been fine to do the `ACLU.xml` change in this pull request. In fact it would have been a little easier for me.

PR #7638, opened 2016-11-19
Update poliisi.fi

Foorack:


jeremyn:
Please:

- [ ] Change [this](https://github.com/Foorack/https-everywhere/blob/e8f30b4ac00894e322ee737bdfa059b4369441a1/src/chrome/content/rules/Poliisi.fi.xml#L2) from `Wrong certificate` to `Invalid certificate`. I know this is a very minor complaint, it's just good to be consistent.

PR #7639, opened 2016-11-19
Added vero.fi

Foorack:
This pull request partially resolves issue #3069

### DIFF
--- a/src/chrome/content/rules/vero.fi.xml
+++ b/src/chrome/content/rules/vero.fi.xml
@@ -1,30 +1,28 @@
 <!--
 
-    Timeout:
-        - portal.vero.fi
-        - prosentti.vero.fi
-        - veronsaajat.vero.fi
-        - veronsaajat2.vero.fi
+	Timeout:
+		- portal.vero.fi
+		- prosentti.vero.fi
+		- veronsaajat.vero.fi
+		- veronsaajat2.vero.fi
 
-    Invalid certificate:
-        - arkisto.vero.fi
-        - kove.vero.fi
-        - www.kove.vero.fi
-        - palaute.vero.fi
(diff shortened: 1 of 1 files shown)


jeremyn:
Please:

- [ ] Remove `target` https://verokortti.vero.fi , because it's not in DNS.
- [ ] Remove `target` https://stat.vero.fi unless you can find a URL that doesn't return a `403`. Otherwise, move it to a new error category `No working URL known`.

Foorack:
@jeremyn 

- Removed `target` https://verokortti.vero.fi.
- Removed `target` https://stat.vero.fi and marked is as `403`.
- Sorted errors alphabetically.

jeremyn:
A problem with marking a URL as `403 Forbidden` the way you did is that it might mean that only the HTTPS version of the page gets a `403`, while the HTTP version works fine. Instead what we really mean is that it might be a valid target, we just don't know any URL that can confirm it, which is why I suggested `No working URL known`. Does that make sense? I admit the wording `No working URL known` is awkward, so you can think of something better, please suggest it.

Foorack:
@jeremyn Yes, it makes sense. Even if `No working URL is known` might be weird to say I think it perfectly describes the issue. It seems like if HTTP returns 200 and HTTPS 403 then it should be classified as 403. If both return 403 and the HTML body does not have any external links then it will be marked as `No working URL is known`.

Does test's **require** the status_code **200**?

Foorack:
I am currently working on rewriting my earlier generator program. It is now completely written in python (compared to before when much of it was bash) and can generate more intelligent rules than ever before. However, due to the many rare cases found I'm thinking of making a flow-chart and rewriting it once again. You did an extremely amazing job on the SL.se pull request. 😄

Issue #6629, opened 2016-08-27
When should rulesets allow unmarked mixed content blocking?


jeremyn:
@fuglede @Hainish @J0WI This issue is for general discussion.

My opinion is that if any mixed content blocking (MCB) happens, it needs to be marked either by disabling the target or marking the entire ruleset with `platform="mixedcontent"`, depending on how many targets are affected. By "any mixed content blocking" I mean that if Firefox lets you [unblock mixed content](https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox#w_unblock-mixed-content), then that target should be marked.

This seems harsh but some motivation is that in pull request #5217, I pushed back on MCB and the contributor @galeksandrp made the valid point that the MCB was only a font and some search branding, which seems minor. However it turned out that this MCB hid an important menu item in a common browser configuration. So even seemingly minor MCB can really break a site, and because a maintainer can't reasonably check all pages in all configurations, we should err on the side of usability.

What do you all think?

fuglede:
I'm not sure it will be possible to make a general process on that. While I never ran the numbers to back this up, my guess would be that a (very) considerable part of the ruleset database accepts active mixed content blocking, and you would be disabling a huge number of perfectly fine rulesets to take care of what I expect on the other hand to be a very limited amount of edge cases.

jeremyn:
I'm not saying we should go back and update existing rulesets. Let's assume they're fine because no one has complained about them (yet).

galeksandrp:
And contributors who want to add a single line to existing ruleset will have to rewrite whole ruleset? Great.

jeremyn:
@galeksandrp I agree it would be annoying for a contributor who wants to make a small update to an existing ruleset to have the maintainer demand they fix a lot of old problems with mixed content. In that case I think it's okay for the maintainer to ask for the other stuff but if the contributor doesn't want to do it, then the maintainer should probably just accept the small update. It depends on how broken things are.

fuglede:
@jeremyn: But if that holds for previous rulesets, then it is also likely to hold for future rulesets.

jeremyn:
@fuglede If we need to make a policy change on how we handle mixed content blocking or anything else, it shouldn't depend on whether we're willing to update the huge number of existing rulesets. Otherwise we would never make any policy changes.

fuglede:
[...]

PR #7646, opened 2016-11-19
Rewrite National-Snow-and-Ice-Data-Center.xml to meet new rules and add exclusion

### LINKED ISSUE #7367: A certain NSIDC page is broken
If I navigate to http://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/

HTTPS Everywhere redirects to

https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/

but then the chart doesn't load. Console message:

Mixed Content: The page at 'https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/' was loaded over HTTPS, but requested an insecure resource 'http://nsidc.org/data/tools/arctic-sea-ice-chart/'. This request has been blocked; the content must be served over HTTPS.

Given that the resource is from the same website, why doesn't HTTPS Everywhere simply load it from https://nsidc.org/data/tools/arctic-sea-ice-chart/ ?

smw-koops:
Closes #7367.

### DIFF
--- a/ruleset-style.md
+++ b/ruleset-style.md
@@ -21,8 +21,18 @@ which subdomains are covered.
 If you know all subdomains of a given domain support HTTPS, go ahead and use a
 left-wildcard, along with a plain rewrite from `"^http:"` to `"^https:"`. Make sure
 to add a bunch of test URLs for the more important subdomains. If you're not
-sure what subdomains might exist, you can iteratively use Google queries and enumerate
-the list of results like such:
+sure what subdomains might exist, you can install the `Sublist3r` tool:
+
+    git clone https://github.com/aboul3la/Sublist3r.git
+    cd Sublist3r
+    sudo pip install -r requirements.txt # or use virtualenv...
+
+Then you can to enumerate the list of subdomains:
+
+    python sublist3r.py -d example.com -e Baidu,Yahoo,Google,Bing,Ask,Netcraft,Virustotal,SSL
+
+Alternatively, you can iteratively use Google queries and enumerate the list of
+results like such:
 
 1. site:*.eff.org
 2. site:*.eff.org -site:www.eff.org
--- a/src/chrome/content/rules/127.net.xml
+++ b/src/chrome/content/rules/127.net.xml
@@ -1,0 +1,56 @@
+<!--
+	Cert error:
+		^
+		www.127.net
+		
+		get test url in mail.163.com:
+			cp.127.net
+			ep.127.net
+			tp.127.net
+			gzcp.127.net
+			gzep.127.net
+			gztp.127.net
+			hzp.127.net
+-->
+
+<ruleset name="127.net">
+
+	<target host="mimg.127.net" />
+		<test url="http://mimg.127.net/index/lib/scripts/base_v5.min.js" />
+	
+	<target host="imglf.nosdn.127.net" />
+		<test url="http://imglf.nosdn.127.net/img/T0VLUWtpQlNHSmVHYlM0QU01cElqRE0vM09ETzRKTEtjK3hROFpqSk9QMm1GWXlaL2EwQ1hBPT0.jpeg" />
+	
+	<target host="imglf0.nosdn.127.net" />
+		<test url="http://imglf0.nosdn.127.net/img/SS9BOUErVTdaNTNOMi9hVk50U0t2cUZ2eVQvOWZRMFdsSElYV0o2ZDIxU2lxdkNZM25qMVl3PT0.jpg" />
+	
+	<target host="imglf1.nosdn.127.net" />
+		<test url="http://imglf1.nosdn.127.net/img/Y0N3cWN5LzNBY3pPaGRhd3VTWFRqSCtSaER4MFNIT0paanBkRjQ3b251SkxUbnFoajkvN2xnPT0.jpg" />
+	
+	<target host="imglf2.nosdn.127.net" />
+		<test url="http://imglf2.nosdn.127.net/img/dHpYYnhNNG03U1dybWdVN293WWZia25SNmhRaUNVOW8.jpg" />
+	
+	<target host="lovepicture.nosdn.127.net" />
+		<test url="http://lovepicture.nosdn.127.net/2234303933346505348" />
+	
+	<target host="mail-userthumb.nosdn.127.net" />
+		<test url="http://mail-userthumb.nosdn.127.net/f67aa1965c10333f78d75e98833c54d3.jpg" />
+	
+	<target host="onegoods.nosdn.127.net" />
+		<test url="http://onegoods.nosdn.127.net/goods/2491/d0f896aac597931a5429d32d57267793.jpg" />
+	
+	<target host="open-image.nosdn.127.net" />
+		<test url="http://open-image.nosdn.127.net/b16f89fd2e31407ab4dbf249243ce550.jpg" />
+	
+	<target host="ursdoccdn.nosdn.127.net" />
+		<test url="http://ursdoccdn.nosdn.127.net/webzj_m163/message_2016072801.js" />
+	
+	<target host="haitao.nosdn1.127.net" />
+	<target host="haitao.nosdn2.127.net" />
+	<target host="haitao.nosdn3.127.net" />
+	<target host="haitao.nosdn4.127.net" />
+	<target host="haitao.nosdn5.127.net" />
+		<test url="http://haitao.nosdn5.127.net/onlineipfe50u411092.jpg" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
--- a/src/chrome/content/rules/163.com.xml
+++ b/src/chrome/content/rules/163.com.xml
@@ -109,8 +109,6 @@
 
 <ruleset name="163.com (partial)">
 
-	<!--	Direct rewrites:
-				-->
 	<target host="163.com" />
 	<target host="analytics.163.com" />
 	<!--target host="1.163.com" /-->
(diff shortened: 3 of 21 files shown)


jeremyn:
Please:

Mark as `503, HTTPS only`:
- [ ] https://noaa_waf.apps.nsidc.org
- [ ] https://soac.apps.nsidc.org

Mark as `Incomplete certificate chain`:
- [ ] https://n5eil01u.ecs.nsidc.org

- [ ] https://www.nsidc.org

Mark as `No working URL known`:
- [ ] https://inside.nsidc.org
- [ ] https://staging.inside.nsidc.org

(Note that the error categories above are in alphabetical order, for example `Incomplete certificate chain` before `Invalid certificate`. Please maintain that in the ruleset.)

Add `target`:
- [ ] https://daacdata.apps.nsidc.org
- [ ] https://smap_op.apps.nsidc.org
- [ ] https://igloo.nsidc.org
- [ ] https://support.nsidc.org

Also:
- [ ] Delete [all of this](https://github.com/smw-koops/https-everywhere/blob/c67f06ec8f557a720e07fcc351e6c96d608d6acf/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L2-L20), we don't need it.
- [ ] Change [this comment](https://github.com/smw-koops/https-everywhere/blob/c67f06ec8f557a720e07fcc351e6c96d608d6acf/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L21) to `Invalid certificate:`
- [ ] Move `www.nsidc.org` directly under `nsidc.org` in the list of `target`s, because we like to move these two to the top as special cases.
- [ ] Delete [the general `securecookie`](https://github.com/smw-koops/https-everywhere/blob/c67f06ec8f557a720e07fcc351e6c96d608d6acf/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L37) or make it more specific, because there are too many domains that don't work with HTTPS.
- [ ] If you decide to rewrite the `securecookie` to be more specific, fix its alignment with the other tags.
- [ ] Delete [this comment](https://github.com/smw-koops/https-everywhere/blob/c67f06ec8f557a720e07fcc351e6c96d608d6acf/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L39)
- [ ] Delete the space [here](https://github.com/smw-koops/https-everywhere/blob/c67f06ec8f557a720e07fcc351e6c96d608d6acf/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L38)

smw-koops:
@jeremyn Done.

jeremyn:
Thanks. Some comments:

* `www.nsidc.org` should be at the top of [this comment list](https://github.com/smw-koops/https-everywhere/blob/ef9eb7b4fa2d5e8e6ce982e7ab8aeb916e380d11/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L9-L13) due to our special sorting rules.
* [This](https://github.com/smw-koops/https-everywhere/blob/ef9eb7b4fa2d5e8e6ce982e7ab8aeb916e380d11/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L16) should be `inside.nsidc.org`.
* [The `securecookie` tags](https://github.com/smw-koops/https-everywhere/blob/ef9eb7b4fa2d5e8e6ce982e7ab8aeb916e380d11/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L36-L45) should be combined in one regex. [Here](https://github.com/EFForg/https-everywhere/blob/fc3e2b0c0e9f567dea056108f72d2e317e437ff6/src/chrome/content/rules/Microsoft.xml#L614) is a particularly complicated example.
* That's a good catch on the `charctic-interactive-sea-ice-graph` problem but I'd really like to avoid the weird downgrade rule. What's wrong with just having the `exclusion`?

smw-koops:
@jeremyn The exclusion is here for a case when the user navigates to https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/, for example by manually changing the URL. The website doesn't redirect the user to http://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/, so a downgrade is neccessary.

jeremyn:
The `securecookie` should be this:

```xml
<securecookie host="((daacdata|smap_op)\.apps|(staging\.)?extranet|igloo|integration|qa|staging|support)\.nsidc.org" name=".+" />
```

which changes it so something is required before `nsidc.org`. The reason for this is obscure but important, see https://github.com/EFForg/https-everywhere/issues/2907 and particularly https://github.com/EFForg/https-everywhere/pull/7527#issuecomment-262405979 for an explanation.

jeremyn:
For https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/ , I think it's not our job to downgrade this URL to work around their HTTPS misconfiguration.

smw-koops:
@jeremyn Will that secure `nsidc.org` cookies too?

smw-koops:
@jeremyn If the user visits https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/, they would get a broken page with no obvious reason. I think it is necessary to fix this flaw instead of hoping the user would do the right thing. If something is broken and turning HTTPS Everywhere off fixes it the user would simply turn HTTPS Everywhere off and it's unlikely that they will turn it on again.

jeremyn:
> Will that secure nsidc.org cookies too?

No.

> If the user visits https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/, they would get a broken page with no obvious reason.

There are a huge number of URLs that are broken because of mixed content. It would be a major increase in the scope for HTTPS Everywhere to downgrade all of them, and there's nothing distinguishing this specific example from all the others.

smw-koops:
@jeremyn Is a separate `securecookie` tag required to secure `nsidc.org` cookies?

jeremyn:
No, we can't include `^nsidc.org` cookies. Including those cookies introduces the problem described in https://github.com/EFForg/https-everywhere/pull/7527#issuecomment-262405979 .

smw-koops:
@jeremyn In this case it's not just a mixed content warning but a completely broken page. See [Visa.xml](https-everywhere/src/chrome/content/rules/Visa.xml), which has a downgrade to prevent 404. I think it makes sense to have a downgrade here.

jeremyn:
There are a lot of pages broken to varying degrees. You can find examples by searching the rulesets and looking for top comments with `Mixed content`. There might be thousands of them and if we downgrade here, we're deciding that all of them really should be downgraded too.

Issue https://github.com/EFForg/https-everywhere/issues/6629 is a somewhat recent discussion on mixed content blocking. If you want to start a new discussion issue, that's fine. I personally would be interested to see what the expected use cases for downgrading are, since it's so uncommon.

smw-koops:
@jeremyn Was there any discussion about `Visa.xml` downgrade?

jeremyn:
Not that I know of. Looking at the downgrade [here](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/rules/Visa.xml#L54-L57):

```xml
	<!--	Doesn't redirect to http automatically (404s).
		Linked to from merchants/.	-->
	<rule from="^https://usa\.visa\.com/(?=about_visa/|download/|personal/|sitewide/)"
		to="http://usa.visa.com/" downgrade="1" />
```

I suppose the comment means that there were hardlinks to these HTTPS URLs in the wild on merchant pages that couldn't be fixed.

If you can show that the `charctic-interactive-sea-ice-graph` URL is hardlinked in the wild, maybe in printed magazine articles or major government websites, then I will agree that the cases are similar and a downgrade is appropriate. But unless it's actually happening though, the cases are not the same, or even if there are actual bad hardlinks but only on someone's personal blog.

smw-koops:
@jeremyn I didn't find any links to https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/. However I don't think the downgrade rule can cause any harm, while preventing accidental visits to the broken page.

smw-koops:
@jeremyn I think this should be discussed by people other than only me and @jeremyn.

jeremyn:
Okay, please feel free to create an issue to discuss it. In the meantime, it'd be good if you could remove the relevant parts from this pull request so we can move forward with what we have. If there's a later consensus to add the downgrade, we can create a new pull request to add that back in.

smw-koops:
@jeremyn I removed the downgrade rule. Check if everything is okay now.

jeremyn:
There is no reason to move `nsidc.org` to a mixed content ruleset, since the only remaining point of dispute is what happens if a user manually types https://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/ into their browser.

smw-koops:
@jeremyn I didn't move it to a mixed content ruleset. I just added a mixed content ruleset that covers the entire `nsidc.org` domain without an exclusion, so the page would only be excluded when mixed content blocking is on.

jeremyn:
`platform="mixedcontent"` means the ruleset is effectively off for every platform except Tor, see [here](https://www.eff.org/https-everywhere/rulesets#mixed-content-blocking-mcb).

jeremyn:
We should also change [this rule](https://github.com/smw-koops/https-everywhere/blob/d690fb6e9bc98c9ab678616c1ddb69022bf54798/src/chrome/content/rules/National-Snow-and-Ice-Data-Center.xml#L44) to:

```xml
	<rule from="^http://www\.nsidc\.org/" to="https://nsidc.org/" />
```

that is, replace `https?` with `http` in the `from` attribute. Does that sound right?

Also, since you're in there anyway, please move the `test` directly under the `exclusion` and add an indent, so it's:

```xml
	<exclusion pattern="^http://nsidc\.org/arcticseaicenews/charctic-interactive-sea-ice-graph/" />
		<test url="http://nsidc.org/arcticseaicenews/charctic-interactive-sea-ice-graph/" />
```

smw-koops:
@jeremyn Is that a new rule? Is it official? I think there is an unfinished discussion on that.

jeremyn:
Is what a new rule?

smw-koops:
@jeremyn https://www.nsidc.org is broken and users often type www before the domain name. Do you want to send them to a broken website?

smw-koops:
@jeremyn indent.

jeremyn:
I recently reviewed a pull request for `Ubuntu-CN.xml` with a `rule` that redirects `.org.cn` URLs to a `.com.cn` because the content is the same but `.org.cn` has an invalid certificate. That's the same problem as here, rewriting `www` to `^`. Here's the `rule` ([link](https://github.com/EFForg/https-everywhere/blob/25e0994c8fa5c3aa07438c8446f7034b15a45eef/src/chrome/content/rules/Ubuntu-CN.xml#L37)):

```xml
	<rule from="^http://(www\.|forum\.|linux\.|paste\.|wiki\.)?ubuntu\.org\.cn/" to="https://$1ubuntu.com.cn/" />
```

It's just an `http` rewrite, not `https?`.

My own opinion is that it's almost always outside scope to rewrite HTTPS URLs. It's the same sort of problem as downgrading. I agree it's arguable, especially in this case without the extra `downgrade` attribute. But the way I asked you do it is consistent with at least this one other rule.

For indenting, it's being discussed (see https://github.com/EFForg/https-everywhere/issues/7707#issuecomment-264001630) and we can talk about doing it other ways if you want, as long as the other way follows some standard.

jeremyn:
When I squash-and-merge, do you mind if I delete all of your commit messages? There are a lot of them.

smw-koops:
@jeremyn Yes, you can delete them.

jeremyn:
Thanks, merged.

PR #7649, opened 2016-11-20
Extract rules to top-level directory.

jkdf2:
The current documentation implies that everything in `src/` is for Firefox. The goal of this PR is to fix that impression.


jeremyn:
I can see what you're getting at here. I'm concerned though that people might see

```
src/chrome/content/rules  The rulesets (for all browsers) live here
```

and mistakenly think that there might actually be different rules for different browsers. Maybe they can expect to see a `src/chrome/content/rules/firefox`? In other words maybe it is introducing confusion that didn't exist before, while clearing up other confusion.

I agree it's a mess. `install-dev-dependencies.sh` softlinks `src/chrome/content/rules` to `rules/`, I guess in part to avoid the confusion you're trying to solve. Probably the best fix is to change things so that `rules/` is the actual directory and `src/chrome/content/rules` is the softlink. Note I'm not asking you to do that here.

jkdf2:
I agree with you that bringing `rules/` to a top-level directory seems to make much more sense. And that sounds like it would not be very difficult to implement. Let me take a look at that approach.

jkdf2:
Done :smile: if this approach works for everyone, it should be merged quickly to avoid conflicts with rule updates.

jkdf2:
The `TEST=fetch` failed only because docker could not run the tests in **all** xml files since it looks like they all changed when, in fact, they only changed directory.

jkdf2:
I guess I can't think of a good way to make that test pass, but the point is it's just looking like every single rule changed just because they moved directories.

jeremyn:
I appreciate your can-do attitude, but for security purposes, we should let one of the EFF staff do this, if they want to.

jkdf2:
Best to tag them or will they come across it eventually?

jeremyn:
@Hainish What do you think about arranging things so that the actual rules live in `rules/` with a soft link to `src/chrome/content/rules`, instead of the reverse we currently have?

cowlicks:
@jkdf2 I think the directory structure of the project will be changing a lot after we have totally transitioned to webextensions. The rulesets should be moved at that time.

PR #7650, opened 2016-11-20
Clean up Chrome manifest.

jkdf2:
1. Removes 'gecko' entry that does not apply to Chrome.

2. Prioritizes and visually groups HTTPS Everywhere extension
information separately from extension code configuration.


Hainish:
@jkdf2 looks good to me, rebase from master?

FYI: The `gecko` key was previously needed for the transition to WebExtensions within Firefox, it looks like it has been causing problems with Chrome more recently.  It is no longer needed within Firefox >= 48.  Sometime after March, ESR will be based on FF 52, and we should be able to safely transition to WebExtensions without `gecko`, but since the FF extension is still using the XPCOM codebase it's fine to remove it for now.

jkdf2:
@Hainish 

Thanks for the feedback. Merged master.

PR #7651, opened 2016-11-20
Update merge-rulesets.py

### LINKED ISSUE #2040: Chrome default.rulesets contains test URLs
Right now we assemble the rulesets for a release one of two ways. For Chrome, we use `utils/merge-rulesets.py`, which concatenates rulesets and removes some comments and whitespace with regexes. For Firefox, we use `utils/make-sqlite.py`, which parses XML, removes comments, whitespace, and certain elements and attributes, adds the `f=` attribute, and serializes it again.

When I added the `<test />` tag, I added stripping for it in `utils/make-sqlite.py` but forgot to add it to `utils/merge-rulesets.py`, which means default.rulesets for Chrome is a little bigger than it needs to be.

In the long run we should consolidate Chrome and Firefox on the same ruleset storage, sqlite. As a short term fix, we should remove `utils/merge-rulesets.*` and have `make-sqlite.py` output both a database and a concatenation of XML, renaming it in the process.

gloomy-ghost:
Related: #2040

### DIFF
--- a/utils/merge-rulesets.py
+++ b/utils/merge-rulesets.py
@@ -52,7 +52,7 @@ def rulesize():
   return len(open(rulesets_fn).read())
 
 def clean_up(rulefile):
-    """Remove extra whitespace and comments from a ruleset"""
+    """Remove extra whitespace, comments and tests from a ruleset"""
     comment_and_newline_pattern = re.compile(r"<!--.*?-->|\n|\r", flags=re.DOTALL)
     rulefile = comment_and_newline_pattern.sub('', rulefile)
     to_and_from_pattern = re.compile(r'\s*(from=)')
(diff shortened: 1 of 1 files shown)


jeremyn:
`clean_up` should probably be rewritten so it does XML parsing/extraction/creation rather than trying to use regexes to remove stuff.

gloomy-ghost:
This is just a quick fix since Chrome's ruleset has been bigger than Firefox's for 1 year

gloomy-ghost:
We have to parse XML rulesets and then translate them into JSON because HTTPS Everywhere for Firefox is using JSON. However, HTTPS Everywhere straightly uses XML on Chrome, and we have Travis-CI for testing every modification of ruleset to make sure that XML is valid, so there shouldn't be a problem. 

Also I'm not sure is the `f` field still needed, since I can't find `"view source XML" for rules` on Chrome (and Firefox actually)

gloomy-ghost:
```bash
$ ls -lh
total 24M
-rw-r--r-- 8.8M default.rulesets
-rw-r--r-- 7.8M rulesets.noTest
-rw-r--r-- 7.4M rulesets.noTest.noF
```

wow

gloomy-ghost:
@Hainish We have about 1MB `test` tags in Chrome's ruleset for years, so I just wanted to check if you would allow this patch before someone rewrite it entirely?

Hainish:
Tested, non-greedy test regex lgtm, merging

PR #7661, opened 2016-11-21
Added vr.fi

Foorack:
This pull request partially resolves issue #3069


jony0008:
Why doesn't it appear in the HTTPSE atlas after the last update?

PR #7669, opened 2016-11-23
TotalFreedom Ruleset

zD12:
Redirect http://TotalFreedom.boards.net to https://TotalFreedom.boards.net

### DIFF
--- a/src/chrome/content/rules/TotalFreedom.boards.net.xml
+++ b/src/chrome/content/rules/TotalFreedom.boards.net.xml
@@ -1,0 +1,4 @@
+<ruleset name="TotalFreedom.boards.net">
+	<target host="totalfreedom.boards.net" />
+	<rule from="^http:" to="https:" />
+</ruleset>


jeremyn:
Thanks for the pull request. This ruleset should really cover all of the `*.boards.net` subdomains, for example see [this Google search](https://www.google.com/#q=site:*.boards.net). Are you interested in doing that?

zD12:
@jeremyn, not exactly possible without inadvertently harming the website itself. 

All '*.boards.net' forums were originally designed for HTTP, not SSL. As a result, forcing all websites to use SSL will break many sites due to mixed traffic (i.e. 'This page is trying to load scripts from unauthenticated sources'). 

Though, if we neglected this (and we shouldn't), you'd want to add:

- *.proboards.com
- *.boards.net
- *.freeforums.net

jeremyn:
Right, we shouldn't actually add a wildcard `*.boards.net`. Wildcards are discouraged by the [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md). Instead we should list working domains individually, and note non-working domains as having problems.

J0WI:
I think in the case of "customer subdomains" it's okay to have wildcards. There might be a huge list that change often, so it's impossible for us to maintain.
Let's use a `platform="mixedcontent"` ruleset for the customer domains and a separate ruleset for CDN hosts.

jeremyn:
@J0WI 

> Let's use a `platform="mixedcontent"` ruleset for the customer domains and a separate ruleset for CDN hosts.

For "customer subdomains" I agree that the only likely problem is mixed content blocking, so moving them all into a `platform="mixedcontent"` should be fine.

What do you mean by "CDN hosts" in this example?

J0WI:
Some resources are loaded from the related `storage.proboards.com` host, which supports HTTPS.

jeremyn:
Oh, okay. Yes, any non-customer domains like `storage`, `www` and so on should go into a non-`mixedcontent` ruleset.

zoracon:
The testing environment has been stale for this PR, trying to readjust by open and closing

zoracon:
Ah the original repo is unknown. Apologies for the lag in reviews. If @zD12 Opens back up make sure to ping me for review

PR #7685, opened 2016-11-24
Replace non-standard for each statement with for

### LINKED ISSUE #7676: Broken on Firefox Nightly
Addon does not work on Firefox Nightly.

I am getting for syntax error in Browser Console.
https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/src/chrome/content/code/IOUtil.js#L130

Disable non-standard for-each:
https://bugzilla.mozilla.org/show_bug.cgi?id=1293305

smw-koops:
Closes #7676.

### DIFF
--- a/src/chrome/content/code/Cookie.js
+++ b/src/chrome/content/code/Cookie.js
@@ -72,7 +72,7 @@ Cookie.prototype = {
     this.value = nv.join('=') || '';
     
     var n, v;
-    for each (p in parts) {
+    for (p in parts) {
       nv = p.split("=");
       switch (n = nv[0].toLowerCase()) {
         case 'expires':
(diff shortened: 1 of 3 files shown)


evilpie:
This patch looks wrong to me. for each .. in and for .. in are different. 
For example:
```
for each (var x in ["a", "b"]) { print(x); } // Prints a, b 
for (var x in ["a", "b"]) { print(x); } // Prints 0, 1 (The array indexes) 
```

Sometimes it's possible to replace for each with [for .. of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

smw-koops:
@evilpie Nice one! This pull request was made in a rush without reading the docs carefully. If ES6 features can be used, then for of is indeed a good choice.

arai-a:
apparently all cases are iterating over array, so you can use `for...of`

smw-koops:
@arai-a I changed `for each...in` to `for...of`. I think that's correct.

Hainish:
lgtm. thanks for the fix @smw-koops!

PR #7692, opened 2016-11-24
sourcemod.net

galeksandrp:


src/chrome/content/rules/sourcemod.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="sourcemod.net">
+	<target host="sourcemod.net" />
+	<target host="www.sourcemod.net" />
+	<target host="docs.sourcemod.net" />
```
J0WI: See also doc.sourcemod.net

galeksandrp: Added.

J0WI:
Thanks!

PR #7693, opened 2016-11-25
Create AppMifile.com.xml

gloomy-ghost:


src/chrome/content/rules/AppMifile.com.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	For other domains belong to xiaomi, see Xiaomi.com.xml
+
+	Invalid Certificate:
+		- bbs.appmifile.com
+		- download.appmifile.com
```
J0WI: I also found u01.appmifile.com

PR #7695, opened 2016-11-25
Add sydsvenskan.se

rycee:


J0WI:
Please see also:
wahlgrenska.stiftelsen.sydsvenskan.se
apps.sydsvenskan.se

src/chrome/content/rules/Sydsvenskan.se.xml
```diff
@@ -0,0 +1,31 @@
...
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+
+	Mixed content:
+
+		- blogg.sydsvenskan.se
```
J0WI: Is it just mixed display content or mixed active content?

rycee: Sorry, I'm not certain what the difference is. Looking at the HTML code for the site it insists on using http for links within blogg.sydsvenskan.se and for img sources.

rycee:
@J0WI Many thanks for having a look at this. Good catch with wahlgrenska.stiftelsen.sydsvenskan.se and apps.sydsvenskan.se. I have force pushed an amended commit that adds these two hosts (one has working SSL, the other not).

src/chrome/content/rules/Sydsvenskan.se.xml
J0WI: Since browsers only block mixed active content by default (scripts, styles, objects...) you can enable this target.

rycee: Ah, that seems true. I've pushed the commit again with the blogg.sydsvenskan.se target enabled.

J0WI:
Thanks!

rycee:
And thank you for your useful review and ample patience :-)

PR #7696, opened 2016-11-26
Update rules.sh

gloomy-ghost:


test/rules.sh
```diff
@@ -29,16 +29,16 @@ if [ $# -gt 0 ] ; then
 fi
 if ! python2.7 test/rules/src/https_everywhere_checker/check_rules.py \
       test/rules/coverage.checker.config; then
-  cat <<EOF
+  echo '''
```
pgerber: Bash doesn't have strings that start with `'''`. Bash just concatenates the multiple strings you're creating.

As long as there is no escaped space in between two strings, they are just concatenated:
```bash
[ 'hello you' = 'h''e'l"l"o\ y'o'u ] && echo "equal!"
equal!
```

jeremyn:
Merged. I also removed the intermediate commit messages, I hope that's okay.

PR #7697, opened 2016-11-26
[Archive.org] Expanded coverage, removed nonexistent domains, documen…

galenide:
…ted Political-TV-Ad-Archive.xml ruleset


J0WI:
Some certs just expired yesterday, so let's give them some time.

J0WI:
Certs have been renewed. Thanks for contributing!

PR #7701, opened 2016-11-27
Create zdoom.xml

WilliamFeely:
Adding the website zdoom.org which is now supporting HTTPS on their servers.
Contains zdoom.org, forum.zdoom.org, and remilia.zdoom.org
There is no www subdomain in the .xml because the site does not normally use the www subdomain, and it is not in the HTTPS certificate.

### DIFF
--- a/src/chrome/content/rules/zdoom.xml
+++ b/src/chrome/content/rules/zdoom.xml
@@ -1,8 +1,12 @@
+<!--
+	Invalid certificate:
+		www.zdoom.org
+-->
 <ruleset name="ZDoom">
-        <target host="zdoom.org" />
-        <target host="forum.zdoom.org" />
-        <target host="remilia.zdoom.org" />
+	<target host="zdoom.org" />
+	<target host="forum.zdoom.org" />
+	<target host="remilia.zdoom.org" />
 
-        <rule from="^http:"
-                to="https:" />
+	<securecookie host=".+" name=".+" />
+	<rule from="^http:" to="https:" />
 </ruleset>
--- a/src/chrome/content/rules/zdoom.xml
+++ b/src/chrome/content/rules/zdoom.xml
@@ -1,0 +1,8 @@
+<ruleset name="ZDoom">
+        <target host="zdoom.org" />
+        <target host="forum.zdoom.org" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Please:
- [ ] We like to note important domains that work under plain HTTP but not under HTTPS in a comment at the top. In this case, http://www.zdoom.org works as a redirect. You can find comment examples in other rulesets, but for this ruleset, add this to the top:
```xml
<!--
	Invalid certificate:
		www.zdoom.org
-->
```
- [ ] Change the leading spaces to a single tab, following our [style guide](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md)
- [ ] Add a general `securecookie` between the `target`s and the `rule`:
```xml
```
- [ ] We like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) pull requests, which means we combine and rewrite pull requests in your name. Is that all right with you for this and any other pull requests you submit to us?

WilliamFeely:
Hi.  I'll update the xml file tomorrow.
Regarding the www subdomain, I sent the webmaster a PM regarding that issue, as I have noticed it too.
As for the securecookie line, is that to be placed in the xml file verbatim?

Regarding the squash-and-merge this is only my second PR I have ever made; the first one was approved and merged a while back.  I do not plan on making a lot of rulesets, so does this apply to me?

jeremyn:
Yes, the `securecookie` is verbatim. It's the same thing that's in the style guide that I linked.

I like to ask before doing a squash-and-merge because it technically rewrites commits in the user's name. I also ask for future permission as well so I don't need to ask again if you submit another pull request.

jeremyn:
Hi @WilliamFeely, will you be able to work on this soon?

WilliamFeely:
Sorry for the delay, I have been having personal issues and it kind of got put on the backburner.
Just updated the .xml file.

jeremyn:
No problem @WilliamFeely, thanks for getting back to this. Please confirm that you're okay with squash-and-merge and that will be it.

WilliamFeely:
Totally fine with the squash-and-merge.

jeremyn:
Thanks again, merged.

PR #7714, opened 2016-11-29
ljsear.ch

galeksandrp:


src/chrome/content/rules/ljsear.ch.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="ljsear.ch">
+	<target host="ljsear.ch" />
+
+	<target host="www.ljsear.ch" />
```
J0WI: This host does not exist.

galeksandrp: Removed.

PR #7715, opened 2016-11-29
servers.ru

galeksandrp:


src/chrome/content/rules/servers.ru.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+stage.servers.ru ¹
+
+
+¹ mismatch
+-->
+<ruleset name="servers.ru">
+	<target host="servers.ru" />
+	<target host="www.servers.ru" />
+	<target host="portal.servers.ru" />
```
J0WI: See also office.servers.ru

galeksandrp: Added.

PR #7735, opened 2016-12-02
M-net: add mined subdomains

numismatika:


src/chrome/content/rules/M-net.de.xml
```diff
@@ -1,9 +1,82 @@
 <ruleset name="M-net Telekommunikations GmbH" default_off="broken">
```
J0WI: Note that this is still `default_off`

numismatika: At the tine the request was made the mixed content broke the page and a few places had redirects to plain. I will check it again.

numismatika: So re-enabled.
I only could find a small mixed-content issue with a tracking element on the board. Shouldn't be an issue.

J0WI:
Travis reports now a bunch of certificate issues. 😿

Folant:
@J0WI https://github.com/EFForg/https-everywhere/pull/7492#issuecomment-281836532

numismatika:
the ones with the broken server hello are probably explained with the linked comment,bu t there are also a few other errors for domains that are broken or seemingly internal.

numismatika:

PR #7737, opened 2016-12-02
add rule for AD4mat

numismatika:


src/chrome/content/rules/AD4mat.xml
```diff
@@ -0,0 +1,34 @@
...
+	TimeOut:
+		- ch.ad4mat.info
+		- it.ad4mat.info
+-->
+<ruleset name="Ad4mat">
+
+	<target host="static.ad4mat.com"/>
+	<target host="ui.net.ad4mat.com"/>
+	<target host="www.ad4mat.com"/>
+	<target host="www.ad4mat.de"/>
```
J0WI: Please also add ^ad4mat.com and ^ad4mat.de

PR #7740, opened 2016-12-02
chetv.ru

galeksandrp:
Issue author: @galeksandrp


gloomy-ghost:
We can add a `securecookie` here, do you want to do it?

galeksandrp:
Added.

PR #7749, opened 2016-12-03
Update Unizar.es.xml

davidabian:


davidabian:
Are there real issues with webmail.unizar.es?

Folant:
Seems so. 
webmailunizares

davidabian:
Thanks, @Folant.

J0WI:
Some hosts are not passing our tests.

PR #7752, opened 2016-12-04
[1823.gov.hk] New ruleset

Marco129:
New ruleset for [1823 Online](http://www.1823.gov.hk)


src/chrome/content/rules/1823.gov.hk.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Invalid certificate:
+		1823.gov.hk
```
J0WI: See also `sc.1823.gov.hk` and `wcis.1823.gov.hk`

PR #7754, opened 2016-12-04
Create hbooker.com.xml

ivysrono:


J0WI:
I also have mixed active content on `www.hbooker.com`

PR #7775, opened 2016-12-05
ren.tv

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/ren.tv.xml
```diff
@@ -0,0 +1,36 @@
...
+
+¹ mismatch
+² refused
+³ timed out
+⁵ expired
+-->
+<ruleset name="ren.tv">
+	<target host="ren.tv" />
+	<target host="www.ren.tv" />
+	<target host="m.ren.tv" />
```
J0WI: I have mixed content issues on m.ren.tv.

galeksandrp: Commented.

PR #7807, opened 2016-12-08
echo.msk.ru

galeksandrp:
Issue author: @galeksandrp


J0WI:
Why is this only for `*.echo.msk.ru`?

PR #7813, opened 2016-12-08
keenetic.net

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/keenetic.net.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+keenetic.net ¹
+www.keenetic.net ¹
```
J0WI: They WFM, see also help.keenetic.net

PR #7828, opened 2016-12-09
openwebtorrent.com

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/openwebtorrent.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+tracker.openwebtorrent.com non-2xx http code
+
+
+-->
+<ruleset name="openwebtorrent.com">
+	<target host="openwebtorrent.com" />
+
+	<target host="www.openwebtorrent.com" />
```
J0WI: This host does not exist.

galeksandrp: Removed.

PR #7832, opened 2016-12-10
ipleak.net

galeksandrp:
Issue author: @galeksandrp


J0WI:
I would be fine with a wildcard in this case, but thanks anyway!

PR #4875, opened 2016-05-21
Create ust.hk

cschanaj:
New ruleset for The Hong Kong University of Science and Technology, including the homepage www.ust.hk and some of its subdomain.

Issue #4389, opened 2016-03-17
Create port specific rules?


Bisaloo:
The domain `halley.ens.fr` has a rule in https everywhere and it works fine most of the time

https://halley.ens.fr/

However, some uses require to connect to port 4550 and then, https everywhere breaks it because ssl is not / badly (?) configured for this port.

https://halley.ens.fr:4550/

I could simply create a PR to disable this subdomain in the rule but my question is more general: 

**Is there a way to specify port specific rules? If no, maybe it should be implemented** (just a suggestion given I wouldn't be able to do it myself)

J0WI:
Are you looking for something like this?
```
<target host="foo.example" />

<rule from="^http://foo\.example:2083/"
           to="https://foo\.example:2083/" />
```

Bisaloo:
Well, that was kind of obvious in hindsight. Thank you!

I submitted a PR (#8969) to fix what lead me to ask this question in the first place.

PR #7836, opened 2016-12-10
Create ust.hk.xml

cschanaj:
Fixed the ruleset style and added more hosts from #4875. An email on the invalid certificate of https://o365.ust.hk has been sent to the university administration, I will add https://o365.ust.hk to the targets once the host is fixed. BTW, https://hrmsxprod.psft.ust.hk serves an invalid certificate and is non-functional on the default port but  a valid one on https://hrmsxprod.psft.ust.hk:8044, any clarification on whether a port specific rule should/ possibly be created #4389?

### DIFF
--- a/src/chrome/content/rules/ust.hk.xml
+++ b/src/chrome/content/rules/ust.hk.xml
@@ -21,6 +21,7 @@
 
 		Timeout:
 			- www.bm.ust.hk
+			- teaching.ust.hk
 			- w9.ust.hk
 
 		MCB:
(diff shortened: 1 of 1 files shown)


cschanaj:
It seems that they are not going to fix https://o365.ust.hk according to the email reply, I will leave the host in the comment as is.

J0WI:
Please have a look at the failing tests.

J0WI:
Thanks!

PR #7839, opened 2016-12-10
rom1504.fr

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/rom1504.fr.xml
+++ b/src/chrome/content/rules/rom1504.fr.xml
@@ -1,4 +1,5 @@
 <!--
+www.rom1504.fr different content
 dbpedia.rom1504.fr ¹
 flying-squid.rom1504.fr ¹
 sharelatex.rom1504.fr different content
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/rom1504.fr.xml
```diff
@@ -0,0 +1,25 @@
...
+	<target host="gitlab-ci.rom1504.fr" />
+	<target host="instant.rom1504.fr" />
+	<target host="logs.rom1504.fr" />
+	<target host="phpmyadmin.rom1504.fr" />
+	<target host="tracker.rom1504.fr" />
+
+	<target host="www.rom1504.fr" />
+
+	<rule from="^http://www\.rom1504\.fr/"
+		to="https://rom1504.fr/" />
```
J0WI: They have not the same content.

galeksandrp: Removed.

PR #7844, opened 2016-12-10
file.pizza

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/file.pizza.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="file.pizza">
+	<target host="file.pizza" />
+
+	<target host="www.file.pizza" />
```
J0WI: This host does not exist.

galeksandrp: Removed.

PR #7848, opened 2016-12-10
youshark.neocities.org

galeksandrp:
Issue author: @galeksandrp


J0WI:
Why is this just for youshark.neocities.org and not neocities.org in general?

PR #7851, opened 2016-12-10
alhur.es

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/alhur.es.xml
```diff
@@ -0,0 +1,25 @@
...
+	<target host="dpelp.alhur.es" />
+	<target host="fiatjaf.alhur.es" />
+	<target host="repos.alhur.es" />
+	<target host="rosetta.alhur.es" />
+	<target host="sheets.alhur.es" />
+	<target host="tabelas.alhur.es" />
+	<target host="temperos.alhur.es" />
+	<target host="templates.alhur.es" />
+
+	<target host="www.alhur.es" />
```
J0WI: This host does not exist

galeksandrp: Removed.

PR #7857, opened 2016-12-11
Update JD.com.xml

ivysrono:


J0WI:
`nj.jd.com` resolves to multiple IPs, some of them are working and some of them are bad configured.

PR #7859, opened 2016-12-11
Update Qh_img.com.xml

ivysrono:


gloomy-ghost:
I think it will be better to rename the ruleset without underscore, but it's ok if you don't want to.

PR #7869, opened 2016-12-11
clodo.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/clodo.ru.xml
```diff
@@ -0,0 +1,267 @@
...
+62-76-45-223.clodo.ru ⁷
+62-76-46-153.clodo.ru ⁷
+62-76-47-249.clodo.ru ⁷
+69421-1.clodo.ru ⁴
+72151-2.clodo.ru ⁴
+85-143-166-11.clodo.ru ⁷
+85-143-166-120.clodo.ru ⁷
+85-143-166-154.clodo.ru ⁷
+85-143-166-6.clodo.ru ⁷
+85-143-166-7.clodo.ru ⁷
```
J0WI: Is there any orderliness between all those numeric subdomains that we can avoid that huge list?

galeksandrp: They are using wildcard domain *.clodo.ru which resolves to webcluster.clodo.ru [62.76.182.7] which responds with protocol error (ERR_EMPTY_RESPONSE) on any domain.

PR #7873, opened 2016-12-12
semver.io

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/semver.io.xml
+++ b/src/chrome/content/rules/semver.io.xml
@@ -1,10 +1,9 @@
+<!--
+www.semver.io different content
+
+-->
 <ruleset name="semver.io">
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/semver.io.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="semver.io">
+	<target host="semver.io" />
+
+	<target host="www.semver.io" />
+
+	<rule from="^http://www\.semver\.io/"
+		to="https://semver.io/" />
```
J0WI: ^ and www serve not the same content.

galeksandrp: Commented.

PR #7874, opened 2016-12-12
Remove NYTimes downgrade rule.

jsha:
Per NYTimes engineer, this is now causing breakage as they migrate to fully
support HTTPS.

### DIFF
--- a/src/chrome/content/rules/NYTimes.xml
+++ b/src/chrome/content/rules/NYTimes.xml
@@ -243,11 +243,4 @@
 	<rule from="^http:"
 		to="https:" />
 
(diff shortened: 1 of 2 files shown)


gloomy-ghost:
lgtm, [here](https://www.nytimes.com/svc/community/V3/requestHandler?callback=jQuery17103594748933300256_1375345664116&method=get&cmd=GetCommentsAll&path=http://www.nytimes.com/2013/08/04/education/edlife/how-georgetown-became-a-gay-friendly-campus.html&offset=1) is a test if someone needs to confirm.

However this is actually a `default_off` ruleset...

PR #7875, opened 2016-12-12
Create priceshoes.com.xml

gloomy-ghost:


jeremyn:
Here's how I would write it:

```xml
<!--
	Time out:
	- priceshoes.com
	- centenario.priceshoes.com
	- licitaciones.compras.priceshoes.com
	- distribuidores.priceshoes.com
	- licitaciones.priceshoes.com
	- proveedor.priceshoes.com
	- proveedores.priceshoes.com

-->
<ruleset name="PriceShoes.com">
	<target host="priceshoes.com" />
	<target host="www.priceshoes.com" />

	<rule from="^http://priceshoes\.com/" to="https://www.priceshoes.com/" />
</ruleset>
```

and the filename would be `PriceShoes.com.xml`.

gloomy-ghost:
Are there any issues with the `securecookie`?

Also I'm also not sure about the `rule`. I would keep it two lines when it's longer than just `from="http:" to ="https:"`.

jeremyn:
The `www` `securecookie` is fine. I'm not sure why I didn't include it in my version.

Two lines is fine if you want. Note that tab formatting doesn't necessarily line things up nicely, depending on how many spaces the viewer uses per tab. So multiline rules can look jagged. Also the viewer can turn on line wrap if they want in their local editor, though not when looking at it in GitHub.

jeremyn:
Thanks, merged. I also deleted the intermediate commit messages.

PR #7892, opened 2016-12-14
Replace securecookie tags

gloomy-ghost:
See ruleset-style.md and discussion at #7512


gloomy-ghost:
Just realized that updating old rulesets which have insufficient test is a big trouble for travis-ci... 

@Hainish What do you think?

jeremyn:
In https://github.com/EFForg/https-everywhere/pull/8128 where @Hainish removed preloaded domains, we just ignored simple fetch errors, in fact we ignored that the tests didn't even complete.

Is the only thing happening in this PR that `.`/`.` in `securecookie`s are rewritten to `.+`/`.+`? I don't think it's worth modifying 800+ rulesets just for that style change.

Hainish:
I think moving forward we should have a test to ensure that new rules use the preferred formatting, but I figure it's okay if old rules stick to the functionally equivalent `"."`

PR #7893, opened 2016-12-14
Update Zhihu.xml

ivysrono:


J0WI:
I also found:
worldcup2014.zhihu.com
sugar.zhihu.com

You may remove the partial flag since most sites are working with https.

PR #7894, opened 2016-12-14
Update duapp.com.xml

ivysrono:


J0WI:
What about other subdomains here?

ivysrono:
@J0WI No matter Ok or MCB, all of them are too much to list.
I do not know how to do more.

PR #7897, opened 2016-12-14
Update Wolfram_Alpha.xml

Folant:


src/chrome/content/rules/Wolfram_Alpha.xml
```diff
@@ -1,102 +1,47 @@
...
+Refused: 
+	community.wolframalpha.com
+    
+Timeout: 
+	blog.wolframalpha.com
+	nba.devel.wolframalpha.com
+	homeworkday.wolframalpha.com
+	m.wolframalpha.com
+	products.wolframalpha.com
+	www2.wolframalpha.com
```
J0WI: developer, m and products are working fine for me.

src/chrome/content/rules/Wolfram_Alpha.xml
```diff
@@ -1,102 +1,47 @@
...
+	<target host="www4b.wolframalpha.com" />
+	<target host="www4c.wolframalpha.com" />
+	<target host="www4d.wolframalpha.com" />
+	<target host="www4f.wolframalpha.com" />
+	<target host="www5a.wolframalpha.com" />
+	<target host="www5b.wolframalpha.com" />
+	
+	<rule from="^http://api-(cn|maps|tw)\.wolframalpha\.com/" to="https://www.wolframalpha.com/" />
+	<rule from="^http://volunteer\.wolframalpha\.com/" to="https://www.wolframalpha.com/" />
+	<rule from="^http://(www[0-9][0-9]\.)wolframalpha\.com/" to="https://www.wolframalpha.com/" />
```
J0WI: Please document why those redirects are needed.
It seems that any `www\d\d` subdomain works, so you can use a wildcard target here.

PR #7900, opened 2016-12-14
Create sysadmins.ru.xml

Folant:


src/chrome/content/rules/sysadmins.ru.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+Invalid certificate:
+	eng.sysadmins.ru
+	forum.sysadmins.ru
```
J0WI: See also mail.sysadmins.ru

PR #7909, opened 2016-12-16
Create nobelprize.org.xml

Folant:


src/chrome/content/rules/nobelprize.org.xml
```diff
@@ -0,0 +1,21 @@
...
+-->
+<ruleset name="nobelprize.org">
+	<target host="nobelprize.org" />
+	<target host="www.nobelprize.org" />
+	<target host="images.nobelprize.org" />
+	<target host="preview.nobelprize.org" />
+	<target host="static.nobelprize.org" />
+	
+	<rule from="^http://nobelprize\.org/" to="https://www.nobelprize.org/" />
+	<rule from="^http://(images|static)\.nobelprize\.org/" to="https://www.nobelprize.org/" />
```
J0WI: Please document why those redirects are needed.

PR #7914, opened 2016-12-17
Update XDA-Developers.xml and Create xda-cdn.com.xml

ivysrono:


ivysrono:
By the way, the XDA-Developers.xml is too complicated to understand.
@jeremyn

PR #7917, opened 2016-12-17
Disabled Memrise ruleset due to breakage

Foorack:
Review and other sections get stuck on loading screen.


J0WI:
I could not find the broken parts, can you give me an example?

Foorack:
Can't find it either. Must have been fixed by now.

PR #7925, opened 2016-12-20
www.streamingeverywhere.com

LiveChief:
Issue author: @LiveChief


src/chrome/content/rules/www.streamingeverywhere.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="www.streamingeverywhere.com">
+	<target host="www.streamingeverywhere.com" />
+
+	<target host="www.www.streamingeverywhere.com" />
```
J0WI: This host does not exist, but you should add `streamingeverywhere.com` (without www prefix).

PR #7929, opened 2016-12-20
Update Gravity4.com.xml

ivysrono:


src/chrome/content/rules/Gravity4.com.xml
```diff
@@ -1,42 +1,15 @@
...
 -->
-<ruleset name="Gravity4.com (partial)">
 
-	<!--	Direct rewrites:
-				-->
-	<!--target host="gravity4.com" /-->
+<ruleset name="Gravity4.com">
+	<target host="gravity4.com" />
+	<target host="www.gravity4.com" />
 	<target host="jira.gravity4.com" />
```
J0WI: os.gravity4.com

J0WI: support.gravity4.com

Issue #7903, opened 2016-12-15
idnes.cz does not work with HTTPS Everywhere enabled


Alex-007-:
Hello,
I can't load following page because of plugin in Firefox and Chrome:
http://technet.idnes.cz/
I mention this page, but it is related to most of the pages on:
idnes.cz
Issue is solved as soon as I disable the addon (it is NOT enough to switch it off). I use "HTTPS everywhere" already long time, but this issue appeared as new.

jeremyn:
Thanks for the report. I've updated the issue title to be more specific.

J0WI:
@Alex-007- I don't have any issues, can you confirm that this has been fixed?

jankratochvil:
idnes.cz does have https support now; although it is still mixed content.  Still the rules are fragile as they test only rajce websites for all idnes.cz sites and rajce appears to have some separate administration (as in the past rajce had https working while the rest of idnes.cz did not).

J0WI:
So can we just solve this by adding `platform="mixedcontent"`?

krejmaj:
All issues with idnes.cz should be fixed except for topkontakt.idnes.cz which is just a redirect

PR #7941, opened 2016-12-21
Remove idnes.cz rules

finwe:
idnes.cz removed https access which renders its sites unreachable with current rules.
Only rajce.idnes.cz and a few other subdomains remains accessible via https


jeremyn:
For reference: this pull request looks relevant to issue https://github.com/EFForg/https-everywhere/issues/7903.

jankratochvil:
idnes.cz has https support now; although it is still mixed content.

finwe:
Closing in favor of potential future PR favoring that.

PR #7943, opened 2016-12-21
Update Yeah.net.xml

ivysrono:


src/chrome/content/rules/Yeah.net.xml
```diff
@@ -14,15 +15,26 @@
 <ruleset name="Yeah.net (partial)">
 
 	<!--	Complications:	-->
+	<target host="yeah.net" />
+	<target host="mail.yeah.net" />
+		<exclusion pattern="^http://(mail\.)?yeah\.net/(?!favicon\.ico)" />
```
J0WI: Please document why this exclusion is required.

ivysrono: Oh, it's complicated.

J0WI:
Thanks!

PR #7948, opened 2016-12-22
Update T411.xml

Harest:
t411 moved from .ch to .li


J0WI:
The .ch host do no longer exist, so they should be removed.

PR #7951, opened 2016-12-23
Remove shopping.rsbp.org.uk

illicitonion:
It's currently broken. I have gotten in touch, but there will naturally
be some lag in getting a fix.

### DIFF
--- a/src/chrome/content/rules/RSPB.xml
+++ b/src/chrome/content/rules/RSPB.xml
@@ -1,7 +1,10 @@
+<!--
+	Shopping cart functionality is broken on shopping.rspb.org.uk because it makes XHR which are hard-coded to http://
+-->
+
 <ruleset name="RSPB">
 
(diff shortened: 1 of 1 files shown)


J0WI:
Could you please use regex exclusions for the specific locations?

gloomy-ghost:
Is this still an issue?

illicitonion:
Sorry, just checked this - the entire subdomain is broken, so I guess I can use a regex exclusion, but it will look like:

```
<target host="shopping.rspb.org.uk" />
<exclusion pattern="^http://shopping\.rspb\.org\.uk/.*" />
```

which seems a little pointless?

illicitonion:
(Each of the pages will load and display, but if you try to interact with them, they are all broken)

gloomy-ghost:
Yes I see, let's just merge it

PR #7957, opened 2016-12-24
Create showapi.com.xml

gloomy-ghost:


jeremyn:
Here's what I would do:

```xml
<!--
	200 with plain HTTP, 404 with HTTPS:
	- docs.showapi.com
	  example: http://docs.showapi.com/jdk7/index.html

	503 with plain HTTP, refused with HTTPS:
	- wap.showapi.com

	No working URL known:
	- static.showapi.com

-->
<ruleset name="ShowAPI.com">
	<target host="showapi.com" />
	<target host="www.showapi.com" />
	<target host="bbs.showapi.com" />
	<target host="route.showapi.com" />

	<rule from="^http://bbs\.showapi\.com/(archiver|data|source|static)/" to="https://bbs.showapi.com/$1/" />
		<test url="http://bbs.showapi.com/archiver/" />
		<test url="http://bbs.showapi.com/data/" />
		<test url="http://bbs.showapi.com/source/" />
		<test url="http://bbs.showapi.com/static/" />
		<!-- This exception disables the implicit test created by the target. -->
		<exclusion pattern="^http://bbs\.showapi\.com/$" />
	<rule from="^http://(www\.|route\.)?showapi.com/" to="https://$1showapi.com/" />
</ruleset>
```

Also the filename would be `ShowAPI.com.xml`.

Do you want to add `dn-showapi.qbox.me` to [`Qbox.me.xml`](https://github.com/EFForg/https-everywhere/blob/3f48ae8e47696d846bb5d809dffaa3af40cb59ab/src/chrome/content/rules/Qbox.me.xml)?

gloomy-ghost:
I was using exceptions to keep the plain rewriting rule, otherwise the rules are basically just repeating subdomains that are already included in `target`s. Maybe complicated rules that don't do anything else than http>https redirecting would be a good method if there were only wildcard targets. 

@Hainish Would it be possible to add something likes a `no_plain_rewriting` attribute in `target`s so that a `target` which is only partially redirected by `rule` won't be redirected by http:>https: rule in the end of the ruleset?

jeremyn:
The ruleset I wrote in https://github.com/EFForg/https-everywhere/pull/7957#issuecomment-278825904 is simple: for these domains, do these rewrites. This is a lot easier to understand in my opinion than saying rewrite everything except these other certain things.

I'm opposed to something like `no_plain_rewriting` because that means that to figure out what the rewrite logic is, I can't just look at the `rule`, I also have to also possibly look at attributes of the `target`s.

Hainish:
I agree with @jeremyn on this - adding an attribute which changes how rewrites are performed also complicates things for our downstream dependencies, like Brave browser or libhttpseverywhere.

jeremyn:
@gloomy-ghost Can we finish this?

gloomy-ghost:
Is there a url for actual resource under http://bbs.showapi.com/source/? I can't find any.

BTW, I still think that listing all subdomains in a plain rewriting `rule` is repeating the `target`s' jobs. If everything is in `rule`s, I can replace all `target`s with a wildcard and it won't change anything. Then why there are `target` tags?

jeremyn:
I can't find anything under http://bbs.showapi.com/source/ either, but it returns a 200 even though it's empty so that's good enough for me.

I don't know how to answer the question in your second paragraph. Maybe a better system would merge the concept of `target` and `rule`.

jeremyn:
- [ ] https://docs.showapi.com is `Different HTTP/HTTPS content`, not `404`
- [ ] The general case `*.showapi.com` works for me (they all redirect to `www`), so you can remove it from the top comment
- [ ] Also `u5335.showapi.com` redirects to https://www.showapi.com like the general `*` case, so you can delete that from the top comment too
- [ ] The `u(\d+)\.showapi\.com` `rule` [here](https://github.com/gloomy-ghost/https-everywhere/blob/ec08d9776076863405cd0afea6ea54acbac35f69/src/chrome/content/rules/showapi.com.xml#L46-L48) doesn't have a target, so please either delete the `rule` or add a `target`
- [ ] I would move the `rule` for `^` [here](https://github.com/gloomy-ghost/https-everywhere/blob/ec08d9776076863405cd0afea6ea54acbac35f69/src/chrome/content/rules/showapi.com.xml#L40-L41) into the larger rule [here](https://github.com/gloomy-ghost/https-everywhere/blob/ec08d9776076863405cd0afea6ea54acbac35f69/src/chrome/content/rules/showapi.com.xml#L43-L44), using `(www\.|admin\.|api\.| ...)?` but if you prefer the style you have now, that's fine with me.

jeremyn:
I made the changes I described in https://github.com/EFForg/https-everywhere/pull/7957#issuecomment-297184435.

@J0WI Can you review this please?

PR #7982, opened 2016-12-27
Create Centralfield.com.xml

cschanaj:


src/chrome/content/rules/Centralfield.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Non-functional hosts
+		Certificate mismatch:
+			- extreme.centralfield.com
+			- mail.centralfield.com
+			- mbox.centralfield.com
+-->
+<ruleset name="Centralfield.com">
+	<target host="centralfield.com" />
+	<target host="www.centralfield.com" />
```
J0WI: Please also add `shop.centralfield.com`

PR #7983, opened 2016-12-27
Added rules for rcdb.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/rcdb.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Roller Coaster DataBase">
+
+	<target host="rcdb.com" />
+	<target host="www.rcdb.com" />
+
+	<rule from="^http://www\.rcdb\.com/" to="https://rcdb.com/"/>
```
J0WI: We don't hijack server side redirects, except when they are broken.

J0WI:
Thanks!

PR #7988, opened 2016-12-28
Create soso.com.xml

ivysrono:


J0WI:
What about other *.soso.com hosts?

PR #8003, opened 2016-12-29
update deezer.com -> default off because it breaks the navigation links

luciancor:
disable deezer.com rules  by default as it breaks navigation links


jwatt:
It was just pointed out to me that this pull request exists after I made my own pull request for Deezer.xml in pull request #8418. However, I investigated *why* it is that deezer.com is broken by the Deezer.xml rules and documented the reasons in the commit in my pull request. I would hope those comments would save time for anyone who wants to get Deezer.xml working in future. Given that, could you cancel this pull request, @luciancor?

J0WI:
closing in favor of #8418

PR #8010, opened 2016-12-30
Ruleset for dpskolkata.com

ankitpati:
Added a ruleset for a [site](https://www.dpskolkata.com/ "Delhi Public School, Ruby Park").

### DIFF
--- a/src/chrome/content/rules/Dpskolkata.com.xml
+++ b/src/chrome/content/rules/Dpskolkata.com.xml
@@ -1,0 +1,8 @@
+<ruleset name="Dpskolkata.com">
+	<target host="dpskolkata.com" />
+	<target host="www.dpskolkata.com" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


J0WI:
This site has some mixed content issues. Can you confirm that this breaks no functionality?

ankitpati:
No, the mixed content issues do not break any functionality. In fact, if the rule is added, users of HTTPS Everywhere will magically see the mixed content issues disappear.

Manually navigating to the HTTPS version gets you the HTML content over HTTPS, but it loads resources over insecure connections because the URLs have HTTP hardcoded into them. HTTPS Everywhere will rewrite all of those URLs, and make the page completely secure.

J0WI:
The problem is, that the mixed content blocker of the browser blocks this content before we can redirect it.

ankitpati:
Mixed content blocker? I believe that requires user action to enable. None of the browsers have it enabled by default. Most will just warn.

J0WI:
All major browsers block mixed active content by default:
https://security.googleblog.com/2011/06/trying-to-end-mixed-scripting.html
https://blogs.msdn.microsoft.com/ie/2011/06/23/internet-explorer-9-security-part-4-protecting-consumers-from-malicious-mixed-content/
https://bugzilla.mozilla.org/show_bug.cgi?id=834836

ankitpati:
Mixed ***active*** content gets blocked. The site serves all scripts over HTTPS when the HTML is requested over HTTPS. I tried browsing to all parts of the site, and Chrome did not report any blocked scripts.

J0WI:
Firefox reports me mixed active content from `http://ajax.googleapis.com/ajax/libs/jqueryui/1.10.3/jquery-ui.js`

ankitpati:
Yes, there are mixed active content warnings, but they did not seem to affect the page in the browsing that I did, in both Chrome and Firefox.

Unrelated to this PR, but something that HTTPS Everywhere should consider: why not rewrite all recognised URLs in a page to HTTPS before the browser has a chance to take over? For example, the specific URL you have mentioned is already known to HTTPS Everywhere, and visiting it directly forces it to HTTPS.

J0WI:
Thanks for your feedback. Therefore I'm fine with merging this.

For security reasons browsers do not allow Add-Ons to bypass their mixed content blocker like this. We could disable the mixed content blocker completely, like Tor does, but this brings MITM risks for all other sites, that are not redirected by HTTPS Everywhere.

ankitpati:
Thank you for the merge.

PR #8012, opened 2016-12-30
add rule for Adventori.com

numismatika:


src/chrome/content/rules/Adventori.com.xml
```diff
@@ -0,0 +1,58 @@
...
+			- d01
+			- d02
+			- d03
+			- d04
+			- d05
+			- d06
+-->
+<ruleset name="Adventori.com">
+
+	<target host="www.adventori.com" />
```
J0WI: Please also add ^adventori.com

J0WI:
ping @numismatika

PR #8020, opened 2016-12-31
Create hearty.me.xml

chennien:


src/chrome/content/rules/hearty.me.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="hearty.me">
+	<target host="hearty.me" />
+	<target host="cdn.hearty.me" />
+	<target host="www.hearty.me" />
```
J0WI: * intro.hearty.me

J0WI:
See also: #8047 #8076 #8077

chennien:
@J0WI: I submitted these domains in separate xml files due to you asked me to do it separately. May I know what should I do next? Thanks a million.

J0WI:
In this case these are just alias domains. So I'm fine with having them in one file.
Otherwise you should mention the related ruleset files in comments, that we are aware of them when we maintain the rules.

J0WI:
I'm closing this, because this host is already preloaded in all major browser.

PR #8040, opened 2017-01-01
Create mohot.fit.xml

chennien:


J0WI:
This host seems not to be active anymore.

PR #8043, opened 2017-01-01
Create dfctaiwan.org.xml

chennien:


jeremyn:
`dfctaiwan.org` is currently preloaded, so we don't need a ruleset for it.

PR #8048, opened 2017-01-01
Update and rename openload.co.xml to Openload.xml

cschanaj:
Updated comments, added more host for `openload.co` to improve ruleset coverage, as well as for `*.oloadcdn.net`.


J0WI:
Thanks!

PR #8064, opened 2017-01-02
Create thenewslens.com.xml

chennien:


J0WI:
I'll prefer #8745 since it covers more subdomains.

PR #8077, opened 2017-01-02
Create hearty.blog.xml

chennien:


jeremyn:
`hearty.blog` is currently HSTS preloaded, so we don't need a ruleset for it.

PR #8112, opened 2017-01-04
Update TouTiao.xml

ivysrono:


src/chrome/content/rules/TouTiao.xml
```diff
@@ -2,36 +2,48 @@
...
+	<!-- Directly: -->
 	<target host="toutiao.com" />
 	<target host="www.toutiao.com" />
-		<test url="http://toutiao.com/a6361374475379998978/" />
-		<test url="http://toutiao.com/i6360544217684509186/" />
-		<test url="http://toutiao.com/i6361599096398021121/" />
 		<test url="http://www.toutiao.com/a6361374475379998978/" />
 		<test url="http://www.toutiao.com/i6360544217684509186/" />
-		<test url="http://www.toutiao.com/i6361599096398021121/" />
-	<exclusion pattern="^http://(www\.)?toutiao\.com/(a|i)(\d+)/" />
```
J0WI: I'm a bit confused about this change: You removed the exclusion but referring to your comments there are still mixed content issues?

ivysrono: Maybe I should delete the comments.
In http mode, the 热门视频 and 精彩图片 in right bar will be broken just like MCB.
HTTPS mode has fixed this problem.

ivysrono: @J0WI OK now?

PR #8115, opened 2017-01-05
Delete Kuantokusta.pt.xml

ekaris:
Server now uses HTTPS with HSTS with long duration, no more need for this rule.


Hainish:
Since Kuantokusta.pt is not preloaded, this site is still vulnerable on first load to the [SSLStrip attack](https://moxie.org/software/sslstrip/).  In order to qualify for removal, this site must be included in the HSTS preload list for all the browsers that we support.  See https://github.com/EFForg/https-everywhere/issues/7126 for additional context

PR #8117, opened 2017-01-05
Create hkicl.com.hk.xml

cschanaj:


src/chrome/content/rules/hkicl.com.hk.xml
```diff
@@ -0,0 +1,55 @@
...
+		Couldn't connect to server:
+			 - hkicl.com.hk
+			 - ebpp.hkicl.com.hk
+			 - uat.ebpp.hkicl.com.hk
+			 - ns10.hkicl.com.hk
+			 - ns11.hkicl.com.hk
+			 - ns12.hkicl.com.hk
+			 - ns9.hkicl.com.hk
+
+		Timeout was reached:
```
J0WI: You don't need to lists hosts, that don't respond on HTTP requests.

PR #8118, opened 2017-01-05
Create files.tugatech.com.pt.xml

djprmf:


J0WI:
@djprmf could you expand the coverage to all hosts of tugatech.com.pt?

djprmf:
@J0WI I have added some more that have been implemented with SSL in the mean time.
Thanks

src/chrome/content/rules/files.tugatech.com.pt.xml
```diff
@@ -1,10 +1,12 @@
 <ruleset name="Files.tugatech.com.pt">
```
J0WI: Please change the name to `tugatech.com.pt`

J0WI: Please add comments for the mismatch on `blog.tugatech.com.pt`, `magazine.tugatech.com.pt` and `webradio.tugatech.com.pt`

src/chrome/content/rules/files.tugatech.com.pt.xml
```diff
@@ -1,10 +1,12 @@
...
 
+	<target host="tugatech.com.pt" />
```
J0WI: Please also add `www.tugatech.com.pt`

djprmf:
@J0WI I have added the subdomains (magazine,webradio,blog) with SSL support, so this can be more easily approved.
Also added the www version of the main domain. Thanks

J0WI:
adding non-working domains as targets won't work. You need to add them as comments, see https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts

djprmf:
@J0WI those targets have been added with SSL support, so they are working.
Check here:
https://www.ssllabs.com/ssltest/analyze.html?d=blog.tugatech.com.pt&s=104.27.157.87&latest
https://www.ssllabs.com/ssltest/analyze.html?d=magazine.tugatech.com.pt&s=104.27.157.87&latest
https://www.ssllabs.com/ssltest/analyze.html?d=webradio.tugatech.com.pt&s=104.27.157.87&latest

J0WI:
Are you the owner of this host?
Then you should read https://hstspreload.org/

PR #8150, opened 2017-01-07
Added rules for vhs-karlsruhe.de.

Jochen-A-Fuerbacher:


src/chrome/content/rules/vhs-karlsruhe.de.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="vhs Karlsruhe">
+
+	<target host="www.vhs-karlsruhe.de" />
+	<target host="moodle.vhs-karlsruhe.de" />
```
J0WI: Please see also:
vhs-karlsruhe.de
upgrade.vhs-karlsruhe.de

PR #8151, opened 2017-01-07
Added rules for vhs-lu.de.

Jochen-A-Fuerbacher:


src/chrome/content/rules/vhs-lu.de.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Volkshochschule Ludwigshafen">
+
+	<target host="www.vhs-lu.de" />
```
J0WI: Please also add vhs-lu.de (without www prefix).

PR #8152, opened 2017-01-07
Added rules for klilu.de.

Jochen-A-Fuerbacher:


src/chrome/content/rules/klilu.de.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Klinikum Ludwigshafen">
```
J0WI: Please also document that HTTPS is refused on mtaschule.klilu.de

PR #8154, opened 2017-01-07
Create l214.com.xml

Bisaloo:


src/chrome/content/rules/l214.com.xml
```diff
@@ -0,0 +1,41 @@
...
+	<target host="boutique.l214.com" />
+	<target host="don.l214.com" />
+	<target host="www.don.l214.com" />
+	<target host="paris.l214.com" />
+	<target host="pour.l214.com" />
+	<target host="visuel.l214.com" />
+	<target host="visuels.l214.com" />
+
+	<rule from="^http://www\.don\.l214\.com/" 
+			to="https://www.l214.com/" />
```
J0WI: I don't get the same content

src/chrome/content/rules/l214.com.xml
```diff
@@ -0,0 +1,41 @@
+<!--
+	Mismatched:
+		l214.com
```
J0WI: I think it's refused

Bisaloo:
Done. They must be updating their website as I don't remember it defaulting to https.

PR #8163, opened 2017-01-08
Create t66y.com.xml

stlwvikx:


src/chrome/content/rules/t66y.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="t66y.com">
+	<target host="t66y.com" />
+	<target host="www.t66y.com" />
+
```
gloomy-ghost: Please add a `securecookie` tag here.

stlwvikx: It's important?

gloomy-ghost: It prevents session cookie leaking from http hijack, can be important for users who have an account and never log out.

gloomy-ghost:
Merged, thanks.

PR #8168, opened 2017-01-09
Update Debian related domains

pabs3:


strugee:
@pabs3 wouldn't it be way more effective to just bug DSA to implement redirects?

pabs3:
I am a member of DSA and we already have redirects for almost every
single debian.org subdomain.

DSA don't control debian.net subdomains (those are for Debian members)
so we can't add redirects there.

The redirects don't help for people who visit subdomains for the first
time with http, which is the main point of https-everywhere IIRC.

strugee:
@pabs3 ah, didn't scroll down far enough to see the debian.net-related changes. That makes sense. debconf.org needs a redirect though - that's why I mentioned them.

The first-visit problem that you mention, however, would be better solved by implementing [HTTP Strict Transport Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) (seems debian.org at least does this) and submitting to the [preload list](https://hstspreload.org/).

Cheers!

pabs3:
I didn't see a mention of debconf.org, DSA don't run debconf.org either
 at this time, except for a few of the static subdomains. The patch has
been waiting so long to be applied that more of debconf.org became TLS
enabled since I wrote it. I'll try to find some time to add more but
I'd rather this patch be merged first.

All the DSA-run servers should use HSTS, if you find any that don't
then that is a bug that we need to fix.

Unfortunately, AFAICT, we are not eligible for the preload list because
we have some subdomains that don't have TLS on them and never will, eg:

http://nossl.people.debian.org/~pabs/

Really though, https-everywhere and hsts preload is a non-scalable
solution, browsers need to be looking at something like DNSSEC+TLSA,
which DSA also does for all our TLS enabled domains.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

strugee:
@pabs3 oh, I thought you meant "debconf.org generally" instead of the literal apex domain `debconf.org`. Sorry about that; my mistake.

> All the DSA-run servers should use HSTS, if you find any that don't then that is a bug that we need to fix.

Good to know, thanks. And thanks for clarifying the preload issue - I'd just assumed everything was on TLS by now.

> Really though, https-everywhere and hsts preload is a non-scalable solution, browsers need to be looking at something like DNSSEC+TLSA, which DSA also does for all our TLS enabled domains.

Agreed that they're not scalable, but don't hold your breath waiting for browsers to implement DANE support. In fact Chrome used to validate TLSA records and dropped support because it turns out the underlying trust model is arguably even worse than the CA system. If you have time [this section][1] from a talk by Moxie Marlinspike does a pretty good job explaining the problems.

 [1]: https://youtu.be/xIiklPyS8MU?t=29m30s

src/chrome/content/rules/DebConf.org.xml
```diff
@@ -14,17 +14,45 @@
...
+	<target host="10years.debconf.org" />
+	<target host="es.debconf.org" />
+	<target host="fr.debconf.org" />
+	<target host="miniconf10.debconf.org" />
+
+	<rule from="^http:"
+		to="https:" />
+
+</ruleset>
+
```
J0WI: Please don't put more that one `<ruleset>` in one file.

pabs3:
I've updated the lists of domains to include everything in DSA's list
of LE-signed domains and also a bunch of DebConf domains too.

https://anonscm.debian.org/cgit/mirror/letsencrypt-domains.git/tree/domains

I've split the self-signed DebConf ruleset into a separate file.

J0WI:
Please have a look at the failing tests. Some redirects are throwing errors.

src/chrome/content/rules/Debian-Administration.org.xml
```diff
@@ -1,15 +1,8 @@
-<!--
```
J0WI: If this host is related, please add a hint to and in Debian.xml

pabs3:
debian-administration.org is not run by DSA or Debian but is related to
Debian in  that it is run by Debian users who are sysadmins and use
Debian and some of them are Debian contributors.

Thanks for pointing out the failing test, github really doesn't make
that obvious, especially when you don't enable JavaScript.

ports-master.d.o only uses TLS for rsync backend stuff, not http, removed.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

pabs3:
I poked the debian-administration.org folks about their expired cert.

J0WI:
Okay, Thank you!

PR #8173, opened 2017-01-10
Create CompilerDev.org.xml

matt11235:
The following domains should be upgraded to HTTPS:

* [compilerdev.org](https://compilerdev.org)
* [www.compilerdev.org](https://www.compilerdev.org)
* [forum.compilerdev.org](https://forum.compilerdev.org)
* [wiki.compilerdev.org](https://wiki.compilerdev.org)

Thanks.


J0WI:
The cert is now only matching `mattcode.me`

PR #8193, opened 2017-01-13
Contrib help

Hainish:
cc @jeremyn 
Still a work in progress

Issue #7717, opened 2016-11-30
When downgrade rules are appropriate


ghost:
Continuing discussion in #7646, when is it appropriate to use downgrade rules. Downgrade rules should be whitelisted, showing that they should be avoided if possible, however it's not clear when they are allowed.

For instance, are they allowed to:

1. Avoid 404 error?
2. Avoid invalid certificate?
3. Avoid redirect to a wrong page?
4. Avoid broken page UX if mixed content blocking is enabled?
5. Avoid broken page functionality if mixed content blocking is enabled?
6. Avoid broken page UX even if mixed content blocking is disabled?
7. Avoid broken page functionality even if mixed content blocking is disabled?

jeremyn:
Pinging @fuglede @gloomy-ghost @Hainish @J0WI .

jeremyn:
I want to note here that @smw-koops was nice enough to remove their downgrade from pull request #7646 so we could move forward with it while this discussion is happening.

J0WI:
From a technical aspect they are appropriated whenever a (protocol) relative link does not behavior the same on https than on http.
I don't know if there are any scenarios with mixed content, I think most will hit point 2.

J0WI:
I think we can get rid of downgrade rules anyway, see #8833.

jeremyn:
Getting rid of all of them is the best solution if possible.

Hainish:
Thanks @J0WI @jeremyn. Can I close this?

jeremyn:
I think this issue can be closed. Maybe another should be opened by someone, sometime, to remove the technical ability to create downgrades at all, and remove any related documentation too.

PR #8195, opened 2017-01-13
Initial commit of Pleasuredome.xml

kerframil:
This ruleset covers the Pleasuredome Tracker and its associated forums.

### DIFF
--- a/src/chrome/content/rules/Pleasuredome.xml
+++ b/src/chrome/content/rules/Pleasuredome.xml
@@ -1,4 +1,4 @@
-<ruleset name="Pleasuredome">
+<ruleset name="Pleasuredome" default_off="self-signed">
 	<target host="pleasuredome.org.uk" />
 	<target host="www.pleasuredome.org.uk" />
 	<target host="forum.pleasuredome.org.uk" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Pleasuredome.xml
```diff
@@ -0,0 +1,16 @@
...
+	<target host="www.pleasuredome.org.uk" />
+	<target host="forum.pleasuredome.org.uk" />
+
+	<test url="http://pleasuredome.org.uk/" />
+	<test url="http://www.pleasuredome.org.uk/" />
+	<test url="http://forums.pleasuredome.org.uk/" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^https?://pleasuredome\."
```
J0WI: We don't redirect https domains.

src/chrome/content/rules/Pleasuredome.xml
```diff
@@ -0,0 +1,16 @@
...
+	      to="https://www.pleasuredome." />
```
J0WI: This is not a valid TLD

kerframil:
Sorry, I didn't realise that self-signed certs are disallowed. Several parties have tried - and failed - to convince the webmaster to make use of a properly signed certificate. Therefore, this ruleset will remain inadmissible and I am closing the request.

J0WI:
They are not disallowed, but then the rule should be `default_off`.

kerframil:
Incidentally, it is neither mentioned [here](https://www.eff.org/https-everywhere/rulesets) nor [here](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) that redirects from https URLs are a QA violation. Further, some existing rulesets exploit the fact that it can be done.

```
$ cat *.xml | perl -ne 'print if /rule from="\^?https/ and not /<!--\s*<?rule/' | wc -l
98
```

Why does HTTPS-Everywhere honour such rules (downgrade attribute notwithstanding) if it is not considered to be an acceptable practice?

J0WI:
The documentation will soon be improved: https://github.com/EFForg/https-everywhere/pull/8193/files#diff-6a3371457528722a734f3c51d9238c13R230

We hope to get rid of the existing cases, but sadly some servers are really bad configured (see also discussion started in #7717).

PR #8198, opened 2017-01-14
Update Upaiyun.com.xml

gloomy-ghost:


jeremyn:
- [ ] Move the `www.upaiyan.com` rewrite `rule` down to directly above the general `rule` and with the same indentation. Use only one line for each `rule`. Keep the `target` where it is.
- [ ] Use `Secure connection failed` instead of `PROTOCOL_ERROR`

Also, it looks like the returned pages from www.upaiyun.com and www.upyun.com have the same content, but you might add a comment saying how these domains are related: are they the same company, etc. For example it might not be clear to a non-English speaker why `getdropbox.com` redirects to `dropbox.com` even though the domains have similar letters.

jeremyn:
Why did you add the `target` for `upyun.com`? Shouldn't that be a different ruleset?

Also the comment `<!-- Same as www.upaiyun.com but has a valid certificate -->` is a little vague. Are these the same company?

gloomy-ghost:
It's splitted now. Two domains are operated by the same company, and have the same DNS record, share the same ICP license number.

jeremyn:
I've reviewed the rulesets and updated checklist https://github.com/EFForg/https-everywhere/pull/8198#issuecomment-278781206.

jeremyn:
I've pushed changes to complete the checklist in https://github.com/EFForg/https-everywhere/pull/8198#issuecomment-278781206. I also noted `^upaiyun.com` as `Invalid certificate` in `Upaiyun.com.xml`.

@J0WI Can you review this please?

PR #8208, opened 2017-01-15
Add Tribecastudios.co.uk.xml

wjt:
Just a local gym. I didn't notice any mixed-content warnings etc. when signing into my account, booking a session, and generally poking around the site.

### DIFF
--- a/src/chrome/content/rules/Tribecastudios.co.uk.xml
+++ b/src/chrome/content/rules/Tribecastudios.co.uk.xml
@@ -1,0 +1,20 @@
+<!--
+	Nonfunctional hosts in *.tribecastudios.co.uk:
+
+		blog.tribecastudios.co.uk (r)
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Tribecastudios.co.uk">
+	<target host="tribecastudios.co.uk" />
+	<target host="www.tribecastudios.co.uk" />
+
+	<!-- accessing the bare domain with https: redirects to http://www.zingfit.com/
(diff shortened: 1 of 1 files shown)


J0WI:
Please also add `blog.tribecastudios.co.uk`

wjt:
Chrome, Firefox, Epiphany and HTTPie all report an invalid response (not even a certificate error) when accessing https://blog.tribecastudios.co.uk/ , so I don't think that's a good idea.

J0WI:
We usually try list each host, either as a target or otherwise as `Nonfunctional hosts in *.tribecastudios.co.uk:`. This will make maintenance easier for us and sysadmins can see the issue in the HTTPS Everywhere atlas.

wjt:
Ah, I understand now, thanks. I'll take a look.

wjt:
How's that? I've also contacted the website to suggest they force HTTPS anyway.

src/chrome/content/rules/Tribecastudios.co.uk.xml
```diff
@@ -0,0 +1,20 @@
...
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Tribecastudios.co.uk">
+	<target host="tribecastudios.co.uk" />
+	<target host="www.tribecastudios.co.uk" />
+
+	<!-- accessing the bare domain with https: redirects to http://www.zingfit.com/
+	     so explicitly add www.
```
J0WI: Please see https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects

J0WI:
That's great, thanks!

PR #8209, opened 2017-01-16
Remove redundant personal site rule (HSTS Preloaded)

ajhaydock:
I operate both sites so if there are any queries with the rules please let me know.

Thanks :)

### DIFF
--- a/src/chrome/content/rules/Alexhaydock.co.uk.xml
+++ b/src/chrome/content/rules/Alexhaydock.co.uk.xml
@@ -1,6 +1,0 @@
-<ruleset name="Alexhaydock.co.uk">
-  <target host="alexhaydock.co.uk" />
-  <target host="www.alexhaydock.co.uk" />
-
-  <rule from="^http:" to="https:" />
-</ruleset>


strugee:
@ajhaydock I'm not affiliated with HTTPS Everywhere and this probably won't prevent this PR from being merged, but you may want to consider splitting this commit into two (see [atomic commits][1]) and then making both commits have [good commit messages][2] that are under 50 characters :)

Cheers!

 [1]: https://www.freshconsulting.com/atomic-commits/
 [2]: http://chris.beams.io/posts/git-commit/

ajhaydock:
@strugee Thanks for the feedback. :) Not sure if I can go back and split the commit up after attempting this pull request, but I'll definitely bear it in mind for future ones. I have a few more sites to commit later.

strugee:
@ajhaydock if you split it up and force-push to GitHub, the new commits will show up here. That being said I don't know offhand how easy it is to _split_ commits (reordering and squashing aren't _super_difficult, but I don't know about _splitting_). :)

J0WI:
Since both hosts are already HSTS preloaded, we can safely remove them from HTTPS Everywhere.

ajhaydock:
That is true. Hadn't considered that. Would you like me to rework this commit to remove my rule instead?

J0WI:
That would be great!

ajhaydock:
That ought to do it :smile:

J0WI:
Thanks!

PR #7358, opened 2016-10-16
created rule for dji.com

schmiedecker:

PR #8217, opened 2017-01-18
Schmiedecker master

Hainish:


jeremyn:
Thanks @Hainish , but why did you create a new pull request instead of using #7358 ?

Hainish:
@jeremyn I could have merged and then committed, but I was already in the review process in my local repo when I made this additional commit so I found it easier to create a new PR

PR #7746, opened 2016-12-03
Update and rename GHacks.net.xml to ghacks.net.xml

ekaris:

PR #8221, opened 2017-01-19
Update GHacks.net.xml

redranamber:


gloomy-ghost:
Thanks.

- [ ] Every target has an implicit test, so please remove the unnecessary `test`s.
- [ ] cdn.ghacks.net has some mixed content blocking, maybe we can just redirect it to www.ghacks.net?

redranamber:
It seems you tried browsing to the cdn subdomain. It's not something that people would normally do, but the other pages on the site all load content from cdn.ghacks.net. The rest of the pages I checked didn't have any issues loading it over https with this rule on so it would be better to leave it in.

jeremyn:
FYI this is a duplicate of PR https://github.com/EFForg/https-everywhere/pull/7746, so be sure to close one of these PRs after the other one is merged.

jeremyn:
I agree with @gloomy-ghost that we have to consider the behavior at http://cdn.ghacks.net even though users would not normally go there directly. Note that some of these broken `cdn` URLs show up at https://www.google.com/#q=site:cdn.ghacks.net, so they are not entirely hidden.

Looking at the Dev Tools > Network tab for http://www.ghacks.net, it looks like the important content from the `cdn` subdomain comes from `http://cdn.ghacks.net/wp-content/`. So I recommend writing the ruleset to only cover `http://cdn.ghacks.net/wp-content/*`, not all of `http://cdn.ghacks.net`.

gloomy-ghost:
Oops, missed the notification. I have some rulesets of CDN domains that are even more radical than this one, like #7281 and #7484. However they are just pending forever because the trade-off is hard to be made.

Some expanding thoughts: Many CDNs offer lazy caching which only loads and caches resources from a origin when they actually received requests. It causes as many MCBs as the origin has, but frankly I don't except a CDN domain serving pages instead of resources, and securing the whole CDN sounds really fascinating.

redranamber:
How do I get it to only rewrite /wp-content/ and /wp-includes/ I'm not getting it right.

jeremyn:
@redranamber `target`s can't be complicated like that, they have to be just domains. You want to work with the `rule`s. Take a look at https://github.com/EFForg/https-everywhere/pull/7957#issuecomment-278825904 for an approach I recently used in a similar situation.

redranamber:
ERROR src/chrome/content/rules/GHacks.net.xml: No rule or exclusion applies to test URL http://cdn.ghacks.net/

What happened there?

jeremyn:
Each `target` creates an implicit `test`, so you can imagine that your `cdn` `target` automatically creates a `test` like `<test url="http://cdn.ghacks.net/" />`. However `test`s are supposed to go with a `rule` or an `exclusion`, but you don't have a `rule` or `exclusion` that applies to this `test`.

The workaround is to add an `exclusion` specifically for this test. This is what I did in the `showapi.com` example with:

```xml
		<!-- This exception disables the implicit test created by the target. -->
		<exclusion pattern="^http://bbs\.showapi\.com/$" />
```

redranamber:
This is done in case you missed it.

gloomy-ghost:
Oh, sorry for the delay.

Please delete the tests for www.ghacks.net and deals.ghacks.net, add appropriate indent for exclusion/rule and newlines between target/securecookie/rule according to the style guide. Then we are ready to go :)

redranamber:
Is this right now? The style guide isn't clear on tabs or spacing.

gloomy-ghost:
> Use tabs and double quotes (", not ').

Maybe it's just not prominent enough I guess.

Can you move the exclusion right under the target it applies to, and also the tests (with properly indent)?

redranamber:
It would probably help to add a more complex rule as an example, that one doesn't even have the securecookie.

gloomy-ghost:
Merged, thanks!

Issue #3798, opened 2016-01-07
[Idea] Tor Onion services


frafra:
Some websites are providing access as a Tor Onion service, like:
- DuckDuckGo - http://3g2upl4pq6kufc4m.onion/
- Facebook - https://facebookcorewwwi.onion/
- Tor Project - http://bt3ehg7prnlm6tyv.onion/
- Wikileaks - http://gjlng65kwikileax.onion/

Roger Dingledine (original developer of Tor along with Nick Mathewson and Paul Syverson) says something like "Tor Onion services vs HTTPS only websites is like HTTPS vs HTTP few years ago".
-- https://media.ccc.de/v/32c3-7322-tor_onion_services_more_useful_than_you_think#video&t=1041

I think it would be cool to have an option in HTTPS Everywere which instead of replacing http:// with https:// could replace the domain with the equivalent .onion domain.

I submitted a new issue on https-everywhere because I know this program is a collaboration between EFF and The Tor Project and the aim is to making browsing more secure, so I think there could be an interest for this kind of feature.

> HTTPS Everywhere is a Firefox, Chrome, and Opera extension that encrypts your communications with many major websites, making your browsing more secure. [...]
> HTTPS Everywhere is produced as a collaboration between The Tor Project and the Electronic Frontier Foundation.
> -- https://www.eff.org/https-everywhere

Would it makes sense to you? Or do you think it would be better to develop a different project, maybe using the same codebase of https-everywhere?

numismatika:
If you have it as a separate project and the user runs both addons i see a conflict of rewrites,
Having it in HTTPS-Everywhere needs a way to figure out if TOR is active, since otherwise the redirects are useless. Maybe implement it as an extra platform-tag?

frafra:
I would prefer to integrate such feature in HTTPS Everywhere if possible.

`platform="tor"` looks goods:
> Note: This should be something like platform=tor, to indicate that these rules should not go through the fetch tester on the non-Tor Internet because they will fail. As an approximation to that, we just use platform="mixedcontent", since Tor Browser is a mixedcontent platform, and the https-everywhere-checker will skip non-default platforms.
> - [src/chrome/content/rules/FacebookCoreWWWi.onion.xml](https://github.com/EFForg/https-everywhere/blob/61cf352307131540f6b2ab3e45ed2164e00e981a/src/chrome/content/rules/FacebookCoreWWWi.onion.xml#L4)
[...]

PR #8223, opened 2017-01-19
[ProtonMail Hidden Service] New ruleset

lucaspetter:
ProtonMail just launched a Tor hidden service that requires HTTPS.

https://protonmail.com/blog/tor-encrypted-email/
https://techcrunch.com/2017/01/19/protonmail-adds-tor-onion-site-to-fight-risk-of-state-censorship/
http://www.theregister.co.uk/2017/01/19/protonmail_launches_tor_hidden_service/

### DIFF
--- a/src/chrome/content/rules/ProtonMail.com.xml
+++ b/src/chrome/content/rules/ProtonMail.com.xml
@@ -1,6 +1,6 @@
 <!--
 	Other ProtonMail rulesets:
-		- ProtoniRockerXow.onion.xml
+		- Protonirockerxow.onion.xml
 
 	Server sends includeSubDomains in STS header.
 
--- a/src/chrome/content/rules/ProtonMail.com.xml
+++ b/src/chrome/content/rules/ProtonMail.com.xml
@@ -1,4 +1,6 @@
 <!--
+	Other ProtonMail rulesets:
+		- ProtoniRockerXow.onion.xml
 
 	Server sends includeSubDomains in STS header.
 
--- a/src/chrome/content/rules/ProtoniRockerXow.onion.xml
+++ b/src/chrome/content/rules/ProtoniRockerXow.onion.xml
@@ -1,0 +1,25 @@
+<!--
+	Other ProtonMail rulesets:
+		- ProtonMail.com.xml
+
+	This ruleset covers ProtonMail hidden services on Tor.
+	Note: This should be something like platform=tor, to indicate
+	that these rules should not go through the fetch tester on the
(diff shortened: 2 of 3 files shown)


lucaspetter:
The Travis build fails due to a fetch error. This ruleset uses `platform="mixedcontent"` in the same way as other hidden service rulesets (eg. [Facebook](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/FacebookCoreWWWi.onion.xml), [Intercept](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Y6xjgkgwj47us5ca.onion.xml)), so isn't the test supposed to not fetch the URL? I don't understand this.

```
INFO Finished comparing http://protonirockerxow.onion/ -> https://protonirockerxow.onion/. Rulefile: src/chrome/content/rules/ProtoniRockerXow.onion.xml.
ERROR src/chrome/content/rules/ProtoniRockerXow.onion.xml: Fetch error: http://protonirockerxow.onion/ => https://protonirockerxow.onion/: (6, 'Could not resolve host: protonirockerxow.onion')
```

jeremyn:
@lucaspetter HTTPS Everywhere tests are known to fail against .onion addresses. See the discussion in issue #3798. The Facebook and Intercept rulesets you link each fail with `fetch-test.sh` locally and I expect would fail with Travis. I assume they were added as special cases, but I'm not sure.

@Hainish Do you think we should add ProtonMail's new .onion URL as a special case? I think we should.

lucaspetter:
@jeremyn Thanks, now I understand.

Hainish:
@jeremyn Since we've already made exceptions for the Intercept and Facebook, I think it's fine to add Protonmail to the list.

lucaspetter:
@jeremyn @Hainish Where does this exception need to get added?

Hainish:
@lucaspetter I think the way you've added it is correct, but I'll leave it up to the digression of the ruleset maintainers.

jeremyn:
@lucaspetter Please make these small changes:

- [ ] Note `*.protonirockerxow.onion` as `Invalid certificate` at the top
- [ ] Change the filename to `Protonirockerxow.onion.xml` and also update the comment in `ProtonMail.com.xml`
- [ ] Remove the `Direct rewrites` comment, we don't need it

lucaspetter:
@jeremyn Okay, I've made the changes.

jeremyn:
Thanks. This looks fine, other than the fetch error, which we have decided to ignore. Merged.

lucaspetter:
Thanks

PR #8127, opened 2017-01-07
Adding the hsts-prune utility

Hainish:

PR #8224, opened 2017-01-19
status.cubecraft.net fix.

marcoslater:
Remove special case for status.cubecraft.net as its SSL cert is now valid.

Also should be HSTS preloaded in some web browsers by now, and could potentially be removed from HTTPS Everywhere. Up to you guys. - cc @jeremyn


J0WI:
We have now a script that removes all preloaded hosts (see #8127), so if this host fits our needs it will be removed.

PR #8128, opened 2017-01-07
Removing rulesets which are preloaded in the major browsers

Hainish:

PR #5109, opened 2016-06-11
Update Python.org.xml

davidabian:

PR #8226, opened 2017-01-20
Adding additional test to ensure that added targets are not HSTS prel…

Hainish:
…oaded

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ env:
   - TEST=chromium
   - TEST=rules
   - TEST=fetch
+  - TEST=preloaded
--- a/test/travis.sh
+++ b/test/travis.sh
@@ -66,6 +66,12 @@ if [ "$RULESETS_CHANGED" ]; then
     # --privileged is required here for miredo to create a network tunnel
     docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" --privileged httpse bash -c "service miredo start && test/fetch.sh"
   fi
+
+  if [ "$TEST" == "preloaded" ]; then
+    echo >&2 "Ensuring rulesets do not introduce targets which are already HSTS preloaded."
+    docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" node bash -c "cd /opt/utils/hsts-prune && npm install && node index.js"
+    [ `git diff --name-only | wc -l` -eq 0 ]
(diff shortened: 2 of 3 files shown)


Hainish:
Tests failing, fixed by https://github.com/EFForg/https-everywhere/pull/8241

Hainish:
Ready for review @cooperq

cooperq:
lgtm

jeremyn:
@Hainish `hsts-prune` in Travis is causing a problem in https://travis-ci.org/EFForg/https-everywhere/jobs/195715999.

Hainish:
@jeremyn It's not a bug, it's a feature :] See https://github.com/EFForg/https-everywhere/pull/5109#issuecomment-275574332

Hainish:
One thing that would make it easier to see which targets to remove is if `hsts-prune` actually listed the targets it would remove, so you don't have to run it yourself and can just report that back to the user.  I'll work on that.

jeremyn:
@Hainish The current status as I understand it is that `hsts-prune` makes Travis fail but it doesn't say on what domains. With a ruleset like in #5109 with a lot of domains, and with HSTS preloading requiring a bunch of different checks, it's not reasonable to have to check every domain manually because of an `hsts-prune` failure. Also it looks like this failure isn't tied into `fetch-test.sh` or `test.sh`. `hsts-prune` is only documented in a secondary [README.md](https://github.com/EFForg/https-everywhere/blob/a95453f4f9eb5a32dc0719f71f6106a8503f31dc/utils/hsts-prune/README.md) under `utils/hsts-prune/`, with no description of output, and the short instructions say to run `npm` and `node` which I assume many contributors will not have immediately available or know how to install or use like bash or python.

If this is an accurate description of the situation, then I feel pretty strongly that it's premature to move this into Travis and it should be removed quickly from the automatic tests. Until it's well-integrated into the current development dependency installation (`install-dev-dependencies.sh`) and local testing workflow (`fetch-test.sh`), we should just run it manually every now and then and do bulk removals like in https://github.com/EFForg/https-everywhere/pull/8128.

Hainish:
@jeremyn ^ was true until moments ago with https://github.com/EFForg/https-everywhere/commit/07b8e33f9e3fcebb5b43fa9feda3e1f0a5d4b3da

The output of the travis test now shows exactly which domains to remove from the PR.

Hainish:
As for the documentation, I'll be adding that to https://github.com/EFForg/https-everywhere/pull/8193

Issue #8251, opened 2017-01-26
Exclusion rule for squarespace.com subdomain sg-links


dtauerbach:
Hello, I work for Squarespace (and was a former EFF technologist). We are getting some complaints about the subdomain sg-links.squarespace.com which is hosted by sendgrid and not us and it's non-trivial to get a cert for it on their side. The upshot is poor user experience for HTTPS Everywhere users clicking certain links from email, like, say, unsubscribe. I am going to write a rule to exclude this subdomain from the Squarespace partial rule, & send a PR referencing this issue; if there is another preferred workflow beyond the repo guidelines that I missed please let me know.

jeremyn:
I'm closing this issue since the associated PR is merged.

PR #8252, opened 2017-01-26
[#8251] Exclusion rule for sg-links.squarespace.com

dtauerbach:
See https://github.com/EFForg/https-everywhere/issues/8251.

I disabled the chrome selenium test since selenium wasn't playing nicely on my fedora dev environment, but I tested the exclusion rule locally by building a crx, in addition to running the unit tests.


Hainish:
Hey @dtauerbach! \o

Looks like this isn't matching the indentation style of the rest of the ruleset.  Could you correct that?  Also note we now have an `.editorconfig` file which specifies the proper indentation style.

Otherwise lgtm

PR #8256, opened 2017-01-27
Update Hetzner-Online.xml

chkpnt:


J0WI:
Our current style guide recommends to avoid wildcards.
Also, this breaks mail.hetzner.de

Can you list single targets for hetzner.de instead?
Please also add a note, that your-server.de is used for customers and therefore contains various subdomains without a valid cert.

J0WI:
ping @chkpnt

jeremyn:
I'm closing this due to no response from @chkpnt .

PR #8261, opened 2017-01-28
[ACLU of Arizona] New ruleset

galenide:
Note: There is another site, [Changing MCSO](http://www.changingmcso.org/), that is described as "An ACLU of Arizona site" but which does not support HTTPS. As of now, this site is not mentioned or included in the ruleset.

### DIFF
--- a/src/chrome/content/rules/ACLU-of-Arizona.xml
+++ b/src/chrome/content/rules/ACLU-of-Arizona.xml
@@ -2,6 +2,9 @@
 
 	For other ACLU coverage, see ACLU.xml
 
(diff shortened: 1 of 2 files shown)


jeremyn:
Thanks. Please:

- [ ] Note `owa.acluaz.org` in the top comment as `Different HTTP/HTTPS content`. Ordinarily I might not include an internal email site, but the `owa` site uses cookies with domain `.acluaz.org`, so this comment justifies why the `securecookie` `host` is restricted to `www`.

jeremyn:
Thanks, merged.

PR #8269, opened 2017-01-29
[Octopuspop.com] Add ruleset

fuglede:
This is a follow-up to https://github.com/EFForg/https-everywhere/pull/6102 in which various parts of the site was covered, but in which some of them still had some issues. In the same pull request, the developer later notified us that those issues have now been fixed.


J0WI:
Thanks!

PR #8270, opened 2017-01-29
Add new york zoos

illicitonion:


src/chrome/content/rules/Bronxzoo.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Bronxzoo.com">
+	<target host="bronxzoo.com" />
```
J0WI: Please see also:
www.bronxzoo.com
app.bronxzoo.com
apps.bronxzoo.com
pages.bronxzoo.com

src/chrome/content/rules/Centralparkzoo.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Centralparkzoo.com">
+	<target host="centralparkzoo.com" />
```
J0WI: Please see also:
www.centralparkzoo.com
chicks.centralparkzoo.com
pages.centralparkzoo.com

src/chrome/content/rules/Prospectparkzoo.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Prospectparkzoo.com">
+	<target host="prospectparkzoo.com" />
```
J0WI: Please see also www.prospectparkzoo.com

src/chrome/content/rules/Queenszoo.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Queenszoo.com">
+	<target host="queenszoo.com" />
```
J0WI: Please see also www.queenszoo.com

illicitonion:
Added where appropriate.

Skipping pages.bronxzoo.com because the http version redirects to bronxzoo.com and the https version has an invalid certificate.

Skipping apps.bronxzoo.com, chicks.centralparkzoo.com, pages.centralparkzoo.com because both the http and https versions are redirects to bronxzoo.com (or for pages.centralparkzoo.com, centralparkzoo.com). If they start being used for something, who knows whether they will be validly https'd, but right now there's no reason for people to be visiting them.

It feels like adding comments saying "Here are some subdomains for which DNS technically exists, but which don't function at all" seems weird and redundant.

J0WI:
@illicitonion we use the documentation for future maintenance on the rules and for the HTTPS Everywhere Atlas.

J0WI:
Thanks and sorry for the delay!

PR #8271, opened 2017-01-30
Create home.cern.xml

Folant:


jeremyn:
Thanks. Can you please just add these changes to [`CERN.xml`](https://github.com/Folant/https-everywhere/blob/b65fdf38f1c70e41fca68889e5d1f334eaee2b8f/src/chrome/content/rules/CERN.xml) instead?

PR #8272, opened 2017-01-30
Update Forbes.com.xml

Folant:


src/chrome/content/rules/Forbes.com.xml
```diff
@@ -1,80 +1,21 @@
- <!--
-	For problematic rules, see Forbes_Magazine-mixed.xml.
-
-
-	Other Forbes rulesets:
-
-		- Forbes_img.com.xml
-		- Forbes_Magazine.xml
```
J0WI: Please keep this reference

src/chrome/content/rules/Forbes.com.xml
```diff
@@ -1,80 +1,21 @@
- <!--
-	For problematic rules, see Forbes_Magazine-mixed.xml.
```
J0WI: You can remove this file since its content is obsolete.

src/chrome/content/rules/Forbes.com.xml
```diff
@@ -1,18 +1,25 @@
-<!--
-Invalid certificate: 
+<!---
+Other Forbes rulesets:
+	- Forbes_img.com.xml
+	- Forbes_Magazine.xml
```
J0WI: `Forbes_img.com.xml` does not exist, but `Forbes_Magazine.xml` can be kept.

PR #8273, opened 2017-01-30
Add DNC.org.xml and Democrats.org.xml

jeremyn:


src/chrome/content/rules/DNC.org.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	See also Democrats.org.xml.
+
```
J0WI: See also:
hrb.dnc.org
gems.dnc.org
sendgrid.dnc.org
tanto.dnc.org

src/chrome/content/rules/Democrats.org.xml
```diff
@@ -0,0 +1,52 @@
...
+		- centrodevotacion.democrats.org
+		- convention.democrats.org
+		- ppapi.democrats.org
+
+	Time out:
+		- asdc.democrats.org
+		- developers.democrats.org
+		- directory.democrats.org
+		- factivists.democrats.org
+
```
J0WI: See also:
developers.democrats.org
dtmc.democrats.org
future.democrats.org
live.democrats.org

jeremyn: https://developers.democrats.org is already in the file so I'm not sure what you want me to do with that subdomain.

src/chrome/content/rules/DNC.org.xml
jeremyn: Both https://gems.dnc.org and the HTTP version time out for me, so I don't think we should include or even mention it.

PR #8278, opened 2017-01-31
Add Depressionforums.org.xml

jeremyn:


jeremyn:
There's a lot of mixed content blocking for this site so I've added `platform="mixedcontent"`. The mixed content blocking looks harmless, but most of the site requires a login which I don't have, so I can't be sure.

I've contacted the depressionforums.org staff at https://www.depressionforums.org/forums/contact/ to see if they can help us evaluate this ruleset, ideally so we can remove the `platform` attribute and enable the ruleset by default for everyone.

jeremyn:
I haven't heard back from the depressionforums.org staff so I'm fine with leaving this as `platform="mixedcontent"`. Thanks.

PR #8281, opened 2017-01-31
Create dxy.com.xml

ivysrono:


ivysrono:
@J0WI This is the best Chinese medical science website, and works fine for android.
Review it please.

src/chrome/content/rules/dxy.com.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+	Invalid cert:
+		www.dxy.com
+		i.dxy.com
+		yao.dxy.com
```
J0WI: bbs.dxy.com
hz.dxy.com

PR #8287, opened 2017-01-31
Add IslamReligion.com.xml

jeremyn:


gloomy-ghost:
https://ms.mailserver.islamreligion.com, it's an internal mail system though.

jeremyn:
I don't want to add an internal mailserver URL like that.

PR #8289, opened 2017-01-31
Add ING.org.xml

jeremyn:


src/chrome/content/rules/ING.org.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Invalid certificate:
+	- dipl.ing.org
```
gloomy-ghost: Here is a wildcard DNS record... Every subdomains that don't actually exist are redirected to https://ing.org/wp-signup.php?new=$DOMAIN

jeremyn: You're right, good catch. Removed.

PR #8302, opened 2017-01-31
Add IslamWeb.net.xml

jeremyn:


src/chrome/content/rules/IslamWeb.net.xml
```diff
@@ -0,0 +1,33 @@
...
+	- kids.islamweb.net
+	- kidsen.islamweb.net
+	- quran.islamweb.net
+	- tube.islamweb.net
+
+-->
+<ruleset name="IslamWeb.net">
+	<target host="islamweb.net" />
+	<target host="www.islamweb.net" />
+	<target host="articles.islamweb.net" />
```
Hainish: It looks like there's some mixed content on these.  The top banner isn't loading for the HTTPS endpoints.

Hainish: This is common for *all* the HTTPS subdomains under this domain.  I think it's probably worth the tradeoff, though.  Accepting changes.

PR #8305, opened 2017-02-01
Add IslamAcademy.net.xml

jeremyn:


src/chrome/content/rules/IslamAcademy.net.xml
```diff
@@ -0,0 +1,26 @@
...
+-->
+<ruleset name="IslamAcademy.net">
+	<target host="islamacademy.net" />
+	<target host="www.islamacademy.net" />
+	<target host="almajd.islamacademy.net" />
+	<target host="benaa.islamacademy.net" />
+	<target host="www.benaa.islamacademy.net" />
+	<target host="courses.islamacademy.net" />
+	<target host="forum.islamacademy.net" />
+	<target host="www.forum.islamacademy.net" />
```
Hainish: looks like this is not a valid cert...

Hainish: Oh, I see the rule rewrite now.  LGTM

PR #8313, opened 2017-02-01
mikbill.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/mikbill.ru.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+mikbill.ru ¹
```
J0WI: WFM

PR #8316, opened 2017-02-01
sets default_off for Sanger Institute

jrandall:
The Sanger institute web site is broken in various ways (the most obvious being that all icons are broken and represented by placeholder boxes) when using https everywhere.


src/chrome/content/rules/Wellcome-Trust-Sanger-Institute.xml
```diff
@@ -1,4 +1,4 @@
-<ruleset name="Wellcome Trust Sanger Institute (partial)">
+<ruleset name="Wellcome Trust Sanger Institute (partial)" default_off="broken">
```
J0WI: Please use "redirects to http" as reason.

jrandall:
@J0WI change made as requested - agree that one way to look at the reason it is broken is that the site keeps redirecting back to http.

Issue #8314, opened 2017-02-01
fix rule for ietf.org


AxelNennker:
https is not enforced on http://xml2rfc.tools.ietf.org/
Please either add xml2rfc or amend the tools.ietf.org rule

Bisaloo:
@J0WI, could you close this issue since you merged #8321 please?

PR #8321, opened 2017-02-01
Update IETF.xml

Bisaloo:


Bisaloo:
Notes to reviewer:

- `ns0.ietf.org` redirects to an outside domain. Should it be added in the rewrite rules?
- `ztp.meeting.ietf.org`, `etherpad.sql1.ietf.org` and `permatrac.sql1.ietf.org` return `401`

Bisaloo:
Fixes #8314

Bisaloo:
Ok, travis and I still disagree but I think the mistake is his in this case.

`curl -v https://noc.ietf.org/` indeed complains about invalid certificate but using firefox, the certificate is fine.

gloomy-ghost:
noc.ietf.org doesn't serve a full certificate chain, maybe you can reproduce it on Firefox by using a brand-new profile.

Bisaloo:
You're right! I removed `noc.ietf.org` and added a comment.

J0WI:
Thanks!

PR #8348, opened 2017-02-02
Add Ctrl blog rules

da2x:


J0WI:
This host has been submitted to the HSTS preload list, but is not yet included. So for now this LGTM, but we may remove the ruleset when the host is on that list.

PR #8361, opened 2017-02-02
Add ImamReza.net.xml

jeremyn:


Bisaloo:
I get MCB from `fonts.googleapis.com`. 

Could you add a comment to show that it's been checked and that MCB on this site does not hinder navigation?

jeremyn:
I disagree with adding a comment to say that some specific MCB is not a problem. If someone in the future wants to verify that it's not a problem, they can check it themselves, and in fact it may be a problem then.

Also in this specific case I don't understand the language of the site and can't really say what effect the missing font styling has on navigation. The `https` and `http` versions of the site seem similar enough, in the small number of pages I've checked.

Bisaloo:
Would you be okay with simply adding a comment saying there is active mixed content from `fonts.googleapis.com` then? Without judgment call about the effect on site navigation of appearance.

Right now, it looks like an oversight. If I were to update this ruleset, I would ask myself why we have a ruleset that trigger mixed content blocking.

jeremyn:
No, I disagree with adding a comment about it at all. If you were to update this ruleset in the future and saw an `fonts.googleapis.com` MCB error on the site, you should evaluate it at that time and either keep the domain or remove it based on your observations then. You should not rely on existing comments. There is no need to "ask yourself why we have a ruleset that trigger mixed content blocking", because the answer is always the same: the maintainer thought it worked well enough at the time.

If there is surprising or hard-to-discover behavior, then maybe a comment is justified. An example might be if a domain is failed for MCB, but it's not clear what the MCB breaks -- a forum in a subsection of the site, maybe. In that case the comment can act as a test case for future maintainers to check before passing the domain.

jeremyn:
Thanks.

PR #8368, opened 2017-02-02
Add TheIsmaili.org.xml

jeremyn:


src/chrome/content/rules/The.Ismaili.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	See also TheIsmaili.org.xml
+
+	No working URL known:
+		www.the.ismaili
```
Bisaloo: I get `Invalid certificate`

jeremyn: I've removed `www.the.ismaili` entirely, since it has an invalid certificate for `https` and a `404` for `http`.

PR #8369, opened 2017-02-02
Add Albalagh.net.xml

jeremyn:


src/chrome/content/rules/Albalagh.net.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Note that https://www.albalagh.net/bookstore is a subsite of the main site
+	so be sure to test changes at both "/" and "/bookstore".
```
J0WI: That's now `albalaghbooks.com` and I don't find any relation

jeremyn:
Updated.

PR #8370, opened 2017-02-02
Add Wahiduddin.net.xml

jeremyn:


gloomy-ghost:
http://mail.wahiduddin.net says "Please visit: http://wahiduddin.net", and https://mail.wahiduddin.net shows https://wahiduddin.net after ignore the mismatch error. Would it be abandoned? I think at least we can add a securecookie for wahiduddin.net.

jeremyn:
I'm not sure what you're asking for with `mail`, could you please be more specific? Also I don't want to add a `securecookie` for `wahiduddin.net` while there are certificate problems at https://mail.wahiduddin.net.

gloomy-ghost:
It seems like that `mail.` doesn't function properly anyway, so I think the certificate issue on `mail.` can be ignore while considering a `securecookie`.

jeremyn:
The site owner runs an [email newsletter](http://wahiduddin.net/saki/) and it's possible that the `mail` subdomain is important for that in some way we don't know about, so I don't think we should ignore `mail`.

gloomy-ghost:
alright, merging

PR #8375, opened 2017-02-03
Add QuranAcademy.io.xml

jeremyn:


src/chrome/content/rules/QuranAcademy.io.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Refused:
+	- www.quranacademy.io
```
J0WI: I got a mismatch

See also `courses` and `reader`.

PR #8376, opened 2017-02-03
Add Al-Shia.org.xml

jeremyn:


J0WI:
There are some new translations, could you please update the list of targets?

jeremyn:
Updated.

PR #8377, opened 2017-02-03
Add AllahsQuran.com.xml

jeremyn:


gloomy-ghost:
The donate banner at left is blocked, should I be concerned about this?

jeremyn:
On https://allahsquran.com without MCB (or on unencrypted http://allahsquran.com ) I see two donate ads, one to the left and a banner at the bottom. With MCB, the left one disappears but the bottom one stays. Both with and without MCB I see the social media links at the right. Also at https://allahsquran.com/read/index.php the MCB blocks a donate ad at the top. On all pages there is a "DONATE" text link in the footer. Even on plain HTTP pages, the URL is for https://donateforislam.com which doesn't work (`Unable to connect`), though the plain HTTP URL does work.

Does that match what you're seeing? In that case, while I don't like blocking donation links, there is still at least one donation link available on every page, and the links are broken anyway. So I think the ruleset is okay, but what do you think?

gloomy-ghost:
The social media links are block for me:
image

I'm ok with this but just want to be clear before merging.

jeremyn:
I checked again and the social links to the right are now blocked for me by MCB. I'm not sure why they worked before. Anyway I'm generally okay with blocking social links.

gloomy-ghost:
Alright, merged.

PR #8399, opened 2017-02-03
Update Normalesup.org.xml

Bisaloo:


J0WI:
Note that this host may not work in near future, because the host is using a certificate with the deprecated SHA1 hash.

PR #8415, opened 2017-02-04
Create monde-diplomatique.fr.xml

Bisaloo:


Bisaloo:
The domain `mondediplo.net` belongs to the same paper.

Should it be added here or in a new ruleset?

J0WI:
Thanks!
Please add a new file for `mondediplo.net` and use comments to show the relationship between them.

Issue #4768, opened 2016-05-13
Broken site: deezer.com not functioning with default settings


AlfonsoMuskedunder:
Apologies in advance if this is the wrong place to talk about broken sites; if not please inform me and i'll refrain from it once and for all.

The music streaming website deezer.com only works once I untick "deezer.com" in the list of Stable Rules. Is that how it is supposed to be? (I'm on Windows 10 and Chrome 50.0.2661.102 m.)

PS: Thanks to all dev's for this awesome extension!

Edit: This is the same in Firefox 46.0.1

J0WI:
Could you tell us more what exactly is broken when you enable the rule?
Do you have any errors in your browser console?

AlfonsoMuskedunder:
@J0WI Thank you for your reply. With the rule enabled, menus from which tracks are being selected simply won't open.

Here is a screenshot when the rule is disabled. I just clicked on "Alben":
![deezer_functioning](https://cloud.githubusercontent.com/assets/10466043/15294810/076b6488-1b8e-11e6-9e53-fa4c75fb923c.jpg)

With the rule enabled, the list of albums simply won't appear:
![deezer_broken_console](https://cloud.githubusercontent.com/assets/10466043/15294844/2fc48342-1b8e-11e6-86f4-dd40241b4853.jpg)

J0WI:
You need an account to reproduce this, right?

AlfonsoMuskedunder:
That's correct. Regular accounts are free, however.

gyphub:
In addition, the search on the site does not work if the rule for Deezer.com is enabled.
![screen_search_not_working](https://cloud.githubusercontent.com/assets/13639560/16484331/f93f72de-3eb8-11e6-8579-eaa559b8efa4.png)

If the rule is disabled, search works okay:
![screen_rule_disabled](https://cloud.githubusercontent.com/assets/13639560/16484546/04e3b2d4-3eba-11e6-9cc1-9dad4c7e38b3.png)

jwatt:
For me enabling HTTPS Everywhere somehow results in this error in the console:

`ReferenceError: reference to undefined property ma.https [Learn More]  vb.js:87:1679`

( http://connect.facebook.net/en_GB/sdk/vb.js )

When that exception is thrown it seems to interrupt the deezer scripts sufficiently badly that playing music on the site no longer works (clicking play buttons has no effect).

jw243:
[...]

PR #8418, opened 2017-02-06
Update Deezer.xml (#4768)

jwatt:
Disable ruleset since the navigation bar is broken.

### DIFF
--- a/src/chrome/content/rules/Deezer.xml
+++ b/src/chrome/content/rules/Deezer.xml
@@ -1,4 +1,24 @@
-<ruleset name="Deezer.com">
+<!--
+Disabled because:
+
+Currently Deezer's website *generally* redirects as follows:
+
+	https://deezer.com/      301 Moved Permanantly (to www.deezer.com)
+	https://www.deezer.com/  302 Found             (to plain-http URL)
+	http://www.deezer.com/   200 OK
+
+That is, if we apply an http -> httpS rule to [www.]deezer.com most paths will
+end up being loaded via plain-http anyway.  The one (known) exception is the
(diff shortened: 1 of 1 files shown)


jwatt:
Note that the comments added in the commit explain why HTTPS Everywhere doesn't work with deezer.com.

J0WI:
Thanks for debugging this!
What about static resources from `e-cdns-files.dzcdn.net`? Do we run in any problems if we only redirect them?

jwatt:
Sure thing!
I don't see any requests to `e-cdns-files.dzcdn.net` while navigating around deezer.com. I do see requests to `cdn-images.deezer.com` and `e-cdn-files.deezer.com` which I added commented out to Deezer.xml (the comment explains why).

J0WI:
I have such requests on `https://www.deezer.com/us/` without being logged in.

jwatt:
It seems when you're NOT logged in, `deezer.com` uses HTTP*S*. It's only after you're logged in that it redirects everything to plain-http. So I think all the sub-resources you're seeing load when loading `https://www.deezer.com/us/` will be made via HTTP*S*, including those from `e-cdns-files.dzcdn.net`. In fact it seems that `e-cdns-files.dzcdn.net` **only** serves content over HTTP*S*. For example, [this CSS file](https://e-cdns-files.dzcdn.net/cache/css/bootstrap-a66521b84dab22cf9d1ac7a7fc40cc10.css) will not load (404) if you change the URL scheme to plain-'http'.

J0WI:
So maybe we could rewrite some resources to this host but since this host behaves very bad on https I'm fine with disabling it.

PR #8429, opened 2017-02-07
bither.net

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/bither.net.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+www.bither.net ⁴
```
J0WI: WFM

galeksandrp: Invalid cert chain

PR #8431, opened 2017-02-07
breadwallet.com

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/breadwallet.com.xml
+++ b/src/chrome/content/rules/breadwallet.com.xml
@@ -1,10 +1,6 @@
 <ruleset name="breadwallet.com">
 	<target host="breadwallet.com" />
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/breadwallet.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="breadwallet.com">
+	<target host="breadwallet.com" />
+
+	<target host="www.breadwallet.com" />
+
+	<rule from="^http://www\.breadwallet\.com/"
+		to="https://breadwallet.com/" />
```
J0WI: Why is this redirect needed?

galeksandrp: Definitely, not needed now.

PR #8471, opened 2017-02-09
Update Distributed.net

gopherit:


src/chrome/content/rules/Distributed.net.xml
```diff
@@ -1,24 +1,12 @@
-<!--
-	Nonfunctional subdomains:
-
```
J0WI: There are still som non-functional hosts, please document them:
cvs.distributed.net
ftp.distributed.net
www1.distributed.net
lists.distributed.net
n0cgi.distributed.net

J0WI:
Thanks!

PR #8473, opened 2017-02-09
[FileZilla-project.org] Added download, lib, and trac subdomains

galenide:


src/chrome/content/rules/FileZilla.xml
```diff
@@ -1,16 +1,13 @@
-<!--
-	Nonfunctional hosts in *filezilla-project.org:
-
-		- trac	(refused)
-
--->
 <ruleset name="FileZilla-project.org (partial)">
```
J0WI: Please remove the `(partial)` flag, since this cover all known hosts.

src/chrome/content/rules/FileZilla.xml
```diff
@@ -1,16 +1,13 @@
...
+	<target host="download.filezilla-project.org" />
 	<target host="forum.filezilla-project.org" />
+	<target host="lib.filezilla-project.org" />
 	<target host="svn.filezilla-project.org" />
+	<target host="trac.filezilla-project.org" />
 	<target host="wiki.filezilla-project.org" />
-	<target host="www.filezilla-project.org" />
 
 
 	<securecookie host="." name="." />
```
J0WI: Please use the `.+` expression as recommend by our style guide.

PR #8489, opened 2017-02-12
Update KickassTorrents.xml

Folant:


src/chrome/content/rules/KickassTorrents.xml
```diff
@@ -21,15 +23,12 @@
...
-	<target host="www.kickass.la" />
-	
+
 	<!--kat.am forks-->
 	<target host="kickass.cd" />
 	<target host="www.kickass.cd" />
-	<target host="kickass.mx" />
-	<target host="www.kickass.mx" />
+
+	<target host="katcr.co" />
```
J0WI: Please also add `www.katcr.co`

Folant: Overlooked it, thanks!

PR #8497, opened 2017-02-14
Update 3.cn.xml

ivysrono:


src/chrome/content/rules/3.cn.xml
```diff
@@ -1,14 +1,10 @@
 <!--
-Get most of them in item.jd.com/.*.html
-Relate: See JD.com.xml
-Mismatch:
-	- ^
-	- st
-	
+	Get most of them in item.jd.com/.*.html
+	Relate: JD.com.xml
```
J0WI: You should also update the reference list there.

ivysrono: ?
I don't understand.

J0WI: Sorry, I missed that this file is already listed in `JD.com.xml`

PR #8520, opened 2017-02-15
uniongang.net

galeksandrp:
Issue author: @galeksandrp


J0WI:
This site contains some mixed content. Does this break anything?

galeksandrp:
1 social network share button
1 websocket ad
2 http ad

J0WI:
Thanks for the feedback. I would consider this as trivial.

PR #8521, opened 2017-02-15
novafilm.tv

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/novafilm.tv.xml
+++ b/src/chrome/content/rules/novafilm.tv.xml
@@ -1,0 +1,37 @@
+<!--
+bb.novafilm.tv ⁵
+www.bb.novafilm.tv ⁵
+blog.novafilm.tv ⁵
+oc.s02e13.rus.eng.dvdrip.novafilm.tv ⁵
+feedback.novafilm.tv ¹
+malcolm.novafilm.tv ²
+morena.novafilm.tv ⁵
+ns1.novafilm.tv ⁵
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/novafilm.tv.xml
```diff
@@ -0,0 +1,37 @@
...
+supernatural.s05e07.hdtvrip.rus.novafilm.tv ⁵
+supernatural.s05e08.hdtvrip.rus.novafilm.tv ⁵
+supernatural.s05e09.hdtvrip.rus.novafilm.tv ⁵
+supernatural.s05e10.hdtvrip.rus.novafilm.tv ⁵
+supernatural.s05e11.hdtvrip.rus.novafilm.tv ⁵
+shop.novafilm.tv ¹
+smsvote.novafilm.tv ⁵
+static.novafilm.tv ⁵
+video.novafilm.tv ⁵
+wash.novafilm.tv ⁵
```
J0WI: Wildcard DNS record

galeksandrp: What about https://github.com/EFForg/https-everywhere/pull/10286#discussion_r125181351

J0WI: Same practice: I would prefer a comment about this behavior. The target list can be reduced to those hosts that are really in use. A list of hosts makes no sense here.

PR #8523, opened 2017-02-15
Added rules for ble.de.

Jochen-A-Fuerbacher:


J0WI:
Is this related to bund.de?

Jochen-A-Fuerbacher:
It is the domain of a federal office of Germany and it is also listed on the bund.de website. However it doesn't contain any bund.de domain. I can put a comment to this file referencing the main bund.de ruleset, if you think that this would be useful.

J0WI:
That would be great!

J0WI:
Travis detected an error on `download.ble.de`. Is this subdomain no longer in use or do we just need a more specific test url?

J0WI:
Thanks!

PR #8526, opened 2017-02-15
Remove platform="mixedcontent" for KrebsOnSecurity

q--:
MCB issues were resolved some time ago when the site moved  to a different DDoS mitigation provider.

This commit also changes the `<securecookie>` tag to be conformant with the style guide.


J0WI:
Thanks!

PR #8537, opened 2017-02-17
Update Crowdin.com.xml

gloomy-ghost:


src/chrome/content/rules/Crowdin.com.xml
```diff
@@ -1,48 +1,29 @@
...
-	* Server sends no certificate chain, see https://whatsmychaincert.com
-
-
-	Insecure cookies are set for these hosts:
-
-		- crowdin.com
-		- support.crowdin.com
-
 -->
 <ruleset name="Crowdin.com (partial)">
```
J0WI: I think we can remove the partial flag now

PR #8538, opened 2017-02-18
Add DepFile.com

gopherit:


gopherit:
The Travis CI "fetch" test failed on http://secure.depfile.com --> https://secure.depfile.com with this error:secure (Travis Fetch error: (60, 'SSL certificate problem: unable to get local issuer certificate') )

...even though it appears to be a valid rewrite.  http returns a blank page.  https returns the correct page.  I added the "secure" subdomain to the documentation even though it seems to work fine.

J0WI:
Thanks!
The certificate chain on secure.depfile.com is incomplete, this will break the site in some situations.

PR #8540, opened 2017-02-18
Austrian Post

MilanKral:


J0WI:
I also found:
intranet.post.at
karriere.post.at
ideenwettbewerb.post.at
versandmanager.post.at
paolapanel.post.at
nachschau.post.at
bc.post.at
intranet.post.at

Furthermore, it seems that they have a subdomain for each postcode. Any other subdomain just shows the main site. So I propose to document this behavior and to use a wildcard target here. All known hosts can be used as test urls.

J0WI:
@MilanKral how to you think about using a wildcard target here?

PR #8544, opened 2017-02-19
Add multistre.am

nixxquality:


src/chrome/content/rules/multistre.am.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="multistre.am">
+	<target host="multistre.am" />
```
J0WI: Could you also add www.multistre.am?

nixxquality: There seems to be a bug with the website where https://www.multistre.am/ redirects to the HTTP version, which is why I didn't.
The www. only redirects to non-www anyway. Are you sure that I should?

J0WI: `https://www.multistre.am/` redirects to `http://multistre.am/` and we will redirect it to `https://multistre.am/`. That's totally fine.

J0WI:
Thanks!

PR #8548, opened 2017-02-19
Add rules for 20min.ch

J0WI:


jeremyn:
`20min-tv.ch.xml`:

- [ ] Add `test` for `feedsfast.20min-tv.ch` or mark it as `No working URL known` (root domain gets `403`)

`20min.ch.xml`:

- [ ] Sublist3r doesn't find any `proxy\d.20min.ch` domains, so we can delete `proxy\d` entirely.
- [ ] Delete `static01.20min.ch` entirely because it is `refused` with both HTTPS and plain HTTP.
- [ ] Fix misspelling of `varios` to `various`.

`20minuti.ch.xml`:

- [ ] I get a `502` for https://20minuti.ch , not an `http redirect`. Please update the comment.
- [ ] Add `target` for https://epaper.20minuti.ch . (`epaper` actually seems fine without any mixed content problems, so another option is to turn the current file into `20minuti.ch-mixedcontent.xml` and make a new `20minuti.ch.xml` just for `epaper`. I'm okay with it either way.)

J0WI:
`static01.20min.ch` WFM over plain HTTP

jeremyn:
`static01` works for me now over plain HTTP.

Issue #8487, opened 2017-02-12
Ruleset issue: sci-hub.bz


githubtefo:
https://sci-hub.bz redirect to https://moscow.sci-hub.ac/ and the site is down, so I need to deactivate http to https conversion (http://sci-hub.bz works fine).

Bisaloo:
Fixed by https://github.com/EFForg/https-everywhere/pull/8561.

Bisaloo:
@J0WI, can you close this as well please? Thanks!

PR #8561, opened 2017-02-20
Update sci-hub

Bisaloo:


Bisaloo:
I don't think it's this extension job to redirect users to another mirror. I would like to hear your thoughts on this though.

jeremyn:
Correct me if I'm wrong, but Sci-Hub's mission is legally gray, which means the domains go up and down as the owners try to avoid legal problems. As of right now https://en.wikipedia.org/wiki/Sci-Hub shows the "Current status" as "Down over https." and "Up over http", whatever that means. The rapid changing means that this is not really a great site for HTTPS Everywhere, because we don't react to changes very quickly.

I also don't like relating subdomains to each other unless there's a way to verify they are related. The new ruleset has "sci-hub.io" and "sci-hub.ac", and how do we know that these are related to each other? This goes doubly for redirecting one site to another, which to be clear, isn't done in the new ruleset.

I wonder if in this case we're not doing more harm than good by trying to cover it, if it encourages people to disable HTTPS Everywhere when they have trouble accessing the site. Maybe we should just delete this ruleset altogether. What do you think? Also @gloomy-ghost @Hainish @J0WI.

Bisaloo:
I think your argument makes sense and I would understand the choice of deleting this ruleset.

On the other hand, HTTPS-Everywhere currently covers [kickass-torrents](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/KickassTorrents.xml) or [T411](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/T411.xml) which suffer from exactly the same problems. 

Furthermore, precisely because those websites are in the gray area/illegal, their users may need extra protection against surveillance.

Another reasonable option in my opinion would be to `default_off` the ruleset. That way, only users who know how to enable/disable a ruleset would do it and they probably wouldn't disable the whole extension in case of an issue with this particular website.

J0WI:
I saw a lot of file sharing sites that already implemented preloaded HSTS, so there will be a problem with those domains anyway, when they get banned.
This specific change here seems less risky since we just redirect to https and not changing the host. If we have more complex redirects that likely change in near future we should avoid them.

ilabdsf:
I have uploaded alternative PR #8650, there are more working rules.

Need to think about redirecting to other domains, but removing, for example, sci-hub.bz is not a thing I agree with.

Bisaloo:
Can we prioritize this PR please? The subdomain is down and this rule is currently causing problems, as reported in #8487.

J0WI:
Can you merge this with #8650?

Bisaloo:
> Can you merge this with #8650?

I'd rather not since they are redirecting the user to other mirrors. As I explained in my first comment, I don't think we should do that.

J0WI:
What I prefer in #8650 is the coverage of TLD and subdomains.
I'm okay with just adding simple http => https redirects.

PR #8563, opened 2017-02-21
Update 115.com.xml

ivysrono:


src/chrome/content/rules/115.com.xml
```diff
@@ -1,10 +1,21 @@
 <!--
+	Issues after login:
+		Only http by server:
+			- ^115.com
```
J0WI: HTTPS works, but there is a JS redirect back to HTTP

src/chrome/content/rules/115.com.xml
```diff
@@ -1,10 +1,21 @@
...
+		Use Mismatch sources https://static.115cdn.com/static/pc/d_155.html
+			- home.115.com
+		Refresh home page continued:
+			- msg.115.com
+			- q.115.com
```
J0WI: I could not confirm this, can you give examples?

src/chrome/content/rules/115.com.xml
ivysrono: Which JS?

src/chrome/content/rules/115.com.xml
ivysrono: Could you login?

J0WI: I'm not logged in

ivysrono: 18911129518
3Ykgvx3K

J0WI:
It seems that they changed the login procedure, so I can no longer reproduce the issues.

PR #8565, opened 2017-02-21
Create secitem.xml

bezcitu:


J0WI:
This site has some unrelated CSP issues, but it's preloaded in all major browsers anyway.

PR #8572, opened 2017-02-21
update Ukeer.de.xml

J0WI:


jeremyn:
Add `target` (these both look like aliases for `webmail`):
- [ ] https://mail.ukeer.de
- [ ] https://push.ukeer.de

J0WI:
added

PR #8575, opened 2017-02-21
eztvstatus.com

galeksandrp:


src/chrome/content/rules/eztv.ag.xml
```diff
@@ -1,6 +1,15 @@
 <ruleset name="eztv.ag">
```
J0WI: You may change the name of this rule to somewhat generic.

PR #8580, opened 2017-02-21
torrentz.site

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/torrentz.site.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="torrentz.site">
+	<target host="torrentz.site" />
+
+	<target host="www.torrentz.site" />
+
+	<rule from="^http://www\.torrentz\.site/"
+		to="https://torrentz.site/" />
```
J0WI: Why is this redirect needed?

PR #8587, opened 2017-02-21
[ArtStation] Add ruleset

synthtech:


src/chrome/content/rules/ArtStation.com.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="ArtStation">
+	<target host="artstation.com" />
+	<target host="*.artstation.com" />
```
J0WI: Our style guide recommend to avoid wildcards. Can you document the reason why you used one here?

synthtech: The majority of subdomains are user-created sites. The only official ones that I could find are `www`, `cdn`, `cdna`, and `cdnb`.

PR #8588, opened 2017-02-21
[Kitsu] Add ruleset

synthtech:


src/chrome/content/rules/Kitsu.io.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	Not resolved:
+		- www.kitsu.io
+
+-->
+<ruleset name="Kitsu (partial)">
```
J0WI: Why is this rule partial?

synthtech: Since `www` wasn't working I thought I should label it as partial. I can change it if it's not needed.

J0WI: Referig to your comment www does not exist, so the host is fully covered with those rules.

synthtech: Ok, I changed it then

PR #8594, opened 2017-02-21
diply.com.xml

NationaISecurityAgency:


J0WI:
What is the reason for using a wildcard here? Our style guide recommends to list all targets separately whenever possible. A wildcard can break some subdomains, like `dash.diply.com` in this case.

In our style guide you can also find some handy tools to create rules and searching more subdomains.

PR #8601, opened 2017-02-21
artofproblemsolving.com.xml

NationaISecurityAgency:
Solves: https://github.com/EFForg/https-everywhere/issues/3069#issuecomment-263136274


src/chrome/content/rules/artofproblemsolving.com.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="ArtOfProblemSolving.com">
+	<target host="artofproblemsolving.com" />
+	<target host="*.artofproblemsolving.com" />
```
J0WI: Is there a specific reason for a wildcard or can we avoid it?

PR #8604, opened 2017-02-21
tinkernut.com.xml

NationaISecurityAgency:


src/chrome/content/rules/tinkernut.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="tinkernut.com">
+	<target host="tinkernut.com" />
+	<target host="www.tinkernut.com" />
```
J0WI: Please document the mismatch on `clone.tinkernut.com`

PR #8605, opened 2017-02-21
bitcoinpaperwallet.com.xml

NationaISecurityAgency:
https://github.com/EFForg/https-everywhere/issues/3069#issuecomment-262264505

### DIFF
--- a/src/chrome/content/rules/bitcoinpaperwallet.com.xml
+++ b/src/chrome/content/rules/bitcoinpaperwallet.com.xml
@@ -1,0 +1,6 @@
+<ruleset name="bitcoinpaperwallet.com">
+	<target host="bitcoinpaperwallet.com" />
+	<target host="www.bitcoinpaperwallet.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


src/chrome/content/rules/bitcoinpaperwallet.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="bitcoinpaperwallet.com">
+	<target host="bitcoinpaperwallet.com" />
+	<target host="www.bitcoinpaperwallet.com" />
```
J0WI: Mismatch on www

J0WI:
@Hainish this build shouldn't pass, because the cert served by `www.bitcoinpaperwallet.com` only matches `bitcoinpaperwallet.com`.

Hainish:
This may be because I've merged a PR which was a shallow clone, looking into it now

Hainish:
Re-running it results in a failure - I'll re-run all of the PRs that have been made since the last failure.

src/chrome/content/rules/bitcoinpaperwallet.com.xml
terrorist96: No more mismatch for me.

J0WI: I still have one

Hainish: So do I.

PR #8606, opened 2017-02-21
ouedkniss.com.xml

NationaISecurityAgency:


src/chrome/content/rules/ouedkniss.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="ouedkniss.com">
+	<target host="ouedkniss.com" />
+	<target host="www.ouedkniss.com" />
```
J0WI: See also forum.ouedkniss.com

J0WI:
Thanks!

PR #8608, opened 2017-02-21
teamup.com.xml

NationaISecurityAgency:


src/chrome/content/rules/teamup.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="teamup.com">
+	<target host="teamup.com" />
+	<target host="www.teamup.com" />
```
J0WI: apidocs.teamup.com (mismatch)
blog.teamup.com
calendar.teamup.com

PR #8615, opened 2017-02-21
greenaddress.it.xml

NationaISecurityAgency:


jeremyn:
Thanks for the pull request. The tests failed because the domains in this ruleset are [HSTS preloaded](https://hstspreload.org), and we try not to include those in HTTPS Everywhere. So, I'm closing this pull request.

PR #8625, opened 2017-02-22
Fixed BND rules to pass tests.

Jochen-A-Fuerbacher:


src/chrome/content/rules/Bundesnachrichtendienst.xml
```diff
@@ -4,6 +4,6 @@
 
     <rule from="^http://bnd\.bund\.de/"
         to="https://www.bnd.bund.de/" />
```
J0WI: Can we also get rid of this unnecessary redirect?

PR #8630, opened 2017-02-22
remove Wyenet.xml

J0WI:
Still no HTTPS support


jeremyn:
I think we should have a `(partial)` ruleset for:

- [ ] https://jupiter.wyenet.co.uk (maybe)
- [ ] https://wallace.wyenet.co.uk
- [ ] https://webmail.wyenet.co.uk

What do you think?

J0WI:
done (test failed because of #9107)

PR #8635, opened 2017-02-22
wikifin.be

jordidg:


jordidg:
Site forces redirect to http if no language cookie is present, causing fetch-test to fail

PR #8641, opened 2017-02-22
remove Auditorius.ru.xml

J0WI:
still no HTTPS support


jeremyn:
Please:

- [ ] Remove the reference from [`Audtd.com.xml`](https://github.com/EFForg/https-everywhere/blob/10586fb78d4c0fa9a9138f0070800545ae7949fa/src/chrome/content/rules/Audtd.com.xml)

J0WI:
Done

jeremyn:
I found a URL for `track.audtd.com` through Google and when I visited it with this ruleset active, it redirected forever. I've haven't been able to reproduce the problem but now I don't trust that these domains will work consistently with HTTPS. Also most of the other subdomains for `audtd.com` from `Sublist3r` look internal or at least non-user-facing. Maybe we should just delete `Audtd.com.xml` entirely. What do you think?

J0WI:
I have no idea what `audtd.com` is for, so I'm very happy with just removing it.

jeremyn:
Thanks, merged.

PR #8652, opened 2017-02-22
New ruleset for Vinyl-on-demand.com, which enforces HTTPS rules incor…

aladar42:
…rectly. If a user should for example click a direct non-https link to a release and purchased it, the HTTPS would never be enforced, leaving the customer details exposed. Please see comment for more info.


J0WI:
Thanks!

PR #8667, opened 2017-02-24
Delete zubu.re.xml

Folant:
Host does not exist


galeksandrp:
@Folant In this and later your PR's where host does not exist did you mean NXDOMAIN?

PR #8671, opened 2017-02-24
Delete Widgetbox.com.xml

Folant:
widgetserver.com is for sale
widgetbox.com does not exist


jeremyn:
I've merged the deletion. Just to note:

I agree that `widgetserver.com` and its subdomains are either gone or for sale.

There are some `widgetbox.com` subdomains still around though:

* http://www.widgetbox.com redirects to http://www.flite.com. https://www.widgetbox.com times out.
* http://files.widgetbox.com , http://files-origin.widgetbox.com , http://img.widgetbox.com all exist but fail with both plain HTTP and HTTPS.

I removed the original commit message when I squash-and-merged.

PR #8744, opened 2017-02-24
theintercept.com.xml

NationaISecurityAgency:


jeremyn:
Thanks, but it looks like `theintercept.com` is HSTS-preloaded, and we don't want to add domains that are preloaded, so I'm closing this issue.

PR #8746, opened 2017-02-24
Create studyjamscn.com.xml

ivysrono:


src/chrome/content/rules/studyjamscn.com.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+	Invalid cert:
```
J0WI: It was refused for me.

src/chrome/content/rules/studyjamscn.com.xml
```diff
@@ -0,0 +1,23 @@
...
+		^studyjamscn.com
+
+	MCB:
+		www.studyjamscn.com
+			- Break captcha on signup page.
+			- Break baidu share js on lots pages.
```
J0WI: So do we need to add `platform="mixedcontent"`?

ivysrono: The MCB bug has been fixed now.

PR #8747, opened 2017-02-24
Create 4pda.ru.xml

ivysrono:


src/chrome/content/rules/4pda.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="4pda.ru">
+	<target host="4pda.ru" />
```
J0WI: Please also add:
www.4pda.ru
4pda.to
s.4pda.to

PR #8751, opened 2017-02-25
Update InoReader.xml

ivysrono:


src/chrome/content/rules/InoReader.xml
```diff
@@ -1,9 +1,8 @@
 <ruleset name="InoReader">
-    <target host="www.inoreader.com"/>
     <target host="inoreader.com"/>
+    <target host="www.inoreader.com"/>
```
J0WI: See also:
beta.inoreader.com
blog.inoreader.com
forum.inoreader.com
status.inoreader.com

PR #8768, opened 2017-02-25
Delete Life-in-the-Fast-Lane.xml

Folant:
Host does not exist


Folant:
Very strange indeed. Fails in Russia/Ukraine/Iran. Closing it then.
lifl

PR #8800, opened 2017-02-25
Added dragonsmoke.cloud rule

TheOpenSourceNinja:
I've recently set up the web site dragonsmoke.cloud with HTTPS, so I thought I'd submit a rule for HTTPS Everywhere.

### DIFF
--- a/src/chrome/content/rules/Dragonsmoke.cloud.xml
+++ b/src/chrome/content/rules/Dragonsmoke.cloud.xml
@@ -1,0 +1,15 @@
+<!--
+	Nonfunctional hosts in *.dragonsmoke.cloud:
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Dragonsmoke.cloud">
+	<target host="dragonsmoke.cloud" />
+	<target host="www.dragonsmoke.cloud" />
+
+	<rule from="^http://(www\.)?dragonsmoke\.cloud/" to="https://www.dragonsmoke.cloud/" />
+</ruleset>


src/chrome/content/rules/Dragonsmoke.cloud.xml
```diff
@@ -0,0 +1,15 @@
...
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Dragonsmoke.cloud">
+	<target host="dragonsmoke.cloud" />
+	<target host="www.dragonsmoke.cloud" />
+
+	<rule from="^http://(www\.)?dragonsmoke\.cloud/" to="https://www.dragonsmoke.cloud/" />
```
J0WI: Why are you forcing to redirect to the www subdomain? HTTP is working on both and the redirect works on the servers side.

J0WI:
You should have a look on https://hstspreload.org/ which makes a rule in HTTPS Everywhere obsolete.

src/chrome/content/rules/Dragonsmoke.cloud.xml
TheOpenSourceNinja: The server redirects you to www no matter what, so there's no reason we should redirect to the non-www version. Might as well just skip straight to www.

J0WI: See our upcoming guideline:
https://github.com/EFForg/https-everywhere/blob/3b8017cdd8f36337b817e07b67944a4f0340e376/CONTRIBUTING.md#snapping-redirects

TheOpenSourceNinja:
Already submitted to hstspreload.org :)
I don't think it entirely obsoletes HTTPS Everywhere rules though. HTTPS Everywhere can run on current & outdated browser versions, whereas the HSTS preload list with my site added will presumably only be pushed to future browser versions.
Since I set up dragonsmoke.cloud primarily as an exercise in making my web server as secure as possible, I'm trying to eliminate (as far as possible) the "insecure first connection" problem by adding to both HTTPS Everywhere and hstspreload.org.

jeremyn:
@TheOpenSourceNinja Are you still interested in working on this?

jeremyn:
I'm closing this pull request due to lack of response from @TheOpenSourceNinja . Also, both https://dragonsmoke.cloud and https://www.dragonsmoke.cloud fail HTTPS, so there is no ruleset to make here anyway.

PR #8830, opened 2017-02-26
Delete Newtention.net.xml

Folant:
Host does not exist


jeremyn:
@Folant We like to [squash-and-merge](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) pull requests, which means we combine and rewrite pull requests in your name. Is that all right with you for this and any other pull requests you submit to us?

PR #8832, opened 2017-02-26
Update and rename ImageShack.us.xml to ImageShack.xml

ivysrono:


src/chrome/content/rules/ImageShack.xml
```diff
@@ -0,0 +1,33 @@
+<!--
+	Note:
+		ImageShack Changed policy, stopped image sharing service.
+		So, a.imageshack.us and img\d+.imageshack.us are 404 now, nomatter http or https.
+
+	Chain Contains anchor:
+		blog.imageshack.com
+
+	Mismatch:
+		api.imageshack.us
```
J0WI: But you also added it as target

src/chrome/content/rules/ImageShack.xml
```diff
@@ -0,0 +1,33 @@
+<!--
+	Note:
+		ImageShack Changed policy, stopped image sharing service.
+		So, a.imageshack.us and img\d+.imageshack.us are 404 now, nomatter http or https.
+
+	Chain Contains anchor:
+		blog.imageshack.com
```
J0WI: This is not a functional issue, you can add it as a target.

src/chrome/content/rules/ImageShack.xml
```diff
@@ -0,0 +1,33 @@
...
+
+	<target host="imageshack.us" />
+	<target host="www.imageshack.us" />
+	<target host="imagizer.imageshack.us" />
+	<target host="post.imageshack.us" />
+
+	<!-- Show 404 page: -->
+	<target host="a.imageshack.us" />
+		<test url="http://a.imageshack.us/img254/8219/40789751.png" />
+	<target host="*.imageshack.us" />
```
J0WI: Mismatch on some subdomains, like profile.imageshack.us and reg.imageshack.us.

PR #8835, opened 2017-02-26
Create umontpellier.fr.xml

Bisaloo:
This is a low priority PR because as opposed to what I initially thought, most subdomains are secured by default.

The only true redirection here is for `*.edu.umontpellier.fr`.

Please note that I also skipped some subdomains that looked internal or purely make for testing purposes.

### DIFF
--- a/src/chrome/content/rules/umontpellier.fr.xml
+++ b/src/chrome/content/rules/umontpellier.fr.xml
@@ -10,6 +10,10 @@
 		app.umontpellier.fr (503)
 		federation.umontpellier.fr (503)
 
+	Redirect to http:
+		edu.umontpellier.fr
+		*.edu.umontpellier.fr
+
 -->
 <ruleset name="Université de Montpellier">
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/umontpellier.fr.xml
```diff
@@ -0,0 +1,45 @@
...
+	<target host="cs.umontpellier.fr" />
+	<target host="ebureau.umontpellier.fr" />
+
+	<target host="edu.umontpellier.fr" />
+	<target host="*.edu.umontpellier.fr" />
+
+	<!-- List of subdomains here: https://edu.umontpellier.fr/ -->
+
+		<test url="http://sciences.edu.umontpellier.fr/" />
+		<test url="http://oenologie.edu.umontpellier.fr/" />
```
J0WI: I get redirected back to http on those sites.

Bisaloo:
@J0WI , https://edu.umontpellier.fr now redirects to http as well. I swear this wasn't the case earlier (wouldn't travis have detected it anyways?). Is it okay to wait a couple of days to see how the situation evolves?

I will ping you back soon if you wish.

Bisaloo:
Unfortunately, I don't think the `http` redirection is going to be removed so I guess it can be merged, @J0WI .

PR #8838, opened 2017-02-26
Update and reactivate Sciencemag.org.xml

Bisaloo:
The cert chain has been fixed.

I know I could optimize those regexes but I am not sure what the style guide recommends in this case.

PS: sorry about the terrible commits names, I always click too fast...

### DIFF
--- a/src/chrome/content/rules/Sciencemag.org.xml
+++ b/src/chrome/content/rules/Sciencemag.org.xml
@@ -101,7 +101,7 @@
 		  to="https://www.sciencemag.org/careers/for-employers" />
 	
 	<rule from="^http://(webinar|webinars)\.sciencemag\.org/"
-		  to="http://www.sciencemag.org/custom-publishing/webinars" />
+		  to="https://www.sciencemag.org/custom-publishing/webinars" />
 	
 	<rule from="^http://xxfiles\.sciencemag\.org/"
 		  to="https://www.sciencemag.org/projects/xxfiles" />
--- a/src/chrome/content/rules/Sciencemag.org.xml
+++ b/src/chrome/content/rules/Sciencemag.org.xml
@@ -2,18 +2,13 @@
 	Refused:
 		sciencemag.org
 		aaasmember.sciencemag.org
-		advances.sciencemag.org
 		breakthroughs.sciencemag.org
 		hwmaint.advances.sciencemag.org
 		apps.sciencemag.org
-		beta.sciencemag.org
 		blogs.sciencemag.org
 		comments.sciencemag.org
 		hwmaint.sciencemag.org
-		immunology.sciencemag.org
 		modencode.sciencemag.org
-		news.sciencemag.org
-		newsimages.sciencemag.org
 		newsquiz.sciencemag.org
 		robotics.sciencemag.org
 		sageke.sciencemag.org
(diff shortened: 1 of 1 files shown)


Bisaloo:
Ok, I'm stuck here. This is getting more complex than what I expected. It looks like most of my non-trivial redirection don't work on the pages of a subdomain.

http://sciencecareers.sciencemag.org/ redirects to http://www.sciencemag.org/careers/

- https://www.sciencemag.org/careers/career-resources
- https://www.sciencemag.org/careers/graduate-programs

are valid domains

but 
- http://sciencecarrers.sciencemag.org/career-resources
- http://sciencecarrers.sciencemag.org/graduate-programs

`404`

What should I do in this case?

jeremyn:
Think of a `target` as just saying "if the URL is for this domain, then pay attention because you might need to rewrite it" and the `rule` as saying "if the URL matches this regex, then rewrite it like this". For example your rule [here](https://github.com/Bisaloo/https-everywhere/blob/f4e6e1572c9f054db0b620e91e6ecf26672788b1/src/chrome/content/rules/Sciencemag.org.xml#L97-L98):

```xml
	<rule from="^http://(nextwave|sciencecareers|sciencecareerst)\.sciencemag\.org/"
		  to="https://www.sciencemag.org/careers" />
```

gives instructions for rewriting `sciencecareers.sciencemag.org` URLs, but you never told HTTPS Everywhere to pay attention by creating a `target` for `sciencecareers.sciencemag.org`.

For the specific error: the `rules` Travis job [here](https://travis-ci.org/EFForg/https-everywhere/jobs/205584868) fails because you don't have enough `test`s. For example your `rule` for `^http://(nextwave|sciencecareers|sciencecareerst)\.sciencemag\.org/` needs three `test`s, for http://nextwave.sciencemag.org , http://sciencecareers.sciencemag.org , and http://sciencecareerst.sciencemag.org . The only `test` for this `rule` that you have is the [implicit `test`](https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md) from the `nextwave` `target`. If you add the two other `target`s following the discussion above, then you'll get the two other tests for free.

Bisaloo:
Thank you @jeremyn but I don't think this answers my question. You are giving me general advice on how to write a ruleset with tests but although valid, it doesn't solve the issue at hand where the server redirection is not handled in a usual way.

That's why I didn't add more tests yet (travis errors are pretty self-explanatory now). Because I am not even sure I should write a rewrite rule in this case.

jeremyn:
I don't understand your question in https://github.com/EFForg/https-everywhere/pull/8838#issuecomment-282581671, but it seems to have to do with a subdomain that has the problem I described in https://github.com/EFForg/https-everywhere/pull/8838#issuecomment-282581671. Please fix the known problem and then we can go from there.

Bisaloo:
Please don't merge this ruleset yet! 

The issue described in https://github.com/EFForg/https-everywhere/pull/8838#issuecomment-282581671 has not been fixed!

Bisaloo:
Actually, please disregard my last comments. This ruleset should be fine now and is ready for review.

jeremyn:
I haven't deeply reviewed this ruleset but here are some thoughts so far, please let me know what you think.

A lot of the top comment domains are bad for different reasons than what's given. For example, I get an `invalid certificate` for https://advances.sciencemag.org/ , not `refused`. Maybe this has changed in the two months since you submitted this.

We need to be really careful with special redirecting from one subdomain to a specific path in another subdomain. For example you redirect https://poster.sciencemag.org to https://www.sciencemag.org/custom-publishing/posters , but this breaks http://poster.sciencemag.org/sem/ . Another example is you redirect https://news.sciencemag.org to https://www.sciencemag.org/news , but http://news.sciencemag.org/node/92691/backlinks should actually redirect to https://www.sciencemag.org/node/92691/backlinks.

I'm tempted to just say get rid of all the special redirects and only include `target`s that don't have special handling (`www`, `oascentral`, `survey`) and redirects that we feel pretty safe about (`^`).

Bisaloo:
No, I think you are right. 

I am slightly disappointed that we can't safely add more rules but this is probably the best we can do and it's already a step in the good direction.

jeremyn:
Here's what I came up with, please let me know if you disagree:

Remove from `Bad certificate`, don't add anywhere else:
- [ ] https://labs.sciencemag.org

Remove from `No working URL known`, don't add anywhere else:
- [ ] https://china.sciencemag.org
- [ ] https://dmad.sciencemag.org (404)
- [ ] https://newsqa.sciencemag.org
- [ ] https://stage.sciencemag.org

Remove from `Refused`, add to `Bad certificate`:
- [ ] https://robotics.sciencemag.org

Remove from `Refused`, don't add anywhere else:
- [ ] https://tutil.sciencemag.org

Remove `Target`, add to `No working URL known`:
- [ ] https://oascentral.sciencemag.org

Remove `Target`, don't add anywhere else (I'm not sure we should add these when we can't check past the login page):
- [ ] https://cts.sciencemag.org
- [ ] https://survey.sciencemag.org

Remove from `Time out`, add to `Refused`:
- [ ] https://breakthroughs.sciencemag.org
- [ ] https://comments.sciencemag.org
- [ ] https://modencode.sciencemag.org
- [ ] https://newsquiz.sciencemag.org
- [ ] https://poster.sciencemag.org/sem/
- [ ] https://sciencecareers.sciencemag.org
- [ ] https://sciencecareers.sciencemag.org/meetings
- [ ] https://sciencecareerst.sciencemag.org
- [ ] https://specialprojects.sciencemag.org

Bisaloo:
- I added targets that land on a login page to the top comment under `Private subdomain`. I prefer listing all working subdomains for completeness.
- For your last point, those subdomains are already in `Refused`. Did you mean something else?

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/8838/commits/d4011d089c83546fa262978cd9d069853ec31fa8. I'm not sure what you're asking in your last comment (https://github.com/EFForg/https-everywhere/pull/8838#issuecomment-296809460).

jeremyn:
Thanks, merged.

PR #8846, opened 2017-02-28
Add pckeyboard.com

gopherit:


src/chrome/content/rules/pckeyboard.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="pckeyboard.com">
```
J0WI: Please also add a target for `pckeyboard.com` (without `www.`)

gopherit: Done.  Thanks.

PR #8847, opened 2017-02-28
Ulozenka.cz

MilanKral:


jeremyn:
Thanks for the pull request, but please rewrite it without using a wildcard. Instead, list each subdomain as a separate `target`. You can then get rid of the `test`s.

jeremyn:
Please:

Add `target`:
- [ ] https://api.ulozenka.cz
- [ ] https://beta.ulozenka.cz
- [ ] https://tracking.ulozenka.cz

Also:
- [ ] Remove the top comment, we don't need it.
- [ ] Sort the `target`s in alphabetical order starting from the top level domain at the right reading left, moving `^` and `www` to the top of their group. So in this case, to be very specific, the order should be: `^`, `www`, `api`, `beta`, `m`, `partner`, and `tracking`.

jeremyn:
Thanks, merged.

Issue #8845, opened 2017-02-28
Videos on CBS Local don't show


OrganizedStalking:
http://sanfrancisco.cbslocal.com/2017/02/24/judge-orders-california-to-release-papers-discussing-risk-of-cell-phone-use/

ivysrono:
I find several rulesets show red.
First fix:
#8848

numismatika:
The Anv(ato) player is loaded via plain http and therefore i assumed active mixed content as the problem.
We can't secure that anv.bz stuff as it doesn't support TLS yet.

Bisaloo:
Fixed by #14656 as well @J0WI

PR #8848, opened 2017-02-28
Update AddThis.xml

ivysrono:


J0WI:
I wonder how this is related to #8845?
You dind't change anything functionally except removing one weird exclusion for the main site. The second exclusion is even more weird by the way. I wonder why this is passing our tests, but I won't touch it either.

ivysrono:
Sorry, Maybe this is one bug coused by 5.2.10 on my Firefox just like #8851 .
5.2.11 works great.

PR #8851, opened 2017-02-28
Update googletagservices.com.xml

ivysrono:


ivysrono:
@Hainish @jeremyn @J0WI @gloomy-ghost 
Is it one bug? I'm not sure.
The test keep http without this PR.
I cannot understand.

PR #8855, opened 2017-03-01
[Vancity] Add new rules due to invalid certs

lucaspetter:
See discussion at #7580.

There are invalid certificates at `https://vancityinsurance.com/` and `https://myreloadable.ca/`. But we should still redirect from `http://` to `https://www.` because the `www.` subdomains do have valid certs.

cc @J0WI


lucaspetter:
Thanks!

PR #8862, opened 2017-03-01
[Post.at.xml] add test urls

J0WI:


jeremyn:
- [ ] Change `recircts` to `redirects` [here](https://github.com/J0WI/https-everywhere/blob/c0a06fc42c51cc7b668f0347cc04e71c70b5bf32/src/chrome/content/rules/Post.at.xml#L5)

Looks good otherwise.

J0WI:
done

PR #8866, opened 2017-03-02
fix #8488

### LINKED ISSUE #8488: Unable to access Cable test tool: http://www.dslreports.com/tools/puma6
Trying to access [Cable Modem Puma 6 test](http://www.dslreports.com/tools/puma6) tries to load it using HTTPS, which does not exist on the site. 

Exclusion for that link is desired.
I will try to raise this issue with the site owner if I can find where to do so.

Thanks!

J0WI:


jeremyn:
The `exclusion` `pattern` should end in a `$`. Otherwise this looks good.

J0WI:
done

jeremyn:

PR #8872, opened 2017-03-03
Create Post852.com.xml

cschanaj:


src/chrome/content/rules/Post852.com.xml
```diff
@@ -0,0 +1,20 @@
...
+		Couldn't resolve host name:
+			 - cdn.post852.com
+
+		Timeout was reached:
+			 - dev1.post852.com
+
+		Mixed content blocking (MCB) tiggered:
+			 - www.post852.com
+-->
+<ruleset name="Post852.com">
```
J0WI: Please rename to `Post852.com (partial)`

cschanaj: done. thanks.

PR #8874, opened 2017-03-03
Update and rename StatistikamtBayern.xml to Bayern.de.xml

hlqbjxcg:


J0WI:
There are a lot of more subdomains on `*.bayern.de`. Please also use tabs to indent lines.

PR #8881, opened 2017-03-03
Update and rename Visa_Europe.xml to VisaEurope.xml

ivysrono:


src/chrome/content/rules/VisaEurope.xml
```diff
@@ -0,0 +1,28 @@
+<!--
+	For other Visa coverage, see Visa.xml.
+
+	Failed:
+		visaeurope.pl
+		www.visaeurope.pl
+-->
+<ruleset name="VisaEurope">
+	<target host="visaeurope.com" />
+	<target host="www.visaeurope.com" />
```
J0WI: apprentices.visaeurope.com
futures.visaeurope.com
vision.visaeurope.com

Issue #2324, opened 2015-07-26
Add some hosters/domain registrars


rugk:
- [x] https://www.domaindiscount24.com
- [x] https://rage4.com/
- [x] https://www.royaldns.net/
- [x] https://www.transip.de/
- [x] https://www.rb-media-group.de/
- [ ] https://www.rrpproxy.net/
- [ ] https://www.omc.net/
- [ ] https://www.mk.de/
- [ ] https://www.mcs.de/
- [ ] https://joker.com/
- [ ] https://www.internetx.com
- [ ] https://www.epag.de/
- [ ] https://www.edns.de/
- [ ] https://domainexpress.de/
- [x] https://desec.io
- [x] https://www.dotplex.de/
- [x] https://www.cps-datensysteme.de
- [x] https://www.core-networks.de/
- [x] https://www.123domain.eu/

J0WI:
It would be great to get a pull request from you!
https://github.com/EFForg/https-everywhere/blob/master/ruleset-style.md

rugk:
If I get time I will do.

rugk:
dotplex.de already has a rule: https://github.com/EFForg/https-everywhere/blob/fe756cd650cdc4b81a25b93982fe86be801037f6/src/chrome/content/rules/Dotplex.de.xml

zneaa:
@rugk joker.com.xml https://github.com/EFForg/https-everywhere/pull/8891

zneaa:
@rugk cps-datensysteme.de.xml https://github.com/EFForg/https-everywhere/issues/2324

zneaa:
@rugk core-networks.de.xml https://github.com/EFForg/https-everywhere/issues/8893

jeremyn:
I've closed pull request https://github.com/EFForg/https-everywhere/pull/8892 without merging it, so `cps-datensysteme.de` still needs a ruleset.

RemakingEden:
@J0WI I'm slightly confused on some of these older issues. I'm new to contributing so i'm sorry if this is a basic question. Ive just tested a lot of these and they all seem to resolve to https automatically. 

Should I check all of them and if they all resolve automatically should this be closed? Or if thye resolve automatically, is it still good to make rulesets for them?

Thanks for the help!

rugk:
[...]

PR #8893, opened 2017-03-03
core-networks.de.xml

zneaa:
https://github.com/EFForg/https-everywhere/issues/2324


J0WI:

Issue #7659, opened 2016-11-21
https://abcnews.go.com/ is broken, keeps redirecting to http


securingyour:

zneaa:
PR solves this issue: https://github.com/EFForg/https-everywhere/pull/8894

PR #8894, opened 2017-03-03
ABC_News.xml default breaks site

zneaa:
Reported in https://github.com/EFForg/https-everywhere/issues/7659

### DIFF
--- a/src/chrome/content/rules/ABC_News.xml
+++ b/src/chrome/content/rules/ABC_News.xml
@@ -25,7 +25,7 @@
 			- abcnews.go.com from pixel.quantserve.com
 
 -->
-<ruleset name="ABC News (partial)">
+<ruleset name="ABC News (partial)" default_off="breaks site">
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ABC_News.xml
```diff
@@ -25,7 +25,7 @@
 			- abcnews.go.com from pixel.quantserve.com
 
 -->
-<ruleset name="ABC News (partial)">
+<ruleset name="ABC News (partial)" default_off="breaks site">
```
J0WI: Please use `JS redirect loop` or `JS redirect to http`

J0WI: @zneaa could you change this, that we have this fix in our next release?

J0WI:
I made this minor change myself, but now somebody else have to review this.
cc @Hainish @jeremyn @gloomy-ghost

PR #8898, opened 2017-03-03
[OAuth.net] Update, rename and reactivate ruleset

Bisaloo:


jeremyn:
Please:

- [ ] Change the `ruleset` `name` to `OAuth.net`
- [ ] Delete the [`cert`](https://github.com/Bisaloo/https-everywhere/blob/06c1d8c90aca4e2e163f63af56823f3f3429a5f2/src/chrome/content/rules/OAuth.net.xml#L13) comment, we don't need it
- [ ] Delete all `test`s
- [ ] I think the `rule` would be better as
```xml
	<rule from="^http://(www|blog|oauthwiki|tokens|tumble)\.oauth\.net/"
		to="https://oauth.net/" />
```
This looks a little more normal to me. What do you think?

jeremyn:

PR #8551, opened 2017-02-19
Update srg-ssr.xml

J0WI:

PR #8901, opened 2017-03-04
Update and rename srg-ssr.xml to SRG-SSR.xml

ivysrono:


J0WI:
This conflicts with #8551 were I made some additional changes.

PR #8902, opened 2017-03-04
[Interpol.int] Update, rename and reactivate ruleset

Bisaloo:


src/chrome/content/rules/Interpol.int.xml
```diff
@@ -0,0 +1,33 @@
+<!-- 
+	Non-functional:
```
J0WI: Please double check them, at least `cyber.`, `s.` and `secure.` are working for me.
Could you also add the type of error those hosts have with HTTPS?

Bisaloo: My bad. They didn't work over `http` so I simply assumed they were non-functional. But they indeed work when asked over `https`.

Bisaloo:
I updated the top comment but I did not include the working subdomains you found for the reason listed in the top comment.

- `cyber.` and `s.` are internal
- `secure.` calls a secure connection failed for me

PR #8755, opened 2017-02-25
Update and rename TheChurchofJesusChristofLatterdaySaints.xml to LDS.org.xml

ivysrono:

PR #8906, opened 2017-03-04
Update and Rename Apple.xml to Apple.com.xml and so on.

ivysrono:


ivysrono:
@J0WI Thanks.
There are two PRs In order to find the file more easily: #8908 #8755 .
Would you review them please?

PR #8909, opened 2017-03-04
[LiveStrong.com] Update, rename, and remove mixedcontent platform

Bisaloo:
This one is a bit weird. `sublist3r` finds plenty of subdomains (284 to be exact) like:

- mushroom.livestrong.com
- sciatica.livestrong.com
- kidneys.livestrong.com
- fennel.livestrong.com

but they all redirect to http://www.livestrong.com so they will get caught by the rewrite rule. I didn't want to put an unnecessary wildcard.


J0WI:
Can you also have a look at `*.livestrongcdn.com`?

Bisaloo:
I am not entirely sure what I should do here:

- Add `*.livestrongcdn.com`. The cert is a wildcard one and all subdomains look correctly configured
- On the other hand, only the subdomains I added seem to be used on the pages I looked at

J0WI:
That's totally fine, thanks!

PR #8912, opened 2017-03-05
Create eshuyuan.net.xml

ivysrono:


src/chrome/content/rules/eshuyuan.net.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="eshuyuan.net">
+	<target host="www.eshuyuan.net"/>
+
+	<target host="www.tushuku.net"/>
+
+	<rule from="^http://www\.tushuku\.net/"
+			to="https://www.eshuyuan.net/"/>
```
J0WI: Why is this redirect required?

src/chrome/content/rules/eshuyuan.net.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="eshuyuan.net">
+	<target host="www.eshuyuan.net"/>
+
+	<target host="www.tushuku.net"/>
```
J0WI: Please also add the non-www aliases as targets.

src/chrome/content/rules/eshuyuan.net.xml
ivysrono: They are equal.
tushuku.net does't support https one month ago.

PR #8936, opened 2017-03-06
Update Sinaimg.cn.xml

ivysrono:


J0WI:
I'm also getting a mismatch on `www.sinaimg.cn`, I don't know why Travis is not catching this. cc @Hainish

ivysrono:
snipaste_20170307_193044

I can redirect www to wwws.
However, I can't get a mismatch.

jeremyn:
I'm also getting a mismatch:

>www.sinaimg.cn uses an invalid security certificate.
>
>The certificate is only valid for the following names: *.akamaihd.net, *.akamaized-staging.net, *.akamaized.net, *.akamaihd-staging.net, a248.e.akamai.net
>
>Error code: SSL_ERROR_BAD_CERT_DOMAIN

Maybe this is a situation where the site uses Akamai outside the Chinese GFW, something else inside the GFW that doesn't get a mismatch, and @ivysrono is checking inside the GFW and @J0WI and I are checking outside of it?

J0WI:
I think it's because the host resolves to different IPs and the servers are not configured the same.

ivysrono:
@J0WI Ok on wwws.sinaimg.cn ?

J0WI:
That one seems always fine.

Issue #7330, opened 2016-10-13
Cloudfront.net breaks video


terrorist96:
http://www.vegaspbs.org/election-2016/roundtable-discussions/

`Cloudfront.net` breaks video.

theel0ja:
Screenshot about issue:

![Screenshot about what's happening (Using Vivaldi]
(https://i.imgur.com/K94jTNm.png)

terrorist96:
http://www.pbs.org/wnet/tavissmiley/interviews/presidential-candidates-dr-jill-stein-gary-johnson-pt-1/?show=28582

Jim-G:
Same problem with pbs.org videos, e.g., http://www.pbs.org/video/2365911408/
"cannot load M3U8: crossdomain access denied (2048)

J0WI:
Does #8940 fix this?

terrorist96:
@J0WI IDK, it hasn't been merged yet and HTTPSE hasn't been updated with the merge for me to test.

J0WI:
@terrorist96 you can test it by just using HTTPS manually on this site.

terrorist96:
It works.

terrorist96:
This shoulda been closed.

PR #8940, opened 2017-03-06
Create PBS.com.xml

Bisaloo:


jeremyn:
The "real" site is https://pbs.org, not https://pbs.com, which is just a redirect. Can you please fill this out for pbs.org?

Bisaloo:
I may try to sort out the mess in the `PBS.xml` ruleset but it looks like an ugly can of worms I am not so keen to open.

So feel free to merge as it is unless you are ready to possibly wait a bit.

Bisaloo:
I gave it a first go but I still have a lot of subdomains to go through. You probably prefer waiting until I'm done before reviewing.

Bisaloo:
Before I keep going, I am a little concerned about the travis fetch error

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://pbs.org/ => https://pbs.org/: Too many redirects while fetching 'https://pbs.org/'

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://www.pbs.org/ => https://www.pbs.org/: Too many redirects while fetching 'https://www.pbs.org/'

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://cgi.pbs.org/ => https://www.pbs.org/: Too many redirects while fetching 'https://www.pbs.org/'

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://newshour-tc.pbs.org/ => https://www.pbs.org/: Too many redirects while fetching 'https://www.pbs.org/'

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://pbs.com/ => https://pbs.com/: Too many redirects while fetching 'https://pbs.com/'

> ERROR src/chrome/content/rules/PBS.xml: Fetch error: http://www.pbs.com/ => https://www.pbs.com/: Too many redirects while fetching 'https://www.pbs.com/'

Even if I finish going through the remaining subdomain, it won't fix this.

jeremyn:
`Too many redirects` is what happens when a HTTPS request redirects back to the corresponding plain HTTP request. The add-on then rewrites the plain HTTP request back to HTTPS and it gets in a loop.

What's strange here is that when I visit https://www.pbs.org in a browser, it doesn't redirect. However with `curl https://www.pbs.org -v` , I get a `307 Temporary Redirect`. Maybe this site is handling requests differently based on their user agent?

For reference, I can reproduce the problem with this simple ruleset:

```xml
<ruleset name="Public Broadcasting Service (test)">
	<target host="www.pbs.org" />

</ruleset>
```

I'm not sure what to do about it. Maybe the problem really is temporary and we just need to wait a few days.

Bisaloo:
This is definitely odd...

I don't think it is linked to the user agent because

```sh
curl -A 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0' https://www.pbs.org/
```

still returns `307`.

Same as you, I don't get redirected on firefox or chromium but strangely, I do on the Tor browser.

jeremyn:
I also get a 307 redirect in Tor (and curl), and not in Firefox or Chromium.

[This](https://stackoverflow.com/questions/3806486/curl-cookie-and-307-redirect-issue) StackOverflow discussion (not directly related to this issue) suggested that maybe a cookie or something was being set on the first connection, resulting in a 307 the first time and 200 afterwards. I tried with a new Firefox profile and still did not get a 307. I also tried with the "Low" security setting in Tor, and allowed cookies, and got the redirect even on the second time I visited the URL.

I'm not sure what else to try. @gloomy-ghost @Hainish @J0WI, do you have any ideas why https://www.pbs.org/ returns a 307 using Travis, curl, and Tor, but returns a 200 using Firefox or Chromium?

J0WI:
This servers sniffs for `upgrade-insecure-requests: 1` header:
```
$ curl -IH "upgrade-insecure-requests: 1" https://www.pbs.org/
```

Bisaloo:
Good catch @J0WI, thank you! 

The only option I can think of right now is splitting this ruleset in `PBS.xml` and `PBS-problematic.xml` and put `platform="firefox chromium"` in `PBS-problematic.xml`

I am not particularly fond of splitting rulesets but I don't think we have much of a choice in this case.

jeremyn:
Since this is just a problem with testing, I would rather whitelist it.

Bisaloo:
> Since this is just a problem with testing

But it isn't. It will cause redirect loops in Tor as well.

Sure, you could argue that rulesets causing redirect loops are detected and disabled by HTTPS-Everywhere but I don't really like it. This feature is like a last chance safety net to prevent breakage when the server starts redirecting to `http` without any warning.

But IMO, we should never commit/merge rulesets knowing they will create that kind of issues.

Hainish:
This is a way for websites to say "if your browser allows your request to be protected fully using UIR, allow connecting via HTTPS.  If not, downgrade to HTTP"

This is a way for a website that doesn't know how to rewrite page content so as to avoid MCB to ensure the page loads correctly.

Once Tor Browser moves to Firefox ESR 52, all our supported browsers will support this feature.  We'll then be able to merge this (as well as change our testing suite to deliver the relevant header).

src/chrome/content/rules/PBS.xml
```diff
@@ -33,19 +32,71 @@
...
 <ruleset name="Public Broadcasting Service (partial)" platform="mixedcontent">
 
-	<target host="*.pbs.org" />
+	<target host="pbs.org" />
+	<target host="www.pbs.org" />
+	<target host="account.pbs.org" />
+	<target host="autodiscover.pbs.org" />
+	<target host="bento.cdn.pbs.org" />
+	<target host="gchrome.cdn.pbs.org" />
+	<target host="ga.video.cdn.pbs.org" />
```
J0WI: image.pbs.org
player.pbs.org
www-tc.pbs.org

src/chrome/content/rules/PBS.xml
```diff
@@ -33,19 +32,71 @@
...
+
+	No working URL known:
+		adfs.pbs.org
+		admin.pbs.org
+		aolbrand.pbs.org
+		audio.pbs.org
+		boards.pbs.org
+		(www.)cdn.pbs.org
+		chrome.pbs.org
+		jaws.pbs.org
```
J0WI: `http://jaws.pbs.org/localization/`

Bisaloo:
@J0WI , I am perfectly aware this ruleset is not ready to be merged. I am waiting for the new Tor Browser version to come out to keep going if that's okay with you.

jeremyn:
@J0WI Thanks for taking this over.

J0WI:
I didn't want to take over your review, I just stumbled over this in #7330.

jeremyn:
@J0WI In case it wasn't clear, I was being genuinely appreciative, not sarcastic or anything.

PR #8956, opened 2017-03-07
Update MFi.re.xml

ivysrono:


src/chrome/content/rules/MFi.re.xml
```diff
@@ -1,15 +1,13 @@
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://www.mfi.re/ => https://www.mfi.re/: (51, "SSL: no alternative certificate subject name matches target host name 'www.mfi.re'")
 	For other MediaFire coverage, see MediaFire.xml.
```
J0WI: This is now MediaFire.com.xml

Issue #529, opened 2014-09-02
Audit stable rulesets that cause Mixed Content Blocking


pde:
I just ran the updated MCB test scripts on the stable ruleset library.  Below are the domains that triggered MCB.  We should check which ones actually break things.  Coming down the pipeline I'll also have detection for cert breakage, and a run of these tests on the master branch.  @2d1 making sure you see this.

MCB triggered: http://immunityinc.com (defined)
MCB triggered: http://www.immunityinc.com (defined)
MCB triggered: http://sitemasonmail.com.moses.com (defined)
MCB triggered: http://sitemasonmail.com (defined)
MCB triggered: http://karwansaraypublishers.com (defined)
MCB triggered: http://waffles.fm (defined)
MCB triggered: http://specialforces.com (defined)
MCB triggered: http://www.specialforces.com (defined)
MCB triggered: http://emsisoft.com (defined)
MCB triggered: http://eiseverywhere.com (defined)
MCB triggered: http://www.eiseverywhere.com (defined)
MCB triggered: http://jabber.ru (defined)
MCB triggered: http://www.amway.com (defined)
MCB triggered: http://bluehost.com (defined)
MCB triggered: http://123rf.com (defined)
MCB triggered: http://hostmonster.com (defined)
MCB triggered: http://nationalarchives.gov.uk (defined)
MCB triggered: http://www.nationalarchives.gov.uk (defined)
MCB triggered: http://www.globaltestsupply.com (defined)
MCB triggered: http://globaltestsupply.com (defined)
MCB triggered: http://chronicle.com (defined)
MCB triggered: http://www.esrb.org (defined)
MCB triggered: http://clkads.com (defined)
MCB triggered: http://www.qca.qualcomm.com (defined)
MCB triggered: http://www.e-rewards.com (defined)
MCB triggered: http://opticsinfobase.org (defined)
MCB triggered: http://digitalforensicsmagazine.com (defined)
MCB triggered: http://www.digitalforensicsmagazine.com (defined)
MCB triggered: http://www.ning.com (defined)
MCB triggered: http://www.hubspot.com (defined)
MCB triggered: http://www.hubspot.net (defined)
MCB triggered: http://guardianeatright.co.uk (defined)
MCB triggered: http://www.guardianeatright.co.uk (defined)
MCB triggered: http://dynamitedata.com (defined)
MCB triggered: http://www.dynamitedata.com (defined)
MCB triggered: http://citizen.org (defined)
[...]

PR #8960, opened 2017-03-07
Mixed check

galeksandrp:
If you are not impressed by https://github.com/franciskim/check-mixed-content/commit/master
```bash
check-mixed-content --depth=1 --url {}
```
can be replaced by https://github.com/bramus/mixed-content-scan/commit/master
```bash
~/.composer/vendor/bin/mixed-content-scan https://{} --no-crawl --format=json | jq -e 'select(.level==400)'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,4 @@ services:
 script:
   - test/travis.sh
 env:
-  - TEST=firefox FIREFOX=firefox-dev
-  - TEST=firefox FIREFOX=firefox-latest
-  - TEST=firefox FIREFOX=firefox-esr-latest
-  - TEST=chromium
-  - TEST=rules
-  - TEST=fetch
-  - TEST=preloaded
+  - TEST=mixed
--- a/src/chrome/content/rules/freerunet.ru.xml
+++ b/src/chrome/content/rules/freerunet.ru.xml
@@ -1,7 +1,7 @@
 <!--
 www.freerunet.ru self signed
 -->
-<ruleset name="freerunet.ru" platform="mixedcontent">
+<ruleset name="freerunet.ru">
 	<target host="freerunet.ru" />
 	<target host="www.freerunet.ru" />
 	
--- a/test/travis.sh
+++ b/test/travis.sh
@@ -79,6 +79,11 @@ if [ "$RULESETS_CHANGED" ]; then
     docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" node bash -c "cd /opt/utils/hsts-prune && npm install && node index.js"
     [ `git diff --name-only | wc -l` -eq 0 ]
   fi
+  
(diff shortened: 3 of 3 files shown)


J0WI:
This cannot replace manual review, but it could be a help.
*ping* @Hainish

.travis.yml
```diff
@@ -5,10 +5,4 @@ services:
 script:
   - test/travis.sh
 env:
-  - TEST=firefox FIREFOX=firefox-dev
-  - TEST=firefox FIREFOX=firefox-latest
-  - TEST=firefox FIREFOX=firefox-esr-latest
-  - TEST=chromium
-  - TEST=rules
-  - TEST=fetch
-  - TEST=preloaded
```
Hainish: why?

src/chrome/content/rules/freerunet.ru.xml
```diff
@@ -1,7 +1,7 @@
 <!--
 www.freerunet.ru self signed
 -->
-<ruleset name="freerunet.ru" platform="mixedcontent">
+<ruleset name="freerunet.ru">
```
Hainish: let's keep changes granular.  this belongs in another PR

test/travis.sh
```diff
@@ -79,6 +79,11 @@ if [ "$RULESETS_CHANGED" ]; then
     docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" node bash -c "cd /opt/utils/hsts-prune && npm install && node index.js"
     [ `git diff --name-only | wc -l` -eq 0 ]
   fi
+  
+  if [ "$TEST" == "mixed" ]; then
+    echo >&2 "Ensuring rulesets do not introduce targets containing active mixed content."
+    docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" node bash -c "cd /opt && apt-get update && apt-get install -y libxml2-utils && npm install -g galeksandrp/check-mixed-content#245e56ad5fd5412bdad986276911e5460d9e48c6 && xmllint --xpath '//ruleset[not(@platform)]/target' $RULESETS_CHANGED | grep -q -v '^$' || exit 0 && xmllint --xpath '//ruleset[not(@platform)]/target' $RULESETS_CHANGED | sed 's/<target host=\"//g' | sed 's&\"/>&\n&g' | grep -v '*' | xargs -n1 -i bash -c 'check-mixed-content --depth=1 --url {}' | grep ' has active mixed content!' && exit 1"
```
Hainish: This belongs in a `utils/` subpath, as we've done before for other tests.

J0WI:
If it's possible to add labels depending on Travis results there would be a lot of use cases for it.
E.g. redirects that are not the same on http/https as in the rule, or when there is a big difference in the content.

J0WI:
Maybe we can use this tool to fix #529?

PR #8963, opened 2017-03-08
Create and Update JapanTimes.co.jp.xml

cschanaj:

PR #8964, opened 2017-03-08
Create and Update JapanTimes.co.jp.xml

cschanaj:
This is a re-open of #8963 (closed) because of a mistake.


src/chrome/content/rules/JapanTimes.co.jp.xml
```diff
@@ -0,0 +1,58 @@
...
+
+		SSL peer certificate or SSH remote key was not OK:
+			 - www.japantimes.co.jp
+			 - realestate.japantimes.co.jp
+			 - support.japantimes.co.jp
+
+		Peer certificate cannot be authenticated with given CA certificates:
+			 - bookclub.japantimes.co.jp
+			 - genki.japantimes.co.jp
+			 - ij.japantimes.co.jp
```
J0WI: bookclub, genki and ij have an incomplete cert chain and redirect to HTTP

src/chrome/content/rules/JapanTimes.co.jp.xml
```diff
@@ -0,0 +1,58 @@
...
+			 - bugs.japantimes.co.jp
+			 - cms.japantimes.co.jp
+			 - features.japantimes.co.jp
+			 - ipm-archives.japantimes.co.jp
+			 - showcase.japantimes.co.jp
+			 - aomori.showcase.japantimes.co.jp
+			 - great-tamba.showcase.japantimes.co.jp
+			 - women.japantimes.co.jp
+
+		SSL peer certificate or SSH remote key was not OK:
```
J0WI: https-everywhere is not related to SSH

cschanaj: Sorry!! See https://github.com/EFForg/https-everywhere/pull/8973#discussion_r106060205.

src/chrome/content/rules/JapanTimes.co.jp.xml
cschanaj: Fixed. 

For research purpose, why a `Incomplete certificate chain` has to be treated differently? When I use `curl` to test against the hosts, it throws an error before the redirections. However, Firefox does not. Is this a reason for that?

J0WI: Firefox caches the intermediate certificates, but if you access the site with a brand new profile you get a mismatch too.

cschanaj: I see. thanks!!

PR #8966, opened 2017-03-08
Add SHAttered.io

q--:
One of the domains mentioned in #3069

### DIFF
--- a/src/chrome/content/rules/SHAttered.io.xml
+++ b/src/chrome/content/rules/SHAttered.io.xml
@@ -1,0 +1,15 @@
+<ruleset name="SHAttered.io">
+
+	<target host="shattered.io" />
+	<target host="www.shattered.io" />
+
+
+	<securecookie host=".+" name=".+" />
+
+	<!--
+		Works only without "www." over HTTPS
+	-->
+	<rule from="^http://(?:www\.)?"
+		to="https://" />
+
+</ruleset>


q--:
What am I doing wrong? I already tried changing the rule once (changed `http:\/\/` to `http://`) and it didn't work. It's not one of the things given in the Travis output either:

> ../src/chrome/content/rules/SHAttered.io.xml fails to validate
>
>Two very common reasons for this are the following:
>
>- Missing caret (^) in 'from' attribute: it should be "^http:" and not "http:".
>
>- Missing trailing slashes in 'from' or 'to' when specifying full hostnames: it should be "https://eff.org/" and not "https://eff.org".

And when I test the RegExp in the developer console (`new RegExp(prompt())`) it works just fine.

Why is my rule considered invalid?

q--:
Just [found](https://github.com/EFForg/https-everywhere/blob/54c5c953560ad9c42479c9acd1f5d092ac9734e7/utils/relaxng.xml#L62) the validation rule... Why is always including the hostname required, even if it would be useless as is the case with, for example, this pull request?

J0WI:
@q-- it's required to make clear what you want to do with the `www` host. A valid syntax could be:
```
	<rule from="^http://www\.shattered\.io/"
		to="https://shattered.io/" />
	<rule from="^http:"
		to="https:" />
```

jeremyn:
@q-- Are you still interested in working on this pull request?

jeremyn:
Due to lack of response from @q-- , I'm closing this pull request in favor of my new pull request https://github.com/EFForg/https-everywhere/pull/14027.

PR #8969, opened 2017-03-09
[ENS.fr] Update ruleset and add exclusion pattern

Bisaloo:


Bisaloo:
The tests are bound to fail on this one because the exclusion pattern for `halley.ens.fr` fixes an internal use of this domain.

I am also pretty sure I am still missing a lot of perfectly valid subdomains but `Sublist3r` returns a huge list and most of them are internal so it's a difficult process to filter the public ones out of it.

src/chrome/content/rules/ENS.fr.xml
```diff
@@ -37,6 +43,7 @@
 		- www.lmd ³
 		- www.lps ²
 		- lumiere ³
+		- www.nuit
```
J0WI: Missing reason

src/chrome/content/rules/ENS.fr.xml
```diff
@@ -167,8 +187,8 @@
 	<securecookie host="^(?:www-mail\.(?:biologie|magbio)|www\.cof|crypto(?:bib)?|\.?halley|\.mail\.(?:jourdan|lumiere)|www\.gdr-meso\.phys)\.ens\.fr$" name=".+" />
 
 
-	<rule from="^http://(dg|ecocampus)\.ens\.fr/"
-		to="https://www.$1.ens.fr/" />
+	<rule from="^http://dg\.ens\.fr/"
```
J0WI: Please add a comment for this

src/chrome/content/rules/ENS.fr.xml
```diff
@@ -110,28 +117,42 @@
...
+	<target host="www.documents.ens.fr" />
+	<target host="ecocampus.ens.fr" />
 	<target host="www.ecocampus.ens.fr" />
 	<target host="git.eleves.ens.fr" />
 	<target host="www.eleves.ens.fr" />
 	<target host="hack.ens.fr" />
 	<target host="halley.ens.fr" />
+	
+		<!-- Used to access papers via google scholar. Causes SSL_ERROR_RX_RECORD_TOO_LONG -->
+		<exclusion pattern="^http://halley\.ens\.fr:4550/" />
```
J0WI: You can use `http://halley.ens.fr:4550/robots.txt` as a test url to remove it from the whitelist.

src/chrome/content/rules/ENS.fr.xml
Bisaloo: https://github.com/Bisaloo/https-everywhere/blob/873973e277908d928a9b969ea519aca6eb6a4d89/src/chrome/content/rules/ENS.fr.xml#L71

This comment may seem wrong at the moment because `dg.ens.fr` is unreachable over both `http` and `https` but it is not usually the case. I am assuming this is just a temporary issue and it will be back to "normal" (invalid certificate) soon.

src/chrome/content/rules/ENS.fr.xml
J0WI: Please also remove it from `utils/ruleset-whitelist.csv`

Bisaloo: I was waiting for you to review. Cherry picked and rebased my commits.

PR #8973, opened 2017-03-09
Create Cyberport.hk.xml

cschanaj:


src/chrome/content/rules/Cyberport.hk.xml
```diff
@@ -0,0 +1,87 @@
...
+			 - mail2.cyberport.hk
+			 - map.cyberport.hk
+			 - mdmnsag.cyberport.hk
+			 - mdmstore.cyberport.hk
+			 - mdmxdm.cyberport.hk
+			 - osug.cyberport.hk
+			 - signup.cyberport.hk
+			 - tryipv6.cyberport.hk
+
+		SSL peer certificate or SSH remote key was not OK:
```
J0WI: SSH is not related to https-everywhere

cschanaj: This is a mistake because I have overloaded libcurl error in testing. Replaced `SSL peer certificate or SSH remote key was not OK` with `SSL peer certificate was not OK`. Thanks.

PR #8984, opened 2017-03-10
Exclude farmer.golang.org

josharian:
This is a public record of failed builder runs.
It uses a self-signed cert, and there is
no plan for that to change.
See the discussion at https://github.com/golang/go/issues/16442.

### DIFF
--- a/src/chrome/content/rules/Go_Lang.org.xml
+++ b/src/chrome/content/rules/Go_Lang.org.xml
@@ -6,6 +6,7 @@
 
 	<target host="golang.org" />
 	<target host="*.golang.org" />
+	<exclusion pattern="^http://farmer.golang.org/" />
 
 		<test url="http://blog.golang.org/" />
 		<test url="http://http2.golang.org/" />


josharian:
I'm going to pause on this for the moment, since it sounds like farmer.golang.org might be getting real certs. Two bits of friendly feedback from my first-time-contributor experience:

It'd be useful to document in the testing part of the readme how to set up for running rule tests--in particular, which python packages are required. I spent a while trying to get the tests to run before giving up. IIRC, I ultimately failed due to missing Levenshtein package. (Maybe it was a PYTHONPATH problem? Any which way, my "figure it out" time quota expired around there.)

It'd also be useful to document in the rules page how to write a test / what the test entries mean / what attributes a test can have. I wasn't sure how to add a test that farmer.golang.org *didn't* get rewritten.

Thanks!

jeremyn:
Thanks for the pull request @josharian.

All the necessary development packages should have been installed by running `bash install-dev-dependencies.sh` as described in [`README.md`](https://github.com/EFForg/https-everywhere/blob/899fffe6e87dac4b63b585f04e57b776d51419ac/ruleset-style.md).

We have a [`ruleset-testing.md`](https://github.com/EFForg/https-everywhere/blob/51b0f13a73ce02321fd5d0dceb2a1237ab043519/ruleset-testing.md) describing tests. In this specific case it would be enough to add a `test` for the specific `exclusion` URL:

```xml
		<test url="http://farmer.golang.org/" />
```

This is also a problematic ruleset because of the wildcard, which I know you didn't add. We prefer to list subdomains with individual `target`s where possible, instead of using wildcards, partly to prevent exactly the problem you were trying to fix here.

Anyway this is just FYI. I know our documentation is a little scattered. You are welcome to ask for help in pull requests if you get stuck.

josharian:
Thanks, @jeremyn. As of https://go-review.googlesource.com/38798, farmer.golang.org supports https courtesy of letsencrypt, so I'm abandoning this.

jeremyn:
Thanks for letting us know.

PR #8988, opened 2017-03-11
Merge Couchsurfing.xml and couchsurfing.com.xml

Bisaloo:


Bisaloo:
I should probably add a comment saying that http://couchsurfing.org redirects to http://www.couchsurfing.com but I'm not what would be more suited: in the top comment or right before the related targets?

J0WI:
As long as couchsurfing.org supports HTTPS I'm ok with this.

PR #8990, opened 2017-03-11
Update System76.com.xml

Bisaloo:


jeremyn:
I haven't reviewed this in detail, but we definitely should not rewrite `mail.system76.com` to `fastmail.com`. Cross-site rewrites can be a security problem. And what if they change their email host?

PR #7767, opened 2016-12-05
Update MicroAd, Inc related domains

cschanaj:

PR #9004, opened 2017-03-12
Create MicroAd.co.jp.xml

cschanaj:
Related: #7767

### DIFF
--- a/src/chrome/content/rules/MicroAd.co.jp.xml
+++ b/src/chrome/content/rules/MicroAd.co.jp.xml
@@ -1,0 +1,30 @@
+<!--
+	For more MicroAd, Inc related ruleset, see MicroAdInc.com.xml
+
+	Non-functional hosts
+		Timeout was reached:
+			 - mail1.microad.co.jp
+			 - ns1.microad.co.jp
+
+		Peer certificate cannot be authenticated with given CA certificates:
+			 - president.microad.co.jp
+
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/MicroAd.co.jp.xml
```diff
@@ -0,0 +1,30 @@
+<!--
+	For more MicroAd, Inc related ruleset, see MicroAdInc.com.xml
```
J0WI: Please also update the reference list on MicroAdInc.com.xml
Can you also remove the reference to `MicroAdInc.com.xml` there, because this would be a reference to itself.

src/chrome/content/rules/MicroAdInc.com.xml
```diff
@@ -1,5 +1,6 @@
 <!-- 
 	MicroAd, Inc related rulesets:
+		- MicroAd.co.jp.xml
 		- MicroAd.jp.xml
 		- MicroAd.net.xml
 		- MicroAdInc.com.xml
```
J0WI: Can you remove the reference to MicroAdInc.com.xml here, because this would be a reference to itself.

cschanaj: Removed self-reference in `MicroAdInc.com.xml`. Thanks!!

PR #9015, opened 2017-03-12
exclude www.mmo-champion.com because of mcb

J0WI:


PekaFat:
When it will in chrome store? I can't browse mmo-champ because all css is broken now...

PR #9021, opened 2017-03-12
[SpamGourmet.com] Update and rename ruleset

Bisaloo:
The subdomains 

- gourmet.spamgourmet.com
- gourmet7.spamgourmet.com
- gourmet8.spamgourmet.com
- gourmet9.spamgourmet.com

all redirect to www.spamgourmet.com and could have a rewrite rule but I am assuming they are only used for testing/dev purposes.


src/chrome/content/rules/SpamGourmet.com.xml
```diff
@@ -0,0 +1,22 @@
...
+-->
+
+<ruleset name="SpamGourmet">
+
+	<target host="spamgourmet.com" />
+	<target host="www.spamgourmet.com" />
+	<target host="secure.spamgourmet.com" />
+
+	<rule from="^http://secure\.spamgourmet\.com/"
+		to="https://www.spamgourmet.com/index.pl/" />
```
J0WI: Please remove the `index.pl/` here.

PR #9022, opened 2017-03-12
Create imgwe.com.xml

ivysrono:


src/chrome/content/rules/imgwe.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="imgwe.com">
+	<target host="imgwe.com" />
+	<target host="www.imgwe.com" />
+		<test url="http://imgwe.com/img/677.jpg" />
```
gloomy-ghost: I get 404 for this URL

PR #9031, opened 2017-03-12
transparency.org.ru

galeksandrp:


src/chrome/content/rules/transparency.org.ru.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+azbuka.transparency.org.ru ¹
```
J0WI: redrabbit.transparency.org.ru
new.transparency.org.ru
old.transparency.org.ru

galeksandrp: Added.

PR #9048, opened 2017-03-13
Update NetEase.xml

ivysrono:


src/chrome/content/rules/NetEase.im.xml
```diff
@@ -0,0 +1,21 @@
...
+	Related:
+		NetEase.com.xml
+
+	MCB:
+		bbs.netease.im
+
+	Timeout:
+		^netease.im
+		www.netease.im
+		dev.netease.im
```
J0WI: dev WFM, but I get a HTTP redirect on ^ and www

src/chrome/content/rules/NetEase.im.xml
```diff
@@ -0,0 +1,21 @@
...
+-->
+
+<ruleset name="NetEase.im">
```
J0WI: This rule may be flagged as partial

Issue #8959, opened 2017-03-07
Add echo.opera.com to Opera rules


anewuser:

RobGraham72:
https://github.com/EFForg/https-everywhere/pull/9051

advaith1:
You can create a pull request [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Opera.xml).

Bisaloo:
This is fixed by PR #6938 (once it's merged)

PR #9051, opened 2017-03-13
add echo.opera.com

RobGraham72:
https://github.com/EFForg/https-everywhere/issues/8959


J0WI:
I included this in my pending #6938

PR #9056, opened 2017-03-13
asana.com

### LINKED ISSUE #7668: HTTPS rewrite for connect.asana.com causes 404
Hi, when opening this link from a newsletter it brings to a 404 error page instead of the HTTPS one :   http://connect.asana.com/1L9Q-4MG6Y-NJ386K-2DOSJU-1/c.aspx

> I disabled the rule affecting that particular site in my own copy of HTTPS Everywhere by clicking on the HTTPS Everywhere toolbar button and unchecking the rule for that site.

galeksandrp:
fixes #7668

### DIFF
--- a/src/chrome/content/rules/Asana.xml
+++ b/src/chrome/content/rules/Asana.xml
@@ -76,7 +76,6 @@ pixel.asana.com different content
 <ruleset name="asana.com">
 	<target host="asana.com" />
 	<target host="www.asana.com" />
-	<target host="*.asana.com" />
 	<target host="1310app.asana.com" />
 	<target host="8atzywvvgloeqjvr.asana.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Asana.xml
```diff
@@ -1,49 +1,415 @@
...
+⁷ protocol error
 -->
-<ruleset name="Asana.com (partial)">
-
-	<!--	Direct rewrites:
-				-->
+<ruleset name="asana.com">
 	<target host="asana.com" />
+	<target host="www.asana.com" />
+	<target host="*.asana.com" />
```
J0WI: This wildcard will break all hosts you commented above.

galeksandrp: Removed.

PR #9064, opened 2017-03-13
Update The-Nature-Conservancy.xml

Bisaloo:


Bisaloo:
The cert for https://voice.nature.org/ expired right between my commits , how random is that?

src/chrome/content/rules/Nature.org.xml
```diff
@@ -0,0 +1,74 @@
+<!--
+	Bad certificate:
```
J0WI: It would be better to have more detailed information, like cert mismatch, expired, chain issues etc.

src/chrome/content/rules/Nature.org.xml
```diff
@@ -0,0 +1,74 @@
...
+	<target host="gift.nature.org" />
+	<target host="giftplanning.nature.org" />
+	<target host="www.giftplanning.nature.org" />
+	<target host="gifts.nature.org" />
+	<target host="global.nature.org" />
+	<target host="help.nature.org" />
+	<target host="my.nature.org" />
+	<target host="support.nature.org" />
+	<target host="stage.nature.org" />
+	<target host="water.nature.org" />
```
J0WI: This host has mixed content issues.

PR #9072, opened 2017-03-14
create Memcached.org.xml

J0WI:


jeremyn:
- [ ] It looks like the `build` port is 8010, not 8080 as you say in the top comment

Looks good otherwise.

jeremyn:

PR #9073, opened 2017-03-14
[Instructables.com] Update, rename and reactivate ruleset

Bisaloo:


src/chrome/content/rules/Instructables.com.xml
```diff
@@ -0,0 +1,43 @@
...
+
+	<target host="instructables.com" />
+	<target host="www.instructables.com" />
+	<target host="cdn.instructables.com" />
+	<target host="ssl.instructables.com" />
+
+	<securecookie host="^\.instructables\.com$" name=".+" />
+
+	<rule from="^http://instructables\.com/"
+		to="https://www.instructables.com/" />
```
J0WI: This redirect is not necessary.

PR #9080, opened 2017-03-14
add more subdomains for DynamicYield.com

numismatika:


hzsolt94:
Hi @numismatika !

You've added a lot of domains, which has invalid certificate. HTTPS Everywhere does not include breaking rules, so please do not include those domains. The standard practice is to list the non-working domains in the comments.

numismatika:
It is just temporary until i get back to my computer and have some free time to mend this file

PR #9083, opened 2017-03-14
Add api.ooyala.com to Ooyala.xml

numismatika:
- add api-subdomain

### DIFF
--- a/src/chrome/content/rules/Ooyala.xml
+++ b/src/chrome/content/rules/Ooyala.xml
@@ -34,6 +34,7 @@
 <ruleset name="Ooyala.com (partial)" platform="mixedcontent">
 
 	<target host="ak-c.ooyala.com" />
+	<target host="api.ooyala.com" />
 	<target host="backlot.ooyala.com" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Do you have an example of an `api.ooyala.com` URL where the response isn't empty? It's a `200`, which is fine, but it would be nice to add a (stable) `test`.

numismatika:
https://api.ooyala.com/docs/api_scratchpad?url=&method=GET&api_key=

jeremyn:
Good find, but let's use https://api.ooyala.com/docs/v2/ , which I found from your example but is simpler.

Can you please change the ruleset so it has this?

```xml
	<target host="api.ooyala.com" />
		<test url="https://api.ooyala.com/docs/v2/" />
```

Otherwise this pull request is ready to merge.

numismatika:
Coming right up, in general we should revisit all the ooyala domains and the exclusion later.

jeremyn:
Thanks, merged.

PR #9095, opened 2017-03-15
Add Freerangekitten.com.xml as a simple ruleset (controlled by EFF) a…

Hainish:
…nd use it for the URL rewrite test

### DIFF
--- a/src/chrome/content/rules/EFF.xml
+++ b/src/chrome/content/rules/EFF.xml
@@ -4,6 +4,7 @@
 		- Copyright-Watch.org.xml
 		- Dear_FCC.org.xml
 		- Democracy.io
+		- Freerangekitten.com.xml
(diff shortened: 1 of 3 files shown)


jeremyn:
Great, thanks @Hainish.

Do you want to make a note in the file that this is an EFF site and/or is used in our tests and/or that empty responses are expected?

Also, you can remove the current no-information top comment. (I know it's added by the provided ruleset generator.)

Hainish:
@jeremyn comment added in both `EFF.xml` and `Freerangekitten.com.xml`

Hainish:
This failed due to a certificate expiration for https://globalchokepoints.org/, which I believe is a problem we need to fix.  I'll follow up with the internal team which handles this and comment on this issue when that's resolved.

Hainish:
The certificate problem is fixed, re-running fetch test.

PR #9105, opened 2017-03-16
Create theinquirer.net.xml

Folant:


Folant:
``['src/chrome/content/rules/Busybox.net.xml', 'src/chrome/content/rules/busybox.net.xml'] failed case-insensitivity testing.
Rules exist with identical case-insensitive names, which breaks some filesystems.``
Honestly, I don't get the logic behind this.

J0WI:
I fixed this in #9107

PR #9107, opened 2017-03-16
remove duplicated Busybox.net.xml

J0WI:
This duplicated rule is currently breaking our builds, see #6494


NickCraver:
This breaks me locally as well (generating an apparently random diff), everyone doing git commits locally is likely affected due to hooks.

PR #9108, opened 2017-03-16
[Ulule.com] Remove wildcard and platform + rename ruleset

Bisaloo:


Bisaloo:
Tests should be re-run. See #9107.

PR #9110, opened 2017-03-16
Urgent: fix redirect loops for meta.*.stackexchange.com

NickCraver:
The `meta.*.stackexchange.com` domains have moved to `*.meta.stackexchange.com` to for HTTPS support via a wildcard. This now does the appropriate redirects to the new homes. Child `blog.*.stackexchange.com` sites (actually only gaming) are unchanged and will not move...they continue to downgrade.

Long-term, we can narrow this entire rule set down to the blogs, all other domains are getting HTTPS support and HSTS headers.

This is an urgent issue since users of HTTPS Everywhere are currently getting redirect loops from us rolling out HTTPS support here. It's also my first time modifying these rulesets, if I'm doing something completely incorrectly please let me know and I'll fix it ASAP, or feel free to do so on your end, whatever fixes redirects for our combined user base faster. Thanks!

cc @bardiharborow @Bisaloo @jeremyn

### DIFF
--- a/src/chrome/content/rules/Stack-Exchange.xml
+++ b/src/chrome/content/rules/Stack-Exchange.xml
@@ -7,7 +7,7 @@
 
 		- www.stackapps.com ==> Mismatched
 		- mathoverflow.com ==> No response (only a redirect to mathoverflow.net anyway)
-		- meta.*.stackexchange.com ==> Mismatched
+		- meta.*.stackexchange.com ==> Mismatched (now *.meta.stackexchange.com)
 		- blog.*  ==> Mixed Active Content
 		- www.stackoverflow.blog ==> Mismatched
 
@@ -46,13 +46,16 @@
 		<test url="http://unix.stackexchange.com/" />
 		<test url="http://writers.stackexchange.com/" />
 
-		<!-- deeper domains cause certificate mismatch errors - there is also a downgrade rule below -->
-		<exclusion pattern="^http://([\w.-]+)\.([\w-]+)\.stackexchange\.com"/>
+		<!-- old child meta domains cause certificate mismatch errors -->
+		<exclusion pattern="^http://meta\.([\w-]+)\.stackexchange\.com"/>
 		<test url="http://meta.unix.stackexchange.com/" />
 		<test url="http://meta.opendata.stackexchange.com/" />
-		<test url="http://blog.gaming.stackexchange.com/" />
 		<test url="https://meta.unix.stackexchange.com/" />
 		<test url="https://meta.opendata.stackexchange.com/" />
+
+		<!-- gaming blog causes certificate mismatch errors -->
+		<exclusion pattern="^http://blog\.gaming\.stackexchange\.com"/>
+		<test url="http://blog.gaming.stackexchange.com/" />
 		<test url="https://blog.gaming.stackexchange.com/" />
 
 		<!-- but this one should not -->
@@ -138,11 +141,13 @@
 	<securecookie host="^teststackoverflow\.com$" name=".+" />
 
 <!-- Rules -->
-	
-	<!-- https links from other pages to these will cause MCB for important elements, hence the downgrades -->
-	<rule from="^https://([\w.-]+)\.([\w-]+)\.stackexchange\.com/" 
-		to="http://$1.$2.stackexchange.com/" downgrade="1" />
 
+	<!-- meta.* sites moved to *.meta - we can safely redirect to their new equivalents, which support https -->
+	<rule from="^https://meta\.([\w-]+)\.stackexchange\.com/" to="https://$1.meta.stackexchange.com/" />
+	
+	<!-- old gaming blog link does not support https -->
+	<rule from="^https://blog\.gaming\.stackexchange\.com/" to="http://blog.gaming.stackexchange.com/" downgrade="1" />
+	
 	<rule from="^http://www\.stackoverflow\.blog/" to="https://stackoverflow.blog/" />
 	
 	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


NickCraver:
I don't think the failure here is related to my commit, it appears to be from https://github.com/EFForg/https-everywhere/pull/6494 earlier, which also gave me odd behavior locally. It was a hurdle to running tests. I assumed it was my setup, but it appears to have broken every PR since on the build system as well.

Failure log:
```
['src/chrome/content/rules/Busybox.net.xml', 'src/chrome/content/rules/busybox.net.xml'] failed case-insensitivity testing.
```

Vogel612:
seems this is basically blocked by #9107 regarding the travis-check

jeremyn:
I've merged #9107, so that problem is gone.

@NickCraver I can merge this as-is, but a few comments, mainly for your reference and the reference of anyone else reading this:

I'm a little unhappy with the https://blog.gaming.stackexchange.com `downgrade`. That subdomain has a specific `exclusion` already, so I think the point of the downgrade is to change HTTPS to plain HTTP if someone navigates to an explicit HTTPS link. That's not really the purpose of HTTPS Everywhere. But, it's a subset of the wildcard `downgrade` that already exists in this ruleset, so that's fine with me in this specific situation.

If possible, in [this](https://github.com/NickCraver/https-everywhere/blob/69d11b346fc318677e58463204cc7768b9263011/src/chrome/content/rules/Stack-Exchange.xml#L145) comment:

```xml
	<!-- meta.* sites moved to *.meta - we can safely redirect to their new equivalents, which support https -->
```

add a link to a StackExchange blog post or something saying this redirect is actually safe to do. I believe you, since you're the site administrator, but a reviewer looking at this ruleset a year from now without context might not know what to do with that assertion.

Did you test your changes in the `HTTPSEverywhereUserRules/` directory as described [here](https://www.eff.org/https-everywhere/rulesets#testing)?

Let me know when you're ready and I'll merge this.

NickCraver:
@jeremyn All good notes, thanks! I was me trying to strictly maintain the old behavior for blog which really isn't necessary - `https://` links to it *never* could have worked before, and so shouldn't exist. I removed the downgrade and added a meta post link. A larger blog post [is in progress](https://github.com/NickCraver/nickcraver.github.com/issues/26), but it'll take me a bit.

I tested using the automated suite, but don't I have/use Firefox on my Macbook which was the easiest place to deal with this repo.

src/chrome/content/rules/Stack-Exchange.xml
```diff
@@ -138,10 +140,10 @@
...
-	
-	<!-- https links from other pages to these will cause MCB for important elements, hence the downgrades -->
-	<rule from="^https://([\w.-]+)\.([\w-]+)\.stackexchange\.com/" 
-		to="http://$1.$2.stackexchange.com/" downgrade="1" />
+
+	<!-- meta.* sites moved to *.meta - we can safely redirect to their new equivalents, which support https -->
+	<!-- details: https://meta.stackexchange.com/questions/292058/network-wide-https-its-time -->
+	<rule from="^https://meta\.([\w-]+)\.stackexchange\.com/" to="https://$1.meta.stackexchange.com/" />
 
 	<rule from="^http://www\.stackoverflow\.blog/" to="https://stackoverflow.blog/" />
```
Bisaloo: I think this rule can be removed. This domain is now properly handled with the trivial rewrite.

This is unrelated to the issue at hand though and it doesn't necessarily has to make it in this particular PR.

jeremyn: The redirect from `meta.*` to `*.meta` doesn't happen in HTTPS though, in fact it can't because they don't have valid certificates for `meta.*`. So I'm okay with rewriting it here.

Bisaloo: My bad, something went wrong with my comment. I was talking about the rule line 148:

```xml
<rule from="^http://www\.stackoverflow\.blog/" to="https://stackoverflow.blog/" />
```

jeremyn: That's not handled by the trivial rewrite, because it's sending `www` to `^`.

jeremyn:
@NickCraver I've merged the commits.

@Hainish Can we do a release soon so we can get these changes out?

src/chrome/content/rules/Stack-Exchange.xml
Bisaloo: Yes, what I meant is that https://www.stackoverflow.blog is valid now. Sure, it's 301 redirected to https://stackoverflow.blog/ but according to the ruleset style guidelines, we should use the trivial rewrite in this case, shouldn't we?

jeremyn: Oh, I see what you mean. You're right, we shouldn't have this rule. However the whole ruleset needs review and we don't need to remove that in this pull request.

Hainish:
@jeremyn I can make a release today.  I usually default against Friday releases but I won't be able to make one next week.

Hainish:
Release made!  I didn't open a `release-forthcoming` ticket for it, since this was an unexpected release and I cut some of the unnecessary steps out.

jeremyn:
Thanks @Hainish.

NickCraver:
Thanks everyone, we greatly appreciate getting this dealt with swiftly! Also kudos on the entire project: tests, checks, hooks, etc. here. The repo itself is great for new users completely unfamiliar with the structure (e.g. me) to contribute.

Hainish:
@NickCraver thanks for the kind words.  You're lucky you jumped in when you did - it was just yesterday that we merged in a comprehensive buildout of our documentation.

Issue #7662, opened 2016-11-21
Rejecting targets without a working HTTP 200/3XX URL


jeremyn:
I've started rejecting `target`s where we can't find even one working `HTTP 200` URL, for example see https://github.com/EFForg/https-everywhere/pull/7636#issuecomment-261987564 . Often both HTTP and HTTPS return the same non-`200` code, like `403`. My suggestion is to note these URLs under a new error category `No working URL known`.

@fuglede @gloomy-ghost @Hainish @J0WI What do you all think?

EDIT: We should count `HTTP 3XX` URLs as well as `200`. I've updated the issue description.

jeremyn:
I noticed @gloomy-ghost 's comment in https://github.com/EFForg/https-everywhere/pull/7481#issuecomment-260155645 , where they wrote:

> I'm quite aggressive on adding targets. And I think that these kind of risks are still acceptable since they are easy to fix as well.

Problems caused by adding internal URLs may not be easy to fix. Internal URLs that are used to manage corporate firewalls, communications, VPNs etc may depend on complex applications from a third party. If a sysadmin finds a problem with one of these sites, they are going to spend a bunch of time narrowing the problem down to HTTPS Everywhere and then just disable it. If it ends up wasting enough of the third party vendor's (Dell, Cisco etc) time, the vendor will simply advise their customers not to use HTTPS Everywhere. The odds that someone from the third party will contact us to work out the problem are very low.

Another problem is that the "easy to fix" approach means we are willing to disable rules because of problems we can't test just because someone asks us to do that. That opens an easy attack path.

gloomy-ghost:
The targets I asked to add in #7481 already have working https index pages
(for login). So the risks of adding these targets should be similar with
adding non-English sites which require logins. (I don't know will the other
reviewers apply accounts to test sites like those sites though)

Also the reporting rate of non-English sites will be even lower than
internal sites in English because internal site's users at least have good
English skills.
e.g. My first PR to HTTPS Everywhere #6765, fixed a login problem of an
Alexa top-100 site in China, which has existed for more than 2 years.

jeremyn:
I'm not sure how to measure risk here. I think that it's more likely a corporate-type web application will have trouble than a site intended for the general public, and that it's less likely a corporate user will report a problem than a general public user.

jeremyn:
We should count HTTP 3XX URLs as well as 200. I've updated the issue description. I've updated the issue description and first comment.

[...]

PR #9113, opened 2017-03-17
[ClamAV.net] Add subdomains

Bisaloo:


jeremyn:
Instead of `No working URL known`, these should be:
- [ ] https://intel.clamav.net is `Time out`
- [ ] https://snapshot.clamav.net is `Invalid certificate`
- [ ] https://stats.clamav.net is `Time out`
- [ ] https://win.clamav.net is not in DNS and can be removed entirely

Also: 
- [ ] Change `Dropped` to `Time out`
- [ ] You can remove ` (partial)` from the `name`

Bisaloo:
I tend to think that no working URL known should always get the precedence compared to other comments.

A given subdomain may be unreachable over `https` due to an invalid certificate but does it matter if this subdomain in not in use anyways and 404/times out/returns a blank page over `http`?

jeremyn:
`No working URL known` means that there's nothing technically wrong with the connection, but we didn't add it because we couldn't find one working 2xx/3xx response at that subdomain using HTTPS, so we're not comfortable rewriting it. So, if that subdomain gets an actual HTTPS error like `Invalid certificate`, then we shouldn't use `No working URL known` because there is actually something wrong with the connection.

I proposed `No working URL known` in issue https://github.com/EFForg/https-everywhere/issues/7662 and you can see a longer discussion there.

jeremyn:

Bisaloo:
@jeremyn, I think it would be useful to have another section for URL that don't work even over `http`. What do you think?

If a ruleset only has subdomains in this category, it means there is nothing wrong with `https` and the domain is fully covered, with all the functioning subdomains.

This is useful because most of the sites have subdomains they don't use anymore but that are still listed by `Sublist3r`. I think we are unnecessarily bloating the top comment with those.

Another option would be to ignore those subdomains completely but it would probably be more confusing when people come back later to update this ruleset.

jeremyn:
I'm not sure what you're asking for. You want add subdomains to the top comment even if they don't work under plain HTTP? I don't want to add more stuff to the top comments if we can avoid it.

Bisaloo:
Okay, let's take for example `snapshot.clamav.net`:

- http://snapshot.clamav.net times out
- https://snapshot.clamav.net is mismatched

Until now, I would add `snapshot.clamav.net` to the list of `No working URL known`. However, you convinced me this is not the purpose of this section.

I am simply wondering what I should do when I encounter this kind of behaviour. Should the subdomain still be listed? Under which section?

jeremyn:
If both HTTPS and plain HTTP are broken, I wouldn't mention the subdomain at all, even if they are broken for different reasons.

PR #9118, opened 2017-03-18
Update ServiceNow.com.xml

gopherit:
Adding docs subdomain.  Removing knowledge -> community rewrite rule.

### DIFF
--- a/src/chrome/content/rules/ServiceNow.com.xml
+++ b/src/chrome/content/rules/ServiceNow.com.xml
@@ -57,17 +57,15 @@
 	<target host="servicenow.com" />
 	<target host="community.servicenow.com" />
 	<target host="developer.servicenow.com" />
+	<target host="docs.servicenow.com" />
 	<target host="express.servicenow.com" />
+	<target host="knowledge.servicenow.com" />
 	<target host="partners.servicenow.com" />
 	<target host="www.partners.servicenow.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ServiceNow.com.xml
```diff
@@ -55,19 +54,16 @@
 	<!--	Direct rewrites:
 				-->
 	<target host="servicenow.com" />
+	<target host="www.servicenow.com" />
 	<target host="community.servicenow.com" />
 	<target host="developer.servicenow.com" />
+	<target host="docs.servicenow.com" />
 	<target host="express.servicenow.com" />
+	<target host="knowledge.servicenow.com" />
```
J0WI: This host has not the same content on https as on http.

gopherit: @J0WI can you please provide examples of https and http that aren't the same?  I cannot find any.  I removed the rule from knowledge to community, but http://knowledge is redirected to https://knowledge by their server.

J0WI: Ah sorry, HTTPS Everywhere currently is rewriting to a different location, but you just fixed this.

gopherit: Great.  Yeah, it's been a bug for a long time.

PR #9121, opened 2017-03-18
Update Qlogo.cn.xml

ivysrono:


ivysrono:
Only available in China even pass tests by Travis+Tor+SSL LABS

PR #9126, opened 2017-03-19
Update Cyberport.hk.xml

cschanaj:
Added a new rule for `^` due to the following observation. In this case HTTPSE cannot rewrite the `www` URL and the page is loaded over plaintext connection. Related: #8973.

Firefox on Debian:
`http://cyberport.hk/ (302)` -> `http://cyberport.hk/en (301)` -> `http://www.cyberport.hk/en (200)`

Firefox on Windows 10:
`http://cyberport.hk/ (302)` -> `http://cyberport.hk/en (200)`

### DIFF
--- a/src/chrome/content/rules/Cyberport.hk.xml
+++ b/src/chrome/content/rules/Cyberport.hk.xml
@@ -84,7 +84,7 @@
 	<target host="tenantbus.cyberport.hk" />
 	<target host="webmail.cyberport.hk" />
 
-	<rule from="^http://(www\.)?cyberport\.hk/"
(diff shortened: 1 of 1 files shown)


jeremyn:
For me on Ubuntu it redirects from http://cyberport.hk (`302`) to http://cyberport.hk/en (`200`), as you describe for Windows 10. But that's not really important for this change in my opinion.

Please:

- [ ] Change the `from` attribute in the complex `rule` to `^http://cyberport\.hk/`. This is easier to read, and `www` is caught by the general `rule` anyway.

jeremyn:
Thanks, merged.

PR #9137, opened 2017-03-19
[YourLogicalFallacyIs.com] Update and rename ruleset

Bisaloo:


jeremyn:
- [ ] Change `ruleset` `name` to `YourLogicalFallacyIs.com`, without `(partial)`

jeremyn:
FYI:

`(partial)` in the `name` is supposed to tell the user that the current page they are looking at is only partially covered. It doesn't necessarily mean that we have only covered some of the subdomains that exist.

A good use case for `(partial)` is:

```xml
<!--
	Invalid certificate:
		example.com
		www.example.com

-->
<ruleset name="Example.com (partial)">
	<target host="img.example.com" />

</ruleset>
```

Assume that there are `img` resources on every page but the user never directly visits `img`. If the `name` didn't have `(partial)`, then when the user looks at the list of active rulesets when they're on http://example.com/something, they might be confused about why the URL itself isn't HTTPS. But with `(partial)` they know that it's only partially covered, so there is no (or less) confusion.

A bad use case for `(partial)` is:

```xml
<!--
	Invalid certificate:
		beta.example.com

-->
<ruleset name="Example.com (partial)">
	<target host="example.com" />
	<target host="www.example.com" />

</ruleset>
```

Assume there is no overlap between `beta` and `^`/`www`. Either the user will either never see the active ruleset (on `beta`) or they will see the active ruleset and everything is covered (on `^`/`www`). They might be confused if they see `(partial)` when looking at `^`/`www` when everything is covered.

jeremyn:
Thanks, merged.

PR #9143, opened 2017-03-20
[GLAD.org] Remove wildcard, rename ruleset and remove platform

Bisaloo:


jeremyn:
Add `target`:
- [ ] https://secure.glad.org with `test` https://secure.glad.org/page/contribute/donate

Also:
- [ ] Remove `ward.glad.org` entirely. I don't get the `Secure connection failed` error but it seems to be for an internal service and I don't like including those.
- [ ] Remove `mail.glad.org:8443` as well, it's internal, and also I get connection reset with plain HTTP
- [ ] Remove `gateway.glad.org` as well: internal, and I get a time out with plain HTTP.
- [ ] Change `ruleset` `name` to `GLAD.org`

jeremyn:
Thanks, merged.

PR #9144, opened 2017-03-20
Create GLAAD.org ruleset

Bisaloo:


jeremyn:
- [ ] https://dev.glaad.org
 
Mark as `Time out`:
- [ ] https://files.glaad.org
- [ ] https://nytheater.glaad.org

Add `target`:
- [ ] https://twitter.glaad.org

jeremyn:
Thanks. The way I like to show the certificate chain problem is by including the subdomain in the `Invalid certificate` category but with an extra note, like this:

```xml
<!--
	Invalid certificate:
		allychallenge.glaad.org
		dev.glaad.org
		fightforourfuture.glaad.org
		peaceandacceptance.glaad.org
		spiritday.glaad.org
		twitter.glaad.org (incomplete certificate chain)
-->
```

because the problem is actually an invalid certificate but the extra note points out the problem, which may not be obvious when testing in a browser.

jeremyn:
Thanks, merged.

PR #9150, opened 2017-03-20
Create francetvinfo.fr.xml

Bisaloo:


Bisaloo:
Mixed content breaks the live feed so I guess it's better to put it as `mixedcontent` for now.

PR #9152, opened 2017-03-20
Create nsaneforums.com.xml

kronblase:
Mentioned in #3069

### DIFF
--- a/src/chrome/content/rules/nsaneforums.com.xml
+++ b/src/chrome/content/rules/nsaneforums.com.xml
@@ -2,6 +2,6 @@
         <target host="nsaneforums.com" />        
         <target host="www.nsaneforums.com" />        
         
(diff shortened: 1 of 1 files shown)


kronblase:
What have I done wrong?

Bisaloo:
Travis gives you a hint at what went wrong when you click on [Details](https://travis-ci.org/EFForg/https-everywhere/builds/213124467).

The part that went wrong here is:
> - Missing trailing slashes in 'from' or 'to' when specifying full hostnames: it should be "https://eff.org/" and not "https://eff.org".

You have the trailing slash in your `from` url but not in your `to`.

What you want here is:

```xml
         <rule from="^http://nsaneforums\.com/" to="https://www.nsaneforums.com/" />
```

PR #9160, opened 2017-03-20
Create hardwareluxx.ru.xml

kronblase:


src/chrome/content/rules/hardwareluxx.ru.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="hardwareluxx.ru">
```
J0WI: This site has mixed content issues, please add `platform="mixedcontent"`

kronblase:
I just realized that adding platform = "mixedcontent" breaks the script and it no longer works. I do not have enough knowledge to fix it.

src/chrome/content/rules/hardwareluxx.ru.xml
kronblase: I just realized that adding **platform = "mixedcontent"** breaks the script and it no longer works. I do not have enough knowledge to fix it.

J0WI:
What script?

kronblase:
hardwareluxx.ru.xml

J0WI:
And what does it *break*?

kronblase:
https not work anymore. Not on Tor Browser, not on Firefox.

kronblase:
Opened. He does not work

J0WI:
The ruleset will be shipped with the next release of HTTPS Everywhere.
Rulesets that have the mixedcontent flag will not be enabled by default in normal browsers when `extensions.https_everywhere.enable_mixed_rulesets` is set to `false`. This is to avoid broken pages. However, in Tor Browser it will be enabled.

kronblase:
I'm sorry but I'm a little confused. There will be two ruleset? One for tor browser with platform = "mixedcontent" and one for a normal browser without platform = "mixedcontent"? Simply now, my ruleset in the folder HTTPSEverywhereUserRules with the value platform = "mixedcontent" opens the http page of the site hardwareluxx.ru.

J0WI:
`platform="mixedcontent"` means that the rule is disabled by default, except in Tor Browser which ignores this attribute.

You can enable all `platform="mixedcontent"` rules in normal Firefox by setting `extensions.https_everywhere.enable_mixed_rulesets` to `true`.

kronblase:
But this addon is not only for Tor Browser. In a regular browser, these ruleset will simply not be active by default and the site will open as http.

J0WI:
Yes, that's true. Otherwise users will end up with a broken site, because of the mixed content blocker.
Users that don't care about this can flip the pref.

Issue #9145, opened 2017-03-20
Add rule for gov.au & scamwatch.gov.au


BK1233:
Can rules be added for www.gov.au and www.scamwatch.gov.au

PR #9162, opened 2017-03-21
Create SCAMWATCH.gov.au.xml

cschanaj:
Related: #9145


Bisaloo:
@J0WI, this was created in the wrong folder.

PR #9171, opened 2017-03-21
xn--90aijkdmaud0d.xn--p1ai

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/xn--90aijkdmaud0d.xn--p1ai.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="xn--90aijkdmaud0d.xn--p1ai">
```
J0WI: Please use a readable ruleset name (I think Unicode should work here).

PR #9178, opened 2017-03-21
Move SCAMWATCH.gov.au.xml to the rules folder

Bisaloo:
No changes to the file. I am simply moving it to the right location.


J0WI:
Oops, thanks!

PR #9180, opened 2017-03-22
Remove donate.crossorigin.me mismatch

technoboy10:
The subdomain no longer exists, so including it in the file doesn't really make sense. (Source: I maintain [crossorigin.me](https://github.com/technoboy10/crossorigin.me))


jeremyn:
Thanks, merged.

PR #9183, opened 2017-03-22
[Frugalware.org] Remove wildcards, rename and reactivate ruleset

Bisaloo:


Bisaloo:
just noticed it's in #3069

src/chrome/content/rules/Frugalware.org.xml
```diff
@@ -0,0 +1,38 @@
...
+	<securecookie host="^.+\.frugalware\.org$" name=".+" />
+
+	<rule from="^http://lists\.frugalware\.org/"
+		to="https://frugalware.org/mailman/listinfo/" />
+
+	<rule from="^http://packages\.frugalware\.org/"
+		to="https://frugalware.org/packages/" />
+
+	<rule from="^http://planet\.frugalware\.org/"
+		to="https://frugalware.org/" />
```
J0WI: This does not work for deep links.

src/chrome/content/rules/Frugalware.org.xml
```diff
@@ -0,0 +1,38 @@
...
+		dors.frugalware.org
+		git.frugalware.org
+		lists.frugalware.org
+		packages.frugalware.org
+		planet.frugalware.org
+		wiki.frugalware.org
+
+	Private subdomain:
+		webmail.frugalware.org
+
```
J0WI: There are also some mirrors like `www\d.frugalware.org` or `rsync\d.frugalware.org`

src/chrome/content/rules/Frugalware.org.xml
```diff
@@ -0,0 +1,38 @@
...
+		darcs.frugalware.org
+		dors.frugalware.org
+		git.frugalware.org
+		lists.frugalware.org
+		packages.frugalware.org
+		planet.frugalware.org
+		wiki.frugalware.org
+
+	Private subdomain:
+		webmail.frugalware.org
```
J0WI: As long as we can get a HTTP 200 response we should add it anyway. Maybe some people there are using HTTPS Everywhere too ;)

src/chrome/content/rules/Frugalware.org.xml
Bisaloo: What do you mean? Would you have a test url to explain?

src/chrome/content/rules/Frugalware.org.xml
Bisaloo: The argument against adding private subdomains was that we don't know if the rewrite rule will break what's behind the login page: maybe there is active mixed content.

src/chrome/content/rules/Frugalware.org.xml
J0WI: e.g. http://planet.frugalware.org/index.php?post_id=27 will not work with this rule.

src/chrome/content/rules/Frugalware.org.xml
J0WI: LGTM, apart from this

Bisaloo: It looks like none of them are working over https, would you accept something like:

```xml
	Non functional over https for various reasons:
		www\d.frugalware.org
		rsync\d.frugalware.org
```

?

J0WI: Yes, that's okay

PR #9184, opened 2017-03-22
Create ruleset Indapass.hu.xml (login provider)

hzsolt94:
Indapass.hu is a single sign on provider of the index.hu page group. They have working SSL, but don't use it by default. (Not even for logins.)


J0WI:
Thanks!

PR #9211, opened 2017-03-22
murman.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/murman.ru.xml
```diff
@@ -0,0 +1,87 @@
+<!--
+murman.ru ⁴
+www.murman.ru ⁴
```
J0WI: Please double check the list, some have only chain issues.

J0WI:
Your changes did not really address my feedback.

PR #9232, opened 2017-03-23
[Mailvelope.com] Add subdomains and rename ruleset

Bisaloo:


J0WI:
Mismatch on `fosdem2014.mailvelope.com` (broken because of HSTS).

PR #9248, opened 2017-03-26
[Roblox] Add some subdomains

mark-otaris:
The subdomains blog.roblox.com and corp.roblox.com support HTTPS, as do a few other subdomains.


J0WI:
Thanks!

PR #9249, opened 2017-03-26
Update Doodle.xml

J0WI:


J0WI:
@Hainish seems like this didn't trigger the labeler.

Hainish:
Not sure why the labeller missed that on first pass.  It may be that one of the other PRs that were checked in the same run of the labeller was malformed, thus causing the node script to crash.  Seems to have labelled it correctly on this pass.

jeremyn:
@J0WI Can this be rewritten without the wildcard `target`?

J0WI:
Nope, see comment about "Premium users"

jeremyn:
I think we should make this a `(partial)` ruleset and limit it to company domains, like I suggested for ning.com domains in PR https://github.com/EFForg/https-everywhere/pull/9536#issuecomment-341248319.

J0WI:
The site has changed a lot, so I think we can get rid of the wildcard now.

J0WI:
Done.

jeremyn:
Thanks, merged.

PR #9262, opened 2017-03-26
Add RateYourMusic.com.xml

afarrow:


src/chrome/content/rules/RateYourMusic.com.xml
```diff
@@ -0,0 +1,34 @@
...
+			Dutch
+			Finnish
+			Italian
+			Japanese
+			Portuguese
+			Spanish
+			Turkish
+-->
+<ruleset name="RateYourMusic.com">
+	<!-- request with www. is rewritten automatically to remove www. so we don't need a rule for that -->
```
J0WI: Please add www.rateyourmusic.com anyway.

afarrow: Ok, I did that. I also recently found some sites related to this entry:
- https://e.snmc.io/ is the CDN for rateyourmusic.com and the three sites below (The certificate is only valid for that subdomain of snmc.io)

The company that runs rateyourmusic.com is also creating three new related websites: 
- https://sonemic.com/
- https://glitchwave.com/
- https://cinemos.com/

My question is this: should I rewrite this file to more generally cover all these sites or should I split them out into 5 different files (the existing one, one for the CDN and one for each of the three other sites)? I apologize if this isn't the right place to ask this kind of question.

J0WI: Thanks!

Please create new files and separate PRs for those hosts. You can still making a comment to show the relationship between them.

PR #9266, opened 2017-03-27
Adding geekzone.co.nz CDN

freitasm:
Adding geekzone.co.nz subdomains cdn, images, avatar

### DIFF
--- a/src/chrome/content/rules/Geekzone.xml
+++ b/src/chrome/content/rules/Geekzone.xml
@@ -1,8 +1,7 @@
 <ruleset name="geekzone.co.nz (partial cdn)">
 	<target host="cdn.geekzone.co.nz" />
 	<target host="images.geekzone.co.nz" />
-	<target host="avatar.geekzone.co.nz" />
-
+	
 	<rule from="^http:"
 		to="https:" />
 </ruleset>
--- a/src/chrome/content/rules/Geekzone.xml
+++ b/src/chrome/content/rules/Geekzone.xml
@@ -1,6 +1,25 @@
+<!--
+        Served over HTTP:
+			feeds.geekzone.co.nz
+			m.geekzone.co.nz
+			price.geekzone.co.nz
+			www.geekzone.co.nz
+
+		Available over HTTPS (301 redirect from HTTP)
+			cdn.geekzone.co.nz
+			images.geekzone.co.nz
+			live.geekzone.co.nz
+			testip.geekzone.co.nz
+
+		Available over HTTPS only, no need for rewrite
+			api.geekzone.co.nz
+			metaweblog.geekzone.co.nz
+-->
 <ruleset name="geekzone.co.nz (partial cdn)">
 	<target host="cdn.geekzone.co.nz" />
 	<target host="images.geekzone.co.nz" />
+	<target host="live.geekzone.co.nz" />
+	<target host="testip.geekzone.co.nz" />
 	
 	<rule from="^http:"
 		to="https:" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Geekzone.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="geekzone.co.nz (partial cdn)">
+	<target host="cdn.geekzone.co.nz" />
+	<target host="images.geekzone.co.nz" />
```
J0WI: See also:
geekzone.co.nz
www.geekzone.co.nz
feeds.geekzone.co.nz
live.geekzone.co.nz
m.geekzone.co.nz
price.geekzone.co.nz

You should list the issues in a comment block on the top if they are not working with HTTPS.

freitasm: Thanks. Thinking of adding something like this:



Is it clear enough? 

Do we need to list specific pages that are available over HTTPS on the main domain, even though it is mainly served over HTTP?

J0WI: GitHub strips HTML comments from a comment when you don't format them as code ;)
If you are looking for examples you should check out our [CONTRIBUTING.md](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md).

freitasm: Ooops, true.

What I meant is if a domain is on both http and https (for example registration, login and private message pages) should I have rules for those or just list as a comment with something like "301 redirect for these pages)?

At the moment we just really want to enforce the target domains ...

Cheers

J0WI: If the 301 is from https to http you should add a comment that the host redirects on https.
If the redirect is from http to https you shoud just add it as target.

freitasm: Thanks. Yes, I can list live.geekzone.co.nz and testip.geekzone.co.nz as 301 for the whole sub-domain. But what about domains with only certain pages (login) redirect to https? Shall I list these URLs on target list or list in the comment only?

J0WI: You can use regex in the `rule` and `exclusion` tags. If it's not worth to make complex patterns, because deep links work only randomly over https, I would just declare it as non-working.

freitasm: Updated file now...

src/chrome/content/rules/Geekzone.xml
```diff
@@ -0,0 +1,26 @@
...
+			price.geekzone.co.nz
+			www.geekzone.co.nz
+
+		Available over HTTPS (301 redirect from HTTP)
+			cdn.geekzone.co.nz
+			images.geekzone.co.nz
+			live.geekzone.co.nz
+			testip.geekzone.co.nz
+
+		Available over HTTPS only, no need for rewrite
```
J0WI: This is no protection. Everyone can run a malicious service that responds on http.

freitasm: Thanks for your comment J0WI. I don't quite understand the context - is it related to the "Available over HTTPS only" line or the ones before?

J0WI: It's related to the `Available over HTTPS only` section. You should add them as targets.

freitasm:
Made some changes... How's this looking now?

J0WI:
LGTM, Thanks!

freitasm:
Thanks for this.

Cheers

PR #9267, opened 2017-03-27
[Mailinator.com] Update rewrite rules

Bisaloo:
I left the wildcard on purpose because I would argue that in this case, it is adequate.

- The website is secured by a wildcard certificate `*.mailinator.com`
- It looks like the website is configured so that any subdomain is a valid mirror of https://www.mailinator.com. For example, I tried https://idk.mailinator.com/ and https://thatsaninterestingbehaviour.mailinator.com/ and they work fine.

### DIFF
--- a/src/chrome/content/rules/Mailinator.com.xml
+++ b/src/chrome/content/rules/Mailinator.com.xml
@@ -7,20 +7,13 @@
 
 -->
 <ruleset name="Mailinator.com">
+
 	<target host="mailinator.com" />
 	<target host="*.mailinator.com" />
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Mailinator.com.xml
```diff
@@ -7,20 +7,13 @@
 
 -->
 <ruleset name="Mailinator.com">
+
 	<target host="mailinator.com" />
 	<target host="*.mailinator.com" />
```
J0WI: We require at least 3 test urls for wildcard targets.

PR #9268, opened 2017-03-27
atlas.sk new domains

MilanKral:


jeremyn:
Please:

- [ ] Remove [this comment](https://github.com/MilanKral/https-everywhere/blob/5940585d9d8beb953e83271c85ce8e9ea6fd71b4/src/chrome/content/rules/atlas.sk.xml#L5) for `(www.)atlas.sk`
- [ ] Sort all lists of subdomains according to the contributing guide [here](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#target-ordering)

jeremyn:
Thanks for the update, but it's still not sorted quite right. The idea is to read the domain components from right to left, so for example `katalog.menu.atlas.sk` and `mapa.menu.atlas.sk` should be next to each other under `Nonfunctional subdomains`, and the same with the `aktualne.atlas.sk` and `sport24.aktualne.atlas.sk` `target`s.

jeremyn:
Thanks, merged.

PR #9270, opened 2017-03-27
Pluska.sk

MilanKral:


src/chrome/content/rules/Pluska.sk.xml
```diff
@@ -0,0 +1,65 @@
...
+		to="https://www.pluska.sk/ipeknebyvanie/" />
+
+	<rule from="^http://emma\.pluska\.sk/"
+		to="https://www.pluska.sk/emma/" />
+
+	<rule from="^http://zdravie\.pluska\.sk/"
+		to="https://www.pluska.sk/izdravie/" />
+
+	<rule from="^http://izdravie\.sk/"
+		to="https://www.pluska.sk/izdravie/" />
```
J0WI: This does not work for deep links.

src/chrome/content/rules/Pluska.sk.xml
```diff
@@ -0,0 +1,65 @@
...
+
+	<rule from="^http://www\.izdravie\.sk/"
+		to="https://www.pluska.sk/izdravie/" />
```
J0WI: This does not work for deep links.

PR #9273, opened 2017-03-27
Fix rewrite rule for code.nsnam.org

labdsf:


jeremyn:
Thanks for your pull request. Why are you making this change? https://code.nsnam.org/ looks okay to me.

ilabdsf:
@jeremyn 
Because everything except index does not work.
For example, http://code.nsnam.org/ns-3-dev is ok, but https://code.nsnam.org/ns-3-dev/ is 404.

jeremyn:
I think I understand. In that case what you probably want are rewrite rules for `^http://code.nsnam.org/$` and `^http://code.nsnam.org/index.cgi/`, and any other URL that works, and no general rewrite rules. I don't think you should rewrite `^http://code.nsnam.org/$` to `^http://code.nsnam.org/index.cgi/`.

ilabdsf:
@jeremyn General rewrite rule is for www.nsnam.org.

labdsf:
@jeremyn Ping. I think I have fixed all the problems, is this change going to be merged?

jeremyn:
I looked at this again and think that, despite what I wrote in https://github.com/EFForg/https-everywhere/pull/9273#issuecomment-290124897, your changes seem safe.

Thanks again, merged.

labdsf:
@jeremyn It is probably because I force-pushed to this branch, so now you reviewed a different, updated version.

PR #9277, opened 2017-03-27
centrum.sk add domains

MilanKral:


src/chrome/content/rules/centrum.sk.xml
```diff
@@ -2,40 +2,44 @@
...
+	<target host="redir.centrum.sk" />
 	<target host="reg.centrum.sk" />
+	<target host="slovniky.centrum.sk" />
+	<target host="sutaze.centrum.sk" />
+	<target host="tv.centrum.sk" />
 	<target host="user.centrum.sk" />
+	<target host="zena.centrum.sk" />
+
+	<rule from="^http://centrum\.sk/"
+		to="https://www.centrum.sk/" />
```
J0WI: Please add a note why this redirect is required.

MilanKral: http://centrum.sk redirects to http://www.centrum.sk

centrum.sk doesn't support HTTPS

www.centrum.sk supports HTTPS

J0WI: Could you put this as a comment into the ruleset?

PR #9283, opened 2017-03-28
eztv.wf

galeksandrp:
Issue author: @galeksandrp


J0WI:
This seems related to `eztv.ag.xml` and `EZTV.xml`. Could you merge them or linking in comments?

PR #9287, opened 2017-03-28
volistob.ru

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/volistob.ru.xml
+++ b/src/chrome/content/rules/volistob.ru.xml
@@ -1,6 +1,8 @@
+<!--
+www.volistob.ru mismatch
+-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/volistob.ru.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="volistob.ru">
+	<target host="volistob.ru" />
+
+	<target host="www.volistob.ru" />
+
+	<rule from="^http://www\.volistob\.ru/"
```
J0WI: Please document the mismatch.

J0WI:
The cert only matches ^.

Issue #9166, opened 2017-03-21
Clean up duplicate-whitelist.txt


Bisaloo:
The file [`duplicate-whitelist.txt`](https://github.com/EFForg/https-everywhere/blob/master/utils/duplicate-whitelist.txt) has outdated entries. 

For example, [`elementaryos.org`](https://github.com/EFForg/https-everywhere/blob/master/utils/duplicate-whitelist.txt#L79) doesn't seem to be an issue anymore ([proof](https://www.eff.org/https-everywhere/atlas/?elementaryos.org)).

I think it's difficult to ask contributors to know about this file and to think about checking it each time they submit a PR.

The best solution in my opinion would be to have an automated way to update this file. Two options come to mind:
- a script that run on a regular basis and remove outdated entries
- a small addition to travis checks to ask the contributor to remove an entry if it is in the whitelist but that's not needed after their changes.

jeremyn:
We have a script that removes whitelist entries, [`utils/ruleset-coverage-whitelist-cleanup.sh`](https://github.com/EFForg/https-everywhere/blob/83d2966432c03a9e0df2f21acfa60e5754cf850f/utils/ruleset-coverage-whitelist-cleanup.sh). The normal process it that someone runs it every now and then, which you can see looking through [past issues and pull requests](https://github.com/EFForg/https-everywhere/search?q=whitelist&type=Issues&utf8=%E2%9C%93).

It's not really urgent to clean up the file. We should not confuse contributors by asking them to modify that file for each pull request.

Bisaloo:
I may be mistaken but it looks like this script only cleans up [`ruleset-coverage-whitelist.txt`](https://github.com/EFForg/https-everywhere/blob/83d2966432c03a9e0df2f21acfa60e5754cf850f/utils/ruleset-coverage-whitelist.txt), which is a different file from the one I am talking about : [`duplicate-whitelist.txt`](https://github.com/EFForg/https-everywhere/blob/master/utils/duplicate-whitelist.txt)

jeremyn:
Oh, you're right. I didn't know about that file. It looks like it's used [here](https://github.com/EFForg/https-everywhere/blob/445df151702555218cda70aab9aff092f1bfd214/utils/trivial-validate.py#L39).

We still shouldn't ask random contributors to look at the file. I agree there should be some cleanup process on that file, manually or automated.

PR #9306, opened 2017-03-28
Create duplicate-whitelist-cleanup.sh

Bisaloo:
Probably not the prettiest code but it apparently works. I'll submit a PR with the output as soon as possible.

### DIFF
--- a/utils/duplicate-whitelist-cleanup.sh
+++ b/utils/duplicate-whitelist-cleanup.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/bin/sh
 
 cat utils/duplicate-whitelist.txt | while read -r line; do
     REPEATS=`grep -F "target host=\"$line\"" src/chrome/content/rules/*.xml | cat | wc -l`
--- a/utils/duplicate-whitelist-cleanup.sh
+++ b/utils/duplicate-whitelist-cleanup.sh
@@ -5,4 +5,6 @@ for host in `cat utils/duplicate-whitelist.txt`; do
     if [ $REPEATS -gt 1 ]; then
         echo $host
     fi
-done >| utils/duplicate-whitelist.txt
+done >| temp.txt
+
+mv --force temp.txt utils/duplicate-whitelist.txt
--- a/utils/duplicate-whitelist-cleanup.sh
+++ b/utils/duplicate-whitelist-cleanup.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-cat utils/duplicate-whitelist.txt | while read -r host; do
+for host in `cat utils/duplicate-whitelist.txt`; do
     REPEATS=`grep -F "target host=\"$host\"" src/chrome/content/rules/*.xml | cat | wc -l`
     if [ "$REPEATS" -gt "1" ]; then
         echo "$host"
--- a/utils/duplicate-whitelist-cleanup.sh
+++ b/utils/duplicate-whitelist-cleanup.sh
@@ -1,3 +1,5 @@
+#! /bin/sh
+
 cat utils/duplicate-whitelist.txt | while read -r line; do
     REPEATS=`grep "target host=\"$line\"" src/chrome/content/rules/*.xml | wc -l`
     if [ "$REPEATS" -gt "1" ]; then
--- a/utils/duplicate-whitelist-cleanup.sh
+++ b/utils/duplicate-whitelist-cleanup.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
-cat utils/duplicate-whitelist.txt | while read -r line; do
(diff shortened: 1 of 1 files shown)


Bisaloo:
Fixes #9166.

Also @jeremyn.

Bisaloo:
Uhm, there is apparently still some issues. Eg. [`*.ebaystatic.com`](https://www.eff.org/https-everywhere/atlas/?ebaystatic.com).

J0WI:
*ping* @Hainish

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
```
Hainish: In other scripts we standardize on `#!/bin/sh` (without the space) so we probably want to preserve this.

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,8 @@
...
+
+cat utils/duplicate-whitelist.txt | while read -r line; do
```
Hainish: I prefer the following:

    for line in `cat utils/duplicate-whitelist.txt`; do

More succinct, no need to pipe

Hainish: I also might just call this `host` instead of `line` to be more specific

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,8 @@
...
+    REPEATS=`grep -F "target host=\"$line\"" src/chrome/content/rules/*.xml | cat | wc -l`
```
Hainish: No need for the `cat` pipe here, you can pipe directly to `wc -l`

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,8 @@
...
+    if [ "$REPEATS" -gt "1" ]; then
```
Hainish: No need for the surrounding quotes here.

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,8 @@
...
+        echo "$line"
```
Hainish: No need for the surrounding quotes here.

utils/duplicate-whitelist-cleanup.sh
Hainish: on either side of the `-gt`

Bisaloo:
I had to create a buffer file otherwise the loop reads the file while it is being overwritten. I do not recall having this problem with `while` but it's fine now.

Bisaloo:
I also assumed this script would be run from the root of the repo, is that correct? Or should I use `utils` as working directory?

utils/duplicate-whitelist-cleanup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+for host in `cat utils/duplicate-whitelist.txt`; do
+    REPEATS=`grep -F "target host=\"$host\"" src/chrome/content/rules/*.xml | wc -l`
+    if [ $REPEATS -gt 1 ]; then
+        echo $host
+    fi
+done > temp.txt
```
J0WI: I would use `mktemp`, but I think the temp file isn't necessary anyway.

Hainish:
The changes look good, other than what @J0WI points out - you should be able to pipe directly to the file as before, no?  If not, `mktemp` is preferred.

Also, I'll be squashing the PR down to one commit for brevity.  Just to lyk.

Bisaloo:
No, I tried and it doesn't work if I pipe it directly.

I think it is due to the fact that the shell creates a fork when you use a pipe, as I did earlier. This is explained in this [stackoverflow answer](http://stackoverflow.com/a/13764018).

Bisaloo:
I'd like to check some details before you merge this please. I think `while read` is actually the proper way to handle this kind of situation.

I think variables should also be quoted.

I am currently looking for references for those two points.

Hainish:
Since the output of the `wc -l` command is treated as an integer, the quotes are unnecessary (and stylistically odd IMO, since you're using the `-gt` comparison operator.)

Bisaloo:
> Since the output of the wc -l command is treated as an integer, the quotes are unnecessary (and stylistically odd IMO, since you're using the -gt comparison operator.)

Gotcha for this part but I should still keep them around $line, shouldn't I?

Hainish:
> Indeed, it should be an issue here because the file shouldn't have spaces but I think it's better if we stick to these good practices. What do you think?

If the file is expected to have one `host` per whitespace, I think it's more readable the way it is.

Hainish:
> Gotcha for this part but I should still keep them around $host, shouldn't I?

I'm fine with either way.

Bisaloo:
okay, in that case, you can go ahead and squash and merge then.

PR #9311, opened 2017-03-29
Update Nextag.xml

ivysrono:


ivysrono:
#8899 has removed Altmetric.com , why it show again.

src/chrome/content/rules/Nextag.xml
```diff
@@ -1,50 +1,52 @@
...
+	Different http/https content:
+		www.nextag.com
+		www42.nextag.com
+		merchants.nextag.com
+		travel.nextag.com
+
+		www.nextag.com.au
+		www.nextag.de
+		www.nextag.fr
+		www.nextag.it
```
J0WI: There is also ^ and a m. subdomain, but I could not reproduce "Different http/https content".

src/chrome/content/rules/Nextag.com.xml
```diff
@@ -4,24 +4,30 @@
...
+	Redirect to www:
+		^nextag.com
+
 	Show empty page:
 		www.nextag.com
 		www42.nextag.com
 
+		m.nextag.ca
 		www.nextag.com.au
 		www.nextag.fr
```
J0WI: I still have no empty page for them.

ivysrono: I get different results from ISP and Tor each time.
I'm not so sure now.

src/chrome/content/rules/Nextag.com.xml
```diff
@@ -0,0 +1,60 @@
+<!--
+	403:
+		merchants.nextag.com
+		travel.nextag.com
```
J0WI: Secure connection failed for me

src/chrome/content/rules/Nextag.com.xml
```diff
@@ -0,0 +1,60 @@
+<!--
+	403:
+		merchants.nextag.com
```
J0WI: WFM

src/chrome/content/rules/Nextag.com.xml
```diff
@@ -0,0 +1,60 @@
...
+		travel.nextag.com
+
+		www.nextag.de
+		www.nextag.fr
```
J0WI: both WFM

PR #9318, opened 2017-03-29
Update CERN.xml

ivysrono:


src/chrome/content/rules/CERN.xml
```diff
@@ -256,93 +234,55 @@
...
 
-	<!--	Protocol-relative redirect from user.web:
-							-->
-	<rule from="^https://home\.web\.cern\.ch/"
-		to="http://home.web.cern.ch/" downgrade="1" />
-
-		<test url="https://home.web.cern.ch/" />
-
-		<exclusion pattern="^http://home\.web\.cern\.ch/" />
+	<target host="home.web.cern.ch" />
```
J0WI: Could you also add `home.cern` and `www.home.cern` as targets?

Folant: @J0WI 
They are already here.
https://github.com/EFForg/https-everywhere/blob/eb3ea1ec764fd6fcc957a491d645f388a748545e/src/chrome/content/rules/CERN.xml#L250-L251

PR #9320, opened 2017-03-29
Update and rename Link-Plus-Catalog.xml to iii.com.xml

ivysrono:


src/chrome/content/rules/iii.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Too many redirects:
+		^	( equal to www )
```
J0WI: There is a bunch of additional subdomains.

src/chrome/content/rules/iii.com.xml
```diff
@@ -0,0 +1,37 @@
...
+	Failed:
+		csdirect.iii.com
+		linkplusaccess.iii.com
+-->
+<ruleset name="iii.com">
+	<target host="iii.com" />
+	<rule from="^http://iii\.com/" to="https://www.iii.com/" />
+
+	<target host="*.iii.com" />
+	<rule from="^http://(\w+)\.iii\.com/" to="https://$1.iii.com/" />
```
J0WI: A wildcard will break most subdomains.

PR #9322, opened 2017-03-29
Add domains of the MPI MOLGEN

paulmenzel:


paulmenzel:
I chose to extend the existing file, instead of creating a new one.

Additionally, I have no idea, if this could be grouped differently to improve the performance or if it affects the performance at all.

koops76:
Duplicate of #11225.

paulmenzel:
The question is, what is the duplicate of. ;-) Thank you for the update. The tests should now also pass.

PR #9338, opened 2017-03-30
[LastPass.com] Remove wildcard

Bisaloo:


jeremyn:
Sublist3r shows a lot of subdomains that aren't covered by the new ruleset that were covered by the old ruleset, like https://lp-push-server-900.lastpass.com and https://rodan.lastpass.com .

Bisaloo:
You're right. I don't know how I missed this...

I started updating ruleset but I realised it's going to be tedious to add all subdomains like 

- lp-push-server-\d.lastpass.com
- herndon-\d.lastpass.com
- do-\d.lastpass.com

without a left wildcard (since middle wildcards are discouraged).

Should I just close this PR?

jeremyn:
I don't think we should close this. The way to go is to use a wildcard and then use a regex in the rule. We did something like that in [360.cn.xml](https://github.com/EFForg/https-everywhere/blob/9698e64a2de7cf37509ab13ba9dcfd5bd4f84a95/src/chrome/content/rules/360.cn.xml#L98-L103).

jeremyn:
I'd like to move this pull request forward.

jeremyn:
- [ ] Fix misspelled `contet`
- [ ] Use `.+`/`.+` for the `securecookie`. (`.+` `host` should be equivalent to the current `^(?:.+\.)?lastpass\.com$` `host`.)

jeremyn:
Thanks, merged.

PR #9312, opened 2017-03-29
new-ruleset labeller

galeksandrp:
image

@Hainish

PR #9345, opened 2017-03-30
new-ruleset labeller 2

galeksandrp:
[Without fun pics for now](https://github.com/EFForg/https-everywhere/pull/9312)

### DIFF
--- a/utils/labeller/index.js
+++ b/utils/labeller/index.js
@@ -133,6 +133,7 @@ function github_process_prs(res, pr_cb){
             }
 
             var file_fetches = [];
+            var newRuleset;
 
             // Out of the list of files for this PR, figure out the minimum Alexa ranking for each
             _.each(files, function(file){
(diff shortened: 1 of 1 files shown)


utils/labeller/index.js
```diff
@@ -133,6 +133,7 @@ function github_process_prs(res, pr_cb){
             }
 
             var file_fetches = [];
+            var newRuleset;
```
Hainish: Please follow the naming convention of this file for variables (no camel case).

utils/labeller/index.js
```diff
@@ -154,9 +155,24 @@ function github_process_prs(res, pr_cb){
                     }
                   });
                 });
+                
+                if (!fs.existsSync('../../' + file.filename)) {
```
Hainish: In most contexts this is run, this won't exist.  The README suggests running this script from within `docker`, which lacks the context of the rest of the project.  Also, this will only work if you're doing a `git pull` before each run.  There must be a way within the github api to tell if a file is new or not.

PR #9348, opened 2017-03-31
[TheVerge.com] Update and rename ruleset

Bisaloo:
I don't think the other subdomains found by `Sublist3r` really exist or are still in use.


Folant:
#9326 😅

PR #9349, opened 2017-03-31
Update Craigslist.org.xml

ivysrono:


src/chrome/content/rules/Craigslist.org.xml
```diff
@@ -1,54 +1,38 @@
...
 		- Craigslist.ca.xml
 		- Craigslist.co.uk.xml
 
+	(May 2016) Craigslist only supports https on its .org, .ca, and .co.uk
+	domains. geo.craigslist.org redirects to regional sites by sending protocol-
+	relative redirects such as "//toronto.craigslist.ca". Unfortunately, it does this
+	even when the destination doesn't support https, such as helsinki.craigslist.fi,
+	making a downgrade to http couse travis error:
+	Too many redirects while fetching 'http://geo.craigslist.org/'
+	So exclusion only.
```
J0WI: I'm not sure if this is still true, this needs further testing.
`toronto.craigslist.ca` WFM on HTTPS and also `https://geo.craigslist.org/iso/fi` redirects correctly.

J0WI:
Maybe @jeremyn have time for a quick additional review on the `geo.craigslist.org` thing.

jeremyn:
With the new ruleset I get https://geo.craigslist.org/iso/fi -> (302) http://helsinki.craigslist.fi , which seems fine.

With the add-on disabled I get http://geo.craigslist.org/iso/fi -> (301) https://geo.craigslist.org/iso/fi, and with the old ruleset enabled the downgrade actually causes a loop, so removing the downgrade seems strictly better than not removing it.

J0WI:
Thanks!

PR #9352, opened 2017-03-31
Create EurActiv.com.xml

Bisaloo:


src/chrome/content/rules/EurActiv.com.xml
```diff
@@ -0,0 +1,31 @@
+<!--
+	No working URL known:
+		ads.euractiv.com
+		events.euractiv.com
```
J0WI: events. and pdf. are both 404/different content

PR #9354, opened 2017-03-31
Add mangagamer.com

nixxquality:


nixxquality:
This website is a web store so you should treat it with higher priority.

J0WI:
Could you please do the steps described here?
https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains

nixxquality:
There's a slight problem with that; it seems MangaGamer have a wildcard DNS record but only mangagamer.com and www.mangagamer.com are recognized in the SSL certificate.
Also, all the subdomains just point to the same front page.
What do you want me to do?

J0WI:
Okay, therefore I'm already happy when you document this behavior with a comment.

J0WI:
Thanks!

PR #9365, opened 2017-03-31
Update AJC.com.xml

afarrow:


afarrow:
Travis gave this error: ERROR src/chrome/content/rules/AJC.com.xml: Fetch error: http://www.ajc.com/ => https://www.ajc.com/: (7, 'Failed to connect to www.ajc.com port 443: Connection refused')

I think I get why this check failed: http://www.ajc.com:443 fails to load. But https://www.ajc.com loads just fine. What am I doing wrong?

src/chrome/content/rules/AJC.com.xml
```diff
@@ -1,27 +1,17 @@
...
-		to="https:" />
-
+<ruleset name="AJC.com (partial)">
+	<target host="ajc.com" />
+	<target host="www.ajc.com" />
+	<target host="myaccount.ajc.com" />
+	<target host="store.ajc.com" />
+	<target host="subscribe.ajc.com" />
+
+	<rule from="^https?://ajc\.com/" to="https://www.ajc.com/"/>
```
J0WI: We don't touch https

src/chrome/content/rules/AJC.com.xml
```diff
@@ -1,27 +1,17 @@
...
-
-	Problematic hosts in *ajc.com:
-
-		- *.blog *
-
-	* Mismatched
+	Other Atlanta Journal Constitution rulesets:
+		* MyAJC.com.xml
 
+	Numerous unsupported subdomains (>300 found by Sublist3r)
```
J0WI: Are all of them active? Is there any regularity?

src/chrome/content/rules/AJC.com.xml
afarrow: Huh. I can find a number of rulesets (78 cases in 69 rulesets) that have rule(s) that touch https. What I was solving was that when you request a page over https on www.ajc.com, it often seems to add https to other ajc.com links on that page regardless of whether the subdomain supports https. This can be a problem because it will create links like https://ajc.com/news/breaking-news/breaking-house-passes-obamacare-repeal-after-months-infighting/5oBa0i6IOViIvx2FPISdnJ/ which when clicked results in a invalid cert error. The rewriting fixes this by adding the www. which does support https. 

So if you don't know of any way to do this in an acceptable way then I guess I'll just close this PR because the other subdomains I added default to https and the real value added here is encrypting the www. requests. And then I guess we need to created an issue that fixes all the existing rulesets that have rules that touch https.

J0WI: We try to get rid of those cases. Could you make a comment about this behavior that we are aware of this bug when maintaining the ruleset?

afarrow: Sorry but I'm not understanding what you said fully. Do you want me to remove the behavior that touches https but put a comment in this ruleset noting the issue that prevents us from adding HTTPS Everywhere support to more parts of the site?

J0WI: I want you to put a comment in this ruleset to explain why touching https is required in this case.
Maybe you can also point to some example where we can reproduce the issue.

src/chrome/content/rules/AJC.com.xml
afarrow: I just re-ran Sublist3r and it found on 186 domains.  52 of those domains are of the form *.blog.ajc.com (50 of those seem to be active, 2 for testing purposes). Another probably 80 or so are domains that aren't meant to be directly accessed (ex: peer1-xen5.ajc.com ). Then they've got the rest which are subdomains like internships.ajc.com or teamplayer.ajc.com . So besides the blogs, there isn't much regularity.

Here's a link to that output from Sublist3r in case you're interested: https://pastebin.com/XL24jkSW

PR #9369, opened 2017-04-01
Update IMDB.xml

Bisaloo:
I voluntarily omitted all subdomains redirecting to `www.imdb.com`.

This domain configuration is far from being fully ready for HTTPS and hopefully, I will be able to come back to this ruleset to complete it eventually.

### DIFF
--- a/src/chrome/content/rules/IMDB.xml
+++ b/src/chrome/content/rules/IMDB.xml
@@ -28,12 +28,14 @@
 		i.imdb.com
 		posters.imdb.com
 
+	Different content http/https:
+		pro.imdb.com (test url: http://pro.imdb.com/signup/v4/help)
+
 -->
 <ruleset name="IMDB (partial)">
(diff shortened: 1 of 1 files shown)


Bisaloo:
Travis fails because `too many redirects`for https://pro.imdb.com/ and https://secure.imdb.com/. I tested with `curl` and indeed, I get 2-3 `302 redirects` but it is not a redirect loop.

What should I do here?

J0WI:
You can make an exclusion for `^https://pro.imdb.com/$` and add some better test urls.

Bisaloo:
My issue with this approach is that those subdomains apparently work fine and this is just a testing issue. Unless I misunderstood what you mean...

Bisaloo:
Ok, I completely removed `pro.imdb.com` because the behaviour differ between `http` and `https` for deep links.

PR #9370, opened 2017-04-01
Create Ofo.so.xml

cschanaj:


src/chrome/content/rules/Ofo.so.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+	Non-functional hosts
+		Couldn't connect to server:
+			 - cms.ofo.so
+			 - mail.ofo.so
+
+		4xx client error:
+			 - common.ofo.so
+			 - img.ofo.so
+			 - one.ofo.so
```
J0WI: Try custom test urls, like `https://common.ofo.so/app/`

cschanaj: done. thanks!

J0WI: `http://img.ofo.so/apk/`
You can easily search for test urls using search engines.

cschanaj: Sorry that I cannot find working URL(s) for `img` and `one` other than the one you have suggested. I tried Google `inurl:img.ofo.so`, `site:img.ofo.so` and even `img.ofo.so`, none of these give me a working URL. What kind of search query  did you use?

J0WI: I used "img.ofo.so" :P

PR #9373, opened 2017-04-01
Create DirectActionEverywhere.com.xml

Bisaloo:


src/chrome/content/rules/DirectActionEverywhere.com.xml
```diff
@@ -0,0 +1,25 @@
...
+	Refused:
+		discourse.directactioneverywhere.com
+
+-->
+<ruleset name="DirectActionEverywhere.com">
+
+	<target host="directactioneverywhere.com" />
+	<target host="www.directactioneverywhere.com" />
+
+	<target host="static.dxetech.org" />
```
J0WI: sendy.dxetech.org

Add comments about ^ and www

PR #9377, opened 2017-04-01
[Yale.edu] Add e360.yale.edu

Bisaloo:
**This is not a complete update of the `Yale.edu` ruleset. I simply noticed a subdomain that can and therefore should be secured.**

### DIFF
--- a/src/chrome/content/rules/Yale_University.xml
+++ b/src/chrome/content/rules/Yale_University.xml
@@ -31,8 +31,8 @@
 -->
 <ruleset name="Yale.edu (partial)">
 
-	<!--	Direct rewrites:
-				-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Yale_University.xml
```diff
@@ -31,8 +31,8 @@
 -->
 <ruleset name="Yale.edu (partial)">
 
-	<!--	Direct rewrites:
-				-->
+	<target host="e360.yale.edu" />
```
J0WI: Please put it in alphabetical order.

Bisaloo: It is? The next block are subdomains of `law.yale.edu`

PR #9399, opened 2017-04-03
Create nytcn.me.xml

ivysrono:


src/chrome/content/rules/nytcn.me.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	For other New York Times coverage, see NYTimes.xml.
```
J0WI: This rule needs to be added to the list in NYTimes.xml

PR #9400, opened 2017-04-03
Update New_York_Times.com.xml

ivysrono:


src/chrome/content/rules/NewYorkTimes.com.xml
```diff
@@ -1,16 +1,11 @@
...
-	<!--	Complications:
-				-->
+<ruleset name="NewYorkTimes.com">
 	<target host="newyorktimes.com" />
 	<target host="www.newyorktimes.com" />
 
-
-	<rule from="^http://(?:www\.)?newyorktimes\.com/"
+	<rule from="^http://(www\.)?newyorktimes\.com/"
 		to="https://www.nytimes.com/" />
```
J0WI: Could you add a note about the http redirect?

PR #9409, opened 2017-04-04
Update AnarchistNews.xml

jtracey:


src/chrome/content/rules/AnarchistNews.xml
```diff
@@ -1,4 +1,4 @@
-<ruleset name="AnarchistNews.org" default_off="self-signed">
+<ruleset name="AnarchistNews.org">
   <target host="www.anarchistnews.org" />
   <target host="anarchistnews.org" />
```
J0WI: @jtracey please also add `podcast.anarchistnews.org` and a comment about the mismatch on `action.anarchistnews.org`.

J0WI:
Thanks!

PR #9417, opened 2017-04-04
Update 126.net.xml

ivysrono:


src/chrome/content/rules/126.net.xml
```diff
@@ -1,15 +1,19 @@
 <!--
-	Only avaiable in China:	126.net_CN.xml
```
J0WI: Why did you remove that file?

src/chrome/content/rules/126.net.xml
```diff
@@ -1,15 +1,19 @@
...
 
+	Incomplete cert chain:
+		l.bst.126.net
+-->
 <ruleset name="126.net">
-
 	<target host="img3.126.net" />
 		<test url="http://img3.126.net/kaola/150512/js/jquery-1.4.2.js" />
 	<target host="easyread.ph.126.net" />
-		<test url="http://easyread.ph.126.net/0jkjwGfV7I4EJXnYlRBkVg==/7917077456266817441.jpg" />
```
J0WI: Why did you remove this test url? We only have an `AccessDenied` on this host.

src/chrome/content/rules/126.net.xml
ivysrono: Sorry , I got mismatched 5 days ago in China. My mistake.

ivysrono:
#9471

PR #9426, opened 2017-04-04
Update INYT.com.xml

ivysrono:


src/chrome/content/rules/INYT.com.xml
```diff
@@ -1,57 +1,44 @@
...
-
 	<securecookie host="^\." name="^(?:__utm|_ga$)" />
 	<securecookie host="^\w" name="." />
 
-
 	<!--	Redirect keeps args but
-		not forward slash:
-					-->
+		not forward slash:	-->
 	<rule from="^http://(?:www\.)?inyt\.com/+"
```
J0WI: Could you also remove the non-capturing group?

J0WI: Maybe we need to get rid of this redirect completely. It seems only to work inside `http://www.inyt.com/subscriptions/`, but then you have mixed content issues on `global.nytimes.com`.

PR #9436, opened 2017-04-05
Add DaButtonFactory.com

PaulCapron:
[DaButtonFactory.com](https://DaButtonFactory.com) is an online button maker. It offers a simple REST API for button images generation, where an image is returned in response to an HTTP GET request. (The button settings, such as the text, size, and colors, are included in the requested URL query string.)

This API/hotlinking facility is used on thousands of sites around the web.
The use of HTTPS is recommended, and the site and its API are fully available over HTTPS. 
However, since Da Button Factory uses a shared SSL certificate (provided by Cloudflare), we do not redirect HTTP requests to HTTPS, to not mess with old browsers not supporting SNI.
Since the HTTPS everywhere plugin is only available for browsers supporting SNI, it is OK to add a ruleset here.

### DIFF
--- a/src/chrome/content/rules/DaButtonFactory.com.xml
+++ b/src/chrome/content/rules/DaButtonFactory.com.xml
@@ -1,0 +1,9 @@
+<ruleset name="DaButtonFactory.com">
+	<target host="dabuttonfactory.com" />
+	<target host="www.dabuttonfactory.com" />
+
+	<!-- For Cloudflare cookies. DaButtonFactory itself does not set any. -->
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks!
You may use preloaded HSTS for this service. This feature is only implemented by clients, that support SNI anyway, so they will connect over HTTPS without any redirect. For the submission a redirect is required, but maybe you can use some user agent specific rewrites for this.

PaulCapron:
Thanks for the suggestion and the merging!

PR #9437, opened 2017-04-05
Create NYT_Style.com.xml

ivysrono:


src/chrome/content/rules/NYT_Style.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	For other New York Times coverage, see NYTimes.xml.
+
+	Not exist:
+		^
+		www
+
+	Timeout by travis:
+		m.cn.nytstyle.com
```
J0WI: Maybe a custom test url works?

J0WI:
I have some mixed content issues. Can you confirm that this breaks nothing?

ivysrono:
@J0WI Where are the mixedcontent issues？

J0WI:
on $ from graphics8.nytimes.com and js.nyt.com

PR #9438, opened 2017-04-05
[Mastodon.social] New ruleset

lucaspetter:
Site serves HSTS header with 1 year `max-age` and `includeSubdomains`, but no `preload`.

### DIFF
--- a/src/chrome/content/rules/Mastodon.social.xml
+++ b/src/chrome/content/rules/Mastodon.social.xml
@@ -3,6 +3,8 @@
 	<target host="*.mastodon.social" />
 
 	<test url="http://assets.mastodon.social/assets/fluffy-elephant-friend-6b47d8e924332955795ff4b2d8fc446437d26b28bfc67d6be2a4d88995ab2c1f.png" />
+	<test url="http://assets.mastodon.social/assets/logo-5701d8d94bca66668f49a2803c984e960dfe0bf93a71f2d86be9455e66505cee.png" />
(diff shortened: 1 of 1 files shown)


J0WI:
We require at least 3 test urls for wildcards.

lucaspetter:
I've added 2 more test urls.

J0WI:
Okay, that's valid now. But I was more thinking about three different subdomains as test urls.

lucaspetter:
I was going by how the site serves HSTS with `includeSubdomains`, therefore we might as well have a wildcard target because all subdomains will be covered anyway.

It looks like the site has only 2 host names: `mastodon.social` and `assets.mastodon.social`. Do you think I should just replace the wildcard with `assets.mastodon.social`?

J0WI:
e.g. `www.mastodon.social` WFM.

Could you add a comment about the HSTS with includeSubdomains into the ruleset?

lucaspetter:
Okay, but are you saying I should keep the wildcard and add a test URL for `www.`? Or remove the wildcard and add targets for `www.` and `assets.`?

J0WI:
I think both variants are valid, but personally I prefer the variant without a wildcard.

lucaspetter:
Thanks for clarifying. Based on the docs, and due to the site's HSTS `includeSubdomains`, I've kept the wildcard and added a test URL for `www.mastodon.social` as well as a comment about the HSTS.

J0WI:
Thanks!

PR #9451, opened 2017-04-06
Update NYTimes.xml

ivysrono:


src/chrome/content/rules/NYTimes-Problematic.xml
```diff
@@ -0,0 +1,114 @@
...
+	<!--	Redirect drops args and forward slash:	-->
+	<target host="travel.nytimes.com" />
+	<rule from="^http://travel\.nytimes\.com/(?:$|\?.*)"
+		to="https://www.nytimes.com/section/travel" />
+		<test url="http://travel.nytimes.com/?" />
+		<test url="http://travel.nytimes.com//?" />
+
+	<rule from="^http://travel\.nytimes\.com/"
+		to="https://www.nytimes.com/" />
+		<test url="http://travel.nytimes.com/section/realestate" />
```
J0WI: Could you also add a test url with the `travel.nytimes.com/YYYY/MM/DD/opinion/\w+.html` scheme?

PR #9453, opened 2017-04-06
Update OpenBSD.org.xml

massoc:
adopt eff style guidelines (alphabetical order etc), update working hosts for many new *openbsd.org subdomains

### DIFF
--- a/src/chrome/content/rules/OpenBSD.org.xml
+++ b/src/chrome/content/rules/OpenBSD.org.xml
@@ -1,26 +1,26 @@
-<!--
-	Nonfunctional hosts in *openbsd.org:
-
-		- cvsweb ²
-		- ftp.eu, ftp2.eu, &c ¹²
-		- man ²
-
-	¹ Mismatched
-	² Refused
-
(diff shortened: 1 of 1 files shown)


massoc:
I should note the reason for removing the ¹ and ² style comment notation, I was getting complaints from fetch-test.sh about encoding "ERROR Exception parsing testmeOpenBSD.org.xml: Input is not proper UTF-8, indicate encoding !" depending on which text editor I used.

So I thought plain ASCII comments should be easier for any editor to handle.

J0WI:
Thanks!

massoc:
Hi,

Update for OpenBSD.org.xml: man.openbsd.org and ftp2.eu.openbsd.org have been enabled for TLS, firmware.openbsd.org remains with a mismatched certificate.

Tested with curl --tlsv1.2 $host, Chrome 59.0.3071.86 and Firefox 53.0.3:
General browsing and file download work fine for:
https://man.openbsd.org
https://ftp2.eu.openbsd.org

I have attached a pull request with the update, the pull request is at: #10375

PR #9461, opened 2017-04-07
Update derstandard.at.xml

aried3r:
I'm not sure what exactly is meant by the mismatches here, but from my testing the whole page of der standard.at can be accessed via https now. I'm just not sure how to handle https://dst.at (see also https://www.ssllabs.com/ssltest/analyze.html?d=dst.at).


J0WI:
Thanks for the update!
The mismatch comment lists subdomains that serves a mismatching cert. dst.at has the same error.

PR #9496, opened 2017-04-11
Imhd.sk

MilanKral:


src/chrome/content/rules/Imhd.sk.xml
```diff
@@ -0,0 +1,57 @@
+<!--
+	imhd.sk public transport
+-->
+<ruleset name="Imhd.sk">
+	<target host="imhd.sk" />
+	<target host="*.imhd.sk" />
+	<target host="m.*.imhd.sk" />
```
J0WI: Can we avoid wildcard targets/test urls here?

PR #9497, opened 2017-04-11
[MoS.org] Merge, update, and rename ruleset

Bisaloo:


src/chrome/content/rules/MoS.org.xml
```diff
@@ -0,0 +1,45 @@
...
+	Time out:
+		exhibits.mos.org
+		library.mos.org
+
+	Private subdomain:
+		kronos.mos.org
+		login.mos.org
+
+	Too many redirects:
+		legacy.mos.org
```
J0WI: Use a custom test url.

src/chrome/content/rules/MoS.org.xml
```diff
@@ -0,0 +1,45 @@
...
+		starwars.mos.org
+		cst.mos.org
```
J0WI: Both redirect back to HTTP.

src/chrome/content/rules/MoS.org.xml
Bisaloo: I am not sure what you mean here. I check with `curl -v -L`and actually, all 3 of those subdomains redirect to `http`.

PR #9508, opened 2017-04-12
[350.org] Remove wildcard, update rules and rename ruleset

Bisaloo:


Bisaloo:
- Why can't travis resolve http://pacific.350.org? It works fine for me. (I removed it anyways because it redirects to http://world.350.org which doesn't support https).

- What should I do with `go.350.org`? It already shows up in `Bitly_branded_short_domains.xml`. I think it would make more sense to group all subdomains related to one domain in a single ruleset.

J0WI:
I can confirm that `pacific.350.org` works. Could you try again after rebasing against current master?

Bisaloo:
@J0WI, I don't understand, I don't see any conflicts for this file.

J0WI:
The test script is taken from your branch, so your tests have not been updated since Apr 12. Rebasing your branch uses the current test scripts from our master, this might fix some strange issues.

PR #9510, opened 2017-04-12
Create PDFMerge.com.xml

cschanaj:
Related: #3069


src/chrome/content/rules/PDFMerge.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	Non-functional hosts
+		SSL peer certificate was not OK:
```
J0WI: They have a wildcard DNS, you can use whateverfoobar.pdfmerge.com

cschanaj: I have removed the comment accordingly. May I know how to test for this in general? thanks.

PR #9526, opened 2017-04-13
Create TestPolitique.fr.xml

Bisaloo:


jeremyn:
- [ ] https://www.testpolitique.fr should be `Invalid certificate`, not `Secure connection failed`

PR #9533, opened 2017-04-14
Create CitizenPost.fr.xml

Bisaloo:


J0WI:
What about the mixed content issues here?

Bisaloo:
I haven't observed any missing functionality with `https`.

I don't think it does hinder the site look too much but if you don't feel good about it, I can always add `platform='mixedcontent'`.

PR #9536, opened 2017-04-14
Update ning.com rulesets

J0WI:


jeremyn:
We shouldn't use a wildcard `target` here. Some of the user domains have mixed content blocking problems, for example https://slimoco.ning.com .

I would combine both rulesets into a single `Ning.com (partial)` ruleset that only covers the company domains like https://www.ning.com and https://blog.ning.com.

J0WI:
I didn't notice major mixed content issues on user domains. Do you have more specific examples?

jeremyn:
https://slimoco.ning.com/ looks okay to me now, but I remember it being badly broken when I checked it before.

* https://beachescyclingclub.ning.com/ has a broken listing under "BCC ON STRAVA" in the lower right.
* https://earlyguitar.ning.com/ has a broken photo display in the middle top.
* https://erythromelalgiawarriors.ning.com/ redirects to HTTP.
* https://smaspace.ning.com/ has a missing "G+" button in the top right.

I'm sure you can find others. There are many `*.ning.com` domains and I assume users can create new ones easily.

J0WI:
Are you fine with moving the wildcard/customer targets to the `mixedcontent` rule?

jeremyn:
Yes, it should be okay to move the wildcard/customer `target`s to the `mixedcontent` ruleset, because it's unlikely that some specific customer domain has some other HTTPS problem.

J0WI:
Okay, I've rebased this PR.

jeremyn:
`Ning.xml`:

Remove these, so they are covered by `Ning-falsemixed.xml`:
- [ ] https://build.ning.com (and add to the top comment `(mixed content, secured by us)`)
- [ ] https://developer.ning.com (and add to the top comment `(mixed content, secured by us)`)
- [ ] https://socialnetworkmain.ning.com (this looks like a customer domain)

Also:
- [ ] https://static.ning.com returns a `404` and needs a `test`
- [ ] Add `target` for https://cdn.ning.com with a `test` for http://cdn.ning.com/wp-content/themes/ning/assets/img/ui/black/s-icons/custom.svg
- [ ] Delete the `securecookie` or restrict it to the non-`^` `target`s in this ruleset

`Ning-falsemixed.xml`:

Remove `exclusion`, to match the changes to `Ning.xml`:
- [ ] https://build.ning.com (and add `test` under `covered by this ruleset`)
- [ ] https://developer.ning.com (and add `test` under `covered by this ruleset`)
- [ ] https://socialnetworkmain.ning.com

Also:
- [ ] Change `Nonfunctional subdomain` to `Subdomain times out`, or otherwise indicate why it doesn't work, for https://status.ning.com
- [ ] Move the `test` for `www` above `about`
- [ ] Change `Subdomains controlled by customers` to `Examples of subdomains controlled by customers`

Also `smaspace.ning.com` looks abandoned ("Domain: smaspace.ning.com is free for sale") so maybe you want to swap that out for another active customer domain.

J0WI:
Done.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/9536/commits/5a46165a130677795fc8d184a69d72b3d7dbb77f.

J0WI:
I didn't add `socialnetworkmain.ning.com` because it's currently down. Comments have been updated.

jeremyn:
I'm not sure what you mean about https://socialnetworkmain.ning.com. It shouldn't be excluded in `Ning-falsemixed.xml`, because it's not included in `Ning.xml` anymore. That's the remaining item in the list.

I'm getting a 302 redirect to a 500 now for https://socialnetworkmain.ning.com , but I remember that at some point I got the sort of "This domain is available" landing page like you see at for example http://asdf.ning.com/ , so I assume http://socialnetworkmain.ning.com isn't associated with the Ning company but is just some customer domain.

jeremyn:
Thanks, merged.

PR #9543, opened 2017-04-15
Update and rename AmazonAWS.xml to AmazonWebServices.xml

ivysrono:


ivysrono:
@J0WI The regex on amazon is so complicated.
Sorry, I have no time to do more.

PR #9546, opened 2017-04-15
[GPGTools.org] Remove wildcard and rename ruleset

Bisaloo:


src/chrome/content/rules/GPGTools.org.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+	Different content http/https:
+		support.gpgtools.org
```
J0WI: Can you give an example?

Bisaloo: Uhm, no, you're right. For some reason, I thought https://support.gpgtools.org didn't redirect properly. I will update it.

PR #9549, opened 2017-04-15
OpenX.net:

numismatika:
- switch to wildcard

### DIFF
--- a/src/chrome/content/rules/OpenX.net.xml
+++ b/src/chrome/content/rules/OpenX.net.xml
@@ -4,9 +4,7 @@
 
 	Nonfunctional hosts in *openx.net:
 
-		- (www.)? *
-
-	* Shows blog.openx.org; self-signed, CN: *.openx.com
+		- (www.) -  self-signed & expired & mismatched, CN: *.openx.com
 
 
 	Insecure cookies are set for these domains: ᶜ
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
 -->
```
J0WI: The comments are not up to date.

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
...
 <ruleset name="OpenX.net (partial)">
```
J0WI: This rule is no longer partial.

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
...
 
-	<target host="bid.openx.net" />
-	<target host="eu-u.openx.net" />
-	<target host="gb-d.openx.net" />
-	<target host="netline-d.openx.net" />
-	<target host="r.openx.net" />
-	<target host="u.openx.net" />
-	<target host="us-ads.openx.net" />
-	<target host="us-u.openx.net" />
+	<target host="*.openx.net" />
```
J0WI: Please add a hint why a wildcard is required here.

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
...
 
 		<!--	Set cookies without Secure:
 							-->
-		<!--test url="http://u.openx.net/w/1.0/sc?r=http%3A%2F%2Fox-d.openxadexchange.com%2Fw%2F1.0%2Frs%3Fcc%3D1%26as%3D1%26cb%3D1" /-->
-		<!--test url="http://us-u.openx.net/w/1.0/pd?plm=1&amp;ph=00000000-0000-0000-0000-000000000000" /-->
-
+		<test url="http://u.openx.net/w/1.0/sc?r=http%3A%2F%2Fox-d.openxadexchange.com%2Fw%2F1.0%2Frs%3Fcc%3D1%26as%3D1%26cb%3D1" />
+		<test url="http://us-u.openx.net/w/1.0/pd?plm=1&amp;ph=00000000-0000-0000-0000-000000000000" />
+		<test url="http://bid.openx.net/" />
+		<test url="http://www.openx.net/" />
```
J0WI: Can you add all previous targets as additional test urls?

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
...
 
 	<!--	Not secured by server:
 					-->
 	<!--securecookie host="^\.openx\.net$" name="^(?:i|pd)$" /-->
 
 	<securecookie host="." name="." />
```
J0WI: Please use `.+`

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -18,29 +18,22 @@
...
 
-
 	<rule from="^http:"
 		to="https:" />
-
+			<exclusion pattern="^http://www\.openx\.net/" />
```
J0WI: This should be before the redirect rule.

src/chrome/content/rules/OpenX.net.xml
numismatika: The question is, when does "partial" need to be part of the name and when is it not required?
I could argue that since subdomains like www do not work, it is still partial.

J0WI: We had a discussion about this somewhere, but I cannot find that issue anymore.
When major parts of the site cannot be secured, or only the resources that are fetched in background (so the https:// is not visible to the user), it makes sence to have the `(partial)` flag.
In this case the `www` host has no content, so it's trivial. In other cases `www` is the most important host.

src/chrome/content/rules/OpenX.net.xml
J0WI: Please update the comment for `(www.)?openx.net`.

src/chrome/content/rules/OpenX.net.xml
```diff
@@ -15,31 +15,27 @@
...
 
 	<!--	Not secured by server:
 					-->
 	<!--securecookie host="^\.openx\.net$" name="^(?:i|pd)$" /-->
 
-	<securecookie host="." name="." />
-
+	<securecookie host=".+" name=".+" />
 
+	<exclusion pattern="^http://www\.openx\.net/" />
```
J0WI: You could add a redirect for `(www.)?openx.net` to `openx.com`

numismatika: It is still self-signed and expired.
I will add a comment about it.

PR #9551, opened 2017-04-15
add rule for Blizzard.com

numismatika:


src/chrome/content/rules/Blizzard.com.xml
```diff
@@ -0,0 +1,125 @@
...
+			- us.version.blizzard.com
+			- vpn.connect.blizzard.com
+			- vcse.blizzard.com
+			- vcse03.blizzard.com
+			- vcse04.blizzard.com
+			- vcse05.blizzard.com
+			- vcse06.blizzard.com
+
+-->
+<ruleset name="Blizzard.com">
```
J0WI: This rule is only partial. Major party of the site cannot be secured.

numismatika: yea unfortunately, but we haven't exactly enforced that in the past for all pull requests.

J0WI: What I wanted to say is: could you add the `(partial)` string to the ruleset name?

PR #9554, opened 2017-04-16
Add ruleset for parkrun.com and subdomains

alexmuller:
parkrun.org.uk is fetching one tiny image from parkrun.com which is causing mixed-content warnings, so this commit fixes that.

parkrun.com has a valid certificate for parkrun.com and *.parkrun.com.

### DIFF
--- a/src/chrome/content/rules/Parkrun.com.xml
+++ b/src/chrome/content/rules/Parkrun.com.xml
@@ -1,0 +1,15 @@
+<ruleset name="parkrun.com">
+
+    <target host="parkrun.com" />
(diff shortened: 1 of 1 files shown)


alexmuller:
@J0WI Added all except the `status` subdomain because it doesn't have a valid cert. Thanks!

J0WI:
@alexmuller could you add a comment about the mismatch on status?

alexmuller:
@J0WI Sure, done! Apparently it's no longer a mismatch, it now doesn't connect for me.

J0WI:
Thanks!

PR #9556, opened 2017-04-16
[NWF.org] Update subdomains and reactivate ruleset

Bisaloo:


Bisaloo:
I know there is a mixed content warning on https://www.nwf.org but this URL (with `https`) is sometimes used in links from itself and on google.

PR #9562, opened 2017-04-16
Create LaPoste.fr.xml

Bisaloo:
`Sublist3r` finds more than 800 subdomains so I obviously didn't add them all.

I simply browsed through the front page of https://www.laposte.fr and added all subdomains I could find there.


J0WI:
Please try to add some more and mark the rule as 'partial'.

PR #9563, opened 2017-04-16
Tagesanzeiger.ch:

numismatika:
- add more subdomains

### DIFF
--- a/src/chrome/content/rules/Tagesanzeiger.ch.xml
+++ b/src/chrome/content/rules/Tagesanzeiger.ch.xml
@@ -1,38 +1,299 @@
 <!--
-	Nonfunctional hosts:
+		Most of the subdomains are selected top stories, not all of them are properly secured with a valid cert though
 
-		- blog ¹
-		- epaper ¹
-		- jobs ²
-		- marktplatz ¹
-		- weiterbildung ³
-		- www ⁴
+	Cert expired:
+		- staging.mobile.tagesanzeiger.ch
+		
+	Cert mismatch:
+		- www.1zu12.tagesanzeiger.ch
+		- www.abo.tagesanzeiger.ch
+		- www.abseits.tagesanzeiger.ch
+		- www.actioncam.tagesanzeiger.ch
+		- www.akwdebatte.tagesanzeiger.ch
+		- www.alttrifftjung.tagesanzeiger.ch
+		- www.antarktis.tagesanzeiger.ch
+		- www.arbeit.tagesanzeiger.ch
+		- www.arboretum.tagesanzeiger.ch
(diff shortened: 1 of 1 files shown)


J0WI:
I think this host is similar to #8548: They just use a wildcard DNS and sub domains for some top stories.

numismatika:
It id the same media holding so very likely.

J0WI:
Could you add a hint about this behaviour in the comments?

J0WI:
@numismatika I agree with this. But now you could also add `foo.tagesanzeiger.ch` and `bar.tagesanzeiger.ch` to the list. I think adding this comment is the best we can do here.

numismatika:
@J0WI agreed.

src/chrome/content/rules/Tagesanzeiger.ch.xml
```diff
@@ -1,38 +1,318 @@
...
+		- longform.tagesanzeiger.ch
+		- marktplatz.tagesanzeiger.ch
+		- seminare.tagesanzeiger.ch
+		- tippinho.tagesanzeiger.ch
+		- webspecial.tagesanzeiger.ch
+		- editor.webspecial.tagesanzeiger.ch
+		- weiterbildung.tagesanzeiger.ch
 
-	Fully covered hosts:
+	Redirect to plain:
```
J0WI: This section is now empty.

src/chrome/content/rules/Tagesanzeiger.ch.xml
```diff
@@ -1,38 +1,318 @@
...
+	<target host="wwwt.tagesanzeiger.ch" />
+	<target host="wwww.tagesanzeiger.ch" />
+	<target host="wwwww.tagesanzeiger.ch" />
+	<target host="zalando.tagesanzeiger.ch" />
+	<target host="zomm.tagesanzeiger.ch" />
+	<target host="zoom.tagesanzeiger.ch" />
+	
+	
+	
+	<!-- TEST -->
```
J0WI: Please clean this up

src/chrome/content/rules/Tagesanzeiger.ch.xml
```diff
@@ -1,38 +1,318 @@
...
+	<target host="leer.tagesanzeiger.ch"/>
+	<target host="leichtathletik.tagesanzeiger.ch"/>
+	<target host="lichtbild.tagesanzeiger.ch"/>
+	<target host="lift.tagesanzeiger.ch"/>
+	<target host="listen.tagesanzeiger.ch"/>
+	<target host="m.tagesanzeiger.ch"/>
+	
+	
+	
+	<!-- TEST2 -->
```
J0WI: Please clean this up

PR #9567, opened 2017-04-16
Wemfbox.ch:

numismatika:
- add more subdomains
- remove wildcard


numismatika:

PR #9568, opened 2017-04-16
add rule for Eyeonx.ch

numismatika:


src/chrome/content/rules/eyeonx.ch.xml
```diff
@@ -0,0 +1,30 @@
+<!--
+		Cert expired:
+		-	event.eyeonx.ch
```
J0WI: See also `converter.eyeonx.ch`

numismatika: every subdomain of eyeonx.ch is answered for by a wildcard dns on deliberate typos.
What about exactly this domain?

J0WI:
Referring to Travis there are some more expired certs.

numismatika:

PR #9570, opened 2017-04-16
add rule for Mnet-Ad.net

numismatika:


J0WI:
Do you have a valid test url for this?

numismatika:
Should be part of Tagesanzeiger.ch iirc.

PR #9574, opened 2017-04-17
[CharityWater.org] Remove wildcard and increase coverage

Bisaloo:


src/chrome/content/rules/CharityWater.org.xml
```diff
@@ -0,0 +1,48 @@
...
+		maps.charitywater.org
+		sandbox.charitywater.org
+		staging.charitywater.org
+		wazi.charitywater.org
+		wazi-d.charitywater.org
+
+	Private subdomain:
+		sensors.charitywater.org
+		www-d.charitywater.org
+
```
J0WI: cdn.charitywater.org

PR #9583, opened 2017-04-18
Update Recode.net.xml

J0WI:
See also #7344


jeremyn:
@J0WI Is this new PR because Recode.net has improved their HTTPS coverage and now we don't need all those exclusions in my older PR?

J0WI:
Exactly. And I also added `events.recode.net`.

jeremyn:
A person from Vox Media (which controls Recode.net) recently created PR https://github.com/EFForg/https-everywhere/pull/9700 for Vox domains. I think you and I should close our Recode PRs in favor of #9700 . What do you think?

iamcarrico:
Sorry, I didn't see this PR— I just went through all our properties that are defaulting to HTTPS now.

J0WI:
@iamcarrico could you increase the coverage of targets in your pending PR?
E.g. here I also added `events.recode.net`. If a host doesn't support HTTPS we use to list them in comments for future maintenance.

iamcarrico:
I am doing a full audit of our subdomains and will add that in. 

Note: I am moving cross country this week— so implementation will be slightly delayed.

jeremyn:
@J0WI What do you want to do with this PR?

J0WI:
Since there is no progress in #9700 I would love to see this merged.

jeremyn:
Mark as `Invalid certificate` or similar:
- [ ] https://photos.recode.net

Add `target`:
- [ ] https://live.recode.net
- [ ] https://on.recode.net

Also:
- [ ] Remove the `securecookie` tag or make it more specific (because we have a failing subdomain)

iamcarrico:
Sorry y'all— I have not been able to get back to this. The structure is unfamiliar, and every time I tried I kept getting bogged down. I can say for certain that all of our sites are 100% HTTPS as of now though.

jeremyn:
@iamcarrico That's fine, thanks for following up. If you are going to step back from making Vox rulesets indefinitely, please put a note in your PR #9700 and close it.

Also, just FYI, for HTTPS Everywhere we consider domains rather than sites, so for example https://photos.recode.net has an invalid certificate and so we won't include it in our rulesets. From the certificate error it looks like you are relabeling a SmugMug site under `recode.net`, so I can understand if you don't consider it one of your sites, but for us, it is. The best way to signal that all of your sites are covered by HTTPS is to add them to the HSTS preload list. When that happens, the domains should eventually be removed from HTTPS Everywhere.

iamcarrico:
I'll make a ticket to remove that domain— I think it is just legacy gunk. Thank you.

J0WI:
`on.recode.net` is already handled by `Bitly_branded_short_domains.xml`

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/9583/commits/0a6f921bba21b30cde8f7cbf711cba53d06e4be3.

There are other cookies at https://www.recode.net besides `^_chorus_geoip_continent$` that are secure: false. You can just use a `.+` for the `securecookie` `name`.

J0WI:
Done.

jeremyn:
Thanks, merged.

Issue #9576, opened 2017-04-17
.


PekaFat:

J0WI:
Are all external links with `https://vk.com/away.php?to=` affected or just those to donationalerts?

PekaFat:
I think all. Tested pastebin.com and wowhead.com - same result. But for redirect w/o https-everywhere you need use `&post=-145213353_23&cc_key=` too.

swarmer:
I can confirm that HTTS everywhere breaks VK redirects. VK has a very advanced redirect system with cookies for reasons that are beyond understanding of mere mortals like me. Anyway, with HTTPS everywhere enabled browser doesn't send cookies in one of the requests.

It goes like this:
```
 -> https://vk.com/away.php?to=blahblah&cc_key=
 <- Response: set-cookie:remixsec_redir=blahblah; path=/; domain=.vk.com
    And 302's to away.vk.com
 -> http://away.vk.com/away.php
    Here the browser usually sends the cookie shown above, but not with HTTPS everywhere enabled.
    With no cookie set, redirect fails.
```

J0WI:
@swarmer can you confirm that the redirect works when you use HTTPS on `away.vk.com`?

swarmer:
Yes, when I manually go to https://away.vk.com/away.php, I'm redirected to the address that was saved in the cookie previously.

J0WI:
Thanks!
So I think #9584 will fix this without making any exclusion.

PR #9584, opened 2017-04-18
#9576

J0WI:
See #9576

### DIFF
--- a/src/chrome/content/rules/VK.xml
+++ b/src/chrome/content/rules/VK.xml
@@ -7,7 +7,6 @@
 
 	Self-signed:
 		- pp.vk.com
-
 -->
 <ruleset name="VK.com">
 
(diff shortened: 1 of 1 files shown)


J0WI:
Maybe we can re-enable the securecookie rule together with `away.vk.com`, see https://github.com/EFForg/https-everywhere/issues/9576#issuecomment-295256665

Hainish:
I don't have an account and can't sign up for one without a phone number, so I can't double-check on this.

@jeremyn do you have a VK account?

I'm willing to merge without a review for the sake of expediency, since this comes from a trusted user and is a top-100.  But first I'd like to see if I can find someone in the office here that I can test this account with.  In the future it would be nice to allow a day or so before flagging urgent ruleset bugs.

jeremyn:
I don't have a VK account.

J0WI:
Just to clarify: I was making these changes based on the feedback in #9576. Since the final fix is just adding a target I don't expect any regressions. The first attempt with exclusions was more risky.

PR #9411, opened 2017-04-04
Create Element14.com.xml

cschanaj:
Related: #3069, #4073, #4224 . 

Remark: 
 - `au.element14.com`, `cn.element14.com`, `in.element14.com` etc do not pass the Travis build because of the different status codes, however, I cannot reproduce this issue using `curl` in the command line. (why?)

PR #9603, opened 2017-04-19
Create rules for Element14.com, related #9411

cschanaj:
Related: #9411


cschanaj:
@J0WI It looks like that whatever the first target is will render the Travis build fail even if working test URLs are supplied. Any suggestion?

J0WI:
uhm thats sad.
Maybe we can use a wildcard target and a rule that ony redirects those language subdomains. Then we can add this file to  the `ruleset-coverage-whitelist.txt` to make Travis happy. (without having tested this myself)

cschanaj:
The ruleset 'passes' Travis test now, this should close #9411 if it also pass the review. thanks!

P.S. I notice that  `ruleset-coverage-whitelist.txt` is not sorted correctly, is this expected?

J0WI:
I think initially they were sorted by the second col (ruleset names).
If you want to sort them correctly, you can contribute to `utils/ruleset-coverage-whitelist-cleanup.sh`

PR #9604, opened 2017-04-19
Delete Risk.io.xml

cschanaj:
`risk.io` and `*.risk.io` are gone from DNS. BTW, do we require (or have a script to check that) all rulesets contain at least one resolvable target currently?


J0WI:
This check is done by Travis with `utils/validate.sh` and `utils/relaxng.xml`.

cschanaj:
It seem to me that `utils/validate.sh` only check if all rulesets contain at least one non-wildcard target. 

By `resolvable` I mean `have an existing DNS record`, say from `https://dns.google.com` etc. If all targets are gone from DNS, a rule is unlikely ever applied (thus ought to be updated/ removed). Currently, I  notice only some 40+ rulesets having no resolvable targets, so I don't think this is urgent. But in long term it seem better to have one such script to perform check regularly, e.g. quarterly imho.

J0WI:
[Here](https://github.com/EFForg/https-everywhere/blob/master/utils/relaxng.xml#L32) we require at least one valid target.
If the target also has a DNS record and a active HTTP(S) service is checked by our fetch tests.

cschanaj:
I see, thanks for your explanations!

Issue #9582, opened 2017-04-18
Audit CACert rulesets


J0WI:
Since Let's Encrypt has been launched some `platform="cacert"` may need an update to be enabled for all users.

clairew:
Hi @J0WI, this is my first time contributing to open source. Would like some pointers on how to get started with this task! Could you elaborate? Thank you so much, and sorry if I'm asking for a lot of details.

J0WI:
Thanks for picking up this task!
First I recommend to read our [CONTRIBUTING.md](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md). We have some quality standards for rulesets.

Then you can choose any ruleset that contains `platform="cacert"`. If the ruleset already uses our current code style, you can just update it. If not, you need to rewrite those parts to pass our automated tests.
[`Toofishes.net.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Toofishes.net.xml) seems to be an easy one to start with.

Please use a separate branch/PR for each ruleset if they are not related.

clairew:
Thanks for the pointers!

cschanaj:
I have made a very simple script to test the hosts in `cacert` rulesets. (since the script is so simple, I will assume there are false results). thanks.

```bash
#!/bin/bash

REGEX1='platform=\"cacert\"'
REGEX2='<target host= *\"\([-.a-zA-Z0-9]*\)\" \?\/>'
REGEX3='s/^[\t ]*<target host= *\"\([-.a-zA-Z0-9]*\)\" \?\/> *$/\1/g'

OUTPUT_PHASE1='/tmp/cacert_flist.txt'
OUTPUT_PHASE2='/tmp/target_hlist.txt'
OUTPUT_PHASE3='./output_list.txt'

# Fetch the current master
git clone 'https://github.com/EFForg/https-everywhere.git'

# Collect the list of file containing "platform=\"cacert\""
grep -l "$REGEX1" ./https-everywhere/rules/*.xml > "$OUTPUT_PHASE1"

# Obtain the first target
while read rfile;
do
	grep "$REGEX2" "$rfile" | sed "$REGEX3" | head -n 1
done < "$OUTPUT_PHASE1" > "$OUTPUT_PHASE2"

while read host;
do
	curl -s -o /dev/null "https://$host" && echo "$host" 
done < "$OUTPUT_PHASE2" > "$OUTPUT_PHASE3"
``` 
[...]

PR #9609, opened 2017-04-19
Update OzLabs.org.xml

cschanaj:
Related: #9582


src/chrome/content/rules/OzLabs.org.xml
```diff
@@ -0,0 +1,62 @@
...
+			 - yaboot.ozlabs.org
+
+		Incomplete certificate chain error:
+			 - bilbo-bmc.ozlabs.org
+
+		4xx client error:
+			 - patchwork-proxy.ozlabs.org
+
+		5xx server error:
+			 - anton.ozlabs.org
```
J0WI: see `http://anton.ozlabs.org/blog/`

PR #9628, opened 2017-04-20
[Nature.com] Remove wildcard and update subdomains

Bisaloo:
It's a though one. I may be missing some rules but it shouldn't break anything (which matters more IMO).

Please also note that real https coverage for this domain is quite new (the cert is valid from 04/13 and I don't remember seeing it in https before today) so it might still be in progress.


Bisaloo:
For some reason, travis times out on http://data.nature.com.

It's quite strange because it worked for a6bad68 but after just changing the ruleset name in 877c629, it stopped working.

src/chrome/content/rules/Nature.com.xml
```diff
@@ -0,0 +1,152 @@
...
+		press.nature.com
+		proceedings.nature.com
+		protocols.nature.com
+		reports.nature.com
+		search.nature.com
+		split.nature.com
+		structbio.nature.com
+		support.nature.com
+		xml.nature.com
+		xps.nature.com
```
J0WI: my.nature.com

Bisaloo: Given the message over `http`:

> my.nature.com has been discontinued

> Following a beta pilot, my.nature.com discontinued service on April 14th, 2014.

> We apologize for any inconvenience caused.

is it really necessary to include it?

J0WI: IMHO: Since this host is still up there is no excuse to not support https.

Bisaloo: Looks like they noticed something was up and the subdomain is now unreachable even over `http`.

PR #9643, opened 2017-04-22
[Wire.com] Update ruleset

lucaspetter:
Changes:
- Prior to this PR, the `<rule>` tags only redirected a few hosts to https, despite the wildcard target. They didn't cover Wire's main web app at `app.wire.com`. The new `<rule>` redirects everything to https.
- Added a target for Wire's Piwik analytics domain, `wire.innocraft.cloud`.
- Added test URLs for all subdomains I could find.
- Wildcard securecookie tag.
- Ruleset is no longer "partial".

### DIFF
--- a/src/chrome/content/rules/Wire.com.xml
+++ b/src/chrome/content/rules/Wire.com.xml
@@ -1,44 +1,20 @@
-<!--
-	Nonfunctional subdomains:
-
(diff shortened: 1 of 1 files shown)


lucaspetter:
Thanks!

PR #9647, opened 2017-04-22
Bild.de: secure more subdomains

numismatika:


src/chrome/content/rules/Bild.de.xml
```diff
@@ -12,14 +12,12 @@
...
 
 		^ ¹
-		- bilder ²
 		- epaper ³
 		- impressum ¹
 		- wetter ⁴
 		- www ²
 
 	¹ Dropped
-	² 504, akamai
```
J0WI: www has still a reference to ²

PR #9648, opened 2017-04-22
Mookie1.com:  add more subdomains

numismatika:


src/chrome/content/rules/Mookie1.com.xml
```diff
@@ -1,44 +1,126 @@
 <!--
-	Nonfunctional subdomains:
-
-		- (www.)
```
J0WI: Please readd a comment about `(www\.)?`.

J0WI:
This needs to be rebased.

PR #9649, opened 2017-04-22
Update AdBlock.xml

cschanaj:


J0WI:
`chromeadblock.com` does not work for me at all.

cschanaj:
Maybe `chromeadblock.com` has changed their server configuration. I decided to remove the `target` from this rule completely for the moment.

cschanaj:
Maybe `chromeadblock.com` has changed their server configuration. I decided to remove the `target` from this rule completely for the moment.

PR #9652, opened 2017-04-22
Update APA.org.xml

afarrow:
Loading the site's main page and some other pages on the subdomains https://www.apa.org, https://www2.apa.org and https://books.apa.org over https triggers a JavaScript alert that complains that "An error occurred while loading the featured jobs.". But it doesn't seem to affect functionality at all so I think the benefits outweigh the side-effects.

### DIFF
--- a/src/chrome/content/rules/APA.org.xml
+++ b/src/chrome/content/rules/APA.org.xml
@@ -14,7 +14,6 @@
 	<target host="apa.org" />
 	<target host="www.apa.org" />
 	<target host="books.apa.org" />
-	<target host="cesaoas.apa.org" />
 	<target host="coaportal.apa.org" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/APA.org.xml
```diff
@@ -1,21 +1,33 @@
 <!--
-	Nonfunctional hosts in *apa.org:
-
-		- doi ʳ
-		- psycnet ʰ
```
J0WI: Please keep those comments, they are still valid.

afarrow: Fixed

J0WI:
Travis reported an incomplete cert chain at `join.apa.org`

PR #9654, opened 2017-04-22
Add Miniclip.com.xml

J0WI:


jeremyn:
Sublist3r shows a lot more domains than what you have here. Do you want to add some more? Or we could just mark this as `(partial)`.

J0WI:
Are they all still in use?

jeremyn:
I checked one at random, https://payments.miniclip.com , and it seemed okay. I'm not going to check all of them unless it's to review your work.

J0WI:
Okay, let's go for the partial flag.

PR #9655, opened 2017-04-23
Create GreenChef.com.xml

Bisaloo:


src/chrome/content/rules/GreenChef.com.xml
```diff
@@ -0,0 +1,21 @@
+<!--
+	No working URL known:
+		cdn2.greenchef.com
```
J0WI: just search for `site:cdn2.greenchef.com`

Issue #9631, opened 2017-04-20
Translate FAQ


dvd2000:
Continuing from #9466. 
@Hainish 

As suggested, we should probably support the FAQ translation (as Privacy Badger does). Initially the translation process was there: [EFForg/www-l10n](https://github.com/EFForg/www-l10n), but this repo was not very active so we should consider making the translations in the https-everywhere repository.

We have two options:

1) Do all the job on GitHub through pull requests. In this case we should create a folder for each language inside the docs folder (docs/en_US, docs/fr, docs/it, ...) and just translate the .md files directly without changing the Markdown syntax. Here the problem is about the review of the translations (how can we assure the quality of the translations?).

2) Move the translation process to Transifex (EFF already as an account there). The problem in this case is about the creation of the .md files from the Transifex translation files. Also, it isn't very common to translate markdown files with Transifex. On the other side, Transifex offer a great platform for translator that will probably increase contributions.

What should we do? Do you have any other option?

Bisaloo:
It is likely to be a quite small file and I think it's possible for one person to translate the whole file so Transifex is probably overkill (in addition with the issues you mentioned) so Github and PR sounds fine to me.

dvd2000:
Ok, at the moment there is also some documentation for the developers in the docs folder (README, development, rulesets).
Should we move these documents somewhere else?
Otherwise we can keep them there and organize the docs folder like this:

```
+-- README.md
+-- development.md
+-- rulesets.md
+-- en_US
|   +-- faq_en.md
+-- fr
|   +-- faq_fr.md
+-- it
|   +-- faq_it.md
```

and so on for the other languages.

Hainish:
@dvd2000 I'd suggest the following, almost identical schema:

```
+-- README.md
+-- en_US
|   +-- faq.md
|   +-- development.md
|   +-- rulesets.md
+-- fr
|   +-- faq.md
+-- it
|   +-- faq.md
```

And then maybe create a symlink from `default` to `en_US`, and make a note in the README.md that `en_US` is the canonical source.

dvd2000:
@Hainish  I will write some instructions to help the translators, should I put them in docs/README ?

Hainish:
Yes please

dvd2000:
@Hainish I've written some instructions (PR #9670), tell me if you think I should add something else

dvd2000:
I'll start to work on the italian translation

pipboy96:
[...]

PR #9670, opened 2017-04-24
Documented translation process

dvd2000:
This should help the translators.

Do you think I should add something else?


dvd2000:
Related to issue #9631

docs/README.md
```diff
@@ -5,3 +5,15 @@ The markdown files contained in this path provide documentation for contributing
...
 
 Copy the output, excluding the header on the first line, to the source of the relevant page within the CMS.  Note that some of the pages are dynamically generated and are not generated from templates contained here.
+
+# Translation of the FAQ
+
+#### Do you want to translate the FAQ to your language?
+
+Fork this repository and, if it doesn't exist yet, create a folder with the name of your language code (*it*, *fr*, ...) in the same level of the *en_US* and *default* folders. Then copy the faq.md file from the *en_US* folder into your language folder and translate it **without touching at the Markdown syntax**. When you're done, create a pull request to merge your changes.
+
+Initially the translation was happening in [this repository](https://github.com/EFForg/www-l10n) so if you're lucky you may find the FAQ already translated in your language there. In this case you can use the old translation as a starting point for yours, but please, take some time to review it.
```
Hainish: I wouldn't add this since I don't think there actually are any useful translations in that repo, and also because the FAQ is likely to change soon, leaving any translations that might be there out-of-date.

docs/README.md
```diff
@@ -5,3 +5,15 @@ The markdown files contained in this path provide documentation for contributing
     pandoc faq.md
 
 Copy the output, excluding the header on the first line, to the source of the relevant page within the CMS.  Note that some of the pages are dynamically generated and are not generated from templates contained here.
+
+# Translation of the FAQ
+
+#### Do you want to translate the FAQ to your language?
```
Hainish: We should generalize this, so that it is not specific to the FAQ.  Other docs include `development.md`, `rulesets.md`

Hainish:
In general, I think we may want to discourage translations until we get a chance to do a comprehensive review of the FAQ.  We don't want to render translators work useless by changing the base document significantly in the near future, we'd rather those cycles be deployed *after* we update the FAQ.

dvd2000:
@Hainish Ok, so I will wait for the FAQ to be updated. We can leave this PR open until then. I will apply the requested changes. It would be nice if you could notify me when you think we can start translating the new FAQ.

zoracon:
This was resolved and the conversation around what new translations to use are ongoing for 2019. Closing out, thank you.

PR #9684, opened 2017-04-26
Added backloggery.com

miyalys:


J0WI:
This site has mixed content issues and needs that mixedcontent platform attribute.

PR #9687, opened 2017-04-26
RTVS

MilanKral:


J0WI:
Can you confirm that this does not break radio and television live streams?

MilanKral:
Live streaming works.

J0WI:
Thanks!

PR #9695, opened 2017-04-28
[ScienceDirect.com] Update subdomains and add platform (fix #9689)

### LINKED ISSUE #9689: ScienceDirect breaking with HTTPS Everywhere
www.sciencedirect.com is partially broken when using HTTPS Everywhere. I suppose the website uses several insecure connections, such as to els-cdn.com (the Elsevier CDN). When letting Chrome "Load unsafe script", the site works as expected.

It may be problematic to resolve this issue on the side of the website because it also makes calls to educational institutions servers to check if users have access rights to certain articles. These calls, at least for some institutions, are insecure. Therefore, I suppose a different rule should be used in HTTPS Everywhere, right?

Example: https://www.sciencedirect.com/science/article/pii/S0004370215000843

Thanks for your help.

Bisaloo:
Major parts of the website have mixed content. I think this option is safer than exclusions.

### DIFF
--- a/src/chrome/content/rules/sciencedirect.com.xml
+++ b/src/chrome/content/rules/sciencedirect.com.xml
@@ -41,6 +41,13 @@
 	<target host="holdings.sciencedirect.com" />
 	<target host="pdn.sciencedirect.com" />
 	<target host="rss.sciencedirect.com" />
+	
+	<target host="sdfestaticassets.sciencedirectassets.com" />
+		<test url="http://sdfestaticassets.sciencedirectassets.com/prod/9b58ca69938c5ab521a75fffc20b1f4d285d94a0/images/logo.png" />
+	<target host="sdfestaticassets-us-east-1.sciencedirectassets.com" />
+		<test url="http://sdfestaticassets-us-east-1.sciencedirectassets.com/prod/9b58ca69938c5ab521a75fffc20b1f4d285d94a0/images/logo.png" />
+	<target host="sdfestaticassets-us-west-2.sciencedirectassets.com" />
+		<test url="http://sdfestaticassets-us-west-2.sciencedirectassets.com/prod/9b58ca69938c5ab521a75fffc20b1f4d285d94a0/images/logo.png" />
 
 	<rule from="^http:" to="https:" />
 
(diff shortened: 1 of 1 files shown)


Bisaloo:
Should I add a downgrade rule to fix the fact that the website sometimes links to `//help.sciencedirect.com` even though it is not working over https?

Bisaloo:
Wow, that website is a mess... The URL on Google for `www.info.sciencedirect.com` is over `https`.

Bisaloo:
@J0WI, @jeremyn I'm sorry to ping you but can we prioritize this PR? I've heard numerous complaints about the breakage of this top-1k website.

src/chrome/content/rules/sciencedirect.com.xml
```diff
@@ -1,32 +1,54 @@
...
-    topics.sciencedirect.com
+	Time out:
+		alpha.sciencedirect.com
 
-Mixed content:
-    attach.sciencedirect.com
+	Invalid certificate:
+    	ars.sciencedirect.com
+    	binary-services.sciencedirect.com
+		info.sciencedirect.com
```
J0WI: That one WFM.

src/chrome/content/rules/sciencedirect.com.xml
```diff
@@ -1,32 +1,54 @@
...
+
+	Secure connection failed:
+    	top25.sciencedirect.com
 
-Refused connection:
-    help.sciencedirect.com
-    top25.sciencedirect.com
 -->
-<ruleset name="sciencedirect.com">
+<ruleset name="ScienceDirect.com" platform="mixedcontent">
```
J0WI: Maybe you could add #9689 as a reference for the mixedcontent platform.

src/chrome/content/rules/sciencedirect.com.xml
Bisaloo: I just tried and it's now refused for me.

J0WI: I get a 301 to `/www.elsevier.com`

Bisaloo: Over `http`? Ok, I do as well.

Unfortunately, it doesn't redirect for deep links so I can't write a simple rule. 

Deep links apparently support `https` so I may be able to write a more complex rule but I'd like to get this merged today if possible, to be included in the next release.

I will investigate and see what can be done for `info.sciencedirect.com`.

test url: https://info.sciencedirect.com/techsupport/books/catalogs/food-science-nutrition-books-catalog.pdf

PR #9706, opened 2017-04-29
Create House.gov.xml

afarrow:
And removed house.gov rules from other files

I didn't include representatives' subdomains in this PR because going through 435 or more subdomains was more work that I wanted to do for one PR. I put all the subdomains found by Sublist3r that looked like they were for representatives here: https://pastebin.com/KPLFcAY3. Someone else is welcome to go through those and make sure they all work and create their own PR

### DIFF
--- a/src/chrome/content/rules/House.gov.xml
+++ b/src/chrome/content/rules/House.gov.xml
@@ -48,6 +48,7 @@
 	<target host="oce.house.gov" />
 	<target host="oversight.house.gov" />
 	<target host="periodical.house.gov" />
+	<target host="policy.house.gov" />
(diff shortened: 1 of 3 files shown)


afarrow:
So I was double-checking my work and I came across an issue that I'm unsure of how to resolve. 11 of the targets in the House.gov.xml ruleset support HTTPS on their main page but redirect to HTTP when visiting any other page. For example https://agriculture.house.gov works and is served over HTTPS but https://agriculture.house.gov/about/ redirects to HTTP. 

So for these 11 subdomains that I've found, do I make exclusions for each one so the rule only applies to the main page or do I remove the targets altogether?

PR #9707, opened 2017-04-29
Update Init7 rules

J0WI:
#9582


J0WI:
I rebased my branch, so tests are passing now.

jeremyn:
`Fiber7.ch.xml`:

- [ ] This ruleset should be merged into `Init7.xml`. All of the covered domains redirect to somewhere under https://www.init7.net.

`Init7.xml`:
- [ ] https://ipv6.init7.net should be `Refused`, not `Mismatch`
- [ ] Mark https://customerlists.init7.net as `Refused`
- [ ] Mark https://notomys.init7.net as `Refused`
- [ ] I'm not sure what the `Customers` note means. Could you clarify that in the ruleset?

J0WI:
>  https://ipv6.init7.net should be Refused, not Mismatch

Nope, it's accessible, but only over IPv6.
I've addressed all other points.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/9707/commits/20b9bca827d398071e7abf4d168d0bae75c729f2.

I'm not able to easily test https://ipv6.init7.net but I marked it as okay anyway because it doesn't really matter.

I think we can remove the `*.fiber7.init7.net`  domains and `Customers` comment entirely. The `*.fiber7.init7.net` domains time out with both HTTPS and plain HTTP, which is enough to leave them out.

J0WI:
> I think we can remove the *.fiber7.init7.net domains and Customers comment entirely. The *.fiber7.init7.net domains time out with both HTTPS and plain HTTP, which is enough to leave them out.

This depends on whether the customer have configured a public HTTP service or not.

jeremyn:
I'm confused by the PTR comment. If I test one of the `*.fiber7.init7.net` domains found by Sublist3r with `dig [some IP].fiber7.init7.net any`, I only get an A record, and no PTR record. So I'm not really sure what you're saying with the comment.

Also the comments in this section should be testable and I'm not sure what a test case is for this.

J0WI:
Use `dig -x [smoe IP]`

jeremyn:
Okay, the PTR record is actually for host `[reverse of some IP].in-addr.arpa` and can be found with `dig [reverse of some IP].in-addr.arpa any`. I'm still not sure what you're trying to say with the comment.

J0WI:
This host are managed by the customers, so it' up to them if they want to support HTTPS or not. It's impossible to list all of them.

jeremyn:
How about instead of "Customers (PTR Resource Records)" we say "User-controlled subdomains not covered here"?

J0WI:
Like so?

jeremyn:
That's fine, thanks. Merged.

PR #9714, opened 2017-04-30
Update Kubieziel.de.xml

cschanaj:
Related: #9582


src/chrome/content/rules/Kubieziel.de.xml
```diff
@@ -1,17 +1,13 @@
-<ruleset name="Kubieziel.de" platform="cacert">
-
-	<!--	Direct rewrites:
-				-->
+<!--
+	Non-functional hosts
+		SSL peer certificate was not OK:
+			 - mx.kubieziel.de
```
J0WI: the site has a wildcard dns record and HSTS, but no wildcard cert.

PR #9718, opened 2017-04-30
Update Bitcoin-Contact.org.xml

cschanaj:
Related: #9582

### DIFF
--- a/src/chrome/content/rules/Bitcoin-Contact.org.xml
+++ b/src/chrome/content/rules/Bitcoin-Contact.org.xml
@@ -2,7 +2,7 @@
 	Mixed Content Blocking (MCB) tiggered:
 		- bitcoin-contact.org
 -->
-<ruleset name="Bitcoin-Contact.org" platform="mixedcontent">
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Bitcoin-Contact.org.xml
```diff
@@ -1,35 +1,13 @@
...
-
+	Mixed Content Blocking (MCB) tiggered:
 		- bitcoin-contact.org
-
 -->
-<ruleset name="Bitcoin-Contact.org" default_off="expired" platform="cacert">
-
-	<!--	Direct rewrites:
-				-->
+<ruleset name="Bitcoin-Contact.org" platform="mixedcontent">
```
J0WI: Is there any critical mixed content that breaks functionality? I only see blocked tracking scripts.

cschanaj: Those are basically tracking script, I will remove the `mixedcontent` attribute.

PR #9725, opened 2017-04-30
[gov.hk] New ruleset

Marco129:
HKSAR portal (GovHK) finally supports HTTPS 🇭🇰

### DIFF
--- a/src/chrome/content/rules/GovHK.xml
+++ b/src/chrome/content/rules/GovHK.xml
@@ -18,9 +18,7 @@
 		- voterinfo.gov.hk.xml
 -->
 <ruleset name="GovHK">
-	<target host="gov.hk" />
 	<target host="www.gov.hk" />
 
-	<rule from="^http://(www\.)?gov\.hk/"
-		to="https://www.gov.hk/" />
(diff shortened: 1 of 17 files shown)


src/chrome/content/rules/GovHK.xml
```diff
@@ -0,0 +1,26 @@
...
+		- HKGov-HYD.xml
+		- HKGov-ITF.xml
+		- HKGov-LAD.xml
+		- HKGov-LCSD.xml
+		- HKGov-TID.xml
+		- hkpl.gov.hk.xml
+		- hkqf.gov.hk.xml
+		- hongkongpost.gov.hk.xml
+		- jobs.gov.hk.xml
+		- voterinfo.gov.hk.xml
```
J0WI: You should also add a reference in these rulesets to this list.

src/chrome/content/rules/GovHK.xml
```diff
@@ -0,0 +1,26 @@
...
+-->
+<ruleset name="GovHK">
+	<target host="gov.hk" />
```
J0WI: ^gov.hk is not a valid domain.

PR #9731, opened 2017-04-30
Create rule for Gamesandpals

philip94:
Create Gamesandpals.com.xml. Does not seem to have any nonfunctional hosts.


J0WI:
The cert expired recently. Let's see if they fix this.

J0WI:
cert has not been renewed.

jeremyn:
All three domains in the ruleset are not in DNS, so there's nothing to add here. Closing.

PR #9734, opened 2017-05-01
[viu.tv] New ruleset

Marco129:
New ruleset for [ViuTV](https://www.viu.tv)

### DIFF
--- a/src/chrome/content/rules/viu.tv.xml
+++ b/src/chrome/content/rules/viu.tv.xml
@@ -6,6 +6,7 @@
 	<target host="static.viu.tv" />
 
 	<exclusion pattern="^http://(www\.)?viu.tv/encore/" />
+		<test url="http://viu.tv/encore/now-report" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/viu.tv.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="ViuTV">
+	<target host="viu.tv" />
+	<target host="www.viu.tv" />
+	<target host="edm.viu.tv" />
+	<target host="event.viu.tv" />
+	<target host="static.viu.tv" />
+
+	<exclusion pattern="^http://(www\.)?viu.tv/encore/" />
```
J0WI: You need to escape all the dots.
Please also add a comment about the reason for the exclusion.

Marco129: Ruleset updated. Thanks!

J0WI: Thanks!

PR #9743, opened 2017-05-01
Create rule for pcgamer

philip94:
Create Pcgamer.com.xml. Apart from the two main subdomains, all other hosts seem nonfunctional. Error type indicated for each.


J0WI:
This site has some mixed active content issues.

6u621jpq6ybdw:
@philip94 you only need to add platform="mixedcontent"

PR #9747, opened 2017-05-01
Update pocketgamer.biz

philip94:
Took off the main subdomains, as it had too many redirections.


src/chrome/content/rules/Pocketgamer.biz.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Nonfunctional hosts in *.pocketgamer.biz:
+	
+	HTTPS Connection refused:
+	    www2.pocketgamer.biz
+	
+	Too many redirects to HTTPS:
```
J0WI: In fact, they redirect back to HTTP.

PR #9758, opened 2017-05-01
Create rule for Touchgen

philip94:
Create Touchgen.com.xml. Site contains MCB, used a test URL of the site's logo.

### DIFF
--- a/src/chrome/content/rules/Touchgen.net.xml
+++ b/src/chrome/content/rules/Touchgen.net.xml
@@ -6,6 +6,8 @@
 
 	<rule from="^http:" to="https:" />
 
-	<test url="http://www.touchgen.net/wp-content/uploads/2016/09/logo.png" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Touchgen.net.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Touchgen" platform="mixedcontent">
```
J0WI: Is there only mixed passive content? Then we can ignore it.

philip94: Yes, all the Javascript seems to load fully on HTTPS. The mixed content is on "https://mail.touchgen.net". Only a few images show the alt text. I added some more test URLs for the images that do not show up.

philip94: Took off the mixedcontent as you requested. The previous commit did not pass the rules test. Should be fine now.

PR #9765, opened 2017-05-02
Update Stalkr.net.xml

cschanaj:
Related: #9582

### DIFF
--- a/src/chrome/content/rules/Stalkr.net.xml
+++ b/src/chrome/content/rules/Stalkr.net.xml
@@ -5,11 +5,6 @@
 			 - smtp.stalkr.net
 			 - sshgw.stalkr.net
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Stalkr.net.xml
```diff
@@ -1,20 +1,37 @@
...
+	Non-functional hosts
+		Peer certificate cannot be authenticated with given CA certificates:
+			 - ns0.stalkr.net
+			 - smtp.stalkr.net
+			 - sshgw.stalkr.net
 
-		- blog *
-
-	* Blogger
+		Incomplete certificate chain error:
```
J0WI: Please double check, I think this has been fixed.

PR #9635, opened 2017-04-21
 	Re-activate and Update CNBC.com related ruleset

cschanaj:
`Sublist3r` output more than 400+ sub-domains, only added those which are obvious on `www.cnbc.com`. thanks. 

 - [x] cnbc.com
 - [x] www.cnbc.com
 - [x] apps.cnbc.com
 - [x] data.cnbc.com
 - [x] fm.cnbc.com
 - [x] futuresnow.cnbc.com
 - [x] media.cnbc.com
 - [x] portfolio.cnbc.com
 - [x] pro.cnbc.com
 - [x] sc.cnbc.com
 - [x] video.cnbc.com
 - [x] watchlist.cnbc.com
 - [x] webcast.cnbc.com

Remark: Some videos are broken because of `Demdex.net` and `Dynamic Yield.com` rulesets, disable them manually fix the issues #9641 .

PR #9772, opened 2017-05-03
Update CNBC.com related ruleset

### LINKED ISSUE #9641: Ruleset 'Demdex.net' breaks videos on CNBC.com
It looks like that `fast.nbcu.demdex.net` is causing this issue on http://www.cnbc.com/latest-video. See https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Demdex.net.xml#L26

Remark: there is also a CORS issues with `dynamicyield.com`, I am not sure if this is related.

With `Demdex.net` On:

With `Demdex.net` Off:

cschanaj:
@J0WI This is a re-open of #9635. Could you please review and merge this instead?

I have put the rule for `www` to `partial` because a simple redirect occasionally fall-back and disable the whole ruleset automatically on `mixedcontent` platform. I rather having partial coverage in this case.

Besides, I closed #9641 for the moment. As mentioned in #9635, I cannot reproduce #9641 on Firefox (without flash player). The page is broken even if HTTPS Everywhere is disabled on my Firefox. I suspect there are some issues on their side because it works perfectly on `Chromium` even if all `ruleset` are enabled.  

Apologies for the mess.


J0WI:
LGTM, Thanks!

PR #9780, opened 2017-05-03
Create invisiblethings.org.xml

InstantGratimification:


J0WI:
I can't find any mixed content, so I'm not sure why you've added the platform here.

vuiiiiuc:
@J0WI 

> I can't find any mixed content, so I'm not sure why you've added the platform here.

https://blog.invisiblethings.org/2013/11/26/windows-7-seamless-gui-integration.html

J0WI:
That's only passive content, so you don't need to add a platform.

7x15:
@J0WI

Oh, right, can you change it? I deleted my initial account so I can't modify it myself.

sadfumm:
@J0WI

I made a PR that solves these problems https://github.com/EFForg/https-everywhere/pull/10172

J0WI:
Thanks!

PR #9785, opened 2017-05-04
Update Umea_Hackerspace.se.xml

cschanaj:
Related: #9582

### DIFF
--- a/src/chrome/content/rules/Umea_Hackerspace.se.xml
+++ b/src/chrome/content/rules/Umea_Hackerspace.se.xml
@@ -1,46 +1,25 @@
 <!--
-	Umeå Hackerspace
-
-
-	Fully covered hosts in *umeahackerspace.se:
-
-		- (www.)?
-		- social
-		- stuff
-		- wiki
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Umea_Hackerspace.se.xml
```diff
@@ -1,46 +1,25 @@
...
-	* Secured by us
-
+	Non-functional hosts
+		Peer certificate cannot be authenticated with given CA certificates:
+			 - asbest.umeahackerspace.se
+			 - gram.umeahackerspace.se
+			 - gumboda.umeahackerspace.se
+			 
+		Mixed Content Blocking (MCB) tiggered:
+			 - from font.google.com on umeahackerspace.se
```
J0WI: Fonts are considered as trivial, so we can ignore this.

cschanaj: Removed the comment and `mixedcontent` platform accordingly. thanks!

PR #9788, opened 2017-05-04
Malwarebytes.com

fr33tux:
Add rules for Malwarebytes domains.

Another PR to fix [malwarebytes.org legacy rules](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Malwarebytes.xml) will follow.


fr33tux:
Indeed, done!

J0WI:
Thanks!

PR #9793, opened 2017-05-05
Remove old and legacy malwarebytes.org rules

fr33tux:
malwarebytes.org is deprecated and the rule no longer needed (superseeded by https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Malwarebytes.com.xml )


J0WI:
As long as the host is still up we keep this rule. Otherwise the redirect from malwarebytes.org to malwarebytes.com is not secured.

PR #9809, opened 2017-05-06
Create Cahiers-Antispecistes.org.xml

Bisaloo:
Mixed content only breaks alignment of language buttons and can probably be safely ignored.


J0WI:
Nope, you can't click on the es/it buttons anymore.

Bisaloo:
I don't really want to put it as mixedcontent just because of this.

I will try to reach out to them and see if we can get it fixed.

PR #9837, opened 2017-05-08
Create Demisexuality.org.xml

38407625:


src/chrome/content/rules/Demisexuality.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Demisexuality.org">
+	<target host="demisexuality.org" />
+	<target host="www.demisexuality.org" />
+	<target host="demisexuality.freeforums.net" />
```
J0WI: This host should be in a separate ruleset for `*.freeforums.net`

PR #9838, opened 2017-05-08
Fixed compatibility with breitbandmessung.de

kosekmi:


kosekmi:
My bad, issuing new request

Issue #9842, opened 2017-05-08
Audit default_off rulesets


cschanaj:
Type: ruleset issue

Domain: *

Similar to #9582, I suspect 168 out of 537 `default_off="self-signed"` rulesets can be re-activated. I added a date to PRs which haven't been merge after being reviewed for a long time.

### Self-signed

 - [ ] 112cafe.com
 - [x] www.aaai.org #9824 
 - [x] connect.aber.ac.uk #10116 
 - [x] airs.com #9815 
 - [x] all-about-security.de #11087 
 - [x] alltop.com #9862 
 - [x] www.anarchistnews.org #9409 `(4 Apr 2017)`
 - [ ] ancud.de
 - [x] atomicinsights.com #10194 
 - [x] av-comparatives.org #10407 
 - [ ] badgeville.com
 - [x] crosswire.org #9845 
 - [ ] bartleby.com
 - [ ] store.fileforum.com
 - [x] bioinfobank.com #9853 
 - [x] bioinfo.pl #9853
 - [ ] vz46.brenet.de
 - [x] b-link.bucknell.edu #10083 
 - [x] alumni.buffalostate.edu  #10121 
 - [x] cn.calameo.com #9869 
 - [ ] channeladvisor.com
 - [x] members.chicagotribune.com #10068 
 - [x] coinjabber.com #9872 
 - [ ] allure.com
 - [x] couragecampaign.org #9858 
 - [ ] creativevirtual.com
 - [x] darkreading.com #9952 
 - [x] dealzon.com #11086 
 - [x] dave.org.uk #9907 
 - [ ] denx.de
 - [x] devconf.cz #9951
 - [ ] digitalmars.com
 - [x] diygadget.com #10105 
 - [ ] dlisted.com
 - [ ] drwho.virtadpt.net
 - [x] global.duke.edu #10102 
 - [ ] echothrust.com
 - [ ] egyptindependent.com
 - [ ] energylivenews.com
 - [ ] engelschall.com
 - [ ] esportsea.com
 - [x] europe-consommateurs.eu #9855 
 - [x] evoliatis.com #10103 
 - [ ] exiledonline.com
 - [x] fdopk.org #10118 
 - [ ] files.poulsander.com
 - [x] g10code.com #10015 
 - [x] fotoforensics.com #9954 
 - [x] downloads.rc.fas.harvard.edu #9878 
 - [ ] hiig.de
 - [ ] how-to-box.com
 - [ ] huxo.co.uk
[...]

PR #9854, opened 2017-05-09
Re-activate and Update MSFN.org.xml

cschanaj:
Related: #9842


Tallefer:
"Follow content" is not working, instead of a subscription options menu a small popup is shown with "There was a problem loading this content." Some other site functions are similarly affected.

J0WI:
@Tallefer could you please open an issue?

Tallefer:
I will, but I didn't find any clear guidelines for doing that. :) Could you show me some proper one to copy its style?

J0WI:
That one has quite a complex regex, but that's how we make exclusions: https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/amazon.in.xml

PR #9862, opened 2017-05-09
Re-activate and Update Alltop.com.xml

cschanaj:
Related: #9842


src/chrome/content/rules/Alltop.com.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Remark: *.alltop.com has a wildcard DNS record and 
+		a wildcard certificate.
+-->
+<ruleset name="AllTop.com">
+	<target host="alltop.com" />
+	<target host="*.alltop.com" />
```
J0WI: Is there also a reason to use a wildcard target or can we list all hosts?

cschanaj: The wildcard DNS basically redirect `https://foo.alltop.com` to `http://alltop.com/?keywords=foo`. And there is 700+ sub-domain support HTTPS correctly. I do not want this to overwhelm reviewers and future maintainers. If you prefer, I can rebase it.

See https://gist.github.com/cschanaj/f0fdfa8f0535ded83603b1dfe1827b4e (Only checked automatically).

PR #9871, opened 2017-05-10
Create copilotgps.com.xml

coliff:


src/chrome/content/rules/copilotgps.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="copilotgps.com">
+
+	<target host="copilotgps.com" />
+	<target host="www.copilotgps.com" />
+	<target host="blog.copilotgps.com" />
```
J0WI: I also found:
apphelp.copilotgps.com
my.copilotgps.com
support.copilotgps.com

coliff: Good spot! I've now added those.

J0WI: Thanks!

PR #9877, opened 2017-05-11
[Delta.com] fix #9876

### LINKED ISSUE #9876: delta.com just times out with HTTPS Everywhere enabled in Firefox
Whenever I try to load delta.com with HTTPS Everywhere enabled in Firefox it just times out, doesn't load anything. This has been going on for a long time, maybe a year? It's mentioned in this old issue:

https://github.com/EFForg/https-everywhere/issues/4705

Happens on both Windows and OS X.

Bisaloo:


Bisaloo:
Not sure why the tests are failing here. Could someone have a look please?

cschanaj:
@Bisaloo It seems that `content.delta.com` behave differently when an `user-agent` is supplied. I got `403` when I `curl` with 
```
Mozilla/5.0 (X11; Linux x86_64; rv:24.0) Gecko/20140429 Firefox/24.0 Iceweasel/24.5.0
```
and 
```
Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
```
but not with `curl` default `user-agent`
```
libcurl/7.51.0
```

Can you verify this?

Bisaloo:
Hum, nope. I don't think I can reproduce what you are describing. In the 3 cases, I get `200`.

cschanaj:
Did you tried with 
```bash
curl -I $url -A $useragent
```
or 
```bash
curl -v $url -A $useragent -o /dev/null 
```

I get the same result as yours with the 1st command, but not with the 2nd one.

Bisaloo:
Yay! The new tests work fine. No more false positive.

J0WI:
Could you also add the subdomains of this host?

PR #9884, opened 2017-05-12
Update default_off="mismatched" rulesets

cschanaj:
Related: #9906


Bisaloo:
You should also add `expired`.

cschanaj:
We have around 218 variants of error messages... I think we should figure out a list of keyword first...

cschanaj:
FYI, I would like to work on updating the `default_off` messages first. Then modify `relaxng.xml` to enforce the grammar, please refer to 9906 if you are interested.

cschanaj:
Revert 6a3fc30f89e9a95c1123167f80e5c003ec491fe7 because it was caused by #9891

PR #9889, opened 2017-05-12
[CrimethInc.com] Update subdomains

Bisaloo:


src/chrome/content/rules/CrimethInc.com.xml
```diff
@@ -1,36 +1,13 @@
...
--->
-<ruleset name="CrimethInk.com (partial)">
-
-	<target host="*.crimethinc.com" />
-
-
-	<rule from="^http://cloudfront\.crimethinc\.com/"
-		to="https://d33bhpqk8nwto8.cloudfront.net/" />
-
-	<rule from="^http://thecloud\.crimethinc\.com/"
```
J0WI: That host is still up.

PR #9891, opened 2017-05-12
Removing hsts-preloaded targets and rulesets (via hsts-prune utility)

Hainish:


Bisaloo:
Great! Thank you!

However, I had noticed domains that have been preloaded but they weren't removed. For example:

- [twitter.com](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Twitter.xml)
- [aeon.co](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Aeon.co.xml)

and some others I lost on the way.

Any idea why?

Hainish:
@Bisaloo this may be for various reasons.  One is that they aren't preloaded on *all* the browsers that we support.  Another possible reason is that they aren't delivering the "preload" directive in the `Strict-Transport-Security` header - this makes these preloads prone to removal in future iterations of the list, so we don't want to remove them from HTTPS Everywhere and leave them unprotected.

Hainish:
Checking now, the absence of the `preload` directive seems to be the case.

Hainish:
I've just talked with the Tor Project web admins, they've added the preload directive back to `torproject.org` and associated subdomains.

Bisaloo:
@Hainish, the `preload` directive is indeed back but they also need to add the `includeSubDomains` directive.

Hainish:
@Bisaloo I've talked with them about this and they are unwilling to do so.  I'm unsure why.

PR #9895, opened 2017-05-13
[VDownloader.es] Update rewrite rule

Bisaloo:
Clean up after #9891

### DIFF
--- a/src/chrome/content/rules/VDownloader.xml
+++ b/src/chrome/content/rules/VDownloader.xml
@@ -1,14 +1,15 @@
+<!--
+	Invalid certificate:
+		mail.vdownloader.es
+
+-->
 <ruleset name="VDownloader">
 
-	<!--	* for cross-domain cookie.	-->
 	<target host="vdownloader.es" />
-
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/VDownloader.xml
```diff
@@ -1,14 +1,15 @@
...
-
-	<securecookie host="^\.?vdownloader\.com$" name=".*" />
-
+	<target host="www.vdownloader.es" />
 
 	<!--	es redirects to com	-->
-	<rule from="^http://(www\.)?vdownloader\.(?:com|es)/"
-		to="https://$1vdownloader.com/" />
+	<rule from="^http://(www\.)?vdownloader\.es/"
+		to="https://vdownloader.com/" />
```
J0WI: We don't need to force non-www, just use a simple http to https redirect here.

Bisaloo: It's not the same subdomain

J0WI: Okay, I see.

PR #9896, opened 2017-05-13
Delete AgileBits.xml

Bisaloo:
Cleanup after #9891.

I couldn't find any working subdomains for `agilewebsolutions.com`.


Bisaloo:
@jeremyn, I'm sorry but I spoke too fast. It looks like this one needs to be merged as well to fix travis.

jeremyn:
I noticed it too, thanks for this and the other cleanup issues.

PR #9898, opened 2017-05-13
Update list of EFF rulesets

Bisaloo:
Cleanup after #9891


jeremyn:
@Hainish What's going on with the EFF site https://www.ripmixmake.org ? Is the `Secure Connection Failed` problem temporary or can we delete it from the ruleset?

Hainish:
I didn't encounter the problem when attempting a local connection on my browser.  Re-running fetch test.

Hainish:
Looks like the `www` subdomain is not responding correctly.  This should be fine to remove.

PR #9900, opened 2017-05-13
Delete Invasion.com.xml

Bisaloo:
Cleanup after #9891.

The website doesn't appear to use this cdn anymore.


jeremyn:
https://invasion-turnonsocial.netdna-ssl.com is technically still a valid target, but it's pretty obscure and I'm okay with deleting this ruleset to reduce future maintenance.

Issue #9906, opened 2017-05-13
Enforce Grammar for default_off Messages


cschanaj:
## Reason

Unlike the `platform` attribute, there is not any grammatical requirement on the `default_off`attribute, This make the process of performing automated check to re-activate `disable ruleset` difficult. 

To illustrate the situation, we have `mismatch`, `mismatches`, `mismatched` and their capitalization variants for the `mismatched` error. These variants make it hard for contributors to perform audit. See #9582, #9842 

A particularly bad example could be [`BufferedIO.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/BufferedIO.xml). It should be a `platform=cacert` ruleset, but the message is given in `default_off`. (It is a `mismatch` now)

## Goal

The goal is to standardize the `default_off` attributes and to enforce the grammar with [`relaxng.xml`](https://github.com/EFForg/https-everywhere/blob/master/utils/relaxng.xml) such that automated audit can be done easier.

## Steps

1. Modify `default_off` attributes in the existing `ruleset` to an agreed set of `keyword`, add the `ruleset` to [`ruleset-coverage-whitelist.txt`](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-coverage-whitelist.txt) if necessary. See https://github.com/EFForg/https-everywhere/pull/9884.

2. Update [`CONTRIBUTING.md`](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md) to explain the `keyword` we use. 

3. Enforce the grammar with [`relaxng.xml`](https://github.com/EFForg/https-everywhere/blob/master/utils/relaxng.xml).

### Patterns
Each keyword should be separated by a comma and a space
```
(keyword1|keyword2)(, (keyword1|keyword2))*
```

### Note 
Some rule use `'` (single quote) instead of `"` (double quote), it will be great if you can do a replacement as wel; :)

### Keywords
```
List of keywords (feel free to suggest new ones)

regional
refused
timeout
cert-algo
cert-chain // curl: (60) SSL certificate problem: unable to get local issuer certificate
expired
self-signed
mismatched
ssl-error // reset, or any other ssl error not stated above 
status-unexpected // 200 Error pages
status-others // other than 200, i.e. 4xx/5xx
loops // secure connection redirects to plaintext
content-different // visual/ contextual difference, also for different status code
breaks-site // in a board sense.
breaks-third-parties
cors  // CORS issues 
[...]

PR #9926, opened 2017-05-14
Re-activate, Rename and Update prnt.sc.xml

cschanaj:
#9906


src/chrome/content/rules/prnt.sc.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="prnt.sc" platform="mixedcontent">
+	<target host="prnt.sc" />
+	<target host="www.prnt.sc" />
+	<target host="ads.prnt.sc" />
+		<test url="http://ads.prnt.sc/ads/adtrue/728x90/index.html" />
```
J0WI: app.prnt.sc

cschanaj: done. this `ruleset` is longer a `default_off`, it is safe to merge this regardless of the result of #9906. thanks.

PR #9929, opened 2017-05-14
Update default_off="regional" rulesets

cschanaj:
#9906

### DIFF
--- a/src/chrome/content/rules/126.net_CN.xml
+++ b/src/chrome/content/rules/126.net_CN.xml
@@ -1,8 +1,11 @@
 <!--
 	Avaiable in global:	126.net.xml
+
+	Remark: 
+		This ruleset contains hosts that available in China ONLY.
 -->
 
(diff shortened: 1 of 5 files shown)


jeremyn:
Problems with rulesets for sites in China can be more complicated than `This ruleset contains hosts that available in China ONLY.` For example in some cases I've seen where we get different behavior because a different CDN is used inside China vs outside it and the HTTPS certificates are only configured correctly for one of the CDNs.

@gloomy-ghost and @ivysrono , it would be good to get your feedback on this pull request when you have the time.

ivysrono:
I don't think we should add too much complicate comments, they are too hard to maintain and update.

BTW, all "Only avaiable in China" use *_CN.xml
Should you update them to *regional.xml ?

ivysrono:
There is only one ruleset now: Baidu_CN.xml

cschanaj:
Closing this, discussion will be continued in #9906

PR #9931, opened 2017-05-14
 	Update KeyGhost.com.xml default_off attribute

cschanaj:
#9906


J0WI:
This rule won't work in any browser, so I'm okay with just removing it.

PR #9938, opened 2017-05-14
Add target *.freeforums.net to ProBoards.xml

38407625:


src/chrome/content/rules/ProBoards.xml
```diff
@@ -40,6 +40,8 @@
 	<target host="storage.proboards.com" />
 	<target host="www.proboards.com" />
 
+	<target host="*.freeforums.net" />
```
J0WI: Please also add `freeforums.net` (without any subdomain) and a comment, that the wildcard is required because of user generated subdomains.

38407625:
I've added freeforums.net and a comment about the wildcard.

J0WI:
Thanks!

PR #9944, opened 2017-05-15
Torproject.xml: removing deprecated exclusion, simplifying ruleset

Hainish:


src/chrome/content/rules/Torproject.xml
```diff
@@ -2,17 +2,10 @@
 
 	<target host="torproject.org" />
 	<target host="*.torproject.org" />
```
J0WI: Our own guideline requires to add at least 3 test urls for wildcards.

Hainish: Ah since I didn't add this wildcard from the original ruleset I missed this.  Added the three test urls.

PR #9956, opened 2017-05-18
Changed the developer email address in manifest.json from eff.softwar…

jgillula:
…e.projects@gmail.com to info@eff.org


Bisaloo:
Good idea! But it probably should be https-everywhere-eff@eff.org instead.

jgillula:
I talked to @Hainish and he said he'd prefer a different email address, so I'll close this PR and let him introduce a new one.

PR #9960, opened 2017-05-19
[ORCID.org] Remove exclusions

Bisaloo:


J0WI:
I have some issues on `qrcode.orcid.org`

src/chrome/content/rules/ORCID.org.xml
```diff
@@ -11,6 +14,9 @@
 		<test url="http://testurl2.orcid.org/" />
 		<test url="http://testurl3.orcid.org/" />
 
+		<exclusion pattern="^http://qrcode\.orcid\.org/" />
+			<test url="http://qrcode.orcid.org" />
```
J0WI: Missing trailing slash

Bisaloo: thank you!

PR #9963, opened 2017-05-19
Move simplemachines.xml to the rules directory

strugee:
Seemed to be in the wrong place.


redranamber:
You can close this I created #9973 which also fixes the errors in the rule.

strugee:
Neat, thanks @redranamber!

PR #9972, opened 2017-05-20
Added rules for flightradar24.com.

Jochen-A-Fuerbacher:


J0WI:
Seems like they have a wildcard DNS entry/cert.

Jochen-A-Fuerbacher:
What do you suggest to do here? Some of the (wildcard) subdomains have valid certs, others don't. So matching them with `<target host="*.flightradar24.com" />` will break some subdomains.

J0WI:
Well, all 4th+ level sub domains have no valid cert, I think you don't need to list them separately. Just add a comment about the behavior.
For the targets I would prefer to remove the cryptic subdomains that are not really used. Or you use a wildcard with exclusions for the known non-working ones.

Issue #9696, opened 2017-04-28
Add Rule for Simplemachines.org


BK1233:
Simplemachines/SMF has just enabled HTTPS but it is not by default.

PR #9973, opened 2017-05-20
Simplemachines fixes

redranamber:


J0WI:
There is an error with `media.simplemachinesweb.com`

redranamber:
It seems to be working correctly when I test it. I have the rule enabled in Firefox and it's rewriting it correctly. There isn't anything at media.simplemachinesweb.com itself but the test url "http://media.simplemachinesweb.com/smf/default/images/theme/backdrop.png" is working, so I think Travis is doing something wrong.

J0WI:
Travis is also checking `https://media.simplemachinesweb.com/`. You need to make an exclusion for `^http://media\.simplemachinesweb\.com/$` to avoid this error (please also document the error with a comment in the ruleset).

redranamber:
Not sure what you wanted commented but the test is passing now.

J0WI:
That's fine for me, thanks!

cschanaj:
this should close #9696

PR #9979, opened 2017-05-21
Update Newgrounds.xml

mrjohnson22:
Capture subdomains (like user pages: <username>.newgrounds.com)

-----------

Newgrounds has dedicated subdomains for user profile pages. These pages have the same general layout as the www site, which includes a login button with username/password fields. Without this ruleset, http links to user profile pages will remain in http, so any logins made from those pages will be unencrypted.

### DIFF
--- a/src/chrome/content/rules/Newgrounds.xml
+++ b/src/chrome/content/rules/Newgrounds.xml
@@ -1,7 +1,12 @@
 <ruleset name="Newgrounds (partial, broken)" default_off="reported broken">
 
 	<target host="newgrounds.com"/>
+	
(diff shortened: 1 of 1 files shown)


J0WI:
Please note that this rules is still `default_off`.
Could you add the reason for the wildcard as a comment in the ruleset? Please also provide at least 3 test urls.

mrjohnson22:
Made a new commit that adds some tests. I had also tested a local build with the ruleset changes applied, and it worked as expected.

I did notice that this rule is off by default. That made me less hesitant to propose this change since it was already reported as broken, so the change is unlikely to make things any worse.

PR #9988, opened 2017-05-21
create a rule file for MerNetwork

fabianfrz:
It is working but bug in Firefox seem to break it so I have disabled it:
spectacle k10018

It works if `security.mixed_content.block_active_content` is `false` in `about:config`.
So it can be enabled as soon as this will work without changing this setting.


fabianfrz:
The yandex thing breaks the travis test 👎 .

src/chrome/content/rules/MerNetwork.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="MerNetwork">
+  <target host="mernetwork.com" />
```
J0WI: Please also add `www.mernetwork.com`.
Please use tabs to indent lines.

fabianfrz:
@J0WI thanks for that, I have added the host and it looks good.

J0WI:
Thanks!

fabianfrz:
no problem - I am happy that this extension exists ;)

PR #9999, opened 2017-05-22
Create AirMap.xml

jcoscia:
Ruleset for airmap.com and airmap.io. ~~Left-wildcards were used, as all domains I found supported HTTPS.~~


J0WI:
Thanks!

PR #10011, opened 2017-05-23
Update Baidu.xml

ivysrono:


jeremyn:
- [ ] Move the https://wapp.baidu.com `target` between `wapbaike` and `wenku`.
- [ ] I think we need to keep the `exclusion` for https://tieba.baidu.com . http://tieba.baidu.com/kb/ redirects to http://static.tieba.baidu.com which eventually shows an error page, but https://tieba.baidu.com/kb/ redirects to something under https://gsp0.baidu.com which eventually gets `Refused`. https://tieba.baidu.com/zjstv/ does the same thing.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/10011/commits/3d4f93eed341c54d92481c247189395fc4b55230.

I checked some more `tieba` URLs. https://tieba.baidu.com/skype and https://tieba.baidu.com/acer redirect to HTTP so we can't include those. Unfortunately I really don't think we can include `tieba` right now, there are too many problems and Baidu is a top-100 site so we have to be really careful.

ivysrono:
#10046

PR #10049, opened 2017-05-28
Added rules for stadtwiki.net.

Jochen-A-Fuerbacher:


Jochen-A-Fuerbacher:
The tests that fail are from the Yandex.ru ruleset, that I didn't change.

src/chrome/content/rules/stadtwiki.net.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Invalid certificate:
+		www.stadtwiki.net
+		allmende.stadtwiki.net
+		ettlingen.stadtwiki.net
+		tr.ka.stadtwiki.net
+
+	Redirect loop:
```
J0WI: Please double check, I think this has been fixed.

Jochen-A-Fuerbacher: Thanks, added them as targets.

PR #10053, opened 2017-05-28
[AFSP.org] Update rewrite rule and targets

Bisaloo:


jeremyn:
- [ ] https://www.afsp.org redirects to HTTP. Add a rewrite `rule` from `www.afsp.org` to `^afsp.org.`

PR #10061, opened 2017-05-29
[Ntop.org] Update and reactivate ruleset

Bisaloo:


Bisaloo:
I didn't to go through all subdomains because it seems that they didn't set up `https` at all but I still checked all subdomains previously covered.

J0WI:
At least the shop subdomain WFM

Bisaloo:
#10843

PR #10067, opened 2017-05-29
Fix Argument list too long error

joshka:
22,000 files was too many to handle as a glob when checking whether an existing rule matches a new one. This addresses the issue by using `find -exec` instead of a glob.

Manually tested on macOS Sierra. I'm assuming the syntax will likely be the same, but not aware of the cross platform requirements of this project enough to make a call.


J0WI:
Thanks!

PR #10069, opened 2017-05-30
 	Update Fr33tux.org.xml 

cschanaj:
Delete `Fr33tux.org.xml` since it is currently preloaded, for completion of #9582. See https://hstspreload.org/?domain=fr33tux.org Related: #10056

### DIFF
--- a/src/chrome/content/rules/Fr33tux.org.xml
+++ b/src/chrome/content/rules/Fr33tux.org.xml
@@ -3,6 +3,8 @@
 	<target host="fr33tux.org" />
 	<target host="www.fr33tux.org" />
 	<target host="p.fr33tux.org" />
+	<target host="tor.fr33tux.org" />
+	<target host="wp.fr33tux.org" />
 
 
 	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


jeremyn:
I agree that `fr33tux.org` shows as preloaded at https://hstspreload.org but it doesn't appear to be preloaded in the current stable versions of the browsers that HTTPS Everywhere supports. We should be able to find `fr33tux.org` with subdomains in these files:

- [ ] https://chromium.googlesource.com/chromium/src.git/+/58.0.3029.110/net/http/transport_security_state_static.json
- [ ] https://dxr.mozilla.org/mozilla-release/source/security/manager/ssl/nsSTSPreloadList.inc
- [ ] https://gitweb.torproject.org/tor-browser.git/tree/security/manager/ssl/nsSTSPreloadList.inc?h=tor-browser-45.8.0esr-6.5-2

A quick way to check for preloading is to search for the domain in the `Query domain` search field in chrome://net-internals/#hsts in the latest stable version of Chrome. If it's not found there then there's no need to keep looking.

J0WI:
This is still not preloaded in chrome.

cschanaj:
@jeremyn @J0WI  I have re-activated the ruleset since I do not want to have too many pending PR. Let's wait `hsts-prune` to remove this ruleset later. thanks!

J0WI:
Please also add:
```
tor.fr33tux.org
wp.fr33tux.org

PR #10079, opened 2017-05-31
active.domains

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/active.domains.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="active.domains">
+	<target host="active.domains" />
+	<target host="my.active.domains" />
+
+	<target host="www.active.domains" />
+
+	<rule from="^http://www\.active\.domains/"
```
J0WI: WFM

galeksandrp: Fixed.

PR #10081, opened 2017-05-31
domainer.ru

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/domainer.ru.xml
+++ b/src/chrome/content/rules/domainer.ru.xml
@@ -1,0 +1,17 @@
+<!--
+ns4.domainer.ru ²
+market.domainer.ru different content
+
+
+² refused
+-->
+<ruleset name="domainer.ru">
+	<target host="domainer.ru" />
+
+	<target host="www.domainer.ru" />
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/domainer.ru.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+ns4.domainer.ru ²
+market.domainer.ru different content
```
J0WI: http gives me a 301 redirect to https

src/chrome/content/rules/domainer.ru.xml
```diff
@@ -0,0 +1,17 @@
...
+
+² refused
+-->
+<ruleset name="domainer.ru">
+	<target host="domainer.ru" />
+
+	<target host="www.domainer.ru" />
+
+	<rule from="^http://www\.domainer\.ru/"
+		to="https://domainer.ru/" />
```
J0WI: Both are working for me over https, so this is not necessary.

J0WI: How do you think about this?

J0WI:
Thanks!

PR #10087, opened 2017-06-01
[BaseKit.com] Remove broken targets

Bisaloo:
I'm not entirely sure what to do here. I usually wouldn't use https-everywhere to fix broken websites but it was already here so...

### DIFF
--- a/src/chrome/content/rules/basekit.com.xml
+++ b/src/chrome/content/rules/basekit.com.xml
@@ -1,13 +1,11 @@
 <!--
 	Invalid certificate:
-		basekit.com
 		custom-css.basekit.com
 		docs.basekit.com
 		resources.basekit.com
 
 	Mixed content:
 		www.basekit.com
-		odense.basekit.com
 
 	Time out:
 		knowledge.basekit.com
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/basekit.com.xml
```diff
@@ -1,79 +1,32 @@
...
-
-	Insecure cookies are set for these hosts: ᶜ
-
-		- odense.basekit.com
-
-	ᶜ See https://owasp.org/index.php/SecureFlag
-
+	Invalid certificate:
+		basekit.com
+		custom-css.basekit.com
```
J0WI: This is an active target

src/chrome/content/rules/basekit.com.xml
```diff
@@ -1,79 +1,32 @@
...
 	<target host="resellers.basekit.com" />
 	<target host="scottishculture.basekit.com" />
-	<target host="www.basekit.com" />
-
-	<!--	Complications:
-				-->
-	<target host="custom-css.basekit.com" />
-	<target host="resources.basekit.com" />
-
+	<target host="try.basekit.com" />
```
J0WI: See also:
```
support.basekit.com
apidocs.basekit.com
knowledge.basekit.com
docs.basekit.com
partners.basekit.com
trivia.basekit.com
```

It seems that customers get a subdomain (which works over HTTPS), so this rule may need a wildcard.

src/chrome/content/rules/basekit.com.xml
Bisaloo: Yes, see the rewrite rule.

src/chrome/content/rules/basekit.com.xml
Bisaloo: > It seems that customers get a subdomain (which works over HTTPS), so this rule may need a wildcard.

Some of them may have mixed content issues (for example https://odense.basekit.com/). Should I still add a wildcard with exclusion and hope it doesn't break too many customers subdomains?

J0WI: As far as I understand this host is a website builder. It's unlikely that there are major differences between all the sites or customers are allowed to insert active content. In this example only Google Fonts are affected.
Let's add the wildcard.

PR #10094, opened 2017-06-01
mastername.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/mastername.ru.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+mastername.ru/: (52, 'Empty reply from server')
+www.mastername.ru/: (52, 'Empty reply from server')
```
J0WI: I was not able to reproduce. If this is a Travis related error, you may add the ruleset to the whitelist.

galeksandrp: Added.

PR #10102, opened 2017-06-02
Re-activate, Rename and Update Duke.edu.xml

cschanaj:
Rewrite `Duke.edu.xml` completely. 

~Is there any tool to convert [`regex`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Duke-University.xml#L98) without wildcards to a list of matched targets?~ It is so difficult to update this ruleset without missing existing targets. #9842 

Remark: sub-domain enumeration is not performed.


src/chrome/content/rules/Duke.edu.xml
```diff
@@ -0,0 +1,147 @@
...
+	<target host="studentaffairs.duke.edu" />
+	<target host="styleguide.duke.edu" />
+	<target host="theaterstudies.duke.edu" />
+	<target host="today.duke.edu" />
+	<target host="www.today.duke.edu" />
+	<target host="travel.duke.edu" />
+	<target host="trinity.duke.edu" />
+	<target host="undergrad.duke.edu" />
+	<target host="undergraduateresearch.duke.edu" />
+	<target host="wiki.duke.edu" />
```
J0WI: I won't annoy you, but I just stumbled over `lists.duke.edu` and `oie.duke.edu`, which are not yet in the list of targets. :)

cschanaj:
@J0WI I have added them back to the list of `target`. thanks!

PR #10118, opened 2017-06-03
[fdopk.org] Update and reactivate ruleset

Bisaloo:
#9842


jeremyn:
- [ ] Remove `webmail` entirely. It's tempting to add the `target` because the plain HTTP version of the page has a login form that isn't protected by HTTPS, but we can't test if it works.
- [ ] Change the `ruleset` `name` from `Farmers Development Organization` to `FDOPK.org`

jeremyn:
Thanks, merged.

PR #10146, opened 2017-06-04
[CNNumerique.fr] Update and reactivate ruleset

Bisaloo:


src/chrome/content/rules/CNNumerique.fr.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Invalid certificate:
+		live.contribuez.cnnumerique.fr
+		tes.cnnumerique.fr
```
J0WI: That one is also listed as a target

J0WI: See also `up.cnnumerique.fr`

PR #10152, opened 2017-06-04
Updated eduroam.org.

Jochen-A-Fuerbacher:


J0WI:
Please add the reference to `eduroam.de.xml`

PR #10162, opened 2017-06-05
Delete MyOpenID.xml

cschanaj:
Remove rules for `*.myopenid.com` as it is gone from DNS. (at least for `^` and `www`)

Related: #9906


jeremyn:
- [ ] Replace the existing `JanRain.xml` entry in the whitelist [here](https://github.com/cschanaj/https-everywhere/blob/da36a3b94589d748800a7b1b45d06b3d54571103/utils/ruleset-coverage-whitelist.txt#L2787), instead of adding a new entry [here](https://github.com/cschanaj/https-everywhere/blob/da36a3b94589d748800a7b1b45d06b3d54571103/utils/ruleset-coverage-whitelist.txt#L6252).

PR #10175, opened 2017-06-05
[Unsearcher.org] Reactivate ruleset

Bisaloo:


jeremyn:
- [ ] Remove the [`securecookie`](https://github.com/Bisaloo/https-everywhere/blob/37f182e6776cac5c827296d532d0dd5ac715bfad/src/chrome/content/rules/Unsearcher.org.xml#L6-L7) tag, because it doesn't look like they use cookies.

PR #10181, opened 2017-06-05
Delete Woahh.xml

Bisaloo:


jeremyn:
https://www.woahh.com actually exists and works with HTTPS, though it returns a `404` with a blank page. But I agree this ruleset should be deleted.

PR #10160, opened 2017-06-05
Rename and Update XE.com.xml, Fix #10030, #10156

cschanaj:
This should close #10030 and close #10156

PR #10184, opened 2017-06-05
Add complicated rule for 'www.xe.com' and 'community.xe.com'

cschanaj:
@jeremyn I added some coverage of working paths for `www` and `community` here. Are you still interested in reviewing this? #10160

### DIFF
--- a/src/chrome/content/rules/XE.com.xml
+++ b/src/chrome/content/rules/XE.com.xml
@@ -25,6 +25,13 @@
 		<test url="http://www.xe.com/gen/css/default.7.css" />
 	<target host="community.xe.com" />
 		<exclusion pattern="^http://community\.xe\.com/(blog|forum|rss-feed|s3|$)" />
+		<!-- match exclusion -->
+		<test url="http://community.xe.com/blog/xe-market-analysis" />
+		<test url="http://community.xe.com/forum/xe-currency-data/api-usage" />
+		<test url="http://community.xe.com/rss-feed" />
+		<test url="http://community.xe.com/s3/files/styles/large/s3/image_1022.jpg?itok=GI0FUOMO" />
+
+		<!-- do not match exclusion -->
 		<test url="http://community.xe.com/community/sites/all/modules/contrib/jquery_update/replace/ui/external/jquery.cookie.js" />
 		<test url="http://community.xe.com/misc/jquery.once.js?v=1.2" />
 		<test url="http://community.xe.com/modules/system/system.menus.css?oqvkiv" />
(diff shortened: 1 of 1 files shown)


jeremyn:
Thanks for working on this after my suggestion in the earlier pull request.

For `community`:

In this specific case, I would instead list the things you want to exclude instead of list the things you want to include. Here's the list of paths after http://community.xe.com that I know of:

/$ , /ar , /blog , /community , /de , /es , /forum , /it , /ja , /misc , /modules , /node , /pt , /rss-feed , /s3 , /sv , /themes , /user

So, I would:

1) get rid of the `rule` you have [here](https://github.com/cschanaj/https-everywhere/blob/ceaf7c6b4683fac2f66be0b8f248fbaa68a99d95/src/chrome/content/rules/XE.com.xml#L40-L41)
2) add `community\.` to the earlier `rule` [here](https://github.com/cschanaj/https-everywhere/blob/ceaf7c6b4683fac2f66be0b8f248fbaa68a99d95/src/chrome/content/rules/XE.com.xml#L21-L22)
3) modify the `exclusion` [here](https://github.com/cschanaj/https-everywhere/blob/ceaf7c6b4683fac2f66be0b8f248fbaa68a99d95/src/chrome/content/rules/XE.com.xml#L30) to `<exclusion pattern="^http://community\.xe\.com/(blog|forum|rss-feed|s3|$)" />`, and modify the `test`s appropriately.

What do you think? We can look at the rest of it once this part is settled.

cschanaj:
done. please let me know if further changes are necessary!

jeremyn:
Thanks. Just some small things:
- [ ] Change the description at the top from `Secure connection redirects to plaintext` to something like `Multiple problems`. For example, https://www.xe.com redirects to plain HTTP, https://www.xe.com/tools.php gets a `404` page which itself has mixed-content blocking, etc.
- [ ] Remove the trailing space at the end of [this line](https://github.com/cschanaj/https-everywhere/blob/ac8e33b200ed5f66072eb6ea30e646378cde9a87/src/chrome/content/rules/XE.com.xml#L41).
- [ ] (Optional) Remove four of the five test cases for `s.xe.com`, unless each test case is testing some specific problem.
- [ ] (Optional) Same for the `do not match exclusion` test cases for `community.xe.com`, delete most of those too.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/10184/commits/28fd601654a316e8176935c8ccb63412bf1eaa17. I added one more extremely minor thing I'd like changed. I'm sorry for not mentioning it before.

cschanaj:
done! thanks!

jeremyn:
Thanks again, merged.

PR #10196, opened 2017-06-06
Re-activate, Rename and Update BEL.fi.xml

cschanaj:
#9842


jeremyn:
Remove entirely:
- [ ] https://lan.bel.fi (authentication error with plain HTTP, fails HTTPS)

Add `target`:
- [ ] https://bacchus.bel.fi and remove from `Timeout`

Also:
- [ ] Change `Couldn't connect to server` to `Connection refused`
- [ ] Change `Timeout was reached` to `Timeout`
- [ ] Change `SSL peer certificate was not OK` to `Invalid certificate`
- [ ] (Optional) I personally like having the comment categories in order, so arrange it so that first is `Connection refused`, then `Invalid certificate`, then `Timeout`.

cschanaj:
@jeremyn 

 - Removed `lan` and moved `bacchus` to the list of `target`. 

 - The `categories` are basically overloaded from `libcurl` error message and their order comes from the `libcurl` error code. I see little value sorting the `categories` alphabetically because it follows the logical order currently. For example, you cannot have a `timeout` before a `refused`, not a `invalid cert` before `timeout`, etc.  Therefore, I would like to keep the current order. 

 - Modified the `category` names.

Thanks.

jeremyn:
I've merged your changes, thanks.

Ordering helps for various reasons, for example it's deterministic, it looks orderly, and it can help when a person is browsing through the file looking for something. About whether "curl"/"time" order or alphabetical order is preferable, I think alphabetical order is easier to remember and understand and explain to casual/novice contributors.

In some cases the time ordering of errors is obvious, for example clearly DNS errors are encountered before invalid certificate errors, but in other cases it's unclear. For example the difference between `Connection refused` and `Timeout` can be the result of [firewall settings](https://serverfault.com/a/521360) on the server, with the problem encountered by the user at the same point in the connection. In this case ordering the errors seems less based on timing and more based on an arbitrary decision by the curl developers to list one type of problem before another.

It's also not clear to me from [the curl documentation](https://curl.haxx.se/libcurl/c/libcurl-errors.html) which browser errors correspond to which curl error. So even if we use that documentation as the source of truth, it's not obvious how to interpret or explain it, to me.

However this is all a relatively minor point, especially for a small ruleset like this, so I merged it anyway.

cschanaj:
> It's also not clear to me from the curl documentation which browser errors correspond to which curl error. 

I agree that `libcurl` errors are not as clear/ obvious as the browser errors. For [example](https://gist.github.com/cschanaj/e0cf854e15d31bdc1a1acc29b504a50c#file-badssl-com-xml-L15), multiple browser errors are grouped together by default (Noted that `incomplete cert chain` fall under  `Peer certificate cannot be authenticated with given CA certificates` by default). 

> So even if we use that documentation as the source of truth, it's not obvious how to interpret or explain it, to me.

It should be better if [`ERRORBUFFER`](https://curl.haxx.se/libcurl/c/CURLOPT_ERRORBUFFER.html) is being used (a.k.a. `curl` command line errors) for categorizations instead, e.g.

```
$ curl https://expired.badssl.com
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.haxx.se/docs/sslcerts.html
```
But this require some inefficient `grep` and `strcmp` to categorize different errors. Since many automated ruleset generators use `libcurl` as the backend AFAIK, I simply overloaded the `libcurl` errors when I made my own. 

I shall change the categorization labels later. For the moment, I would like to keep existing PRs as-is until they are being reviewed.

PR #10204, opened 2017-06-06
Re-activate, Rename and Update SpringFiles.com.xml

cschanaj:
#9842 `*springfiles.nl` seems to be gone from DNS.


src/chrome/content/rules/SpringFiles.com.xml
```diff
@@ -0,0 +1,27 @@
+<!--
+	Non-functional hosts
+		SSL peer certificate was not OK:
```
J0WI: Wildcard DNS

PR #10205, opened 2017-06-07
[Airbnb] Remove right and left wildcards

Bisaloo:


J0WI:
I would keep the wildcards here, because Airbnb owns most TLD, even if they are not all secured. It's unlikely to break any third party site that are not fishy.
We have the same behavior for other high traffic sites like Google.

Bisaloo:
There is no loss of coverage in the modifications I made.

Bisaloo:
Also, the contributing guide says that right wildcards are "*highly discouraged*".

J0WI:
Yes, the current list of TLDs is incomplete. Here are some random examples:
* airbnb.at
* airbnb.be
* airbnb.com.br
* airbnb.co.nz
* airbnb.fi
* airbnb.lu
* airbnb.mx
* airbnb.no
* airbnb.se
* airbnb.co.ve

Bisaloo:
Even though they are included in the wildcard, there was no rule to rewrite those hosts.

J0WI:
Well, yes. There is no regression in this PR. But what are your doubts to not expand the coverage?

Bisaloo:
I know I could merge the two first rewrite rules but I think it is clearer that way.

J0WI:
Well, I'm sure there are many more TLD, but le't go with this for now. Thanks!

PR #10206, opened 2017-06-07
Re-activate and Update North_Devon_Theatres.org.uk.xml

cschanaj:
#9842


jeremyn:
- [ ] Remove the entire top comment. `mx1` and `remote` are probably internal/private subdomains that aren't worth documenting, especially since they don't work.

cschanaj:
Thanks!!

PR #10207, opened 2017-06-07
Delete ReformaBTL.xml

cschanaj:
#9842 Only `^` support HTTPS correctly and redirects to `www.` However, `www` redirects secure connection to plaintext. Delete `ReformaBTL.xml` because of the limited coverage. See https://gist.github.com/cschanaj/10f7f3be8828b645a8f5ab0260c2cd1b


cschanaj:
The ruleset for `*informabtl.com` on `gist` is meant to demonstrate the situation described above. In case an update is preferred over a deletion, I will rebase this PR with the ruleset on `gist`. Sorry for the confusion incurred.

jeremyn:
Okay that's fine. I'd rather delete it though than have such a limited ruleset.

cschanaj:
Thanks!!

PR #10222, opened 2017-06-07
Delete CalChamber.com.xml

Bisaloo:
Gone from DNS.


jeremyn:
This ruleset should be for `calchamber.com` but all of the domains in the ruleset are misspelled `calchamer.com`. For example [this line](https://github.com/EFForg/https-everywhere/blob/4e5c09f5efe3d5d98193c366d46905cc44c09fd9/src/chrome/content/rules/CalChamber.com-falsemixed.xml#L9):

```xml
	<target host="store.calchamer.com" />
```

is almost certainly supposed to be

```xml
	<target host="store.calchamber.com" />
```

We should fix the ruleset instead of delete it.

Bisaloo:
superseded by #11261.

PR #10256, opened 2017-06-08
apteka.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/apteka.ru.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+www.apteka.ru mixed content
```
J0WI: WFM (redirect to ^)

galeksandrp: Added.

PR #10265, opened 2017-06-09
 	Merge, disable WinZip.com related rulesets 

cschanaj:
#9842 Remove rule for `*winzip.com` because of its limited support. 
 - `www` redirects to HTTP on `$` and serves some active mixed content on other paths. 
 - `^` redirects to `www`. 
 - `store` redirects to `support.cleverbridge.com`

See https://gist.github.com/cschanaj/ef7da481b2fb84e3acfe15fc94a11a97 for the status of other non-functional sub-domains. Thanks.


J0WI:
I feel always uncomfortable with removing non-working rulesets like this, because I always hope that they will fix it one day.

cschanaj:
Sadly, it always take a long time (or even forever) for site like this moving to HTTPS. In some case, they even prefer to redirect everything over HTTPS to plaintext in order to avoid the hassle of migrating. 

In this particular case, `www` is causing infinite loop issues on `mixedcontent` platform (see https://www.winzip.com/win/en/). So I guess they are moving away from HTTPS. If you prefer to keep this rule, I can rebase to `default_off` both ruleset in this PR.

J0WI:
I hope that sites are forced to move to HTTPS anyway, because of the penalty by browsers and search engines.

So let's just disable it.

cschanaj:
@J0WI done. thanks!

src/chrome/content/rules/WinZip.com-falsemixed.xml
```diff
@@ -1,10 +1,9 @@
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://winzip.com/ => https://winzip.com/: (60, 'SSL certificate problem: unable to get local issuer certificate')
 	For rules not causing false/broken MCB, see WinZip.com.xml.
 
+	www redirects to plaintext, e.g. https://www.winzip.com/win/en/
```
J0WI: Please add the fact, that this redirect is JS based.

J0WI:
You can still merge the rulesets, they are both default off anyway.

cschanaj:
@J0WI rebased accordingly, thanks!!

PR #10271, opened 2017-06-09
Re-activate and Update Techmeme.com.xml

cschanaj:
#9842

### DIFF
--- a/src/chrome/content/rules/Techmeme.com.xml
+++ b/src/chrome/content/rules/Techmeme.com.xml
@@ -1,17 +1,19 @@
 <!--
 	Non-functional hosts
-		SSL peer certificate was not OK:
+		Invalid certificate:
+			 - alt.techmeme.com
 			 - ftp.techmeme.com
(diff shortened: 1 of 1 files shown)


jeremyn:
- [ ] Put all five problem subdomains in the top comment under the single problem type `Invalid certificate`
- [ ] Add a `target` and `rule` for http://search.techmeme.com [here](https://github.com/cschanaj/https-everywhere/blob/bfc28548a67bebf6835915b6420e87026a15b15a/src/chrome/content/rules/Techmeme.com.xml#L15) like this:

```xml
    <target host="search.techmeme.com" />
 
    <rule from="^http://search\.techmeme\.com/" to="https://www.techmeme.com/search" />
```

cschanaj:
done. thanks!!

jeremyn:
Thanks for making the changes, but `search.techmeme.com` should be in the top comment too, so people know what the special `rule` is for.

jeremyn:
Thanks again, merged.

PR #4936, opened 2016-05-27
Fix #1930

galeksandrp:

PR #4931, opened 2016-05-26
Fix #3073

galeksandrp:
and #3392

PR #10286, opened 2017-06-10
dotabuff.com

galeksandrp:
Issue author: @galeksandrp

### DIFF
--- a/src/chrome/content/rules/dotabuff.com.xml
+++ b/src/chrome/content/rules/dotabuff.com.xml
@@ -1,0 +1,124 @@
+<!--
+www.8t.dotabuff.com ¹
+www.9it.dotabuff.com ¹
+www.de.dotabuff.com ¹
+www.dotamax.dotabuff.com ¹
+www.en.dotabuff.com ¹
+2.en.dotabuff.com ¹
+5.en.dotabuff.com ¹
+oehardgamewww.en.dotabuff.com ¹
+terriblereality.en.dotabuff.com ¹
+www.fr.dotabuff.com ¹
+www.i8t.dotabuff.com ¹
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/dotabuff.com.xml
```diff
@@ -0,0 +1,124 @@
...
+www.webmail.dotabuff.com ¹
+www.www.dotabuff.com ¹
+2.www.dotabuff.com ¹
+bg.www.dotabuff.com ¹
+en.www.dotabuff.com ¹
+es.www.dotabuff.com ¹
+w.gowww.www.dotabuff.com ¹
+ru.www.dotabuff.com ¹
+zh.www.dotabuff.com ¹
+www.zh.dotabuff.com ¹
```
J0WI: They have a wildcard DNS entry

J0WI: And also a wildcard cert

galeksandrp: https://github.com/EFForg/https-everywhere/pull/4936#issuecomment-222307982
https://github.com/EFForg/https-everywhere/pull/4931#issuecomment-222306666

J0WI: I would prefer a comment about this behavior. 4th+ level domain will mismatch anyway.
The target list can be reduced to those hosts that are really in use.
A list of hosts makes no sense here.

galeksandrp: > 4th+ level domain will mismatch anyway
Why would you sure about that? Even on same IP SNI can differ certificates for different domains.

PR #10289, opened 2017-06-10
[Loctudy.fr] Update default_off reason

Bisaloo:
I didn't bother adding more targets. The purpose of this PR is just to help sorting out `defaut_off='failed ruleset test'` rulesets.


jeremyn:
I found these domains on Google: https://loctudy.fr , https://www.loctudy.fr , https://tourisme.loctudy.fr , https://mediatheque.loctudy.fr , https://port.loctudy.fr , and https://sport.loctudy.fr . All of them are mismatched. Since none of the domains work with HTTPS, we should delete the ruleset.

Bisaloo:
I am totally fine with deleting it. I even think we should. But i am a bit unclear on the deletion policy here. Does that mean that all `default_off="mismatched"` rulesets should be deleted?

jeremyn:
I don't know if there's a "policy", but my own opinion is that rulesets are sometimes marked `default_off` instead of deleted because we want to give the site owner some time to fix their site before we just delete the work that a ruleset represents.

How do we know when a ruleset should be deleted after being `default_off`? If the site is still broken the next time someone, like you, happens to look at the ruleset, then it's probably ready for deletion.

PR #10291, opened 2017-06-10
Delete JCore.fr.xml

Bisaloo:


jeremyn:
The site seems okay to me with HTTPS, even if there isn't much there.

PR #10302, opened 2017-06-11
 	Update lighthouseapp.com related rulesets

cschanaj:
#9906


J0WI:
Can we expand this to `*.lighthouseapp.com`?

cschanaj:
@J0WI It seems that only `help` gives certificate error. So, I think we can apply a wildcard `target` here. See https://gist.github.com/cschanaj/3bbefb3500577076aa63b9ac2e88c054.

PR #10309, opened 2017-06-11
Re-activate, Rename and Update HTC.com.xml

cschanaj:
#9906 no sub-domain enumeration is performed.

### DIFF
--- a/src/chrome/content/rules/HTC.com.xml
+++ b/src/chrome/content/rules/HTC.com.xml
@@ -1,8 +1,14 @@
+<!--
+	Invalid certificate:
+		 - htc.com
+-->
 <ruleset name="HTC.com (partial)">
 	<target host="htc.com" />
 	<target host="www.htc.com" />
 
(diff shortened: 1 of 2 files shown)


jeremyn:
- [ ] I have multiple (small) issues with the new ruleset, and because the ruleset is so short, I've rewritten it instead of providing a point-by-point list of issues. Please replace what you have with this:

```xml
<!--
    Invalid certificate:
        htc.com
    
-->
<ruleset name="HTC.com (partial)">
    <target host="htc.com" />
    <target host="www.htc.com" />

    <rule from="^http://htc\.com/"
            to="https://www.htc.com/" />
    <rule from="^http:" to="https" />
</ruleset>
```

cschanaj:
thanks!

jeremyn:
Oops, I had `to="https"` without the `:` in the original version of the checklist, sorry about that. I've fixed it in the checklist so anyone looking at this in the future doesn't use a broken example.

PR #10319, opened 2017-06-11
Create SmackJeeves.xml

nachanon:
This pull request partially enforces HTTPS on the main domain of Smack Jeeves webcomic hosting site ([www.smackjeeves.com](https://www.smackjeeves.com/)), as it has stable TLS support with [cPanel](https://cpanel.com/) certificate for a while now.

Note: Image server ([img2.smackjeeves.com](http://img2.smackjeeves.com/)) used throughout the site is not covered by this rule as it does not support HTTPS; hence "_(partial)_" was added to the rule name.


J0WI:
Thanks!
There are some mixed content issues, but as far as I can see nothing is broken, so LGTM.

PR #10072, opened 2017-05-30
brew.sh: expand existing ruleset

DomT4:

PR #10328, opened 2017-06-12
makecrx.sh: improve support for macOS

DomT4:
An attempt to handle a couple of issues with this script when using macOS, stumbled across in https://github.com/EFForg/https-everywhere/pull/10072#issuecomment-305370157.

It's written in such a way that it shouldn't mess with anything on the Linux side of development but I will spin up a VM a little later today to triple check that.

Locally after this change:
```
~> PYTHONPATH="/private/tmp/https-venv/lib/python2.7/site-packages" makecrx.sh
Building chrome version 2017.6.5
Generating ruleset DB
Validation of included rulesets completed.

Validation of rulesets against relaxng.xml succeeded.
Creating ruleset library...
Removing whitespaces and comments...
chrome/content/rules/default.rulesets passed XML validity test.
Unsigned package has sha1sum: 2386676cfd3157dcb3592697748439866238e909
Total included rules:    22831
Rules disabled by default:     3811
Created pkg/https-everywhere-2017.6.5~pre.crx
```

Or if you don't have a GNU `sed` available:
```
~> PYTHONPATH="/private/tmp/https-venv/lib/python2.7/site-packages" makecrx.sh
Building chrome version 2017.6.5
Generating ruleset DB
Validation of included rulesets completed.

Validation of rulesets against relaxng.xml succeeded.
Creating ruleset library...
Removing whitespaces and comments...
chrome/content/rules/default.rulesets passed XML validity test.
Unsigned package has sha1sum: 2386676cfd3157dcb3592697748439866238e909
Cannot find a GNU sed. Please install one!
```

Theoretically anyone who has run the `install-dev-dependencies.sh` script shouldn't be in a situation where they don't have a GNU `sed` installed, but it's perfectly possible for GNU `sed` to be either `sed` or `gsed` on macOS hence the name logic.

### DIFF
--- a/makecrx.sh
+++ b/makecrx.sh
@@ -98,7 +98,7 @@ trap 'rm -f "$pub" "$sig" "$zip"' EXIT
 # zip up the crx dir
 cwd=$(pwd -P)
 (cd "$dir" && ../../utils/create_xpi.py -n "$cwd/$zip" -x "../../.build_exclusions" .)
-echo >&2 "Unsigned package has sha1sum: `sha1sum "$cwd/$zip"`"
+echo >&2 "Unsigned package has sha1sum: $(openssl dgst -sha1 "$cwd/$zip" | awk '{print $2}')"
 
 # signature
 openssl sha1 -sha1 -binary -sign "$key" < "$zip" > "$sig"
--- a/makecrx.sh
+++ b/makecrx.sh
@@ -115,8 +115,16 @@ crmagic_hex="4372 3234" # Cr24
 version_hex="0200 0000" # 2
 pub_len_hex=$(byte_swap $(printf '%08x\n' $(ls -l "$pub" | awk '{print $5}')))
 sig_len_hex=$(byte_swap $(printf '%08x\n' $(ls -l "$sig" | awk '{print $5}')))
+
+# Case-insensitive matching is a GNU extension unavailable when using BSD sed.
+if [[ "$(sed --version 2>&1)" =~ "GNU" ]]; then
+  sed="sed"
+elif [[ "$(gsed --version 2>&1)" =~ "GNU" ]]; then
+  sed="gsed"
+fi
+
 (
-  echo "$crmagic_hex $version_hex $pub_len_hex $sig_len_hex" | sed -e 's/\s//g' -e 's/\([0-9A-F]\{2\}\)/\\\\\\x\1/gI' | xargs printf
+  echo "$crmagic_hex $version_hex $pub_len_hex $sig_len_hex" | $sed -e 's/\s//g' -e 's/\([0-9A-F]\{2\}\)/\\\\\\x\1/gI' | xargs printf
   cat "$pub" "$sig" "$zip"
 ) > "$crx"
 #rm -rf pkg/crx


makecrx.sh
```diff
@@ -115,8 +115,19 @@ crmagic_hex="4372 3234" # Cr24
...
 pub_len_hex=$(byte_swap $(printf '%08x\n' $(ls -l "$pub" | awk '{print $5}')))
 sig_len_hex=$(byte_swap $(printf '%08x\n' $(ls -l "$sig" | awk '{print $5}')))
+
+# Case-insensitive matching is a GNU extension unavailable when using BSD sed.
+if [[ "$(sed --version 2>&1)" =~ "GNU" ]]; then
+  sed="sed"
+elif [[ "$(gsed --version 2>&1)" =~ "GNU" ]]; then
+  sed="gsed"
+else
+  echo "Cannot find a GNU sed. Please install one!"
```
DomT4: If desired this could be a more explicit instruction to simply go & execute the `install-dev-dependencies.sh` script, which handles this, presuming you use Homebrew on macOS.

DomT4:
> It's written in such a way that it shouldn't mess with anything on the Linux side of development but I will spin up a VM a little later today to triple check that.

Works fine on Ubuntu 17.04:

screen shot 2017-06-14 at 02 31 44

makecrx.sh
```diff
@@ -98,7 +98,7 @@ trap 'rm -f "$pub" "$sig" "$zip"' EXIT
 # zip up the crx dir
 cwd=$(pwd -P)
 (cd "$dir" && ../../utils/create_xpi.py -n "$cwd/$zip" -x "../../.build_exclusions" .)
-echo >&2 "Unsigned package has sha1sum: `sha1sum "$cwd/$zip"`"
+echo >&2 "Unsigned package has sha1sum: $(openssl dgst -sha1 "$cwd/$zip" | awk '{print $2}')"
```
J0WI: We could also use `shasum` to avoid further dependencies.

makecrx.sh
J0WI: I would prefer a BSD/POSIX compatible solution anyway. We could use `awk` or `perl`.

J0WI:
cc @Hainish to hear his thoughts.

makecrx.sh
DomT4: You use `openssl` below to generate the signature so I figured it was reasonably safe to reuse.

makecrx.sh
DomT4: I suspect `awk` would have the same problem in terms of being very feature-limited on BSD-based systems compared to GNU-based systems. There may well be a way with `perl` though.

Hainish: I would encourage installing GNU `sed` as a dependency in `install-dev-dependencies` rather than creating a conditional block here.  Devs really should be running that first thing, before ever building.  If we can, from the perspective of this script, make it so that `sed` is transparently the GNU version, that would be best.  If not, I suppose this block is okay.

DomT4: Thanks for the review.

It's kinda complex on macOS and with Homebrew, which is the only package manager supported by that script. By default `install-dev-dependencies` installs `gnu-sed` which `brew` then installs as `gsed`, so for most people on macOS the GNU `sed` will be called that, but `brew` also offers a `--with-default-names` option to install GNU `sed` as just `sed`. Consequently there's some ambiguity as to what it is likely to be called.

We could trim the `echo` out here & just assume people are running that dependencies script, which seems reasonable, but I think the naming logic is probably something that needs to remain in some shape or form.

@J0WI was probably right that ideally this could be handled by something like `perl` which would be more cross-platform with less risk in terms of unexpected variability, but my `perl` knowledge is pretty awful sadly.

Hainish:
Fair enough.  I'd remove the `echo`, otherwise LGTM.

DomT4:
Change pushed. Let me know if anything else is wanted/needed here & will adjust accordingly.

Hainish:
Thank you for the contribution

PR #10358, opened 2017-06-13
Update FedoraProject.org related rulesets

cschanaj:
#9906 sub-domain enumeration is not performed.


J0WI:
See also `apps.fedoraproject.org`, they maintain a list of some services/subdomains [here](https://github.com/fedora-infra/apps.fp.o/blob/develop/data/apps.yaml).

cschanaj:
`app` is currently HSTS preloaded. Added sub-domains from the list accordingly. thanks!

PR #10366, opened 2017-06-13
Re-activate and Update msu.edu.xml

cschanaj:
#9906 Partial sub-domain enumeration is performed, which produce a 300+ line ruleset I would like to avoid. Only a relatively small portion of sub-domains are included.

### DIFF
--- a/src/chrome/content/rules/msu.edu.xml
+++ b/src/chrome/content/rules/msu.edu.xml
@@ -1,29 +1,21 @@
 <!--
 	Other Michigan State University rulesets:
+		+ Animal_Law.info.xml
+ -->
+<ruleset name="Michigan State University (partial)">
+	<target host="msu.edu" />
+	<target host="www.msu.edu" />
+	<target host="admissions.msu.edu" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/msu.edu.xml
```diff
@@ -1,29 +1,25 @@
...
-     https://mail1.eff.org/pipermail/https-everywhere-rules/2012-April/001096.html
-     -->
-<ruleset name="Michigan State University (partial)" default_off="Have not caught all exclusion patterns">
-    <target host="msu.edu" />
-    <target host="*.msu.edu" />
-
-    <!-- several subdomains do not have a certificate or their certificates are
-         self-signed, so we have to exclude them from the ruleset. -->
-    <exclusion pattern="^http://(?:grad|acadgov|trustees|president|provost|ur|rha)\." />
-    <exclusion pattern="^http://www\.(?:vprgs|vpfo|vps|canr)\." />
```
J0WI: Can we keep the existing exclusions as comments?

PR #10371, opened 2017-06-13
[Warez-BB.org] Update and reactivate ruleset

Bisaloo:
I don't have a test url for `img9.warez-bb.org` but I left it because it was already there and I am assuming the previous contributor checked it was working.


src/chrome/content/rules/Warez-BB.org.xml
```diff
@@ -0,0 +1,24 @@
...
+		img8.warez-bb.org
+		img10.warez-bb.org
+-->
+<ruleset name="Warez-BB.org">
+
+	<target host="warez-bb.org" />
+	<target host="www.warez-bb.org" />
+	<target host="img9.warez-bb.org" /><!-- Needs test url -->
+	<target host="img11.warez-bb.org" />
+		<test url="http://img11.warez-bb.org/wbb3_theme/logos/cellpic_bkg.jpg" />
```
J0WI: test url WFM for img8-12

Bisaloo: Good catch! I didn't think about that.

PR #10373, opened 2017-06-13
[FixingTao.com] Remove broken target

Bisaloo:


src/chrome/content/rules/FixingTao.com.xml
```diff
@@ -1,10 +1,6 @@
 <ruleset name="Fixing Tao.com">
 
-	<!--	Direct rewrites:
-				-->
 	<target host="fixingtao.com" />
-	<target host="www.fixingtao.com" />
```
J0WI: Could you add a comment for it?

PR #10382, opened 2017-06-14
[Tribler.org] Update and reactivate ruleset

Bisaloo:
#9842


jeremyn:
- [ ] https://dispersyreporter.tribler.org
- [ ] https://dl.tribler.org
- [ ] https://old.forum.tribler.org
- [ ] https://kayapo.tribler.org
- [ ] https://svn.tribler.org
  
Add `target`:
- [ ] https://jenkins.tribler.org

Bisaloo:
Do you have a test url for http://dl.tribler.org? I get a test page over `http`.

jeremyn:
I don't have a test URL other than the root page, which looks like a VMware ESXi "Welcome" page. As you know, we normally don't have test URLs for problematic domains in the top comment.

Do you think we should remove this URL entirely because it's both problematic and apparently internal?

Bisaloo:
Yes, I think it should be removed entirely because we have no evidence it is currently used.

jeremyn:
That's fine. Merged.

PR #10383, opened 2017-06-14
[NaughtyDog.com] Update and reactivate ruleset

Bisaloo:
#9842


jeremyn:
- [ ] https://feedback.naughtydog.com
- [ ] https://forums.naughtydog.com

Also:
- [ ] Add `target` for `naughtydog.com` and a `rule` to rewrite it to `www`

jeremyn:
Thanks, however `naughtydog.com` should still be in the top comment, otherwise it's not clear why the special `rule` is needed.

Bisaloo:
I wasn't sure how to explain it. Let me know if you would prefer something different.

jeremyn:
I prefer your original text of `Different content http/https:`.

Bisaloo:
I think we should be a bit more verbose for this case because we don't usually add rules when content differs between `http` and `https`.

jeremyn:
Okay. I think your description is a little wordy. How about:

```
	Different HTTP/HTTPS content:
		naughtydog.com (503 status for HTTPS, 301 to www.naughtydog.com for HTTP)
```

jeremyn:
Thanks, merged.

PR #10385, opened 2017-06-14
[History.com] Merge rulesets and remove wildcard

Bisaloo:
I messed up in #9818 so this is a new PR.


gloomy-ghost:
I get timeout for https://history.com, but since the fetch test passed I am ok with it. However the wildcard securecookie may break something as it secures all cookies under history.com. Can you remove the wildcard securecookie?

PR #10386, opened 2017-06-14
Add rules for ruby-doc.org, update rules for ruby-lang.org

amiryal:
~~Also renamed `Ruby-lang.org.xml` → `Ruby-lang.xml`.~~

~~I hope it’s OK to include both domains in the same file, even though their WHOIS records show different contacts.~~

Update: see [my comment below](https://github.com/EFForg/https-everywhere/pull/10386#issuecomment-309361754), ruby-doc.org got its own rules file, rules for ruby-lang.org got updated.

### DIFF
--- a/src/chrome/content/rules/Ruby-doc.org.xml
+++ b/src/chrome/content/rules/Ruby-doc.org.xml
@@ -1,0 +1,17 @@
+<!--
+	Nonfunctional hosts in *.ruby-doc.org:
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+
+	See also Ruby-lang.org.xml
+-->
+<ruleset name="Ruby-doc.org">
+	<target host="ruby-doc.org" />
+	<target host="www.ruby-doc.org" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
--- a/src/chrome/content/rules/Ruby-lang.org.xml
+++ b/src/chrome/content/rules/Ruby-lang.org.xml
@@ -1,28 +1,43 @@
 <!--
-	Nonfunctional subdomains:
+	Nonfunctional hosts in *.ruby-lang.org:
 
-		- doc *
+		ᵐ doc (redirects to docs)
+		ᵐ redmine (mirrors bugs)
 
-	* Reset
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
 
-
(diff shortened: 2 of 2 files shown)


amiryal:
The tests are failing because of `ruby-lang.org` and `redmine.ruby-lang.org`. Maybe I should have added `ruby-doc.org` in a separate rules file after all.

J0WI:
Could you please fix the errors exposed by Travis?
The affected hosts can be added to the `Nonfunctional subdomains` list. You may also redirect `^ruby-lang.org` to `https://www.ruby-lang.org/` since this is the expected behavior.

amiryal:
I’ve found out that the domains (ruby-lang.org and ruby-doc.org) do not seem to belong to the same entities, so I created a separate file for ruby-doc.org after all.

While I was in the area, I also updated the rules for ruby-lang.org as you suggested.

src/chrome/content/rules/Ruby-lang.org.xml
```diff
@@ -2,28 +2,34 @@
...
 
 -->
-<ruleset name="ruby-lang.org (partial)">
+<ruleset name="ruby-lang.org">
 
 	<target host="ruby-lang.org" />
-	<target host="bugs.ruby-lang.org"/>
-	<target host="redmine.ruby-lang.org"/>
-	<target host="www.ruby-lang.org"/>
+	<target host="*.ruby-lang.org" />
```
J0WI: This will break various subdomains, including redmine and doc.

src/chrome/content/rules/Ruby-doc.org.xml
```diff
@@ -0,0 +1,15 @@
+<!--
```
J0WI: Could you add a "See also Ruby-lang.org.xml" comment?

src/chrome/content/rules/Ruby-lang.org.xml
```diff
@@ -2,28 +2,34 @@
 	Nonfunctional subdomains:
```
J0WI: Could you add a "See also Ruby-doc.org.xml" comment?

src/chrome/content/rules/Ruby-lang.org.xml
amiryal: Right, if redirected blindly to HTTPS on the same domain, they will break on certificate mismatch. In my latest rebase, I chose to add specific redirects for those: redmine → bugs, doc → docs.

I also added the “See also” comments per your suggestion.

J0WI:
Thanks!

PR #10389, opened 2017-06-14
Re-activate, Rename and Update Nruns.com.xml

cschanaj:
#9906


jeremyn:
- [ ] I don't see any cookies so you can remove the new `securecookie` tag

cschanaj:
thanks!!!

PR #10393, opened 2017-06-14
Delete MNGinteractive.com.xml

cschanaj:
#9906 None of `^`, `www`, `secure.extras` or `extras` work over HTTPS


jeremyn:
https://extras.mnginteractive.com works with HTTPS for me, but it just redirects to http://www.digitalfirstmedia.com which times out with HTTPS. The other `mnginteractive.com` domains don't work with HTTPS. So I'm okay with deleting the entire ruleset.

Merged.

PR #10307, opened 2017-06-11
Re-activate, Rename and Update WebType.com.xml

cschanaj:
#9906

PR #10407, opened 2017-06-15
[AV-Comparatives.org] Update and reactivate ruleset

Bisaloo:
#9842


jeremyn:
Please rewrite this without the wildcard. Even if the certificate has a wildcard, individual domains may fail for other reasons.

Bisaloo:
But wildcard DNS is literally the only case where we can cover all possible subdomains only through a wildcard ruleset.

How do you propose we protect this domain otherwise?

jeremyn:
We can protect it by listing the specific domains like we normally do. https://github.com/EFForg/https-everywhere/pull/10307#issuecomment-308918535 is a current example of this.

I personally think there is almost no good use case for wildcards in rulesets. This has been discussed before, unfortunately I can't find a link. I know others disagree.

jeremyn:
Okay I think I get what you are asking. What you do in this case is try to find URLs that are meaningfully different and then protect those specifically. You can do this by seeing which URLs show up in Google searches, and by running Sublist3r and visiting all of the URLs returned. "Junk" URLs like `adsfadsfas.example.com` that are covered by the wildcard just don't get protected.

Bisaloo:
It get the idea behind what you are proposing but IMO, it would make sense only if the website didn't chose this configuration on purpose.

I have always assumed they did that for analytical reasons. They would post `site1.example.com` on site1, `site2.example.com` on site2, etc. You can see it as some sort of referrer.

If what I am describing is true, then we should use a wildcard because it is the only way we can protect subdomains that are created on the fly.

jeremyn:
I'm not sure what you mean by "analytical reasons". Are you talking about https://www.av-comparatives.org in particular or all wildcard certificate sites in general?

Bisaloo:
I was talking about [analytics](https://en.wikipedia.org/wiki/Analytics). Now that you say it, I am not sure `analytical` can be used in that sense...

I am talking about wildcard DNS+certificate sites in general.

jeremyn:
I don't think we can assume a wildcard URL is for analytics. `*.github.io` and `*.wordpress.com` use wildcard certificates but that's not for analytics.

Bisaloo:
This is why I said wildcard **DNS**+certificate.

Bisaloo:
Let's say I am the owner of `av-comparatives.org`. I want to put up a couple of links to www.av-comparatives.org on the web to advertise it to new users. But I would like to be able to know where people found about www.av-comparatives.org. So I post site1.av-comparatives.org on site1, site2.av-comparatives.org on site2. Everybody will reach www.av-comparatives.org in the end but my server log will reveal which link people used to get there. And I will know where my ads have been more effective.

jeremyn:
I get what you're saying now about DNS+certificate, thanks.

That's an unusual way to handle analytics. Most sites are going to use stuff like tokens in the URL, the `referer` tag, or things like that. For `av-comparatives.org` in particular, Sublist3r reports only 19 subdomains including stuff like  `ww` and `wwww`. I don't think they are using most of these domains for analytics.

jeremyn:
Also, https://feedbacksystem.av-comparatives.org times out while http://feedbacksystem.av-comparatives.org works.

jeremyn:
Please let me know when this is ready for a "formal" review.

Bisaloo:
@jeremyn, should be ready now

jeremyn:
Thanks, merged.

PR #10190, opened 2017-06-06
Delete openDesktop.xml

Bisaloo:
That ruleset has been bothering me for a long time. Most domains are broken or for sale. The rest just are mainly redirects.

I will maybe add the few domains that are working in separate rulesets.

PR #10408, opened 2017-06-15
Move armservers.com from WordPress-blogs.xml to its own ruleset

Bisaloo:
Keeping each domain in a separate ruleset in way better for maintaining. Otherwise, we end with half broken rulesets that are so huge and messy that no one wants to update them, as in #10190 or [`Nederland.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Nederland.xml).


Bisaloo:
Before I keep going, what's your take on this @Hainish, @jeremyn, @J0WI?

Also, it's going to be a big PR. Is that okay or should I somehow try to split it in many smaller PR?

jeremyn:
I agree that we should break this ruleset up into new rulesets grouped by the logical relationships between domains. The fact that they are hosted on WordPress is an irrelevant implementation detail.

I definitely recommend multiple PRs. Each PR should create the new ruleset(s) for a group of related domains and also delete the matching domains from the WordPress ruleset. For example there should be a `Create ARMservers.com.xml` PR that creates the `ARMservers.com.xml` ruleset and deletes the `armservers.com` `target`s from the WordPress ruleset.

J0WI:
I think if there are just multiple identical aliases in one ruleset, like `foo.com`, `foo.net` and `foo.org` it's okay. But in the case here and in the other linked examples it's too much and a pain to maintain.

+1 for separate PRs.

Bisaloo:
Okay, I will add PR one by one then so that I don't have to solve conflicts each time you merge a PR.

PR #10439, opened 2017-06-16
[Gitorious.org] Update ruleset

Bisaloo:
Gitorious.org shut down. All other subdomains are gone from DNS.


migueldemoura:
https://gitorious.org/ is still up as a read-only version

PR #10441, opened 2017-06-17
Re-activate, Rename and Update TurnTo.com.xml

cschanaj:
#9906


J0WI:
See also `*.turntonetworks.com`

PR #10442, opened 2017-06-17
Re-activate and Update respekt.cz.xml

cschanaj:
#9906 no sub-domain enumeration is performed.


jeremyn:
The `default_off` message is `Broken login form, may work seamlessly with platform='mixedcontent'`. Did you test logging in?

cschanaj:
I did not test logging in since I do not have an account. `www` redirects to HTTPS with `301`, so I assume it should works well over HTTPS. Should I add a `mixedcontent` attribute?

cschanaj:
@jeremyn `^` redirects (302) to `www` while `www` redirects (301) to HTTPS. So I guess the site it moving to and encouraging HTTPS. IMHO, It should be safe to merge.

jeremyn:
Merged.

cschanaj:
thanks a lot!

PR #10450, opened 2017-06-17
[Pthree.org] Simplify rewrite rule

Bisaloo:


Bisaloo:
I'm confused. I get as mismatch for https://pthree.org/ on firefox but it works fine on chromium.

Bisaloo:
I'm confused. I get a mismatch for https://www.pthree.org/ on firefox but it works fine on chromium.

jeremyn:
It looks like Chrome is okay with a `www` request serving a certificate that doesn't itself contain `www` but only `^`. I think this is known behavior for Chrome, meaning I very vaguely remember encountering this before, but I can't find any reference for it. @gloomy-ghost @Hainish @J0WI can any of you confirm this?

jeremyn:
@Bisaloo In any case, failing in curl or Firefox counts as failing for the purpose of writing rulesets.

Bisaloo:
> @Bisaloo In any case, failing in curl or Firefox counts as failing for the purpose of writing rulesets.

I fixed it instead of closing this issue so that `Pthree.org.xml` gets removed from `ruleset-coverage-whitelist.txt`. I also fixed the regex style.

I am still interested in knowing the cause of this.

Hainish:
We just tested this with a domain we own using Let's Encrypt.  Indeed, if there's a cert misconfiguration and `www.` is a missing subject, Chrome auto-redirects to `.`.  This shows up in the developer console as `cancelled` since it never gets to the HTTP layer, it's cancelled immediately upon recognizing the invalid certificate.

PR #10463, opened 2017-06-17
Delete Pro186.com.xml

Bisaloo:
https://pro186.com/ returns `HTTP 200` but would need a test url to be included.


jeremyn:
For these pull requests that delete `default_off` rulesets, I consider deleting the disabled ruleset to be a different step from creating a new ruleset for related domains. So for example the only `target` this ruleset has is https://cdn.pro186.com , so that's all I need to check, and that domain is not in DNS so I approved the deletion. Whether https://pro186.com supports HTTPS or not doesn't matter to me. Maybe this will save you some time going forward.

PR #6771, opened 2016-09-06
add more hosts to swissgov.xml

J0WI:

PR #10468, opened 2017-06-17
separate ag.ch

J0WI:


J0WI:
@jeremyn as requested in #6771 this is the first split-off.

jeremyn:
- [ ] I think the `ruleset` `name` should be `ag.ch`, not `Kanton Aargau`. I prefer using domain names instead of descriptions for `ruleset` `name`s because descriptions can require language and cultural knowledge that the user may not have and shouldn't need.
- [ ] I don't think `ag.ch.xml` and `swissgov.xml` should point at each other. I understand there is precedent, for example in [`US-government.xml`](https://github.com/EFForg/https-everywhere/blob/9bf49cbeb4696cdea76df23f685da1d11031f693/src/chrome/content/rules/US-government.xml), but this is a bad decision. Federal, state, and city jurisdictions in the US are often entirely independent from each other and so there is no reason that someone looking at for example [`Colorado_Secretary_of_State.xml`](https://github.com/EFForg/https-everywhere/blob/9bf49cbeb4696cdea76df23f685da1d11031f693/src/chrome/content/rules/US-government.xml#L39) needs a chain of references to [`City_of_Seattle.xml`](https://github.com/EFForg/https-everywhere/blob/9bf49cbeb4696cdea76df23f685da1d11031f693/src/chrome/content/rules/US-government.xml#L35). Maybe Swiss cantons have a tighter relationship than US jurisdictions but if not then we shouldn't have these references.

J0WI:
The reason for linking this with swissgov is [this](https://github.com/EFForg/https-everywhere/pull/10468/files#diff-e9b2273115483f1defc7b3a3468f5c39R10) list that fits in no other ruleset.

The reason for the ruleset name is, that [some](https://github.com/EFForg/https-everywhere/pull/10468/files#diff-e9b2273115483f1defc7b3a3468f5c39R317) targets use another domain that the official [shortcut].ch

jeremyn:
I'm saying `ag.ch.xml` shouldn't be linked to `swissgov.xml`, so a list in `swissgov.xml` isn't relevant. But for what it's worth, if we were reviewing `swissgov.xml`, I would also argue that that list should be removed from it.

All the targets in `ag.ch.xml` use the domain `ag.ch`.

J0WI:
So where should the list be?

I want to have a consistence in related ruleset, so if we name this `ag.ch` how do we name rulesets that contains more than one domain?

jeremyn:
I'm saying the canton rulesets/domains aren't related to the federal ruleset, so they shouldn't be linked to the federal ruleset, and that list shouldn't be there. Where that list should be is unrelated. There are thousands/millions of other unrelated domains that also don't support HTTPS, and I wouldn't put them in `swissgov.xml` either.

The ruleset name depends on what's in the ruleset but I think where possible the ruleset name should be the domain name.

jeremyn:
@J0WI I want to move on with this PR somehow.

I disagree with the `ag.ch.xml` <-> `swissgov.xml` link but if you really want it here, then I'm okay with keeping it.

I do however think that the `ruleset` `name` for `ag.ch.xml` should be `ag.ch`, not `Kanton Aargau`. I don't understand your counterargument in https://github.com/EFForg/https-everywhere/pull/10468#issuecomment-318632367.

J0WI:
Okay I've chaged the ruleset name.
I'd like to keep the comment. IMHO more documentation than required shouldn't be punished.

jeremyn:
Merged, thanks.

PR #10474, opened 2017-06-18
Re-activate and Update ThinkProgress.org.xml

cschanaj:
#9906 this site somehow do not work really well on Firefox...


J0WI:
WFM on Firefox.

PR #10486, opened 2017-06-18
Add Raiffeisen.at.xml

J0WI:


jeremyn:
Could you please be more explicit in the domains at the top? It makes testing easier. For example:

```xml
<!--
	Nonfunctional hosts in *.raiffeisen.at:
		- raiffeisen.at (h) 
		- www.raiffeisen.at (h) 
		- finanzieren.raiffeisen.at (m) 
		- www.finanzieren.raiffeisen.at (m) 
		- meinbanking.raiffeisen.at (m) 
		- www.meinbanking.raiffeisen.at (m) 
		- vorsorgen.raiffeisen.at (m) 
		- www.vorsorgen.raiffeisen.at (m) 
		- wohnen.raiffeisen.at (m) 
		- www.wohnen.raiffeisen.at (m) 

...
-->
```

Also I don't know what to test specifically for  `[region]` in `(www\.)?[region].raiffeisen.at (f)`. If you could list all the regions or at least give some examples that would be good.

I also got time outs for some of the top comment items, such as https://finanzieren.raiffeisen.at , which you have listed as a mismatch. Please do a general review of the entire PR, and then I'll review it.

J0WI:
Rebased.

jeremyn:
- [ ] https://raiffeisen.at seems okay, you can make it a `target` and remove the comment. (It seems to always redirect to `www`.)
- [ ] https://filialen.raiffeisenbank.at is `(f)`, not `(m)`.
- [ ] After more thought I think we should leave out the `www.[region]` and `salzburg` comments entirely. Since the ruleset is `(partial)`, we don't need them. The problem is that it requires local and/or German knowledge to know whether a specific word is a region or not, for example `vorsorgen` or `gastein`, so it can be hard to tell whether a particular domain is classified correctly.

J0WI:
> The problem is that it requires local and/or German knowledge to know whether a specific word is a region or not, for example vorsorgen or gastein, so it can be hard to tell whether a particular domain is classified correctly.

That's exactly why such a comment is useful. I just stored the result of my own little research in the comments. We already have this practice with Chinese, Russian and many other language specific sites. Otherwise future contributors have to do some research their self to figure out whether the is a systematic HTTPS support or not.

jeremyn:
Maintainers should be able to verify everything in a ruleset. We also shouldn't show favoritism to certain languages or cultures, except for allowing English as the working language of the project, so this means changes should be straightforward and standardized where possible. Rulesets are not the places for personal research notes.

This is a security issue, otherwise you can imagine malicious contributors sneaking in confusing or hostile stuff that reviewers don't see due to language or cultural barriers.

It's also a maintenance issue. I think comments need to be verified like code, except comments are worse because they aren't checked by a program except in a few specific cases. So comments should be minimized and restricted. We should also imagine the ruleset being reviewed years from now by other people, wondering what a comment means or whether it's important.

jeremyn:
This PR is ready to go other than the discussion we're having about the comment. It would be nice if this could go out with the release scheduled for later this week.

src/chrome/content/rules/Raiffeisen.at.xml
```diff
@@ -5,6 +5,7 @@
 		- [region].raiffeisen.at (f)
 		- www.[region].raiffeisen.at (f)
 			- e.g. salzburg.raiffeisen.at
+			  more examples on http://www.raiffeisen.at/oesterreich/1006622610903_359752870146184405-359752870146184405-NA-30-NA.html
```
J0WI: @jeremyn that's all I found. =/

jeremyn:
From your list at http://www.raiffeisen.at/oesterreich/1006622610903_359752870146184405-359752870146184405-NA-30-NA.html there are only four of these regional subdomains: `bgld`, `noe`, `salzburg`, and `stmk`. Let's just list them individually and remove the `[region]` comment.

J0WI:
Well, the list is not complete. If you search on Google you will find much more.

jeremyn:
Okay, well in that case I think we should delete all the region stuff, as I said in the checklist above. Or, you can include the region subdomains that you have tested. We really shouldn't be saying that a bunch of URLs that we haven't tested fail for some reason anyway.

jeremyn:
@J0WI I'd like to do something with this one way or another. Should we get one of the other maintainers/staff in here to cast a "tie-breaking vote"?

J0WI:
Yes I also think we got stuck.
@gloomy-ghost could you help out here?

gloomy-ghost:
When we separate regional subdomains from the others, we assume these subdomains are used for the same service, and are supposedly handled using the same logic. Under this assumption, listing them as a separate category can help to update the ruleset when https availability is changed. 

As a maintainer who mainly _review_ the changes to a ruleset, I don't really mind comment notes intended to provide some information that might help future contributors _updating_ a ruleset. In other words, I'm ok with it as long as the information provided is true, but I won't specifically ask for it when it's not there.

jeremyn:
I understand @gloomy-ghost to be saying that if some identifiable group of subdomains behave in the same way, then you can simply name the group in the comment when describing its behavior. Maybe a good example of this is something like `cdn01.example.com` through `cdn99.example.com`. I would agree that someone could just test a few of these, and if all the test cases have for example an invalid certificate with HTTPS and a default Apache page for plain HTTP, you could just say `cdn[01..99].example.com` as `Invalid certificate` in the top comment. It would the same if the group were "`jan`, `feb`, `mar`..." or something else that is an obvious grouping.

In this case though the category is `region` and I don't think that is clear. (Region of where? Austria?) As I say in the checklist, unless you are familiar with the language and culture you would not be able to tell at a glance whether `vorsorgen` or `gastein` is a region or not. (`vorsorgen` is not, `gastein` is.) So a maintainer who is not familiar with the language and culture who is skimming Google or Sublist3r and sees `vorsorgen` will not be able to tell whether it is covered by the `region` comment or not.

J0WI:
We already had a quite long discussion here and we both agreed to a "tie-breaking vote".

jeremyn:
@J0WI But @gloomy-ghost did not say who they agreed with, instead they wrote some text that was not directly related to the problem we're discussing. They argue that grouping is okay to do, and I agree that groups are sometimes fine, but my problem here is that the group itself is unclear and depends on language/cultural knowledge.

I don't want to be a bad sport. If @gloomy-ghost clearly says that they support the current version of the pull request, meaning they "vote" with you, then I'll merge this without further complaint. (Though I won't consider it "binding" on future pull requests.)

gloomy-ghost:
I think we can merge it as is, since it's intended to note information for contributors who are likely to update this partial ruleset in the future. We will need to verify all targets anyway if we are going to remove the partial flag.

jeremyn:
All right, thanks both of you. Merged.

J0WI:
Thanks!

PR #10495, opened 2017-06-19
 	Re-activate and Update Avropa.se.xml 

cschanaj:
#9906

### DIFF
--- a/src/chrome/content/rules/Avropa.se.xml
+++ b/src/chrome/content/rules/Avropa.se.xml
@@ -4,4 +4,3 @@
 
 	<rule from="^http:" to="https:" />
 </ruleset>
-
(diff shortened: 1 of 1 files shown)


jeremyn:
This one works for me. What's wrong with it?

cschanaj:
I was testing a newly created script to identify `default_off` rulesets which all `target` are not working over HTTPS. `^` and `www` are `reset` over HTTPS with my script. 

However, it seems to be a false-positive case. `curl` does not show any type of errors on both `^` and `www`. I should be double checked before submitting this PR. 

Rebase this PR to re-activate the ruleset. thanks!!

jeremyn:
- [ ] Please delete the [empty line](https://github.com/cschanaj/https-everywhere/blob/7e4d89b3afbc72b2f2cc79cb1eccc4c6fbd9c0be/src/chrome/content/rules/Avropa.se.xml#L7) at the end

cschanaj:
@jeremyn done. thanks!

jeremyn:
Thanks, merged.

PR #10520, opened 2017-06-20
Delete multiple default_off rulesets (^V)

cschanaj:
#9906 https://github.com/EFForg/https-everywhere/pull/10516#issue-237085032


galenide:
With regard to the `Volunteer_Centers_of_Michigan.xml` ruleset, the following may be of interest. According to [this HTTP status code checker](https://httpstatus.io/), the URLs http://mivolunteers.org/ and http://www.mivolunteers.org/ each generate a 302 redirect to `/` and then a 301 redirect to `https://mnaonline.org/civic-engagement/volunteer-centers-of-michigan` (In some cases, the URL http://www.mivolunteers.org/ may generate a single 301 redirect to `https://mnaonline.org/civic-engagement/volunteer-centers-of-michigan`.) At the same time, changing the ruleset so that requests to the `mivolunteers.org` and `www.mivolunteers.org` domains are redirected to `https://mnaonline.org/civic-engagement/volunteer-centers-of-michigan` may or may not work well, particularly if things change in the future. (Another possibility might be to extend the [Michigan_Nonprofit_Association.xml](https://github.com/EFForg/https-everywhere/blob/release/src/chrome/content/rules/Michigan_Nonprofit_Association.xml) ruleset so that it also handles requests to the `mivolunteers.org` and `www.mivolunteers.org` domains.)

cschanaj:
@galenide As far as I am concerned, `^` and `www` does not work on `/` (`timeout`) over HTTPS so I removed the ruleset `Volunteer_Centers_of_Michigan.xml`. Personally, I do not like complicated rewrite rule since they are likely to cause problem in the long term. 

In this case, `mivolunteers.org` redirects to another domain `mnaonline.org` over HTTP. So, I prefer to a simple `rm` here.

PR #10524, opened 2017-06-20
[Worldreader.org] Update ruleset

Bisaloo:


src/chrome/content/rules/Worldreader.org.xml
```diff
@@ -1,31 +1,19 @@
...
-	*Refused
-
-
-	Problematic hosts *worldreader.org:
-
-		- cdn *
-
-	* Cloudfront
+	Time out:
+		bot.worldreader.org
```
J0WI: m.worldreader.org

PR #10528, opened 2017-06-20
Move WordPress.tv from WordPress-blogs.xml to its own ruleset

Bisaloo:


J0WI:
@Bisaloo maybe you could prioritize those hosts in `WordPress-blogs.xml` that causes the `top-1k` label.

PR #6035, opened 2016-08-01
Updates to LinuxFoundation subdomains

emsearcy:
- not all paths on the go.linuxfoundation.org vanity domain work on go.pardot.com
- eventsstg.linuxfoundation.org is a private site behind HTTP auth and is not testable
- LDN has been gone for several years
- events.linuxfoundation.org now only redirects to HTTP for / and /events/*
- collabprojects supports SSL (though is just a redirect to www.linuxfoundation.org now)
- www.linuxfoundation.org is HTTPS only with HSTS; no more mixed-SSL and http:// redirects
- developerbugs.linuxfoundation.org now provides "secure" flag in cookie

PR #10537, opened 2017-06-20
Update LinuxFoundation.xml

jeremyn:
This pull request includes changes from PR https://github.com/EFForg/https-everywhere/pull/6035, but that PR became complicated. The primary intended changes here are to:

* remove references to `eventsstg`
* remove `exclusion`s for `www`
* remove `exclusion`s for `collabprojects`
* update `exclusion`s for `events`
* update `go.linuxfoundation.org` handling

There are also minor changes like reordering items and removing comments.

### DIFF
--- a/src/chrome/content/rules/LinuxFoundation.xml
+++ b/src/chrome/content/rules/LinuxFoundation.xml
@@ -66,7 +66,9 @@
 		<test url="http://go.linuxfoundation.org/e/6342/TheLinuxFoundation/ymc8q/578968329" />
 		<test url="http://go.linuxfoundation.org/l/6342/2016-03-23/33c3kc" />
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http://linuxfoundation\.org/" to="https://linuxfoundation.org/" />
+	<rule from="^http://(www|admin|automotive|collabprojects|developerbugs|events|identity|ldn|lists|lsbbugs|training)\.linuxfoundation\.org/"
+		to="https://$1.linuxfoundation.org/" />
+	<rule from="^http://lists\.linux-foundation\.org/" to="https://lists.linux-foundation.org/" />
 
 </ruleset>
--- a/src/chrome/content/rules/LinuxFoundation.xml
+++ b/src/chrome/content/rules/LinuxFoundation.xml
@@ -20,17 +20,10 @@
 
 	Partially covered hosts in *linuxfoundation.org:
 
-		- collabprojects ʰ
 		- events ʰ
-		- eventsstg ʰ
 		- www ʰ
 
 	ʰ Some pages redirect to http
-
-
-	Insecure cookies are set for these domains:
-
-		- developerbugs.linuxfoundation.org
 
 -->
 <ruleset name="Linux Foundation.org (partial)">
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/LinuxFoundation.xml
```diff
@@ -20,18 +20,11 @@
 
 	Partially covered hosts in *linuxfoundation.org:
 
-		- collabprojects ʰ
 		- events ʰ
-		- eventsstg ʰ
 		- www ʰ
```
J0WI: Is this comment still up to date? There are no exclusions for www.

src/chrome/content/rules/LinuxFoundation.xml
```diff
@@ -40,100 +33,39 @@
 	<target host="lists.linux-foundation.org" />
```
J0WI: Can we expand the coverage for `linux-foundation.org`?

J0WI: Same for `linuxfoundation.org`, but therefore this can be done in a follow up PR.

src/chrome/content/rules/LinuxFoundation.xml
jeremyn: I removed `www` in https://github.com/EFForg/https-everywhere/pull/10537/commits/d16b3613b69abb6d8f35ae28c3cb15cb37e1d0d1.

src/chrome/content/rules/LinuxFoundation.xml
jeremyn: This PR is just bringing over changes from PR https://github.com/EFForg/https-everywhere/pull/6035. I don't want to get into expanding coverage beyond that.

emsearcy:
Is the final rule for http->https is going to attempt to fetch go.linuxfoundation.org paths other than / and /[el] over https? (there is no valid https cert for that domain)

As an example URL, what is the behavior of http://go.linuxfoundation.org/sysadminday2016 with this pull request? It should not redirect to https://go.linuxfoundation.org/sysadminday2016 nor to https://go.pardot.com/sysadminday2016

The two methods I attempted to handle this were first the negative lookahead:
```<exclusion pattern="^http://go\.linuxfoundation\.org/(?![el]/)[^?]" />```
...and second by moving go.linuxfoundation.org to a different xml so that it wouldn't be affected by the final rule.

emsearcy:
In case it's helpful, I've created #10565 which contains the same changes as #6035 but without the last commit, so that the only affected domains are Linux Foundation ones (this had been one of the latter requests on that thread, to take out that last commit).

jeremyn:
@emsearcy Yes, the general rule at the end was too broad for `go.linuxfoundation.org`, thanks for the catching that. I've replaced it with more targeted `rule`s.

jeremyn:
@emsearcy Regarding your other PR https://github.com/EFForg/https-everywhere/pull/10565, I think that `X vanity domains` is an antipattern because domains should be grouped by logical relationships (related organizations) rather than underlying implementation (Pardot, Bitly).

A motivation for this new PR is that our original discussion in PR https://github.com/EFForg/https-everywhere/pull/6035 dragged on, which I admit was partly my fault. I'm trying to move this firmly out of my queue and so I don't want to restart the discussion here.

That said, you are welcome to exactly replace the changes in either of your PRs with the changes from this PR and then @J0WI can merge that PR in favor of this one, if @J0WI wants to. That way you can get the credit for the changes, if you want.

emsearcy:
No worries; I understand your concerns about antipattern and won't bother restarting that discussion! I've closed 10565, too. This one looks good to me.

Issue #10378, opened 2017-06-14
Progress #9906


cschanaj:
Type: other

The statistics will be updated for  **the first release every month** ~odd releases and the latest release since 5.2.15~  until #9906 is closed. See [`simple-statistics.go`](https://github.com/cschanaj/lambda/blob/master/src/simple-statistics.go) for the source code used.

## `ruleset`
| `tag` | `platform` | `#platform` | `default_off` | `#default_off` |
| :---: | :---: | :---: |  :---: |  :---: | 
| 2017.12.6 | 1208 | 2 | 5675 | 104 |
| 2017.11.21 | 1202 | 2 | 5724 | 104 |
| 2017.10.4 | 1210 | 2 | 6199 | 110 |
| 2017.9.12 | 1210 | 2 | 6259 | 118 |
| 2017.8.15 | 1232 | 2 | 6351 | 120 |
| 5.2.21 | 1272 | 3 | 6564 | 131 |
| 5.2.19 | 1279 | 3 | 3647 | 134 |
| 5.2.17 | 1298 | 3 | 3893 | 182 |
| 5.2.15 | 1354 | 3 | 3980 | 198 |

## `platform`
| `tag` | `mixedcontent` | `cacert` | `cacert mixedcontent` | 
| :---: | :---: | :---: |  :---: |
| 2017.12.6 | 1194 | 14 | 0 |
| 2017.11.21 | 1187 | 15 | 0 |
| 2017.10.4 | 1194 | 16 | 0 |
| 2017.9.12 | 1193 | 17 | 0 |
| 2017.8.15 | 1210 | 22 | 0 |
| 5.2.21 | 1234 | 36 | 2 |
| 5.2.19 | 1239 | 38 | 2 |
| 5.2.17 | 1255 | 41 | 2 |
| 5.2.15 | 1244 | 104 | 6 |

## `default_off`
| `tag` | `failed ruleset test` |
| :---: | :---: |
| 2017.12.6 | 3955 |
| 2017.11.21 | 3967 |
| 2017.10.4 | 4276 |
| 2017.9.12 | 4313 |
| 2017.8.15 | 4384 |
| 5.2.21 | 4561 |
| 5.2.19 | 1620 |
| 5.2.17 | 1706 |
| 5.2.15 | 1735 |

## `target`
| `tag` | `target` | `#target` | `target (*.)` | `target (.*)` | duplicated `target` |
| :---: | :---: | :---: |  :---: |  :---: | :---: |
[...]

PR #10538, opened 2017-06-21
Disabling rulesets that fail the comprehensive fetch test across mult…

Hainish:
…iple locations (Singapore, London, San Francisco) and across the span of a week


jeremyn:
@Bisaloo and @cschanaj have been patiently going through existing `default_off` rulesets. I strongly suggest we get their input before merging in 5000 more of these rulesets. I'm not sure what system they are using to review the existing `default_off` rulesets but we shouldn't interfere with that without their input.

jeremyn:
:frowning: please revert your changes until they can respond.

jeremyn:
For example, several recent PRs now require conflict resolutions, for example https://github.com/EFForg/https-everywhere/pull/10523 and https://github.com/EFForg/https-everywhere/pull/10518.

Hainish:
It's fine to overwrite the changes here if there are more granular fixes to be made.

I've announced that this was happening numerous times.  We've done this in years past, as well.

jeremyn:
Five days ago @cschanaj wrote in https://github.com/EFForg/https-everywhere/issues/9906#issuecomment-308930531 that:

> it is fine if someone create `default_off` ruleset as usual at the moment, *as long as it is not a massive update* like https://github.com/EFForg/https-everywhere/pull/9884

I'm not sure where you announced this would happen but it would have been helpful for you to coordinate your efforts in issues https://github.com/EFForg/https-everywhere/issues/9582 , https://github.com/EFForg/https-everywhere/issues/9842 , https://github.com/EFForg/https-everywhere/issues/9906 and maybe others.

These volunteers have done a tremendous amount of work over months working on these `default_off` rulesets and I think it is entirely reasonable for you to pull back your changes for now and try to minimize the burden on them.

cschanaj:
This PR doubled the number of `default_off` ruleset, see https://github.com/EFForg/https-everywhere/issues/10378#issue-235757362. Around 30% of the rulesets are `default_off` now, I don't think this is ideal at all. Can you modify the script to delete ruleset as in https://github.com/EFForg/https-everywhere/pull/10516#issue-237085032?

 Besides, I think you should whitelist (at least for `alexa-top-1k`), see some popular site like [`9gag.com.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/9gag.com.xml) (`alexa top-189`) got disabled.

cschanaj:
@Hainish This is also failing Travis builds, see #10540

Hainish:
> Around 30% of the rulesets are default_off now, I don't think this is ideal at all.  Can you modify the script to delete ruleset as in #10516 (comment)?

@cschanaj This used to be around 10%.  This means that at least 20% of rulesets were enabled by default when they were causing problems with at least one of their targets.  Although obviously the majority of the changes here are moving from enabled to disabled, there are some rulesets that have been re-enabled after long having been disabled.  Most of these only have one or a few hosts.  If we delete rulesets rather than disabling them, this 1) makes it so that users can't manually re-enable these rulesets, and 2) makes it so that future runs of the `full-fetch-test` tool will not automatically re-enable them, if these sites fix their TLS configuration.  However, if we decide this is causing unnecessary clutter in terms of stale rulesets never being cleared out, we can definitely modify the script and remove them.

> Besides, I think you should whitelist (at least for alexa-top-1k), see some popular site like 9gag.com.xml (alexa top-189) got disabled.

It seems important that the fetch test *not* whitelist rulesets that are in the alexa top 1k, for the very fact that HTTPS Everywhere may well be causing problems for these sites.  Your example is one with many hosts and just one failure.  But what happens when we have one of the top 1k sites with only a few hosts, and it is causing a failure?  We won't necessarily see it unless someone reports it.  Better, in this case, to disable that rule completely until someone can take a look at it.

> @Hainish thephoenixbay.com.xml is failing Travis due to this PR.

Fixed this issue and tests seem to be passing now.

cschanaj:
@Hainish thanks for fixing Travis!

Bisaloo:
I was expecting a specific issue before that happens but okay. For the record and for the next time you run this : 

- it would be good if the tests could simply comment the broken targets. It is not always necessary to disable the entire ruleset. It may take very long for contributors to go through disabled rulesets so it may result in an unnecessary loss of coverage. Now that we are removing most wildcards ruleset, I think it is a realistic option.

- it would be useful to be able to identify the disabled rulesets *on a specific iteration of the tests* by a simply `grep` or something. So instead of `default_off='failing ruleset tests`, `default_off='failing ruleset tests $DATE` would be nice.

Hainish:
@Bisaloo I will attempt to modify the script to implement (1).  For (2), great idea as well.

cschanaj:
@Hainish  I ran the following command and noted that around 900+ rulesets failed the `fetch` test more than once having multiple `Disabled by https-everywhere-checker because: ...`. Is this expected? For example, [`Cafegate.com.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Cafegate.com.xml) and [`ADP_Retirement_Services.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ADP_Retirement_Services.xml)

```bash
user@host:~/https-everywhere/rules$ for file in *; do  TMP=$(grep 'Disabled by https-everywhere-checker because' $file | wc -l); if [ $TMP -gt 1 ]; then echo $file; fi; done > ~/failed_more_than_once.txt
```

Bisaloo:
Also, as @jeremyn pointed it out, it would have been really useful for us to know when this was going to happen. It's true you mentioned it in multiple issues but I had no idea it would be coming so soon.

If you could open an issue to announce it next time, as you do for the releases, I would really appreciate it.

cschanaj:
I agree with @Bisaloo, It could be better if it could be announce, say, 2-3 weeks (even 1-2 months) before, since a quite number of my PRs get merge conflict now. It is sort of annoying to resolve them all.

cschanaj:
IMO #10545 might worth some attentions.

Hainish:
> @Hainish I ran the following command and noted that around 900+ rulesets failed the fetch test more than once having multiple Disabled by https-everywhere-checker because: .... Is this expected? For example, Cafegate.com.xml and ADP_Retirement_Services.xml

This is because currently there is no graceful way to remove comments from previous runs of the disabling `fetch-test` tool.

@Bisaloo In the future I'd like to have a this happen continuously, say every week.  We've gotten *lots* of reports to the effect of "HTTPS Everywhere breaks my site" and these issues sometimes linger for months.  That's only the ones that are actually reported.  If we're able to have a good system in place which keeps the ruleset coverage accurate on a continual basis, those problems wouldn't arise as much.  Plus, every diff would be much much smaller - note that the last time this was run was two years ago.  This will result in very few PRs resulting in merge conflicts.

We have to account for temporary and regional outages, as well.  My ideal schedule would be:

1. Monday: San Francisco `full-fetch-test` run
2. Tuesday: Singapore `full-fetch-test` run
3. Wednesday: UK `full-fetch-test` run

and only disable those rulesets (or targets, if we change the script) that are problematic for each one of them.  This is pretty much what we did here, but not on such a strict schedule.

Resolving the merge conflicts should be easy.  The more granular changes that you've been working on in pull requests should always overwrite the `full-fetch-test` disables.

Bisaloo:
> @Bisaloo In the future I'd like to have a this happen continuously, say every week. We've gotten lots of reports to the effect of "HTTPS Everywhere breaks my site" and these issues sometimes linger for months. That's only the ones that are actually reported. If we're able to have a good system in place which keeps the ruleset coverage accurate on a continual basis, those problems wouldn't arise as much. Plus, every diff would be much much smaller - note that the last time this was run was two years ago. This will result in very few PRs resulting in merge conflicts.

This sounds good! It's also worth noticing that we sometimes get issues from users saying that HTTPS-Everywhere is breaking a website and that the webmasters obviously know it because there is a warning on the page 'Please disable HTTPS-Everywhere to use this site'. But no one bothered reporting it... So I am totally with you on the fact that we can't trust reports to detect broken websites.

Bisaloo:
What's the deal with [`GitLab.com`](https://github.com/EFForg/https-everywhere/blob/9bf49cbeb4696cdea76df23f685da1d11031f693/src/chrome/content/rules/GitLab.com.xml)? I would think that 308 subdomains are supposed to pass the tests.

Hainish:
@Bisaloo https://github.com/efforg/https-everywhere/blob/bb84ab16a4d95134c72aae487d9ce7fb1410ddd0/test/rules/src/https_everywhere_checker/check_rules.py#L144 is the relevant line here.  If the `http` endpoint results in a `2XX`, and the `https` endpoint does not, then count this as a failure.

Looks like the response codes have changed since I ran the test, though.  http://www.gitlab.com/ returns a `301` for me currently.

Bisaloo:
https://travis-ci.org/EFForg/https-everywhere/jobs/245518072

Hainish:
https://github.com/efforg/https-everywhere/blob/bb84ab16a4d95134c72aae487d9ce7fb1410ddd0/test/rules/src/https_everywhere_checker/http_client.py#L379 seems to be the culprit.  308 redirects are not followed.  

This is a relatively new HTTP code, standardized in 2015: https://tools.ietf.org/html/rfc7538

This has been fixed in https://github.com/EFForg/https-everywhere/commit/c284a7cbaa4c2849fe4f619e1afa55a0f595de1f and I've restarted the fetch test.

Hainish:
@Bisaloo I think you need to rebase for this to take effect.

cschanaj:
@Hainish I think that `target` with `Could not resolve host` error can be safely removed from the ruleset during the fetch test, e.g. [`PayPal.xml#L4`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/PayPal.xml#L4). If a `target` is gone from DNS, there is little reason to keep it.

Hainish:
@cschanaj several of the `.onion` sites that we support resulted in a `Could not resolve` due to lack of Tor in our testing environment.  This is another reason why we want to have a whitelist that covers fetch tests (per https://github.com/EFForg/https-everywhere/issues/10227):

https://github.com/EFForg/https-everywhere/commit/77ee5d282d7e8f8f9abaed6f2baa8582b84ec5e4

Bisaloo:
Just FYI, the tests led to false positive because they ran on an outdated version of the rules. See for example https://github.com/EFForg/https-everywhere/pull/10560#issuecomment-310397117. I am not sure what can be done about it, excepted merging the changes as soon as the tests finished running.

cschanaj:
Is it possible to trace which `commit` the checker is ran against? I am afraid there is more than one false-positive case...

Bisaloo:
I am confused about this:

> This is because currently there is no graceful way to remove comments from previous runs of the disabling fetch-test tool.

Did you run the tests against disabled rulesets? In this case, how come broken rulesets like [`BBC-mismatches`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/BBC-mismatches.xml) are not caught in the tests?

I see [a setting in the tests](https://github.com/EFForg/https-everywhere/blob/master/test/rules/src/https_everywhere_checker/check_rules.py#L285) to include/ignore `default_off` rulesets but I can't find anything that would explain why the tests only run against *some* `default_off` rulesets.

Bisaloo:
I don't understand the issue with [`P_Kimber.net.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/P_Kimber.net.xml).

https://pkimber.net works fine for me when accessed with Firefox, Chromium or curl.

Hainish:
@cschanaj the `full-fetch-test` was run against ee4304af7e9d48c939925c630bfbbf96b0a247bb

We are running the `full-fetch-test` across three cities through the course of a week.  It's necessary to stagger the tests, so we don't get false positives from temporary outages.  It's also necessary for all three scans to begin from the same commit, to ensure that when we run the intersect across these three cities, we get an accurate result.

For this PR, I rebased from `master` and this generated a number of merge conflicts.  For all these, I obviously adopted the files in `master` since these have been thoroughly checked.  But a merge conflict would have only resulted if the same portion of the file was modified in both `master` and `full-fetch-test`.  So it seems that some files had modifications in different parts of the file, removed from where `full-fetch-test` disables the rule and adds comments.

In the future, it would be optimal to, before merging, from `HEAD` on `master` do a `git diff --name-only <base commit>` (in this case ee4304af7e9d48c939925c630bfbbf96b0a247bb)

With this list, after applying the patch from the `full-fetch-test`, `git checkout HEAD <file>` for each of those files.

Hainish:
The above can also be done retroactively, so I'll look for those changes now.

Bisaloo:
I am not sure there is an easy way to go around that but some commented targets are still tested by `https-everywhere-checker`.

For example, `crm.leboncoin.fr` [`leboncoin.fr.xml`](in https://github.com/EFForg/https-everywhere/edit/master/src/chrome/content/rules/leboncoin.fr.xml).

Hainish:
>> This is because currently there is no graceful way to remove comments from previous runs of the disabling fetch-test tool.

> Did you run the tests against disabled rulesets? In this case, how come broken rulesets like BBC-mismatches are not caught in the tests?

https://github.com/EFForg/https-everywhere-full-fetch-test/blob/master/test.sh is the relevant file here.  The `full-fetch-test` removes the `default_off` attribute only for rulesets which have failed the ruleset test in the past.  If they have been disabled for other reasons, we leave them alone and don't bother checking them.

Hainish:
> I don't understand the issue with P_Kimber.net.xml.
> https://pkimber.net works fine for me when accessed with Firefox, Chromium or curl.

This is what I get:

user@browsing:~$ curl https://pkimber.net/
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104

Bisaloo:
```sh
user@host:~$ curl https://pkimber.net/
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.4.6 (Ubuntu)</center>
</body>
</html>
```
and 

```sh
user@host:~$ curl -L -s https://pkimber.net/ -w "%{http_code}" -o /dev/null
200
```

Hainish:
@Bisaloo I believe you.  I'm just saying this is probably due to a deeper issue - maybe the SSL library you're using.

Can you do a `curl --version` for me?

Bisaloo:
```sh
user@host:~$ curl --version
curl 7.47.0 (x86_64-pc-linux-gnu) libcurl/7.47.0 GnuTLS/3.4.10 zlib/1.2.8 libidn/1.32 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz TLS-SRP UnixSockets
```

Hainish:
I figured as much - I'm using `curl` based on `OpenSSL`:

user@browsing:~$ curl --version
curl 7.38.0 (x86_64-pc-linux-gnu) libcurl/7.38.0 OpenSSL/1.0.1t zlib/1.2.8 libidn/1.29 libssh2/1.4.3 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API SPNEGO NTLM NTLM_WB SSL libz TLS-SRP

Bisaloo:
In this simple case, https://pkimber.net/ redirects to https://www.pkimber.net/ which should be working for you so we can just rewrite both targets to https://www.pkimber.net/

Bisaloo:
I apologize for pestering you but your PR dug out a lot of strange cases.

Could you test [`leboncoin.fr.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/leboncoin.fr.xml)? I can't reproduce the time out found by `https-everywhere-checker` and it passes travis' tests just fine.

Hainish:
> In this simple case, https://pkimber.net/ redirects to https://www.pkimber.net/ which should be working for you so we can just rewrite both targets to https://www.pkimber.net/

`curl https://www.pkimber.net/` works for me.

We've flagged this kind of rewriting an anti-pattern, due to the ability for it to cause problems if a site owner changes their site configuration down the line.  In this very simple case, I think it's okay to make an exception.  But we shouldn't make a habit out of it.

Hainish:
@Bisaloo the rules that have been disabled were failing the fetch test across all three locations, *but* the comments are derived from the San Francisco test.  Looking now, I can see that `https://leboncoin.fr/` was not problematic in the London test, but several other domains there were.

When I just re-enabled and tested `https://github.com/EFForg/https-everywhere/blob/83f6bb5233dd5cc49175cbd44b0fccf973242a42/src/chrome/content/rules/leboncoin.fr.xml`, here was my result:

```shell
user@https-everywhere ~/workspace/https-everywhere (master) $ sudo docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="src/chrome/content/rules/leboncoin.fr.xml" --privileged httpse bash -c "service miredo start && test/fetch.sh"
 * Starting Teredo IPv6 tunneling daemon miredo                                                                                                                                                             [ OK ] 
INFO Finished comparing http://leboncoin.fr/ -> https://leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://beta.leboncoin.fr/ -> https://beta.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://compteperso.leboncoin.fr/ -> https://compteperso.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://comptepro.leboncoin.fr/ -> https://comptepro.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://corporate.leboncoin.fr/ -> https://corporate.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://deliv.leboncoin.fr/ -> https://deliv.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://ftp.leboncoin.fr/ -> https://ftp.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img0.leboncoin.fr/ -> https://img0.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img1.leboncoin.fr/ -> https://img1.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img2.leboncoin.fr/ -> https://img2.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img3.leboncoin.fr/ -> https://img3.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img4.leboncoin.fr/ -> https://img4.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img5.leboncoin.fr/ -> https://img5.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img6.leboncoin.fr/ -> https://img6.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://img7.leboncoin.fr/ -> https://img7.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://lesincroyablesrencontres.leboncoin.fr/ -> https://lesincroyablesrencontres.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://mobile.leboncoin.fr/ -> https://mobile.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://ns1.leboncoin.fr/ -> https://ns1.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://rss.leboncoin.fr/ -> https://rss.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://sondage.leboncoin.fr/ -> https://sondage.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://static.leboncoin.fr/ -> https://static.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://support.leboncoin.fr/ -> https://support.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://www.leboncoin.fr/ -> https://www.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://www2.leboncoin.fr/ -> https://www2.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
INFO Finished comparing http://crm.leboncoin.fr/ -> https://crm.leboncoin.fr/. Rulefile: src/chrome/content/rules/leboncoin.fr.xml.
ERROR src/chrome/content/rules/leboncoin.fr.xml: Fetch error: http://deliv.leboncoin.fr/ => https://deliv.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
ERROR src/chrome/content/rules/leboncoin.fr.xml: Fetch error: http://ns1.leboncoin.fr/ => https://ns1.leboncoin.fr/: (28, 'Connection timed out after 20001 milliseconds')
ERROR src/chrome/content/rules/leboncoin.fr.xml: Fetch error: http://rss.leboncoin.fr/ => https://rss.leboncoin.fr/: (28, 'Operation timed out after 30001 milliseconds with 0 bytes received')
ERROR src/chrome/content/rules/leboncoin.fr.xml: Fetch error: http://support.leboncoin.fr/ => https://support.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
ERROR src/chrome/content/rules/leboncoin.fr.xml: Fetch error: http://crm.leboncoin.fr/ => https://crm.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
INFO Finished in 150.74 seconds. Loaded rulesets: 1, URL pairs: 25.
Test URL test failed.
```

Hainish:
In London:

```html
<!--
Disabled by https-everywhere-checker because:
Fetch error: http://deliv.leboncoin.fr/ => https://deliv.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
Fetch error: http://ns1.leboncoin.fr/ => https://ns1.leboncoin.fr/: (28, 'Connection timed out after 20000 milliseconds')
Fetch error: http://rss.leboncoin.fr/ => https://rss.leboncoin.fr/: (28, 'Operation timed out after 30001 milliseconds with 0 bytes received')
Fetch error: http://support.leboncoin.fr/ => https://support.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
Fetch error: http://crm.leboncoin.fr/ => https://crm.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')

-->
```

In Singapore:

```html
<!--
Disabled by https-everywhere-checker because:
Fetch error: http://deliv.leboncoin.fr/ => https://deliv.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
Fetch error: http://ns1.leboncoin.fr/ => https://ns1.leboncoin.fr/: (28, 'Connection timed out after 20003 milliseconds')
Fetch error: http://rss.leboncoin.fr/ => https://rss.leboncoin.fr/: (28, 'Operation timed out after 30001 milliseconds with 0 bytes received')
Fetch error: http://support.leboncoin.fr/ => https://support.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')
Fetch error: http://crm.leboncoin.fr/ => https://crm.leboncoin.fr/: (60, 'SSL certificate problem: unable to get local issuer certificate')

-->
```

PR #10544, opened 2017-06-21
Update Alipay.com.xml

cschanaj:


J0WI:
Note that this rule is still `default_off`

cschanaj:
Closing this is much more involving than I have imagined and I simply do not have the time. See #10545 for more similar ruleset.

Issue #10545, opened 2017-06-21
Re-activate top-100 from #10538


cschanaj:
#10536 #10538 I cannot do this all myself. Please mark this as a Good Volunteer Work. Thanks!

### top-100
 - [x] ./https-everywhere/rules/Adobe.xml #10597 
 - [x] ./https-everywhere/rules/Alipay.com.xml #10547 
 - [x] ./https-everywhere/rules/Apple.com.xml #10600 
 - [x] ./https-everywhere/rules/BBC.xml #10623 
 - [x] ./https-everywhere/rules/Facebook.xml #10567 
 - [x] ./https-everywhere/rules/Instagram.xml #10569 
 - [x] ./https-everywhere/rules/LinkedIn.xml #10624 
 - [x] ./https-everywhere/rules/Live.xml 10625
 - [x] ./https-everywhere/rules/Mail.ru.xml #10626 
 - [x] ./https-everywhere/rules/PayPal.xml #10566 
 - [x] ./https-everywhere/rules/Pinterest.xml #10548 
 - [x] ./https-everywhere/rules/PopAds.xml #10570 
 - [x] ./https-everywhere/rules/ThePirateBay.xml #10549 
 - [x] ./https-everywhere/rules/Tumblr.xml #10550 
 - [x] ./https-everywhere/rules/Twitch.tv.xml #10568 
 - [x] ./https-everywhere/rules/Yahoo.xml #10627 
 - [x] ./https-everywhere/rules/Yahoo_Japan.xml #9955 
 - [x] ./https-everywhere/rules/YouTube.xml #10564 
 - [x] ./https-everywhere/rules/amazon.co.jp.xml #10546 
 - [x] ./https-everywhere/rules/hao123.com-mixedcontent.xml #10629

Bisaloo:
`facebook.com` is preloaded. Proofs: 

- https://hstspreload.org/?domain=facebook.com
- https://securityheaders.io/?q=facebook.com&followRedirects=on

Bisaloo:
[`paypal.com` is preloaded](https://hstspreload.org/?domain=paypal.com) as well but is [missing the `preload` header.](https://hstspreload.org/?domain=paypal.com)

Hainish:
Note: the relevant HSTS files to be checked for inclusion are those in the browsers we support:

[...]

PR #10549, opened 2017-06-21
Re-activate and Update ThePirateBay related rulesets

cschanaj:
#10545

### DIFF
--- a/src/chrome/content/rules/ThePirateBay.xml
+++ b/src/chrome/content/rules/ThePirateBay.xml
@@ -1,50 +1,36 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://kanyebay.com/ => https://kanyebay.com/: (28, 'Connection timed out after 20000 milliseconds')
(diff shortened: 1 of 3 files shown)


src/chrome/content/rules/ThePirateBay.xml
```diff
@@ -1,50 +1,34 @@
...
+	<target host="thepiratebay.org" />
+	<target host="www.thepiratebay.org" />
 	<target host="tpb.run" />
 	<target host="www.tpb.run" />
-	
-	<rule from="^http://www\.proxybay\.la/"
-		to="https://proxybay.la/" />
-
-	<target host="thepiratebay-proxylist.org" />
-	<target host="www.thepiratebay-proxylist.org" />
```
J0WI: Those targets have been removed, but they aren't covered by `thepiratebay-proxylist.org.xml` (which is disabled anyway).

cschanaj:
@J0WI I have re-activated more related ruleset.  Thanks!

PR #10561, opened 2017-06-21
[Deimos.fr] Update and reactivate ruleset

Bisaloo:


src/chrome/content/rules/Deimos.fr.xml
```diff
@@ -1,38 +1,17 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://piwik.deimos.fr/ => https://piwik.deimos.fr/: (6, 'Could not resolve host: piwik.deimos.fr')
-Fetch error: http://wiki.deimos.fr/ => https://wiki.deimos.fr/: (28, 'Operation timed out after 0 milliseconds with 0 out of 0 bytes received')
-
 	Mixed content:
+		wiki.deimos.fr
```
J0WI: But the target is active...

PR #10564, opened 2017-06-21
[Youtube] Reactivate ruleset

Bisaloo:
I am assuming the time out is temporary and the subdomain is eventually going to be fixed.

### DIFF
--- a/src/chrome/content/rules/YouTube.xml
+++ b/src/chrome/content/rules/YouTube.xml
@@ -12,6 +12,10 @@
         <!-- Main domains -->
 	<target host="youtube.com" />
 	<target host="*.youtube.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/YouTube.xml
```diff
@@ -1,21 +1,42 @@
...
 		- .youtube.com
 
 -->
-<ruleset name="YouTube" default_off='failed ruleset test'>
+<ruleset name="YouTube">
 
         <!-- Main domains -->
 	<target host="youtube.com" />
 	<target host="*.youtube.com" />
-	<target host="*.ytimg.com" />
```
J0WI: I would keep the wildcard here. They could add/remove more mirrors at any time.

PR #10565, opened 2017-06-21
LinuxFoundation Pardot bugfixes and updates

emsearcy:
An updated copy of #6035 since the thread got too convoluted.

Commit 7e99e2bf8 includes several fixes to Linux Foundation domains, the most important of which is a broken "vanity URL" tracker domain for Pardot.

Due to a concern about the negative lookahead syntax, the next commit fb64b5f00 creates a new  standalone Pardot domain xml, similar to bitly, which allows us to avoid the negative lookahead.

29970 contains style updates I learned about in the course of contributing this change.


emsearcy:
Closed in favor of #10537

PR #10567, opened 2017-06-22
Re-activate and Update Facebook.xml

cschanaj:


J0WI:
I think this is also a candidate for a wildcard, but let's just fix the tests in this PR.

PR #10583, opened 2017-06-22
Delete Muttscomics.com.xml

Bisaloo:
Redirects to another domain so not likely to get fixed.


J0WI:
I'm not sure if this is a valid argument.

Bisaloo:
Okay, let me explain my point of view:

This website seem to have moved to a new domain (namely `mutts.com`). The old domain probably only exists for legacy purposes and redirects to the new one. This means that it `https` doesn't work, it is not very likely they get out of their way to fix it on a domain they are going to retire at some point anyways.

We have many rulesets in this situation and I think they can be removed.

But if you don't feel comfortable with this, feel free to close this PR.

J0WI:
Well since Let's Encrypt is automated and free, there is no overhead to secure any (obsolete) alias domain.

How about creating a ruleset for the new domain and add the old unsecured one to the list of non-functional hosts? Then the admin may get aware of it and we can still remove it in future.

jeremyn:
I just merged some deletion PRs like this one (https://github.com/EFForg/https-everywhere/pull/10574, https://github.com/EFForg/https-everywhere/pull/10582) and didn't realize there was a discussion going on about it.

I see no problem with deleting these. If someone wants to notify the webmaster, then send them an email instead of putting a note in one of our rulesets. They probably already know their own site is failing HTTPS anyway.

If anyone cares about coverage for the redirect target domain, then they can make a new issue or note the domain in https://github.com/EFForg/https-everywhere/issues/3069.

I disagree with adding a comment to ruleset A that some other random domain B redirects to it. By this logic anyone could create a domain that redirects to microsoft.com, google.com etc and get their domain included in these ruleset files. And even if the redirecting site could be proven to be associated with the redirect target (e.g. Google definitely owns the site redirecting to google.com), this is the sort of comment that will never go away and can puzzle or frustrate future maintainers.

PR #10589, opened 2017-06-22
Delete Shoutback-Concepts.xml

Bisaloo:


J0WI:
The main domain is still in use, but does not HTTPS at all.

PR #10592, opened 2017-06-22
Delete Auto_VPS.net.xml

Bisaloo:
Gone from DNS.


cschanaj:
@Bisaloo FYI, I have got a list of rulesets which I suspected are gone from DNS, see https://gist.github.com/cschanaj/c6ce8302a347f85bae7990fbebd8b47c

Bisaloo:
Okay, what are you planning to do with them? Mass delete PR or do you need help for manual review?

cschanaj:
I cannot confirm that all of them are really gone (but very likely). It will be great if you can help to check some of them if you have the time.

Bisaloo:
Could you add the list to an actual git repo or something so that I can submit PR to tell you which domains I have checked? Or you can open an issue about it, as you did in the past.

cschanaj:
#10595

PR #10596, opened 2017-06-23
Delete Acrocomcontent.com.xml

cschanaj:
#10595

Issue #10595, opened 2017-06-23
Delete rulesets which are all gone from DNS


cschanaj:
I suspected a number of rulesets having all `target` gone from DNS: https://github.com/EFForg/https-everywhere/pull/10516#issuecomment-311502592

Please refer to [`gone-from-dns-list.txt`](https://github.com/cschanaj/https-everywhere-issues/blob/master/gone-from-dns-list.txt)

Bisaloo:
`Caisse-Epargne.fr.xml` is not gone from DNS.

koops76:
Maybe we should wait for some time to let them renew their domains before deleting the rulesets?

cschanaj:
@koops76 Thank you for helping in resolving this issue. Please note that not all the domain listed in the above are completely gone from DNS. Occasionally, you might want to update a ruleset instead of deletion.

Bisaloo:
@cschanaj, any idea why your script didn't pick up  #10738, #10732, #10729?

cschanaj:
For #10738, #10732 The items from the list only include ruleset which

 - `Disabled by https-everywhere-checker` once only
 - Number of occurrence of `Could not resolve host:` == number of non-wildcard `target`

As for #10729, I guess it is because of the improper indentation (with space)

cschanaj:
Let's wait for the discussion in https://github.com/EFForg/https-everywhere/pull/10516#issuecomment-311387749 complete before submitting new deletion PRs.

jeremyn:
@cschanaj I merged your recent deletion PRs but from https://github.com/EFForg/https-everywhere/pull/10516#issuecomment-317886379 it sounds like `full-fetch-test` should get rid of these rulesets automatically, so you don't need to submit PRs to delete them. What do you think?

PR #10597, opened 2017-06-23
Re-activate and Update Adobe.xml, Delete Acrocomcontent.com.xml

cschanaj:
#10545, #10595, close #10596


src/chrome/content/rules/Adobe.xml
```diff
@@ -199,7 +189,6 @@ Fetch error: http://max.adobe.com/ => https://max.adobe.com/: Too many redirects
 <ruleset name="Adobe.com" default_off='failed ruleset test'>
```
J0WI: The rule is still off.

src/chrome/content/rules/Adobe.xml
```diff
@@ -50,6 +40,7 @@ Fetch error: http://max.adobe.com/ => https://max.adobe.com/: Too many redirects
 
 		- c00 ²
 		- community **
+		- discoverwem (timeout)
```
J0WI: The target is still active.

PR #10600, opened 2017-06-23
Re-activate and Update Apple.com.xml

cschanaj:
#10545

### DIFF
--- a/src/chrome/content/rules/Apple.com.xml
+++ b/src/chrome/content/rules/Apple.com.xml
@@ -1,10 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Apple.com.xml
```diff
@@ -150,7 +148,7 @@ Fetch error: http://uptodate.apple.com/ => https://uptodate.apple.com/: (28, 'Op
 	<target host="feeds.itunes.apple.com" />
 	<target host="files.itunes.apple.com" />
 	<target host="genius.itunes.apple.com" />
-	<target host="genius-2.itunes.apple.com" />
+	<!-- <target host="genius-2.itunes.apple.com" /> -->
```
J0WI: Please add a notice why this has been disabled.

PR #10627, opened 2017-06-23
Reactivate Yahoo.xml

koops76:
#10545


cschanaj:
Please remove the complicated rules for `totaltravel.com` and `totaltravel.co.uk` as well. Besides, I think that you are whitelisting the wrong file.

PR #10629, opened 2017-06-23
Reactivate hao123.com-mixedcontent.xml

koops76:
#10545


cschanaj:
See also [`Halo123.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Halo123.xml)

PR #10642, opened 2017-06-23
Delete CyanogenMod.org-falsemixed.xml

koops76:
#10595


cschanaj:
see also [`CyanogenMod.org.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/CyanogenMod.org.xml)

koops76:
@J0WI #11420

PR #10644, opened 2017-06-23
Delete Hotfile.xml

koops76:
#10595


J0WI:
The host has a valid DNS record, but also a new owner (feel free to add a rule for `wheretowatch.com`). So I assume it's okay to delete it.

PR #10648, opened 2017-06-23
Delete Intencysrv.com.xml

Bisaloo:
All subdomains are gone from DNS or time out over both http and https.

#10595


J0WI:
Some WFM, but no https support.

PR #10659, opened 2017-06-24
Remove old exceptions from Phaser.io.xml.

### LINKED ISSUE #6907: Partial rule for phaser.io breaks some content
If you go to http://phaser.io/tutorials/making-your-first-phaser-game/part3, then the CSS and a lot of the images are broken because they're changed to HTTPS but the server has a certificate error, like this one: http://phaser.io/content/tutorials/making-your-first-phaser-game/part4.png

bardiharborow:
After manually forcing my browser to treat Phaser.io as a HSTS preloaded site, and testing all the previously affected URLs, and just wandering around the site generally, all the issues seem to be resolved. Closes #6907 (an old ticket that presumably caused the exceptions to be created, but hasn't been closed).

### DIFF
--- a/src/chrome/content/rules/Phaser.io.xml
+++ b/src/chrome/content/rules/Phaser.io.xml
@@ -1,33 +1,6 @@
-<ruleset name="Phaser.io (partial)">
-
-	<!--	Direct rewrites:
-				-->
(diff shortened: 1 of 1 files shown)


J0WI:
Thanks!

PR #10662, opened 2017-06-24
Delete On-web.fr.xml

Bisaloo:


J0WI:
Needs rebasing after #11239

Bisaloo:
I was previously removing planet-work.com because it is preloaded but I am leaving this job to `HSTS-prune`. So this PR simply removes `On-web.fr.xml` now.

J0WI:
But some hosts in `*.on-web.fr` support HTTPS.

Bisaloo:
You are right, I must have overlooked that.

PR #10666, opened 2017-06-24
[Freenode] Update and reactivate ruleset

Bisaloo:


Bisaloo:
@J0WI, I forgot to remove the `default_off`. See #11150.

PR #10676, opened 2017-06-25
Delete Browser_Hacker.com.xml

Bisaloo:
Gone from DNS.

#10595


J0WI:
Can't confirm, WFM over https.

Bisaloo:
@J0WI, strange. It's still gone from DNS for me. Are you talking about `^`, `www` or another subdomain?

J0WI:
@Bisaloo  `$ curl -I https://www.browserhacker.com`

Bisaloo:
Okay, it works with `curl` but no with firefox nor chromium. What does that mean?

This is just out of curiosity. I think it is a sufficient reason to close this PR.

PR #10679, opened 2017-06-25
Update and reactivate ThePirateBay.xml

Bisaloo:
I would like to use this PR to discuss the case of thepiratebay. Currently, the numerous mirrors are secured across 9 rulesets, without any apparent logic behind it.

First of all, we should stop redirecting dead mirrors to working ones, as in [`thepiratebay-legacy.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/thepiratebay-legacy.xml). It is not this extension job to track the current working mirror. It is very time consuming and it could introduce a security risk (some mirrors may be compromised and we don't want to be responsible for sending users on them). This point was already discussed in https://github.com/EFForg/https-everywhere/pull/8561.

Also, in this case, I am not sure if it would better to group all domain in a single ruleset, that would require frequent but somewhat easy updates or split them in a myriad of smaller rulesets, which would have to be updated less frequently.

### DIFF
--- a/src/chrome/content/rules/ThePirateBay.xml
+++ b/src/chrome/content/rules/ThePirateBay.xml
@@ -1,44 +1,32 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://kanyebay.com/ => https://kanyebay.com/: (28, 'Connection timed out after 20000 milliseconds')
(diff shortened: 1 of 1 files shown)


Bisaloo:
Duplicate of #10549.

PR #10684, opened 2017-06-25
[OpenClipart.org] Fix mismatch and reactivate ruleset

Bisaloo:


src/chrome/content/rules/OpenClipart.org.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="Open Clipart.org">
+
+	<target host="openclipart.org" />
+	<target host="www.openclipart.org" />
+
+	<securecookie host="^\.openclipart\.org$" name=".+" />
+
+	<rule from="^http://(www\.)?openclipart\.org/"
+		to="https://openclipart.org/" />
```
J0WI: Please document the mismatch.

PR #10687, opened 2017-06-25
[amazon.in.xml] fix #10515

### LINKED ISSUE #10515: Amazon India Domain Not Compatible
Hello everyone!

I have been testing Amazon's Indian domain, amazon.in, with https-everywhere add-on and found out that it actually breaks the site's functionality!

**Here's how**:
The homepage works fine and hence the rule seems fine too! But when you dig deep and navigate to say a section say, "Today's Deal" and there you will see many items disappear with an infinite loading circle! Almost 90 percent of the whole page become inaccessible!

**Note**;
Amazon US seems to have been handing Indian version a bit differently. Since amazon.com suffers no such issues as mentioned above.

Thank you for your time and I welcome your support.

Dev

J0WI:


J0WI:
@jeremyn this is a follow-up of #7419 to fix #10515

jeremyn:
I can confirm that the changes here fix the problem described in #10515, but

- [ ] How can I reproduce the `HTTP 405` you describe?

J0WI:
`$ curl -I http://www.amazon.in/cross-app-features/ags-best-deal-widget/handler/get-best-deals.html`

jeremyn:
I can confirm the `405` from curl, but I get a `200` from Firefox when browsing to the URL directly which also uses `GET`. So there's more to the `405` than just a `GET`. 

If you want to just say the `405` only happens in curl but not browsers, that's fine with me.

J0WI:
done

jeremyn:
Thanks, merged.

PR #10692, opened 2017-06-25
Delete ATech_Media.com-falsemixed.xml

koops76:
#10595


cschanaj:
see also [`ATech_Media.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ATech_Media.xml)

PR #10703, opened 2017-06-25
Delete Bunadformenn.info.xml

koops76:
#10595


src/chrome/content/rules/Bunadformenn.info.xml
```diff
@@ -1,27 +0,0 @@
...
-			- cdn.bunadformenn.info
-
-
-	Nonfunctional subdomains:
-
-		- ^	(502)
-		- www	(times out)
-
--->
-<ruleset name="bunadformenn.info (partial)" default_off='failed ruleset test'>
```
J0WI: This domain is still up

PR #10727, opened 2017-06-26
Delete Codespeak.xml

koops76:
#10595


J0WI:
This host is still up.

koops76:
@J0WI It has a StartCom certificate and uses SHA-1. It doesn't open in latest Chrome. I guess it should be deleted, since it breaks the website by redirecting it to broken HTTPS.

PR #10733, opened 2017-06-26
Update CAcert.xml

koops76:
#10227

### DIFF
--- a/src/chrome/content/rules/CAcert.xml
+++ b/src/chrome/content/rules/CAcert.xml
@@ -1,43 +1,17 @@
-
-<!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


cschanaj:
`platform="cacert" default_off="cacert"` is not equivalent to `platform="cacert"` AFAIK.

Hainish:
@cschanaj we need to make a policy decision for CAcert sites.  These should be `default_off`, since we don't want these sites to be inaccessible by browsers in default configuration.  We also want them to pass the fetch test, so that they aren't disabled on the next run of https://github.com/EFForg/https-everywhere-full-fetch-test.  Adding `default_off="cacert"` seems the easiest way to do this.  Keeping `cacert` as the value of the `platform` attribute makes sense to me as well, since the `platform` attribute is often used as a selector for specific system configurations to turn rulesets on.

See: https://github.com/EFForg/https-everywhere/issues/10227#issuecomment-311159236

PR #10739, opened 2017-06-26
Added KissAnime.ru ruleset

Greydus:
Added a ruleset for [KissAnime](http://kissanime.ru)


jeremyn:
@Greydus Thanks for your pull request, unfortunately it looks like https://kissanime.ru redirects to http://kissanime.ru. This means we shouldn't make a ruleset for it, because that would cause a redirect loop. (The similar [`KissAnime.to.xml`](https://github.com/EFForg/https-everywhere/blob/7e4193f45aeb4633e79cdfa56abe1178bf746b39/src/chrome/content/rules/KissAnime.to.xml#L22) is disabled by default for this reason.) So, I'm closing this pull request.

PR #10757, opened 2017-06-27
[RYLWarfare.net] Update and simplify ruleset

Bisaloo:


J0WI:
WFM

Bisaloo:
That's so weird. It had no DNS records for the two full-fetch-tests. No DNS records when I tested it. But now, it works again?

PR #10768, opened 2017-06-27
Delete Svetlodarsk.com.xml

koops76:
#10595


gloomy-ghost:
svetlodarsk.com can be resolved, still not available in https though.

koops76:
@gloomy-ghost I guess it can be deleted anyway if it doesn't support HTTPS.

Issue #10782, opened 2017-06-28
Re-activate top-1k from #10538


cschanaj:
#10538 Noted that this is not as urgent as #10545 .

### top-1k 

- [x] ~./https-everywhere/rules/Automatic_Data_Processing.xml~
- [x] ./https-everywhere/rules/Lowes.com.xml #10893 
- [x] ./https-everywhere/rules/ASUS.xml #10789 
- [x] ./https-everywhere/rules/LinkedIn.xml #10624 
- [x] ./https-everywhere/rules/Live.xml #10625 
- [x] ./https-everywhere/rules/Mail.ru.xml #10626 
- [x] ./https-everywhere/rules/104.com.tw.xml #10787 
- [x] ./https-everywhere/rules/1688.com.xml #10854 
- [x] ./https-everywhere/rules/4Shared.xml #10788 
- [x] ./https-everywhere/rules/9gag.com.xml #10540 
- [x] ./https-everywhere/rules/Alibaba.xml #10651 
- [x] ./https-everywhere/rules/Apache.xml #10778 
- [x] ./https-everywhere/rules/Asana.xml #10650 
- [x] ./https-everywhere/rules/Blastingnews.com.xml #10798 
- [x] ./https-everywhere/rules/Booking.com.xml #10799 
- [x] ./https-everywhere/rules/Box.xml #10806 
- [x] ./https-everywhere/rules/BuzzFeed.xml #10818 
- [x] ./https-everywhere/rules/CapitalOne.com.xml #10855 
- [x] ./https-everywhere/rules/Cisco.xml #10856 
- [x] ./https-everywhere/rules/CodePen.io.xml #10815 
- [x] ./https-everywhere/rules/Doublepimp.xml #10780 
- [x] ./https-everywhere/rules/Dell.xml #10829 
- [x] ./https-everywhere/rules/Delta.com.xml #9877 
- [x] ./https-everywhere/rules/Douban.com.xml #10827 
- [x] ./https-everywhere/rules/DuckDuckGo.xml #10586 
- [x] ~./https-everywhere/rules/Electronic-Arts.xml~
- [x] ./https-everywhere/rules/Evernote.xml #10781 
- [x] ./https-everywhere/rules/Exoclick.xml #10824 
- [x] ./https-everywhere/rules/Expedia.xml #10831 
- [x] ./https-everywhere/rules/Flickr.xml #10610 
- [x] ./https-everywhere/rules/Focus.de.xml #10832 
[...]

PR #10788, opened 2017-06-28
Update 4Shared.xml

koops76:
#10782

### DIFF
--- a/src/chrome/content/rules/4Shared.xml
+++ b/src/chrome/content/rules/4Shared.xml
@@ -32,10 +32,12 @@
 
 	<target host="4shared.com" />
 	<target host="*.4shared.com" />
+	<exclusion pattern="^http://help\.4shared\.com/" />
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/4Shared.xml
```diff
@@ -30,15 +28,14 @@ Fetch error: http://help.4shared.com/ => https://help.4shared.com/: (7, 'Failed
 	ˢ Secured by us, see https://www.paulirish.com/2010/the-protocol-relative-url/
 
 -->
-<ruleset name="4Shared.com" default_off='failed ruleset test'>
+<ruleset name="4Shared.com">
 
 	<target host="4shared.com" />
 	<target host="*.4shared.com" />
```
J0WI: The wildcard will break `help.4shared.com`.

koops76: @J0WI It should be removed but I added an exclusion for now.

PR #10789, opened 2017-06-28
Reenable ASUS.xml

koops76:
#10782


koops76:
https://asus.com and https://zenbook.asus.com are indeed inaccessible.

PR #10790, opened 2017-06-28
Update League_of_Legends.com.xml

koops76:
#10782

### DIFF
--- a/src/chrome/content/rules/League_of_Legends.com.xml
+++ b/src/chrome/content/rules/League_of_Legends.com.xml
@@ -1,8 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/League_of_Legends.com.xml
```diff
@@ -94,11 +92,10 @@ Fetch error: http://dominion.leagueoflegends.com/ => https://dominion.leagueofle
 	* Secured by us
 
 -->
-<ruleset name="League of Legends.com (partial)" default_off='failed ruleset test'>
+<ruleset name="League of Legends.com (partial)">
 
 	<target host="leagueoflegends.com" />
 	<target host="signup.leagueoflegends.com" />
-	<target host="support.leagueoflegends.com" />
```
J0WI: The issue is, that there are too many redirects. If you add a rule to redirect this host to `https://support.riotgames.com/` it should work.

koops76: @J0WI Done.

PR #10793, opened 2017-06-28
Update Intel.xml

koops76:
#10782

### DIFF
--- a/src/chrome/content/rules/Intel.xml
+++ b/src/chrome/content/rules/Intel.xml
@@ -1,9 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Intel.xml
```diff
@@ -58,12 +54,14 @@ Fetch error: http://scoop.intel.com/ => https://scoop.intel.com/: (7, 'Failed to
 	Problematic subdomains:
 		- chi1      (Expired)
 		- embedded.edc ³
+		- firmware ⁴
```
J0WI: It's not a mismatch, it is a issue in the certificate chain. And the target is still active.

koops76: @J0WI Fixed.

PR #10807, opened 2017-06-29
Re-activate and Update Attendesource.com.xml

cschanaj:
#9906 there is only one sub-domain with working DNS record.

### DIFF
--- a/src/chrome/content/rules/Attendesource.com.xml
+++ b/src/chrome/content/rules/Attendesource.com.xml
@@ -1,0 +1,5 @@
+<ruleset name="Attendesource.com">
+	<target host="attendesource.com" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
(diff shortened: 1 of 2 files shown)


J0WI:
How can we verify if the bug described in the `default_off` reason has been fixed?

cschanaj:
@J0WI `^` serves HSTS, so I guess the clients cannot access the plaintext page anyway.

```bash
$ curl -I https://attendesource.com
Date: Tue, 25 Jul 2017 09:01:33 GMT
Server: Apache
Strict-Transport-Security: max-age=7776000; includeSubDomains
x-xss-protection: 0
Set-Cookie: applicationName=attendesource;Domain=.attendesource.com;Path=/;Expires=Wed, 26 Jul 2017 09:01:33 GMT;Secure;HttpOnly
location: https://attendesource.com/user/index.cfm
Cache-Control: no-cache
Pragma: no-cache
Content-Type: text/html;charset=UTF-8

```

J0WI:
Well, if they never visited the host over https, they still could use http. But it's a strong argument.

PR #10808, opened 2017-06-29
Replace Greek-various.xml with split files per domain

kargig:
As requested at https://github.com/kargig/https-everywhere-greek-rules#5
Update these domain xmls.


J0WI:
See https://github.com/EFForg/https-everywhere/pull/12042#issuecomment-343746834

kargig:
Looks like me and @koops76 are trying to do the same thing in different PRs. I don't know why @J0WI didn't make his comment about split PRs here in the first place instead on #12042, but anyway it seems to me that this PR is of no use any more.

I'm closing it. 
Thanks @koops76 for creating all those PRs for the Greek websites

PR #10809, opened 2017-06-29
[HollywoodOutlaws.com] Update default_off and merge rulesets

Bisaloo:


J0WI:
Since this host does not support any TLS version I'm okay with removing it.

PR #10813, opened 2017-06-30
Use for-of and Array.filter instead of general `for`

yfdyh000:
A by-product for understanding the code of the project. This is a readability cleanup, although there may be no significant advantage and may need to QA test (I have do it).

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -90,8 +90,8 @@ var wr = chrome.webRequest;
 var loadStoredUserRules = function() {
   var rules = getStoredUserRules();
   var i;
-  for (i = 0; i < rules.length; ++i) {
-    all_rules.addUserRule(rules[i]);
+  for (let rule of rules) {
+    all_rules.addUserRule(rule);
   }
   log('INFO', 'loaded ' + i + ' stored user rules');
 };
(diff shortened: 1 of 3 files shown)


bershan2:
Disclaimer: I do not know HTTPS Everywhere code (and never contributed to this repository), everything I write is supported only by the sources I cite. Think critically of this comment!

@yfdyh000 Did you check that `for ... of` is supported by all browser versions that extension?

For example, [HTTPS Everywhere aims to support all Chromium versions after 45](https://github.com/EFForg/https-everywhere/blob/3d21c7109723060e6260cc81f357ae9c67992458/chromium/manifest.json), while [iteration over objects came only in version 51](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Chrome_note_3). But I think rules are [represented as objects](https://github.com/EFForg/https-everywhere/blob/203375992d37e878d585e49d050fea283f51e390/src/chrome/content/code/HTTPSRules.js):
```
function Rule(from, to) {
      this.to = to;
      this.from_c = from; // This will become a RegExp after compilation
}
```
May be, the regular `for` is necessary for backward-compatibility... But I can't tell: I do not know HTTPS Everywhere code.

yfdyh000:
Indeed, I forgot it, though once consult the document.

koops76:
@yfdyh000 We already had `for...of` loops in the codebase so I guess this is OK.

yfdyh000:
@koops76 Thank you for the reminder that I did not notice it before.

So, the `"minimum_chrome_version": "45"` should be updated? I have not carefully tested the current codebase with older Chrome versions.

koops76:
@yfdyh000 I already did.
https://github.com/EFForg/https-everywhere/pull/10852/files#diff-a951caac7ef604fdff31e05ef8ca45d4R31

koops76:
@yfdyh000 #10864

J0WI:
cc @Hainish

Hainish:
lgtm

PR #10820, opened 2017-06-30
Re-activate and Update South_China_Morning_Post.xml

cschanaj:
#9906

### DIFF
--- a/src/chrome/content/rules/South_China_Morning_Post.xml
+++ b/src/chrome/content/rules/South_China_Morning_Post.xml
@@ -3,7 +3,7 @@
 		^scmp.com
 		www.scmp.com
 -->
-<ruleset name="South China Morning Post">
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/South_China_Morning_Post.xml
```diff
@@ -3,12 +3,9 @@
 		^scmp.com
 		www.scmp.com
 -->
-
-<ruleset name="SouthChinaMorningPost" default_off="incomplete certificate chain">
-
+<ruleset name="South China Morning Post">
```
J0WI: Can we add the `partial` flag? Most parts of the main site are not secured.

cschanaj: done! thanks!

PR #10823, opened 2017-07-01
Update Abuse.ch.xml

gopherit:
Ruleset clean-up, add and remove some subdomains.

### DIFF
--- a/src/chrome/content/rules/Abuse.ch.xml
+++ b/src/chrome/content/rules/Abuse.ch.xml
@@ -3,9 +3,11 @@
 	<target host="www.abuse.ch" />
 	<target host="feodotracker.abuse.ch" />
 	<target host="mail.abuse.ch" />
+	<target host="palevotracker.abuse.ch" />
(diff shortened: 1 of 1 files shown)


J0WI:
@gopherit please rebase your branch and solve the conflicts.

gopherit:
@J0WI done.

src/chrome/content/rules/Abuse.ch.xml
```diff
@@ -1,14 +1,16 @@
-<ruleset name="Abuse.ch">
-
-	<target host="spyeyetracker.abuse.ch" />
```
J0WI: This host does still support HTTPS. Let's keep it.

gopherit: Even if it's been discontinued?  Same story with palevotracker.abuse.ch

J0WI: As long as the hosts is accessible over HTTP we secure it. Otherwise you could still do nasty things with it.

PR #10827, opened 2017-07-01
Re-activate and Update Douban.com.xml

cschanaj:
#10782

### DIFF
--- a/src/chrome/content/rules/Douban.com.xml
+++ b/src/chrome/content/rules/Douban.com.xml
@@ -1,37 +1,12 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://thing.douban.com/ => https://thing.douban.com/: (6, 'Could not resolve host: thing.douban.com')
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Douban.com.xml
```diff
@@ -54,37 +30,15 @@ Fetch error: http://img6.douban.com/ => https://img6.doubanio.com/: (60, 'SSL ce
...
-	<target host="thing.douban.com" />
 	<target host="wap.douban.com" />
 	<target host="www.douban.com" />
 	<target host="ypy.douban.com" />
 
-	<!--	Complications:
-				-->
-	<target host="fm.douban.com" />
-	<target host="img3.douban.com" />
-	<target host="img6.douban.com" />
```
J0WI: Please try img[1-9], some are still up.

cschanaj:
@J0WI 

Only `img2.douban.com` has a DNS record, but it redirects to `img3` over HTTP (which is gone from DNS) while `curl` gives a `Server returned nothing (no headers, no data)` error over HTTPS. 

So, I guess that `img2` is already obsolete.

PR #10831, opened 2017-07-01
Update Expedia.xml

koops76:
#10782

### DIFF
--- a/src/chrome/content/rules/Expedia.xml
+++ b/src/chrome/content/rules/Expedia.xml
@@ -1,25 +1,24 @@
 <!--
-
-	Other Expedia rulesets:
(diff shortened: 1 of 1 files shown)


J0WI:
Looks like a false positive. The host WFM in browser, but requests by curl are dropped.

J0WI:
Can you retry with a static resource as test url? Otherwise we need to whitelist it.

koops76:
@J0WI I CBA to look for a URL. Sorry, look for it yourself.

cschanaj:
Be noted that this is still `default_off`

koops76:
@cschanaj Done.

Issue #9621, opened 2017-04-19
Release 5.2.16 forthcoming


Hainish:
cc @jsha @cooperq @pde @J0WI @fuglede @semenko @loveithateit @diracdeltas @jeremyn

Scheduled for 5/4

Bisaloo:
Can we HSTS-prune this time please? Is it simply an oversight or is there a technical reason it wasn't done last time?

It would be really helpful for us contributors. I sometimes start working on updating a ruleset before realising it's now preloaded. It's just a waste of time.

J0WI:
#8833 is resolved, no more downgrade rules 🎉

Hainish:
I'm unable to make a release today, and we don't release on Fridays, so this will have to wait until Monday.

Hainish:
release made!

Hainish:
I didn't get the chance to include the HSTS script into the release process.  I'll run this separately tomorrow

Bisaloo:
> I didn't get the chance to include the HSTS script into the release process. I'll run this separately tomorrow

@Hainish, I think you forgot. It would be great if you have the time to do this!

Hainish:
I did! I'll try to get to it today, thanks for the reminder.

On May 12, 2017 2:39:46 AM PDT, Bisaloo <notifications@github.com> wrote:
>> I didn't get the chance to include the HSTS script into the release
>process. I'll run this separately tomorrow
>
>@Hainish, I think you forgot. It would be great if you have the time to
>do this!
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/EFForg/https-everywhere/issues/9621#issuecomment-301032175

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Hainish:
@Bisaloo https://github.com/EFForg/https-everywhere/pull/9891 is the result of hsts-prune.

Issue #7253, opened 2016-10-01
Could we add rules to redirect one https to another?


ivysrono:
Examples:

For work correctly: #6894 

For work better: #6889

anewuser:
You can use this extension to write custom redirects: https://github.com/einaregilsson/Redirector#download-links

ivysrono:
@anewuser I am using it. However, it does not work in Android, does not for everyone.

jeremyn:
I don't like redirecting domains, for example see the discussions in #2709 and #6905 . I might consider it if both domains were related and this could be verified. For example, I might allow redirects between the various Wikimedia projects listed [here](https://wikimediafoundation.org/wiki/Special:SiteMatrix).

We've already discussed this though so I'll let others contribute.

anewuser:
@ivysrono Subscribe to this issue to know when Android support for Redirector is implemented: https://github.com/einaregilsson/Redirector/issues/29

PR #10839, opened 2017-07-01
Only modify HTTP requests

koops76:
I don't think we have any HTTPS to HTTPS requests and downgrades are disabled so let's only add a HTTP request handler.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -5,7 +5,7 @@
  * @param url: a relative URL to local XML
  */
 function loadExtensionFile(url, returnType) {
-  const xhr = new XMLHttpRequest();
+  var xhr = new XMLHttpRequest();
   // Use blocking XHR to ensure everything is loaded by the time
   // we return.
   xhr.open("GET", chrome.extension.getURL(url), false);
@@ -53,7 +53,7 @@ storage.get({httpNowhere: false}, function(item) {
 });
 chrome.storage.onChanged.addListener(function(changes, areaName) {
   if (areaName === 'sync' || areaName === 'local') {
-    for (const key in changes) {
+    for (var key in changes) {
       if (key === 'httpNowhere') {
         httpNowhereOn = changes[key].newValue;
         updateState();
@@ -75,21 +75,22 @@ chrome.webNavigation.onCompleted.addListener(function() {
 * Load stored user rules
  **/
 var getStoredUserRules = function() {
-  const oldUserRuleString = localStorage.getItem(USER_RULE_KEY);
-  let oldUserRules = [];
+  var oldUserRuleString = localStorage.getItem(USER_RULE_KEY);
+  var oldUserRules = [];
   if (oldUserRuleString) {
     oldUserRules = JSON.parse(oldUserRuleString);
   }
   return oldUserRules;
 };
-const wr = chrome.webRequest;
+var wr = chrome.webRequest;
 
 /**
  * Load all stored user rules
  */
 const loadStoredUserRules = function() {
-  const rules = getStoredUserRules();
-  for (const i = 0; i < rules.length; ++i) {
+  var rules = getStoredUserRules();
+  var i;
+  for (i = 0; i < rules.length; ++i) {
     all_rules.addUserRule(rules[i]);
   }
   log('INFO', 'loaded ' + i + ' stored user rules');
@@ -104,13 +105,13 @@ loadStoredUserRules();
  * active: extension is enabled and rewrote URLs on this page.
  * disabled: extension is disabled from the popup menu.
  */
-const updateState = function() {
+var updateState = function() {
   chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
     if (!tabs || tabs.length === 0) {
       return;
     }
-    const applied = activeRulesets.getRulesets(tabs[0].id)
-    let iconState = "inactive";
+    var applied = activeRulesets.getRulesets(tabs[0].id)
+    var iconState = "inactive";
     if (!isExtensionEnabled) {
       iconState = "disabled";
     } else if (httpNowhereOn) {
@@ -134,12 +135,12 @@ const updateState = function() {
  * @param params: params defining the rule
  * @param cb: Callback to call after success/fail
  * */
-const addNewRule = function(params, cb) {
+var addNewRule = function(params, cb) {
   if (all_rules.addUserRule(params)) {
     // If we successfully added the user rule, save it in local 
     // storage so it's automatically applied when the extension is 
     // reloaded.
-    const oldUserRules = getStoredUserRules();
+    var oldUserRules = getStoredUserRules();
     // TODO: there's a race condition here, if this code is ever executed from multiple 
     // client windows in different event loops.
     oldUserRules.push(params);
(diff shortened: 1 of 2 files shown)


Bisaloo:
Actually, there are rewrites `https` -> `https`. Now the question is: do we want to keep them? 

https://github.com/EFForg/https-everywhere/issues/1327

```sh
grep 'from="^https' *.xml | wc -l
```
> 72

koops76:
@Bisaloo Please send a list of files which have them. I may rewrite these rules as a part of this PR.

Bisaloo:
```
7Headlines.com.xml
AdultShopping.com.xml
Aeriagames.xml
Agol.dk.xml
AJC.com.xml
Austin_Fit_Magazine.com.xml
basekit.com.xml
Blip.xml
BME.hu_incomplete.xml
BME.hu.xml
CafePress.xml
Caltech.xml
cambridgeshire.gov.uk.xml
Cashback.co.uk.xml
Cato-Institute.xml
Comic-Con-Intl.xml
Cru.org.xml
DailyDot.xml
Dell.xml
Demonoid.xml
Deutsche-Rentenversicherung.de.xml
Epic-Systems.xml
Epson.xml
Fasthosts.xml
First_Class_Magazine.se.xml
Ghostery_metrics_optout.xml
GoogleMaps.xml
Google.xml
InkFrog.com.xml
Intelliworks_Chat.com.xml
Kettering.gov.uk.xml
Khalsa-Credit-Union.xml
Kiva.org.xml
KKH.de.xml
Knappschaft-Bahn-See.xml
Linuxaria.com.xml
LiveJournal.xml
Minijob-zentrale.de.xml
MoveOn.xml
Mutterschaftsgeld.de.xml
National_Geographic.xml
Necuhb.xml
Ntop.org.xml
Numergy.xml
Pair-Networks.xml
Parkerpen.com.xml
Perfect_Audience.xml
Plymouth-University.xml
Sandbag.xml
Sberbank_of_Russia.xml
SPajIT.xml
spectator.co.uk.xml
Stack-Exchange.xml
Stallman.org.xml
Telus.xml
Tesco.xml
Unblocked.xml
Unity.xml
US-Dept-of-Veterans-Affairs.xml
VEVO.xml
Virus_Bulletin.xml
Yandex.xml
Yimg.com.xml
yWorks.com.xml
ZeniMax-Media.xml
```

Note that there are sometimes good reasons for these rewrites. Hopefully, it will documented more often than not.

As explained in the issue I linked earlier, these rules sometimes prevent breakages (such as use of protocole relative links `//example.com`). It's more a matter of policy whether we want to keep them or not.

koops76:
Let's ask @Hainish and others' opinions.

jeremyn:
I don't like rewriting HTTPS domains because that goes beyond HTTPS Everywhere's mission. At the same time there may be some edge cases I'm not thinking of where rewriting HTTPS is a good idea.

Probably the first step is to update each of the relevant rulesets, in separate PRs, like in https://github.com/EFForg/https-everywhere/issues/9842#issue-227004120 for `default_off` rulesets. Then we can see what we're left with and what changes are appropriate or possible. Note that I'm not committing to reviewing these PRs myself, so it may be a while before someone looks at them.

koops76:
Closing until #10843 is done.

cschanaj:
I agree that this is a policy decision, noted that we don't have any downgrade rule anymore, see https://github.com/EFForg/https-everywhere/issues/9621#issuecomment-299241902. Re-writing HTTPS to HTTPS should be safe as long as the users are not exposed to plain-text connection. Let's see @Hainish opinion on this.

Besides, @koops76 why do you need to modify the code base for this? I just have a quick look of the diff on GitHub, the modifications are mostly changing `var` to `const`, why are these neccessary for resolving this issue?

koops76:
@cschanaj 
background.js:
https://github.com/EFForg/https-everywhere/pull/10839/files#diff-81d05cf2685349d7bbbe9f394bd74506R573

manifest.json:
https://github.com/EFForg/https-everywhere/pull/10839/files#diff-a951caac7ef604fdff31e05ef8ca45d4R40

var -> const/let replacement probably should have been in its own pull request.

cschanaj:
@koops76 I think it will be much better to have a separate PR for the replacement.

koops76:
@cschanaj I cleaned out all these var to const/let changes from this PR.

koops76:
Again, closing until #10843 is done.

koops76:
@cschanaj I moved JS style changes into #10852.

jeremyn:
@cschanaj If nothing but HTTP is needed then I'm okay with restricting the permissions to just HTTP. This improves security, for example it would make it significantly more difficult for a malicious contributor to sneak an unexpected HTTPS rewrite into a ruleset.

If we make this change to the extension permissions, then we should also add a check that the `rule` `from` attribute begins with `^http`. Otherwise a mistake will manifest as a permission error instead of a value error. We can do that in the XML DTD as described [here](https://stackoverflow.com/a/27862735). Eventually we can make the `^http` optional/implied.

koops76:
@jeremyn A malicious contributor can do *anything*, this doesn't prevent malicious code modification.

jeremyn:
A malicious contributor has to get their changes past a reviewer first. A reviewer might overlook something like this

```xml
<rule from="https://good\.example\.com/" to="https://evil.example.org/" />
```

buried in a large PR, especially if the PR requires a lot of discussion or is rebased by the contributor, but the reviewer will almost certainly notice unexpected changes to `.js` or `.json` files.

cschanaj:
@jeremyn This is also true for HTTP rewrite, for example

```xml
<rule from="http://good\.example\.com/" to="https://good.example.org.localhost/" />
```
will block all the connection to `good.example.com` over plain-text. IMHO, a probable way to remedy this is to restrict rewrite such that `from` and `to` have to be over the same domain (TLD plus one). 

We should allow rewrite like (same domain)
```xml
<rule from="http://good\.example\.com/" to="https://secure.example.com/" />
```
but not (cross domain)
```xml
<rule from="http://good\.example\.com/" to="https://good.example.org/" />
```

Though I guess a significant number of existing rulesets will fail this requirement.

koops76:
@cschanaj That would break lots of rules that redirect from a custom domain to a canonical CDN domain to fix certificate mismatch errors.

jeremyn:
@cschanaj I agree that a reviewer could also miss a malicious HTTP->HTTPS rewrite, however HTTPS->HTTPS rewrites are potentially more dangerous, because even sites with poor HTTPS support will usually make sure critical stuff like logins go through HTTPS. There is almost no chance that a POST request to `login.example.com` with a user's name and password will go through HTTP, which means that it's safe from any malicious HTTP rewrites by HTTPS Everywhere.

cschanaj:
@koops76 Yes, but a CDN can be used to serve malicious content as well. 

P.S. Personally, I am unwilling to see https://github.com/EFForg/https-everywhere/pull/10839#issuecomment-312495796 implemented, since it will break too many existing rulesets.

@jeremyn I see your point that rewriting HTTPS might have some security impacts to the users. I agree that we should audit/ update existing ruleset with `HTTPS->HTTPS`. Supposed that we can get ride of all `HTTPS->HTTPS` rewrites in #10843, I agree that _https-everywhere_ should rewrite HTTP request only.

koops76:
@cschanaj I'm talking only about the rulesets that redirect from HTTP-only custom domain to a HTTPS canonical CDN domain.

jeremyn:
FYI we have (collectively) discussed cross-domain rewrites, including CDN-related, several times, for example see the links in https://github.com/EFForg/https-everywhere/issues/7253#issuecomment-250918673.

cschanaj:
@koops76 

Never mind. IMHO, https://github.com/EFForg/https-everywhere/pull/10839#issuecomment-312495796 is not a good idea in the short term because we have limited manpower to review pending PRs. Supposed that we restrict cross domain rewrite, we have to either update/ whitelist existing ruleset. The pending PRs might block new ruleset from getting merged. Also, I am really reluctant to see more whitelist are created.

koops76:
#77

Issue #10843, opened 2017-07-01
If possible, rewrite the rulesets in a way that does not require HTTPS->HTTPS rules.


koops76:
Related: #10839
- [ ] 7Headlines.com.xml
- [ ] Aeriagames.xml
- [ ] AJC.com.xml
- [ ] basekit.com.xml
- [ ] Blip.xml
- [ ] CA-State-Board-of-Equalization.xml
- [ ] Caltech.xml
- [ ] cambridgeshire.gov.uk.xml
- [ ] Care-Net.xml
- [ ] Comic-Con-Intl.xml
- [ ] Cru.org.xml
- [ ] DailyDot.xml
- [ ] Epic-Systems.xml
- [ ] Fasthosts.xml
- [ ] First_Class_Magazine.se.xml
- [ ] Google.xml
- [ ] HealthCheckUSA.xml
- [ ] InkFrog.com.xml
- [ ] Intelliworks_Chat.com.xml
- [ ] Khalsa-Credit-Union.xml
- [ ] KKH.de.xml
- [ ] Knappschaft-Bahn-See.xml
- [ ] MoveOn.xml
- [ ] MyEdAccount.Com.xml
- [ ] MYEDDEBT.com.xml
- [ ] National_Geographic.xml
- [ ] Necuhb.xml
- [ ] Numergy.xml
- [ ] Pair-Networks.xml
- [ ] Parkerpen.com.xml
- [ ] Perfect_Audience.xml
- [ ] Plymouth-University.xml
- [ ] Sberbank_of_Russia.xml
- [ ] SPajIT.xml
- [ ] spectator.co.uk.xml
- [ ] Telus.xml
- [ ] US-Dept-of-Veterans-Affairs.xml
- [ ] VEVO.xml
- [ ] Volgistics.xml
- [ ] Yimg.com.xml
- [ ] yWorks.com.xml
## Done
- [x] AdultShopping.com.xml (#11215)
- [x] Agol.dk.xml (#11210)
- [x] Austin_Fit_Magazine.com.xml (#11216)
- [x] BME.hu_incomplete.xml (#11217)
- [x] BME.hu.xml (#11217)
- [x] CafePress.xml (#10843)
- [x] Cashback.co.uk.xml (#11220)
- [x] CoolCart.Net.xml (#11024)
- [x] CuidadoDeSalud.gov.xml (#11024)
- [x] Dell.xml (#10829)
[...]

PR #10849, opened 2017-07-02
Update Ghostery related rulesets

cschanaj:
#9906


Bisaloo:
Also #10843.

koops76:
@cschanaj Thanks!

src/chrome/content/rules/Ghostery_Enterprise.com.xml
```diff
@@ -1,37 +1,28 @@
...
+			 - ghosteryenterprise.com
+			 - www.ghosteryenterprise.com
+			 - livescan.ghosteryenterprise.com
+
+		Peer certificate cannot be authenticated with given CA certificates:
+			 - cdn.ghosteryenterprise.com
+			 - help.ghosteryenterprise.com
+			 - my-cdn.ghosteryenterprise.com
+			 - site-cdn1.ghosteryenterprise.com
+			 - site-cdn2.ghosteryenterprise.com
```
J0WI: What about 3 and 4?

src/chrome/content/rules/Evidon.xml
```diff
@@ -1,61 +1,26 @@
...
+			 - my.betteradvertising.com
+			 - blog.evidon.com
+			 - my.evidon.com
 -->
-<ruleset name="Evidon Inc (partial)" default_off='failed ruleset test'>
-
-	<!--	Direct rewrites:
-				-->
+<ruleset name="Evidon Inc (partial)">
+	<!-- *betrad.com -->
```
J0WI: Please add at least ^and www (same for evidon)

src/chrome/content/rules/Ghostery_Enterprise.com.xml
J0WI: Thanks!
You can also user RegEx for such comments btw.

cschanaj: I don't prefer regex because I am used to extract the commented domain in the following way. Thanks anyway!!

```bash
grep ' - ' Ruleset.xml | sed 's/\s*- \(.*\).*/\1/g'
```

PR #10856, opened 2017-07-02
Update Cisco.xml

koops76:
#10782

### DIFF
--- a/src/chrome/content/rules/Cisco.xml
+++ b/src/chrome/content/rules/Cisco.xml
@@ -1,18 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Cisco.xml
```diff
@@ -52,6 +39,7 @@ Fetch error: http://solucionespyme.cisco.com/ => https://solucionespyme.cisco.co
 		- eir-stage ˢ
 		- home ᵈ
 		- homesupport ᵈ
+		- solucionespyme ᵐ
```
J0WI: The target is still active

koops76: @J0WI Fixed.

PR #10860, opened 2017-07-02
Replace all occurences of <securecookie host="^\w" name="." /> with <securecookie host=".+" name=".+" />

koops76:


cschanaj:
@koops76 

Given that the expressions are equivalent, I see little urgency and necessity from this PR. I am inclined to think this is a policy decision. 

Besides, https://github.com/EFForg/https-everywhere/pull/10860/commits/ee0f9d375fa10d9fa6715a61fa253c1da0fd4a89 seems unrelated.

koops76:
@cschanaj 

Removing preloaded entries was neccessary for the `preloaded` test to pass.

It seems that none of the modified files are in the whitelist, else `rules` test would have failed.

I probably should have opened an issue and let @Hainish or someone else with write access to the repo to do the replacement but I was bored.

Which changes were not equivalent? Those where I removed `securecookie` tags other than `<securecookie host=".+" name=".+" />`? `<securecookie host=".+" name=".+" />` tag would have secured all cookies any way so these were redundant.

PR #10864, opened 2017-07-03
Change minimum required Chromium version to 51

koops76:
In [/chromium/background.js](https://github.com/EFForg/https-everywhere/blob/master/chromium/background.js#L288):
`for (let ruleset of potentiallyApplicable) {`

`for...of` support came only in Chromium 51. ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Chrome_note_3))

See also: #10852, #10813.


koops76:
Superseded by #10852.

PR #10867, opened 2017-07-03
disable HTTPS for tosc.iacr.org, and support of ia.cr

fabrice102:
`tosc.iacr.org` does not support HTTPS.

IACR also uses the domain `ia.cr`.


fabrice102:
`tosc.iacr.org` now supports HTTPS. This pull request is no more useful.

Issue #10536, opened 2017-06-20
Release 5.2.20 forthcoming


Hainish:
cc @jsha @cooperq @pde @J0WI @fuglede @semenko @loveithateit @diracdeltas @jeremyn

Scheduled for 7/5

Hainish:
Release made: https://github.com/EFForg/https-everywhere/pull/10887

PR #1036, opened 2015-02-10
Disable failing rulesets

jsha:

PR #10869, opened 2017-07-04
Update AM_Best.com.xml, Fix Problematic Rewrite

cschanaj:
@Hainish Any idea why [`AM-Best-Company.xml#L11`](https://github.com/EFForg/https-everywhere/blob/b7650c06f79e61eefdc34f3b7e6a686d990b8e2b/src/chrome/content/rules/AM-Best-Company.xml#L11) passed the fetch test and got re-activate in #10538 while it failed in #1036 ? 

P.S. Possibly #10536 because this is introducing a problematic rewrite in the next release.

P.S.2: AFAIK, this this is the only `target` in the code match does not have any eTLD.


Bisaloo:
From what I have seen, I *think* that the new version of the tests doesn't check the `from` pattern, only the `to`.

PR #10871, opened 2017-07-05
Use URL object instead of document.createElement('a')

koops76:
https://developer.mozilla.org/en/docs/Web/API/URL

While the minimum supported Chrome version is 52, [the extension currently requires at least Chrome 51](https://github.com/EFForg/https-everywhere/pull/10864), so I don't think it's a problem.


J0WI:
cc @Hainish

Hainish:
Merged in 50dc6e30e69c7463269e4887d0699c7bae2d15aa

PR #10872, opened 2017-07-05
Trivialize direct rewrite rules

cschanaj:
A list of candidates is generated such that

 - Neither `default_off` nor `platform`
 - No linked rulesets (no occurrence of the string `.xml`)
 - Exclude all rulesets (8) failing Travis
   * 3dglassesfree.com.xml
   * Ebi.ac.uk.xml
   * RoboForm.com.xml
   * UpsideOut.com.xml
   * CounterPunch.org.xml
   * Kursbuero.de.xml
   * Toshiba.co.jp.xml
   * xServers.xml
 - `#rule` == `#target` == `#non-wildcard target` == 1

Note: The following ruleset from the PR require might some attention (non effective rewrite)
 - E-rewards.com.xml
 - Myftp.utechsoft.com.xml
 - Owl.English.Purdue.edu.xml

Remark: The number of rulesets satisfying these conditions is much lesser than my expectation. IMHO, this is not urgent at all. 

Remark: The code I used is written in Go, I might provide the code to reproduce this change if necessary.

### DIFF
--- a/src/chrome/content/rules/254a.com.xml
+++ b/src/chrome/content/rules/254a.com.xml
@@ -6,7 +6,6 @@
 	<securecookie host="^d\.254a\.com$" name=".+" />
 
 
-	<rule from="^http://d\.254a\.com/"
-		to="https://d.254a.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/3Play-Media.xml
+++ b/src/chrome/content/rules/3Play-Media.xml
@@ -7,7 +7,6 @@
 	<target host="account.3playmedia.com" />
 
 
-	<rule from="^http://account\.3playmedia\.com/"
-		to="https://account.3playmedia.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/419_Eater.com.xml
+++ b/src/chrome/content/rules/419_Eater.com.xml
@@ -19,7 +19,6 @@
 	<securecookie host="^forum\.419eater\.com$" name=".+" />
 
 
-	<rule from="^http://forum\.419eater\.com/"
-		to="https://forum.419eater.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/AAU.at.xml
+++ b/src/chrome/content/rules/AAU.at.xml
@@ -3,7 +3,6 @@
 	<target host="pervasive.aau.at" />
 
 
-	<rule from="^http://pervasive\.aau\.at/"
-		to="https://pervasive.aau.at/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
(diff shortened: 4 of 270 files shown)


J0WI:
Yes, let's just review your Go script instead off 300+ changes.

cschanaj:
@J0WI For your reference: [`TrivializeSimpleRules.go`](https://github.com/cschanaj/covfefe/blob/master/src/TrivializeSimpleRules.go)

J0WI:
> btw, is there any way to measure the performance of this extension before and after these change?
Idk, maybe @Hainish or @jsha know something about performance measurements.

I just had a quick look at your go script, but I think some cases are not handled:
* `from=a to=b` rules (source and target are not the same)
* Rulesets that contain a non-trivial redirect (like `from=a/path to=a/path`) and a generic redirect will break other parts.

Also [this](https://github.com/cschanaj/covfefe/blob/master/src/TrivializeSimpleRules.go#L57) regex looks not fail-safe to me. Could you try to be more XML-like, e.g. `<rule\s+from="[^"]+"\s+/>`?

cschanaj:
@J0WI I have modified the script, it matches trivial rewrite (exact matches) only with an additional condition that no `exclusion`. 

Only (1) ruleset is excluded because it failed Travis
 - `Miretail.com.xml`

J0WI:
LGTM, @Hainish do you have any doubts to merge this?

cschanaj:
@Hainish if you know any method to benchmark the rewrite performance (of this extension), please also let me know.

Hainish:
We don't have any current tools for benchmarking.  We may want to utilize https://benchmarkjs.com/ for this purpose, building it into our selenium tests.

There are a lot of good reasons for trivializing rules, aside from performance.  For instance, some downstream dependencies only use the trivialized rules.  It's also simpler to read.

Bisaloo:
@cschanaj, I am also wondering the impact this could have on memory, which is one of the major complaints about HTTPS-Everywhere.

Hainish:
@Bisaloo @cschanaj trivial rules reduce memory usage, see https://github.com/EFForg/https-everywhere/blob/94de81a773a6e6936ec23850c61d8b7e421606f8/chromium/rules.js#L10-L32

I expect they also increase performance, since the regex is less complex.

cschanaj:
@Hainish @Bisaloo FYI, nearly half (15k+) rules in the code base are _complicated rules_. To reduce memory usage by trivializing rule might require a significant amount of works.

Bisaloo:
An potential easy bulk fix I have noticed looking at old rulesets: remove snapping redirects. Many rulesets with only `^` and `www` as targets redirect to `https://www.$DOMAIN` when `https://$DOMAIN` works as well.

cschanaj:
@Bisaloo I made an experimental PR on the snapping redirects in #11019, 27 out of 1254 (around 2.14%) files redirects `^` and `www` to `www` with working HTTPS connection. I agree that we should fix these ruleset.

PR #10891, opened 2017-07-06
Ignore "rules" directory itself changes when build

yfdyh000:
needs_update always be true for me, perhaps because of "Clean up obsolete ruleset databases" or anything else. This causes "Generating ruleset DB" to always be used when build.


yfdyh000:
supplanted by https://github.com/EFForg/https-everywhere/pull/10894.

PR #10897, opened 2017-07-06
Add ESLint Travis task

koops76:
Note that this doesn't include any ESLint rules and ESLint will always succeed until we'll add the rules. I will create 3 more PRs with different ESLint rules: Standard, Airbnb and Google.


koops76:
ping @Hainish

Hainish:
We've generally put tools like this in [utils/](https://github.com/EFForg/https-everywhere/tree/master/utils)

koops76:
@Hainish It is in `/test/`.

Hainish:
I mean https://github.com/koops76/https-everywhere/blob/725aaa538e85dfc2fb695d63e79a81ca1a296a14/chromium/package.json

koops76:
@Hainish Or do you want ESLint to be installed under `/utils/`? Usually dev dependencies are installed under `node-modules`, just like regular dependencies.

koops76:
@Hainish Yeah, `package.json` is supposed to be in `/chromium/`.

koops76:
@Hainish https://docs.npmjs.com/files/package.json

koops76:
@Hainish Also, `node_modules`, `.gitignore` and `package.json` should be excluded from the crx, just in case.

test/travis.sh
```diff
@@ -42,11 +42,19 @@ set -e
 
 if ! $ONLY_RULESETS_CHANGED; then
   echo >&2 "Core code changes have been made."
+
+  if [ "$TEST" == "eslint" ]; then
+    echo >&2 "Running ESLint."
+    docker run --rm -ti -v $(pwd):/opt node bash -c "cd /opt/chromium && npm install && npm test"
+    [ `git diff --name-only | wc -l` -eq 0 ]
```
Hainish: This command does not generate a diff, it only reports errors (since the `eslint` `--fix` flag is not used).  Therefore line 48 should be sufficient, 49 can be removed.

koops76: @Hainish Done.

koops76:
@Hainish How can I exclude a file or a directory from the crx?

Hainish:
This *is* a utility run over the `chromium/` codebase.  If there are other utilities added in the future, they would also need to be in their own separate paths in `utils/`.  This is necessary to keep the repo tidy, not just throwing a `package.json` in the `chromium/` codebase for this particular use case.

koops76:
@Hainish We may need `package.json` later, when we will be adding unit tests ran in Node environment using a test runner like Mocha, but OK, I will remove it for now.

koops76:
@Hainish Done.

koops76:
@Hainish Ready for merge.

Hainish:
@koops76 for some reason, the `eslnit` run isn't working as intended.  Even with a junk commit in https://github.com/EFForg/https-everywhere/pull/10899, the `eslint` test passes: https://travis-ci.org/EFForg/https-everywhere/jobs/250977450

koops76:
@Hainish Of course. There are no rules set up to lint against.

Hainish:
@koops76 this was properly failing previously, before moving to `utils/`

koops76:
@Hainish Try now.

koops76:
@Hainish It works.
https://travis-ci.org/EFForg/https-everywhere/jobs/250990887
https://travis-ci.org/EFForg/https-everywhere/jobs/250991973

PR #10900, opened 2017-07-07
Create a generic ruleset-whitelist csv file, separating coverage and …

Hainish:
…fetch whitelisting into different flags (#10227)


koops76:
Maybe that should have been JSON?

Hainish:
We decided on CSV format in #10227, besides since people will be manually adding these lines it's more easily read/modified.

PR #10903, opened 2017-07-07
[LINCS.fr] Add subdomain

Bisaloo:


src/chrome/content/rules/LINCS.fr.xml
```diff
@@ -6,9 +6,12 @@
 
 	<target host="lincs.fr" />
 	<target host="www.lincs.fr" />
+	<target host="salles.lincs.fr" />
```
J0WI: See also `lists.lincs.fr`(m)

PR #10905, opened 2017-07-07
Update NameCheap.xml

koops76:


src/chrome/content/rules/NameCheap.xml
```diff
@@ -22,9 +19,15 @@ Fetch error: http://hosting.www.namecheap.com/ => https://hosting.www.namecheap.
...
-		- hosting.www
 		- manage.www
 
+	Nonfunctional hosts in *namecheap.com:
+
+		- hosting.www (dropped)
+
+	Problematic hosts in *namecheap.com:
+
+		- manage.www (untrusted root)
```
J0WI: Only chain issues

koops76: @J0WI Done.

PR #10912, opened 2017-07-07
Restore some W3C rules

sampablokuper:
Commit 9bf49cbeb4696cdea76df23f685da1d11031f693 disabled the entire ruleset,
apparently on the basis that test.w3.org no longer exists on the server. I
agree that test.w3.org no longer exists, but to disable the entire ruleset
because of just one of its subdomains seems to be throwing the baby out with
the bathwater.

Most of the other subdomains still exist and have HTTPS available, so I
re-enabled them.

I deleted the test.w3.org subdomain because it no longer exists.

I left two subdomains disabled because I am unsure how to handle
them:
- the "people" subdomain, which serves different content depending on
  whether it is accessed via HTTP or HTTPS.
- the "validator-suite" subdomain, which redirects to a page hosted by
  GitHub.


Bisaloo:
Thank you for reactivating this ruleset. It's fine to put `validator-suite.w3.org` in the main ruleset because the certificate is valid.

`people.w3.org` on the other hand should be removed entirely.

Also, you can keep comments about problematic subdomains in the main ruleset as well.

Bisaloo:
> Commit 9bf49cb disabled the entire ruleset,
apparently on the basis that test.w3.org no longer exists on the server. I
agree that test.w3.org no longer exists, but to disable the entire ruleset
because of just one of its subdomains seems to be throwing the baby out with
the bathwater.

Also, for your information, the process to automatically check rulesets is currently being revised and it's possible that in the future, only the broken targets will be removed instead of disabling the whole ruleset.

sampablokuper:
@Bisaloo wrote:
> Thank you for reactivating this ruleset. It's fine to put validator-suite.w3.org in the main ruleset because the certificate is valid.
>
> people.w3.org on the other hand should be removed entirely.
>
> Also, you can keep comments about problematic subdomains in the main ruleset as well.

Done :)

J0WI:
Thanks!

PR #10920, opened 2017-07-07
[NamePros.com] Remove preloaded ruleset.

bardiharborow:
This domain meets the [guideline for removal of HSTS preloaded properties](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#hsts-preloaded-rules).


Zenexer:
I'm in charge of HSTS for `*.namepros.com`; I'm also an active HTTPS Everywhere user.  Will this require any changes on our end to maintain the same level of security?

bardiharborow:
@Zenexer as koops76 said in the other ticket, we're simply removing preloaded rulesets as redundant because preloading is actually a stronger guarantee than HTTPS-everywhere provides, e.g. the existing rule only protects `namepros.com` and `www.namepros.com` whereas preloading protects `*.namepros.com`.

Zenexer:
Awesome, thanks.

bardiharborow:
Fixed by 6ef2a16c0e8fc67cc6d0f25807106c72ca7df9fe.

Issue #10888, opened 2017-07-06
Release 5.2.21 forthcoming


Hainish:
cc @jsha @cooperq @pde @J0WI @fuglede @semenko @loveithateit @diracdeltas @jeremyn

Scheduled for 7/18 or 7/19

koops76:
Oops. @Hainish you included broken code from #10852. Roll it back in the release branch but not in master.

Hainish:
@koops76 this was not included in the `5.2.20` release, and was not included in the (release)[https://github.com/EFForg/https-everywhere/tree/release] branch, which is only used for the HTTPS Everywhere Atlas.

koops76:
I made a PR reverting all changes except the minimum required Chromium version: #10895

BK1233:
Just wondering a lot of rulesets have suddenly failed and been put into the experimental section?

Hainish:
@BK1233 see https://github.com/EFForg/https-everywhere/pull/10538

cschanaj:
@Hainish can you run `hsts-prune` in the next release?

Hainish:
@cschanaj done, 6ef2a16c0e8fc67cc6d0f25807106c72ca7df9fe

cschanaj:
@Hainish thanks!!

Hainish:
Done: https://github.com/EFForg/https-everywhere/pull/11238

PR #10923, opened 2017-07-07
Whitelist platform="cacert" from fetch-test

cschanaj:
on those `default_off='failed ruleset test'` https://github.com/EFForg/https-everywhere/issues/10227#issuecomment-311159236, #10733, #10888


cschanaj:
Closing because it is better to fix this by adding the root certificate of `cacert` to the fetch test.

PR #10927, opened 2017-07-08
Update travis.sh

koops76:
- Clarify that ESLint is only ran on the Chromium version of the extension.

### DIFF
--- a/test/travis.sh
+++ b/test/travis.sh
@@ -68,9 +68,7 @@ if [ "$RULESETS_CHANGED" ]; then
   if [ "$TEST" == "rules" ]; then
     echo >&2 "Performing comprehensive coverage test."
     docker_build
-    docker run --rm -ti -v $(pwd):/opt httpse python utils/ruleset_filenames_validate.py
-    docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/validate.sh"
-    docker run --rm -ti -v $(pwd):/opt httpse bash -c "test/rules.sh"
+    docker run --rm -ti -v $(pwd):/opt httpse bash -c "python utils/ruleset_filenames_validate.py && utils/validate.sh && test/rules.sh"
   fi
 
   if [ "$TEST" == "fetch" ]; then
--- a/test/travis.sh
+++ b/test/travis.sh
@@ -46,7 +46,6 @@ if ! $ONLY_RULESETS_CHANGED; then
   if [ "$TEST" == "eslint" ]; then
     echo >&2 "Running ESLint on the Chromium extension."
     docker run --rm -ti -v $(pwd):/opt node bash -c "cd /opt/utils/eslint && npm install && ./node_modules/.bin/eslint ../../chromium --fix"
-    [ `git diff --name-only | wc -l` -eq 0 ]
   fi
 
   if [ "$TEST" == "firefox" ]; then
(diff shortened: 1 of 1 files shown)


koops76:
@Hainish ping

Hainish:
@koops76 the second point seems unnecessary to me.  `travis.sh` is run within the travis environment.  This is not intended to be run on a development environment, so fixing problems *after* they're detected doesn't actually do much.  And if there are problems detected, the test in its previous form would fail as expected.

koops76:
@Hainish Fixed.

koops76:
@Hainish Can we combine these three lines
```
docker run --rm -ti -v $(pwd):/opt httpse python utils/ruleset_filenames_validate.py
docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/validate.sh"
docker run --rm -ti -v $(pwd):/opt httpse bash -c "test/rules.sh"
```
into
```
docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/ruleset_filenames_validate.py && utils/validate.sh && test/rules.sh"
```
?

Hainish:
`--fix` is still present, but unnecessary.  The error code without `--fix` should be sufficient to fail the test when errors are encountered.

koops76:
@Hainish Fixed.

test/travis.sh
```diff
@@ -83,7 +83,6 @@ if [ "$RULESETS_CHANGED" ]; then
   if [ "$TEST" == "preloaded" ]; then
     echo >&2 "Ensuring rulesets do not introduce targets which are already HSTS preloaded."
     docker run --rm -ti -v $(pwd):/opt -e RULESETS_CHANGED="$RULESETS_CHANGED" node bash -c "cd /opt/utils/hsts-prune && npm install && node index.js"
-    [ `git diff --name-only | wc -l` -eq 0 ]
```
Hainish: This line is necessary, since it uses the utility in line 85 (which does not produce an error code if changes are made) to determine if a ruleset is preloaded, and throw an error code if so.

Hainish:
> @Hainish Can we combine these three lines

Docker is so light-weight that subsequent runs of a prebuilt image should not add much overhead at all.  I find this more readable, as well.

koops76:
@Hainish I already changed it, should it be changed back? Also can Python scripts be run directly without passing them to `python` as command line argument?

test/travis.sh
koops76: @Hainish Ah, ok.

Hainish:
Can we make these changes granular please?  Instead of changing multiple test clauses, just change the comment first, then we can discuss whether other things need to be changed?

koops76:
@Hainish BTW, hsts-prune always hangs when it finishes instead of exiting on Windows. Please look into why it happens.

koops76:
@Hainish OK.

koops76:
@Hainish I thought `[ `git diff --name-only | wc -l` -eq 0 ]` generates a patch that the developer can apply for some reason, this is why I added it and added `--fix` switch to ESLint command.

Hainish:
@koops76 we don't support Windows as a development environment, but would welcome a Dockerfile for `hsts-prune` that specifies which environment it can safely be run in.

koops76:
@Hainish It runs correctly but hangs at the very end. Also why do you not support running `hsts-prune` on Windows?

Hainish:
@koops76 Windows isn't worth our time supporting, not many developers use it, there's a high overhead of maintenance already while only supporting macOS and Debian/Ubuntu/Fedora, and within Windows it's possible to set up a development VM in the other supported environments with no monetary and minimal labor cost.

Hainish:
merged in be2b26ba7fefc5544c67f0c549ad2399d812f037

koops76:
@Hainish `--rm` means that the image is recreated each time. Either remove `--rm` or combine the lines like I did.

Hainish:
@koops76 `--rm` means that the container is deleted after the process finishes.  The image is not recreated.  Container creation is extremely lightweight, and it's more readable this way

PR #10928, opened 2017-07-08
Update and rename PayPal.xml to PayPal.co.uk.xml

koops76:
The entire paypal.com domain is preloaded, only paypal.co.uk isn't


bardiharborow:
That's incorrect. `paypal.com` and `www.paypal.com` are manually preloaded with `include_subdomains` set to `false`.

koops76:
@bardiharborow WTF. Is that even possible?

bardiharborow:
@koops76 it is for large domains that raise a ticket with the Chrome team directly.

PR #10933, opened 2017-07-08
Do not close the popup if extension is enabled/disabled

koops76:


J0WI:
ping @Hainish

Hainish:
This is not a desired behavior.  Popup should close.

Also, since this not an issue particular to webextensions for Firefox, but rather a general webextensions issue, we should just use the `chrome` label for now.  This label should probably change to `webextensions` to be more generic, and we can create a `webextensions-chrome` or just another `chrome` label.

PR #10948, opened 2017-07-08
Update Eventim.xml

J0WI:


J0WI:
#11203

jeremyn:
This ruleset should be broken up into several files based on subdomain.

J0WI:
Yes, I already started to do this with `Parter.ru.xml`. This is primary to solve #11203, which is more important.

jeremyn:
So let's do "bad right-wildcards > good multiple files" instead of "bad right-wildcards > bad single file > good multiple files". We don't need the intermediate step.

J0WI:
Sure, but this would consume a lot of time, while still having right-wildcards in the next releases.

J0WI:
I still think it's easier to have this merged before splitting the file. With the right-wildcard this will result in duplicated targets (I think that's not even allowed by our tests) or we have to maintain a bunch of extra exclusions.

jeremyn:
In your last comment https://github.com/EFForg/https-everywhere/pull/10948#issuecomment-368271038, did you mean "With**out** the right-wildcard..."?

J0WI:
No I mean with.
Let's say I would create a rule for `eventim.fr` now. This targets are also covered by `eventim.*`and `www.eventim.*` in this rule. I want to avoid that.

jeremyn:
You could avoid that problem by making all the changes in this PR.

J0WI:
Would you review such a huge PR?

jeremyn:
That is a fair question. I haven't committed to reviewing anything yet.

However, there should be roughly the same complexity and review requirements whether all the domains are in one ruleset, or broken into several rulesets. I mean that if you have `example.com`, `example.org`, etc, there is just as much to review whether they are all in `Example.xml` or whether you have `Example.com.xml`, `Example.org.xml` etc. So it isn't any "huger" either way.

One of the things that make the combined approach difficult to review is that it is more difficult to do a partial review, because it is all mixed together. If the rulesets are split then you can fully approve `Example.com.xml` and not worry about it again, because the file doesn't get edited, but if it is combined then you have to constantly check for regressions, or people sneaking in extra stuff.

So anyway I don't know what to advise you to do. It's possible you'll spend a ton of time reworking the PR and then it will still just sit there for a while.

jeremyn:
Please add a note when you are done making changes.

J0WI:
I'm already done.

jeremyn:
`Eventim.co.il.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.co.uk.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] `(p)` is undefined
- [ ] Add `target` for https://staging.eventim.co.uk
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.com.br.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] https://eventim.com.br should be `(m)`, not `(t)`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.cu.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] https://eventim.cu should be `(t)`, not `(m)`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.de.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.fr.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] https://eventim.fr should be `(m)`, not `(t)`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.nl.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.no.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.pl.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] https://m.eventim.pl should be `(t)`, not `(m)`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.se.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.sk.xml`:

- [ ] Change the reference to `see Eventim.com.xml`
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s

`Eventim.xml`:

- [ ] Remove https://eventim.no because it's covered in another ruleset
- [ ] Remove https://eventim.pl and https://m.eventim.pl because they're covered in another ruleset
- [ ] Remove https://eventim.sk because it's covered in another ruleset
- [ ] Add `target` for https://api.eventim.com with `test` http://api.eventim.com/eticket/api/html/eticket/download/assets/vendor/css.escape.js
- [ ] Add `target` for https://join.eventim.com with `test` http://join.eventim.com/robots.txt
- [ ] Add `target` for https://redbullairrace.eventim.com
- [ ] Add `test` http://content.eventim.com/static/uploaded/at/files/280b1f3b8225e4e68ccf54194f76142a.pdf (or some better URL)
- [ ] Remove or limit `host`s for the `securecookie` tag to the working `target`s
- [ ] Rename `ruleset` to `Eventim.com`
- [ ] Rename file to `Eventim.com.xml`

J0WI:
Done.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/10948/commits/0bc173714fa7af1d4574bf30121356c290dd8643.

For https://ticketnews.eventim.co.uk I get:

>Secure Connection Failed
>
>An error occurred during a connection to ticketnews.eventim.co.uk. SSL received a record that exceeded the maximum permissible length. Error code: SSL_ERROR_RX_RECORD_TOO_LONG


You've marked this as `plaintext response` but I would call it some variation of `secure connection failed`. Are you getting something different?

jeremyn:
I'm not really sure what you mean by `plaintext response`. Does that normally mean an HTTPS to HTTP redirect?

J0WI:
The HTTPS version of this host returns an unencrypted HTTP response (which leads in an error).

jeremyn:
I'm still not sure what you mean. In my case I'm getting a `Secure Connection Failed` error, with no response -- no response headers or body. What are you seeing?

J0WI:
Firefox hides everything behind a GUI (for good), but you can easily verify with:
http://ticketnews.eventim.co.uk:443/

The server is listening on HTTPS port 443, but returns plain HTTP.

jeremyn:
Thanks for clarifying, the reason makes sense now.

In this case using `plaintext response` for `secure connection failed` seems analogous to using `self-signed` or `certificate expired` instead of `invalid certificate`. I disagree with this extra level of detail in both cases, and I think that in this case we should use `secure connection failed`. However if you feel strongly that this should be `plaintext response`, we can keep it here. Please let me know what you want to do.

J0WI:
I prefer to keep the details.

jeremyn:
Okay, merged.

J0WI:
Thanks!

PR #10964, opened 2017-07-09
[Lemde.fr] Default_off ruleset (fix #10962)

### LINKED ISSUE #10962: Fix rules on lemonde.fr website
1/ Install on Google Chrome HTTPS everywhere extensions
2/ Go to http://www.lemonde.fr/ homepage
3/ The CSS https://asset.lemde.fr/data/0.14.2/css/cadre-page.css ressource can be downloaded with HTTPS.

Note: same issue on website http://libresglisses.blog.lemonde.fr/2017/07/08/surf-une-ceremonie-paddle-out-a-la-memoire-de-jack-oneill/

Please, don't force HTTPS on lemde.fr domain and subdomains.

Bisaloo:
Related: https://github.com/EFForg/https-everywhere/pull/10876 and https://github.com/EFForg/https-everywhere/issues/10962.

Another option would have been to remove the broken subdomain `s1.lemde.fr` but we would need to remove `s2.lemde.fr` as well because it looks like they are used as mirrors.

Also, because lemonde.fr has broken protocol-relative URL in many places in its source code, I would rather not take any chances.

### DIFF
--- a/src/chrome/content/rules/Lemde.fr.xml
+++ b/src/chrome/content/rules/Lemde.fr.xml
@@ -1,5 +1,5 @@
 <!--
-	For more Le Monde related rules, see Le_Monde.fr.xml
+	For more Le Monde related rules, see LeMonde.fr.xml
 
 	Invalid certificate:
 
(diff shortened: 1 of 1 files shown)


Bisaloo:
[`LeMonde.fr.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/LeMonde.fr.xml) should probably be `default_off` for the same reason btw.

J0WI:
Since this is now off anyway you could also merge the hosts as comments into `LeMonde.fr.xml `.

PR #10973, opened 2017-07-09
[Wallabag.org] Remove exclusion

Bisaloo:


src/chrome/content/rules/Wallabag.org.xml
```diff
@@ -1,54 +1,18 @@
...
-		- css, on:
-
-			- doc from $self *
-			- doc from fonts.googleapis.com *
-
-		- favicon on support from www *
-
-	* Secured by us
+	Invalid certificate:
+		v2.wallabag.org
```
J0WI: Wildcard DNS

PR #10976, opened 2017-07-09
[Gitlab.io] Simplify rewrite rule

Bisaloo:


src/chrome/content/rules/GitLab.io.xml
```diff
@@ -2,20 +2,20 @@
...
+		<test url="http://www.gitlab.io/"/>
+		<test url="http://pages.gitlab.io/"/>
+		<test url="http://maxlaumeister.gitlab.io/"/>
+		<test url="http://moshmage.gitlab.io/" />
+		<test url="http://weel.gitlab.io/" />
 
 	<securecookie host=".+" name=".+"/>
 
-	<!-- (6/2016) GitLab.io doesn't support TLS on 4th level domains.
-		See http://docs.gitlab.com/ee/pages/README.html#limitations -->
```
J0WI: This is why we shouldn't change it.

Bisaloo: Right. I forgot *again* that left-hand wildcards can be more than one level deep. I will try and look if that comment is still valid.

J0WI:
ping @Bisaloo

Bisaloo:
> When using Pages under the general domain of a GitLab instance (*.example.io), you cannot use HTTPS with sub-subdomains. That means that if your username/groupname contains a dot, for example foo.bar, the domain https://foo.bar.example.io will not work. This is a limitation of the HTTP Over TLS protocol. HTTP pages will continue to work provided you don't redirect HTTP to HTTPS.

[Retrieved from the Wayback Machine](https://web.archive.org/web/20170121211447/http://docs.gitlab.com/ee/pages/README.html#limitations).

I understand the issue now. It was bugging me because I didn't think you could create subdomains for your GitLab Pages but the dot in the username stuff makes sense. I will now need to find someone with a dot in their username. Or create a test account.

Bisaloo:
I am not sure if this limitation applies to `*.gitlab.io` pages or only to custom domains hosted on GitLab pages.

J0WI:
🔮 `http://jordi.torne.gitlab.io/csvio-f90/`

Bisaloo:
Good job! I added a more detailed comment to explain why this should be kept as it is.

J0WI:
I was surprised myself that searching for `site:*.*.gitlab.io` works.

PR #10977, opened 2017-07-09
add rule for TK Lindau

numismatika:
Telekommunikation Lindau supports TLS.
The energy company that owns TKL doesn't.


src/chrome/content/rules/TK-Lindau.xml
```diff
@@ -0,0 +1,308 @@
...
+		- voicegw-9.tk-lindau.de
+		- voicegw-90.tk-lindau.de
+		- voicegw-91.tk-lindau.de
+		- voicegw-92.tk-lindau.de
+		- voicegw-93.tk-lindau.de
+		- voicegw-94.tk-lindau.de
+		- voicegw-95.tk-lindau.de
+		- voicegw-97.tk-lindau.de
+		- voicegw-98.tk-lindau.de
+		- voicegw-99.tk-lindau.de
```
J0WI: Next time you should use a regex for such comments.

PR #10983, opened 2017-07-09
[Cooliris.com] Simplify ruleset

Bisaloo:


src/chrome/content/rules/Cooliris.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Invalid certificate:
+		blog.cooliris.com
+
+-->
+<ruleset name="Cooliris">
+
+	<target host="cooliris.com" />
+	<target host="www.cooliris.com" />
```
J0WI: `on.cooliris.com`

Bisaloo: It's secured in `Bitly_branded_short_domains.xml`. Should I move it here? Or I can also a comment about it.

J0WI: No, that's fine.

PR #10988, opened 2017-07-09
[Binsearch] Simplify ruleset

Bisaloo:


Bisaloo:
This looks fine to me. I'd rather whitelist it than adding a useless complex rule.

PR #10992, opened 2017-07-09
Remove preloaded domains from PayPal.xml and move PayPal.co.uk into a separate ruleset

koops76:


koops76:
@J0WI Request review.

PR #10560, opened 2017-06-21
Delete Mozilla_Persona.xml and reactivate Mozilla.xml

Bisaloo:
Service discontinued (https://mail.mozilla.org/pipermail/persona-notices/2016/000005.html)

PR #10993, opened 2017-07-09
Update Mozilla.xml

koops76:


Bisaloo:
@J0WI, see https://github.com/EFForg/https-everywhere/pull/10560#issuecomment-310397117. I removed the outdated comment but forgot to remove the `default_off`.

koops76:
@J0WI #11051

Bisaloo:
Other than that, I think it's good to get rid of the wildcard here. We have to update this ruleset every other month to add new exclusion.

And who knows how long those subdomains have been broken before something involved in the project hears about them?

cschanaj:
@koops76 do you mind to merge `krakenbenchmark.mozilla.org` from [`Mozilla-self-signed.xml#L7`](https://github.com/cschanaj/https-everywhere/blob/master/src/chrome/content/rules/Mozilla-self-signed.xml#L7) into `Mozilla.xml`? It is no longer `self-signed`.

koops76:
@cschanaj Done.

koops76:
@J0WI Request review.

J0WI:
I forgot to close this after I merged #11051. I would still like to keep the wildcard here, because Mozilla is actively working on HTTPS/HSTS support for all hosts.

koops76:
@J0WI We should avoid wildcards whenever possible.

koops76:
@J0WI Review before the release please.

koops76:
ping @J0WI.

gloomy-ghost:
Given Mozilla is actively supporting https and the subdomains haven't been migrated are mostly legacy sites that don't receive much visitor, I find it ok to keep the wildcard. At least we know that Mozilla is not going launch a new site using http, so the cost of this approach is still acceptable.

jeremyn:
@gloomy-ghost @J0WI Do either of you have suggestions on what we should do with this PR?

J0WI:
I vote for `wontfix`. Mozilla is actively working on mozilla.org to get it on the HSTS preload list.

Bisaloo:
This PR:

- has conflicts
- has no real changes compared to the `master` version (securecookie and stylistic changes only)
- cannot be updated because the submitter deleted their account
- does not pass the tests (missing `test url`)

Additionnally, two maintainers have voted to close it. Closing.

PR #11000, opened 2017-07-10
[HTTPwatch.com] Remove wildcard

Bisaloo:


J0WI:
See also `bugs.extatic.org`

PR #11005, opened 2017-07-10
[Powervoip.com] Add new rule

libertate:
Added new https rule for powervoip.com

### DIFF
--- a/src/chrome/content/rules/Powervoip.com.xml
+++ b/src/chrome/content/rules/Powervoip.com.xml
@@ -6,6 +6,7 @@
 	client.powervoip.com refused
 	sip.powervoip.com timeout
 	download.powervoip.com refused
+	versions.powervoip.com refused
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Powervoip.com.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Nonfunctional hosts in *.powervoip.com:
+
+	ftp.powervoip.com refused
+	dns.powervoip.com timeout
+	client.powervoip.com refused
+	sip.powervoip.com timeout
+	download.powervoip.com refused
```
J0WI: `versions.powervoip.com` is also refused

libertate: Added

J0WI:
Thanks!

PR #11013, opened 2017-07-11
[IDRIX.fr] Update and reactivate ruleset

Bisaloo:


src/chrome/content/rules/IDRIX.fr.xml
```diff
@@ -1,43 +1,30 @@
...
-
-		- (www.)	(works; mismatched, CN: ssl3.ovh.net)
-
-
-	Observed cookie domains:
-
-		- ^
-		- prod
-		- www
+	2017/06/11: STS includeSubDomains header.
```
J0WI: In this case a wildcard makes sense.

PR #11021, opened 2017-07-11
Create IslamStory.com.xml

sbafio:


J0WI:
Deleted user/repo, rule does not fir our style guide.

PR #11024, opened 2017-07-11
Trivialize snapping redirects (^C, experimental)

cschanaj:
No ruleset needed to be excluded.

### DIFF
--- a/src/chrome/content/rules/CuidadoDeSalud.gov.xml
+++ b/src/chrome/content/rules/CuidadoDeSalud.gov.xml
@@ -5,6 +5,5 @@
 	<securecookie host="www\.cuidadodesalud\.gov"
 			name=".+" />
 
-	<rule from="^(?:http://(?:www\.)?|https://)cuidadodesalud\.gov/"
-		to="https://www.cuidadodesalud.gov/" />
+	<rule from="^http:" to="https:" />
 </ruleset>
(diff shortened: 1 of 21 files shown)


src/chrome/content/rules/CuidadoDeSalud.gov.xml
```diff
@@ -5,6 +5,5 @@
 	<securecookie host="www\.cuidadodesalud\.gov"
 			name=".+" />
 
-	<rule from="^(?:http://(?:www\.)?|https://)cuidadodesalud\.gov/"
```
J0WI: Oh, that's weird. There are around 18 rules like this.

J0WI:
For some reason there are `Disabled by https-everywhere-checker because:` comments in some rulesets, but this is unrelated to your change. Maybe another cleanup task.

J0WI:
Sorry, I just discovered that there are some outdated comments in the ruleset, like `Cert only matches www` or `mismatch on ^`. I'm not sure if we can change them by script.

cschanaj:
@J0WI I don't think removing the outdated comment can be done trivially by script. Unlike rewriting direct rewrite as in #10872, removing snapping rewrite actually introduce new rules to the code base . In some rare case (I don't have an example), it can be problematic.

PR #11025, opened 2017-07-11
Update pps rules

J0WI:


jeremyn:
This ruleset should be broken up into several files based on subdomain.

J0WI:
I don't agree here. In this case, all hosts are just aliases. It makes not much sense to maintain equal rules.

jeremyn:
They are not all aliases. The main sites

* https://piratenpartei.ch
* https://pirateparty.ch
* https://piraten-aargau.ch
* https://piraten-basel.ch

are visibly different from one another. https://discourse.piratenpartei.ch exists but not https://discourse.pirateparty.ch . I'm sure there are many other differences.

Issue #11035, opened 2017-07-12
Progress of Trivializing Snapping Rewrite


cschanaj:
#10872 I would like to partition the massive change into a number of PRs to make sure that Travis build successful. The progress will be slow because trivializing such rules require performing and checking HTTPS connections. 

Example of snapping rewrite (I am not sure if this is the correct word)
```xml
<ruleset name="$Domain">
    <target host="$domain" />
    <target host="www.$domain" />

    <rule from="^http://(?:www\.)?$escape($domain)/"
            to="https://www.$domain/" />
</ruleset>
``` 

Go script used to perform the trivialization: [`rewrite-simple-rules.go`](https://github.com/cschanaj/trivialize-rules/blob/master/src/rewrite-simple-rules.go)

### Excluded Files
Failed Travis Build

 - AAMC.org.xml
 - Aarhus.dk.xml
 - Adbusters.org.xml
 - BulkSMS.xml
 - Dream_News.jp.xml
 - DiscountTheatre.xml
 - Givskudzoo.dk.xml
 - Ibqonline.com.xml
 - IMedia-Communications.xml
 - Institut_fur_Internet-Sicherheit.xml
 - Lokus.no.xml
 - Metrotransit.org.xml
 - NCTA.com.xml
 - National_Academy_of_Engineering.xml
 - Orion.xml
 - Powerbuy.co.th.xml
 - Warner-Artists.xml
 - WestMidlandsPolice.xml
 - Wagner.xml

Preloaded

 - ~Express-VPN.xml~ (deleted via 6ef2a16c0e8fc67cc6d0f25807106c72ca7df9fe)

cschanaj:
Hooray!! Close this because of completion of the task.

PR #11036, opened 2017-07-12
Trivializing Snapping Rewrite (^0 -> ^9)

cschanaj:
#11035

### DIFF
--- a/src/chrome/content/rules/1177.se.xml
+++ b/src/chrome/content/rules/1177.se.xml
@@ -8,7 +8,6 @@
 	<target host="www.1177.se" />
 
 
-	<rule from="^http://(?:www\.)?1177\.se/"
-		to="https://www.1177.se/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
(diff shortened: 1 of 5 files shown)


jeremyn:
The way I verified this was to manually check in a browser that the `^` subdomain for each ruleset, for example https://1177.se , supports HTTPS. It would be nice if there were some way to verify this without having to manually check.

Maybe you could modify your Go program to produce a Bash script that runs `curl -I` for each of the `^` subdomains, and then include these scripts in their corresponding PR. We can then verify that the list of domains in the script matches the list of domains modified in the PR and then run the script to check the network connection.

cschanaj:
@jeremyn @J0WI I have modify the Go script to produce a list of `target` for each modified ruleset, and create a bash script to verify the list of target support HTTPS via `curl -I`, see https://github.com/cschanaj/trivialize-rules/commit/d59ebec457ccf59897f2d2c124f724cb0872d339. 

Or you might want to try https://github.com/cschanaj/xhttpse2 if you think the speed of the script is too low (since the size of the upcoming PRs can be huge).

jeremyn:
Thanks @cschanaj .

PR #11054, opened 2017-07-14
[All-inkl] Remove right-wildcard and simplify ruleset

Bisaloo:
https://github.com/EFForg/https-everywhere/issues/11203

### DIFF
--- a/src/chrome/content/rules/All-Inkl.xml
+++ b/src/chrome/content/rules/All-Inkl.xml
@@ -6,6 +6,16 @@
 	For sale:
 		all-inkl.cn
 		all-inkl.fm
+		all-inkl.ph
+		all-inkl.vg
+		all-inkl.ws
+
+	No content:
+		all-inkl.pw
+		all-inkl.sy
+
+	Different owner:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/All-Inkl.xml
```diff
@@ -1,23 +1,52 @@
+<!--
+	No working URL known:
+		all-inkl.es
+		www.all-inkl.es
+
+	Invalid certificate:
+		(www.)?all-inkl.ch
+		(www.)?all-inkl.de
+		(www.)?all-inkl.it
+		(www.)?all-inkl.uk
```
J0WI: There are much more like this

Bisaloo: you mean more TLDs?

J0WI: yes

Bisaloo: @J0WI, done. I checked every single ccTLD. They don't own all of them so the previous rewrite wasn't really safe/correct.

PR #11073, opened 2017-07-14
wylsa.com

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/wylsa.com.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+blog.wylsa.com ¹
+
+
+¹ mismatch
+-->
+<ruleset name="wylsa.com">
+	<target host="wylsa.com" />
+
+	<target host="www.wylsa.com" />
```
J0WI: Does not exist.

galeksandrp: Deleted.

PR #11081, opened 2017-07-15
[adfoc.us] Add ruleset.

bardiharborow:


src/chrome/content/rules/adfoc.us.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="AdFoc.us">
+
+	<target host="adfoc.us" />
+	<target host="cdn.adfoc.us" />
+	<target host="www.adfoc.us" />
```
J0WI: See also `forum.adfoc.us` (needs the mixed content platform)

bardiharborow: Done.

PR #11106, opened 2017-07-15
[StatusPage.io] Remove outdated rules

Bisaloo:


src/chrome/content/rules/StatusPage.io.xml
```diff
@@ -4,34 +4,12 @@
...
-		- www
-		- *		(per-client subdomains)
-
 -->
 <ruleset name="StatusPage.io">
 
 	<target host="statuspage.io" />
 	<target host="*.statuspage.io" />
 	<target host="metastatuspage.com" />
-	<target host="status-page-blog.herokuapp.com" />
```
J0WI: Why is this target outdated?

Bisaloo: Outdated probably isn't the proper term for this one. But it is already covered in [Heroku.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Heroku.xml).

PR #11109, opened 2017-07-15
Create webpack.js.org.xml

koops76:


J0WI:
How about other hosts in js.org?

koops76:
`js.org` is a public suffix.

J0WI:
Lookls like you can easily fetch them from here: https://github.com/js-org/dns.js.org/blob/master/cnames_active.js
HTTPS seems to be default for new hosts.

koops76:
@J0WI Anyway, each of them should be in its own ruleset, since it's a public prefix.

koops76:
@J0WI Merge this if you can, please.

J0WI:
We should use the same practice as we use for `Github-Pages.xml` or `GitLab.io.xml`: wildcard with exclusions.

koops76:
@J0WI it's a *public prefix*, bruh, just like `.com`. We can't be sure every (or most) `.js.org` domains will support HTTPS. Not all of them are powered by CloudFlare and GitHub Pages, and some of them may still not support HTTPS correctly.

J0WI:
These are also in the [Public Suffix List](https://publicsuffix.org/list/public_suffix_list.dat):
```
github.io
githubusercontent.com
gitlab.io
```
But we maintain wildcard rules for them. In the case of `js.org` we even know which subdomains exists and which are problematic, see https://github.com/EFForg/https-everywhere/pull/11109#issuecomment-317007072.

koops76:
@J0WI Ok, if you're going to maintain the rule and constantly add exclusions, I'll do it this way.

J0WI:
We can automate it from https://raw.githubusercontent.com/js-org/dns.js.org/master/cnames_active.js

koops76:
Duplicate of #11718.

PR #11115, opened 2017-07-15
Update 360.cn.xml

koops76:


src/chrome/content/rules/360.cn.xml
```diff
@@ -15,10 +7,12 @@ Fetch error: http://msoftdl.360.cn/ => https://msoftdl.360.cn/: (60, 'SSL certif
 
 
 	Nonfunctional hosts in *360.cn:
+		- 8.360.cn             ᴵ
```
J0WI: Does not exist anymore

koops76: @J0WI Done.

PR #7924, opened 2016-12-19
The Guardian: add remaining subdomains

numismatika:
The guardian just announced to switch completely to TLS:

https://www.theguardian.com/info/developer-blog/2016/nov/29/the-guardian-has-moved-to-https

Probably needs work on 

- [x] 		- Gu.com.xml
- [x] 		- Guardian.xml
- [x] 		- guardianapps.co.uk.xml
- [x] 		- Guardian_Escapes.com.xml
- [x] 		- Guardian_Offers.co.uk.xml
- [x] 		- Guardian_Subscriptions.co.uk.xml
- [x] 		- Guim.co.uk.xml

as well

PR #11122, opened 2017-07-15
Update The_Guardian.com.xml

koops76:


cschanaj:
related: #7924

J0WI:
#7924 needs work and conflicts anyway, so merging this as a bugfix to re-enable.

Issue #5377, opened 2016-06-28
costco.com broken with https everywhere 


zeugmatis:
Cotsco is one of the biggest retailers in the US: http://www.costco.com ...when viewing images of products with https everywhere enabled, viewing product images by clicking the plus sign in the lower left corner of the main image brings up an applet that goes blank and cannot click any other images or even the "x" in the upper left to close the pop-up window - effectively breaking the site.

zeugmatis:
BTW tested in recent versions of Chrome and Firefox (both recent stable and nightly) on OSX El Capitan.

tomwys:
I got "Access Denied" when clicking that link. Is this only me, or website is completely broken?

zeugmatis:
Please copy/paste the URL?  As opposed to generating randomness, how about fixing the bug?

zeugmatis:
I'm figuring the website is broken - I did let Costco know as well.  Such a large US company though, they did not respond.

jeremyn:
FYI, I can specifically reproduce the problem by going here:

http://www.costco.com/Costco-Cash-Card.product.10024438.html

clicking the magnifying glass in the lower right part of the picture, below the card, and observing that clicking the X in the top right of the interstitial that appears does nothing other than throw messages to the JavaScript console. Turning off the [Costco-static.com ruleset](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/Costco-static.com.xml) fixes the problem.

@fuglede @J0WI What's the fix here? Just disable the static ruleset entirely?

fuglede:
@jeremyn: I can not actually reproduce the issue as described in your comment: indeed I can not close the bigger version of the image, but disabling the ruleset (or HTTPS Everywhere in its entirety for that matter) does not appear to be sufficient to fix the faulty behaviour. I have not looked into it in any greater detail than that, but this also means that I can not quite tell what causes the issue right here and now. In general, the proper fix will depend on how you want to prioritize security versus breakage: it would be nice to pin down the issue exactly, for example to a specific static serving folder which you could then exclude, but often that can be rather involved, and if noone is around to do the job, and if the issue is likely to affect a lot of users (as in this case), it might very well be preferable to simply disable the ruleset. Different people will have different opinions, and I don't think anything has been written down on the matter, but my impression is that most of us prefer to avoid the usability issues even if it costs a bit of security somewhere else.

[...]

PR #11133, opened 2017-07-16
Re-enable and add Costco.xml to ruleset-whitelist.csv

koops76:


Bisaloo:
Related: #5377

J0WI:
Could you also increase the coverage of this rule? Some subdomains can be secured now.

J0WI:
BTW: I think the 403 error was because of the issue with Akamai hosts. Fetch test should pass now.

koops76:
@J0WI Request review.

PR #11141, opened 2017-07-16
Update Postbank.xml

koops76:


src/chrome/content/rules/Postbank.xml
```diff
@@ -14,22 +11,14 @@ Fetch error: http://direkt.postbank.de/ => https://direkt.postbank.de/: (6, 'Cou
 		- direkt
```
J0WI: Please also update the comments.

koops76: @J0WI Done.

PR #11151, opened 2017-07-16
[Asciinema.org] Fix mismatch and reactivate ruleset

Bisaloo:


src/chrome/content/rules/Asciinema.org.xml
```diff
@@ -1,27 +1,17 @@
...
-
-		- blog *
-
-	* Works; mismatched, CN: *.herokuapp.com
+	Invalid certificate:
+		www.asciinema.org
+		blog.asciinema.org
 
 -->
 <ruleset name="Asciinema.org (partial)" default_off='failed ruleset test'>
```
koops76: Remove `default_off='failed ruleset test'`.

src/chrome/content/rules/Asciinema.org.xml
```diff
@@ -46,6 +36,9 @@ Fetch error: http://www.asciinema.org/ => https://www.asciinema.org/: (51, "SSL:
 	<rule from="^http://blog\.asciinema\.org/"
```
koops76: Change `<securecookie host="^asciinema\.org$" name=".+" />` to `<securecookie host=".+" name=".+" />` unless securing only the top domain cookies is your intent.

src/chrome/content/rules/Asciinema.org.xml
```diff
@@ -1,27 +1,17 @@
...
+<ruleset name="Asciinema.org (partial)">
 
 	<target host="asciinema.org" />
-	<target host="blog.asciinema.org" />
 	<target host="www.asciinema.org" />
+	<target host="blog.asciinema.org" />
+	<target host="staging.asciinema.org" />
+
 		<!--
 			Avoid user-visible paths:
```
J0WI: Why are we doing this?

Bisaloo: I'm guessing it's to avoid sending the user to a "strange" URL. They may freak out seeing they landed on a different domain than the one they requested.

J0WI: I fine with removing the complicated exclusion in favor of https and simplicity.

PR #11152, opened 2017-07-16
[Firehol.org] Remove broken target and default_off

Bisaloo:


src/chrome/content/rules/firehol.org.xml
```diff
@@ -1,24 +1,21 @@
...
+
 	<target host="firehol.org" />
 	<target host="www.firehol.org" />
 	<target host="iplists.firehol.org" />
 	<target host="lists.firehol.org" />
 	<target host="netdata.firehol.org" />
 	<target host="netdata1.firehol.org" />
 	<target host="netdata2.firehol.org" />
-	<target host="netdata3.firehol.org" />
 	<target host="test.firehol.org" />
```
koops76: Add `<securecookie host=".+" name=".+" />`.

HTTPSNowhere: @koops76 One question, in general when should we add that line?

koops76: @HTTPSNowhere In most cases.

PR #11154, opened 2017-07-16
[PressFreedomFoundation.org] Remove broken target and reactivate ruleset

Bisaloo:


src/chrome/content/rules/PressFreedomFoundation.org.xml
```diff
@@ -0,0 +1,17 @@
...
+
+	Redirects to freedom.press (preloaded)
+
+-->
+<ruleset name="Press Freedom Foundation.org" >
+
+	<target host="pressfreedomfoundation.org" />
+	<target host="www.pressfreedomfoundation.org" />
+
+	<securecookie host="^\.pressfreedomfoundation\.org$" name=".+" />
```
koops76: Replace `<securecookie host="^\.pressfreedomfoundation\.org$" name=".+" />` with `<securecookie host=".+" name=".+" />`.

PR #11157, opened 2017-07-16
Added energy-charts.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/energy-charts.de.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="energy-charts.de">
+
+	<target host="energy-charts.de" />
+	<target host="www.energy-charts.de" />
+	<target host="dev1.energy-charts.de" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

src/chrome/content/rules/energy-charts.de.xml
```diff
@@ -0,0 +1,9 @@
...
+	<rule from="^http:"
+		to="https:" />
```
koops76: There's no need to split this into two lines.

Issue #11114, opened 2017-07-15
List of all rulesets disabled for failing ruleset test (4301 rulesets, updated 09/19/2017)


ghost:
For everybody's convenience I created a list of all rulesets disabled for failing ruleset test, sorted by Alexa rank.

https://gist.github.com/koops76/770ae067734ae17394761b1f79988a25

Updated 09/19/2017, commit c5ee856.

cschanaj:
@koops76 I have compiled another list separately in [`default-off-alexa-list.csv`](https://gist.github.com/cschanaj/8c76c5a217e8d2a775e71566fdb40355). It includes all `default_off` ruleset with a valid alexa ranking, it also fix some domain detection problem in your list. Hopes this will help.

koops76:
@cschanaj Your list only includes top 1m Alexa. Also I think it misses some of the disabled rulesets.

cschanaj:
Yes, my intention is only to complement your list on the items with domain detection problem, e.g. [`Rakuten.co.jp.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Rakuten.co.jp.xml). Thank you for clarifying this!!

koops76:
@J0WI Review the PRs from here please.

J0WI:
Sadly the list here is gone, so this issue has no value anymore. If anybody could create a new list, our wiki might be a better place.

PR #11159, opened 2017-07-16
Create ArsMathematica.net.xml

8baige:


koops76:
@8baige Can you try to fix the rulesets from this list: #11114

src/chrome/content/rules/ArsMathematica.net.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="ArsMathematica.net">
+	<target host="arsmathematica.net" />
+	<target host="www.arsmathematica.net" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

J0WI: No cookies are set and all hosts are secured.

PR #11162, opened 2017-07-16
Create CedricVillani.org.xml

8baige:


src/chrome/content/rules/CedricVillani.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="CedricVillani.org">
+	<target host="cedricvillani.org" />
+	<target host="www.cedricvillani.org" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #11163, opened 2017-07-16
Create RedditCommentSearch.com.xml

8baige:


src/chrome/content/rules/RedditCommentSearch.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="RedditCommentSearch.com">
+	<target host="redditcommentsearch.com" />
+	<target host="www.redditcommentsearch.com" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #11164, opened 2017-07-16
Create PolymathProjects.org.xml

8baige:


src/chrome/content/rules/PolymathProjects.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="PolymathProjects.org">
+	<target host="polymathprojects.org" />
+	<target host="www.polymathprojects.org" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #11165, opened 2017-07-16
Create recettes.de.xml

8baige:


src/chrome/content/rules/recettes.de.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="recettes.de">
+	<target host="recettes.de" />
+	<target host="www.recettes.de" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #11167, opened 2017-07-16
Added ewrdb.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/ewrdb.com.xml
```diff
@@ -0,0 +1,22 @@
...
+	Timeout:
+		- mx0.ewrdb.com
+-->
+<ruleset name="European Water Ride DataBase">
+
+	<target host="ewrdb.com" />
+	<target host="www.ewrdb.com" />
+	<target host="de.ewrdb.com" />
+	<target host="nl.ewrdb.com" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #11169, opened 2017-07-16
Added fkfev.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/fkfev.de.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Not found:
+		piwik.fkfev.de
+
+-->
+<ruleset name="FKF e.V.">
+
+	<target host="fkfev.de" />
+	<target host="www.fkfev.de" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

src/chrome/content/rules/fkfev.de.xml
```diff
@@ -0,0 +1,13 @@
...
+	<rule from="^http:"
+		to="https:" />
```
koops76: There's no need splitting this into two lines.

src/chrome/content/rules/fkfev.de.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Not found:
+		piwik.fkfev.de
```
J0WI: See also:
dev.fkfev.de
new.fkfev.de

PR #11178, opened 2017-07-16
Add ESLint Travis task (WebExtensions)

koops76:
Like #10897 but for WebExtensions! 😎


koops76:
ping @Hainish

yfdyh000:
@koops76 No "webextensions" in "env"?

koops76:
@yfdyh000 Fixed. Also fixed for `master` in #11190.

Hainish:
It is unnecessary to add this separately to the [`webextensions`](https://github.com/EFForg/https-everywhere/tree/webextensions) branch, because

1. The syntax for [`webextensions`](https://github.com/EFForg/https-everywhere/tree/webextensions) should not differ from the syntax for the current Chrome extension in any significant way
2. We already have a `.eslint.json` file in `master`.  Adding this separately (and with different indentation, to note) will cause merge conflicts down the line.  It's better to modify this file in `master` instead.

PR #10337, opened 2017-06-12
[BritishGas.co.uk] Remove IP and simplify rewrite rule

Bisaloo:
I know I am missing subdomains but I just wanted to remove the IP, not do a complete update of this ruleset.

PR #11187, opened 2017-07-17
Update https_everywhere_checker/check_rules.py, Fix #10877

### LINKED ISSUE #10877: Test if target is a valid host
See #10869

cschanaj:
~Blocked by [`Xxxbunker.com.xml#L44`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Xxxbunker.com.xml#L44)~. See also #11185 Fix #10877. 

Hopefully before #11240 since we have a huge number of PR these days.

P.S. It is assumed that `target` matched the regex in  [`relaxng.xml#L36`](https://github.com/EFForg/https-everywhere/blob/master/utils/relaxng.xml#L36)

### DIFF
--- a/test/rules/src/https_everywhere_checker/rules.py
+++ b/test/rules/src/https_everywhere_checker/rules.py
@@ -1,4 +1,6 @@
+from tldextract import tldextract
 from urlparse import urlparse
+
 
 import regex
 
@@ -178,6 +180,28 @@ class Ruleset(object):
 						self.filename, test.url))
 				self.determine_test_application_run = True
 		return self.test_application_problems
+
+	def getTargetValidityProblems(self):
+		"""Verify that each target has a valid TLD in order to prevent problematic rewrite
+			 as stated in EFForg/https-everywhere/issues/10877. In particular, 
+			 right-wildcard target are ignored from this test.
+
+			 Returns an array of strings reporting any coverage problems if they exist,
+			 or empty list if coverage is sufficient.
+			 """
+		problems = self._determineTestApplication()
+		# Next, make sure each target has a valid TLD
+		for target in self.targets:
+			# Ignore right-wildcard targets
+			if target.endswith(".*"):
+				continue
+				
+			# Extract TLD from target if possible
+			res = tldextract.extract(target)
+			if res.suffix == "":
+				problems.append("%s: Target '%s' missing gTLD/ ccTLD" % (self.filename, target))
+				
+		return problems
 
 	def getCoverageProblems(self):
 		"""Verify that each rule and each exclusion has the right number of tests
--- a/test/rules/src/https_everywhere_checker/rules.py
+++ b/test/rules/src/https_everywhere_checker/rules.py
@@ -1,8 +1,8 @@
 from tldextract import tldextract
 from urlparse import urlparse
 
-
 import regex
+import socket
 
 class Rule(object):
 	"""Represents one from->to rule element."""
@@ -195,11 +195,25 @@ class Ruleset(object):
 			# Ignore right-wildcard targets
 			if target.endswith(".*"):
 				continue
+
+			# Ignore if target is an ipv4 address
+			try:
+				socket.inet_aton(target)
+				continue
+			except:
+				pass
+
+			# Ignore if target is an ipv6 address
+			try:
+				socket.inet_pton(socket.AF_INET6, target)
+				continue
+			except:
+				pass
 				
 			# Extract TLD from target if possible
 			res = tldextract.extract(target)
-			if res.suffix == "":
-				problems.append("%s: Target '%s' missing gTLD/ ccTLD" % (self.filename, target))
+			if res.domain == "" or res.suffix == "":
+				problems.append("%s: Target '%s' missing eTLD" % (self.filename, target))
 				
 		return problems
 
--- a/test/rules/src/https_everywhere_checker/rules.py
+++ b/test/rules/src/https_everywhere_checker/rules.py
@@ -212,8 +212,10 @@ class Ruleset(object):
 				
 			# Extract TLD from target if possible
 			res = tldextract.extract(target)
-			if res.domain == "" or res.suffix == "":
+			if res.suffix == "":
 				problems.append("%s: Target '%s' missing eTLD" % (self.filename, target))
+			elif res.domain == "":
+				problems.append("%s: Target '%s' containing entire eTLD" % (self.filename, target))
 				
 		return problems
 
--- a/test/rules/coverage.checker.config
+++ b/test/rules/coverage.checker.config
@@ -4,6 +4,7 @@
 [rulesets]
 rulesdir = src/chrome/content/rules
 check_coverage = true
+check_target_validity = true
 check_nonmatch_groups = true
 check_test_formatting = true
 include_default_off = false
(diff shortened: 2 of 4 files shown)


Bisaloo:
Note: at some point, the support for right-sided wildcards should probably be removed (see https://github.com/EFForg/https-everywhere/issues/1327#issuecomment-178076565).

But it may be hard for now because we still have:

```sh
grep -l '<target host=".*\.\*" />' *.xml | wc -l
41
```

rulesets using them.

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -179,6 +181,28 @@ def _determineTestApplication(self):
...
+
+			 Returns an array of strings reporting any coverage problems if they exist,
+			 or empty list if coverage is sufficient.
+			 """
+		problems = self._determineTestApplication()
+		# Next, make sure each target has a valid TLD
+		for target in self.targets:
+			# Ignore right-wildcard targets
+			if target.endswith(".*"):
+				continue
```
Hainish: Since IPv4 and IPv6 addresses are [valid subjects](https://tools.ietf.org/html/rfc3280), we should add a conditional here which ignores these targets as well.

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -179,6 +181,28 @@ def _determineTestApplication(self):
...
+				
+			# Extract TLD from target if possible
+			res = tldextract.extract(target)
+			if res.suffix == "":
+				problems.append("%s: Target '%s' missing gTLD/ ccTLD" % (self.filename, target))
```
Hainish: I would change this error message to `"%s: Target '%s' missing eTLD"`

Hainish: Let's add another conditional. `res.domain == ""` should append to `problems` as well, since we don't want targets which include entire eTLDs.

cschanaj:
> Let's add another conditional. res.domain == "" should append to problems as well, since we don't want targets which include entire eTLDs.

@Hainish when running

```bash
$ export TEST=rules
$ ./test.sh
```
Travis gives the following errors if we require `res.domain == ""`

```
ERROR src/chrome/content/rules/PoliceUK.xml: Target 'police.uk' missing eTLD
ERROR src/chrome/content/rules/GOV.UK.xml: Target 'gov.uk' missing eTLD
```

We will need to whitelist these files after this is merge. 

P.S. I expect Travis to run `rules` test even if "Core code changes have been made." Should this be fixed as well?

test/rules/src/https_everywhere_checker/rules.py
Bisaloo: They are valid but I am not convinced it is a good idea to include them in HTTPS-Everywhere rulesets.

[I had a look at rulesets including IP addresses a month ago](https://github.com/EFForg/https-everywhere/pulls?q=is%3Apr+ip+is%3Aclosed+author%3ABisaloo) and most of them were oudated. Best case scenario was the IP address is not responding on port 80. But worst case scenario, HTTPS-Everywhere was hijacking the DNS resolution and redirecting to the wrong domain, which is quite bad (eg [here](https://github.com/EFForg/https-everywhere/pull/10337)).

Adding IP addresses heavily increases the maintenance load without any clear benefit (very very few people will access a website by typing the IP address instead of the domain name).

cschanaj:
@Hainish requested changes are made accordingly, please make a second review when you have the time. thanks.

test/rules/src/https_everywhere_checker/rules.py
Hainish: Regardless of whether we actually have IP addresses in the rulesets, this conditional should be all-encompassing.  That means matching any *potentially* valid ruleset targets.  It would be trivial to match for this.

Hainish:
> Travis gives the following errors if we require `res.domain == ""`
> 
> ```
> ERROR src/chrome/content/rules/PoliceUK.xml: Target 'police.uk' missing eTLD
> ERROR src/chrome/content/rules/GOV.UK.xml: Target 'gov.uk' missing eTLD
> ```

This should have a different error message, and not be included in the same conditional statement, since `missing eTLD` is incorrect in this case.  It also seems incorrect to me to include an eTLD as a target, and in this case it's caused problems. See https://www.sps.gov.uk/ and https://www.lakedistrict.gov.uk/.  Sidestepping this by whitelisting is acceptable for now.

>P.S. I expect Travis to run rules test even if "Core code changes have been made." Should this be fixed as well?

This is unexpected.  Ideally, core code changes should not affect the validity of rulesets.

cschanaj:
@Hainish I have changed the error message in 2f73a16760bcbac8dd43ee2e8881afb157bdf18e. Another PR #11597 will fix Travis build after this PR is merged. thanks!

cschanaj:
@Hainish please help to merge this if this look good for you. thank you!

cschanaj:
ping @Hainish for reviewing the requested changes plus merging  this and #11597

Hainish:
Reviewing now

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -179,6 +181,44 @@ def _determineTestApplication(self):
...
+			# Ignore if target is an ipv4 address
+			try:
+				socket.inet_aton(target)
+				continue
+			except:
+				pass
+
+			# Ignore if target is an ipv6 address
+			try:
+				socket.inet_pton(socket.AF_INET6, target)
```
Hainish: Very nice trick.

test/rules/src/https_everywhere_checker/rules.py
koops76: `^(25[0-5]|2[0-4][0-9]|[01]?[0-9]{1,2})(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]{1,2})){3}$` is the regex for IPv4 addresses (four numbers from 0 to 255 separated by dots).

koops76: IPv6 address regex:
`^(([0-9a-f]{1,4}:){7,7}[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,7}:|([0-9a-f]{1,4}:){1,6}:[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-f]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|[0-9a-f]{1,4}:((:[0-9a-f]{1,4}){1,6})|:((:[0-9a-f]{1,4}){1,7}|:)|fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-f]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))$`

I do not guarantee that it will work.

cschanaj:
many thanks!!

PR #11199, opened 2017-07-17
Update Spring.io.xml

koops76:
#11026


afarrow:
FYI I closed #11026 because it had a merge conflict (I was working off an outdated repository) and opened #11028 which was merged. So I think you can close this issue unless you added something different than what I had (from a quick look it looks like your pull request is basically the same as mine that was merged).

PR #11056, opened 2017-07-14
[Kriso] Remove left- and right-wildcards

Bisaloo:
https://github.com/EFForg/https-everywhere/issues/11203

PR #11200, opened 2017-07-17
Simplify and default_off Krisostomus.xml

cschanaj:
this break page and use undesirable right-wildcard.


cschanaj:
close as duplicated with #11056

PR #11202, opened 2017-07-18
Simplify KickNews.xml

cschanaj:
#11203


J0WI:
Could you please also look for subdomains?

cschanaj:
@J0WI done. thanks!

PR #11204, opened 2017-07-18
Delete Tickengo.xml

cschanaj:
- `tickengo.com` and `www.tickengo.com` redirect to `wingz.me`
 - `tickengo.fr` and `www.tickengo.fr` redirect to `covoiturage-professionnel.com`
 - `fr.tickengo.ca` gone from DNS


J0WI:
We didn't had a rule for `*.wingz.me` yet, so I just created #11326.

Issue #11203, opened 2017-07-18
Remove right-wildcard targets (if possible)


cschanaj:
**This issue might take a long long long time to complete since a rewrite is not always possible.**

Using a modified command from https://github.com/EFForg/https-everywhere/pull/11187#issuecomment-315766782
```bash
egrep -l '<target\s+host=\s*"[^"]*\.\*"\s*/>' *.xml
```

- [x] EBay.xml
- [x] Eventim.xml #10948 
- [x] GoogleImages.xml (e8bb995071c473f4cb30596f42dcaf0a9b523d3a)
- [x] GoogleMaps.xml (30ac44802700ac9c964d53e447e99a75c181e60e)
- [x] GoogleShopping.xml (4c82624193b46e9079067124ced78d1e66713cf7) 
- [ ] Google.tld_Subdomains.xml
- [x] GoogleVideos.xml
- [ ] Google.xml
- [x] Live-Nation.xml (#14255)
- [x] Ticketmaster.xml (#19215)
- [x] Wyndham.xml #12672
- [x] All-Inkl.xml #11054 
- [x] Bytename.xml #11242 
- [x] Capital-One-Connect.xml #11237 
- [x] Directbox.xml #11246 
- [x] EKomi.xml #11302 
- [x] F5-mismatches.xml #12519 
- [x] Foresters.xml #11205 
- [x] GlobalSign.xml #11201 
- [x] GymGlish.xml, GymGlish-problematic.xml #11278 
- [x] ~HSBC.xml~ (does not use a right-wildcard)
- [x] Jako-O.xml #11248 
- [x] KickNews.xml #11202 
- [x] Kimberly-Clark.xml #11249 
- [x] Krisostomus.xml #11056 ~#11200~
- [x] LoveFilm.xml #11207 
- [x] Nokia.xml #11285 
- [x] Nordea.xml #11250 
- [x] Planet-Work.xml #11239
- [x] Red_Bull.xml #11852 
- [x] Rochesterhomepage.net.xml #11551 
- [x] SamKnows.xml #11206 
- [x] Servage.net.xml #11245 
- [x] ShopStyle.xml, ShopStyle-mismatches.xml #12521
- [x] Siemens.com.xml #11451 
- [x] SimFlight-problematic.xml #12616 
- [x] STRATO.xml #11279 
- [x] Thumbshots.xml #11765 
- [x] Tickengo.xml #11204 
- [x] Trusted-Shops.xml #12522 
[...]

PR #11206, opened 2017-07-18
Simplify SamKnows.xml

cschanaj:
#11203


cschanaj:
@J0WI There is 140+ subdomains for `*samknows.com`, I do not want to go through them here. I added some domains from the `Sublist3r` output to improve the coverage. `^` and `www` are the only subdomains on `*samknows.eu`. Thanks!

PR #11208, opened 2017-07-18
[Unblocked] Add domains and simplify ruleset

Bisaloo:
#10843

### DIFF
--- a/src/chrome/content/rules/Unblocked.xml
+++ b/src/chrome/content/rules/Unblocked.xml
@@ -15,7 +15,6 @@
 	<test url="http://rarbg.unblocked.red/" />
 	<test url="http://isohunt.unblocked.red/" />
 
-	<!-- Old Domain -->
 	<target host="unblocked.li" />
 	<target host="*.unblocked.li" />
 	<test url="http://unblocked.li/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Unblocked.xml
```diff
@@ -1,33 +1,73 @@
 <!--
-	HSTS: max-age=15552000; preload || Not in the list yet
+	Other domains:
+		unblocked.pub
```
J0WI: WFM

src/chrome/content/rules/Unblocked.xml
```diff
@@ -1,33 +1,73 @@
...
+		unblckd.biz
```
J0WI: Cannot resolve

J0WI:
I think it's safe to keep a wildcard here, because every mirror get his own subdomain.

src/chrome/content/rules/Unblocked.xml
Bisaloo: Yeah, I was too lazy to add it but I will do it.

src/chrome/content/rules/Unblocked.xml
Bisaloo: WFM

Bisaloo:
> I think it's safe to keep a wildcard here, because every mirror get his own subdomain.

I didn't see this comment. I came here to ask you this. Glad to see we're on the same page.

PR #11210, opened 2017-07-18
[Agol.dk] Remove https rewrite

Bisaloo:
#10843


Bisaloo:
@J0WI, FYI, we don't get notifications when you add tags. Conflicts would probably be solved faster if you could simply ping us as well.

J0WI:
Ok, I'll try to ping you in future. Btw: you should prefer rebaseing over merge commits in the case of a conflict.

Bisaloo:
> Btw: you should prefer rebaseing over merge commits in the case of a conflict.

Is there a technical reason for this? Or is it simply to avoid merge commits?

I usually work on Github for simple cases so I also solve conflicts via the web interface when possible. I only fetch, rebase and force push when it's too complicated.

J0WI:
It's simply to avoid merge commits, because they blow up the repo. But I'm not sure if this is still the case after I squashed your commits.

PR #11221, opened 2017-07-18
Update Demonoid.xml

koops76:
#10843

### DIFF
--- a/src/chrome/content/rules/Demonoid.xml
+++ b/src/chrome/content/rules/Demonoid.xml
@@ -2,6 +2,7 @@
 	<target host="demonoid.pw" />
 	<target host="www.demonoid.pw" />
 
+	<!-- Old domain -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Demonoid.xml
```diff
@@ -2,13 +2,6 @@
 	<target host="demonoid.pw" />
 	<target host="www.demonoid.pw" />
 
-	<!-- Old Domain-->
-	<target host="demonoid.ph" />
-	<target host="www.demonoid.ph" />
-	<rule from="^https?://(www\.)?demonoid\.ph/" to="https://$1demonoid.pw/" />
```
J0WI: You can just remove the `s` here, everything else works.

koops76: Why should we be redirecting it to a new domain?

PR #11223, opened 2017-07-18
Rtcquickstart.org

MilanKral:


src/chrome/content/rules/RTCQuickStart.org.xml
```diff
@@ -0,0 +1,11 @@
+<!--
++	https://www.rtcquickstart.org has certificate mismatch, but
++	http://www.rtcquickstart.org redirects to http://rtcquickstart.org
```
koops76: Remove this comment.

src/chrome/content/rules/RTCQuickStart.org.xml
```diff
@@ -0,0 +1,11 @@
...
++	https://rtcquickstart.org has correct certificate
+-->
+<ruleset name="RTCQuickStart.org">
+	<target host="rtcquickstart.org" />
+	<target host="www.rtcquickstart.org" />
+
```
koops76: Add a rule that redirects from www.rtcquickstart.org to rtcquickstart.org.

PR #11228, opened 2017-07-18
[VirusBulletin.com] Simplify and reactivate ruleset

Bisaloo:
#10843


koops76:
@Bisaloo Can you also work at the rulesets from my failed ruleset test list (https://gist.github.com/koops76/770ae067734ae17394761b1f79988a25)?

PR #11229, opened 2017-07-18
Delete NYDailyNews.xml

koops76:
The certificate chain is incomplete, there are CORS and mixed content issues breaking the website, the rules are extremely complicated and when you remove the rules that do not work, almost nothing is left. I guess it's too broken to fix.


koops76:
Replaced by #11514.

PR #11232, opened 2017-07-18
Update Migros-Bank.xml

boleros:
Some cert issues were fixed, some hosts are gone.

### DIFF
--- a/src/chrome/content/rules/Migros-Bank.xml
+++ b/src/chrome/content/rules/Migros-Bank.xml
@@ -18,10 +18,12 @@
 
 	<target host="banquemigros.ch" />
 	<target host="www.banquemigros.ch" />
-	
+
 	<test url="http://www.migrosbank.ch/" />
-	<test url="http://blog.migrosbank.ch/" />
 	<test url="http://standorte.migrosbank.ch/" />
+	<test url="http://blog.migrosbank.ch/" />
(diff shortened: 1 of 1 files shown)


koops76:
@boleros See https://github.com/EFForg/https-everywhere/blob/master/ruleset-testing.md.

src/chrome/content/rules/Migros-Bank.xml
```diff
@@ -1,42 +1,35 @@
 <!--
 	For other Migros coverage, see Migros.xml.
-
-	Mismatch:
-		- blog.bancamigros.ch
-		- p2p.bancamigros.ch
-		- blog.banquemigros.ch
-		- p2p.banquemigros.ch
```
J0WI: Please keep those comments, they are still up to date.

J0WI: The p2p domain is still mismatching too.

boleros: OK, I will let get back to you in a few days, when the hosts are actually gone.

J0WI: Are you involved? Then you should have a look at preloaded HSTS:
https://hstspreload.org/

J0WI: Seems like they are gone now.

PR #11237, opened 2017-07-18
Delete Capital-One-Connect.xml

cschanaj:
All target `timeout` over HTTPS. #11203


J0WI:
They even timeout on http.

PR #11243, opened 2017-07-19
Delete Sandbag.xml

Bisaloo:
All subdomains time out over both http and https.

#10843


J0WI:
Subdomains are still up.

Bisaloo:
I found more subdomains but they are mismatched. Did you find valid targets?

J0WI:
Nope, thanks!

PR #11244, opened 2017-07-19
Update and rename ToSDR.org.xml to ToS_DR.xml

koops76:


Bisaloo:
Actually, I renamed this one. I personally prefer when the ruleset name is the same as the domain name. That way, you can guess the domain name just by looking at the list of files. 

Only exception is when several domains are covered in a single ruleset.

J0WI:
Underscores are not preferable (https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#filenames)

koops76:
@Bisaloo @J0WI Done.

Bisaloo:
Also, why did you change targets ordering?

koops76:
@Bisaloo oops.

PR #11251, opened 2017-07-20
Create CableForum.co.uk.xml

sampablokuper:


src/chrome/content/rules/CableForum.co.uk.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Nonfunctional hosts in *.cableforum.co.uk:
```
J0WI: `m.cableforum.co.uk`

sampablokuper:
@J0WI [wrote](https://github.com/EFForg/https-everywhere/pull/11251#pullrequestreview-51930390):
> This breaks the Google search box due mixed content blocking.

Thanks again for your scrutiny :)

Assuming you mean the search box on the site's home page: on at least some modern browsers, it seems broken regardless. For me, at least, adding the rule to HTTPS Everywhere doesn't seem to make that Google Custom Search instance any more or less broken than it is without the rule.

PR #11253, opened 2017-07-20
Create Reprieve.org.uk.xml

sampablokuper:


src/chrome/content/rules/Reprieve.org.uk.xml
```diff
@@ -0,0 +1,18 @@
...
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Reprieve">
+	<target host="reprieve.org.uk" />
+	<target host="www.reprieve.org.uk" />
```
J0WI: Please also add `act.reprieve.org.uk`

src/chrome/content/rules/Reprieve.org.uk.xml
```diff
@@ -6,6 +6,8 @@
 	r: connection refused
 	s: self-signed certificate
 	t: timeout on https
+
+        r: act.reprieve.org.uk (gives 404 error)
```
J0WI: You need to use a custom test url, like `http://act.reprieve.org.uk/page/s/join`

PR #11254, opened 2017-07-20
add rule for RadioTamazuj

numismatika:


src/chrome/content/rules/RadioTamazuj.xml
```diff
@@ -0,0 +1,29 @@
...
+		- browse.radiotamazuj.org
+		- investigations.radiotamazuj.org
+		- m.radiotamazuj.org
+		- new.radiotamazuj.org
+		- news.radiotamazuj.org
+		- old.radiotamazuj.org
+		- w.radiotamazuj.org
+		- week.radiotamazuj.org
+		- wqw.radiotamazuj.org
+		- ww.radiotamazuj.org
```
J0WI: Wildcard DNS

numismatika: so what?
It didn't reply for the full set of subdomains from sublistr and one ip for multiple domains is quite common

J0WI: It makes no sense to list those domains, because you could likely add any domain, including `invalidinvalidinvalid.radiotamazuj.org`. Please just add a note that the host has a wildcard DNS record.

PR #11255, opened 2017-07-20
add rule for SudanTribune.com

numismatika:


J0WI:
@numismatika can you confirm that mixed content is trivial here?

numismatika:
@J0WI it is content from Google Pagead and OpenX. I think breaking advertisments is ok.,

J0WI:
Thanks!

PR #11258, opened 2017-07-20
re-enable Skype.xml

J0WI:


J0WI:
#10782

J0WI:
@jeremyn could you review this to close #10782?

jeremyn:
- [ ] https://downloads.skype.com returns a 404 with a JavaScript redirect to https://go.skype.com. (https://download.skype.com , without the `s`, does that too.) Either change the comment or delete the subdomain entirely.
- [ ] https://secure.skype.com is `Refused` and shouldn't be a `target`.
- [ ] https://downloads-akm.skype.com is not in DNS, so remove it from the top comment.
- [ ] https://mystatus.skype.com is not in DNS, so remove it from the top comment.
- [ ] https://shop.skype.com should be marked as `Refused`.
- [ ] Combine `Nonfunctional subdomains` and `Problematic domains` in one category with some appropriate name.
- [ ] Delete the `CDN buckets`, `Insecure cookies`, and `Not secured by server` comments. I don't know if they are correct but I don't like having them either way.

J0WI:
I was not able to reproduce the `refused` on `secure.skype.com`.
I also found the `Skype_assets.com.xml` ruleset and some more hosts.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/11258/commits/b43a4818b2a7c404e905bd2ad6b8d9ab2ee57b5e.

J0WI:
>  It doesn't seem appropriate to redirect www.skypeassets.com to secure.skypeassets.com. If you accept the mismatch for www, it doesn't redirect itself.

It doesn't redirect itself, but they are identical.

jeremyn:
I get a `503` at http://www.skypeassets.com but a `504` at http://secure.skypeassets.com , so they are not identical.

koops76:
@J0WI Both are server failure codes.

J0WI:
Take any other example like:
```
http://www.skypeassets.com/apollo/2.0.1366/images/components/navigation/singin-sprite-v3.png
https://secure.skypeassets.com/apollo/2.0.1366/images/components/navigation/singin-sprite-v3.png
http://www.skypeassets.com/apollo/2.0.1366/css/scom.css
https://secure.skypeassets.com/apollo/2.0.1366/css/scom.css
```

jeremyn:
@J0WI I still don't really trust a blanket rewrite from `www` to `secure`. Also http://skype.com redirects to https://www.skype.com/en/ (for me) which pulls in files from `secure`, not `www`, so I don't think we're losing much security but not rewriting `www`.

J0WI:
We already had it in `Skype_assets.com.xml`, so we loose coverage if we remove it. Also, we want https everywhere, so if anybody access the `www` for any reason, it's better to secure it than not (or have a cert mismatch error).

jeremyn:
`Skype_assets.com.xml` was off by default before so I'm treating this mostly like a new ruleset.

I agree that so far all the files I've seen match URLs between the two subdomains, but the subdomains are not the same, given the difference I found earlier. You are guessing that the sites otherwise match based on manually testing some files but I don't think that is enough information.

I'm not sure what else to do here.

jeremyn:
I'd like to move this along somehow. Could you please remove the redirect for `www.skypeassets.com`? Otherwise I can remove myself from the PR and someone else can work with you on it.

J0WI:
Okay, I removed it.

jeremyn:
Thanks, merged.

PR #11267, opened 2017-07-21
[Elsevier.com] Add targets

Bisaloo:


src/chrome/content/rules/Elsevier.xml
```diff
@@ -17,131 +17,54 @@
...
 	<target host="api.elsevier.com" />
 	<target host="auth.elsevier.com" />
 	<target host="covers.elsevier.com" />
 	<target host="dev.elsevier.com" />
+	<target host="developers.elsevier.com" />
 	<target host="www.developers.elsevier.com" />
-	<!--target host="ees.elsevier.com" /-->
-	<!--target host="www.ees.elsevier.com" /-->
+		<exclusion pattern="^http://www\.developers\.elsevier\.com/action/" />
+			<test url="http://www.developers.elsevier.com/action/devprojects" />
```
J0WI: Please add a hint why

Bisaloo: Some paths were still redirecting to `http` when I tested it AFAIR but it doesn't seem to be the case anymore.

PR #10965, opened 2017-07-09
Trivialize capturing direct rewrite rules

cschanaj:
- move away from wildcard target if possible
 - let fetch test do its job!

P.S.  ~The Go code is not ready for review yet, please do not merge.~ 

The source code is available now. [`TrivializeWildcardSimpleRules.go`](https://github.com/cschanaj/covfefe/blob/master/src/TrivializeWildcardSimpleRules.go). 

Updates: 

A significant number of ruleset (35) failed the fetch test after the rewrite (see list below). @J0WI , @Hainish should I rewrite and `default_off='failed ruleset test'`ing them in another PR? @jeremyn Sorry for pinging you mistakenly.

 - InfluAds.xml
 - 24-7_Customer.xml
 - Financial-Content.xml
 - Marhab.xml
 - Blaze.com.xml
 - NJ_Edge.Net.xml
 - Enecto.com.xml
 - Ironwhale.com.xml
 - Hitachi_Data_Systems.xml
 - ChampionCasino.net.xml
 - Crimtan.xml
 - Friesland_Bank.nl.xml
 - AWcloud.net.xml
 - Mesh-Internet.xml
 - Brand_Embassy.xml
 - San_Francisco_Marathon.xml
 - Russia.ru.xml
 - Washington_Times.xml
 - Reed_Business_Information_Australia.xml
 - Redcats.xml
 - World_Television.xml
 - Bank_of_Taipei.com.tw.xml
 - AA.net.uk.xml
 - Serverloft.xml
 - Moreover-Technologies.xml
 - AkademiskaHus.se.xml
 - Network-Depot.xml
 - Sanalika.xml
 - Hostname.sk.xml
 - Ve_Interactive.xml
 - Perform.xml
 - University_of_Tennessee_System.xml
 - Nocdirect.com.xml
 - VisibleGains.com.xml
 - Locaweb.xml
 - Blaze.com.xml

PR #11280, opened 2017-07-23
Update Trulia-cdn.com.xml

ivysrono:
#10965
#9473


J0WI:
Maybe you are also motivated to create a rule for `*.trulia.com`?

PR #11309, opened 2017-07-24
Create TomPawlak.org.xml

koops76:


src/chrome/content/rules/TomPawlak.org.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	^ does not exist.
+-->
+
+<ruleset name="TomPawlak.org">
```
J0WI: `www.tompawlak.org`

src/chrome/content/rules/TomPawlak.org.xml
```diff
@@ -1,8 +1,6 @@
-<!--
-	^ does not exist.
--->
-
 <ruleset name="TomPawlak.org">
+	<target host="tompawlak.org" />
```
J0WI: ^ does not exist.

koops76: Does it exist over HTTP?

koops76:
@J0WI `^` is not accessible either over HTTP or HTTPS. `www` is an empty Heroku app. Should we secure it anyway? I also added a rule which redirects from `^` to `www` since many people do not type `www` in the URL (like me).

J0WI:
We don't add targets that do not exist. Please remove `^` completely.
Since `www` returns a 404 you can also just keep a comment for it.

koops76:
@J0WI Done.

PR #11321, opened 2017-07-24
Delete 24-7-Media-clients.xml

Bisaloo:
All targets gone from DNS.


J0WI:
This conflicts for some reason.

Bisaloo:
@J0WI, should I fix `24-7-Media.xml` in this PR or simply whitelist it?

J0WI:
If it's easy to fix, your very welcome to do so. Otherwise I'm okay with whitelisting it.

Bisaloo:
I will probably come back to it at some point because messy rulesets annoy me but I don't have the time right now.

PR #11333, opened 2017-07-25
Update DepFile.com.xml

gopherit:
Adding securecookie tag.

PR #11334, opened 2017-07-25
Add DipFile.com.xml

gopherit:


gopherit:
Consolidating hosts with #11333.

PR #11335, opened 2017-07-25
Re-activate and Update Zuji.com.hk.xml

cschanaj:


J0WI:
This is still disabled.

cschanaj:
@J0WI done. thanks!

PR #11348, opened 2017-07-25
Re-activate and Update US-government.xml

cschanaj:
The following dirty command is being used to update the reference list

```bash
$ grep '\.xml' US-government.xml  | sed 's/\s*- \([^ \t\r\n]*\).*/\1/g' | while read filename; do if [ -f $filename ]; then echo -e '\t\t+' $filename; fi; done | sort
```


cschanaj:
Closed as per discussion in #11510

PR #11352, opened 2017-07-26
Update amazon.co.uk.xml

koops76:
#11114

### DIFF
--- a/src/chrome/content/rules/amazon.co.uk.xml
+++ b/src/chrome/content/rules/amazon.co.uk.xml
@@ -35,8 +35,7 @@
 		slash, but not args:
 	-->
 
-	<rule from="^http://m\.amazon\.co\.uk/[^?]*"
-		to="https://www.amazon.co.uk/gp/aw" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/amazon.co.uk.xml
```diff
@@ -1,8 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Non-2xx HTTP code: http://associates.amazon.co.uk/ (200) => https://associates.amazon.co.uk/ (400)
```
J0WI: Use a test url, like `http://associates.amazon.co.uk/gp/advertising/api/detail/agreement.html`

koops76: @J0WI Done.

J0WI: You also need to add an exclusion for `^http://associates\.amazon\.co\.uk/$`, otherwise the test will still fail.

koops76: @J0WI Done.

PR #11357, opened 2017-07-26
Update IGN.xml

koops76:
#11114


koops76:
@J0WI Request review.

Bisaloo:
This PR has mistakes that cannot be fixed because the submitter has closed their PR. No complicated modifications were made here so we can easily restart from scratch. Closing.

PR #11366, opened 2017-07-26
Delete Dagens_Bedste.dk.xml

Bisaloo:
All targets gone from DNS.


koops76:
@Bisaloo I updated the disabled list (#11114).

Bisaloo:
@koops76, I may have a go at some of them but to be honest, this is not my priority in this project. I am mainly interested in getting a cleaner rulesets codebase (no whitelists, no wildcards, no overly complicated rules, no undue split or pooled rulesets, etc.) so that maintenance is easier and can eventually/maybe get (almost) fully automated

PR #11367, opened 2017-07-26
Update Archive.org_Way_Back_Machine.xml adding subdomain web-beta.archive.org

sdawin:


koops76:
Maybe you can also rename it to `Wayback_Machine.xml`?

sdawin:
in another PR once this one gets merged :)

PR #11373, opened 2017-07-26
Create elong.com.xml

ivysrono:


src/chrome/content/rules/elong.com.xml
```diff
@@ -0,0 +1,23 @@
...
+		^(www.)?
+
+	404：
+		ebooking.elong.com
+		hotel.elong.com
+
+	Strange error page:
+		flight.elong.com
+		lvyou.elong.com
+		trip.elong.com
```
J0WI: Seems like this host has w wildcard DNS anyway.

J0WI: My comment was refering to the whole `*.elong.com`

PR #6316, opened 2016-08-15
CCC: correct comments

numismatika:

PR #11375, opened 2017-07-26
Update CCC.de.xml

cschanaj:
#9582


J0WI:
See also #6316

cschanaj:
@J0WI please feel free to merge #6316 if you prefer. My intention was only to remove [`CCC.de-cacert.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/CCC.de-cacert.xml) from the code base (since a wildcard rewrite for `cacert` is strange). It will be great if it can be done in #6316, however, I do not mind to rebase this afterward in order to achieve the same goal.

J0WI:
closed by #6316

PR #11379, opened 2017-07-26
Update and rename Archive.org_Way_Back_Machine.xml to Wayback_Machine.xml

koops76:
#11367


koops76:
ping @J0WI.

Bisaloo:
This needs changes and the submitter deleted their account. The changes here don't really matter. Closing.

PR #11384, opened 2017-07-26
Delete Nexternal-clients.xml

Bisaloo:
- `blazingthyme.com` is for sale
- `appmarket.nexternal.net` has a valid certificate and can be secured in  Nexternal.xml


src/chrome/content/rules/Nexternal.xml
```diff
@@ -1,22 +1,39 @@
 <!--
-	For clients, see Nexternal-clients.xml.
+	Secure connection failed:
+		host.nexternal.com
+
+	Invalid certificate:
+		nexternal.net
```
J0WI: `www.nexternal.net`

J0WI:
Some tests are failing.

PR #11400, opened 2017-07-26
Update Quran.com.xml

6u621jpq6ybdw:


src/chrome/content/rules/Quran.com.xml
```diff
@@ -1,6 +1,6 @@
 <!--
 	See also:
-	- Quranicaudio.com.xml
+	- QuranicAudio.com.xml
```
J0WI: This file does not exist.

6u621jpq6ybdw:
@J0WI Added.

PR #11437, opened 2017-07-27
Update and rename Windows.net.xml to blob.core.windows.net.xml

koops76:


src/chrome/content/rules/blob.core.windows.net.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	For other Microsoft coverage, see Microsoft.xml.
```
J0WI: Please keep `^` and `www`.
See also:
```
login.windows.net
*.accesscontrol.windows.net
*.dsds.core.windows.net
*.file.core.windows.net

Comment (J0WI) [inline src/chrome/content/rules/blob.core.windows.net.xml]: If you rename the file you also have to update the reference in `Microsoft.xml`.

Comment (koops76) [inline src/chrome/content/rules/blob.core.windows.net.xml]: @J0WI `^` and `www` do not exist. The ones you specified do though.

**Comment by J0WI** [comment]
This rule has some syntax errors.
Please try to find 3 test urls for each wildcard.
Another see also: `*.servicebus.windows.net`.

File: src/chrome/content/rules/blob.core.windows.net.xml (hunk 1 above)
Comment (J0WI) [inline src/chrome/content/rules/blob.core.windows.net.xml]: They do, try
```
curl -I windows.net
curl -I www.windows.net

PR #11452, opened 2017-07-28
Create 6Scan_Security.com.xml and Delete 6Scan.com.xml 

cschanaj:
All subdomains are either gone from DNS, or non-functional over HTTPS.

```xml
<!--
	Non-functional hosts
		SSL peer certificate was not OK:
		- 6scan.com
		- www.6scan.com
-->
<ruleset name="">

</ruleset>
```


J0WI:
Seems like they moved to `6scansecurity.com`. Would you like to replace it with this host?

PR #11454, opened 2017-07-28
 	Update and rename Bohoomil.com.xml to GBConsult.info.xml 

cschanaj:
#9906 Only `^` (after `sublist3r` enumeration) has an existing DNS record, though it is a `mismatched` over HTTPS.


J0WI:
How about redirecting to `gbconsult.info`?

J0WI:
Please also add targets for `*.gbconsult.info` and rename the rule.

PR #11458, opened 2017-07-28
[NetNow.co] Update and merge rulesets

Bisaloo:


Bisaloo:
I added a comment just before the rule so we don't forget about it in case we run the script to remove snapping redirects again.

src/chrome/content/rules/NetNow.co.xml
```diff
@@ -1,50 +1,18 @@
...
-
-
 	Mixed content:
-
-		- css from $self *
-		- fonts from $self *
-		- Images from $self *
-
-	* Secured by us
+		www.netnow.co
```
J0WI: I was not able to reproduce

J0WI: See also `asi.netnow.co`

Bisaloo: I confirm I have MCB issues on `www.netnow.co` on Chromium.

PR #11460, opened 2017-07-28
Create PayMaster.ru.xml

koops76:


src/chrome/content/rules/PayMaster.ru.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Mismatched:
+		- p1.paymaster.ru (paymaster.ru)
```
J0WI: Refused: `info.paymaster.ru`

PR #11462, opened 2017-07-28
Add FreeRADIUS.org

gopherit:


gopherit:
Oh, booo!  Incomplete certificate chain, among other HTTPS server configuration errors.  See [Qualys SSL Labs Test](https://www.ssllabs.com/ssltest/analyze.html?d=freeradius.org) "F" rating.  I should have run this test prior to writing the ruleset.

src/chrome/content/rules/FreeRADIUS.org.xml
```diff
@@ -0,0 +1,22 @@
...
+	Invalid Certificate:
+		- bugs.freeradius.org
+		- ftp.freeradius.org
+		- git.freeradius.org
+		- lists.freeradius.org
+		- mail.freeradius.org
+		- ns2.freeradius.org
+		- power.freeradius.org
+		- luajit.wiki.freeradius.org
+	Problematic HTTPS configuration:
```
Bisaloo: `Problematic HTTPS configuration` is not very informative. Please use `Broken certificate chain` instead or put those subdomains in the `Invalid certificate` section which in note that the chain is missing.

gopherit:
I changed it to "Broken certificate chain," even though there's actually a whole host of issues:
* https://www.ssllabs.com/ssltest/analyze.html?d=wiki.freeradius.org
* https://www.ssllabs.com/ssltest/analyze.html?d=doc.freeradius.org

Both hosts get "F" ratings for several flaws in the HTTPS configuration.

Bisaloo:
I see your edit but I am not sure why tests are passing. The cert chain is indeed broken for `doc` and `wiki`.

https://whatsmychaincert.com/?doc.freeradius.org
https://whatsmychaincert.com/?wiki.freeradius.org

@J0WI, @jeremyn do you have an idea why tests are passing here?

jeremyn:
Both https://doc.freeradius.org and https://wiki.freeradius.org : 

* load for me without a certificate error in a new Firefox profile
* don't report a certificate chain problem at https://www.ssllabs.com
* pass Travis

Other than the failures at https://whatsmychaincert.com they seem okay. I guess there is some problem with https://whatsmychaincert.com and not with `doc` and `wiki`.

Bisaloo:
That makes sense. I shouldn't blindly trust https://whatsmychaincert.com. Testing with a fresh FF profile is indeed a more sensible thing to do.

PR #11465, opened 2017-07-28
[R-project.org] Update subdomains

Bisaloo:


src/chrome/content/rules/R-project.org.xml
```diff
@@ -1,34 +1,29 @@
...
-	Mismatched contents or not supporting HTTPS:
-		* beta.r-project.org
-		* cran.us.r-project.org
-		* debian.cran.r-project.org
-		* ess.r-project.org
-		* latticeextra.r-forge.r-project.org
-		* lists.r-project.org
-		* rhrv.r-forge.r-project.org
-		* search.r-project.org
-		* translation.r-project.org
```
J0WI: Please keep that one as `refused`.

Bisaloo: it's gone from DNS for me. Can you confirm it's working for you over `http`?

J0WI: yes

PR #11475, opened 2017-07-28
Akb.ch.xml

J0WI:


jeremyn:
- [ ] `^` is okay, so remove it from the top comment and remove its special rewrite `rule`
- [ ] https://events.akb.ch is okay, so remove it from the top comment and add a `target` for it
- [ ] Add a `target` for https://portal.akb.ch
- [ ] Add a `target` for https://staging.akb.ch
- [ ] Remove the `securecookie` tag or make it more specific (because we have some failing subdomains)

J0WI:
The `securecookie` rule is only applied on active targets, so that's okay here.

jeremyn:
The potential problem is if one of the working subdomains sets a cookie on `^` that is then requested by one of the failing subdomains. I walk through this sort of problem in https://github.com/EFForg/https-everywhere/pull/7527#issuecomment-262405979. (In my opinion the `securecookie` tag should almost never be used, see https://github.com/EFForg/https-everywhere/issues/2907#issuecomment-245113096.)

J0WI:
All cookies in the listed targets are already served with the secure flag.

jeremyn:
In that case, we can remove `securecookie` entirely.

J0WI:
I don't agree on this. With this argument we could remove the entry rule because the server already redirects to HTTPS. But we want to enforce this client side.

jeremyn:
By "entry rule" I guess you mean the entire `target`. It's not the same though because, as you know, a server redirect to HTTPS still can still require an unencrypted connection from the user that can be attacked. Having a `securecookie` when all the cookies are already secure is roughly equivalent to having an HTTPS Everywhere `target` for a domain that is in the HSTS preload list.

The ideal `securecookie` situation is when all domains are sent through HTTPS, by HSTS preloading or by HTTPS Everywhere, but some of the cookies don't have the secure flag. An attacker might trigger a request that causes an unsecured cookie to be sent, which the attacker can then modify or inspect or whatever. `securecookie` stops this attack but can cause no harm to the legitimate user.

On the other hand, applying `securecookie` to a domain like `^` that might be used by HTTP-only subdomains like `100jahre` is, at best, harmless if the HTTP-only domain doesn't use that cookie. At worst it can break functionality. 

Maybe there is some edge case or functionality that I don't understand. I admit I find `securecookie` logic difficult to reason about sometimes, in particular it's easy for me to see problems but difficult to see advantages, which is one reason I don't like including it.

J0WI:
Well, interestingly `akb.ch` serves a HSTS header with `includeSubDomains` and a long duration (which will break nonfunctional hosts...). So we dan't make anything worse by adding the `securecookie` rule.

jeremyn:
We could make it worse if a user visits a working site (`www`), gets a `^` cookie with the secure flag set by us, and then visits a failing site (`www.100jahre`) that wants to read the cookie, without ever visiting `^` and getting the STS header.

If you really want to include a `securecookie` tag , then you can use the `host` attribute and specifically list subdomains that work with HTTPS except for `^`. This isn't 100% guaranteed to work, because, using `www` as an example, I could repeat my earlier arguments with `www` and subdomains of `www` instead of `^` and subdomains of `^`, but this would be good enough for me.

jeremyn:
If you really want the general `securecookie` here, then we can keep it. I disagree with it but will defer to you in this PR. It's not worth arguing about forever.

jeremyn:
@J0WI Just to clarify, I'm waiting on you to confirm that you want this PR to go as-is.

J0WI:
I confirm. This is blocking me from further work on the `kantonalbanken.xml` rules, so let's continue.

jeremyn:
Merged.

PR #11481, opened 2017-07-29
Delete proxybay.one.xml

koops76:
`proxybay.one` itself is preloaded, and the rest of targets should be done in their own rulesets.

Replaces #11479.


J0WI:
Please create those rulesets before removing this.

koops76:
@J0WI #11478.

koops76:
Duplicate of #11528.

PR #11485, opened 2017-07-29
[DataFile.com] Merge and simplify rulesets

Bisaloo:
Redirect loop


J0WI:
I would prefer to disable it and hopefully we can re-enable it when they finally moved to HTTPS.

jeremyn:
The main pages still have a redirect loop but some of the static elements work with HTTPS, like these from [`datafile.com-resources.xml`](https://github.com/EFForg/https-everywhere/blob/19aff7d76423f1147f0c89f2d656d60cdbe842f2/src/chrome/content/rules/datafile.com-resources.xml#L29-L35)

```xml
<!--	-ve:
		-->
<test url="http://www.datafile.com/css/getpremium_new.css" />
<test url="http://www.datafile.com/favicon.ico" />
<test url="http://www.datafile.com/fonts/helveticaneue-bold/HelveticaNeueCyrBold-webfont.eot" />
<test url="http://www.datafile.com/images/icons/recaptcha/captcha_speak.png" />
<test url="http://www.datafile.com/stylesheet/bundle_b68b43925235522902d187865fd2888a.css" />
```

Can we combine the two ruleset files, remove most of the comments, and just cover the static elements?

src/chrome/content/rules/DataFile.com-resources.xml
```diff
@@ -0,0 +1,23 @@
+<ruleset name="DataFile.com (resources)">
```
J0WI: I think we should name it `partial` to be consistent.

src/chrome/content/rules/DataFile.com-resources.xml
```diff
@@ -0,0 +1,23 @@
...
+	<test url="http://www.datafile.com/images/dflogo_en.png" />
+	<test url="http://www.datafile.com/images/icons/langs/en.png" />
+	<test url="http://www.datafile.com/css/images/ui-bg_highlight-soft_75_cccccc_1x100.png" />
+	<test url="http://www.datafile.com/css/getpremium_new.css" />
+	<test url="http://www.datafile.com/stylesheet/bundle_b68b43925235522902d187865fd2888a.css" />
+	<test url="http://www.datafile.com/javascript/bundle_ac996cee453148b2217dc9a861dbd8d3.js" />
+	<test url="http://www.datafile.com/js/windowAjax.js" />
+
+	<rule from="^http://(www\.)?datafile\.com/(fonts|images|css|javascript|js|stylesheet)/"
+		to="https://$1datafile.com/$2/" />
```
J0WI: `^` redirects to `www`, but HTTPS seems to work there. So we can simplify the rule by redirecting `^` to HTTPS and use complex regexes only for `www`.

src/chrome/content/rules/DataFile.com-resources.xml
```diff
@@ -0,0 +1,23 @@
...
+
+</ruleset>
```
J0WI: See also `wwwtst.datafile.com` and `n[4789]\d` (timeout).

Bisaloo: Do you have an example of a working URL for  a `n[4789]\d` host? It looks like internal subdomains

J0WI:
Meanwhile `wwwtst.datafile.com` got a new cert.

PR #11493, opened 2017-07-29
[Wi-Fi.org] Remove wildcard

Bisaloo:


src/chrome/content/rules/Wi-Fi.org.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Login required:
+		certv1.wi-fi.org
+		login.wi-fi.org
+		certifications.prod.wi-fi.org
+		certifications.staging.wi-fi.org
```
J0WI: `http://certifications.prod.wi-fi.org/pdf/certificate/public/download?cid=WFA65598`
`http://certifications.staging.wi-fi.org/pdf/certificate/public/download?cid=WFA65598`

PR #11495, opened 2017-07-29
Delete FastSSL.net.xml

Bisaloo:
Gone from DNS.


Bisaloo:
https://domlab.net/fastssl.net

J0WI:
How can we know if there are any active subdomains?

Bisaloo:
According to the link I posted, the domain expired in 2016.

PR #11498, opened 2017-07-29
[SSRN.com] Remove wildcard, exclusions and platform

Bisaloo:


src/chrome/content/rules/SSRN.xml
```diff
@@ -1,95 +0,0 @@
-<!--
-		- (www.)ssrnblog.com	(cert: *.myhostcenter.com; presents default cPanel page)
```
J0WI: Please keep this host (mixed content)

PR #11502, opened 2017-07-29
add Globus.ch.xml

J0WI:


jeremyn:
- [ ] Mark https://mails.globus.ch as `Invalid certificate` or similar, and remove its `target`
- [ ] Sort `www` directly below `^`

J0WI:
Done.

PR #11504, opened 2017-07-29
Update Nagano.cz.xml default_off reason

cschanaj:
update `default_off` reason and remove the incomprehensible complicated rule.

### DIFF
--- a/src/chrome/content/rules/Nagano.cz.xml
+++ b/src/chrome/content/rules/Nagano.cz.xml
@@ -1,4 +1,4 @@
-<ruleset name="Nagano.cz" default_off="expired, mismatch, self-signed">
+<ruleset name="Nagano.cz" default_off="ssl-error">
 
 	<target host="nagano.cz"/>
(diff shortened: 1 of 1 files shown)


jeremyn:
I agree that

```xml
	<rule from="^http://([^w]\w*|w[^w]\w*)\.nagano\.cz/"
		to="https://$1.nagano.cz/"/>
```

is hard to read. This ruleset should definitely be improved if it's ever reactivated. However it looks like this site is a directory for pages with URLs like:

* http://af4jf.nagano.cz
* http://ok0be.nagano.cz
* http://ok0epb.nagano.cz

So the `rule` is actually appropriate. We don't need to delete it just because it looks weird.

Would you be okay with just changing the `default_off` reason and leaving the rest of it alone?

cschanaj:
@jeremyn I am fine with changing the `default_off` reason alone. thanks!

jeremyn:
Thanks, merged.

PR #11507, opened 2017-07-29
[PGP.com] Remove wildcard

Bisaloo:


src/chrome/content/rules/PGP.com.xml
```diff
@@ -0,0 +1,34 @@
+<!--
+	Invalid certificate:
+		pgp.com
+		authorize1.pgp.com
+		authorize2.pgp.com
+		download.pgp.com
+		linux-update.pgp.com
+		prm.pgp.com
+		support.pgp.com
```
J0WI: `supportimg.pgp.com`

Issue #11194, opened 2017-07-17
NYDailyNews (partial) breaks images (fonts?)


terrorist96:
http://www.nydailynews.com/entertainment/music/mobb-deep-rapper-prodigy-dead-42-article-1.3262918

`NYDailyNews (partial)` breaks some images (fonts?).

![image](https://user-images.githubusercontent.com/12802056/28289654-70366ad4-6b11-11e7-93d5-a9199e51a635.png)

koops76:
Can you post a developer tools log?

terrorist96:
Could you please give me a step-by-step on how to do that?

koops76:
F12 → Console → right click anywhere → Save as...

koops76:
Also WTF @ your nickname.

terrorist96:
[www.nydailynews.com-1500335216954.zip](https://github.com/EFForg/https-everywhere/files/1154287/www.nydailynews.com-1500335216954.zip)

koops76:
Post it as text without archiving. If it's large post it on Gist.

terrorist96:
Don't trust me? :stuck_out_tongue_closed_eyes: 
https://gist.github.com/terrorist96/c1c9710e3c3a5b855d7e0402e7bfbe15

koops76:
Your log doesn't show any issue. I tried visiting this page with the ruleset both on and off. No difference. Try disabling all extensions except HTTPS Everywhere and see if images are still broken.

koops76:
I have uBlock Origin and Privacy Badger but neither block any images on this page.

terrorist96:
I have those too. But here is a screenshot from a new profile with only HTTPS Everywhere:
![image](https://user-images.githubusercontent.com/12802056/28330441-97996f90-6bbb-11e7-9170-f9286792af34.png)

And here is an image on the same profile with HTTPS Everywhere active but the NYDailyNews rule deactivated:
![image](https://user-images.githubusercontent.com/12802056/28330495-ca94d97a-6bbb-11e7-970c-a68d3c37d044.png)

numismatika:
There is a lot of mixed content warnings scrolling by:

> http://platform.twitter.com/widgets.j
> http://asset.pagefair.com/measure.min.js
> http://assets.nydailynews.com/nydn/js/jquery-ui.js?r=201706290002
> http://c2.taboola.com/nr/nydailynews-nydailynews/newsroom.js
[...]

PR #11514, opened 2017-07-29
Don't rewrite assets.nydailynews.com (#11194)

jeremyn:
This pull request fixes issue #11194 .


src/chrome/content/rules/NYDailyNews.xml
```diff
@@ -27,7 +27,7 @@ Fetch error: http://nydailynews.com/ => http://nydailynews.com/: (60, 'SSL certi
 
 	Problematic subdomains:
 
-		- assets *
+		- assets
```
J0WI: Can you move the comment to the `Nonfunctional` section, with a hint about the issue?

jeremyn: Done.

PR #11542, opened 2017-07-30
[CakeCentral.com] Simplify ruleset

Bisaloo:


src/chrome/content/rules/CakeCentral.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Mixed content:
+		cakecentral.com
+		www.cakecentral.com
+
+-->
+<ruleset name="Cake Central.com (partial)">
+
+	<target host="cdn.cakecentral.com" />
```
J0WI: `cdn1.cakecentral.com`

src/chrome/content/rules/CakeCentral.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Mixed content:
+		cakecentral.com
+		www.cakecentral.com
```
J0WI: They redirect server side to HTTPS. DO you have an example for MCB issues?

Bisaloo: If I didn't include a test URL, it must have been homepage but it looks like it has been fixed now.

PR #11545, opened 2017-07-30
Update GeForce.com.xml

koops76:


src/chrome/content/rules/GeForce.com.xml
```diff
@@ -1,51 +1,28 @@
 <!--
 	For other nVidia coverage, see NVidia.xml.
```
J0WI: There is a bunch of language subdomains that does not match the given certificate, like `cn.geforce.com`, `fr.geforce.com`, `uk.geforce.com` etc.

src/chrome/content/rules/GeForce.com.xml
```diff
@@ -1,51 +1,28 @@
...
+	<target host="www.stagegeforce.geforce.com" />
+	<target host="chn.stagegeforce.geforce.com" />
+	<target host="gbr.stagegeforce.geforce.com" />
+	<target host="gfwsl.stagegeforce.geforce.com" />
+	<target host="twn.stagegeforce.geforce.com" />
+	<target host="demo.uk.geforce.com" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://geforce\.com/" to="https://www.geforce.com/" />
```
J0WI: Please add a comment for the issue on `^`.

src/chrome/content/rules/GeForce.com.xml
koops76: I don't have them rewritten.

J0WI: But could you add a comment about this behavior?

src/chrome/content/rules/GeForce.com.xml
koops76: @J0WI Done.

src/chrome/content/rules/GeForce.com.xml
J0WI: I think we can also remove the `(partial)` flag.

J0WI: ping @koops76

koops76: @J0WI Done.

J0WI:
Thanks!

PR #11553, opened 2017-07-30
Create Sia.tech.xml

afarrow:
Listed in #3069


J0WI:
MCB only affect an SVG image, so I consider it as trivial.

PR #11557, opened 2017-07-30
Delete imwukong.com.xml

Bisaloo:
Gone from DNS.


J0WI:
I'm always careful when deleting "High Value" domains. Let's wait a few weeks to make sure that they don't sell/reuse it.

Bisaloo:
Okay. I don't think the top-1k label is due to the deleted ruleset but to the deleted comment in `Alibaba.xml`.

Issue #7272, opened 2016-10-03
Insufficient test coverage error not triggered by wildcard target and no tests


jeremyn:
The below ruleset, from pull request #7248 , has a `target` with a lefthand wildcard and no tests, so it should trigger the `Ruleset test coverage was insufficient.` error. However it doesn't trigger that error [in Travis](https://travis-ci.org/EFForg/https-everywhere/jobs/164522009) or when I run `fetch-test.sh` against it.

``` xml
<ruleset name="Vimeo Pro.com">
    <target host="vimeopro.com" />
    <target host="www.vimeopro.com" />
    <target host="*.vimeocdn.com" />

    <rule from="^http:" to="https:" />
</ruleset>
```

cschanaj:
#11998

jeremyn:
This is fixed now, thanks @cschanaj .

PR #11560, opened 2017-07-31
Remove false-positive from ruleset-whitelist.csv (full-rules-test)

cschanaj:
As per https://github.com/EFForg/https-everywhere/issues/10227#issuecomment-308940988, replace #11527 

Step to find out false-positive:

 - Create `Dummy.xml` in `rules`
 - `git commit`
 - export TEST=rules
 - ./test.sh
 - `git reset --hard` (back to the automated commit)
 - whitelist the erroneous files from `rules-test` manually.
 - back to step 1 until no error is shown (usually it takes only 1 iterations)


Bisaloo:
Unfortunately, if I got it right, this will only work if we can trust travis to actually fail when the number of tests is not sufficient. But it is not the case, as reported in https://github.com/EFForg/https-everywhere/issues/7272.

I guess it depends on how you consider this whitelist. 

If its function is to auto-pass travis tests when necessary, your changes are fine. But if you think of it as a list of rulesets that don't follow our standard in terms of test urls, this PR will possibly remove true-positive.

cschanaj:
Sadly, the issues in #7272 remain unresolved as described in [`rules.py#L184`](https://github.com/EFForg/https-everywhere/blob/master/test/rules/src/https_everywhere_checker/rules.py#L184)

cowlicks:
@cschanaj any update on this? #11187 is merged

cschanaj:
I will rebase it again later today, do not merge yet.

cschanaj:
it seems that we have a small number of false-positive after #11998 is merged. Anyway, this is ready to be merged.

PR #11561, opened 2017-07-31
Create utils/full-rules-test.sh

cschanaj:
~it will somehow introduce false result, see #11560~ should not be merged before #11998 

ping @Hainish

### DIFF
--- a/utils/full-rules-test.sh
+++ b/utils/full-rules-test.sh
@@ -2,7 +2,7 @@
 
 # Undo whitelisting from `rules-test` if possible
 
-# Change directory to git root; taken from ../test/test.sh
+# Change directory to git root; taken from ../test/rules.sh
 if [ -n "$GIT_DIR" ]
 then
     # $GIT_DIR is set, so we're running as a hook.
--- a/utils/full-rules-test.sh
+++ b/utils/full-rules-test.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Undo whitelisting from `rules-test` if possible
+
+# Change directory to git root; taken from ../test/test.sh
+if [ -n "$GIT_DIR" ]
+then
+    # $GIT_DIR is set, so we're running as a hook.
+    cd $GIT_DIR
+else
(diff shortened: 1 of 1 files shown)


utils/full-rules-test.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Undo whitelisting from `rules-test` if possible
+
+# Change directory to git root; taken from ../test/test.sh
+if [ -n "$GIT_DIR" ]
```
cowlicks: How is this file run? How does `$GIT_DIR` get set? Using something like  `git rev-parse --show-toplevel)` to get the toplevel might make it easier to run this file since it won't depend of mysterious global variables.

cschanaj: I run this by calling `./utils/full-rules-test.sh` in the root directory. It is taken from https://github.com/EFForg/https-everywhere/blob/27f6f230a4f7c7fdff59f20c3642ef3343e2b859/test/rules.sh#L7 

and honestly I do not know where does this variable come from :smile:

cowlicks:
@cschanaj could you please describe the purpose of this PR? It isn't clear to me from the description.

cschanaj:
Currently, we use a whitelist (utils/ruleset-whitelist.csv) to exempt a number of ruleset from Travis test. For some reasons, there are false entry, i.e. we are whitelisting some ruleset unnecessary (as they can pass the Travis without error). This script will remove false whitelisting for rules-test.

Hainish:
@cowlicks for additional context, any ruleset PR will run the `fetch` test on only the targets listed in that PR.

A more comprehensive `full-fetch-test` is something we periodically run, which

1. re-enables all rulesets that have been previously disabled due to fetch test failures
2. tests all enabled rulesets with the fetch test
3. disables all rulesets which fail the fetch test

Here's the code: https://github.com/EFForg/https-everywhere-full-fetch-test

This is done as a batch job, and was run last on [Jun 20](https://github.com/EFForg/https-everywhere/pull/10538).

PR #11564, opened 2017-07-31
Remove duplicated 'newsroom.intel.com'

cschanaj:
remove `newsroom.intel.com` since it is no longer `mixedcontent` and it is duplicated in [`Intel.xml#L141`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Intel.xml#L141)


gloomy-ghost:
Can you also remove the reference at [Intel.xml#L2](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Intel.xml#L2)?

cschanaj:
@gloomy-ghost done. thanks!

PR #11582, opened 2017-07-31
Create Shobohat.com.xml

ihave100openPRs:


jeremyn:
- [ ] I don't see any mixed content blocking that's bad enough to need `platform="mixedcontent"`. Where is the site broken that badly? Otherwise, remove that.

ihave100openPRs:
@jeremyn `https://www.shobohat.com/vb/`

PR #11606, opened 2017-08-01
[FbCDN.net] Remove targets that are gone from DNS

Bisaloo:
Again, it's fine if you need to take some time to review this ruleset.

### DIFF
--- a/src/chrome/content/rules/FbCDN.net.xml
+++ b/src/chrome/content/rules/FbCDN.net.xml
@@ -57,7 +57,8 @@
 				-->
 	<target host="*.xx.fbcdn.net" />
 
-		<test url="http://external-cdg2-1.xx.fbcdn.net/static_map.php?v=29&#x26;osm_provider=2" />
+		<test url="http://external-lht6-1.xx.fbcdn.net/map_tile.php?v=29&osm_provider=2&#x26;x=0&#x26;y=0&#x26;z=0" />
+		<test url="http://external-lht6-1.xx.fbcdn.net/safe_image.php" />
 		<test url="http://profile-a.xx.fbcdn.net/" />
 		<test url="http://profile-a-ord.xx.fbcdn.net/" />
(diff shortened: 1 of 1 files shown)


J0WI:
@jeremyn could you please have an additional look here?

jeremyn:
I agree with all the deletions and changing the `securecookie` to `.+`/`.+` to match the preferred convention in [`CONTRIBUTING.md`](https://github.com/EFForg/https-everywhere/blob/a50b6f21347039b92a12dcebc3d4465fad7167a5/CONTRIBUTING.md#regex-conventions).

The two new `test` URLs return a `404` at their HTTPS equivalents:

* https://external-cdg2-1.xx.fbcdn.net/static_map.php?v=29&#x26;osm_provider=2
* https://static.xx.fbcdn.net/images/emoji.php

Normally we want URLs with `2XX`/`3XX` responses. Can we use other URLs?

Bisaloo:
Done. It's not so easy to find good test urls without leaking any private info with this kind of domain.

src/chrome/content/rules/FbCDN.net.xml
```diff
@@ -73,8 +55,11 @@
 
 	<!--	Direct rewrites:
 				-->
-	<target host=   "*.xx.fbcdn.net" />
+	<target host="*.xx.fbcdn.net" />
 
+		<test url="http://external-cdg2-1.xx.fbcdn.net/map_tile.php?v=29&#x26;osm_provider=2&#x26;x=0&#x26;y=0&#x26;z=0" />
+		<test url="http://external-lht6-1.xx.fbcdn.net/map_tile.php?v=29&#x26;osm_provider=2&#x26;x=0&#x26;y=0&#x26;z=0" />
```
J0WI: I still gets 404s here.

Bisaloo: @J0WI, did you change `&#26;` to `&`?

PR #11614, opened 2017-08-01
Update geekzone.co.nz to cover entire domain

freitasm:
Entire domain served over HTTPS. HTTP requests automatically redirect with 301. HTTPS response headers include HSTS.


J0WI:
This host is already preloaded.

Issue #1998, opened 2015-06-26
Update https-everywhere-checker to disable single target hosts rather than entire rules


jsha:
One common case in https-everywhere-checker is a single hostname that has been disabled or no longer supports HTTPS. Right now the checker will disable the whole ruleset. For large domains with many subdomains, this is more disruptive than is warranted for a single subdomain. It's also correspondingly more likely for such domains.

The checker should be able to automatically comment out single target hosts in rulesets if a fetch test for that hostname fails, and the ruleset is simple (no exclusions, only an http->https rewrite).

reedy:
:+1:

J0WI:
I think this would reduce ruleset maintenance a lot. @Hainish could you have a look at the open PR?

Hainish:
@J0WI I'll be picking up that PR where the contributor left off.

Bisaloo:
This is fixed by #17000

PR #11616, opened 2017-08-01
Cut off problematic domains instead of disabling whole ruleset

galeksandrp:
@Hainish fix for this https://github.com/EFForg/https-everywhere/commit/9bf49cbeb4696cdea76df23f685da1d11031f693

### DIFF
--- a/test/rules/src/https_everywhere_checker/check_rules.py
+++ b/test/rules/src/https_everywhere_checker/check_rules.py
@@ -96,11 +96,11 @@ class UrlComparisonThread(threading.Thread):
 		problems = []
 		for url in task.urls:
 			result = self.processUrl(url, task)
-			if result:
+			if result[0]:
 				problems.append(result)
 		if problems:
 			for problem in problems:
-				logging.error("%s: %s" % (task.ruleFname, problem))
+				logging.error("%s: %s" % (task.ruleFname, problem[0]))
 			if self.autoDisable:
 				disableRuleset(task.ruleset, problems)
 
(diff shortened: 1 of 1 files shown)


cschanaj:
How about commenting the `target` instead? See also the `regex` in [`hsts-prune/index.js#L186`](https://github.com/EFForg/https-everywhere/blob/master/utils/hsts-prune/index.js#L186).

Relate https://github.com/EFForg/https-everywhere/pull/10538#issuecomment-309970098 Conflict #11187

Bisaloo:
https://github.com/EFForg/https-everywhere/issues/1998

J0WI:
@Hainish I marked #1998 as High Priority because it would really reduce the overhead to maintain the rules a lot. Since this PR seems to fix the issue I would love to get some feedback here.

I think this is a blocker and we should not run the script before this issue has been addressed. It's a frustrating work to re-enable  all the rules. May we can start another project to automatically revert parts of 9bf49cbeb4696cdea76df23f685da1d11031f693 and b69a6d7a7baa0ec7842e7a1375d39a29b9192df7 when this has been fixed.

Hainish:
@J0WI thanks for pinging me on this.  I agree, it's far better to disable individual hosts when they are problematic.  I see running the full-fetch-test as a way to periodically ensure quality control for the rulesets, but I can understand that it's a bit like using a hammer to kill a fly.

We should also be careful that upon subsequent runs, those disabled hosts should be able to be checked and enabled again, in case there is a temporary outage.

In https://github.com/EFForg/https-everywhere/pull/10538 there was some discussion on what to do before future runs of the comprehensive fetch test.  I've created a list to better prepare for the next full-fetch-test run: https://github.com/EFForg/https-everywhere/issues/13829

test/rules/src/https_everywhere_checker/check_rules.py
```diff
@@ -96,11 +96,11 @@ def processTask(self, task):
 		problems = []
 		for url in task.urls:
 			result = self.processUrl(url, task)
-			if result:
+			if result[0]:
```
Hainish: I don't see the substantive difference here.  In `processUrl`, if result is not `None`, then it will be defined and have a `0` element.

test/rules/src/https_everywhere_checker/check_rules.py
```diff
@@ -193,16 +193,19 @@ def processUrl(self, plainUrl, task):
...
 	logging.info("Disabling ruleset %s", ruleset.filename)
 	contents = open(ruleset.filename).read()
 	# Don't bother to disable rulesets that are already disabled
 	if re.search("\bdefault_off=", contents):
 		return
-	contents = re.sub("(<ruleset [^>]*)>",
-		"\\1 default_off='failed ruleset test'>", contents)
+	for rule in rules:
+		contents = re.sub('<[ \n]*target[ \n]+host[ \n]*=[ \n]*"%s"[ \n]*\/?[ \n]*>' % (rule.split('/')[2]),
+			"", contents)
```
Hainish: Here, you'll want to remove the entire line, rather than replace just the XML tag with a blank string.

test/rules/src/https_everywhere_checker/check_rules.py
```diff
@@ -193,16 +193,19 @@ def processUrl(self, plainUrl, task):
 			logging.info("Finished comparing %s -> %s. Rulefile: %s.",
 				plainUrl, transformedUrl, ruleFname)
 
-		return message
+		return [message, plainUrl]
 
-def disableRuleset(ruleset, problems):
+def disableRuleset(ruleset, problemsRules):
```
Hainish: You'll probably want to rename this appropriately.

galeksandrp:
> If not, I can take it from here.

@Hainish Feel free to work on this PR. Thanks.

Hainish:
Closing, superseded by #17000. Thanks for getting us started on this @galeksandrp!

PR #11525, opened 2017-07-30
Remove https-everywhere-checker message (2nd run)

cschanaj:
As per https://github.com/EFForg/https-everywhere/pull/11274#issuecomment-318867682

15 of 21 files failed `fetch-test` ... 

```bash
$ grep ERROR 189759654.log | sed 's/ERROR \([^:]*\).*/\1/g' | sort | uniq  
src/chrome/content/rules/7_Springs.com.xml
src/chrome/content/rules/Alfa-Media-Partner.xml
src/chrome/content/rules/Auto_Rims_and_Accessories.xml
src/chrome/content/rules/Brixwork.com.xml
src/chrome/content/rules/BTCZone.xml
src/chrome/content/rules/Covered-California.xml
src/chrome/content/rules/E-Estonia.com.xml
src/chrome/content/rules/Laup.xml
src/chrome/content/rules/Matkahuolto.xml
src/chrome/content/rules/McClatchy_Interactive.com.xml
src/chrome/content/rules/Mirror_Bingo.com-falsemixed.xml
src/chrome/content/rules/Mstar.com.xml
src/chrome/content/rules/Obedovat.sk.xml
src/chrome/content/rules/Packetwerk.com.xml
src/chrome/content/rules/Pan-Islamic_Malaysian_Party.xml
```

PR #11618, opened 2017-08-02
trivialize securecookie (^0 -> ^9)

cschanaj:
GO script used [`trivialize-secure-cookie.go`](https://github.com/cschanaj/lambda/blob/master/src/trivialize-secure-cookie.go)

Similar to previous massive rewrites, subsequent PRs will be much larger (300+ files PRs), please help to review the script if possible. thanks!

```
(1) exclude files which are 'default_off'
(2) exclude files which contain wildcard 'target'
(3) exclude files with 'exclusion' or complicated 'rule'
(4) IF for each `target`, exist at least one 'securecookie#host' match 'target' 
    and 'securecookie#name' is a wildcard. THEN rewrite.
(5) checkout files failing 'fetch-test'
```

P.S. around 3k files are subjected to modification, 1.3k files are ignored because they are `default_off`.

**To reviewer**: if possible, please merge #11525 before this. Otherwise, there might be some huge number of merge conflicts needed to be resolved in future PRs. Thanks!

### DIFF
--- a/src/chrome/content/rules/0x.co.xml
+++ b/src/chrome/content/rules/0x.co.xml
@@ -4,7 +4,7 @@
 	<target host="www.0x.co" />
 
 
-	<securecookie host="^\w" name="." />
+	<securecookie host=".+" name=".+" />
 
 
 	<rule from="^http:"
--- a/src/chrome/content/rules/10antz.co.jp.xml
+++ b/src/chrome/content/rules/10antz.co.jp.xml
@@ -4,7 +4,7 @@
 	<target host="www.10antz.co.jp" />
 
 
-	<securecookie host="^\w" name="." />
+	<securecookie host=".+" name=".+" />
 
 
 	<rule from="^http:"
--- a/src/chrome/content/rules/10dollar.ca.xml
+++ b/src/chrome/content/rules/10dollar.ca.xml
@@ -17,7 +17,7 @@
 					-->
 	<!--securecookie host="^(?:www\.)?10dollar\.ca$" name="^(?:AWSELB$|PHPSESSID)" /-->
 
-	<securecookie host="^\w" name="." />
+	<securecookie host=".+" name=".+" />
 
 
 	<rule from="^http:"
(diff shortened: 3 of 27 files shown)


J0WI:
Please also exclude rules with non-trivial redirect rules or exclusions.

cschanaj:
@J0WI I have modified the GO script via https://github.com/cschanaj/lambda/commit/9d8225d6f96eff536cb11394636835a79205917c and rebased this accordingly. If this is okay, I will work submit larger PRs later. thanks!

J0WI:
I'm a bit unsure about securing all cookies in `platform="mixedcontent"` rules. The rule is not enabled by default, but there might be some edge cases where such bulk changes break cookies.

cschanaj:
how about excluding rules with `platform` as well?

cschanaj:
@J0WI I have partially resolved my questions after some Google search. If I am correct, trivializing `securecookie` is not a trivial task by itself. Reasoning as follow: 

According to https://tools.ietf.org/html/rfc6265

> 5.2.3.  The Domain Attribute

and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

> Contrary to earlier specifications, leading dots in domain names are ignored. If a domain is specified, subdomains are always included.

the `leading dot` in `cookie.domain` are ignored (as implemented in [`chromium/rules.js#L390`](https://github.com/EFForg/https-everywhere/blob/master/chromium/rules.js#L390)), the following type of cookie rule (900+ such)
```xml
<securecookie host="^\." name=".+" />
```
actually never apply (and thus safe to be removed). 

From my understanding of the Mozilla doc, it looks like that all `cookie` are shared by subdomain of `cookie.domain`. It is therefore indeed **dangerous** to apply `securecookie` since a working `target` might have some non-functional subdomains, leading to broken sites. 

That is
```xml
<!--
  cert chain error:
  - img.foo.bar
-->
<ruleset name"Foo.bar">
  <target host="foo.bar" />
  <target host="www.foo.bar" />

  <securecookie host="^(www\.)?foo\.bar$" name=".+" />

  <rule from="^http:" name="https:" />
</ruleset>
```
in fact securing cookie which potentially accessed by `img.foo.bar`

@Hainish could you please help to confirm my assertions when you have the time?

cschanaj:
closing this as trivializing `securecookie` is likely to introduce some unexpected problems.

PR #11633, opened 2017-08-02
re-enable and update RhB.xml

J0WI:


jeremyn:
- [ ] Mark https://apps.rhb.ch as `Refused`, not `Expired`
- [ ] Mark https://erben.rhb.ch as `Refused`, not `Expired`
- [ ] Move all `Expired`/`Self-signed` subdomains under a single `Invalid certificate` category (they generally fail for multiple reasons and it's not worth making distinctions here)
- [ ] Mark https://promo.rhb.ch as `Invalid certificate` and remove its `target`
- [ ] Remove https://s.rhb.ch entirely since it times out with both HTTPS and plain HTTP
- [ ] (Optional) Change `ruleset` `name` to `RhB.ch`

J0WI:
Done

PR #11659, opened 2017-08-03
Create quackwatch.org.xml

6u621jpq6ybdw:


src/chrome/content/rules/quackwatch.org.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="quackwatch.org">
+	<target host="quackwatch.org" />
+	<target host="www.quackwatch.org" />
```
J0WI: `lists.quackwatch.org`

6u621jpq6ybdw:
@J0WI Added.

PR #11666, opened 2017-08-03
Create topachat.com.xml

6u621jpq6ybdw:


src/chrome/content/rules/topachat.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Connection refused:
+		topachat.com
```
J0WI: Could you add a redirect from `^http://topachat\.com/` to `https://www.topachat.com/`?

6u621jpq6ybdw:
@J0WI Added.

J0WI:
You need to add it as target, otherwise the rule is not valid.

6u621jpq6ybdw:
@J0WI Added.

J0WI:
Please add `<target host="topachat.com" />` and remove the test url.

6u621jpq6ybdw:
@J0WI oh right the rule must apply to that domain in the first place, ok fixed now and sorry for that!

PR #11676, opened 2017-08-03
Add OpenNIC.org.xml

koops76:


src/chrome/content/rules/OpenNIC.org.xml
```diff
@@ -0,0 +1,34 @@
...
+
+	<target host="opennicproject.org" />
+	<target host="www.opennicproject.org" />
+	<target host="lists.opennicproject.org" />
+	<target host="api.opennicproject.org" />
+	<target host="servers.opennicproject.org" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://wiki\.opennicproject\.org/.*" to="https://wiki.opennic.org/" />
```
J0WI: `.*` is ugly. Please document why you use this.

koops76: To drop request path.

koops76: @J0WI Done.

PR #11686, opened 2017-08-04
Create teachcraft.net.xml

6u621jpq6ybdw:


src/chrome/content/rules/teachcraft.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="teachcraft.net">
```
J0WI: Please add a hint about the wildcard DNS record.

6u621jpq6ybdw:
@J0WI Added.

PR #11687, opened 2017-08-04
Create GOOD.is.xml

6u621jpq6ybdw:


src/chrome/content/rules/GOOD.is.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Related Ruleset:
+		goodstatic.com.xml
```
J0WI: Does not exist.

src/chrome/content/rules/GOOD.is.xml
```diff
@@ -0,0 +1,19 @@
...
+-->
+<ruleset name="GOOD.is">
+	<target host="good.is" />
+	<target host="www.good.is" />
```
J0WI: `www` is already covered by the wildcard.

src/chrome/content/rules/GOOD.is.xml
```diff
@@ -0,0 +1,19 @@
...
+	<target host="*.good.is" />
```
J0WI: Could you please add a hint why a wildcard is required here?

J0WI: Please note that this could break some hosts, e.g. `shop.good.is`

src/chrome/content/rules/GOOD.is.xml
J0WI: You may add it as a test url, but I think we should avoid wildcrads if possible.

6u621jpq6ybdw:
@J0WI Thanks for the useful comments, discarded the wildcard.

PR #11692, opened 2017-08-04
Update Weibo.com.xml

ivysrono:


src/chrome/content/rules/Weibo.com.xml
```diff
@@ -1,22 +1,21 @@
...
+		.weibo.com
+		game.weibo.com
+		sg2.game.weibo.com
+		yahoo.tw.weibo.com
 
 	MCB:
-		chart.weibo.com
-		e.weibo.com
+		chart.weibo.com		( Too many redirects )
+		e.weibo.com			( https://e.weibo.com/register/register )
```
J0WI: I don't understand why you removed this target

ivysrono: https://e.weibo.com/register/register with http://weibo.com/signup/signup.php

J0WI: Yes, but it was already in the `mixedcontent` ruleset before. So why did you remove it from there?

ivysrono: mixedcontent pages' contents should be https all.
weibo.com/signup/ redirects to http, no https

J0WI: Ok, thanks for clarify.

PR #11706, opened 2017-08-05
add Jakarta Post rules

J0WI:


jeremyn:
`Jakpost.net.xml`:
- [ ] Add `test` for http://img.jakpost.net/c/2017/10/20/2017_10_20_34386_1508471627._medium.jpg

`Thejakartapost.com.xml`:
- [ ] Fix [this comment](https://github.com/J0WI/https-everywhere/blob/0bc3187e11a847933eea7d55ea5a21154d9293f0/src/chrome/content/rules/Thejakartapost.com.xml#L2) referring to `20min`
- [ ] `epapaer` should be `epaper`
- [ ] Mark https://thejakartapost.com as `mixed content` (FYI it has unusual behavior where it displays entirely different content if it detects mixed content blocking)
- [ ] Mark https://events.thejakartapost.com as `mixed content`
- [ ] Mark https://jobs.thejakartapost.com as `mixed content`
- [ ] Add `^`, `events`, and `jobs` as `target`s and make this a mixed content ruleset
- [ ] Remove `(partial)` from the `ruleset` `name`

J0WI:
I kept the `partial` here, because of the redirect on `www`. I removed the `mixed content` comments here, because the platform is already making this clear.

jeremyn:
That's fine, thanks. Merged.

PR #11709, opened 2017-08-05
Create AWSCloud.com.xml

koops76:


src/chrome/content/rules/AWSCloud.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	For other Amazon coverage, see Amazon.xml.
```
J0WI: See also `email.awscloud.com`

koops76: Dropped it seems.

J0WI: Could you add a comment for it?

koops76: @J0WI Done.

PR #11717, opened 2017-08-06
Create HackerFactor.com.xml

6u621jpq6ybdw:
fixes https://github.com/EFForg/https-everywhere/pull/9839


J0WI:
I found only mixed fonts and tracking, so we can safely remove the platform.

6u621jpq6ybdw:
@J0WI Removed.

PR #11718, opened 2017-08-06
Create JS.org.xml

koops76:


koops76:
ping @J0WI

J0WI:
I assume that you generated this ruleset from [here](https://github.com/js-org/dns.js.org/blob/master/cnames_active.js). If you used a script to do so, could you share this please? Then I just have to review the script instead of all rules and we could also use it to update the ruleset in future.

koops76:
@J0WI No, it was manual editing using Sublime Text's multi line selection 😊. I will make the script in some time.

koops76:
@J0WI The JS file has many errors and requires manual editing to correct them. This should be solved on JS.org's side.

J0WI:
Please add a link to the upstream domain list in a comment.

koops76:
@J0WI Done.

PR #11730, opened 2017-08-07
Update makecrx.sh

koops76:
Not sure why do we compute the SHA1 checksum of the unsigned ZIP file, since the signed CRX will have a different checksum.


koops76:
ping @Hainish.

Issue #2311, opened 2015-07-23
Add ftp: to 'Block all HTTP requests'


avlidienbrunn:
Hi!

I'm not 100% sure that I've understood the original intention of the 'Block all HTTP requests' feature, but I personally enable it when I want to make sure that I don't accidentally receive anything over an unencrypted connection.

If that is the intended reason, then I would like to suggest that you add the "ftp:" protocol handler to the blacklist as well, since browsers are capable of serving unencrypted content over FTP too.

jsha:
You've got the right idea about the intent. Probably this should be a whitelist containing just HTTPS rather than a blacklist. Thanks for the bug report!

avlidienbrunn:
Great. When you do, keep in mind that "filesystem:" URIs also trigger the onBeforeRequest event (for some reason) in Chrome. For example, this will trigger the onBeforeRequest on a "filesystem:" URI without ever sending a request and thus shouldn't be blocked:

```
function onInitFs(fs) {
	fs.root.getFile('ex.png', {create: true}, function(fileEntry) {
		fileEntry.createWriter(function(fileWriter) {
			// Create a new Blob and write it to ex.png.
			var blob = new Blob([':)'], {type: 'image/png'});
			fileWriter.write(blob);
			elem=document.createElement("img");
			//Create a filesystem: URL and append it to the current document
			elem.src=fileEntry.toURL();
			document.documentElement.appendChild(elem);
		});
	});
}
window.requestFileSystem  = window.requestFileSystem || window.webkitRequestFileSystem;
window.requestFileSystem(window.TEMPORARY, 1024*1024 /*1MB*/, onInitFs);
```

9Yg1rxeSeha90ZU1:
This is still present and "Block all unencrypted requests" does NOT prevent unencrypted FTP (!)

A proper fix would, at minimum, block the ftp:// protocol scheme when "Block all unencrypted requests" is enabled.

koops76:
Fixed by #11738.

Hainish:
Fixed by #15813 (though Firefox is still broken, I've filed a bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=1471065)

PR #11738, opened 2017-08-07
Block FTP requests if "Block all unencrypted requests" is on

koops76:


J0WI:
This should fix #2311

J0WI:
@Hainish can we ship this for the next release? I think this is a quite important feature.

koops76:
@Hainish Just test it thoroughly, I don't want any bugs to end up in the release because of me. Remember the #10852? I didn't test it at all, same with this one. I rely on you testing it.

koops76:
@J0WI Add this to your kanban I guess.

chromium/background.js
```diff
@@ -227,14 +227,11 @@ function onBeforeRequest(details) {
   const uri = new URL(details.url);
 
   // Should the request be canceled?
-  var shouldCancel = (
-    httpNowhereOn &&
-    uri.protocol === 'http:' &&
-    !/\.onion$/.test(uri.hostname) &&
-    !/^localhost$/.test(uri.hostname) &&
-    !/^127(\.[0-9]{1,3}){3}$/.test(uri.hostname) &&
-    !/^0\.0\.0\.0$/.test(uri.hostname)
```
Hainish: Why remove this?

koops76: @Hainish Look down.

koops76: @Hainish I thought you mean I removed the entire statement. Sorry. `0.0.0.0` is not a valid IPv4 address. Not sure why we have it here.

koops76: I also changed the regexp for IP addresses just now.

Hainish: It *is* a valid IP address.  See https://en.wikipedia.org/wiki/0.0.0.0

> In the Internet Protocol Version 4, the address 0.0.0.0 is a non-routable meta-address used to designate an invalid, unknown or non-applicable target. To give a special meaning to an otherwise invalid piece of data is an application of in-band signaling.
> 
> In the context of servers, 0.0.0.0 means "all IPv4 addresses on the local machine". If a host has two IP addresses, 192.168.1.1 and 10.1.2.1, and a server running on the host listens on 0.0.0.0, it will be reachable at both of those IPs.

For instance, setting up a simple local webserver with python is by default accessible over `0.0.0.0`:

```shell
user@https-everywhere ~/blah $ python -m SimpleHTTPServer &
[1] 16779
user@https-everywhere ~/blah $ Serving HTTP on 0.0.0.0 port 8000 ...

user@https-everywhere ~/blah $ curl 0.0.0.0:8000
127.0.0.1 - - [15/Aug/2017 11:27:48] "GET / HTTP/1.1" 200 -
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN"><html>
<title>Directory listing for /</title>
<body>
<h2>Directory listing for /</h2>
<hr>
<ul>
<li><a href="default.rulesets.gz.base64">default.rulesets.gz.base64</a>
<li><a href="rulesets-signature.sha256.base64">rulesets-signature.sha256.base64</a>
<li><a href="rulesets-timestamp">rulesets-timestamp</a>
</ul>
<hr>
</body>
</html>
user@https-everywhere ~/blah $
```

koops76: @Hainish Thanks for useful information!

koops76: @Hainish Done.

koops76: @Hainish Chrome returns ERR_ADDRESS_INVALID if you attempt to visit http://0.0.0.0/ though.

chromium/background.js
```diff
@@ -568,7 +566,7 @@ function onBeforeRedirect(details) {
 
 // Registers the handler for requests
 // See: https://github.com/EFForg/https-everywhere/issues/10039
-wr.onBeforeRequest.addListener(onBeforeRequest, {urls: ["*://*/*"]}, ["blocking"]);
+wr.onBeforeRequest.addListener(onBeforeRequest, {urls: ["http://*/*", "https://*/*", "ftp://*/*"]}, ["blocking"]);
```
Hainish: Why change this back to `http://*/*` and `https://*/*` rather than keep `*://*/*`?  Is this just for readability?

koops76: Indeed.

Hainish:
This seems not to be blocking unencrypted `ftp` requests when I access e.g. ftp://mirror.fra10.de.leaseweb.net/centos/

Hainish:
For either Chromium or Firefox.  It also doesn't work when I switch the match pattern to `<all_urls>`

Hainish:
This is probably a bug in the WebExtensions APIs.  This would be a good feature to have, let's open bugs the respective implementations.

koops76:
@Hainish I found what caused the problem: sloppy code in `onBeforeRequest` function. Expect fix soon. WebExtensions is not to blame.

koops76:
@Hainish Done. Test and if there are no problems, merge.

koops76:
@Hainish Can you figure out why the tests fail? My testing can't find any issues.

Hainish:
Re-running seems to have resolved the issue

Hainish:
Testing again with ftp://mirror.fra10.de.leaseweb.net/centos/ I see the same behavior - this request is not blocked.  It *is* a problem with WebExtensions.  Add the following lines to the top of `onBeforeRequest`:

```js
console.log('GOT HERE');
console.log(details);
```

With ftp URLs, this is never triggered.

koops76:
@Hainish Try with an extension that cancels all requests.

Hainish:
The hook itself doesn't work.

koops76:
@Hainish File a bug on Chromium and Firefox.

koops76:
Blocked by WebExtensions flaw.

PR #11746, opened 2017-08-08
Create michelonfray.com.xml

denisj:


src/chrome/content/rules/michelonfray.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="michelonfray.com">
+  <target host="michelonfray.com" />
```
J0WI: + `www.michelonfray.com`

denisj: Should i put it even if the site redirect to michelonfray.com?

J0WI: Yes. We also want to secure this redirect.

J0WI:
Thanks!

PR #11756, opened 2017-08-08
Create GinjFo.com.xml

6u621jpq6ybdw:


src/chrome/content/GinjFo.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Invalid Certificate:
+		ginjfo.com
+		openx.ginjfo.com
```
J0WI: Just an incomplete cert chain.

6u621jpq6ybdw:
@J0WI Added.

src/chrome/content/GinjFo.com.xml
J0WI: The comment was right for `^`

koops76:
#11795

PR #11757, opened 2017-08-08
Why not contacting every site first with https instead of making single rulesets?

Raketenwilli:
As you can see in the title this is more a question than a pull request, I hope that's okay to you. If not, just delete it.
So my question is: Why do you make single rulesets instead of asking the domain hoster directly for a ssl certificate and if there is no one, access the site per http? I know this slows down the access time a bit but you even have a much greater chance to get an https connection.


koops76:
Close this please, there's no need merging commits from `master` into `webextensions`.

Raketenwilli:
Just wanted to get an answer...

koops76:
@Raketenwilli It will break lots of websites.

PR #11764, opened 2017-08-08
Create LearnShell.org.xml

opsteroids:


J0WI:
Would you also like to add all other learn* hosts from this project?

6u621jpq6ybdw:
@J0WI Ok, and how do I name the ruleset? Or you mean I create different rulesets for each?

6u621jpq6ybdw:
@J0WI Added the second approach.

J0WI:
Please choose one rule with a complete reference to other rulesets and link just to this file. Otherwise we have to update all the rules if we need to change something.

6u621jpq6ybdw:
@J0WI Added.

J0WI:
Thanks!

PR #11771, opened 2017-08-09
Create JUMIA.ma.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


src/chrome/content/rules/JUMIA.ma.xml
```diff
@@ -0,0 +1,21 @@
+<ruleset name="JUMIA.ma">
+	<target host="jumia.ma" />
+	<target host="*.jumia.ma" />
```
J0WI: Please avoid wildcards if possible.

6u621jpq6ybdw:
@J0WI There are way too many subdomains, and I couldn't fiind a single sub that isn't https, could be wrong through (used Google for subdomain enumeration)

PR #11510, opened 2017-07-29
[US-government] Delete ruleset

Bisaloo:
The dead links are caused by `hsts-prune`.

PR #11783, opened 2017-08-09
[House.gov] Add targets

Bisaloo:
@jeremyn 

#11510

I don't like wildcards any more than you do but I don't think we have a choice here.

### DIFF
--- a/src/chrome/content/rules/House.gov.xml
+++ b/src/chrome/content/rules/House.gov.xml
@@ -29,8 +29,8 @@
 	
 		<!-- Elected representative subdomains -->
 		<test url="http://abraham.house.gov/" />
-		<test url="http://babin.house.gov" />
-		<test url="http://cardenas.house.gov" />
+		<test url="http://babin.house.gov/" />
+		<test url="http://cardenas.house.gov/" />
 		<test url="http://davidson.house.gov/" />
 		<test url="http://ellison.house.gov/" />
 	
--- a/src/chrome/content/rules/House.gov.xml
+++ b/src/chrome/content/rules/House.gov.xml
@@ -1,8 +1,44 @@
+<!--
+	Different content http/https:
+		clerk.house.gov
+
+	Time out:
+		docs.house.gov
+
+	Invalid certificate:
+		kids.clerk.house.gov
+		history.house.gov
+
+	Each congress elected gets their own subdomain.
+
+-->
 <ruleset name="House.gov">
 
-	<target host="democrats-intelligence.house.gov" />
-	<target host="intelligence.house.gov" />
-	<target host="democrats.intelligence.house.gov" />
+	<target host="house.gov" />
+	<target host="*.house.gov" />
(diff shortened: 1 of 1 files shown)


afarrow:
I don't know if you've seen this or if it would be of any use to you but I made a pull request a while back for this domain: #9706

Bisaloo:
No, I did not see that. Thank you!

Do you want to continue your work on this ruleset? Or should I integrate your changes to this PR?

afarrow:
You can just integrate whatever changes you feel are necessary into this PR

jeremyn:
I'm firmly against a wildcard for `house.gov` because the [many issues](https://github.com/EFForg/https-everywhere/search?q=senate.gov&type=Issues&utf8=%E2%9C%93) raised against weird behavior at `senate.gov`.

I would prefer if this PR just made the very minor change of adding `(partial)` to the `name` that I asked for in https://github.com/EFForg/https-everywhere/pull/11510, with @afarrow 's (thorough) changes left in their own PR.

Bisaloo:
Ok, @afarrow, the subdomains I added as test url should be safe to add as targets.

The full list of representatives can be found in the source code of https://www.house.gov/ starting line 400.

For an unclear reason, 4 of them are not linked to the `https` version of their page:
- http://capuano.house.gov
- http://eshoo.house.gov/
- http://matsui.house.gov
- http://rokita.house.gov/

afarrow:
Ok I created a new PR combining our efforts: #11808 

Good find on the list of the subdomains in the source. Saved me a lot of effort. I included some notes (including that one) at the top of the file for future maintainers.

Those 4 subdomains that had their http version linked looked fine over https to me so I included them.

@jeremyn I went ahead and added `(partial)` to the name in my PR since that's such a minor change.

jeremyn:
@afarrow A primary reason to avoid wildcards and instead list individual targets is because listing a target implies that someone has looked at the domain and decided it works with HTTPS. Creating a list of 400+ targets using an automated process defeats that purpose. For example, https://capuano.house.gov has some mixed content blocking issues that may just be stylistic but we should look into it. Some other domains that I checked, don't have mixed content blocking.

I'm also uninterested in reviewing a ruleset where I have to check 400+ domains before I can approve it. Maybe one of other reviewers will do that but it seems unlikely to me.

I have no solution to this problem, I'm just letting you know that I think your PR https://github.com/EFForg/https-everywhere/pull/11808 is going to have a tough time getting reviewed.

afarrow:
Ok, well then what about taking the approach I took in my first PR which is cover the more general subdomains and ignore the ones for the representatives. If I take out the representatives' subdomains, that yields 66 subdomains to check. That's still a lot but a lot more manageable. We could even do them 10 at a time if you would prefer. 

I totally understand what you're saying about not wanting to review >400 subdomains. There's got to be a happy medium we can come to between adding all the subdomains and ignoring this domain.

PR #11788, opened 2017-08-09
[Spreadshirt] Simplify ruleset and fix mismatch

Bisaloo:


src/chrome/content/rules/Spreadshirt.xml
```diff
@@ -78,12 +78,12 @@
 	<test url="http://www.spreadshirt.se/" />
 	<test url="http://wikileaks.spreadshirt.se/" />
 
-	<securecookie host="\w+\.spreadshirt\.(at|be|ch|co\.uk|com|com\.au|de|dk|es|fi|fr|ie|it|net|nl|no|pl|se)$" name=".+" />
-
-	<exclusion pattern="^http://(www\.)?spreadshirt\.(at|be|ch|co\.uk|com|com\.au|de|dk|es|fi|fr|ie|it|net|nl|no|pl|se)/" />
+	<securecookie host=".+" name=".+" />
 
 	<rule from="^http://(\w+)\.spreadshirt\.(at|be|ch|co\.uk|com|com\.au|de|dk|es|fi|fr|ie|it|net|nl|no|pl|se)/"
```
koops76: This rule is unneeded since there is already a catch-all rule doing the same after it.

src/chrome/content/rules/Spreadshirt.xml
```diff
@@ -78,12 +78,12 @@
 	<test url="http://www.spreadshirt.se/" />
```
koops76: Maybe you'll try to de-wildcard this ruleset?

Bisaloo:
@koops76, I'm still working on this PR. I am simply waiting for travis between each changes.

src/chrome/content/rules/Spreadshirt.xml
Bisaloo: I am thinking about it but those two rules are not strictly the same. The last trivial rule also rewrites subdomains deeper than first-level.

I need to check if this could be an issue first.

koops76: Even if the first rule doesn't catch the request, the second one will always do. The rules are functionally the same. I think that the first rule is redundant.

Bisaloo:
@J0WI, I'm done here so if you want to merge, go ahead.

PR #11789, opened 2017-08-09
Firefox: Removing SDK extension tests and generalizing selenium tests…

Hainish:
… to support FF


test/firefox/test_profile_skeleton/prefs.js
```diff
@@ -1,20 +1,9 @@
-// Submit SSL Observatory reports.
-user_pref("extensions.https_everywhere._observatory.enabled", true);
-// Don't show any popups that might get in the way of testing.
-user_pref("extensions.https_everywhere._observatory.popup_shown", true);
-user_pref("extensions.https_everywhere.toolbar_hint_shown", true);
-// Show all logs.
-user_pref("extensions.https_everywhere.LogLevel", 0);
-user_pref("extensions.https_everywhere.log_to_stdout", true);
 // Make it quicker to make manual config changes.
```
jsha: Why is this file only partially deleted?

test/script.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python2.7
+#
+# Run Chromium tests for HTTPS Everywhere
```
jsha: s/Chromium/Selenium/

test/script.py
```diff
@@ -0,0 +1,98 @@
...
+
+if sys.argv[1] == "Firefox":
+    fp = webdriver.FirefoxProfile(sys.argv[2])
+    try:
+        if len(sys.argv) == 4:
+            binary = FirefoxBinary(sys.argv[3])
+            driver = webdriver.Firefox(fp, firefox_binary=binary, log_path=os.devnull)
+        else:
+            driver = webdriver.Firefox(fp, log_path=os.devnull)
+        # Allow the extension time to load
```
jsha: This comment should be right above the time.sleep. But it's a little alarming: HTTPS Everywhere should ideally redirect all relevant requests. If requests that happen early in the browser load skip the redirects, that's a bug. It looks like Chrome doesn't have an equivalent sleep. Do you know why this is necessary in Firefox?

test/script.py
```diff
@@ -0,0 +1,98 @@
...
+
+print ''
+
+driver.get('http://freerangekitten.com')
+
+test_failed = False
+if driver.current_url.startswith('https'):
+    print bcolors.OKGREEN + sys.argv[1] + ": HTTP to HTTPS redirection successful" + bcolors.ENDC
+elif driver.current_url.startswith('http'):
+    print bcolors.FAIL + sys.argv[1] + "Chromium: HTTP to HTTPS redirection failed" + bcolors.ENDC
```
jsha: s/Chromium: //

test/firefox/test_profile_skeleton/prefs.js
Hainish: It's been a while since this commit was made, I'm honestly not sure why.

Hainish: Ah, these are the prefs that are only relevant for the XPCOM extension.  The other prefs are necessary for debugging, installing the extension without a valid signature, etc.

Hainish: These changes belong in https://github.com/EFForg/https-everywhere/pull/11791, I'll move them over.

PR #11791, opened 2017-08-09
Firefox webextensions core

Hainish:
Core changes to WebExtension code to make it work with Firefox.  Reference the `done` column of https://github.com/EFForg/https-everywhere/projects/1

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -61,12 +61,16 @@ chrome.storage.onChanged.addListener(function(changes, areaName) {
     }
   }
 });
-chrome.tabs.onActivated.addListener(function() {
-  updateState();
-});
-chrome.windows.onFocusChanged.addListener(function() {
-  updateState();
-});
+if (chrome.tabs) {
+  chrome.tabs.onActivated.addListener(function() {
+    updateState();
+  });
+}
+if (chrome.windows) {
+  chrome.windows.onFocusChanged.addListener(function() {
+    updateState();
+  });
+}
 chrome.webNavigation.onCompleted.addListener(function() {
   updateState();
 });
@@ -106,6 +110,9 @@ loadStoredUserRules();
  * disabled: extension is disabled from the popup menu.
  */
 var updateState = function() {
+  if (!chrome.tabs) {
+    return;
+  }
   chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
     if (!tabs || tabs.length === 0) {
       return;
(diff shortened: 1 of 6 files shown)


Hainish:
@jsha: reverted 7dbab77ef08d86d8ca0219fa47d7fc54312d0ed9

Hainish:
(As with #11792, just let me know when this looks good but don't merge.  I'll rebase to remove the original commit as well as the revert).

chromium/background.js
```diff
@@ -618,3 +627,38 @@ chrome.runtime.onConnect.addListener(function (port) {
     });
   }
 });
+
+// This is necessary for communication with the popup in Firefox Private
+// Browsing Mode, see https://bugzilla.mozilla.org/show_bug.cgi?id=1329304
+chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
```
jsha: Can you confirm that this will only receive messages from background.js, not from content scripts or other extensions?

Hainish: That's right.  There is a separate IPC API for inter-extension, `chrome.runtime.onMessageExternal`.  See https://developer.chrome.com/extensions/messaging#external

chromium/popup.js
```diff
@@ -65,11 +72,17 @@ function appendRuleLineToListDiv(ruleset, list_div) {
       break;
     }
   }
-  var xhr = new XMLHttpRequest();
-  try {
-    xhr.open("GET", favicon.src, true);
-    label.appendChild(favicon);
-  } catch (e) {}
+  if (favicon.src == "chrome://favicon/") {
+    favicon.src += "https://" + hostnames[ruleset.name];
```
koops76: @Hainish @jsha WTF was this XMLHTTPRequest stuff?

chromium/popup.js
```diff
@@ -65,11 +72,17 @@ function appendRuleLineToListDiv(ruleset, list_div) {
...
+  }
+
+  if (navigator.userAgent.match("Chrome")) {
+    var xhr = new XMLHttpRequest();
+    try {
+      xhr.open("GET", favicon.src, true);
```
koops76: Remove this XMLHTTPRequest stuff. I'm not sure what it was supposed to do at all.

chromium/storage.js
```diff
@@ -1,6 +1,10 @@
 var storage = chrome.storage.local;
```
koops76: This entire file can be deleted and replaced with simple line in each file `storage` is used.

`const storage = chrome.storage.sync || chrome.storage.local`

Hainish: Within for example the tor browser, `chrome.storage.sync` may exist but throw an error.  That's the reason for the `lastError` check.

koops76: File a bug for Tor Browser then.

PR #11792, opened 2017-08-09
Embedded webextension isolated

Hainish:


Hainish:
A failure to the `firefox` tests is expected, resolution depends on https://github.com/EFForg/https-everywhere/pull/11789.  This is rebased on top of https://github.com/EFForg/https-everywhere/pull/11791

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+/**
+ * Import extension settings (custom rulesets, ruleset toggles, globals) from an object
+ * @param settings the settings object
+ */
+async function import_settings(settings){
+  if(settings.changed){
+    // Load custom rulesets and add to storage
+    await new Promise(resolve => {
+      for(let ruleset of settings.custom_rulesets){
+        all_rules.addFromXml((new DOMParser()).parseFromString(ruleset, 'text/xml'));
```
jsha: Please factor out a function load_legacy_custom_rulesets, so that you can call it here after you do `storage.set`.

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+          resolve(false);
+        })
+      })(ruleset_name, settings.rule_toggle[ruleset_name]));
+    }
+    await Promise.all(rule_toggle_promises);
+
+    all_rules.ruleCache = new Map();
+
+    // Set/store globals
+    storage.set({'httpNowhere': settings.prefs.http_nowhere_enabled});
```
jsha: Is there a reason you provide a callback when calling set above, but not here? It seems like we should use the callback on both, to detect success or failure.

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+              }
+            }
+          }
+          resolve(false);
+        })
+      })(ruleset_name, settings.rule_toggle[ruleset_name]));
+    }
+    await Promise.all(rule_toggle_promises);
+
+    all_rules.ruleCache = new Map();
```
jsha: Can you add a comment explaining why this is needed?

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+    });
+
+    // Load all the ruleset toggles into memory and store
+    let rule_toggle_promises = [];
+    for(let ruleset_name in settings.rule_toggle){
+      localStorage[ruleset_name] = settings.rule_toggle[ruleset_name];
+
+      rule_toggle_promises.push((ruleset_name, ruleset_active => {
+        return new Promise(resolve => {
+          for(let host in all_rules.targets){
```
jsha: A couple of comments here:

 - This doesn't seem to do any async work, so why is it wrapped in a Promise?
 - This seems to re-do the work of marking rulesets active or inactive that should happen at startup. Similar to the loading of legacy rulesets I mentioned above, can you just store the settings, and then re-load the rulesets? In fact, since loading the rulesets should include loading legacy_custom_rulesets, you can just do that and don't have to separately invoke the loader for legacy_custom_rulesets.

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+      })(ruleset_name, settings.rule_toggle[ruleset_name]));
+    }
+    await Promise.all(rule_toggle_promises);
+
+    all_rules.ruleCache = new Map();
+
+    // Set/store globals
+    storage.set({'httpNowhere': settings.prefs.http_nowhere_enabled});
+    storage.set({'showCounter': settings.prefs.show_counter});
+    isExtensionEnabled = settings.prefs.global_enabled;
```
jsha: This setting doesn't seem to get stored?

makexpi.sh
```diff
@@ -101,7 +100,25 @@ fi
 # Prepare packages suitable for uploading to EFF and AMO, respectively.
 [ -d pkg ] || mkdir pkg
 rsync -aL --delete --delete-excluded --exclude /chrome/content/rules src/ pkg/xpi-eff
-cp -a translations/* pkg/xpi-eff/chrome/locale/
+
+# The next lines are for embedded WebExtensions, and can be deleted after the full transition to WebExtensions
```
jsha: You should probably include a footer indicating the end of the lines that can be deleted after the full transition.

src/bootstrap.js
```diff
@@ -0,0 +1,111 @@
...
+}
+
+function startup({webExtension}) {
+  webExtension.startup().then(api => {
+    const {browser} = api;
+    browser.runtime.onMessage.addListener((msg, sender, sendReply) => {
+      if (msg == "import-legacy-data") {
+        let globals = get_prefs("extensions.https_everywhere.");
+        let prefs_array = globals.getChildList("", {});
+        if(~prefs_array.indexOf("webextensions-migrated") ? !globals.getBoolPref("webextensions-migrated") : true){
```
jsha: This if statement is overly complicated. In particular, `:?` operators are ideally avoided because they are hard to read and error prone. Embedding a `:?` inside an if makes that especially so. :-) Can you break it out for clarity? Also, the ~ operator is a new idiom in this codebase. I think it would be better not to introduce it in such a large change without strong motivation.

src/bootstrap.js
```diff
@@ -0,0 +1,111 @@
...
+          let rules = get_prefs("extensions.https_everywhere.rule_toggle.");
+
+          let changed = false;
+
+          let prefs = {
+            http_nowhere_enabled:  ~prefs_array.indexOf("http_nowhere.enabled") ? globals.getBoolPref("http_nowhere.enabled") : false,
```
jsha: This pattern of "check if pref exists, return default value if not" should be a small helper function.

src/bootstrap.js
```diff
@@ -0,0 +1,111 @@
...
+
+function startup({webExtension}) {
+  webExtension.startup().then(api => {
+    const {browser} = api;
+    browser.runtime.onMessage.addListener((msg, sender, sendReply) => {
+      if (msg == "import-legacy-data") {
+        let globals = get_prefs("extensions.https_everywhere.");
+        let prefs_array = globals.getChildList("", {});
+        if(~prefs_array.indexOf("webextensions-migrated") ? !globals.getBoolPref("webextensions-migrated") : true){
+          let rules = get_prefs("extensions.https_everywhere.rule_toggle.");
```
jsha: Move `let rules` and `let changed` down to right before they are used.

koops76:
@Hainish Also, can you already code this in StandardJS style or the style of your choice (I recommend Standard)?

chromium/background.js
```diff
@@ -674,5 +683,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
     updateState();
   } else if (message.type == "remove_rule") {
     removeRule(message.object);
+  } else if (message.type == "import_settings") {
```
jsha: Can you add a comment explaining when this is used? I.e., importing from the Options page rather than automatically from the parent extension?

chromium/background.js
Hainish: This task can take quite a while, since it's iterating through all ruleset targets and checking all rules they contain until finding a match.  Creating a new promise for each of these isn't to resolve callbacks, but to start asynchronous program flow for each of these potentially many toggles.

Hainish: Since this is javascript and not actually dedicating separate cores to the problem, it does not necessarily give us any efficiency gain to split this up into multiple iterations of the event loop.  I can refactor this to be synchronous.

chromium/background.js
Hainish: It's true, the extension enabled toggle is not stored (and has never been).  We should fix that but I figured it was outside of the scope of this PR.

src/bootstrap.js
Hainish: I agree for `rules`, but for `changed` it seems this is a var which is featured prominently throughout this scope.  I'd argue that it's appropriate to keep it at the top.

chromium/background.js
jsha: Oh, interesting! I did not know that. It seems weird to apply this toggle only on first run after an upgrade, though. If storing it doesn't work, I'd suggest just removing this line.

src/bootstrap.js
jsha: Sounds reasonable.

chromium/background.js
Hainish: Added.

chromium/background.js
```diff
@@ -674,5 +682,56 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+      storage.set({"legacy_custom_rulesets": settings.custom_rulesets}, resolve);
+    });
+    load_legacy_custom_rulesets(settings.custom_rulesets);
+
+    // Load all the ruleset toggles into memory and store
+    let rule_toggle_promises = [];
+    for(let ruleset_name in settings.rule_toggle){
+      localStorage[ruleset_name] = settings.rule_toggle[ruleset_name];
+
+      (ruleset_name, ruleset_active => {
```
jsha: Did you see my earlier comment about just storing the data and reloading? E.g. `all_rules = new RuleSets(localStorage)`

chromium/background.js
```diff
@@ -674,5 +682,56 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+            if(ruleset.name == ruleset_name){
+              ruleset.active = ruleset_active;
+              return;
+            }
+          }
+        }
+      })(ruleset_name, settings.rule_toggle[ruleset_name]);
+    }
+
+    // This is necessary in order to clear the cache for any previous ruleset lookups
```
jsha: If you do the reload, this won't be necessary anymore.

chromium/background.js
```diff
@@ -674,5 +682,56 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+    all_rules.ruleCache = new Map();
+
+    // Set/store globals
+    await new Promise(resolve => {
+      storage.set({'httpNowhere': settings.prefs.http_nowhere_enabled}, resolve);
+    });
+    await new Promise(resolve => {
+      storage.set({'showCounter': settings.prefs.show_counter}, resolve);
+    });
+    isExtensionEnabled = settings.prefs.global_enabled;
```
jsha: Did you see my comment about removing this since it will only work once, and is therefore inconsistent/confusing?

chromium/background.js
Hainish: Ah! Didn't catch that.  Fixing now.

Hainish:
Ready for review again.

Hainish:
Hold on, for some reason the default rulesets are absent when the second call of `new RuleSets(localStorage)` is made.

koops76:
Do you mutate the argument in the `RuleSets` constructor?

Hainish:
@koops76 not in a way that would cause targets to be lost

koops76:
Try making a copy of everything in `localStorage` and passing that copy into the `RuleSets` constructor.

chromium/background.js
```diff
@@ -674,5 +674,44 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+
+    // Load all the ruleset toggles into memory and store
+    let rule_toggle_promises = [];
+    for(let ruleset_name in settings.rule_toggle){
+      localStorage[ruleset_name] = settings.rule_toggle[ruleset_name];
+    }
+    all_rules = new RuleSets(localStorage);
+
+    // Set/store globals
+    await new Promise(resolve => {
```
koops76: DRY

chromium/options.css
```diff
@@ -0,0 +1,17 @@
+.section-header{
```
koops76: Add a space between the `{` and the selector.

chromium/options.html
```diff
@@ -0,0 +1,22 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <link href="options.css" rel="stylesheet" type="text/css"/>
+    <script src="options.js"></script>
+    <script src="translation.js"></script>
+    <script src="send-message.js"></script>
+  </head>
+  <body>
+    <div id='import-confirmed' i18n="options_imported"></div>
```
koops76: Use `data-i18n` for valid HTML5.

chromium/options.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <link href="options.css" rel="stylesheet" type="text/css"/>
+    <script src="options.js"></script>
```
koops76: Move the scripts to the end of `body` tag.

chromium/options.js
```diff
@@ -0,0 +1,28 @@
+document.addEventListener("DOMContentLoaded", () => {
+
+  let json_data;
+  let import_button = document.querySelector("#import");
```
koops76: `const`?

chromium/options.js
```diff
@@ -0,0 +1,28 @@
...
+
+  function import_json(e) {
+    e.preventDefault();
+
+    let settings = JSON.parse(json_data);
```
koops76: `const`?

chromium/options.js
```diff
@@ -0,0 +1,28 @@
...
+    const reader = new FileReader();
+    reader.onload = event => {
+      json_data = event.target.result;
+      import_button.disabled = false;
+    };
+
+    reader.readAsText(file);
+  });
+
+  document.querySelector("form").addEventListener("submit", import_json);
```
koops76: Maybe find a `form` by `id` instead of simply taking the first one in the document?

chromium/options.js
```diff
@@ -0,0 +1,28 @@
...
+    e.preventDefault();
+
+    let settings = JSON.parse(json_data);
+    sendMessage("import_settings", settings, resp => {
+      document.querySelector("#import-confirmed").style.display = "block";
+      document.querySelector("form").style.display = "none";
+    });
+  }
+
+  document.querySelector("#import-settings").addEventListener("change", event => {
```
koops76: `document.getElementById`?

chromium/options.js
```diff
@@ -0,0 +1,28 @@
...
+
+  let json_data;
+  let import_button = document.querySelector("#import");
+
+  function import_json(e) {
+    e.preventDefault();
+
+    let settings = JSON.parse(json_data);
+    sendMessage("import_settings", settings, resp => {
+      document.querySelector("#import-confirmed").style.display = "block";
```
koops76: `document.getElementById`?

chromium/options.js
```diff
@@ -0,0 +1,28 @@
...
+      document.querySelector("form").style.display = "none";
```
koops76: Maybe find a `form` by `id` instead of simply taking the first one in the document?

chromium/popup.html
```diff
@@ -9,6 +9,8 @@
 
     <link href="popup.css" rel="stylesheet" type="text/css"/>
     <script src="popup.js"></script>
+    <script src="translation.js"></script>
```
koops76: Move the scripts to the end of `body` tag.

chromium/popup.js
```diff
@@ -197,16 +197,6 @@ document.addEventListener("DOMContentLoaded", function () {
       httpNowhereCheckbox.setAttribute('checked', '');
```
koops76: ``httpNowhereCheckbox.checked = true`?

chromium/popup.js
```diff
@@ -197,16 +197,6 @@ document.addEventListener("DOMContentLoaded", function () {
...
     }
   });
-
-  // auto-translate all elements with i18n attributes
-  var elem = document.querySelectorAll("[i18n]");
```
koops76: Use `document.querySelectorAll("[data-i18n]")` instead.

chromium/popup.js
```diff
@@ -197,16 +197,6 @@ document.addEventListener("DOMContentLoaded", function () {
...
-  for (let el of elem) {
-    el.innerHTML = chrome.i18n.getMessage(el.getAttribute("i18n"));
```
koops76: Use `el.getAttribute("data-i18n")` or `el.dataset.i18n`.

Hainish:
I figured it out....

chromium/popup.js
```diff
@@ -197,16 +197,6 @@ document.addEventListener("DOMContentLoaded", function () {
...
-  }
-
-  // other translations
-  e("aboutTitle").setAttribute("title", chrome.i18n.getMessage("about_title"));
```
koops76: `e("aboutTitle").title = chrome.i18n.getMessage("about_title")`

chromium/rules.js
```diff
@@ -185,6 +185,14 @@ function RuleSets(ruleActiveStates) {
 
   // A hash of rule name -> active status (true/false).
   this.ruleActiveStates = ruleActiveStates;
+
+  // Load legacy_custom_rulesets
+  const that = this;
```
koops76: ???

chromium/rules.js
```diff
@@ -185,6 +185,14 @@ function RuleSets(ruleActiveStates) {
...
+  storage.get({legacy_custom_rulesets: []}, item => {
+    for(let legacy_custom_ruleset of item.legacy_custom_rulesets){
```
koops76: `const`?

chromium/translation.js
```diff
@@ -0,0 +1,11 @@
+document.addEventListener("DOMContentLoaded", function () {
+  // auto-translate all elements with i18n attributes
+  var elem = document.querySelectorAll("[i18n]");
+  for (let el of elem) {
+    el.innerHTML = chrome.i18n.getMessage(el.getAttribute("i18n"));
+  }
+
+  // other translations
+  e("aboutTitle").setAttribute("title", chrome.i18n.getMessage("about_title"));
```
koops76: `e("aboutTitle").title = chrome.i18n.getMessage("about_title")`

chromium/translation.js
```diff
@@ -0,0 +1,11 @@
+document.addEventListener("DOMContentLoaded", function () {
+  // auto-translate all elements with i18n attributes
+  var elem = document.querySelectorAll("[i18n]");
+  for (let el of elem) {
+    el.innerHTML = chrome.i18n.getMessage(el.getAttribute("i18n"));
```
koops76: `el.innerText = chrome.i18n.getMessage(el.getAttribute("data-i18n"));`

chromium/translation.js
```diff
@@ -0,0 +1,11 @@
+document.addEventListener("DOMContentLoaded", function () {
+  // auto-translate all elements with i18n attributes
+  var elem = document.querySelectorAll("[i18n]");
```
koops76: `const elements = document.querySelectorAll("[data-i18n]");`

chromium/translation.js
```diff
@@ -0,0 +1,11 @@
...
+  for (let el of elem) {
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
+"use strict";
+
+const CI = Components.interfaces;
+const CC = Components.classes;
+
+function get_prefs(branch_name){
+  let o_prefs = CC["@mozilla.org/preferences-service;1"]
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+                      .getService(CI.nsIPrefService);
+  let o_branch = o_prefs.getBranch(branch_name);
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  return o_branch;
+}
+
+function get_custom_rulesets_array(){
+  var loc = "ProfD";  // profile directory
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  var dir =
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+    .getService(CI.nsIProperties)
+    .get(loc, CI.nsILocalFile)
+    .clone();
+  dir.append("HTTPSEverywhereUserRules");
+  // Check for existence, if not, create.
+  if (!dir.exists()) {
+    return [];
+  }
+
+  var entries = dir.directoryEntries;
```
koops76: `const`/`let`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  var files = [];
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  while(entries.hasMoreElements()) {
```
koops76: `while (`

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+    var entry = entries.getNext();
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+    entry.QueryInterface(CI.nsIFile);
+    files.push(entry);
+  }
+
+  let file_data = []
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  for(let file of files){
+    var data = "";
```
koops76: `let`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+    var fstream = CC["@mozilla.org/network/file-input-stream;1"]
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+        .createInstance(CI.nsIFileInputStream);
+    var sstream = CC["@mozilla.org/scriptableinputstream;1"]
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+        .createInstance(CI.nsIScriptableInputStream);
+    fstream.init(file, -1, 0, 0);
+    sstream.init(fstream);
+
+    var str = sstream.read(4096);
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+
+  return file_data;
+}
+
+function startup({webExtension}) {
+  webExtension.startup().then(api => {
+    const {browser} = api;
+    browser.runtime.onMessage.addListener((msg, sender, sendReply) => {
+      if (msg == "import-legacy-data") {
+        let globals = get_prefs("extensions.https_everywhere.");
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+        if(!pref_retrieve_bool_with_default(globals, "webextension-migrated", false)){
+          let changed = false;
+
+          let prefs = {
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+          };
+
+          if(prefs.http_nowhere_enabled == true ||
+            prefs.global_enabled == false ||
+            prefs.show_counter == true
+          ){
+            changed = true;
+          }
+
+          let rule_toggle = {}
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+          let rules = get_prefs("extensions.https_everywhere.rule_toggle.");
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+          for(let rule_toggle_key of rules.getChildList("", {})){
```
koops76: `const`?

src/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+            rule_toggle[rule_toggle_key]  = rules.getBoolPref(rule_toggle_key);
+            changed = true;
+          }
+
+          let custom_rulesets = get_custom_rulesets_array();
```
koops76: `const`?

Hainish:
@jsha There's a problem with loading the XML for legacy rulesets in the constructor.  The rules themselves aren't stored in `localStorage` and are only loaded on what's currently line 31 of this branch in `background.js`.  So in order to just do `new RuleSets(localStorage)` any time we want to reload rules, we'd have to move that into the constructor as well (along with the lookup for mixed content rules).  I'm uncomfortable with this, because it makes the RuleSets object less agnostic, and we will have to rip this out anyway when we merge the `sign-rulesets` branch, which loads rules and stores them via the network.

I therefore propose that we keep the previous workflow for loading rulesets.

jsha:
That sounds fine for now, although I think the localStorage distinction is a bit of a red herring, since the custom rules get stored in the `chrome.storage` API, which nearly the same thing but a little fancier. I would rephrase it as: The current "load custom rulesets" code expects to be able to call `.addFromXml` on a fully constructed RuleSets object. Calling it from the middle of the constructor winds up being a bit weird. Might be worth refactoring, but not on the blocking path for this PR IMO.

LGTM, modulo @koops76's feedback, and fixing the tests.

Hainish:
@jsha, great, I'll change it back to the way it was and merge master into this branch.  Can you give me a final LGTM before I pull the trigger?

jsha:
Sure, will check back later tonight.

Hainish:
@koops76 thanks for the feedback.  Since we are trying to get this out the door, I'd like to save that for a further refactor.

koops76:
@Hainish No problem.

chromium/background.js
```diff
@@ -660,5 +682,57 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+          if(ruleset.name == ruleset_name){
+            ruleset.active = ruleset_active;
+            return;
+          }
+        }
+      }
+    })(ruleset_name, settings.rule_toggle[ruleset_name]);
+
+    // This line is necessary to clear cached lookups of rulesets already loaded
+    all_rules.ruleCache = new Map();
```
jsha: I think you reverted too far. You had an intermediate branch that looked good where you skipped the multiply-nested for loops above, and called `all_rules = new RuleSet(localStorage)`. That will work fine if you just much `load_legacy_custom_rulesets(...)` below `new RuleSet`

Hainish: @jsha we'd have to copy lines 27-32 as well...

```js
// Allow users to enable `platform="mixedcontent"` rulesets                
var enableMixedRulesets = false;                                           
storage.get({enableMixedRulesets: false}, function(item) {                 
  enableMixedRulesets = item.enableMixedRulesets;                          
  all_rules.addFromXml(loadExtensionFile('rules/default.rulesets', 'xml'));
});                                                                        
```

Hainish: Otherwise we'll have the problem I mentioned in https://github.com/EFForg/https-everywhere/pull/11792#issuecomment-321417318.

Hainish: We could do

```js
all_rules = new RuleSet(localStorage);
all_rules.addFromXml(loadExtensionFile('rules/default.rulesets', 'xml'));
load_legacy_custom_rulesets(settings.custom_rulesets);
```

Which is probably fine for now, but seems like we may run into complications when we are working on merging `sign-rulesets`.  Though I can take care of that problem when we come to it, I suppose.

Hainish:
@jsha ready for final review

chromium/background.js
```diff
@@ -660,5 +682,47 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
...
+chrome.runtime.sendMessage("import-legacy-data", import_settings);
+
+/**
+ * Import extension settings (custom rulesets, ruleset toggles, globals) from an object
+ * @param settings the settings object
+ */
+async function import_settings(settings){
+  if(settings.changed){
+    // Load custom rulesets and add to storage
+    await new Promise(resolve => {
```
koops76: Do we really need to wait until the settings are stored?

Hainish: @koops76 we do for the options ui.  We only want to display the `Settings imported` string once the settings have actually been imported.

koops76:
@Hainish My opinion is: it's OK, but has repetitive code that must be moved into a function and you probably do not need to wait until the settings are stored, or instead you could use `Promise.all`

chromium/background.js
```diff
@@ -207,8 +212,21 @@ AppliedRulesets.prototype = {
     return null;
   },
 
+  addHostnameToTab: function(tabId, ruleset_name, hostname) {
```
jsha: Looks like the "addHostnameToTab" work snuck back in by accident.

chromium/background.js
koops76: Then use `Promise.all` to await all four promises.

chromium/background.js
Hainish: @jsha this is just because this was forked from the first iteration of `firefox-webextensions-core` - should I rebase it?

Hainish: I could rebase from `master`, now that `firefox-webextensions-core` is merged.

koops76: @Hainish Rebase it I guess.

Hainish:
New PR, rebased from master: https://github.com/EFForg/https-everywhere/pull/11800

jsha:
I'll review that, but it probably would have been easier to merge rather than rebase.

Hainish:
The thing is that this PR was a request to merge in to a branch that was eventually rebased before being merged into master, so the whole history got a bit unwieldy. Let's try doing all merges next time.

On August 9, 2017 8:02:49 PM PDT, Jacob Hoffman-Andrews <notifications@github.com> wrote:
>I'll review that, but it probably would have been easier to merge
>rather than rebase.
>
>-- 
>You are receiving this because you modified the open/close state.
>https://github.com/EFForg/https-everywhere/pull/11792#issuecomment-321437766

-- 

PR #11797, opened 2017-08-10
Create an additional lookup method for favicons.

Hainish:


chromium/popup.js
```diff
@@ -72,6 +72,9 @@ function appendRuleLineToListDiv(ruleset, list_div) {
       break;
     }
   }
+  if (favicon.src == "chrome://favicon/") {
+    favicon.src += "https://" + hostnames[ruleset.name];
+  }
 
   if (navigator.userAgent.match("Chrome")) {
     var xhr = new XMLHttpRequest();
```
koops76: Do we need this XHR?

Hainish: I'll have to test without it.

koops76: OK, do it.

PR #11798, opened 2017-08-10
Delete GinjFo.com.xml

koops76:
Both `^` and `www` have incomplete certificate chain.


J0WI:
I get a mismatch on ^. Please just `default_off` the rule.

koops76:
@J0WI What should I set as a reason for `default_off`? `mismatched`?

PR #11806, opened 2017-08-10
Remove platform from *.onion rulesets

Bisaloo:
- `platform="mixedcontent"` rulesets are now `default_off` on Tor browser
- users may want to visit hidden services using a different browser than Tor browser with a custom socks5 proxy and Tor daemon
- if a user tries to reach a hidden service outside of the Tor network, the DNS request will fail so it doesn't matter if we redirect them to the `https` version of the hidden service

Because of these 3 reason, I believe `platform="mixedcontent"` should be removed from the hidden services rulesets and they should simply be added to the fetch whitelist.

Related: https://github.com/EFForg/https-everywhere/issues/3798


koops76:
Related: #11988.

koops76:
@Hainish Merge this except `ruleset-whitelist.csv`. Run the tests first though, I want to see if it works.

Hainish:
Thanks for this!  Superseded by https://github.com/EFForg/https-everywhere/pull/12093

koops76:
@Hainish Close this then I guess.

PR #11807, opened 2017-08-10
Update background.js

koops76:


chromium/background.js
```diff
@@ -679,36 +679,35 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
 // Send a message to the embedded webextension bootstrap.js to get settings to import
 chrome.runtime.sendMessage("import-legacy-data", import_settings);
 
+function setStorage(values) {
+  return new Promise(resolve => {
+    storage.set(values, resolve);
+  }
```
Hainish: Add closing parenthesis here.  Also, I'd rather have this in the `import_settings` function.  Just call await to the promise.  But consolidating the four separate calls is :+1:.

koops76: @Hainish Thanks for pointing it out!

Hainish:
Great now

> Also, I'd rather have this in the import_settings function. Just call await to the promise

koops76:
@Hainish Done.

Hainish:
Thanks!

PR #11824, opened 2017-08-11
Fix #11823

### LINKED ISSUE #11823: With WebExtension in Firefox Nightly "customize mode", HTTPS Everywhere icon's title says "(inactive)" which is misleading
STR:
 1. Start Firefox Nightly
 2. Install latest HTTPS Everywhere WebExtension from https://www.eff.org/files/https-everywhere-test/index.html
(I'm using https-everywhere-2017.7.21.1338-eff.xpi )
 3. Click the Firefox hamburger menu (top right of UI) and choose "Customize..."
 4. Drag the HTTPS Everywhere icon from your toolbar onto the overflow menu (directly below it)

ACTUAL RESULTS:
The HTTPS Everywhere entry in the overflow menu has this text "HTTPS Everywhere (inactive)" which gives the impression that the extension is inactive/disabled somehow. Screenshot:
Screenshot

EXPECTED RESULTS:
The entry should just say "HTTPS Everywhere", probably.

koops76:


koops76:
We may also want to at least internationalize these status strings if we're going to keep them.

koops76:
ping @Hainish

Hainish:
I'd rather fix this without removing the status line.  The `inactive` line is a little misleading, but it will only appear when you're customizing the layout, which is pretty seldom.

PR #11826, opened 2017-08-12
Update connect.me.xml ruleset

trevharmon:
The previous ruleset was pointing to another domain that was
generating a 404. It now simply replaces http with https. Also,
it now only handles www.connect.me.

### DIFF
--- a/src/chrome/content/rules/Connect.me.xml
+++ b/src/chrome/content/rules/Connect.me.xml
@@ -1,6 +1,9 @@
 <ruleset name="Connect.me">
 
 	<!--	Complications:
+                  The domain connect.me is
+                  currently serving an incomplete
+                  certificate chain.
 				-->
   <target host="www.connect.me"/>
 
@@ -10,7 +13,5 @@
 					-->
 	<rule from="^http:"
 		to="https:" />
-
-		<test url="http://www.connect.me/" />
 
 </ruleset>
--- a/src/chrome/content/rules/Connect.me.xml
+++ b/src/chrome/content/rules/Connect.me.xml
@@ -1,22 +1,16 @@
-<!--
-	^connect.me: Expired, missing certificate chain
-	www.connect.me: Refused
-
--->
 <ruleset name="Connect.me">
 
 	<!--	Complications:
 				-->
-  <target host="connect.me"/>
   <target host="www.connect.me"/>
 
 
 	<!--	Redirect keeps path and args,
 		but not forward slash:
 					-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Connect.me.xml
```diff
@@ -1,22 +1,16 @@
-<!--
-	^connect.me: Expired, missing certificate chain
-	www.connect.me: Refused
-
--->
 <ruleset name="Connect.me">
 
 	<!--	Complications:
 				-->
-  <target host="connect.me"/>
```
J0WI: Why did you remove this target?

src/chrome/content/rules/Connect.me.xml
```diff
@@ -1,22 +1,16 @@
...
 	<!--	Redirect keeps path and args,
 		but not forward slash:
 					-->
-	<rule from="^http://(?:www\.)?connect\.me/+"
-		to="https://www.respectnetwork.com/connect-me-2/" />
+	<rule from="^http:"
+		to="https:" />
 
-		<test url="http://connect.me//" />
+		<test url="http://www.connect.me/" />
```
J0WI: All listed targets are implicit test urls.

src/chrome/content/rules/Connect.me.xml
trevharmon: `connect.me` and `www.connect.me` both point to different hosts. At the moment, `connect.me` is undergoing some changes, so the https redirects weren't happening correctly. My intention was to get `www.connect.me` fixed properly and then, when `connect.me` became stable to submit another PR to update that URL.

src/chrome/content/rules/Connect.me.xml
trevharmon: Should they just removed then? At the moment, we aren't needing any wildcards on this domain at this time.

src/chrome/content/rules/Connect.me.xml
J0WI: We use to add comments for non-working hosts. Please document that `connect.me` serves an incomplete certificate chain.

src/chrome/content/rules/Connect.me.xml
J0WI: I would merge it anyway, but since I requested other changes you can also remove this redundant test url.

trevharmon:
Made requested changes to the ruleset definition.

J0WI:
Thanks!

PR #11838, opened 2017-08-12
Plista:

numismatika:
- convert to current styleguide
- simplify rules

### DIFF
--- a/src/chrome/content/rules/Plista.xml
+++ b/src/chrome/content/rules/Plista.xml
@@ -1,18 +1,145 @@
-<ruleset name="plista (partial)">
+<!--
+	Chain issues:
+		- grafana.plista.com
+		- graphite.plista.com
+		- plista573.plista.com
+		- plista87.plista.com
+		- projects.plista.com
 
-	<target host="plista.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Plista.xml
```diff
@@ -1,18 +1,17 @@
 <ruleset name="plista (partial)">
 
 	<target host="plista.com" />
-	<target host="*.plista.com" />
+	<target host="farm.plista.com" />
+	<target host="farm-de.plista.com" />
+	<target host="static.plista.com" />
+	<target host="static-de.plista.com" />
+	<target host="www.plista.com" />
```
J0WI: See also:
```
blog.plista.com
tech.plista.com
jira.plista.com
planet.plista.com
orp.plista.com
mail.plista.com

File: src/chrome/content/rules/Plista.xml [hunk 2]
```diff
@@ -1,18 +1,17 @@
 <ruleset name="plista (partial)">
```
J0WI: Are we still `partial`?

PR #11851, opened 2017-08-13
Remove unnecessary exclusion from CNBC.com.xml

cschanaj:
`http://cnbc.com`  now redirect (301) to `http://www.cnbc.com` which work over HTTPS correctly.


J0WI:
Please also remove `www` from the comments.

PR #11274, opened 2017-07-22
Remove https-everywhere-checker message

cschanaj:
Source code available at [`remove-checker-message.go`](https://github.com/cschanaj/lambda/blob/master/src/remove-checker-message.go), see [`issue-11274-list.txt`](https://gist.github.com/cschanaj/311344e50158c7c2ee9099510e851711) for a list of 241 files excluded from this PR because of failing tests. 

[`Bitly_branded_short_domains.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Bitly_branded_short_domains.xml) lengthened the Travis build time significantly, excluded it for everyone's sake. (This particular ruleset contains **25% (36k of 120k)** of `target` in the whole code base! 0rz...)

PR #11854, opened 2017-08-13
[Buerge-CERT.de] Simplify ruleset

Bisaloo:
@cschanaj, it looks like this one managed to get through your automated comment removal.


cschanaj:
@Bisaloo, some 200+ files failed the `fetch-test`/ `rule-test` are excluded from #11274 and #11525 , see https://github.com/EFForg/https-everywhere/pull/11274#issuecomment-318867682 . There is also some false results, I'll try to run the script again when I have fixed the regex. thanks.

PR #11860, opened 2017-08-13
ZAM:

numismatika:
- comment about another subdomain with invalid cert

### DIFF
--- a/src/chrome/content/rules/Zam.xml
+++ b/src/chrome/content/rules/Zam.xml
@@ -1,30 +1,88 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://lkcss.zamimg.com/ => https://lkcss.zamimg.com/: Too many redirects while fetching 'https://lkcss.zamimg.com/'
-Fetch error: http://lkimg.zamimg.com/ => https://lkimg.zamimg.com/: Too many redirects while fetching 'https://lkimg.zamimg.com/'
-Fetch error: http://lkjs.zamimg.com/ => https://lkjs.zamimg.com/: Too many redirects while fetching 'https://lkjs.zamimg.com/'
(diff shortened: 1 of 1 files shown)


J0WI:
Please note that this rule is `default_off`. Would you like to re-enable it?

J0WI:
ping @numismatika

numismatika:
@J0WI working on it

src/chrome/content/rules/Zam.xml
```diff
@@ -1,30 +1,82 @@
...
+	<target host="loldata.services.zam.com" />
+	<target host="media.services.zam.com" />
+	<target host="media-staging.services.zam.com" />
+	<target host="parkplace-staging.services.zam.com" />
 
 	<target host="cdn.zamimg.com"/>
 	<target host="descss.zamimg.com"/>
 	<target host="desimg.zamimg.com"/>
 	<target host="desjs.zamimg.com"/>
-	<target host="lkcss.zamimg.com"/>
```
J0WI: Can we keep a comment for it?

src/chrome/content/rules/Zam.xml
```diff
@@ -1,30 +1,82 @@
...
 	<target host="lkimg.zamimg.com"/>
-	<target host="lkjs.zamimg.com"/>
```
J0WI: Can we keep a comment for it?

src/chrome/content/rules/Zam.xml
numismatika: both of these seem to no longer be in use on lolking and replaced by content on the .services subdomain

J0WI: But they are still accessible.

numismatika: mhmm i think i removed them for not having proper test urls and non 200 status codes or some sort of error like that, let me check that again

J0WI: You can just keep them as comments.

PR #11879, opened 2017-08-14
Create LarsJung.de.xml

koops76:


src/chrome/content/rules/LarsJung.de.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="LarsJung.de">
+	<target host="larsjung.de" />
+	<target host="www.larsjung.de" />
+	<target host="release.larsjung.de" />
```
J0WI: `m.larsjung.de`

src/chrome/content/rules/LarsJung.de.xml
```diff
@@ -0,0 +1,10 @@
...
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://www\.larsjung\.de/" to="https://larsjung.de/" />
```
J0WI: Please add a comment for the issue on `www`

koops76: Done

PR #11889, opened 2017-08-14
exonit.by

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/exonit.by.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+www.exonit.by ¹
```
J0WI: WFM

galeksandrp: Added.

PR #11898, opened 2017-08-14
shop-dent.ru

galeksandrp:
Issue author: @galeksandrp


src/chrome/content/rules/shop-dent.ru.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+www.shop-dent.ru different content
```
J0WI: WFM, both redirect to ^

galeksandrp: Added.

PR #11922, opened 2017-08-15
add rule for WiWo.de

numismatika:


J0WI:
I took some random samples and found various mixed active content issues.

numismatika:
@J0WI yea there is a warning but did you find anything that "breaks" the page?
VGWort and visualrevenue is tracking thirdparty content that doesn't really harm the user experience if it isn't loaded. On the contrary, it benefits the user.

numismatika:
indeed , since we cannot fix that platform= mixed

PR #11924, opened 2017-08-15
add rule for Volantix.com:

numismatika:


J0WI:
Does this site have any content?

src/chrome/content/rules/Volantix.com.xml
```diff
@@ -0,0 +1,29 @@
+<!--
+		Other Instart Logic rulesets:
+
+		- Insnw.net.xml
+		- InstartLogic.com.xml
```
J0WI: Please also update the reference list there.

numismatika:
It is just referenced by other pages as third party object. Suspect it just serves as tracking host.

J0WI:
@numismatika If they are related the references should be updated on those other files. If they are not related, you can remove the reference list from this rule.

PR #11934, opened 2017-08-15
Add whitelist Travis task

koops76:


koops76:
Test failure is expected.

koops76:
Test success is expected now. 😁

koops76:
@Hainish This is ready for merge.

cschanaj:
@koops76 does this implies that we will have to remove an existing entry manually every time we update the items in the list?

P.S. I am not really good at `git`, please correct me if I am wrong :smile:

koops76:
@cschanaj Yes, it's not done automatically. Also keeping the alphabetic order is now necessary for the tests to pass.

cschanaj:
@koops76 This is a good idea since it will keep the white-list synced. However, there is a few points I am concerned about

- Some contributors, like @Bisaloo and I have been working on the items in the white-list. This PR will reinforce us to run the cleanup script whenever there is a change to the white-listed item. IMHO, this is sort of troublesome if the respective changes are massive.

- Suggestion: there should be an error message suggesting the contributor to run the cleanup script when Travis failed `whitelist`. (enhancement)

- Suggestion: if possible, maybe we should check that _ONLY the changed files are added/ removed from the whitelist_. This will avoid unnecessary or possibly malicious whitelisting. (optional)

I am open to merging this with or without the aforementioned changes. thanks.

Possibly Related: #11597, #11560

koops76:
@cschanaj
1. You can simply ignore `whitelist` test failure until the rest of the PR is finished.

test/travis.sh
```diff
@@ -87,4 +88,11 @@ if [ "$RULESETS_CHANGED" ]; then
   fi
 fi
 
+if $RULESETS_CHANGED || $WHITELIST_CHANGED; then
+  if [ "$TEST" == "whitelist" ]; then
+    utils/ruleset-coverage-whitelist-cleanup.sh
+    [ `git diff --name-only | wc -l` -eq 0 ]
```
cschanaj: @koops76 here? with a brief message I guess.

Hainish:
Adding more tests to `.travis.yml` increases the overall running time of Travis.  In this case, there's no concrete benefit - your tests will fail if the whitelist is out of sync, and that adds a development burdon.  For the longest time we had no cleanup script at all, and that was mostly okay.  The periodic run of the cleanup script, upon every release, is sufficient.

koops76:
@Hainish If the script will be modified to only check changed entries, will you add it to tests?

PR #11865, opened 2017-08-13
Remove https-everywhere-checker message (3rd run)

cschanaj:
As per #11854, related #11274 and #11525 

~I would expect the result from `fetch-test` is chaotic ...~ Not really :smiley:

PR #11942, opened 2017-08-16
Update ruleset-coverage-whitelist-cleanup.sh

cschanaj:
@Hainish it should be the 4th column if we are sorting according to the filenames. It is _sorted by the third column, then by the fourth column_ currently, leading to strange sorting in #11865 thank!

Also, please run the cleanup script for once after this has been merged.


cschanaj:
@Hainish Can you merge this simple fix? ruleset-whitelist.csv is currently sorted in the wrong order.

PR #11945, opened 2017-08-16
NodeJs to remove obsolete https-everywhere-checker messages

cschanaj:


koops76:
@cschanaj Why close?

cschanaj:
@koops76 since some modifications were needed. see #11963

PR #11961, opened 2017-08-17
Update nexusmods.com.xml (fixes #11953)

### LINKED ISSUE #11953: nexusmods.com rule breaks downloading
It seems commit 27fef948c9fd3ed653581b1fa77ff14697b309d7 is breaking downloading on nexusmods.com (on Chrome) since some issues with CORS and Mixed Content.

```
Uncaught DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'http://www.nexusmods.com/fallout4/mods/24491/?tab=2&navtag=http%3A%2F%2Fwww.nexusmods.com%2Ffallout4%2Fajax%2Fmodfiles%2F%3Fid%3D24491&pUp=1' cannot be created in a document with origin 'https://www.nexusmods.com' and URL 'https://www.nexusmods.com/fallout4/mods/24491'.

Mixed Content: The page at 'https://www.nexusmods.com/fallout4/mods/24491' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://www.nexusmods.com/fallout4/ajax/downloadfile?id=106843'. This content should also be served over HTTPS.

Failed to load https://www.nexusmods.com/fallout4/ajax/downloadfile?id=106843: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access.
```

koops76:


cschanaj:
also fix #11970

koops76:
@cschanaj Likely the same issue.

umisery:
How do I apply this fix?

koops76:
@umisery Simply wait until it will be merged, the next release will have it. Until then disable NexusMods.com ruleset.

umisery:
Alright, thanks.

springacres:
@umisery also, make sure to disable the ruleset before you log in to any nexusmods.com site.

umisery:
I have no idea how to do either of those, so I'm just going to wait for it to be merged :P

src/chrome/content/rules/nexusmods.com.xml
```diff
@@ -1,11 +1,11 @@
...
+		- ^
+		- www
+		- wiki
 -->
 
 <ruleset name="NexusMods.com">
-	<target host="nexusmods.com" />
-	<target host="www.nexusmods.com" />
 	<target host="filedelivery.nexusmods.com" />
 	<target host="forums.nexusmods.com" />
```
J0WI: Can we add `staticdelivery.nexusmods.com`?

src/chrome/content/rules/nexusmods.com.xml
```diff
@@ -1,11 +1,11 @@
 <!--
 	Mixed content:
```
J0WI: Is it worth to create a MCB platform rule for them?

J0WI:
To fix the mentioned issues I think we can also add an exclusion. Other parts of the site WFM.

koops76:
@Hainish Merge before the release, NexusMods.com is broken.

Hainish:
We should set a policy on what rules are worthy of blocking release for.  Looks like this one ranks 1,070, hardly a show-blocker.

J0WI:
This should not be a blocker for the current release. #12268 has a much bigger impact than any ruleset.

@koops76 I already commented on your changes. When we remove the entry host we don't fix anything, we just disable it.

cschanaj:
In https://github.com/EFForg/https-everywhere/issues/11953#issuecomment-326191685, it is mentioned that 
> To confirm, Nexus Mods does not currently support HTTPS on the main site as per the following forum thread: https://forums.nexusmods.com/index.php?/topic/5907893-site-problems-an-error-occurred/

A number of users have experienced issues with this site (see #11953, #12324 and #12505). IMHO, We should disable this ruleset if it is impossible to fix. (Possibly with `default_off="other"` and a reference to the forum link). thanks.

src/chrome/content/rules/nexusmods.com.xml
koops76: Maybe but unlikely. I don't think anyone actually has active mixed content blocking disabled after Tor Browser enabled it. I think `platform="mixedcontent"` rulesets are on their way out.

PR #11964, opened 2017-08-17
Created ruleset for theanarchistlibrary

Devices:
Added rules for theanarchistlibrary.org, its subdomains (such as fr.theanarchistlibrary.org) and other domains.


Devices:
Did not see it was already added, sorry.

PR #11966, opened 2017-08-17
Updated theanarchistlibrary

Devices:
Added two new subdomains. Also added three other domains that provide the library in different languages, but are still part of the same project.

### DIFF
--- a/src/chrome/content/rules/TheAnarchistLibrary.org.xml
+++ b/src/chrome/content/rules/TheAnarchistLibrary.org.xml
@@ -17,8 +17,11 @@
 	<target host="fr.theanarchistlibrary.org" />
 	<target host="www.theanarchistlibrary.org" />
 	<target host="www.anarhisticka-biblioteka.org" />
+	<target host="anarhisticka-biblioteka.org" />
 	<target host="anarchistischebibliothek.org" />
+	<target host="www.anarchistischebibliothek.org" />
 	<target host="anarhisticka-biblioteka.net" />
+	<target host="www.anarhisticka-biblioteka.net" />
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/TheAnarchistLibrary.org.xml
```diff
@@ -13,7 +13,12 @@
 	<target host="ru.theanarchistlibrary.org" />
 	<target host="es.theanarchistlibrary.org" />
 	<target host="sv.theanarchistlibrary.org" />
+	<target host="da.theanarchistlibrary.org" />
+	<target host="fr.theanarchistlibrary.org" />
 	<target host="www.theanarchistlibrary.org" />
+	<target host="www.anarhisticka-biblioteka.org" />
+	<target host="anarchistischebibliothek.org" />
+	<target host="anarhisticka-biblioteka.net" />
```
J0WI: Please always add the `www` and non-www aliases of those three new hosts.

Devices: Even if they redirect to the non-www? www.anarchistischebibliothek.org redirects to anarchistischebibliothek.org and anarhisticka-biblioteka.org redirects to www.anarhisticka-biblioteka.org.

J0WI: We want to secure everything, even redirects ;)

Devices: Makes sense. I hope it's good like this.

J0WI: LGTM, Thanks!

PR #11998, opened 2017-08-20
Verify each target has the right number of tests

cschanaj:
blocked by #11187 which has been pending review for a long time... Also, merge #11942 before whitelisting `rules-test`.

See https://gist.github.com/cschanaj/b3c80c935226c55e78013d81d618cd99

Related: #7272


test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -180,15 +180,55 @@ def _determineTestApplication(self):
...
+				needed_count = 3
+
+			if target.endswith(".*"):
+				needed_count = 10
+
+			# non-wildcard target always have a implicit test url, if is it not excluded
+			if not "*" in target and not self.excludes(("http://%s/" % target)):
+				continue
+
+			# '*.example.com' match 'www.example.com' but not 'secure.account.exmple.com'
```
cschanaj: @Hainish does `*.example.com` matches `www.example.com`, `secure.account.example.com` and `example.com`? I am a bit confused seeing https://support.google.com/customsearch/answer/71826?hl=en

Hainish: @cschanaj the logic is here: https://github.com/EFForg/https-everywhere/blob/454fb4a9a8f60ed7db6e0624637574b0534a6a14/chromium/rules.js#L350-L355

So to answer, for a.b.examble.com, *.example.com *is* a matching target.

cschanaj:
@Hainish I have fixed the wildcard matching logic in c255212 and this is ready for an review. thanks!!

P.S. please have a look at #11942 which fix a trivial bug in the ruleset whitelist cleanup script.

Hainish:
Got it, I should have some time in the next few days to review this.

cschanaj:
@Hainish I have extracted the matching logic from `rule.js`, it seems that if `example.*` does not match`example.co.uk` however `example.com` does, i.e. 

```xml
<target host="example.*" />
```
does not work for `example.co.uk`. It seems like a bug to me, can you verify it?

```javascript
var myfunct = function(host) {
	var tmp;
	var results = [];

	// Ensure host is well-formed (RFC 1035)
	if (host.indexOf("..") != -1 || host.length > 255) {
		log(WARN,"Malformed host passed to potentiallyApplicableRulesets: " + host);
		return null;
	}

	// Replace each portion of the domain with a * in turn
	var segmented = host.split(".");
	for (let i=0; i < segmented.length; i++) {
		let tmp = segmented[i];
		segmented[i] = "*";
		// results = results.concat(this.targets[segmented.join(".")]);
		console.log(segmented.join("."))
		segmented[i] = tmp;
	}
	for (var i = 2; i <= segmented.length - 2; ++i) {
		var t = "*." + segmented.slice(i,segmented.length).join(".");
		console.log(t);
		// results = results.concat(this.targets[t]);
	}
}

console.log('**** example.co.uk ****');
myfunct('example.co.uk');

console.log('**** example.com  ****');
myfunct('example.com');
```

```bash
$ nodejs rules.js 
**** example.co.uk ****
*.co.uk
example.*.uk
example.co.*
**** example.com  ****
*.com
example.*

```

cschanaj:
@Hainish It seems that we treat `*` in different position differently, according to my understanding of https://github.com/EFForg/https-everywhere/pull/11998#issuecomment-324275475, it seem that 

| `target` | `regex` interpretation |
| :---: | :---: |
| `*.example.com` | `.+\.example\.com$` |
| `secure.*.example.com` | `^secure\.[^\.]*\.example\.com$` |
| `www.example.*` | `^www\.example\.[^\.]*` |

cschanaj:
@Hainish merge #12091 will fix Travis from this PR. thanks!

cowlicks:
@cschanaj I'm looking into this today! Sorry for the delay.

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -180,15 +180,69 @@ def _determineTestApplication(self):
...
+
+			if target.endswith(".*"):
+				needed_count = 10
+
+			# non-wildcard target always have a implicit test url, if is it not excluded
+			if not "*" in target and not self.excludes(("http://%s/" % target)):
+				continue
+
+			# According to the logic in rules.js available at
+			# EFForg/https-everywhere/blob/master/chromium/rules.js#L350-L355
```
cowlicks: Can you include the commit hash in this URL? This way we can find the referenced code if it moves. You can get a permalink to lines in a specific version of the file by selecting the lines normally, then pressing "y".

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -180,15 +180,69 @@ def _determineTestApplication(self):
...
+			# 
+			pattern = target.replace('.', '\.') # .replace('*', '.+')
+
+			# `*.example.com` matches `bar.example.com` and `foo.bar.example.com` etc.
+			if pattern[0] == '*':
+				pattern = pattern.replace('*', '.+')
+
+			# however, `example.*` match `example.com` but not `example.co.uk`
+			if pattern[len(pattern) - 1] == '*':
```
cowlicks: You can get the last element like this: `pattern[-1]`

cowlicks:
@cschanaj I'm not very familiar with running these tests. Would you mind providing a testcase this PR fixes? I would just run it by pointing the `rulesdir` in `checker.config` to it, right?

cschanaj:
@cowlicks I run the rules test by the following command

```bash
$ export TEST=rules
$ ./test.sh
```

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -180,15 +180,69 @@ def _determineTestApplication(self):
...
-			 that applies to it. TODO: Also check that each target has the right
+		"""Verify that each target, rule and exclusion has the right number of tests
+			 that applies to it. Also check that each target has the right
 			 number of tests. In particular left-wildcard targets should have at least
 			 three tests. Right-wildcard targets should have at least ten tests.
 
 			 Returns an array of strings reporting any coverage problems if they exist,
 			 or empty list if coverage is sufficient.
 			 """
 		problems = self._determineTestApplication()
```
cowlicks: Every "`problem`" gets printed twice, they are created here in `getCoverageProblems`. This happens because `problems` is a reference to `self.test_application_problems` through `self._determineTestApplication`.  Then again in `getTargetValidityProblems` we do `problems = self._determineTestApplication`, so the `problems` from `getCoverageProblems` also get reported through `getTargetValidityProblems`.

I'm not sure why `problems` is assigned like this. I'm guessing it is a mistake in either this function or `getTargetValidityProblems`. Maybe @Hainish can chime in.

cowlicks: @cschanaj You can probably just change this to:
```javascript
self._determineTestApplication()
problems = []
```
that is what other tests do

cowlicks:
@cschanaj  git can be tricky sometimes. I can fix this for you if you'd like.

cowlicks:
I revisited this today. It looks good to me. It should be merged in conjunction with #1209, but #12091 needs to be updated again. Can you update it  @cschanaj ? then we can merge this.

Also, docker on travis is failing for some reason.

cschanaj:
@cowlicks I have rebased this and merged #12091 here. I will update #11560 once this is merged. thanks!

cowlicks:
@cschanaj great work. Thank you.

PR #12002, opened 2017-08-20
Re-enable MIT.xml

koops76:


J0WI:
I preferred #12652 because that was a bugfix-only and was up to date with our master.
You have some additional updates in your PR. It would be great if you could rebase your branch, I'll have a look on it then.

J0WI:
Well, it turns out that the rule is still disabled after #12652

koops76:
@J0WI Done.

src/chrome/content/rules/MIT.xml
```diff
@@ -336,18 +179,16 @@
...
 	<target host="www.ll.mit.edu" />
-
 	<target host="immersion.media.mit.edu" />
 	<target host="www.media.mit.edu" />
-
+	<target host="media.mit.edu" />
 	<target host="mit-amps.mit.edu" />
 	<target host="mit150.mit.edu" />
 	<target host="mitei.mit.edu" />
-	<target host="mitei-members.mit.edu" />
```
J0WI: Please add a comment for it

src/chrome/content/rules/MIT.xml
```diff
@@ -299,31 +142,31 @@
...
-
 	<target host="dc.mit.edu" />
 	<target host="debathena.mit.edu" />
 	<target host="athena.dialup.mit.edu" />
 	<target host="dome.mit.edu" />
 	<target host="dspace.mit.edu" />
-
 	<target host="eecs.mit.edu" />
 	<target host="www.eecs.mit.edu" />
-
```
J0WI: Please add `energy.mit.edu`

src/chrome/content/rules/MIT.xml
```diff
@@ -361,134 +202,67 @@
...
 	<target host="shop.sloanreview.mit.edu" />
 	<target host="subscribe.sloanreview.mit.edu" />
-
 	<target host="stellar.mit.edu" />
 	<target host="stuff.mit.edu" />
 	<target host="teachingexcellence.mit.edu" />
 	<target host="techtv.mit.edu" />
 	<target host="tll.mit.edu" />
 	<target host="wayf.mit.edu" />
-	<!--target host="webmail.mit.edu" /-->
```
J0WI: WFM, please readd it and remove it from the comments.

J0WI: Same for `^`, `www` and `web`

J0WI: Lookls like `www` has still some issues, but it can be redirected to `web`. Please have a look at the duplicates in `MIT-mismatches.xml`

jeremyn:
@J0WI I'd like to re-enable the MIT ruleset. What do you think we should do? We can't work with this PR because it's from @ghost now.

I ran `fetch-test.sh` against the current `MIT.xml` and only see two errors:

```
ERROR rules/MIT.xml: Fetch error: http://oeit-tsa.mit.edu/ => https://oeit-tsa.mit.edu/: (28, 'Connection timed out after 20001 milliseconds')
ERROR rules/MIT.xml: Fetch error: http://mitei-members.mit.edu/ => https://mitei-members.mit.edu/: (51, "SSL: no alternative certificate subject name matches target host name 'mitei-members.mit.edu'")
```

Should I just make another PR that just fixes these minimal problems? I don't want to do an exhaustive check for this huge site.

Also, what do you think about moving the keyserver https://pgp.mit.edu to its own ruleset just for that one thing? It's extra high value and I don't think it should depend on a hundred other MIT domains.

J0WI:
There is a lot of work here I would prefer to keep.
I can merge this in a new branch to keep all the commits. Then you can open a PR to this branch. I don't want to merge this in the master branch as long as it breaks stuff.

jeremyn:
Re-enabling a complicated ruleset really means the whole thing needs to be reviewed, so my suggestion to just fix two things and then re-enable it is not a good approach.

I'm not really interested in reviewing an old, complicated, massive ruleset like this. In other words if you want to resubmit it as your own PR or something, go ahead, but it will probably sit there for a long time.

I would like to move `pgp` to its own ruleset since that is quick to do, and `pgp` is especially important to support.

J0WI:
I already reviewed parts of this PR, so I'll continue to review your changes until this is ready to merge.

jeremyn:
I don't think we should merge commits from people who have deleted their accounts. It's like I wrote in https://github.com/EFForg/https-everywhere/pull/13887/commits/9f9c465b6d56299ad25e998a7eea92e3422cade7 in https://github.com/EFForg/https-everywhere/pull/13887 :

> * We want to avoid including updates from contributors who have withdrawn permission for us to use their work. If a contributor deletes their GitHub account, it's not clear whether they have withdrawn their permission.

So in that case one of us would need to rewrite the old commits under their own name, and the other would have to review it. Either way I personally would need to review all the changes and I don't really feel like doing that.

PR #12011, opened 2017-08-20
Create zeranoe.com.xml

koops76:


src/chrome/content/rules/zeranoe.com.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	HTTP 526:
+		- www
```
J0WI: Same for `gstreamer`

koops76: @J0WI Done.

J0WI:
Let's see if they fix it before the next release.

koops76:
@J0WI At least message them.

PR #12029, opened 2017-08-21
Update and rename Calyx-Institute.xml to CalyxInstitute.xml

koops76:


src/chrome/content/rules/CalyxInstitute.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="CalyxInstitute.org">
+	<target host="calyxinstitute.org" />
+	<target host="www.calyxinstitute.org" />
+	<target host="vanunu.calyxinstitute.org" />
```
J0WI: See also `tor.calyxinstitute.org`

J0WI: ... which serves an expired cert.

koops76: @J0WI Done.

PR #12034, opened 2017-08-21
Update NIIF.xml

koops76:


src/chrome/content/rules/NIIF.xml
```diff
@@ -1,14 +1,13 @@
...
+		- www.hboneplus.hu
 -->
-<ruleset name="NIIF Int&#233;zet (partial)" default_off='failed ruleset test'>
-	<target host="www.hboneplus.hu" />
+
+<ruleset name="NIIF Int&#233;zet (partial)">
+	<target host="niif.hu" />
 	<target host="www.niif.hu" />
 	<target host="www.ca.niif.hu" />
+	<target host="keys.niif.hu" />
```
J0WI: Would you like to search for additional subdomains?

koops76: @J0WI No, the point of this PR is to add the keyserver subdomain to this ruleset.

PR #12042, opened 2017-08-21
Split Greek-various.xml and add targets

koops76:
Based on #10808.


J0WI:
Could you please open one PR per new file? The domains are not related at all and would make the review process to complex.

koops76:
@J0WI I can't. If I will there would be duplicating domains.

J0WI:
You can do like in #12672

PR #12049, opened 2017-08-21
Fix translation code and replace "i18n" with "data-i18n" ("i18n" is not a valid HTML5 attribute)

koops76:


koops76:

Hainish:
Reviewing this now

PR #12083, opened 2017-08-24
Create betafamily.com.xml

ihave100openPRs:


src/chrome/content/rules/betafamily.com.xml
```diff
@@ -0,0 +1,18 @@
+<ruleset name="betafamily.com">
+	<target host="betafamily.com" />
+	<target host="*.betafamily.com" />
```
J0WI: Please document why a wildcard is required here.

ihave100openPRs:
@J0WI Is it alright now?

PR #12094, opened 2017-08-25
add Fail2ban.org.xml

J0WI:


src/chrome/content/rules/Fail2ban.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Fail2ban.org">
+	<target host="fail2ban.org" />
+	<target host="www.fail2ban.org" />
+
```
koops76: Add `<securecookie host=".+" name=".+" />`.

PR #6055, opened 2016-08-03
Added website, Updated Hidden services for Travis

stylersnico:

PR #12095, opened 2017-08-25
Add rulesets from #6055

koops76:


koops76:
None of the `.onion` domains have matching certificates and the non-`.onion` rulesets require rewriting.

PR #12100, opened 2017-08-26
Update Fanhattan.com.xml (disabled twice)

cschanaj:


J0WI:
See also `fan.tv`

cschanaj:
Added a nbw ruleset for `fan.tv`. thanks!

PR #12102, opened 2017-08-26
[104.com.tw] Update ruleset.

bardiharborow:


src/chrome/content/rules/104.com.tw.xml
```diff
@@ -1,44 +1,44 @@
...
-
-Redirects to http:
-	corp
-
-Invalid certificate:
-	^	(equal to www.)
-
--->
-
 <ruleset name="104.com.tw">
```
gloomy-ghost: Please add `104.com.tw`, and move `www.104.com.tw` right after it

bardiharborow: Oops, didn't mean to exclude `^`. Done.

PR #12111, opened 2017-08-26
Update OpenX.org.xml

cschanaj:


src/chrome/content/rules/OpenX.org.xml
```diff
@@ -1,60 +1,37 @@
...
-		- adserver
-
+	Other OpenX.com related rulesets, see
+		+ OpenX.xml
+
+	Non-functional hosts
+		Timeout was reached:
+		- codex.openx.org
+		- confluence.openx.org
+		- dc.openx.org
```
gloomy-ghost: I think we don't need to list a host if it is not functional in both http and https.

Issue #12122, opened 2017-08-26
Add more rulesets for .onion domains from Certificate Transparency list


koops76:
List of all certificates issued for `.onion` domains, based on https://crt.sh/?q=%25.onion.
- [x] 3g2upl4pq6kufc4m.onion (#12150)
- [x] ajarouh33jee5tos.onion (#12154)
- [x] blockchainbdgpzk.onion (#12129)
- [x] ~~blockchatvqztbll.onion~~ (mismatched)
- [x] ~~btcforuml2pcfkyx.onion~~ (expired)
- [x] coinpaymtstgtibr.onion (#12140)
- [x] cyphdbyhiddenbhs.onion
- [x] facebookcorewwwi.onion
- [x] fbcdn23dssr3jqnq.onion
- [x] fbsbx2q4mvcl63pw.onion (#12131)
- [x] fyxifz6xwx3d67a2.onion (#12140)
- [x] k6jyto7zq7xxriux.onion (#12155)
- [x] nytimes2tsqtnxek.onion (#12125)
- [x] osf22p3lmweopgho.onion (#12141)
- [x] ppasset42kropoy6.onion (#12142)
- [x] privacyintyqcroe.onion (#12126)
- [x] propub3r6espa33w.onion (#12142)
- [x] protonirockerxow.onion
- [x] pubapp7v22ykdou3.onion (#12142)
- [ ] pubdrop4dw6rk3aq.onion
- [x] sarajarxpvoenum6.onion (#12156)
- [x] tess2mkorb55pl7i.onion
- [x] tomritterbassljd.onion (#12143)
- [x] usatodayw7vu5egc.onion (#12153)
- [x] xczxxczkjivdfsop.onion
- [x] y6xjgkgwj47us5ca.onion

Bisaloo:
btcforuml2pcfkyx.onion cert expired

J0WI:
I'm closing this since this issue has no more open dependencies.

PR #12125, opened 2017-08-26
Create nytimes2tsqtnxek.onion.xml

Bisaloo:
#12122


koops76:
@Bisaloo Better stick it in `NYTimes.com.xml` though.

Bisaloo:
> @Bisaloo Better stick it in NYTimes.com.xml though.

Why?

koops76:
@Bisaloo Technically the same website.

Bisaloo:
True. But it is quite possible that the onion may be down when the clearnet domain isn't (or *vice versa*). It is more robust to add them in different rulesets in that regard.

PR #12129, opened 2017-08-26
Update and rename Blockchainbdgpzk.onion.xml to Blockchain.xml

koops76:
#12122


cschanaj:
IMHO it is better to keep its original name to avoid confusions with the non-tor blockchain related site.

koops76:
@cschanaj It's preloaded anyway.

cschanaj:
thanks.

PR #12131, opened 2017-08-26
Update FacebookCoreWWWi.onion.xml

koops76:


src/chrome/content/rules/FacebookCoreWWWi.onion.xml
```diff
@@ -1,64 +1,30 @@
...
 	<target host="facebookcorewwwi.onion" />
 	<target host="*.facebookcorewwwi.onion" />
 
-		<test url="http://developers.facebookcorewwwi.onion/" />
-		<test url="http://m.facebookcorewwwi.onion/" />
-		<test url="http://www.facebookcorewwwi.onion/" />
+	<target host="*.fbcdn23dssr3jqnq.onion" />
 
-	<target host="scontent.xx.fbcdn23dssr3jqnq.onion" />
+	<target host="*.fbsbx2q4mvcl63pw.onion" />
```
J0WI: We require at least 3 test urls per wildcard.

koops76: Can you look for two more test URLs?

J0WI: I think there is no way to do this for onion domains.

gloomy-ghost:
```
ERROR src/chrome/content/rules/FacebookCoreWWWi.onion.xml: Not enough tests (1 vs 3) for *.fbcdn23dssr3jqnq.onion
ERROR src/chrome/content/rules/FacebookCoreWWWi.onion.xml: Not enough tests (0 vs 3) for *.fbsbx2q4mvcl63pw.onion
```

PR #12135, opened 2017-08-27
Trivialize capturing direct rewrites

cschanaj:
GO script: [`trivialize-rules.go#L242`](https://github.com/cschanaj/trivialize-rules/blob/4aff5c6493e9ad6e00d5d7305f3157d0ed8304ab/src/trivialize-rules.go#L242)

Excluded: `default_off`, `exclusion`, `failed fetch-test`

### DIFF
--- a/src/chrome/content/rules/Abusix.com.xml
+++ b/src/chrome/content/rules/Abusix.com.xml
@@ -1,10 +1,13 @@
 <ruleset name="abusix.com">
 
 	<target host="abusix.com" />
-	<target host="*.abusix.com" />
+	<target host="abusehq.abusix.com" />
+	<target host="blackholemx.abusix.com" />
+	<target host="leakdb.abusix.com" />
+	<target host="spamfeedme.abusix.com" />
+	<target host="www.abusix.com" />
 
 
-	<rule from="^http://((?:abusehq|blackholemx|leakdb|spamfeedme|www)\.)?abusix\.com/"
-		to="https://$1abusix.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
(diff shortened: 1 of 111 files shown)


bardiharborow:
Are you able to tweak your code to also pickup expressions like `http://(?:(mb\.|tools\.)|www\.)?0nl1ne\.at/`?

cschanaj:
Picking up expression like this is too complicated using regex, I have no intention to do it currently (at least in the short terms). 

However, there is some existing library claim to expand regex with finite matches available (e.g. https://github.com/asciimoo/exrex), we might want to look into that if we are going to perform rule trivialization to a deeper level.

RReverser:
@bardiharborow @cschanaj I've added support for such regexps locally as stated in https://github.com/EFForg/https-everywhere/pull/4400#issuecomment-326046155, will be creating a PR soon.

RReverser:
My only concern with this PR, previous @cschanaj's https://github.com/EFForg/https-everywhere/pull/11729 and well, my own previous https://github.com/EFForg/https-everywhere/pull/4400 is that in neither of them we took care of `securecookie` potentially breaking its functionality.

Problem is, when we have `<target host="*.somehost.com" />` but rules are only like `<rule from="^http://sub\.somehost\.com/" to="..." />`, we change targets to `<target host="sub.somehost.com" />` when that might be incorrect, because even if `<rule />` is not applicable to certain subdomains, `securecookie` for these subdomains might be still important and would break on such change.

RReverser:
@cschanaj Didn't see you had a bunch of other similar PRs in between. Well, I guess securecookie is broken anyway for these cases already and we can continue...

RReverser:
Actually I think I can make my script safer to check securecookie as well and not trivialize ruleset when securecookie wants to match non-literal domains.

RReverser:
Done here: https://github.com/EFForg/https-everywhere/pull/12294. Should cover most of the rewrites from this PR too, but leaving rulesets with any securecookie untouched for now for backward compatibility.

PR #12141, opened 2017-08-27
Create osf22p3lmweopgho.onion.xml

Bisaloo:
#12122


J0WI:
Is there any proof that they are managed by the same owner?

Bisaloo:
As for all onion services, they use an EV cert, is that enough?

PR #12145, opened 2017-08-27
Update hsts-prune/index.js

### LINKED ISSUE #12139: hsts-prune is still pointed to the discontinued Firefox Aurora release
The preload list for Firefox Aurora [hasn't been updated since April](https://hg.mozilla.org/releases/mozilla-aurora/log/tip/security/manager/ssl/nsSTSPreloadList.inc). We'll need to point hsts-prune to Firefox Nightly instead, and make changes to support the new format of `nsSTSPreloadList.inc`.

cschanaj:
ping @Hainish should fix #12139 but sort of dirty.

Travis build failed for unknown reason...

### DIFF
--- a/utils/hsts-prune/index.js
+++ b/utils/hsts-prune/index.js
@@ -71,6 +71,9 @@ const parse_include = include_url => {
       })
       .pipe(split())
       .on('data', line => {
+        line = line.replace(new RegExp('/ 1 /g'), ' true ')
+        line = line.replace(new RegExp('/ 0 /g'), ' false ')
+
         let regex_res = line.match(regex)
         if(regex_res){
           hsts[regex_res[1]] = Boolean(regex_res[2])
@@ -236,7 +239,7 @@ async.parallel({
   versions.esr_major = versions.esr.replace(/_.*/, "");
 
   let stable_url = `https://hg.mozilla.org/releases/mozilla-release/raw-file/FIREFOX_${versions.stable}_RELEASE/security/manager/ssl/nsSTSPreloadList.inc`;
-  let dev_url = `https://hg.mozilla.org/releases/mozilla-aurora/raw-file/tip/security/manager/ssl/nsSTSPreloadList.inc`;
+  let dev_url = `https://hg.mozilla.org/releases/mozilla-beta/raw-file/tip/security/manager/ssl/nsSTSPreloadList.inc`;
   let esr_url = `https://hg.mozilla.org/releases/mozilla-esr${versions.esr_major}/raw-file/FIREFOX_${versions.esr}_RELEASE/security/manager/ssl/nsSTSPreloadList.inc`;
   let chromium_url = `https://chromium.googlesource.com/chromium/src.git/+/${versions.chromium}/net/http/transport_security_state_static.json?format=TEXT`;
 


koops76:
@cschanaj So, Developer Edition is a rebrand of Beta? Interesting.

cschanaj:
Multiple source from Mozilla suggest so. Anything wrong?

- https://wiki.mozilla.org/Firefox/Aurora
- https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Source_Code/Mercurial

jeremyn:
Firefox Developer Edition is a rebrand of Aurora according to https://developer.mozilla.org/en-US/Firefox/Developer_Edition :

> Firefox Developer Edition replaces the Aurora channel in the [Firefox Release Process](https://wiki.mozilla.org/Release_Management/Release_Process).

However the last modified date for `mozilla-aurora` at https://hg.mozilla.org/releases is `2017-05-15` which is too old, with `mozilla-release` having a date of `2017-08-26`. I don't know where else to find the Developer Edition source though. Where is it?

koops76:
@jeremyn The version number is identical to Beta.

jeremyn:
Oh okay, thanks. It looks like Mozilla actually dropped Aurora according to https://hacks.mozilla.org/2017/04/simplifying-firefox-release-channels/ :

> Starting on April 18, the Firefox Aurora channel will stop updating, and over the course of the next several months, the Aurora build will be removed from the train release cycle. [Developer Edition](https://developer.mozilla.org/en-US/Firefox/Developer_Edition) will be based on the Beta build.

I wonder if "based on" means that Mozilla modifies something in the beta source before releasing it as Developer Edition.

koops76:
@jeremyn They probably do. There do not seem to be a source specifically for Developer Edition anywhere. I think we should ask Mozilla.

cschanaj:
According to the article [What is Firefox Nightly?](https://wiki.mozilla.org/Nightly#What_is_Firefox_Nightly.3F), Firefox Aurora is a release in-between the Beta and Nightly versions. It is a policy decision we have to made to choose which version to use (since `hsts-prune` ONLY remove `target` when they are preloaded in all supported browser, simultaneously) . 

Personally, I prefer replacing Aurora with Beta since it is always intended to be the next Release (while Nightly could be _the Release after the next Release_). This guarantees `hsts-prune` remove `target` if and only if the current and next Release preloaded it.

bardiharborow:
I've added some thoughts to #12139.

cschanaj:
@bardiharborow both Beta and Nightly use the new format currently, and the modified code handle both formats. If @Hainish decides that we should use Nightly insteads, I can replace `mozilla-beta` with `mozilla-central` in the URI. thanks!

bardiharborow:
@cschanaj either this did not have the intended effect, or something has changed since this was merged.

The following code injected into `utils/hsts-prune/index.js`:

```javascript
console.log(Object.keys(structs.esr).length, Object.keys(structs.dev).length, Object.keys(structs.stable).length, Object.keys(structs.chromium).length);
```

Results in:

```
25431 0 25431 32798
```

cschanaj:
@bardiharborow a breaking change has been introduced in [3849a13bc8ce](https://hg.mozilla.org/releases/mozilla-beta/rev/3849a13bc8ce) in `2017-09-15` as per https://bugzilla.mozilla.org/show_bug.cgi?id=1382001. We will need to modify our script in order to parse the list correctly.

PR #8748, opened 2017-02-24
DuckDuckGo Onion Service

e5f3c2ea895af0f27667:

PR #12150, opened 2017-08-27
DuckDuckGo Onion Service (#8748)

koops76:


src/chrome/content/rules/3g2upl4pq6kufc4m.onion.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Other DuckDuckGo rulesets:
+		- DuckDuckGo.xml
```
J0WI: Please update the reference list in `DuckDuckGo.xml`

koops76: Done.

Issue #12151, opened 2017-08-27
Trying to add new rule, get 500 error


terrorist96:
I'm trying to alter this rule:
https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/torrentdownloads.me.xml
(https://github.com/EFForg/https-everywhere/edit/master/src/chrome/content/rules/torrentdownloads.me.xml)
to add:
```
    <target host="torrentdownloads.co" />
    <target host="www.torrentdownloads.co" />
```

But I keep getting a 500 error. I reported to Github, but in the meantime, can someone else try to make this edit?
Thanks.

koops76:
#12152

terrorist96:
Thanks!

PR #12152, opened 2017-08-27
Update torrentdownloads.me.xml

koops76:
#12151


terrorist96:
@koops76 for some reason navigating to http://www.torrentsdownload.co/ doesn't redirect to https even though this rule was updated. Do you think why?

J0WI:
This rule is only for `torrentsdownloads.co` (plural).

terrorist96:
@J0WI ah, thanks! That's confusing. Going to `torrentdownloadS.co` redirects to `torrentSdownload.co` so this should be updated.

J0WI:
You're welcome to open an issue or a PR :)

PR #12153, opened 2017-08-27
Update USA_Today.xml and delete USA_Today.com-problematic.xml

koops76:
#11114, #12122.


src/chrome/content/rules/USA_Today.xml
```diff
@@ -1,114 +1,38 @@
...
-	² Secured by us
-	³ Some secured by us; others unsecurable <= 503
-	⁴ Unsecurable <= 503
+	Mismatched:
+		- ee (a248.e.akamai.net)
+		- reg.e (gannett.com)
+		- onlinestore (*.nexternal.com)
 
+	Redirect to HTTP:
+		- college
```
J0WI: Wildcard DNS

koops76: @J0WI Done.

PR #12160, opened 2017-08-28
Use Map instead of Object for RuleSets.targets

cschanaj:
@Hainish this will notably improve performance and memory usage, as suggested in https://github.com/EFForg/https-everywhere/issues/1775#issuecomment-151655897, whereas the performance is measured by https://jsperf.com/es6-map-vs-object-properties/42. 

~Why `Firefox` test always fails?~

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -175,7 +175,7 @@ RuleSet.prototype = {
  */
 function RuleSets(ruleActiveStates) {
   // Load rules into structure
-  this.targets = {};
+  this.targets = new Map();
 
   // A cache for potentiallyApplicableRulesets
   this.ruleCache = new Map();
@@ -213,12 +213,12 @@ RuleSets.prototype = {
     var new_rule_set = new RuleSet(params.host, true, "user rule");
     var new_rule = new Rule(params.urlMatcher, params.redirectTo);
     new_rule_set.rules.push(new_rule);
-    if (!(params.host in this.targets)) {
-      this.targets[params.host] = [];
+    if (!this.targets.has(params.host)) {
+      this.targets.set(params.host, []);
     }
     this.ruleCache.delete(params.host);
     // TODO: maybe promote this rule?
-    this.targets[params.host].push(new_rule_set);
+    this.targets.get(params.host).push(new_rule_set);
     if (new_rule_set.name in this.ruleActiveStates) {
       new_rule_set.active = (this.ruleActiveStates[new_rule_set.name] == "true");
     }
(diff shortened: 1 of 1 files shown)


chromium/rules.js
```diff
@@ -328,9 +333,9 @@ RuleSets.prototype = {
 
     var tmp;
```
cschanaj: BTW, this variable is not being used. Should I remove it in this PR?

koops76: @cschanaj I will remove unused variables in one of my PRs.

Hainish:
Checking on the `firefox-dev` test now.

Hainish:
Despite a [note to the contrary](https://wiki.mozilla.org/Add-ons/Extension_Signing#FAQ), it looks like the developer edition is now strictly enforcing xpi signatures.  I've updated the https://github.com/EFForg/https-everywhere-docker-base URL to point to the unbranded edition of Firefox Dev, so once that is done building I can re-run the tests and we should be good.

koops76:
@Hainish Nightly is unbranded by default and does not enforce extension signing.

koops76:
@Hainish Nightly is unbranded by default and does not enforce extension signing.

cschanaj:
> To disable signature checks, you will need to set the xpinstall.signatures.required preference to "false". 

@Hainish It seems that you'd have to modify `test/script.py` to disable signature checking in the Developer Edition. Did you test it?

Hainish:
@cschanaj this is already set by https://github.com/EFForg/https-everywhere/blob/master/test/firefox/test_profile_skeleton/prefs.js#L10

Hainish:
@cschanaj in the latest Firefox versions this pref does not change the behavior as they've made signature checking manditory.

cschanaj:
@Hainish okay. would you also review and merge this PR?

Hainish:
@cschanaj yes but it'll have to wait until at least tomorrow

cschanaj:
@Hainish never mind, but I would like this to be merged before the next release given the performance boost it will provides. thanks.

Hainish:
This is great.  Efficiency can be further improved by utilizing a [radix trie](https://en.wikipedia.org/wiki/Radix_tree), indexed by the string reversal of the target (moc.elpmaxe) but this is contingent upon #11203

cschanaj:
I agree a radix trie should be the best structure for the domains, but we will have to wait #11203 is completed. So I guess using Map could be a easy fix in the short term. :smile:

koops76:
@Hainish Maybe use a bloom filter or a prefix list?

cschanaj:
AFAIK, a bloom filter gives probabilistic results. And thus using a bloom filter might require a major rewrite.

Hainish:
@koops76 bloom filters are only appropriate solutions when false positives are acceptable.  That's not true here - we don't want to DoS a site unintentionally!

koops76:
@Hainish We can verify positives by hashes or simply a list of acceptable targets.

cschanaj:
And a bloom filter give boolean results. The searching of applicable results is still inefficient.

cschanaj:
@Hainish thanks! I will work on using JSON instead of XML for default.rulesets in a later PR.

PR #12165, opened 2017-08-28
Gone from DNS (MarsHut.net.xml)

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


jeremyn:
`marshut.net` and `www.marshut.net` are not gone from DNS.

PR #12172, opened 2017-08-29
[Vook.com] Update and rename rule to Pronoun.com.xml.

bardiharborow:
`vook.com` now redirects to `pronoun.com`. `store.vook.com` and `tool.vook.com` are dead.


src/chrome/content/rules/Pronoun.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+
+	Problematic hosts in *pronoun.com:
+
+		- blog (expired)
+		- support ᵐ
```
J0WI: + `buy.vook.com`

bardiharborow:
`vook.com` has starting timing out. I'll revisit this PR in a week or so and see what the situation is.

bardiharborow:
@J0WI I've dropped `vook.com`, and we should be good to go.

PR #12173, opened 2017-08-29
[VTAC.edu.au] Create ruleset.

bardiharborow:


src/chrome/content/rules/VTAC.edu.au.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="VTAC.edu.au">
+
+	<target host="vtac.edu.au" />
+	<target host="www.vtac.edu.au" />
```
J0WI: This site contains mixed content issues.

bardiharborow: Hmm. It's a current application period, so they keep messing with the site and those sort of issues seem to keep coming and going. I've marked the ruleset as mixedcontent.

PR #12179, opened 2017-08-29
Delete EBay_static.com-problematic.xml

Bisaloo:
Duplicate of [EBay_static.com.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/EBay_static.com.xml)


J0WI:
The `exclusion pattern` differs

PR #12187, opened 2017-08-29
[0bin.net] Update ruleset.

bardiharborow:
`www` issue no longer exists.


bardiharborow:
Something went wrong with my testing here.

PR #12204, opened 2017-08-30
Create Al-Miftah.com.xml

ihave100openPRs:


jeremyn:
Both `^` and `www` have an invalid certificate, and I can't find any other domains, so there's no ruleset to make here. So, I think this should be closed.

PR #12205, opened 2017-08-30
Create DMV.org.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


src/chrome/content/rules/DMV.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="DMV.org">
+	<target host="dmv.org" />
+	<target host="www.dmv.org" />
+	<target host="local.dmv.org" />
```
J0WI: + `static.dmv.org` (please also add a valid test url for it)

PR #12207, opened 2017-08-30
Create Backdrift.org.xml

christianheinrichs:


christianheinrichs:
According to the Travis CI build at https://travis-ci.org/EFForg/https-everywhere/jobs/269904180#L620 there are "not enough tests".

Help on how to fix this would be appreciated

J0WI:
Can you please add a comment here and in `Backpack.tf.xml` to show the relationship between those files?

christianheinrichs:
@J0WI Sorry for getting so late back to you, but what exactly do you mean? AFAIK, there is no relation between `backdrift.org` and `backpack.tf`.

J0WI:
Oh okay, I don't know why I thought there is a relation.

PR #12208, opened 2017-08-30
Create Omdbapi.com.xml

christianheinrichs:


src/chrome/content/rules/OMDb_API.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="OMDb API">
+	<target host="omdbapi.com" />
+	<target host="www.omdbapi.com" />
```
J0WI: + `img.omdbapi.com`

christianheinrichs: @J0WI Done.

PR #12223, opened 2017-08-30
Create 3refe.com.xml

ihave100openPRs:


jeremyn:
- [ ] Add `platform="mixedcontent"`

ihave100openPRs:
@jeremyn  Thanks for the observation (I didn't add it at the time since only images were affected, but seems that it changed), fixed now!

PR #12244, opened 2017-08-30
Create Muslim_Aid.org.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


src/chrome/content/rules/Muslim_Aid.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Muslim Aid.org">
+	<target host="muslimaid.org" />
+	<target host="www.muslimaid.org" />
+	<target host="qurbani.muslimaid.org" />
```
J0WI: `www.ramadan.muslimaid.org`

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@J0WI I tried it earlier  but it didn't work even with HTTP alone, should I include it nevertheless?

J0WI:
I can confirm that `www.ramadan.muslimaid.org` does not exist, so you can ignore this. But `ramadan.muslimaid.org` seems to be there.

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@J0WI Added

PR #12269, opened 2017-08-30
Delete merge-rulesets.js

koops76:
merge-rulesets.py runs on Windows anyway.


cowlicks:
LGTM the script is not referenced anywhere, and is likely broken.

PR #12275, opened 2017-08-31
[Heroku] Remove outdated exclusion and rule (fix travis)

Bisaloo:
Removing lines made obsolete after https://github.com/EFForg/https-everywhere/commit/8bda7ce0a40d208a7e752e4df56e32b403160071.

### DIFF
--- a/src/chrome/content/rules/Heroku.xml
+++ b/src/chrome/content/rules/Heroku.xml
@@ -17,7 +17,7 @@
 
 	<securecookie host="^[\w\-]+\.herokuapp\.com$" name=".+" />
 
-	<rule from="^http://([\w-]+)\.herokuapp\.com/"
-		to="https://$1.herokuapp.com" />
+	<rule from="^http://([\w-]+\.)?herokuapp\.com/"
+		to="https://$1herokuapp.com" />
 
 </ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Heroku.xml
```diff
@@ -7,15 +7,16 @@
...
+		<test url="http://voices.herokuapp.com/" />
 
 	<!--	Avoiding cross-domain cookies.
 						-->
 	<securecookie host="^[\w\-]+\.herokuapp\.com$" name=".*" />
 
-
-	<rule from="^http://([^/:@\.]+\.)?heroku(app)?\.com/"
-		to="https://$1heroku$2.com/" />
+	<rule from="^http://([^/:@\.]+\.)?herokuapp\.com/"
```
J0WI: Can you please avoid this complex regex?

Bisaloo: @J0WI, this will cause invalid cert warnings on `www.*.herokuapp.com` but it's just a page saying "There's nothing there". Does it matter?

Bisaloo: Or did you just mean 

```xml
    <rule from="^http://([\w-]+)\.herokuapp\.com/"
        to="https://$1.herokuapp.com" />
```

?

J0WI:
Thanks!

PR #12277, opened 2017-08-31
Update devtools-panel.html

koops76:


koops76:
Just in case default text color is white.

koops76:
ping @Hainish @cowlicks.

PR #12279, opened 2017-08-31
No unused vars

koops76:
ping @Hainish.

### DIFF
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -1,9 +1,7 @@
 {
   "env": {
     "browser": true,
-    "commonjs": true,
     "es6": true,
-    "node": true,
     "webextensions": true
   },
   "parserOptions": {
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -10,6 +10,7 @@
   },
   "rules": {
     "indent": ["error", 2],
-    "no-unused-vars": "error"
+    "no-unused-vars": "error",
+    "no-undef": "error"
   }
 }
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -11,6 +11,7 @@
     "sourceType": "script"
   },
   "rules": {
-    "indent": [2, 2]
+    "indent": ["error", 2],
+    "no-unused-vars": "error"
   }
 }
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -1,10 +1,7 @@
 "use strict";
 // Stubs so this runs under nodejs. They get overwritten later by util.js
-var VERB=1;
-var DBUG=2;
-var INFO=3;
-var NOTE=4;
-var WARN=5;
+var VERB = 1;
+var NOTE = 4;
 function log(){}
 
 // To reduce memory usage for the numerous rules/cookies with trivial rules
@@ -331,7 +328,6 @@ RuleSets.prototype = {
     }
     log(DBUG, "Ruleset cache miss for " + host);
 
-    var tmp;
     var results = [];
     if (this.targets.has(host)) {
       // Copy the host targets so we don't modify them.
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -1,7 +1,8 @@
 "use strict";
 // Stubs so this runs under nodejs. They get overwritten later by util.js
-var VERB = 1;
-var NOTE = 4;
+var DBUG = 2;
+var INFO = 3;
+var WARN = 5;
 function log(){}
 
 // To reduce memory usage for the numerous rules/cookies with trivial rules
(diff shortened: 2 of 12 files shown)


koops76:
@cowlicks See if there's more global or exported variables, they will be needed in the future.

chromium/.eslintrc.json
```diff
@@ -1,16 +1,16 @@
 {
   "env": {
     "browser": true,
-    "commonjs": true,
     "es6": true,
-    "node": true,
```
cowlicks: Was node supported at some point?

chromium/rules.js
```diff
@@ -1,10 +1,12 @@
+/* global enableMixedRulesets */
+/* global domainBlacklist */
+/* global exports */
+
 "use strict";
 // Stubs so this runs under nodejs. They get overwritten later by util.js
```
cowlicks: If node isn't supported. Can we remove DBUG, INFO, etc entirely from this file?

chromium/.eslintrc.json
```diff
@@ -1,16 +1,16 @@
...
     "webextensions": true
   },
   "parserOptions": {
     "ecmaVersion": 2017,
     "sourceType": "script"
   },
   "rules": {
-    "indent": [2, 2]
+    "indent": ["error", 2],
```
cowlicks: what does this change?

chromium/.eslintrc.json
koops76: https://github.com/EFForg/https-everywhere/blob/master/rewriter/rewriter.js

chromium/.eslintrc.json
koops76: `"indent": [2, 2]` is not a valid value.

chromium/rules.js
koops76: I don't think so.

chromium/.eslintrc.json
koops76: I removed `commonjs` and `node` from `env` list to make `exported` and `global` work.

cowlicks: Could you add `/* eslint-env node, commonjs, etc  */` to the files where it is appropriate?

chromium/.eslintrc.json
```diff
@@ -1,16 +1,16 @@
...
+    "no-unused-vars": "error",
+    "no-undef": "error"
   }
```
cowlicks: Do you think it would be easier to move some of the globals into the eslintrc like:
```javascript
  "globals": {
    "DBUG"...
```
I'm not sure what the exact format is supposed to looks like, but I think eslint supports it. [docs here](https://eslint.org/docs/user-guide/configuring).

chromium/rules.js
cowlicks: Since this file gets run under node, maybe it would be a good candidate for a file level `/*eslint-env node*/`.

koops76: No, it would make ESLint ignore `/* exported */` directives.

chromium/.eslintrc.json
koops76: No, it would make ESLint ignore `/* exported */` directives.

chromium/.eslintrc.json
koops76: It will cause an error in `util.js` since we will be redefining the already existing global.

chromium/.eslintrc.json
cowlicks: Sorry I'm not very familiar with eslint. Doesn't this file only apply this configuration  to sub-directories of `chromium/`? Since `rewriter.js` isn't in this directory, and also doesn't have any exported directives, why can't we add `/* eslint-env node */` to it?

chromium/rules.js
cowlicks: I'm confused, why do we need to stub *some* global vars like DBUG, INFO, and WARN. But not others like `enableMixedRulesets`, and `domainBlacklist`?

cowlicks: cc @koops76

chromium/.eslintrc.json
koops76: I only changed files under `chromium` in this PR.

chromium/rules.js
koops76: Because only `rules.js` is required by rewriter.

koops76:
@cowlicks Can you tell me all changes I need to make?

chromium/rules.js
cowlicks: I see that. It looks like `rewriter.js` does not end up calling code that needs `domainBlacklist` and `enabledMixedRulesets`, so it doesn't break. Regardless, I don't think this PR would cause problems with `rewriter.js` since the removed `NOTE`, `VERB` etc are not needed in this file.

cowlicks:
@koops76 no more changes needed.

I have some concerns about how un-DRY some parts of this code are, but that is out of scope of this PR.

koops76:
@cowlicks Merge?

koops76:
🎉

PR #12280, opened 2017-08-31
Create www.2chan.net.xml

christianheinrichs:
Partially addresses issue #3069.

### DIFF
--- a/src/chrome/content/rules/www.2chan.net.xml
+++ b/src/chrome/content/rules/www.2chan.net.xml
@@ -1,0 +1,14 @@
+<ruleset name="www.2chan.net">
+	<target host="www.2chan.net" />
+	<target host="cgi.2chan.net" />
+	<target host="dat.2chan.net" />
+	<target host="dec.2chan.net" />
+	<target host="may.2chan.net" />
+	<target host="img.2chan.net" />
+	<target host="ipv6.2chan.net" />
+	<target host="jun.2chan.net" />
+	<target host="nov.2chan.net" />
+	<target host="zip.2chan.net" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
--- a/src/chrome/content/rules/2chan.net.xml
+++ b/src/chrome/content/rules/2chan.net.xml
@@ -1,8 +1,15 @@
 <!--
+	502 status code:
+		- mail.2chan.net
 
 	Insecure cookies are set for these domains:
 		- .2chan.net
 
+	Server not found:
+		- 2chan.net
+
+	Unable to connect:
+		- test.2chan.net
 -->
 <ruleset name="2chan.net">
 	<target host="www.2chan.net" />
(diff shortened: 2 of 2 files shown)


src/chrome/content/rules/www.2chan.net.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="www.2chan.net">
```
I-have-50-open-PRs-to-HTTPS-EVERYWHERE: Don't name the ruleset "www.2chan.net.xml" rather "2chan.net.xml"

src/chrome/content/rules/www.2chan.net.xml
```diff
@@ -0,0 +1,14 @@
...
+	<target host="cgi.2chan.net" />
+	<target host="dat.2chan.net" />
+	<target host="dec.2chan.net" />
+	<target host="may.2chan.net" />
+	<target host="img.2chan.net" />
+	<target host="ipv6.2chan.net" />
+	<target host="jun.2chan.net" />
+	<target host="nov.2chan.net" />
+	<target host="zip.2chan.net" />
+
```
I-have-50-open-PRs-to-HTTPS-EVERYWHERE: Please add <securecookie host=".+" name=".+" />

christianheinrichs:
Thanks for your input!

I am aware that the www prefix in the rules directory is not the way to go when it comes to creating new rulesets, so the only reason for that ruleset name was because 2chan.net is unreachable which can be confirmed by executing `ping 2chan.net` which results in:

> ping: 2chan.net: No address associated with hostname

However, it's fixed now.

Also thank you for reminding a new guy of the `securecookie` tag. The uuc cookie on the subdomains containing cgi, dat, dec, may, img, ipv6, jun, nov and zip should now be secure thanks to this. The _cfduid, posttime cookie on .2chan.net is not secured though.

src/chrome/content/rules/www.2chan.net.xml
christianheinrichs: @I-have-50-open-PRs-to-HTTPS-EVERYWHERE 

Since there doesn't seem to be a way to secure the __cfduid and posttime cookie, should I replace:

```xml
```

With this?

```xml
<securecookie host="^cgi\.2chan\.net$" name=".+" />
<securecookie host="^dat\.2chan\.net$" name=".+" />
<securecookie host="^dec\.2chan\.net$" name=".+" />
<securecookie host="^img\.2chan\.net$" name=".+" />
<securecookie host="^ipv6\.2chan\.net$" name=".+" />
<securecookie host="^jun\.2chan\.net$" name=".+" />
<securecookie host="^may\.2chan\.net$" name=".+" />
<securecookie host="^nov\.2chan\.net$" name=".+" />
<securecookie host="^zip\.2chan\.net$" name=".+" />
```

src/chrome/content/rules/www.2chan.net.xml
christianheinrichs: Thanks for your input!

I am aware that the **www** prefix in the rules directory is not the way to go when it comes to creating new rulesets, so the only reason for that ruleset name was because 2chan.net is unreachable which can be confirmed by executing `ping 2chan.net` which results in:

> ping: 2chan.net: No address associated with hostname

However, it's fixed now.

Issue #12297, opened 2017-08-31
Regexps matching host that is not in targets


RReverser:
These cases were found as part of https://github.com/EFForg/https-everywhere/pull/12294. Either regexp got outdated or host was forgotten to be added to `<target />` list, either way they might be functioning incorrectly:

- [x] [ABC-Online.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ABC-Online.xml): `^http://(?:www\.)?abccomercial\.com/` matches `abccomercial.com`, `www.abccomercial.com` which are not in targets `shop.abc.net.au`, `abccommercial.com`, `*.abccommercial.com`
- [x] [ARM.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/ARM.xml): `^http://(?:www\.)?arm\.com/(?=css/|images/)` matches `arm.com` which are not in targets `*.arm.com`
- [ ] [AkademikerForsakring.se.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/AkademikerForsakring.se.xml): `^http://(?:www\.)?akademikerf[oö]rs[aä]kring\.se/` matches `akademikerforsäkring.se`, `akademikerförsakring.se`, `www.akademikerforsäkring.se`, `www.akademikerförsakring.se` which are not in targets `akademikerforsakring.se`, `www.akademikerforsakring.se`, `akademikerförsäkring.se`, `www.akademikerförsäkring.se`
- [ ] [AlJazeera.com.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/AlJazeera.com.xml): `^http://(m\.)?aljazeera\.com/` matches `m.aljazeera.com` which are not in targets `aljazeera.com`, `www.aljazeera.com`, `staff.balkans.aljazeera.com`, `interactive.aljazeera.com`, `leaderssummit.aljazeera.com`, `staff.liberties.aljazeera.com`, `mediaview.aljazeera.com`, `syhacked.aljazeera.com`, `iforgot.aljazeera.net`
[...]

PR #12300, opened 2017-08-31
[PureFTPd.org] Fix broken rule and add targets

Bisaloo:
#12297


Bisaloo:
@Hainish, how come the tests didn't run here?

PR #12302, opened 2017-09-01
Add normalize-securecookie.js

koops76:


koops76:
@Hainish Please actually run the script yourself.

utils/normalize-securecookie.js
```diff
@@ -0,0 +1,151 @@
...
+      if (err) {
+        reject(err)
+      } else {
+        resolve()
+      }
+    })
+  })
+}
+
+function readdir (path) {
```
RReverser: Just a nit / hint: if you're targetting only the latest Node versions (which I think makes sense anyway for scripts, especially given that we use ES6), you can just do `const readdir = util.promisify(fs.readdir);`. (`util.promisify` is a recently added built-in exactly for wrapping Node.js APIs into Promise ones).

koops76: @RReverser Done.

Hainish:
Thanks! Looks good. I'll add this to the `rules` test as well.

PR #12294, opened 2017-08-31
[Spring cleaning #2] Trivialize even more rules

RReverser:
This is a follow-up to https://github.com/EFForg/https-everywhere/pull/4400.

I've updated the previously submitted script to "explode" (analyze) regular expressions and see if it results in a small list of static URLs that can be trivialized. When it does, hosts of these URLs are extracted into targets and rules are replaced with a single `<rule from="^http:" to="https:" />`.

Note: rulesets with `<securecookie />` are skipped for now because trivializing
those can lead to certain hosts being removed from targets even though cookies
should still be secured there. In future, we can analyze securecookie as well
to detect static matches and trivialize more rulesets.

PR #12310, opened 2017-09-01
Trivialize some securecookie rulesets too

RReverser:
A follow-up to https://github.com/EFForg/https-everywhere/pull/12294.

Figured that we can also safely rewrite rulesets with `securecookie` if list of targets wasn't changed at all - turns out this reveals another 138 rulesets that can be trivialized. cc @Hainish

### DIFF
--- a/utils/trivialize-rules/trivialize-rules.js
+++ b/utils/trivialize-rules/trivialize-rules.js
@@ -84,6 +84,9 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) => 
       let result = `[${tagName}] ${chalk.bold(name)}: ${strings[0]}`;
       for (let i = 1; i < strings.length; i++) {
         let value = values[i - 1];
+        if (value instanceof Set) {
+          value = Array.from(value);
+        }
         value = Array.isArray(value) ? value.join(', ') : value.toString();
         result += chalk.blue(value) + strings[i];
       }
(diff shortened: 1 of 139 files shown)


RReverser:
@cschanaj @Bisaloo @Hainish Does anyone want to merge this?

utils/trivialize-rules/trivialize-rules.js
```diff
@@ -150,42 +151,48 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) =>
...
-  info`trivialized`;
-
-  domains = Array.from(new Set(domains));
+  domains = Array.from(domains);
 
   if (domains.slice().sort().join('\n') !== targets.sort().join('\n')) {
+    if (ruleset.securecookie) {
+      return;
+    }
+
```
Hainish: I'm not following the logic here.  I see this was removed from the top.  I'm assuming you're using a `Set` now for its property of having unique values.  That would make sense if you were, for instance, lumping all the `target`s into the same `Set` with the `securecookie` hosts and then checking coverage by the `rule` via this exploded `domains`.  But I don't see that happening above.  How did this result in actually checking coverage of the `securecookie`s?

Hainish: I can see this now.

This is for when we have a left-wildcard in `target`s, and the `rule` contains multiple subdomains.  In this case, the `target`s have to be expanded to match the newly trivialized `rule`.  But this is complicated if we have a `securecookie`.

So in this more complicated case, we can abort.

I suppose we can we do something like we did with the `rule` tag and explode the regex, seeing if only the expanded `target`s are contained in the `securecookie` regex.  But this is a good iterative step.

RReverser: The only difference with previous PR is that the securecookie rule was relaxed a bit. Previously I was bailing out if ruleset contains any securecookie item, but now also walking through cases where exploded regexp doesn't change list of targets, but can be still trivialized - then it's safe to rewrite such ruleset even though it has securecookie.

So yes, it's just another iteration step that is still guaranteed to be backwards-compatible.

RReverser: The `Set` changes are unrelated - they are only for better warning messages (previously some hosts were repeated several times in one message).

PR #12311, opened 2017-09-01
Configurable counter color

koops76:
ping @Hainish because everyone tells me how bright the color is for them, but I like it as it is.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -60,22 +60,26 @@ var switchPlannerInfo = {};
  *  {
  *    httpNowhere: Boolean,
  *    showCounter: Boolean,
- *    isExtensionEnabled: Boolean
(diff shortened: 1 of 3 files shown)


Hainish:
I don't think we should make this configurable.  If people are distracted by it they can turn it off.

koops76:
@Hainish Disagree. The counter is an important source of information whether any requests have been rewritten. Forcing users who find it annoying to disable it rather than allowing them to change its color is bad UX.

koops76:
I like the current color a lot, but many people said previously that it's distracting, especally on dark background. I think it should be configurable. The current color is the same that is used in Privacy Badger, but it seems many want it to be the same as their extensions', such as uBlock Origin, which badge is grey.

Issue #4369, opened 2016-03-14
Forbid `*` in the middle of the target host


RReverser:
As @jsha pointed in https://github.com/EFForg/https-everywhere/issues/2981#issuecomment-194468323:

> Agreed we want to remove internal \* in target hosts, and remove it from support in the rewriter.

So decided to create an issue for 1) removal of support for this in rewriter code 2) having tests that check against such `<target host />` occurences.

RReverser:
Looks like after https://github.com/EFForg/https-everywhere/issues/2981 was fixed by https://github.com/EFForg/https-everywhere/pull/3603, some new cases appeared:

```
src/chrome/content/rules/Argonne-National-Laboratory.xml:	<target host="www.*.anl.gov" />
src/chrome/content/rules/General-Electric.xml:	<target host="files.*.geblogs.com"/>
src/chrome/content/rules/GPLHost.xml:	<target host="dtc.*.gplhost.com"/>
src/chrome/content/rules/Monster.xml:	<target host="hiring.*.monster.com" />
src/chrome/content/rules/OnSugar.xml:	<target host="secure.*.onsugar.com" />
src/chrome/content/rules/OnSugar.xml:	<target host="www.*.onsugar.com" />
src/chrome/content/rules/Projectplace.com.xml:	<target host="projectplace.*.dimelo.com" />
src/chrome/content/rules/Right-Media.xml:	<!--target host="tm.*.dp.yieldmanager.net" /-->
src/chrome/content/rules/Truenudists.com.xml:	<target host="cdn.*.truenudists.com" />
src/chrome/content/rules/United-States-Department-of-Energy.xml:	<target host="www.*.doe.gov" />
src/chrome/content/rules/University-of-Alaska.xml:	<target host="www.*.alaska.edu" />
src/chrome/content/rules/University-of-Bern.xml:	<target host="www.*.unibe.ch" />
src/chrome/content/rules/University-of-Idaho.xml:	<target host="www.*.uidaho.edu" />
src/chrome/content/rules/University-of-Southampton.xml:	<target host="www.*.soton.ac.uk" />
src/chrome/content/rules/University_of_Houston.xml:	<target host="www.*.uh.edu" />
src/chrome/content/rules/University_of_Maine.xml:	<target host="www.*.umaine.edu" />
src/chrome/content/rules/US-Dept-of-Veterans-Affairs.xml:	<target host="www.*.vaforvets.va.gov" />
[...]

PR #12317, opened 2017-09-01
Fix wildcard-in-the-middle in General-Electric.xml

RReverser:
See https://github.com/EFForg/https-everywhere/issues/4369


src/chrome/content/rules/General-Electric.xml
```diff
@@ -1,6 +1,6 @@
 <ruleset name="General Electric (partial)">
 
-	<target host="files.*.geblogs.com"/>
+	<target host="*.geblogs.com"/>
```
J0WI: Could you also add some (we require at least 3) test urls and a hint, that the blogs itself do not support HTTPS.

RReverser: Doesn't rule itself give such a hint? Its regexp clearly covers only files.*.geblogs.com.

Bisaloo: This is the last PR needed before we can forbid wildcards in the middle. @RReverser, are you interested in addressing @J0WI's comments (ie add `test url`s and a comment *why* other subdomains do not support HTTPS)?

If you don't have time / don't want to work on this, I can take over.

Bisaloo:
@J0WI, I would really like to move forward and finally block wildcards in the middle in our tests. Do you think it's okay if I edit this branch? Otherwise, I can also open a brand new PR for this.

J0WI:
@Bisaloo since @RReverser is not responding I would say it's okay to take it.

src/chrome/content/rules/GEBlogs.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	Invalid certificate:
+		*.files.geblogs.com
+		files.*.geblogs.com
+
+	Other subdomains are mismatched or refused.
```
J0WI: @Bisaloo could you add some more documentation about the behavior of this host?
E.g. why a wildcard is required, but only specific patterns are rewritten.

src/chrome/content/rules/GEBlogs.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	Invalid certificate:
+		*.files.geblogs.com
+		files.*.geblogs.com
```
J0WI: `*.geblogs.com` are also mismatching, but see comment below.

src/chrome/content/rules/GEBlogs.com.xml
Bisaloo: How would you describe this? I don't know how to explain it better than what I did.

J0WI:
@Bisaloo would you like to finish this?

J0WI:
There are two wildcards in the `Invalid certificate` which says more or less the same as `Other subdomains are mismatched or refused.`. But there is also wildcard target, which makes it even more unclear. `geblogs.com` itself is not documented at all.

I would add a short description, that the service provides 2nd level subdomains for its customers and an additional `files` 3rd level subdomain for static content. For their own static content they use a subdomain with `files` in the 2nd and the category in the 3rd level.

Maybe it also helps to ad their "main" site?
```
ge.com
www.ge.com
benefits.ge.com
www.benefits.ge.com
ssologin.corporate.ge.com
ssousflogin.corporate.ge.com
digitalsupport.ge.com
go.digital.ge.com
eddie.ge.com
integrity.ge.com
libraries.ge.com
myapps.ge.com
mypassword.ge.com
mytech.ge.com
onehr.ge.com
www.onehr.ge.com
geaeck.registrar.ge.com
ssocentralck.registrar.ge.com
sc.ge.com
supportcentral.ge.com
travel.ge.com
```

J0WI:
@Bisaloo this seems to be the last bit in https://github.com/EFForg/https-everywhere/issues/4369#issuecomment-326727385. Would you like to finish this?

pipboy96:
:tada:

Issue #12232, opened 2017-08-30
Fix WebExtension Excessive Memory Usage


cschanaj:
Type: other

Re-open of #1775, based on https://github.com/EFForg/https-everywhere/issues/1775#issuecomment-151655897 ordered by priority (personal)

## GOAL

???

## PROPOSAL

Proof of concept PRs, may or may not be get implemented for various reasons. Discussions should be made in their respective PRs.

- Try using a Prefix Tree for RuleSets.targets (#12530)
- ~Try using a LRU Cache for caching (#12591)~ See #3952

### Core
- [x] Use Map instead of Object for RuleSets.targets (#12160) (2017.8.31)
    * improve performance
- [x] Use JSON for default.rulesets instead of XML (#12273) (2017.9.12)
    * reduce the startup time to below 1.5 seconds
    * reduce the memory usage to below 100mb (on Chromium task manager on startup after GC).
    * reduce the size of default.rulesets from 8.4mb to 6.4mb and thus smaller extension size
- [x] Use Map instead of Object for redirectCounter and Fix Memory Leak (#12422) (2017.9.12)
    * fix a small memory leak for long lasting browsing session
    * improve performance
- [x] Use null instead of Empty Set for in RuleSets.ruleCache (#12494, #12577) (2017.10.24)
    * avoid unnecessary memory allocation
- [x] Merge exclusions into a single pattern in default.rulesets (#16092, related #15448) (2018.7.8)
    * reduce memory overhead
- [x] Trivialize rulesets with _static_ `securecookie`, thanks @RReverser  (#16029, #16378) (2018.9.19)
    * removed 300+ wildcard targets, improve lookup speed and memory usage
    * related #16540, #16634 
- [x] Merge securecookies which match any cookie.name in each rulesets (2019.5.2)
    * Refactor and enhance trivialize-cookie-rules.js (#17438, #17578)
    * Update utils/merge-rulesets.py (#17492)
    * Remove/ Update some securecookie with no coverage (#17565)
- [x] Drop support for middle wildcard targets in potentiallyApplicableRulesets (2019.5.2)
    * Forbid and remove wildcards in the middle (#4369, #12319 thanks @Bisaloo)
    * Remove middle wildcard support in rules.js (#17715)
    * reduce runtime overhead (lookup all_rules) significantly
- [x] Triviailize rulesets with right-wildcard targets and trivialize rewrites (#17855, #18068) (2019.6.27)
[...]

PR #12320, opened 2017-09-01
Report and remove overlapping targets

RReverser:
I've noticed many XML have targets that actually overlap and so unnecessarily duplicate each other.

I've extended target checks to test and report this automatically as reducing them might partially help with https://github.com/EFForg/https-everywhere/issues/12232.

### DIFF
--- a/src/chrome/content/rules/3min.xml
+++ b/src/chrome/content/rules/3min.xml
@@ -1,7 +1,6 @@
 <ruleset name="3min" default_off="Invalid Certificate">
   <target host="3min.de"/>
   <target host="*.3min.de"/>
-  <target host="www.3min.de"/>
 
   <securecookie host="^(?:.*\.)?3min\.de$" name=".*" />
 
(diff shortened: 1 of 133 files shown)


RReverser:
Hmm, this is too broad and doesn't take into account that `*` in the end is treated differently than in the beginning. Let me fix it.

RReverser:
Added automated attempt to remove targets that were detected as already covered by others in the same ruleset.

As I said in https://github.com/EFForg/https-everywhere/issues/12322#issuecomment-326631920, I'm not sure if this will always give correct result because sometimes it's the wildcard target that can be avoided, but I think it might be better for the start and can be corrected manually after.

RReverser:
(feel free to merge the check separately from XML changes if that's easier / makes more sense)

src/chrome/content/rules/3min.xml
```diff
@@ -1,7 +1,6 @@
 <ruleset name="3min" default_off="Invalid Certificate">
   <target host="3min.de"/>
   <target host="*.3min.de"/>
-  <target host="www.3min.de"/>
```
J0WI: In some cases we cold replace the target with a test url.

Hainish:
Reviewing this now.

Hainish:
Looks great, thanks again!

PR #12326, opened 2017-09-01
Make chromium.sh work for non-debian distros

cowlicks:
Instead of checking for a distro to determine if we want `chromium-browser` or `chromium` we just check if `which chromium-browser` works, if not we use `chromium`.


cowlicks:
@Hainish this seems pretty harmless so I'm merging.

PR #12364, opened 2017-09-02
Update Optagelse.dk.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


J0WI:
Please also search for subdomains here.

PR #12367, opened 2017-09-02
Update P6R.com.xml

ihave100openPRs:


src/chrome/content/rules/P6R.com.xml
```diff
@@ -1,9 +1,8 @@
 <ruleset name="P6R.com">
-
 	<target host="p6r.com" />
 	<target host="www.p6r.com" />
```
J0WI: `support.p6r.com`

ihave100openPRs:
@J0WI Added

PR #12380, opened 2017-09-02
Update background.js (fixes #12377)

### LINKED ISSUE #12377: Credentials in URL are not Restored
## Step to reproduce

1. Visit http://user:password@example.com:403/path with "Block all unencrypted connection"

2. Expect URL should be rewrote to https://user:password@example.com:403/path

3. Actual URL is https://null:null@example.com:403/path

## Reason

Seemingly wrong scoping in [`chromium/background.js#L323`](https://github.com/EFForg/https-everywhere/blob/master/chromium/background.js#L323)

koops76:


cschanaj:
@koops76 the variable is being used again in [`chromium/background.js#L367`](https://github.com/EFForg/https-everywhere/blob/master/chromium/background.js#L367). Thus the variables have to be declared outside the if statement, see #12381

koops76:
@cschanaj Already done.

chromium/background.js
```diff
@@ -384,7 +386,7 @@ function onBeforeRequest(details) {
     // failing.
     if (shouldCancel) {
       if (!newuristr) {
-        return {redirectUrl: canonical_url.replace(/^http:/, "https:")};
+        return {redirectUrl: details.url.replace(/^http:/, "https:")};
```
cschanaj: IMHO, this is change is not necessary, better not to change the existing behavior.

chromium/background.js
```diff
@@ -320,12 +320,14 @@ function onBeforeRequest(details) {
   // If there is a username / password, put them aside during the ruleset
   // analysis process
   var using_credentials_in_url = false;
+  let tmp_user;
```
cschanaj: var for consistence

chromium/background.js
koops76: This is actually why they are replaced with null but not replaced back.

chromium/background.js
koops76: `let` for ES6.

chromium/background.js
koops76: @cschanaj Even better now, bug is fixed but the existing behavior is kept.

chromium/background.js
cschanaj: not even consistent in the same function ...

koops76: Will be when we will run ESLint with new settings.

cschanaj:
lot more changes than I have expected. I am not going to review this. ping Hainish when you are done, thanks anyway.

koops76:
ping @Hainish.

chromium/background.js
```diff
@@ -361,12 +363,18 @@ function onBeforeRequest(details) {
...
-    // re-insert userpass info which was stripped temporarily
-    const uri_with_credentials = new URL(newuristr);
-    uri_with_credentials.username = tmp_user;
-    uri_with_credentials.password = tmp_pass;
-    newuristr = uri_with_credentials.href;
+  if (using_credentials_in_url) {
+    const canonical_url_with_credentials = new URL(canonical_url);
+    canonical_url_with_credentials.username = tmp_user;
+    canonical_url_with_credentials.password = tmp_pass;
+    canonical_url = canonical_url_with_credentials.href;
```
Hainish: It should be safe to add this as an `else` clause to the conditional immediately below.  If `newuristr` is defined, this mutation of `canonical_url` is simply unused.

This is a minor point and shouldn't block merging, but it makes things a bit easier to follow with the matching `if (!newuristr)` below.

PR #12381, opened 2017-09-02
Update chromium/background.js, Fix #12377

### LINKED ISSUE #12377: Credentials in URL are not Restored

cschanaj:


koops76:
Same changes in #12380, close please.

cschanaj:
well...

PR #12393, opened 2017-09-02
[3min] Delete expired domain.

bardiharborow:
Unsure if this is expired, or just has no working nameservers, but in any case it's very dead.


J0WI:
@bardiharborow could you please rebase your branch?

bardiharborow:
@J0WI good to go.

PR #12402, opened 2017-09-02
Create Homosexuelle-Arbeitsgruppen-Bern.xml

koops76:


src/chrome/content/rules/GayAgenda.lgbt.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="GayAgenda.lgbt">
+	<target host="gayagenda.lgbt" />
+	<target host="www.gayagenda.lgbt" />
```
J0WI: See also:
```
3gang.ch
cominginn.ch
gaybern.ch
ha-bern.ch

koops76: @J0WI Quite a lot of domains since each of them may have subdomains.

koops76: @J0WI Done.

PR #12414, opened 2017-09-03
Create GenderJam.lgbt.xml

koops76:


src/chrome/content/rules/GenderJam.lgbt.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="GenderJam.lgbt">
+	<target host="genderjam.lgbt" />
+	<target host="www.genderjam.lgbt" />
```
J0WI: See also `genderjam.org.uk`

koops76: @J0WI Done.

PR #12422, opened 2017-09-03
Use Map instead of Object for redirectCounter and Fix Memory Leak

cschanaj:
## Step to reproduce the memory leak

1. From chrome://extensions, open the "Inspect views" for HTTPS Everywhere in a new session.
2. In the console, type `redirectCounter`. the console should output something like

```
Object {} // Empty Object
```

3. Visit https://www.nytimes.com/
4. Repeat Step 2, observe 

```
Object {440: 1, 444, 1}
```
5. Repeat Step 3-4, observe `redirectCounter` keep growing without sign of removing previous items
```
Object {440: 1, 444: 1, 598: 1, 601: 1, 632: 1, 635: 1}
```

This affect long lasting browsing sessions. 

**UPDATE** On Mixed Content pages, HTTP request are blocked by MCB from the browser. It is necessary to listen on both HTTP & HTTPS connection to resolve the memory leak (Assuming no "Block all unencrypted requests"). thanks.

P.S. I expected browser to handle redirect loops automatically, however, I am not really sure about the case when an extension rewrite URLs.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -284,7 +284,7 @@ var domainBlacklist = new Set();
 
 // redirect counter workaround
 // TODO: Remove this code if they ever give us a real counter
-var redirectCounter = {};
+var redirectCounter = new Map();
 
 /**
  * Called before a HTTP(s) request. Does the heavy lifting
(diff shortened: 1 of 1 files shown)


cschanaj:
@Hainish this fix a memory leak for long lasting browsing sessions, please review and merge this when you have the time. thanks.

Hainish:
Thank you!

PR #12427, opened 2017-09-03
Create QSpace.lgbt.xml

koops76:


src/chrome/content/rules/QSpace.lgbt.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="QSpace.lgbt">
+	<target host="qspace.lgbt" />
+	<target host="www.qspace.lgbt" />
```
J0WI: See also `qspacearch.com`

koops76: @J0WI Done.

PR #12429, opened 2017-09-03
Create TheRoot.com.xml

Bisaloo:


src/chrome/content/rules/TheRoot.com.xml
```diff
@@ -0,0 +1,28 @@
...
+-->
+<ruleset name="The Root">
+
+	<target host="theroot.com" />
+	<target host="www.theroot.com" />
+	<target host="login.theroot.com" />
+	<target host="mms.theroot.com" />
+	<target host="onehundred.theroot.com" />
+
+	<!-- Blogs. Each user gets their own subdomain? -->
```
J0WI: Confirm. They also have a wildcard DNS record. How about a wildcard?

PR #12453, opened 2017-09-04
[Numergy.com] Simplify and reactivate ruleset

Bisaloo:
#10843


Bisaloo:
Because login.numergy times out on port 80, the tests fail. What should I do? Simply remove it? No one will be able to access it over `http` anyways. Or whitelist this ruleset?

PR #12465, opened 2017-09-05
Update Python.org.xml

koops76:


src/chrome/content/rules/Python.org.xml
```diff
@@ -1,68 +1,76 @@
...
 
-	<securecookie host="^(bugs|cheeseshop|discuss|openbadges\.es|jobs|mail|packages|packaging|pl|planet|speed|staging|staging2|status|testpypi|warehouse|warehouse-staging|)\.python\.org$" name=".+" />
+	<securecookie host=".+" name=".+" />
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http:" to="https:" />
 
+	<!-- HTTP 403 at / -->
+	<exclusion pattern="^http://svn\.python\.org/$" />
```
J0WI: Please add a valid test url/keep the old one

src/chrome/content/rules/Python.org.xml
```diff
@@ -1,68 +1,76 @@
...
+		- hg
 
-	No working URL known:
-		* front.python.org
+	Incomplete certificate chain:
+		- hg.es
+		- lists.es
 
-	Refused:
-		* buildbot.python.org
```
J0WI: please keep this comment

src/chrome/content/rules/Python.org.xml
koops76: The test URL is already here, implicitly. Each target has an implicit test URL of `http://${domain}/`.

src/chrome/content/rules/Python.org.xml
koops76: @J0WI Done.

src/chrome/content/rules/Python.org.xml
J0WI: Yes, but you just excluded this implicit test url. Now we have no tests for `svn.python.org` at all.

koops76: @J0WI Done.

src/chrome/content/rules/Python.org.xml
koops76: @J0WI Can you review as many PRs by me as possible? I won't be available for any corrections for 2 weeks.

PR #12469, opened 2017-09-05
[Zerve.com] Fix expired certificate issue

Bisaloo:


src/chrome/content/rules/Zerve.xml
```diff
@@ -1,30 +1,14 @@
...
-	Problematic subdomains:
-
-		- ^			(cert only matches www)
-		- img-static.cdn	(cloudfront)
-
-
-	Some pages redirect to http.
+	Expired certificate:
+		zerve.com
+		www.zerve.com
```
J0WI: Same for `qa.zerve.com`. Let's see if they fix it before next release.

src/chrome/content/rules/Zerve.xml
```diff
@@ -1,30 +1,14 @@
...
-	<target host="zerve.com" />
-	<target host="*.zerve.com" />
-
-
-	<rule from="^http://(?:www\.)?zerve\.com/(cookie_check\.php|(?:directory/)?(?:css|img)/|raa/raa3_2\.php)"
-		to="https://www.zerve.com/$1" />
+	<target host="img-static.cdn.zerve.com" />
 
 	<rule from="^http://img-static\.cdn\.zerve\.com/"
 		to="https://d3ks0ua4wrpv20.cloudfront.net/" />
```
J0WI: Do we have a test url for this?

Bisaloo: I don't

PR #12470, opened 2017-09-05
Create WhoisIp.ovh.xml

koops76:


src/chrome/content/rules/WhoisIp.ovh.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="WhoisIp.ovh">
+	<target host="whoisip.ovh" />
+	<target host="www.whoisip.ovh" />
+	<target host="ssl.whoisip.ovh" />
```
J0WI: `cdn.where.ovh`

koops76: @J0WI Done.

PR #11528, opened 2017-07-30
Add rulesets for various thepiratebay.org proxies

koops76:
#11481, #11478.

Ping @J0WI.

PR #12481, opened 2017-09-06
[thephoenixbay.com] Update and reactivate ruleset.

bardiharborow:
1. Pulled the latest list from the main site.
2. Pruned all targets that failed the fetch test.
3. Added a `www` target for each domain.
4. Pruned all `www` targets that failed the fetch test.
5. Prune duplicate and preloaded targets.


J0WI:
#11528 tries to split them into different files. As you can see there, not all of them have exact the same subdomains/behavior.

bardiharborow:
@J0WI This mainly touches `thephoenixbay.com.xml`, so I've just reworked the edits to remove duplicated targets from `thephoenixbay.com.xml` rather than touching `proxybay.one.xml`. We may want to rewrite these as individual rulesets later, but this will at least get a chunk of domains secured again.

src/chrome/content/rules/thephoenixbay.com.xml
```diff
@@ -1,623 +1,357 @@
...
+        - unblockbay.co (no https)
+        - unblockbay.link (tls error)
+        - unblockbay.pw (mismatch)
+        - unblockbay.xyz (mismatch)
+        - unblockthepiratebay.net (timeout)
+        - unblockthepiratebay.org (timeout)
+        - unblocktpb.co (no https)
+        - unblocktpb.link (dns dead)
+        - unblocktpb.pw (expired)
+        - unblocktpb.xyz (self-signed)
```
J0WI: You can remove all "dns dead" hosts from the list.

koops76:
This should be split into individual files for each domain prior to merging. I can do it automatically and make a PR to your fork with all of them split out.

bardiharborow:
@koops76 I guess I saw that as a lot of effort and maintenance burden for what are mostly just rapidly outdated throwaway domains, but if that's the way you'd prefer to do it feel free to shoot me a PR. :)

koops76:
@bardiharborow See bardiharborow#1.

koops76:
@bardiharborow See Travis.

bardiharborow:
Travis is currently experiencing data center issues, and that may be contributing to the false positive timeouts.

bardiharborow:
@J0WI This should be mergeable now. I've had to remove 5 rulesets which pass locally, but are consistently failing Travis. I'll put in a separate PR in a moment.

J0WI:
Thanks!

PR #12489, opened 2017-09-06
Rename and update Enhancedsteam.com.xml

christianheinrichs:


jeremyn:
@christianheinrichs Are you still interested in working on this pull request?

christianheinrichs:
@jeremyn Sure, are there subdomains missing or is there something else to fix?

jeremyn:
What you have is basically correct but I would make some (mostly style) changes.

- [ ] Use this with filename `EnhancedSteam.com.xml` :
```xml
<!--
    Invalid certificate:
        translation.enhancedsteam.com
  
--> 
<ruleset name="EnhancedSteam.com">
    <target host="enhancedsteam.com" />
    <target host="www.enhancedsteam.com" />
    <target host="api.enhancedsteam.com" />
    <target host="firefox.enhancedsteam.com" />
 
</ruleset>
```

Please let me know if you have any questions.

christianheinrichs:
@jeremyn Done.

jeremyn:
Your new file doesn't match the one I have in https://github.com/EFForg/https-everywhere/pull/12489#issuecomment-349998359, in particular your version has extra comments and a `securecookie` tag.

christianheinrichs:
@jeremyn The file should now have the requested changes applied properly.

I do have to ask what the rationale behind removing the `redirect` comments is.

I would also like to know why you don't want to keep the `securecookie` tag, when there is at least 2 cookies on Enhanced Steam (`__cfduid` and `SERVERID`) which don't have the secure flag set to true by default and neither are secured by HSTS.

jeremyn:
We don't want extra comments because they require additional maintenance and might be confusing to some people, The redirect comments aren't necessary because they don't matter for the ruleset, and anyone who wants to check the site behavior can do that for themselves. Also in this case the comments are incorrect, because I can search for `site:api.enhancedsteam.com` or `site:firefox.enhancedsteam.com` on Google and find results that don't redirect as you described. Maybe only the root URLs https://api.enhancedsteam.com and https://firefox.enhancedsteam.com do that.

`.+`/`.+` `securecookie`s can cause problems when `^` is a `target` and we don't rewrite every subdomain, like here. See the discussion in https://github.com/EFForg/https-everywhere/issues/13762. If you feel strongly about including a `securecookie`, you can do that but limit the `host` to `^(www|api|firefox)\.enhancedsteam\.com$`. Please let me know whether you want to do that.

christianheinrichs:
I am starting to understand your concerns regarding the whole `securecookie` issue and you certainly bring forward quite a few good arguments. The `.+` wildcard in that context can indeed cause breakage of certain websites.

Although the security gain (if there is one at all) is minimal, I have decided to include the `securecookie` tag in my newest commit with your requested host limitation changes.

jeremyn:
Okay, that's fine. Merged.

PR #12491, opened 2017-09-06
Create ESEA.net.xml

christianheinrichs:


src/chrome/content/rules/ESEA.net.xml
```diff
@@ -0,0 +1,17 @@
...
+		- esea.net → play.esea.net
+		- news.esea.net → play.esea.net
+-->
+<ruleset name="ESEA.net">
+	<target host="esea.net" />
+	<target host="www.esea.net" />
+	<target host="fragshack.esea.net" />
+	<target host="mdl.esea.net" />
+	<target host="news.esea.net" />
+	<target host="play.esea.net" />
```
J0WI: + `static.esea.net` (please also add a valid test url for it)

christianheinrichs: @J0WI Done.

src/chrome/content/rules/ESEA.net.xml
```diff
@@ -10,6 +10,8 @@
 	<target host="mdl.esea.net" />
 	<target host="news.esea.net" />
 	<target host="play.esea.net" />
+	<target host="static.esea.net" />
+		<test url="http://static.esea.net/" />
```
J0WI: This test url returns a 403 status code. Please change to something that response with 200.

christianheinrichs: @J0WI The newest commit on this PR now contains a test URL which should return a 200 status code. From what I understand, the `static.esea.net` is probably exclusively used for the hosting of .cfg files.

PR #12493, opened 2017-09-07
Normalize securecookie results

Hainish:
This PR will likely increase CI running time due to the new `node` docker image requirement.

Since as of this PR `node` is required both when only rulesets are changed *and* when core code changes have been made, it would be faster to simply build node support into https://github.com/EFForg/https-everywhere-docker-base/, since that will be downloaded regardless.

Nonetheless I don't have time to make this change and test it currently, so for the time being we'll have to live with the increased running time.  This would make a good PR for anyone interested.


Hainish:
Rel: #12302

PR #12494, opened 2017-09-08
Use null instead of new Set in RuleSets.ruleCache

cschanaj:
ping @Hainish for a review and merging. thanks!

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -403,44 +403,61 @@ RuleSets.prototype = {
    */
   potentiallyApplicableRulesets: function(host) {
     // Have we cached this result? If so, return it!
-    var cached_item = this.ruleCache.get(host);
-    if (cached_item !== undefined) {
-      log(DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.length);
+    if (this.ruleCache.has(host)) {
+      let cached_item = this.ruleCache.get(host);
+      let len = cached_item ? cached_item.size : 0;
+      log(DBUG, "Ruleset cache hit for " + host + " items:" + len);
       return cached_item;
-    }
-    log(DBUG, "Ruleset cache miss for " + host);
-
-    var tmp;
-    var results = [];
-    if (this.targets.has(host)) {
-      // Copy the host targets so we don't modify them.
-      results = results.concat(this.targets.get(host));
+    } else {
+      log(DBUG, "Ruleset cache miss for " + host);
     }
 
     // Ensure host is well-formed (RFC 1035)
-    if (host.indexOf("..") != -1 || host.length > 255) {
+    if (host.length > 255 || host.indexOf('..') != -1) {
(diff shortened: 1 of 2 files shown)


J0WI:
cc @Hainish

chromium/rules.js
```diff
@@ -403,45 +403,62 @@ RuleSets.prototype = {
...
       return null;
     }
 
+    // Let's search!!
+    let resultSet = new Set();
+
+    // Look for exact matches
+    if (this.targets.has(host)) {
+      let results = this.targets.get(host);
+      if (results) {
```
cowlicks: I think you can drop the `if (this.targets.has(host)) {` above since you do this `if (results) {` check here.

chromium/rules.js
```diff
@@ -403,45 +403,62 @@ RuleSets.prototype = {
...
+        for (let result of results) {
+          resultSet.add(result);
```
cowlicks: A more concise, and readable  way to write this would be like:
```javascript
results.forEach(result => resultSet.add(result));
```

chromium/rules.js
cowlicks: ohhh wait, `results` can be `null` now hmm.

cowlicks:
Removing all the empty sets everywhere is good. However changing the API to return `null` means we have to check to see if result of `potentiallyApplicableRulesets` everywhere.

`null` is a good replacement because it is a singleton, so using it means we don't copy data unnecessarily.  But `null` is bad replacement because it isn't iterable.

What we want is an empty iterator singleton. So instead of `null` I think we could use something like this: 
```javascript
const nullIterable = Object.create(null, {
  [Symbol.iterator]: {
    value: function* () {
      },
    },
});
```

chromium/rules.js
```diff
@@ -403,45 +403,62 @@ RuleSets.prototype = {
...
       segmented[i] = tmp;
     }
+
     // now eat away from the left, with *, so that for x.y.z.google.com we
     // check *.z.google.com and *.google.com (we did *.y.z.google.com above)
-    for (var i = 2; i <= segmented.length - 2; ++i) {
-      var t = "*." + segmented.slice(i,segmented.length).join(".");
-      results = results.concat(this.targets.get(t));
+    for (let i = 2; i <= segmented.length - 2; ++i) {
+      let domain  = "*." + segmented.slice(i, segmented.length).join(".");
```
cowlicks: This is unrelated to this PR, but just wanted to note for posterity. I think an algorithm for finding matching rulesets would be much simpler than this if they rulesets were structured as a prefix tree, where the domains are split into labels and reversed (assuming we can remove the RHS wildcards).

cschanaj:
I'm going to submit a new PR using `nullIterable` soon. thanks!!

cowlicks:
closing in favor of #12577

PR #12495, opened 2017-09-08
Create nv.ua.xml

Folant:


jeremyn:
I'll look at this but please first go through and clean up some of the basic stuff we've discussed (indentation, names).

jeremyn:
Remove entirely:
- [ ] https://mx1.nv.ua (this is too internal/technical to include)

Mark as `Different HTTP/HTTPS content` and remove their `target`s:
- [ ] https://biz.dev.nv.ua
- [ ] https://m.dev.nv.ua
 
On the main page https://nv.ua there is some mixed content blocking (MCB) which hides an ad, which is not a big deal, but the MCB also hides a little graphic link to http://magazine.nv.ua in the upper right. We ordinarily don't want to block this sort of thing, but in this case I don't think it's important enough to use `mixedcontent` or add special handling for the main page.

jeremyn:
Thanks, merged.

PR #12504, opened 2017-09-10
Update .editorconfig

christianheinrichs:


jeremyn:
@christianheinrichs We don't have any `*.pl` (Perl) files that I know of, so please remove that part. Otherwise this looks fine.

@Hainish Can you please double-check that these editor settings look good to you?

christianheinrichs:
@jeremyn The Perl section has been removed, although it's worth mentioning that there is 1 Perl script located [here](https://github.com/EFForg/https-everywhere/blob/master/test/rules/certs_transitive_closure/c_rehash).

EditorConfig doesn't detect the Perl file without the .pl extension anyway

jeremyn:
@christianheinrichs Good catch! I still think we should remove the .pl section though, as you did. We don't want anyone writing new Perl scripts here.

J0WI:
LGTM, let's wait for @Hainish to approve this.

Hainish:
Looks good - thanks @christianheinrichs 

We should probably do a rewrite of https://github.com/EFForg/https-everywhere/blob/master/test/rules/certs_transitive_closure/c_rehash to Python at some point.  This is a short file, doesn't seem too complicated, and we want language consistency with the rest of the project.

jeremyn:
@Hainish I thought the same thing about a rewrite, but it looks like this is a (re?)implementation of a common `c_rehash` application from OpenSSL, for example compare to https://opensource.apple.com/source/OpenSSL/OpenSSL-22/openssl/tools/c_rehash.auto.html.

The whole ruleset checker at `test/rules/` includes many things that you would expect the platform to include, such as the `c_rehash` utility, and a zillion browser root certificates. I suppose it was prepared to be highly portable. Instead of rewriting this `c_rehash` from Perl and otherwise maintaining all these dependencies in the repository indefinitely, we should limit the available supported platforms for the checker so we can rely on system dependencies.

PR #12512, opened 2017-09-10
re-enable bger.ch.xml

J0WI:


jeremyn:
- [ ] https://www.bger.ch seems fine, so make a `target` for it and remove it from the top comment
- [ ] https://expert.bger.ch has different HTTP/HTTPS content, so remove its `target` and add it to the top comment
- [ ] Remove `(partial)` from the `ruleset` `name`

J0WI:
I want to keep the `(partial)` for now, because we can't secure the majority of the site.

jeremyn:
I think `(partial)` has two good uses.

One use is to tell a user that the page they are visiting right now is only partly covered. For example if they are visiting `www`, which makes a lot of requests to `static`, and we only cover `www` but not `static`, `(partial)` tells the user that HTTPS Everywhere is only partly securing that page.

The other use is when for whatever reason we don't want to list or cover all the subdomains for a site. For example maybe some site has a ton of subdomains, and we want to cover one or two really important ones but we don't want to worry about covering or even listing the others. In other words this is an explicit pass on Google/Sublist3r subdomain checking.

Neither of these cases apply here because we've provided broad `bger.ch` coverage and `www` pages seem entirely covered, so I think we should get rid of `(partial)`.

(The reason I'm arguing about this is because I don't want to have to make judgment calls about whether we're providing enough coverage to justify `(partial)` for any particular ruleset.)

jeremyn:
Thanks, merged.

PR #12513, opened 2017-09-10
add MCB rule for 20min.ch

J0WI:


jeremyn:
https://m.20min.ch is redirected to https://m.20min.ch/error but http://m.20min.ch works okay. 

http://m.20min.ch/sport/dossier/europaleague/story/12323607 renders normally but https://m.20min.ch/sport/dossier/europaleague/story/12323607 gets a sort of soft time out, where there are loading dots but the actual content never seems to appear.

https://www.20min.ch and its subpages generally redirect to a plain HTTP version of the page with `?httpsredirect` appended to the URL, for example https://www.20min.ch/sport/ redirects to http://www.20min.ch/sport/?httpsredirect . The HTTPS Everywhere rewrite on the `?httpsredirect` URL seems okay, but this behavior is a little unusual.

I haven't exhaustively examined these domains but so far I don't really trust that either `www` or `m` are safe rewrites even allowing for mixed content blocking, which means that the entire `20min.ch-mixedcontent.xml` ruleset is invalid. What do you think?

J0WI:
Those issues were related to a CORS bug in `20min.ch.xml`.
Since the MCB rule will be disabled by default and explicitly listed in the section "Experimental rules: May cause warnings or breakage. Disabled by default." it should be safe to merge this.

jeremyn:
I don't know what you mean by the CORS

J0WI:
Let's try this.

J0WI:
@jeremyn this is finally passing all tests.

jeremyn:
`20min.ch-mixedcontent.xml`:

This is fine as it is.

`20min.ch.xml`:
- [ ] The `exclusion` and `test` for `api` should have `http`, not `https`.
- [ ] For the `api` `CORS issues` comment, give more information about where to go to see the CORS problem happen
  
Just FYI, for the problems I mentioned in https://github.com/EFForg/https-everywhere/pull/12513#issuecomment-341582722 : https://m.20min.ch/ doesn't redirect to https://m.20min.ch/error for me anymore, and I also don't a "soft time out" for https://m.20min.ch/sport/dossier/europaleague/story/12323607 . I do still get the `?httpsredirect`, but that's not as bad.

J0WI:
I don't see any major mixed content issues, so I just removed the exclusion.

jeremyn:
Thanks, merged.

PR #12517, opened 2017-09-10
re-enable MtSystems.ch.xml

J0WI:


jeremyn:
- [ ] Both https://mtsystems.ch and https://www.mtsystems.ch redirect directly to https://www.mtsystems.com so the rewrite [`rule`](https://github.com/J0WI/https-everywhere/blob/2209a4d8c47347d3c22b7cb90fa7b33dfe2da1e6/src/chrome/content/rules/MtSystems.com.xml#L15-L16) should just be

```xml
	<rule from="^http://(www\.)?mtsystems\.ch/"
		to="https://www.mtsystems.com/" />
```

J0WI:
Isn't this a snappy redirect then?

jeremyn:
Do you mean, am I proposing a snapping redirect? No, I'm not, because the site itself directly redirects `(www.)mtsystems.ch` to `www.mtsystems.com`. It does not redirect `mtsystems.ch` to `mtsystems.com` like the existing rule does, even as an intermediate step. In other words the existing rule incorrectly introduces a redirect that the site itself doesn't do.

J0WI:
Okay, I've changed it.

PR #12518, opened 2017-09-10
Create claymath.org.xml

Folant:


src/chrome/content/rules/claymath.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="claymath.org">
```
J0WI: Please add a hint that this host has a wildcard DNS

PR #12527, opened 2017-09-12
Upaiyun.com.xml: one more exclusion

lilydjwg:


jeremyn:
Thanks for your pull request.

- [ ] Please fix the sort order by putting `jbcdn2` directly after `honggc` in both the `exclusion` regex and the `test` tags.

lilydjwg:
I've fixed the order.

jeremyn:
Thanks, merged.

PR #12530, opened 2017-09-12
Try using a prefix tree for RuleSets.targets

cowlicks:
This is just here for discussion. I haven't benchmarked it yet. NB: this is not a radix trie.

### DIFF
--- a/chromium/tree.js
+++ b/chromium/tree.js
@@ -1,0 +1,73 @@
+// split a hostname and reverse it
+function splitter(dottedName) {
+  return dottedName.split('.').reverse();
+}
+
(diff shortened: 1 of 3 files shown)


chromium/tree.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+let Node = Map;
+
+function Tree() {
+  this._base = new Node();
+}
+
+Tree.prototype = {
+  sentinel: '.', // since '.' should not appear in the label of a hostname
```
cowlicks: We should probably just use a `.data` property on `Node` instead of this.

PR #12536, opened 2017-09-13
Remove Google-mismatches.xml

Bisaloo:


src/chrome/content/rules/Google-mismatches.xml
```diff
@@ -1,27 +0,0 @@
...
-	
-	Problematic domains:
-
-		- (www.)apture.com	(works, mismatched, CN: *.google.com)
-
--->
-<ruleset name="Google (mismatches)" default_off="mismatched">
-
-	<!--	Akamai	-->
-	<target host="js.admeld.com"/>
```
J0WI: `admeld.com` and `www.admeld.com` can be secured.

Bisaloo: https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Admeld.xml

PR #12539, opened 2017-09-13
Remove duplicated target gdata.youtube.com

Bisaloo:
Already covered by the wildcard in [YouTube.xml](https://github.com/EFForg/https-everywhere/blob/release/src/chrome/content/rules/YouTube.xml).


J0WI:
Could you move the test url?

J0WI:
YouTube.xml does not pass our tests.

PR #12544, opened 2017-09-13
Remove GoogleShopping.xml

Bisaloo:
I know this one is hard to review but I tried to really split every step in a single commit, so it should be okay if you review commit by commit.

I can split it in several PR if needed.


Bisaloo:
I know this one is hard to review but I tried to really split every step in a single commit, so it should be okay if you review commit by commit.

Bisaloo:
#11203

Issue #12087, opened 2017-08-25
Use Ed25519 instead of RSA for ruleset signatures.


koops76:
- Higher security with much smaller key size.
- Possibly faster
- Used in respected open source projects ([1](https://en.wikipedia.org/wiki/EdDSA#Software), [2](https://en.wikipedia.org/wiki/Curve25519#Popularity)).

There is already a library that can run in a browser: https://github.com/dchest/tweetnacl-js.

koops76:
ping @Hainish.

jeremyn:
Where in the code do we use cryptographic signatures? I've tried to answer this question with some searches in GitHub, but couldn't.

koops76:
@jeremyn https://github.com/EFForg/https-everywhere/tree/sign-rulesets.

jeremyn:
Thanks @koops76 . 

@Hainish By the way, I disagree with the entire approach of out-of-band updates for rulesets. See my comments in https://github.com/EFForg/privacybadger/issues/1466#issuecomment-315792144 and following for Privacy Badger for why.

koops76:
@jeremyn In turn, I disagree with you, but let's save this for another time. Let's discuss this issue instead.

Hainish:
@koops76 I'd rather use the WebCrypto API than some library, even if it has been publicly audited.  The longevity of crypto APIs implemented directly in the browser seems better assured.  We also need something which is supported by `openssl`, given the signature scripts in `util`:
https://github.com/EFForg/https-everywhere/blob/sign-rulesets/utils/sign-rulesets-standalone.sh
https://github.com/EFForg/https-everywhere/blob/sign-rulesets/utils/sign-rulesets-request.sh
https://github.com/EFForg/https-everywhere/blob/sign-rulesets/utils/sign-rulesets-airgap.sh

ECDSA might work, and also has a shorter key size.

jeremyn:
In the gist https://gist.github.com/tqbf/be58d2d39690c3b366ad by the highly respected security expert Thomas Ptacek, under "Asymmetric signatures" he says

> Use Nacl, Ed25519, or RFC6979.

which supports @koops76 's point. On the other hand it's hard to argue with wanting to use the API included in the browser.

koops76:
@Hainish Signing can be done by a Node script. Do not trade security for using built-in browser features.

koops76:
@Hainish I believe this should be decided either by public discussion or by EFF's crypto experts.

[...]

PR #12548, opened 2017-09-14
Fixes for small bugs and annoyances

cowlicks:
46bc17f fixes the icon text so it now say `HTTPS Everywhere`, `HTTPS Everywhere (blocking)`, or `HTTPS Everywhere (disabled)`. 

720fd78 ensures we only set the badge background color on the correct tab, this isn't really a bug since we only use one color anyway.

869b31f fixes the an error on start up

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -142,10 +142,9 @@ function updateState () {
     if (!tabs || tabs.length === 0) {
       return;
     }
-
     const activeCount = getActiveRulesetCount(tabs[0].id);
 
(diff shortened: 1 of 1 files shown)


koops76:
Change `RegExp` to `new RegExp` at L240 please. I don't want to create a PR for such a minor change.

cowlicks:
@koops76 done thank you

koops76:
@cowlicks Also I would like you to read these two comments: https://github.com/EFForg/https-everywhere/issues/12087#issuecomment-325833464, https://github.com/EFForg/https-everywhere/issues/12087#issuecomment-326403988.

PR #12552, opened 2017-09-14
Let's use the same badge color as uBlock Origin.

koops76:


koops76:
ping @Hainish @cowlicks

Hainish:
This looks good to me - grey doesn't catch the eye as harshly as the previous green does.

PR #12555, opened 2017-09-14
Delete GoogleImages.xml

Bisaloo:
Depends on #12576, #12659.

Same as the previous PR. It may seem complicated but if you review commit by commit, it should be fine. I broke down every single step.

Related: #11203


Bisaloo:
Not sure what's wrong with the tests here.

J0WI:
Thanks!

PR #12556, opened 2017-09-14
Add few more missing vars to rules.js

RReverser:
Following https://github.com/EFForg/https-everywhere/pull/4298, few more variables appeared that are used by rules.js but defined later in other files.

This commit moves missing variables to rules.js so that it can be used standalone as before.

Also, changes var to const for logging levels, removed their duplications from util.js and changed the log stub to fail with error on WARN level (since in rules.js this is used mostly for reporting bad errors, so makes sense to throw when used as Node.js module). In browser, util.js will still override function with proper logging as before.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -34,7 +34,6 @@ try{
 all_rules = new RuleSets(ls);
 
 // Allow users to enable `platform="mixedcontent"` rulesets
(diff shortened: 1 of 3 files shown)


RReverser:
These conflicts are not worth it, closing.

PR #12559, opened 2017-09-14
Prefer const

koops76:
Includes changes by @RReverser.

Doesn't touch non-ES6 `var` declarations yet.

### DIFF
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -11,6 +11,7 @@
   "rules": {
     "indent": ["error", 2],
     "no-unused-vars": "error",
-    "no-undef": "error"
+    "no-undef": "error",
+    "prefer-const": "error"
   }
 }
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -1,4 +1,6 @@
-/* exported enableMixedRulesets */
+/* exported initializeStoredGlobals */
+/* exported urlBlacklist */
+/* exported all_rules */
 /* global RuleSets */
 /* global storage */
 /* global log */
@@ -6,6 +8,9 @@
 /* global NOTE */
 /* global WARN */
 /* global DBUG */
+/* global domainBlacklist */
+// eslint-disable-next-line no-unused-vars
+/* global enableMixedRulesets */
 
 "use strict";
 /**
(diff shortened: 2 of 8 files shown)


koops76:
ping @Hainish, @cowlicks.

cowlicks:
We need to have a real module system instead of specifying globals and exports everywhere. I'll make a PR with these changes ASAP.

koops76:
@cowlicks See [Webpack](https://webpack.github.io/).

koops76:
They were needed to figure out what needs to be exported/imported.

RReverser:
Webpack might be overkill here, I'd suggest going with Rollup.

koops76:
@RReverser I doubt it. Maybe you can explain why you think it is overkill?

RReverser:
Performance-wise, plus webpack has various hacks for CommonJS which extension doesn't use, plus it still includes its own `require` shims even with tree-shaking. For pure ES6 modules Rollup is faster and produces smaller output, which is why it's so often used for libraries.

cowlicks:
@RReverser @koops76 see #12586 for modules. 

@koops76 I am not very fond of this change. Can we just use `const` for global constants? The constants in `util.js` are a good case for this. Otherwise, it seems like we are using `const` to describing existing code usage devoid of context, instead of having the code describe what it should do.... Also const has 2 more letters than let.

koops76:
@cowlicks `const` prevents accidental modification of variable. It should be preferred over `let`.

RReverser:
I agree with @cowlicks, I think using const literally everywhere is noisy and, despite being popular, is rather an antipattern (especially whrn used for things that are still mutable by nature like objects). But then, it's not really up to me to decide.

koops76:
I wish there was a `prefer-let` rule, but there is no such ESLint rule. So I guess that for consistency we should prefer `const` where possible, since we can't prefer `let` instead. It is obvious that we should prefer `let` or `const` over `var` but that depends on namespace PR by @cowlicks since `var` and `let`/`const` behave differently at global scope.

cowlicks:
@koops76 closing this. There are arguments for and against using `const` everywhere, but at the end of the day, it isn't a big deal. So I'd prefer to trust devs to use their best judgement, instead of strictly enforcing a style.

koops76:
@cowlicks Will you also close a PR that adds no-var ESLint rule, requiring either `let` or `const`?

PR #12560, opened 2017-09-14
Quotes

koops76:
ping @Hainish, @cowlicks.

### DIFF
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -11,6 +11,10 @@
   "rules": {
     "indent": ["error", 2],
     "no-unused-vars": "error",
-    "no-undef": "error"
+    "no-undef": "error",
+    "quotes": ["error", "single", {
+      "avoidEscape": true,
+      "allowTemplateLiterals": true
+    }]
   }
 }
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -7,7 +7,7 @@
 /* global WARN */
 /* global DBUG */
 
-"use strict";
+'use strict';
 /**
  * Load a file packaged with the extension
  *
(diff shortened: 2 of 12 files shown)


cowlicks:
I do prefer single quotes, but I think enforcing it like this is too pedantic.

koops76:
@cowlicks Consistency is very important. Reopen this please and tell what changes, if any should be done. I am against keeping random quotes.

@Hainish Do you think we should keep the quotes as they are?

koops76:
Ah, there are too many conflicts. This may not be reopened.

koops76:
This PR is not urgent, do not reopen please.

cowlicks:
@koops76 my comments here are basically the same as #12559 

Some folks prefer `"` some prefer `'`. Forcing folks to use either `"` or `'` will give us some consistency about something that is fairly arbitrary, at the expense of annoying devs.

Personally, if I were making a pull request to an OSS project, I'd be pretty annoyed if the tests failed because I used the wrong kind of quotes that are essentially interchangable.

koops76:
@cowlicks There is such a thing as coding style. Requiring consistent code style for all code in the repository is not something crazy, or uncommon.

Hainish:
@koops76 there's also a balance to strike between enforcing style and allowing flexibility.  If the enforcement of style is too strict, it discourages participation for a benefit which is at best spurious.

koops76:
@Hainish This config is very flexible. It allows exceptions like `"'"` and allows to use template literals.
```json
{
  "quotes": ["error", "single", {
    "avoidEscape": true,
    "allowTemplateLiterals": true
  }]
}
```

koops76:
We are going to adopt StandardJS or another style guide someday.

PR #12577, opened 2017-09-15
Use nullIterable instead of new Set in RuleSets.ruleCache

cschanaj:
@cowlicks this replaces and closes #12494. thanks!

P.S. `eslint` error is unexpected, but it should be fine to fix it here.


chromium/rules.js
```diff
@@ -15,6 +15,15 @@ const trivial_rule_from_c = new RegExp("^http:");
 const trivial_cookie_name_c = new RegExp(".*");
 const trivial_cookie_host_c = new RegExp(".*");
 
+// Empty iterable singleton to reduce memory usage (#12494).
+const nullIterable = Object.create(null, {
+  [Symbol.iterator]: {
+    value: function* () {
+      // do nothing
```
koops76: This can be done in a simpler way.

```js
const nullIterator = {
  next() {
    return { done: true }
  }
}

const nullIterable = {
  [Symbol.iterator]: function () {
    return nullIterator
  }
}
```

Will also save memory.

chromium/rules.js
```diff
@@ -16,13 +16,15 @@ const trivial_cookie_name_c = new RegExp(".*");
...
-  [Symbol.iterator]: {
-    value: function* () {
-      // do nothing
-    }
+const nullIterator = {
+  next : () => { done: true }
+};
+
+const nullIterable = {
+  [Symbol.iterator]: function *() {
```
koops76: what?

koops76: `*`

koops76: Simply change it into `() => nullIterator` I guess.

cschanaj: without `*` it does not work for me.

chromium/rules.js
```diff
@@ -15,6 +15,17 @@ const trivial_rule_from_c = new RegExp("^http:");
 const trivial_cookie_name_c = new RegExp(".*");
 const trivial_cookie_host_c = new RegExp(".*");
 
+// Empty iterable singleton to reduce memory usage (#12494).
+const nullIterator = {
+  next : () => { done: true }
```
koops76: `next: () => ({ done: true })`

koops76: or `next() { return { done: true } }`

chromium/rules.js
koops76: @cschanaj See https://github.com/EFForg/https-everywhere/pull/12577/files#r139175667.

koops76: Remove the * after fixing this issue.

chromium/rules.js
cschanaj: thanks!

chromium/rules.js
koops76: Remove the * now.

chromium/rules.js
```diff
@@ -15,6 +15,17 @@ const trivial_rule_from_c = new RegExp("^http:");
 const trivial_cookie_name_c = new RegExp(".*");
 const trivial_cookie_host_c = new RegExp(".*");
 
+// Empty iterable singleton to reduce memory usage (#12494).
+const nullIterator = {
+  next : () => ({ done: true })
+};
+
+const nullIterable = {
+  [Symbol.iterator]: function *() {
```
koops76:
```js
const nullIterable = {
  [Symbol.iterator] () {
    return nullIterator;
  }
}
```

koops76:
`{*[Symbol.iterator](){}}` is the shortest code for empty iterable but it creates a new iterator each time.

chromium/rules.js
```diff
@@ -15,6 +15,17 @@ const trivial_rule_from_c = new RegExp("^http:");
 const trivial_cookie_name_c = new RegExp(".*");
 const trivial_cookie_host_c = new RegExp(".*");
 
+// Empty iterable singleton to reduce memory usage (#12494).
+const nullIterator = {
+  next : () => ({ done: true })
+};
+
+const nullIterable = {
```
koops76: Simply replace this entire code with `const nullIterable = (function* () {})()`.

chromium/rules.js
cowlicks: Having just one name is nice. How about just:
```javascript
const nullIterable = Object.create(null, {
  [Symbol.iterator]: {
    value: () => ({
      next() {
        return {done: true};
      }
    })
  }
});

```

chromium/rules.js
cowlicks: I choose to create `Object.creat(null, ...` because the created object wouldn't have extra properties, so it would be harder to misuse. This seemed important because we are using it alongside real data.

chromium/rules.js
koops76: @cowlicks `const nullIterable = (function* () {})()` would be better.

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
    */
   potentiallyApplicableRulesets: function(host) {
     // Have we cached this result? If so, return it!
-    var cached_item = this.ruleCache.get(host);
-    if (cached_item !== undefined) {
-      log(DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.length);
+    if (this.ruleCache.has(host)) {
+      let cached_item = this.ruleCache.get(host);
+      let len = cached_item && cached_item != nullIterable ? cached_item.size : 0;
```
cowlicks: just add a `size`  property to the `nullIterable` thing.

chromium/rules.js
koops76: @cowlicks why use `Object.create(null, { ...`?

koops76: @cowlicks This will create a new instance of `{ next() { return {done: true}; } }` each time.

koops76: Currently I have this code:
```js
const nullIterator = {
  next() {
    return { done: true };
  }
};

const nullIterable = {
  [Symbol.iterator]: function () {
    return nullIterator;
  },
  size: 0
};
```

chromium/rules.js
cowlicks: Also, using `Object.create(...` lets us specify other properties we might want, like `size` in this case.

chromium/rules.js
cowlicks: @koops76 see https://github.com/EFForg/https-everywhere/pull/12577#discussion_r139248205

re: new instance of ... thing - that is fine, it should get GC'd immediately anyway

chromium/rules.js
koops76: Would simply doing `{ size: 0 }` cause a conflict?

koops76: Ah, it makes it read only.

koops76:
@cowlicks Look at this code:

```js
const nullIterator = Object.create(null, {
  next: { value: function() { return { done: true }; } }
});

const nullIterable = Object.create(null, {
  [Symbol.iterator]: { value: function() { return nullIterator; } }
});
```

Is it okay to use in your opinion?

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
 
-    var results = [];
+    // Let's search!
+    let resultSet = new Set();
+
     if (this.targets.has(host)) {
       // Copy the host targets so we don't modify them.
-      results = results.concat(this.targets.get(host));
+      let results = this.targets.get(host);
+      results.map(result => resultSet.add(result));
```
cowlicks: I'm a little confused by having both `results` and `resultsSet` variables. You can just do:

```javascript
if (this.targets.has(host)) {
  results = new Set(...results, this.targets.get(host));
```

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
     }
 
     // Ensure host is well-formed (RFC 1035)
-    if (host.indexOf("..") != -1 || host.length > 255) {
-      log(WARN,"Malformed host passed to potentiallyApplicableRulesets: " + host);
-      return null;
+    if (host.length > 255 || host.indexOf("..") != -1) {
```
cowlicks: This makes sense :+1:

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
+      log(WARN, "Malformed host passed to potentiallyApplicableRulesets: " + host);
+      return nullIterable;
     }
 
     // Replace each portion of the domain with a * in turn
-    var segmented = host.split(".");
```
cowlicks: Please avoid renaming and changing this code unless its necessary to the PR. This algorithm is important, but isn't well tested unfortunately.

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
-    for (let i=0; i < segmented.length; i++) {
-      let tmp = segmented[i];
-      segmented[i] = "*";
-      results = results.concat(this.targets.get(segmented.join(".")));
```
cowlicks: This could be something like my above comment, (https://github.com/EFForg/https-everywhere/pull/12577/files#r139258106) or maybe a ternary expression like
```javascript
results = (this.targets.has(segmented.join('.')) ? 
  (new Set([...results, ...this.targets.get(segmented.join('.')]) :
  results);
```

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
+      }
+
+      parts[i] = tmp;
     }
+
     // now eat away from the left, with *, so that for x.y.z.google.com we
     // check *.z.google.com and *.google.com (we did *.y.z.google.com above)
-    for (var i = 2; i <= segmented.length - 2; ++i) {
-      var t = "*." + segmented.slice(i,segmented.length).join(".");
-      results = results.concat(this.targets.get(t));
```
cowlicks: Please revert this rename. Then just use one of the methods I suggested above to update the `results`

chromium/rules.js
```diff
@@ -405,51 +416,61 @@ RuleSets.prototype = {
...
+      let domain = "*." + parts.slice(i, parts.length).join(".");
+      let results = this.targets.get(domain);
+      if (results) {
+        results.map(result => resultSet.add(result));
+      }
     }
 
-    // Clean the results list, which may contain duplicates or undefined entries
-    var resultSet = new Set(results);
-    resultSet.delete(undefined);
```
cowlicks: Keep a `results.delete(undefined)` around to be safe.

koops76: what

cowlicks:
@koops76 re: https://github.com/EFForg/https-everywhere/pull/12577#issuecomment-329908594

It depends on the context. I'd prefer to avoid using 2 names here. I don't think it creating an iterator when iterated matters. It should be garbage collected immediately, and js engines are pretty good at optimizing simple functions like this.

@cschanaj with the size property it would look like:

```javascript
const nullIterable = Object.create(null, {
  [Symbol.iterator]: {
    value: () => ({
      next() {
        return {done: true};
      }
    })
  },
  size: {value: 0},
});

```

cowlicks:
@cschanaj in the future, please try to keep revisions to PR's as the same PR. This makes reviewing easier for me.  It is hard to track what changed and what didn't between two similar PR's so I end up re-reading a lot of code. (I'm referring to this and #12494).

There is no need to worry about force pushes or anything.

cschanaj:
@cowlicks modified accordingly, please check. thanks!!

cschanaj:
@cowlicks Sorry for force pushing a few times. This PR is ready for review and merging now. thanks!

cschanaj:
@cowlicks @Hainish I did not realize that you are modifying rules.js as well. I will rebase this branch in the weekend...

cowlicks:
@cschanaj I can rebase this if you'd like.

cschanaj:
thanks!!

cschanaj:
@cowlicks please merge this if it looks good for you. thank you!!

chromium/rules.js
```diff
@@ -416,63 +430,70 @@ RuleSets.prototype = {
...
-    if (cached_item !== undefined) {
-      util.log(util.DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.length);
+    if (this.ruleCache.has(host)) {
+      let cached_item = this.ruleCache.get(host);
+      util.log(DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.size);
       return cached_item;
+    } else {
+      util.log(DBUG, "Ruleset cache miss for " + host);
     }
-    util.log(util.DBUG, "Ruleset cache miss for " + host);
```
cowlicks: There are some issues with how this was merged, this line should have been kept (it should be `util.DBUG`).

cschanaj:
@cowlicks Travis all passing now, please merge this if it looks good. (It has been sort of annoying to rebase a few time and keeping up with the compatible changes).

cschanaj:
@cowlicks could you please merge this before the next release?

cowlicks:
@cschanaj sorry we won't have time! It will be in the next release. Totally my fault, I forgot to check back on this after you rebased :(

Thank you for your work!

cschanaj:
@cowlicks Never mind. Please be sure to merge this PR before the next release. :smile:

cowlicks:
merged in #13208 thank you!

PR #12586, opened 2017-09-15
Wrap background scripts in modules

cowlicks:
This lets us have all the background scripts nicely namespaced, while making everything both importable by node, and the browser.


chromium/background.js
```diff
@@ -146,7 +140,7 @@ var loadStoredUserRules = function() {
   for (let rule of rules) {
     all_rules.addUserRule(rule);
   }
-  log('INFO', 'loaded ' + i + ' stored user rules');
```
cowlicks: this `'INFO'` thing was actually a bug that got fixed

chromium/background.js
```diff
@@ -703,3 +697,11 @@ async function import_settings(settings) {
...
   }
 }
+
+Object.assign(exports, {
+  enableMixedRulesets,
+  all_rules,
+  initializeStoredGlobals,
+});
+
+})(typeof exports == 'undefined' ? window.background = {} : exports);
```
koops76: @cowlicks It should be `module.exports`, not simply `exports`.
```js
typeof module == 'undefined' ? window.background = {} : module.exports
```

koops76:
This will help a lot with Webpack/Rollup transition.

chromium/background.js
cowlicks: I'm pretty sure what I have here works fine. `exports` refers to `module.exports` see [here](https://nodejs.org/docs/latest/api/modules.html#modules_exports).

koops76: Are you sure?

cowlicks:
@koops76 I don't think we need to transition to webpack/rollup unless we start needing to handle many dependencies. Webextensions are already compressed when sent to browsers. So we don't need minification or anything. I also think Tor would be reluctant to include httpse if it is obfuscated through webpack/rollup.

chromium/background.js
```diff
@@ -1,13 +1,7 @@
...
-/* global RuleSets */
-/* global storage */
-/* global log */
-/* global INFO */
-/* global NOTE */
-/* global WARN */
-/* global DBUG */
+(function(exports) {
 
 "use strict";
```
Hainish: This should probably have global scope, rather than function scope

cowlicks:
I should clarify. There are a few reasons for this PR:
* namespaces are good
* avoid adding eslint directives everywhere
* It also helps people who are new to the code (like me) reason about where various things are coming from
* It allows us to start directly accessing parts of the code in isolation for unittests

chromium/background.js
cowlicks: "I'm pretty sure", would you like to check by running `rewriter.js`?

chromium/.eslintrc.json
```diff
@@ -9,8 +9,15 @@
...
+    "indent": ["error", 2, {"outerIIFEBody": 0}],
     "no-unused-vars": "error",
     "no-undef": "error"
+  },
+  "globals": {
+    "exports": true,
+    "util": true,
+    "storage": true,
+    "background": true,
+    "rules": true
```
Hainish: `incognito` is not present.  It is not used by any other module, either.  So why export it at all?  We should either add it to this list or remove the export and keep the function scope, or keep the export and add it as a global here.  We may also consider renaming the file to keep it consistent with the export name.

chromium/storage.js
```diff
@@ -1,16 +1,22 @@
...
-var storage = chrome.storage.local;
 if (chrome.storage.sync) {
   chrome.storage.sync.set({"sync-set-test": true}, () => {
     if(!chrome.runtime.lastError){
-      storage = chrome.storage.sync;
-      initializeStoredGlobals();
+
+      Object.assign(exports, {
+        get: chrome.storage.sync.get,
+        set: chrome.storage.sync.set,
```
cowlicks: @Hainish this was the only kind of gross part. In other parts of the code we refer to `storage.get/set`, but they are set in an **asynchronous** way. So I can't just do the exports at the end of the file. Looking back now, I think I should just set them with a promise. I'm going to change this.

* directly set these like `window.storage.get/set = ...get/set`
* or do this, I chose

chromium/storage.js
```diff
@@ -1,16 +1,22 @@
...
+
+      background.initializeStoredGlobals();
     }
   });
 }
-if (typeof exports != 'undefined') {
-  exports = storage;
-}
+
+})(typeof exports == 'undefined' ? window.storage = {} : exports);
```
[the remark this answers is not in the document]
cowlicks: Renamed to `store` & `store.js`.

chromium/.eslintrc.json
cowlicks: Renamed `incognito-cache-thing.js` to `incognito.js` and added `incognito` here.

chromium/background.js
cowlicks: done, for other files too

chromium/background.js
```diff
@@ -703,3 +697,11 @@ async function import_settings(settings) {
     });
   }
 }
+
+Object.assign(exports, {
+  enableMixedRulesets,
+  all_rules,
+  initializeStoredGlobals,
```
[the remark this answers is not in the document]
cowlicks: done

koops76:
Maybe we should call `window.background` `window.HTTPSEverywhere` or similar since `background.js` is the main file and `window.background` may be confused with Chrome background window?

Hainish:
:+1:

chromium/rules.js
```diff
@@ -512,7 +511,7 @@ RuleSets.prototype = {
     // observed and the domain blacklisted, a cookie might already have been
     // flagged as secure.
 
-    if (domainBlacklist.has(domain)) {
+    if (background.domainBlacklist.has(domain)) {
```
RReverser: IIUC, this still relies on global `background` variable which makes it impossible to import just `rules` in Node without defining that one first. I tried to fix that with https://github.com/EFForg/https-everywhere/pull/12556 by moving rules options to rules itself, making it self-contained, but there are conflicts now.

RReverser:
> I also think Tor would be reluctant to include httpse if it is obfuscated through webpack/rollup.

FWIW rollup doesn't obfuscate code in any way, it produces clean output as if it would be all written in one file in the first place.

chromium/rules.js
koops76: Or I can make a PR for rules.js adding a dummy `background` object if there is no real one.

koops76:
@RReverser Builds are auditable, they are not minified, simply appended with some code to glue the modules together.

chromium/rules.js
RReverser: @koops76 Exactly. As I said, not directly importable in a modular fashion without polluting globals (just like before, and was fixed in a PR I linked).

RReverser:
Btw, this additionally broke logging from `rules.js` since `log` is not imported from `utils` and not overridden later either.

koops76:
Do you mean it doesn't `require` its dependencies?

koops76:
This was merged a little bit prematurely. Needs some fixes. Will make a PR containing all fixes soon.

RReverser:
@koops76 Yes, but also `background.js` is not importable in Node (it uses Chrome APIs), while these options (`enableMixedRulesets` and `domainBlacklist`) are important for `rules.js` itself even when `background.js` is not present. It would make more sense to move their definitions to `rules.js` itself and export API to change them that `background.js` can use.

RReverser:
Something like:

```js
// rules.js
var areMixedRulesetsEnabled = false;
exports.enableMixedRulesets = () => areMixedRulesetsEnabled = true;
```

```js
// background.js
if (...stored...) rules.enableMixedRulesets();
```

RReverser:
> Builds are auditable, they are not minified, simply appended with some code to glue the modules together.

Yeah, that's what I'm saying about rollup too. It's pretty much like concatenation, just strips `import` / `export` statements while preserving correct JS semantics.

chromium/rules.js
cowlicks: @RReverser can you verify that this is indeed an issue? I am able to import this.

Before this PR, `domainBlacklist` was already `undefined` when this is imported by node. So this should not change anything.

cowlicks: I am able to import `rules.js` fine, and the `rewriter.js` script runs fine for me with cursory testing.

cowlicks:
@RReverser I believe that `enableMixedRulesets` and `domainBlacklist` were inaccessible to `rules.js` when imported from node previously.

However I think that this **does** break logging in the in the rules.js file when running in **browser**. I **don't** think it worked before this PR in **node**.

cowlicks:
@RReverser I agree that `rules.js` should work 100% regardless of whether it is imported by node or the browser. However I think the correct approach is to make `require`ing the background file nullpotent. Then in extension land, the extension can be started by a `start.js` file that is loaded after everything else.

Moving functions around might be a useful thing to do too, but I don't know their use well enough confirm that.

chromium/rules.js
RReverser: @cowlicks Yeah but can you actually use it? I mean, as soon as it goes into one of code paths where it tries to access `background` variable, it will fail with ReferenceError (yay JS, but anyway ESLint would've warned about it if it wouldn't be in the globals array in config).

RReverser:
> I don't think it worked before this PR in node.

Yeah it didn't without also patching globals, that's why I mentioned https://github.com/EFForg/https-everywhere/pull/12556 where I fixed that before (but it wasn't merged as it has various conflicts now).

Not saying it's a new thing that got broken, just that it requires different vars now and needs to be fixed over again.

PR #12596, opened 2017-09-17
Re-add problematic Pirate Bay proxies from #12481.

bardiharborow:
This should pass the rules test once #12481 is merged, but I still don't know why the fetch test works locally by fails on Travis. Are Travis containers run somewhere with censored internet?


bardiharborow:
Wait, I've actually booted up Tor and checked these (they are blocked from where I am... I was running fetch test from a GCE VM) and they are all non functional as pirate bay proxies so maybe we don't care about them.

PR #12613, opened 2017-09-18
[Idenyt.dk] Update ruleset.

bardiharborow:


src/chrome/content/rules/Idenyt.dk.xml
```diff
@@ -1,20 +1,10 @@
-<!--
-	Cloudflare SSL on ^ and www
-
-	Problematic domains:
-		^ ²
-		www ²
-		foto ¹
-		jul2k8 ¹
-		(www.)tv ¹
```
J0WI: Why did you remove all comments? Some are still true.

bardiharborow: @J0WI not sure actually, fixed.

PR #12614, opened 2017-09-18
[Mozdev] Update ruleset.

bardiharborow:


Bisaloo:
Could you add this test url as well please? I think it's good practice to have deep links as test urls for "complex" rules.

http://bugzilla.mozdev.org/enter_bug.cgi?product=forward

PR #12558, opened 2017-09-14
Measure and slightly improve memory usage

RReverser:
Added a simple script for measuring memory usage of the ruleset in "cold" and "hot" states from the console with Node.js.

Also implemented two small optimisations to see how it works.

Results with Node.js 8.5.0 on MacOS (kernel 16.7.0):

**Initial usage** *(rulesets just parsed and ready)*: 42.1 MB -> 40.4 MB
**Maximum usage** *(all the regexps force-compiled)*: 80.2 MB -> 78.1 MB

Note that this includes commit from https://github.com/EFForg/https-everywhere/pull/12556 as it depends on it.

Related issue: https://github.com/EFForg/https-everywhere/issues/12232

PR #12622, opened 2017-09-18
Fix modules

koops76:
Ping @RReverser @cowlicks @Hainish.


RReverser:
Thanks for fixing! Not sure all the duplicated wrappers are worth it TBH, especially given that this still doesn't fix `rules.js` being not directly require-able from Node, still works only in extension unless you do some global patching.

koops76:
Just copy `rules.js` and put a `background.js` shim in the same folder.

RReverser:
No, I already explained in previous PR why it doesn't work that way... `background.js` uses Chrome-specific APIs even at top level, and it's not intended to be executed from Node.js. `rules.js` can and should be just a self-contained module instead of relying on any browser-specific bits.

koops76:
```js
'use strict'

exports = {
  enableMixedRulesets: false,
  urlBlacklist: {
    has () { return false }
  }
}
```

koops76:
That's the shim.

koops76:
@RReverser

RReverser:
Which means the end script has to provide own shims just to require rewriter? Even for things like https://github.com/EFForg/https-everywhere/pull/12558 I would have to keep separate shim and preload it into global before requiring rules.js? I don't really see why this complication is needed when a much simpler solution can be done that wouldn't require any extra effort on consumption side...

koops76:
Hmm, how?

koops76:
I guess a copy of `rules.js` and the shim can be copied into a npm package and kept somewhere in the repo. Scripts may just use it.

RReverser:
Of course, anything can be shimmed, worked around, patched, hacked etc. on consumer side, but why not make it simple in the first place when it's as easy as moving two variables between files? All I want is to be able to launch Node.js, type `require('./rules')` and have it already working.

> Hmm, how?

As wrote here: https://github.com/EFForg/https-everywhere/pull/12586#issuecomment-330318222

Basically make rules.js (or, ideally, each script) independent, just providing APIs for options that can be modified externally.

koops76:
Okay. This PR will fix two bugs then.

cowlicks:
Hey @koops76 @RReverser  I'm working on fixing this, so please hold off for a bit until I can get it out.

koops76:
@RReverser Added standalone `rules.js` support.

chromium/rules.js
```diff
@@ -248,7 +252,7 @@
       var platform = ruletag["platform"]
       if (platform) {
         default_state = false;
-        if (platform == "mixedcontent" && background.enableMixedRulesets) {
+        if (platform == "mixedcontent" && background.enableMixedRulesets.enable) {
```
RReverser: Shouldn't this be without `background.`?

chromium/background.js
```diff
@@ -61,9 +61,9 @@
   all_rules = new rules.RuleSets(ls);
 
   // Allow users to enable `platform="mixedcontent"` rulesets
-  var enableMixedRulesets = false;
+  rules.enableMixedRulesets = false;
```
RReverser: Missed `.enable`?

chromium/rules.js
```diff
@@ -252,7 +250,7 @@
       var platform = ruletag["platform"]
       if (platform) {
         default_state = false;
-        if (platform == "mixedcontent" && background.enableMixedRulesets.enable) {
+        if (platform == "mixedcontent" && enableMixedRulesets.enable) {
```
RReverser: Btw why do you need this indirection and not just `enableMixedRulesets` being a boolean directly?

koops76: This won't work since it needs to be the same object.

koops76: This can be done with `get()` and `set()` methods instead.

RReverser: Or you can just use `exports.enableMixedRulesets`. Then it will be same inside and outside.

cowlicks:
@koops76  would you mind marking this as **WIP**  in the title so we know not to review it?

koops76:
@cowlicks Wait until Travis will finish.

cowlicks:
@koops76 oops sorry. I meant to comment on your ES6 imports PR. Can you mark that is WIP so we know not to review it.

Also, I'm much more in favor of using #12556 to fix this.

koops76:
@cowlicks Done.

koops76:
@cowlicks It's closed, and if you fix all conflicts and merge it anyway, it will cause conflicts for this PR.

cowlicks:
Module fixes are being done in #12627 is this still needed?

PR #12623, opened 2017-09-18
[WIP] ES6 modules

koops76:
Do not merge.


koops76:
ping @RReverser @cowlicks @Hainish.

chromium/.eslintrc.json
```diff
@@ -4,21 +4,21 @@
     "es6": true,
     "webextensions": true
   },
+  "overrides": [
+    {
+      "files": ["background.js", "entry.js", "incognito.js", "rules.js" ,"store.js", "util.js"],
```
RReverser: What's this needed for / why not `module` by default?

RReverser:
Code looks good. Does module syntax work in browser extensions?

koops76:
@RReverser No.

chromium/.eslintrc.json
koops76: `module` disables `/* exported */`

cowlicks:
Is this compatible with node?

koops76:
No, it is currently only compatible with Webpack or Rollup.

koops76:
@Hainish Need help with integrating Rollup with the build system.

RReverser:
@koops76 That should be pretty trivial - just add Rollup config and call it as part of makecrx and makexpi.

koops76:
I'm not sure if Node is available in `makecrx.sh` and `makexpi.sh`.

RReverser:
It is AFAIK, see @Hainish's comment in https://github.com/EFForg/https-everywhere/pull/12493#issue-256083902

cowlicks:
I'm still skeptical about our need for Rollup. 

So it might be better to have a discussion about why we need rollup before implementing it. What advantages does it provide? What are the disadvantages?

koops76:
@cowlicks Rollup is basically a very clever `cat *.js > out.js`. It concatenates the scripts and links them together, automatically dealing with name conflicts. It uses ES6 modules feature to understand how the scripts should be linked. It doesn't minify or obfuscate the produced code.

cowlicks:
@koops76 is there a way with rollup/webpack to load the extension unpacked? I like to develop like this because changes are immediately reflected in the browser. It seems like it would break this.

koops76:
@cowlicks Use watch mode in Rollup or Webpack.

PR #12624, opened 2017-09-18
Add selenium tests for firefox and chrome

cowlicks:
@Hainish ready for review.

The new tests are inside `test/selenium/*`, there are two files for setting up the tests cases, `shim.py` and `util.py`. Define a test case where the browser gets loaded with the extension for each test and is controlled by selenium. Then the `test/selenium/test_*.py` files use these tests.

There are changes to some of the build tools in  [this commit](https://github.com/EFForg/https-everywhere/pull/12624/commits/942ed90ce686348f27d5be570d2ced5b69dae3aa) to make them python2/3 compatible. The new tests are written to py3, no guarantees about py2.

In `setup_travis.sh` we set up travis for each test. In `run_travis.sh` we run each test.

I made changes to `test/travis.sh`. I removed the code paths for the old chromium, firefox, and eslint tests. The browser tests were removed since I added new better ones. The eslint tests were just moved into `run_travis.sh`. The rules, fetch, and preloaded tests are passed through `run_travis.sh` to `test/travis.sh`. 

Note that the selenium tests are not run inside docker. This can be changed if there is a need later. The eslint test is not run inside docker either.

There were also some changes to `chromium/popup.js`. It was showing the UI in a default state, then popuplating it with data later. This caused race conditions. The changes now make the popup wait until it has data to the display the UI.

When this is ready to merge, I'll make a separate issue listing some todos that can be done later, my list:
```
add tests for:
    when js is disabled
    when cookies are disabled settings persist
    test unstable rules shown
    firefox beta tests
```


test/selenium/test_navigation.py
```diff
@@ -0,0 +1,31 @@
...
+        self.driver.get(mixed_url)
+        self.assertTrue(self.driver.current_url.startswith('https'))
+        self.query_selector(mixed_selector).click()
+        self.assertFalse(mixed_success_text in self.query_selector('body').text)
+
+    def test_mixed_content_not_blocked(self):
+        self.driver.get(mixed_url)
+        self.assertTrue(self.driver.current_url.startswith('https'))
+        self.query_selector(mixed_selector).click()
+        self.assertTrue(mixed_success_text in self.query_selector('body').text)
```
cowlicks: @Hainish what else should I add here? We can probably use URLs from badssl.com

cowlicks:
I'd like to eventually break the testing utilities into a few pieces:
* extension configuratinon
* browser shim stuff
* base unittest class
* extension specific testing utilities

test/selenium/test_navigation.py
Hainish: This looks fine for now, let's look at this again when there's time

Makefile
```diff
@@ -0,0 +1,7 @@
+crx:
+	./makecrx.sh
+
+xpi:
```
cowlicks: @Hainish how do you feel about Makefiles? I like to use them for organizing build scripts, and other house keeping stuff. But I probably shouldn't add this here unless we start filling it out more in the future.

koops76: Makefiles should not be used in favor of `package.json` or at least bare .sh files. We should also avoid tying the build process to any specific OS.

Hainish: We've had a project-level Makefile in the past.  I removed it.  It fell into disuse, and since no one payed attention to it, it quickly become stale when things changed.  I'd rather avoid one altogether.

cowlicks:
I'm fixing bugs on firefox now. The main issue is how slow firefox is, so I have to add a lot of gross waits/sleeps.

test/selenium/test_navigation.py
cowlicks: Add js disabled here

cowlicks: add testcase for if cookies disabled persisting settings works

check test/readme.md todo

test/selenium/test_popup.py
```diff
@@ -0,0 +1,54 @@
+from util import ExtensionTestCase
+
+
+class TestPopup(ExtensionTestCase):
+    def test_rule_shown(self):
+        url = 'http://freerangekitten.com'
+        self.load_popup_for(url)
+        el = self.query_selector('#StableRules > div > label > span')
```
cowlicks: get an `#UnstableRules` test case

cowlicks:
this should replace some of the tests in travis.yml for firefox and chrome

chromium/background.js
```diff
@@ -710,6 +712,7 @@ Object.assign(exports, {
   all_rules,
   initializeStoredGlobals,
   urlBlacklist,
+  state,
```
cowlicks: Adding this `state` stuff can be removed. I was using it to check the value of js variables after interacting with UI. But instead, now I just verify the UI is in the correct state after refreshing.

cowlicks:
A job passed finally! https://travis-ci.org/EFForg/https-everywhere/jobs/281447723

cowlicks:
The tests for firefox beta are failing because it doesn't support legacy addons. We should add a webextensions branch test in another PR

cowlicks:
Just solved an issue I was trying to debug.

We were generating a new `dummy-chromium.pem` key in `makecrx.sh` whenever it didn't exist. This isn't under version control, so we'd generate a new one everytime we pushed to travis. Instead of doing this, we now have a static development key, see [this commit](https://github.com/EFForg/https-everywhere/pull/12624/commits/7ae45048d835afe5228fe5368b98d2381b6b2e74).

cowlicks:
Yay! It is passing.

koops76:
@cowlicks To keep the electronic signature, rebase to most recent commit from master and squash the commits yourself.

cowlicks:
@koops76 Sorry, what do you mean by "the electronic signature"?

koops76:
@cowlicks GPG: "This commit was signed with a verified signature."

cowlicks:
I have not purged any files yet that are obsolete, parts of `test/ chromium.sh firefox.sh script.py`.

cowlicks:
hmm I seeing a strange issue in python only where `./makexpi.sh` throws an error validating `./src/chrome/content/rules/Meritrust_CU.org.xml`

chromium/popup.js
```diff
@@ -275,3 +276,12 @@ function setOption_(opt, value, callback) {
   details[opt] = value;
   sendMessage("set_option", details, callback);
 }
+
+function getTab(callback) {
+  let url = new URL(window.location.href);
+  if (url.searchParams.has('tabId')) {
+    var parentId = parseInt(url.searchParams.get('tabId'));
```
koops76: `const parentId = Number(url.searchParams.get('tabId'));`

chromium/popup.html
```diff
@@ -19,13 +19,13 @@
     <h1 data-i18n="about_ext_name"></h1>
 </header>
 
-<section id="disableButton" class="options">
+<section id="disableButton" class="options" style="display:none;">
```
koops76: No need of a semicolon here.

chromium/popup.html
```diff
@@ -19,13 +19,13 @@
...
     <div class="onoffswitch">
         <input id="onoffswitch" type="checkbox" checked>
         <label data-i18n="menu_globalEnable" for="onoffswitch"></label>
     </div>
 </section>
-<section id="HttpNowhere" class="options">
+<section id="HttpNowhere" class="options" style="display:none;">
```
koops76: No need of a semicolon here.

chromium/popup.js
```diff
@@ -116,13 +116,25 @@ function appendRuleLineToListDiv(ruleset, list_div, tab_id) {
   list_div.appendChild(line);
 }
 
+function showHttpNowhereUI() {
+  // Set up checkbox for HTTP nowhere mode
+  getOption_('httpNowhere', false, function(item) {
+    if (item.httpNowhere) {
+      e('http-nowhere-checkbox').setAttribute('checked', '');
+    }
```
koops76: `e('http-nowhere-checkbox').checked = item.httpNowhere;`

cowlicks:
I'm moving the python 2/3 stuff to a separate PR. I don't want to mix in build changes if I don't have too.

test/travis.sh
```diff
@@ -40,53 +36,25 @@ fi
...
-    docker run --rm -ti -v $(pwd):/opt node bash -c "cd /opt/utils/eslint && npm install && ./node_modules/.bin/eslint ../../chromium"
-  fi
-
-  if [ "$TEST" == "firefox" ]; then
-    echo >&2 "Running firefox test suite."
-    docker_build
-    docker run --rm -ti -v $(pwd):/opt -e FIREFOX=/$FIREFOX/firefox/firefox httpse bash -c "test/firefox.sh"
-  fi
-
-  if [ "$TEST" == "chromium" ]; then
```
cowlicks: These are now run from `run_travis.sh`

test/travis.sh
```diff
@@ -40,53 +36,25 @@ fi
...
+if [ "$TEST" == "rules" ]; then
+  echo >&2 "Performing comprehensive coverage test."
+  docker run --rm -ti -v $(pwd):/opt httpse python utils/ruleset_filenames_validate.py
+  docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/validate.sh"
+  docker run --rm -ti -v $(pwd):/opt httpse bash -c "test/rules.sh"
+  docker run --rm -ti -v $(pwd):/opt node bash -c "cd /opt && node utils/normalize-securecookie.js"
+  [ `git diff --name-only | wc -l` -eq 0 ]
 fi
-# Only run test if something has changed.
-if [ "$RULESETS_CHANGED" ]; then
```
cowlicks: We always run the tests now.

This avoids a possible issue where the ruleset validation code changes, but the rulesets don't. So the new validation code would not get tested against existing rulesets.

chromium/popup.html
cowlicks: I think either way is correct. Would you mind motivating why you would like the semi-colon removed?

chromium/popup.html
cowlicks: As above, would you mind motivating this? In future reviews, if a change is stylistic, or if it is a case where both ways are correct, would you mind motivating your comments?

I usually find that small stylistic changes like this can be a time sink, with little benefit.

chromium/popup.js
cowlicks: I think changing `var` to `let` here makes sense. `const` isn't used anywhere else in this file so it might be confusing. 

I'm not sure why `Number` would be better.

cowlicks:
I'm going to merge this since it will make reviewing existing PR's much easier and since most changes do not effect the HE extension itself. The exception is the changes to `chromium/popup.js` and `chromium/popup.html` @Hainish should review these before the next release.

PR #12625, opened 2017-09-19
Update background.js

koops76:


koops76:
ping @Hainish.

chromium/background.js
```diff
@@ -370,18 +370,17 @@ function onBeforeRequest(details) {
     }
   }
 
-  if (using_credentials_in_url) {
+  if (newuristr) {
```
Hainish: This isn't quite what I meant.  Sill use the `if (using_credentials_in_url) {` but inside it add the conditional `if (newuristr) {`, with an `} else {` clause.

koops76:
@Hainish Done.

PR #12627, opened 2017-09-19
re-work #12556

cowlicks:
Fixes some issues from #12586 with some work from #12556 by @RReverser 

Lets rules.js use `enableMixedRulesets` and `domainBlacklist` by moving them into rules.js from background.js.

Stubs out `util.*` so they work normally in the browser again, this was broken in #12556.

### DIFF
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -103,7 +103,7 @@ RuleSet.prototype = {
     if (this.exclusions !== null) {
       for (let exclusion of this.exclusions) {
         if (exclusion.pattern_c.test(urispec)) {
-          log(DBUG, "excluded uri " + urispec);
+          util.log(util.DBUG, "excluded uri " + urispec);
           return null;
         }
       }
@@ -210,7 +210,7 @@ RuleSets.prototype = {
       try {
         this.parseOneXmlRuleset(s);
       } catch (e) {
-        log(WARN, 'Error processing ruleset:' + e);
+        util.log(util.WARN, 'Error processing ruleset:' + e);
       }
     }
   },
@@ -220,7 +220,7 @@ RuleSets.prototype = {
       try {
         this.parseOneJsonRuleset(ruleset);
       } catch(e) {
-        log(WARN, 'Error processing ruleset:' + e);
+        util.log(util.WARN, 'Error processing ruleset:' + e);
       }
     }
   },
@@ -300,7 +300,7 @@ RuleSets.prototype = {
    * @returns {boolean}
    */
   addUserRule : function(params) {
-    log(INFO, 'adding new user rule for ' + JSON.stringify(params));
+    util.log(util.INFO, 'adding new user rule for ' + JSON.stringify(params));
     var new_rule_set = new RuleSet(params.host, true, "user rule");
     var new_rule = new Rule(params.urlMatcher, params.redirectTo);
     new_rule_set.rules.push(new_rule);
@@ -313,7 +313,7 @@ RuleSets.prototype = {
     if (new_rule_set.name in this.ruleActiveStates) {
       new_rule_set.active = (this.ruleActiveStates[new_rule_set.name] == "true");
     }
-    log(INFO, 'done adding rule');
+    util.log(util.INFO, 'done adding rule');
     return true;
   },
 
@@ -323,7 +323,7 @@ RuleSets.prototype = {
    * @returns {boolean}
    */
   removeUserRule: function(ruleset) {
-    log(INFO, 'removing user rule for ' + JSON.stringify(ruleset));
+    util.log(util.INFO, 'removing user rule for ' + JSON.stringify(ruleset));
     this.ruleCache.delete(ruleset.name);
 
 
@@ -336,7 +336,7 @@ RuleSets.prototype = {
       this.targets.delete(ruleset.name);
     }
 
-    log(INFO, 'done removing rule');
+    util.log(util.INFO, 'done removing rule');
     return true;
   },
 
@@ -417,10 +417,10 @@ RuleSets.prototype = {
     // Have we cached this result? If so, return it!
     var cached_item = this.ruleCache.get(host);
     if (cached_item !== undefined) {
-      log(DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.length);
+      util.log(util.DBUG, "Ruleset cache hit for " + host + " items:" + cached_item.length);
       return cached_item;
     }
-    log(DBUG, "Ruleset cache miss for " + host);
+    util.log(util.DBUG, "Ruleset cache miss for " + host);
 
     var results = [];
     if (this.targets.has(host)) {
@@ -430,7 +430,7 @@ RuleSets.prototype = {
 
     // Ensure host is well-formed (RFC 1035)
     if (host.indexOf("..") != -1 || host.length > 255) {
-      log(WARN,"Malformed host passed to potentiallyApplicableRulesets: " + host);
+      util.log(util.WARN,"Malformed host passed to potentiallyApplicableRulesets: " + host);
       return null;
     }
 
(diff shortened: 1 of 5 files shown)


RReverser:
Thanks, this looks pretty great to me. I'll leave just a few comments.

chromium/rules.js
```diff
@@ -3,10 +3,21 @@
 (function(exports) {
 
 // Stubs so this runs under nodejs. They get overwritten later by util.js
-var DBUG = 2;
-var INFO = 3;
-var WARN = 5;
-function log(){}
+if (typeof util == 'undefined') {
```
RReverser: Suggestion: move `util.js` in manifest to the top, so that in extension you would always have global `util` already, and in the Node.js you wouldn't. Then, this entire block can be changed to just `var util = typeof util !== 'undefined' ? util : { log: () => {} };` (such typeof check is valid because `var` declarations get merged with globals).

chromium/rules.js
```diff
@@ -228,7 +239,7 @@ RuleSets.prototype = {
     var platform = ruletag["platform"]
     if (platform) {
       default_state = false;
-      if (platform == "mixedcontent" && background.enableMixedRulesets) {
+      if (platform == "mixedcontent" && enableMixedRulesets) {
```
RReverser: Actually this check here and in other functions won't work because other JS files only modify property on `exports` object but local variable will remain intact. It's better to remove variable for `enableMixedRulesets` and always check `exports.enableMixedRulesets` instead (note that `domainBlacklist` is fine because it's only ever modified through its own methods and not replaced with another value like boolean).

chromium/rules.js
cowlicks: Good  point about moving utils.js, will do.

However, this code is in an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression) so a variable declared here with `var ...` should not be added to the global scope. I'm pretty sure it should still be added manually to `window`.

chromium/rules.js
cowlicks: Good catch! I moved `enableMixedRulesets` and `domainBlacklist` into `rules.settings.*`. So now we refer to `settings.enableMixedRulesests in this file, and in other files, `rules.settings.enableMixedRulesets`. 

I did this because I think direct references to  `exports` should be reserved for actually exporting things.

chromium/rules.js
```diff
@@ -209,7 +220,7 @@ RuleSets.prototype = {
       try {
         this.parseOneJsonRuleset(ruleset);
       } catch(e) {
-        log(WARN, 'Error processing ruleset:' + e);	
+        util.log(util.WARN, 'Error processing ruleset:' + e);
```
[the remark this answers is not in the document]
cowlicks: Here is the WIP require stuff

chromium/rules.js
cowlicks: I don't think `settings` is the best name. These are really just stateful objects.

chromium/rules.js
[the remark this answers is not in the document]
RReverser: > I'm reluctant to add more functionality to this function because it is supposed to be a stub. Especially functionality that diverges from the functionality of the original function.

What about separate function? Something like `rules.reportError` which would throw in Node env and log otherwise.

chromium/rules.js
```diff
@@ -3,9 +3,8 @@
 (function(exports) {
 
 // Stubs so this runs under nodejs. They get overwritten later by util.js
-if (typeof util == 'undefined') {
-  let base = typeof window == 'undefined' ? global : window; // eslint-disable-line
-  Object.assign(base, {
+if (typeof util == 'undefined' || typeof global != 'undefined') {
```
RReverser: This still doesn't have to be that complex - given that it's just a stub, `if (typeof util === 'undefined') var util = { log: () => {} };` would be enough.

chromium/rules.js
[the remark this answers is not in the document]
RReverser: This doesn't need to be in the same PR though, I'm fine with a stub for now (please see comment below though).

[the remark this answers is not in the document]
cowlicks: @RReverser I think it would be better for the function to be configurable by the user,  instead of having it switch behavior base on some global state. So something like `util.log.setBehavior(util.warn, ...)` would be a better API. Or, since this is specific to your needs, you could just monkey patch `console.warn` like:
```javascript
console.warn  = function() {
  throw(arguments);
}
```

Either way, I think adding this feature is out of scope of this PR.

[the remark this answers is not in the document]
koops76: @cowlicks
`throw` is not a function, it is an operator:
```js
throw arguments;
```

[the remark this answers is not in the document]
RReverser: > instead of having it switch behavior base on some global state

Which is why I suggested having something like `rules.reportError` (or, now, `rules.settings.reportError` for consistency).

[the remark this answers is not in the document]
RReverser: Monkey-patching `console.warn` is worse since it affects all the Node.js libs then, not just `rules.js`. Point here is to restore native behaviour of errors (have them thrown and not silently swallowed) which doesn't work when using rules.js as a library because it catches all the errors.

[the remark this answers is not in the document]
RReverser: But yeah, it's out of scope either way.

chromium/rules.js
cowlicks: Your code snippet is definitely satisfactory for this case!

However, I wanted to be extra defensive and make as few assumptions about the global state as possible. There are a lot of of reasons that just `util` could be undefined, but there are very few reasons both `util` would be undefined, and `global` is not undefined.

Like if someone re-arranges the files in the manifest.json, then I won't overwrite the local `util.log`.

chromium/rules.js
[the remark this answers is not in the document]
cowlicks: @RReverser oh sorry I think I misunderstood your API proposal. Having `rules.settings.reportError` be a setting for toggling error reporting would be reasonable. Agreed.

chromium/rules.js
RReverser: I just don't see why `global` matters. Even if someone rearranges files, you'll have `typeof util == 'undefined'` which means you set `log` stub here and later `util.js` overrides it with proper implementation. If you have current order, then `typeof util != 'undefined'` and you just don't override implementation. Hence `typeof util` check seems sufficient, but maybe I don't see the other edge cases you're trying to cover.

RReverser: Another part of my comment was that there is no point in defining `DBUG` etc. since it's either properly defined by `util.js` together with `log` function, or they simply don't matter and can be `undefined` when `log` is just a stub.

RReverser: @cowlicks Could you please make this last small change or, if you still don't think it's reasonable, at least merge this as-is?

cowlicks: I think that `global` matters. The scenario you explain is totally correct, however it is a detail of our current implementation. `global` being present is a pretty strong indicator that we are running in a node environment, which is the only case I would want to stub `util`. This is true regardless of our implementation. Likewise, the reason `util.CONSTANTS` don't cause a problem is a detail of how it is used in this file, which might change.

koops76: @cowlicks Better check for presence of `module` and `module.exports`.

RReverser: @cowlicks 

> global being present is a pretty strong indicator that we are running in a node environment, which is the only case I would want to stub util

I would say the opposite - in Node environment it's especially uncommon to pollute globals. When you're in Node env, a local util variable would be sufficient, which is why I suggested using it over actually putting anything into `global` namespace.

RReverser: Fixed in last commit of https://github.com/EFForg/https-everywhere/pull/12558 (so `util` will be a local stub when rules.js being used as Node.js library, and global variable when part of the extension).

cowlicks: @RReverser

>I would say the opposite - in Node environment it's especially uncommon to pollute globals.

I'm not sure what you mean. I'm referring to the variable called `global` in a node environment, which is the name of the Global Object in node. In the browser it is called `window`. This is why I think `global` being present indicates a node environment. Wouldn't you agree? 

>  When you're in Node env, a local util variable would be sufficient, which is why I suggested using it over actually putting anything into global namespace.

If we stub a local `util` variable, then it would not get overwritten later when we actually load `util.js`. Defining `util` locally instead of globally would also further diverge from the behavior in the browser. Which is undesirable. 

At the end of the day, having to stub this out is a gross hack. It should be removed completely when we have better module loading.

RReverser: > If we stub a local util variable, then it would not get overwritten later when we actually load util.js.

In Node.js it anyway won't because you are using `module.exports` to export utils, so for now it's totally fine to stub as module-level variable.

PR #12631, opened 2017-09-19
Create OpenSSL.org.xml

koops76:


gloomy-ghost:
https://hstspreload.com/api/v1/status/openssl.org preloaded

koops76:
@gloomy-ghost Why did the test succeed then?

gloomy-ghost:
The test is more conservative since it was originally aimed for removing targets from existing rulesets, but if we are adding new rulesets I think there is not much point to add the ruleset and remove it a little while later

Issue #12526, opened 2017-09-12
CBS.xml and CBSNews.xml have conflicting rules


RReverser:
CBSNews.xml added in https://github.com/EFForg/https-everywhere/pull/11933 seems to conflict with CBS.xml which has:

```xml
	<rule from="^http://markets\.cbsnews\.com/$"
		to="https://markets.financialcontent.com/cbsnews" />

	<rule from="^http://markets\.cbsnews\.com/"
		to="https://markets.financialcontent.com/" />
```

As far as I can tell, CBS.xml's rewrite rules are incorrect in this case, and CBSNews' `markets.cbsnews.com` should always take effect.

(also, seems like a good idea to somehow auto-test that different rulesets don't match same URLs)

Bisaloo:
> (also, seems like a good idea to somehow auto-test that different rulesets don't match same URLs)

+1

`encrypted.google.com` had a similar issue (#12544)

Issue #11440, opened 2017-07-27
Include rulesets with different platforms in duplicate checks


Bisaloo:
I think we should review the way tests are currently working to check duplicates.

At the moment, tests will pass even if a target is duplicated in more than one ruleset as long as it is not in more than one ruleset *without platform*. For example, the domain in this PR (https://github.com/EFForg/https-everywhere/pull/11438) was not in `duplicate-whitelist.txt`.

But this can cause problems. In this case, it was pretty obvious that there is mixed content so the ruleset without platform shouldn't have made it to the repo.

But we can think of more tricky cases, where the mixedcontent is not on the landing page but on deeplinks. In that case, User 1 creates a ruleset without `platform="mixedcontent"`. User 2 sees that the website can be secured but doesn't realize there is already a rule for it (`platform="mixedcontent"` rules are disabled by default for most users). But User 2 misses the fact that this website has mixed content and creates a ruleset without `platform`. We are now breaking a website when we could have easily avoided it.

I propose that *all* duplicates require to be in `duplicate-whitelist.txt`, even if one of the rulesets they are in has a `platform`. I know that in some cases, some contributors will secure some resources in a ruleset with `platform` and the rest in `platform="mixedcontent"` but adding such domains in the whitelist is trivial and that's what the whitelist is for anyways (not to mention the fact that to my knowledge, no one is doing that anymore. MB used to do it but they were the only one).

Bisaloo:
To answer @J0WI's comment in https://github.com/EFForg/https-everywhere/pull/11438#issuecomment-318362537:

> Do you have any idea how many rules would be affected by this change?

As far as I can tell, not many. I don't know how the current `duplicate-whitelist.txt` was generated but it already includes some of the cases I am describing (1 ruleset without `platform` + 1 ruleset with `platform="mixedcontent"` such as https://github.com/EFForg/https-everywhere/pull/10910/files) but apparently not all of them.

```sh
egrep '<target\s+host=\s*"' *.xml | cut -d '"' -f 2 | sort | uniq -c | grep -v '1 ' | wc -l
```

returns 230 candidates (with some possible false positives (commented targets).

The current `duplicate-whitelist`@master has 

```sh
wc -l utils/duplicate-whitelist.txt
```
153 targets.

cschanaj:
[...]

PR #12643, opened 2017-09-19
Check conflicting rulesets

RReverser:
As mentioned in https://github.com/EFForg/https-everywhere/issues/12526, rulesets sometimes conflict with each other, resulting in different rewrites.

This PR adds an automated check that detects such cases by trying test URLs against all rulesets at once.

Currently found issues:

- [ ] `src/chrome/content/rules/Geocaching.com.xml`:
  - Test url http://www.geocaching.com/forums/default.aspx can be rewritten in several ways:
    * `src/chrome/content/rules/Geocaching.com-falsemixed.xml`: https://www.geocaching.com/forums/default.aspx
    * `src/chrome/content/rules/Geocaching.com.xml`: https://forums.groundspeak.com/GC/
  - Test url http://www.geocaching.com/forums/default.aspx? can be rewritten in several ways:
    * `src/chrome/content/rules/Geocaching.com-falsemixed.xml`: https://www.geocaching.com/forums/default.aspx?
    * `src/chrome/content/rules/Geocaching.com.xml`: https://forums.groundspeak.com/GC/
  - Test url http://www.geocaching.com/forums/default.aspx?f can be rewritten in several ways:
    * `src/chrome/content/rules/Geocaching.com-falsemixed.xml`: https://www.geocaching.com/forums/default.aspx?f
    * `src/chrome/content/rules/Geocaching.com.xml`: https://forums.groundspeak.com/GC/
  - Test url http://www.geocaching.com/forums/default.aspx?o can be rewritten in several ways:
    * `src/chrome/content/rules/Geocaching.com-falsemixed.xml`: https://www.geocaching.com/forums/default.aspx?o
    * `src/chrome/content/rules/Geocaching.com.xml`: https://forums.groundspeak.com/GC/
- [ ] `src/chrome/content/rules/Google.org.xml`: Test url http://google.org/ can be rewritten in several ways:
  * `src/chrome/content/rules/Google.xml`: https://www.google.org/
  * `src/chrome/content/rules/Google.org.xml`: https://google.org/

### DIFF
--- a/test/rules/src/https_everywhere_checker/check_rules.py
+++ b/test/rules/src/https_everywhere_checker/check_rules.py
@@ -13,6 +13,7 @@ import re
 import sys
 import threading
 import time
+from itertools import imap
 
 from ConfigParser import SafeConfigParser
 
@@ -26,7 +27,7 @@ from rule_trie import RuleTrie
 def convertLoglevel(levelString):
 	"""Converts string 'debug', 'info', etc. into corresponding
 	logging.XXX value which is returned.
-	
+
 	@raises ValueError if the level is undefined
 	"""
 	try:
@@ -36,17 +37,17 @@ def convertLoglevel(levelString):
 
 def getMetricClass(metricType):
 	"""Get class for metric type from config file.
-	
+
 	@raises ValueError if the metric type is unknown
 	"""
 	metricMap = {
 		"markup": metrics.MarkupMetric,
 		"bsdiff": metrics.BSDiffMetric,
 	}
-	
+
 	if metricType not in metricMap:
 		raise ValueError("Metric type '%s' is not known" % metricType)
-	
+
 	return metricMap[metricType]
 
 
@@ -54,22 +55,22 @@ class ComparisonTask(object):
 	"""Container for objects necessary for several plain/rewritten URL comparison
 		 associated with a single ruleset.
 	"""
-	
+
 	def __init__(self, urls, fetcherPlain, fetcherRewriting, ruleset):
 		self.urls = urls
 		self.fetcherPlain = fetcherPlain
 		self.fetcherRewriting = fetcherRewriting
 		self.ruleset = ruleset
 		self.ruleFname = ruleset.filename
-	
+
 class UrlComparisonThread(threading.Thread):
 	"""Thread worker for comparing plain and rewritten URLs.
 	"""
-	
+
 	def __init__(self, taskQueue, metric, thresholdDistance, autoDisable, resQueue):
 		"""
 		Comparison thread running HTTP/HTTPS scans.
-		
+
 		@param taskQueue: Queue.Queue filled with ComparisonTask objects
 		@param metric: metric.Metric instance
 		@param threshold: min distance that is reported as "too big"
@@ -276,7 +277,7 @@ def cli():
 
 	config = SafeConfigParser()
 	config.read(args.checker_config)
-	
+
 	logfile = config.get("log", "logfile")
 	loglevel = convertLoglevel(config.get("log", "loglevel"))
 	if logfile == "-":
@@ -285,7 +286,7 @@ def cli():
 	else:
 		logging.basicConfig(filename=logfile, level=loglevel,
 			format="%(asctime)s %(levelname)s %(message)s [%(pathname)s:%(lineno)d]")
-		
+
 	autoDisable = False
 	if config.has_option("rulesets", "auto_disable"):
 		autoDisable = config.getboolean("rulesets", "auto_disable")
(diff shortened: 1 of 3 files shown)


Bisaloo:
Now that I think about it, I am not convinced this is the best approach. In my opinion, it would be better to report all subdomains that are covered by more than one ruleset, whether it is by conflicting rules or not.

- This would serve as an extension of the current `duplicate-whitelist.txt`. Related: https://github.com/EFForg/https-everywhere/issues/11440
- Better performance (?) https://github.com/EFForg/https-everywhere/issues/12232

RReverser:
> report all subdomains that are covered by more than one ruleset, whether it is by conflicting rules or not

I'm not sure that's reasonable as many rulesets are intentionally split into `(partial)` definitions. At the same time, intersecting rules are interesting as actual bugs and not just optimization because they rely on implicit rule collection order which can result in inconsistent rewrites.

Bisaloo:
> I'm not sure that's reasonable as many rulesets are intentionally split into (partial) definitions and splitting them further can give better performance in cast when most subdomains can degrade to trivial rewrite, but some edge case can't.

True but my feeling is that only concerns a small and ever decreasing number of rulesets.

Furthermore, I am not saying those rules should be removed. But I think they should be listed in the `duplicate-whitelist.txt` file. My understanding is that it's the reason why it was created in the first place.

RReverser:
@bisaloo I just feel that we are trying to solve two different problems, and both checks can be implemented separately. I'm trying to detect and precent addition of specifically conflicting rewrites because they bit me several times already in our alternative internal implementation of HTTPS Everywhere rewriter when I'm going through all the test URLs and compare results with official rewriter as tests. From time to time this results in test failures because database itself contains conflicting rules and different rewriters consume them in different order, and while I could workaround it by comparing entire sets of results and seeing if there's any intersection, it seems worth it to fix database itself for this case since even official rewriter might produce non-deterministic results depending on kinds of optimizations we add.

RReverser:
@Bisaloo So do you mind merging this check separately?

Bisaloo:
I don't have write access to this repo, sorry.

test/rules/src/https_everywhere_checker/check_rules.py
```diff
@@ -394,7 +399,24 @@ def cli():
...
+					continue
+				ruleConflictProblemsExist = True
+				logging.error("%s: Test url '%s' can be rewritten in several ways:\n%s" % (
+					ruleset.filename,
+					test.url,
+					'\n'.join(imap(
+						lambda match: '  %s: %s' % (match.ruleset.filename, match.url),
+						matches
+					))
+				))
```
cowlicks: Would you mind unnesting this a bit to make it easier to read?
```python
                                msg = "%s: Test url '%s' can be rewritten in several ways:\n%s"
				ways = '\n'.join(
					imap(lambda match: '  %s: %s' % (match.ruleset.filename, match.url),
					     matches))
				logging.error(msg % (ruleset.filename, test.url, ways))
```
Something like this would work, or with hanging closing brackets.

cowlicks:
@RReverser When I run this I see the geocaching domains, but not the google domains. Any idea why?

Otherwise, this looks good.

cowlicks:
I think travis is failing because this needs to be rebased. Mas.ter has newer travis-ci stuff

cowlicks:
On failure the same message is printed twice, is this from this PR?:
```bash
+ echo '

Under the new ruleset testing rules (February 2015), any modified ruleset
must have sufficient test coverage. You can often improve test coverage by
adding <test url="..." /> tags, or by restructuring the rule to avoid
wildcard <target> tags. See these documents:
https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#ruleset-style-guide
'

Under the new ruleset testing rules (February 2015), any modified ruleset
must have sufficient test coverage. You can often improve test coverage by
adding <test url="..." /> tags, or by restructuring the rule to avoid
wildcard <target> tags. See these documents:
https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#ruleset-style-guide

+ exit 1
+ rm /tmp/tmp.gZGPL12JqR
```

RReverser:
> On failure the same message is printed twice, is this from this PR?:

First seems to be the `echo` command for the output below (I assume it just runs in verbose mode or something, but anyway unrelated to the PR).

RReverser:
> When I run this I see the geocaching domains, but not the google domains. Any idea why?

I guess Google was fixed since I opened the PR.

RReverser:
> I think travis is failing because this needs to be rebased. Mas.ter has newer travis-ci stuff

For what it's worth, it will be failing even after rebase due to this new check and geocaching rulesets...

J0WI:
This check looks really helpful. Is this PR still alive?

RReverser:
Define "alive". It worked when submitted, not sure if it still does with upstream changes and/or whether maintainers are interested in merging it.

J0WI:
ping @cowlicks @Hainish

Hainish:
I suspect a lot of the conflicts here are because of cleaning up the python scripts with `autopep8` and converting to python 3.6 in the last year.

@RReverser, would you be willing to rebase these changes in a new PR?  Sorry for letting this slip last year, this was about the same time as we were making the big transition to WebExtensions.

PR #12645, opened 2017-09-19
Add Node.js to dev dependencies.

koops76:
We'll need it for Rollup, UglifyES and other Node.js tools.


koops76:
ping @Hainish @cowlicks.

Hainish:
I think in most cases you will not need node for development.  Rollup seems overly complicated for any benefit it provides.  Plus I'm not to keen on the `curl $SOMETHING | bash` callout.

PR #12658, opened 2017-09-20
Add Ganttproject

J0WI:


jeremyn:
It looks like only `www` has mixed content issues. The other domains are fine. Do you want to split this into two rulesets? Otherwise this is ready to merge as-is.

J0WI:
Done.

PR #12607, opened 2017-09-17
Remove GoogleSorry.xml

Bisaloo:

PR #12554, opened 2017-09-14
[WIP] Simplify Google.xml

Bisaloo:
- [x] Remove outdated exclusions
- [ ] Remove unnecessary rules

PR #12659, opened 2017-09-20
[Google] Remove broken target

Bisaloo:
No DNS record.

### DIFF
--- a/src/chrome/content/rules/Google.xml
+++ b/src/chrome/content/rules/Google.xml
@@ -46,7 +46,6 @@
 				- www.google.in **
 				- www.google.kr *
 				- www.google.com.kz *
-				- www.google.com.lk *
 				- www.google.mx *
 				- www.google.sg *
 				- www.google.sl *
--- a/src/chrome/content/rules/Google.xml
+++ b/src/chrome/content/rules/Google.xml
@@ -71,8 +71,8 @@
 
 	<rule from="^http://(www\.)?google\.lk/"
 		to="https://www.google.lk/" />
-		<test url="http://google.lk" />
-		<test url="http://www.google.lk" />
+		<test url="http://google.lk/" />
+		<test url="http://www.google.lk/" />
 
 	<rule from="^http://(?:www\.)?google\.(?:com\.)?kz/"
 		to="https://www.google.kz/" />
--- a/src/chrome/content/rules/Google.xml
+++ b/src/chrome/content/rules/Google.xml
@@ -70,12 +70,10 @@
 		<test url="http://google.co.ug/" />
 		<test url="http://www.google.co.ug/" />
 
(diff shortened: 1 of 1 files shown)


jeremyn:
Here's what I get:

Domain | In DNS?
---|---
http://google.com.kz | Yes
http://www.google.com.kz | Yes
http://google.kz | Yes
http://www.google.kz | Yes
http://google.com.lk | No
http://www.google.com.lk | No
http://google.lk | Yes
http://www.google.lk | Yes

Your change removes coverage for the last two items in the list, http://www.google.lk and http://google.lk. Also, the `.kz` domains can be moved into the next `rule` because http://www.google.com.kz exists.

I think the relevant part of the ruleset should look like this:

```xml
	<rule from="^http://(www\.)?google\.lk/"
		to="https://www.google.lk/" />
		<test url="http://google.lk/" />
		<test url="http://www.google.lk/" />

	<rule from="^http://(?:www\.)?google\.(?:com\.)?(bo|ec|kz|mx|sg|sl|vn)/"
		to="https://www.google.com.$1/" />
		<test url="http://google.com.bo/" />
		<test url="http://www.google.com.bo/" />
		<test url="http://google.com.ec/" />
		<test url="http://www.google.com.ec/" />
		<test url="http://google.com.kz/" />
		<test url="http://www.google.com.kz/" />
		<test url="http://google.com.mx/" />
		<test url="http://www.google.com.mx/" />
		<test url="http://google.com.sg/" />
		<test url="http://www.google.com.sg/" />
		<test url="http://google.com.sl/" />
		<test url="http://www.google.com.sl/" />
		<test url="http://google.com.vn/" />
		<test url="http://www.google.com.vn/" />
```

What do you think?

Bisaloo:
First of all, let me start by saying that the aim of this PR is to fix the tests (https://github.com/EFForg/https-everywhere/pull/12555 and https://github.com/EFForg/https-everywhere/pull/12607) with minimal changes, just so I can keep working on the google related rulesets.

I will come back to it (https://github.com/EFForg/https-everywhere/pull/12554) in the process of my rewrite of the google related rulesets (you can look at my recent PR for the work until now).

> Your change removes coverage for the last two items in the list, http://www.google.lk and http://google.lk. 

You are right, I should fix this.

> Also, the .kz domains can be moved into the next rule because http://www.google.com.kz exists.

I think the rewrite you are proposing is problematic as well because it rewrites `(www\.)?google.kz` to `www.google.com.kz`, which is unnecessary because both `google.kz` and `www.google.kz` work fine over `https`.

I am convinced this specific ruleset is the worst in the google family and needs an almost complete rewrite. It is probably more reasonable to fix this issue in another PR.

jeremyn:
Merged, thanks.

PR #12663, opened 2017-09-20
Rename and update Github.xml

christianheinrichs:
Noticed that not all cookies were being secured, also added some subdomains.


Bisaloo:
`github.com` is preloaded. Not sure why this passes the tests.

bardiharborow:
Could be something to do with `www` not serving HSTS, but I need to confirm.

christianheinrichs:
@Bisaloo Are we talking about the [Chromium HSTS preload list](https://cs.chromium.org/codesearch/f/chromium/src/net/http/transport_security_state_static.json) or the [Mozilla HSTS preload list](https://dxr.mozilla.org/comm-central/source/mozilla/security/manager/ssl/nsSTSPreloadList.inc)? 

Either way, in both cases the `github.com` entry seems to be present and I might need to remove it from this ruleset.

@bardiharborow Did you figure it out?

gloomy-ghost:
@Hainish `github.com` is present in esr52, any ideas that the test passed?

bardiharborow:
@Bisaloo @gloomy-ghost see https://github.com/EFForg/https-everywhere/pull/12145#issuecomment-331695961.

cschanaj:
see https://github.com/EFForg/https-everywhere/pull/12145#issuecomment-331702092. fixed by #12748

christianheinrichs:
I tried to secure the `github.com` cookies by adding:

```xml
<securecookie host="^github\.com$" name=".+" />
<securecookie host="^\.github\.com$" name=".+" />
```

But this only works when `github.com` is included via `target` XML tag.

If there's a way to enable the secure flag for all cookies on `github.com` without having it included as target, please let me know.

bardiharborow:
@christianheinrichs The cookies should not need securing. HSTS guarantees that `*.github.com` will _never_ be accessed over an unsecured connection – essentially a permanent `<securecookie host=".+" name=".+" />` for the domain.

wesdawg:
`github.com` is not preloading for me in Chrome (latest stable on Linux)
github_hsts
Same http redirect on www

Am I missing something, or is this a _really bad_ bug in Chrome ?

bardiharborow:
@wesdawg that 307 redirect _is_ the preloading. If you look more closely, it will say `307 Internal Redirect`, indicating that it's internally generated by Chrome.

src/chrome/content/rules/GitHub.xml
```diff
@@ -1,16 +1,31 @@
...
+	Redirects:
+		- codeload.github.com → github.com
+		- learn.github.com → services.github.com
+		- raw-origin.github.com → github.com
+		- teach.github.com → services.github.com
+		- training.github.com → services.github.com
+
+	Secure connection failed:
+		- ssh.github.com
 -->
```
gloomy-ghost: Since we already have github.com proloaded, I think it's fine to remove these. Maybe you could just leave a comment to state that it's HSTS preloaded and thus doesn't need to be in the ruleset.

christianheinrichs: Done.

gloomy-ghost: Actually you don't need to mention the subdomains at all, as includeSubDomains is one of the requirements to get preloaded

christianheinrichs: The newest commit should have fixed that.

PR #12678, opened 2017-09-22
Remove EasyCottages.com from Wyndham.xml

cschanaj:
seemingly no longer in use.


J0WI:
Please also remove the redirect rule here: https://github.com/cschanaj/https-everywhere/blob/03af7487aeeb0b284ee2239a17941a9963747eba/src/chrome/content/rules/Wyndham.xml#L154

cschanaj:
@J0WI thanks!!

PR #12708, opened 2017-09-22
Update Jason_Davies.com.xml (adding <securecookie host=".+" name=".+" /> )

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


jeremyn:
We don't need to add this `securecookie` tag, because these domains don't seem to set cookies.

As a general rule we shouldn't include stuff that we can't test, including `securecookie`s on domains that don't set cookies.

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
Thanks @jeremyn took note of it!

PR #12714, opened 2017-09-22
Delete sign-rulesets-airgap.sh

koops76:
ping @Hainish @cowlicks.

This doesn't make signing any more secure.

Quoting https://github.com/EFForg/https-everywhere/issues/12087#issuecomment-325048106:
> Your "air-gap" scheme is no more secure than simply signing on a Internet-connected machine.
A compromised machine can output **any hash**, tricking you into *signing **any hash***.
Also it's easier to sneak a dodgy ruleset into the GitHub repository than trying to fool your "air-gap" scheme. Are you going to personally read each ruleset file?


Hainish:
This does, we're not changing this process.  It prevents malware on a developer machine from automatically making a malicious signature of rulesets.  Malware which is detected is able to be removed (or the development machine rebuilt) befope a malicious signature is made, since the signing keys are separated by a physical airgap

koops76:
Are you sure you will be able to notice the malware before it's too late?

Hainish:
Certainty is never certain

koops76:
What are you going to do if you find any malware? I hope you won't simply delete the executable and continue using this machine for signing.

PR #12719, opened 2017-09-23
[regexr.com] Add ruleset.

bardiharborow:


src/chrome/content/rules/regexr.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+
+	Problematic hosts in *regexr.com:
+
+		- mail (mismatched)
+		- ww (mismatched)
+		- wwww (mismatched)
```
J0WI: Wildcard DNS

bardiharborow:
@J0WI ah didn't think of that.

PR #12724, opened 2017-09-23
Create SelfAuthoring.com.xml

koops76:


koops76:
ping @J0WI.

src/chrome/content/rules/SelfAuthoring.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="SelfAuthoring.com">
```
J0WI: Mismatch on `mail.selfauthoring.com`

PR #12728, opened 2017-09-23
Create NirSoft.net.xml

koops76:


koops76:
ping @J0WI.

J0WI:
I saw some mixed content issues in very old blog posts, but I consider this as trivial.

PR #12738, opened 2017-09-23
Delete make-json.py

koops76:


koops76:
ping @Hainish @cowlicks.

cowlicks:
We also use `make-json.py` inside `makecrx.sh`. What is the purpose of this script and the `rulesets.json` file it generates?

koops76:
@cowlicks Seems to be a leftover.

cowlicks:
Would you mind:
* removing the usage from makecrx.sh
* rewording the commit messages to be more descriptive
* explaining when/why the usage of `make-json.py` was deprecated

koops76:
@cowlicks I think it was deprecated when we moved back to XML.

PR #12739, opened 2017-09-23
Update merge-rulesets.sh

koops76:


koops76:
ping @Hainish @cowlicks.

utils/merge-rulesets.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/sh
 
-# Merge all the .xml rulesets into a single "default.rulesets" file -- this
-# prevents inodes from wasting disk space, but more importantly, works around
-# the fact that zip does not perform well on a pile of small files.
+# Merge all the .xml rulesets into a single "default.rulesets" file -- this prevents inodes from wasting disk space, but more importantly,
+# works around the fact that zip does not perform well on a pile of small files.
```
Hainish: Why this change?

utils/merge-rulesets.sh
```diff
@@ -1,9 +1,11 @@
...
 
 python2.7 ./utils/merge-rulesets.py
 
 RULESETS=chrome/content/rules/default.rulesets
+
+# We make default.rulesets at build time, but it shouldn't have a variable timestamp.
+touch -r src/install.rdf src/$RULESETS
```
Hainish: I don't see how this is a substantive improvement.  Care to explain?

koops76: I moved this out of the Python script into the shell script.

utils/merge-rulesets.sh
koops76: There's no need to limit line length to 80 chars.

koops76: All modern editors allow the user to choose the word wrap.

Hainish: Yet most styles still prefer it.  Take PEP 8, which instructs wrapping at 79 characters.  Wrapping a comment line can be quite ugly in some editors.  Let's keep the wrapping, it's orthogonal to the separate change in this PR.

koops76: @Hainish Done.

Hainish:
Looks good, thanks @koops76

PR #12741, opened 2017-09-24
Require === and !==

koops76:
ping @Hainish @cowlicks.

See https://eslint.org/docs/rules/eqeqeq.

This change requires testing.


cschanaj:
This changes might have some issues when handling `undefined` which required to be fixed, in particular

```bash
$ nodejs
> undefined !== null;
true
> undefined != null;
false
```

koops76:
@cschanaj Can you find any?

cschanaj:
AFAIK, the parsing of JSON use 

```
if (variable != null)
```

 which can be trivially rewrote by 

```
if (variable)
``` 

for the same functionalities. I am not sure about the rest of the code.

koops76:

chromium/rules.js
```diff
@@ -101,7 +101,7 @@ RuleSet.prototype = {
   apply: function(urispec) {
     var returl = null;
     // If we're covered by an exclusion, go home
-    if (this.exclusions !== null) {
+    if (this.exclusions) {
```
cowlicks: Why was this changed?

chromium/rules.js
```diff
@@ -250,20 +250,20 @@ RuleSets.prototype = {
 
     // Read user prefs
     if (rule_set.name in this.ruleActiveStates) {
-      rule_set.active = (this.ruleActiveStates[rule_set.name] == "true");
+      rule_set.active = this.ruleActiveStates[rule_set.name];
```
cowlicks: Does `rule_set.active`  go from being a bool, to a string?

chromium/rules.js
```diff
@@ -250,20 +250,20 @@ RuleSets.prototype = {
...
     }
 
     var rules = ruletag["rule"];
     for (let rule of rules) {
-      if (rule["from"] != null && rule["to"] != null) {
+      if (rule["from"] && rule["to"]) {
```
cowlicks: This code block will not execute in cases when it did before in the case where `rule["from"]` and `rule["to"]` are falsey, but not undefined. e.g:
```javascript
0 != null && false != null  // true
0 && false  // false
```

chromium/rules.js
```diff
@@ -250,20 +250,20 @@ RuleSets.prototype = {
...
         rule_set.rules.push(new Rule(rule["from"], rule["to"]));
       }
     }
 
     var exclusions = ruletag["exclusion"];
-    if (exclusions != null) {
+    if (exclusions) {
       for (let exclusion of exclusions) {
-        if (exclusion != null) {
+        if (exclusion) {
```
cowlicks: likewise, if `exclusion  = false`.
```javascript
exclusion != null // true
exclusion  // false
```

chromium/rules.js
```diff
@@ -273,9 +273,9 @@ RuleSets.prototype = {
     }
 
     var cookierules = ruletag["securecookie"];
-    if (cookierules != null) {
+    if (cookierules) {
```
cowlicks: In these cases, can you please explain why the variable will never be falsey but not `undefined` or `null`?

chromium/rules.js
```diff
@@ -312,7 +312,7 @@ RuleSets.prototype = {
     // TODO: maybe promote this rule?
     this.targets.get(params.host).push(new_rule_set);
     if (new_rule_set.name in this.ruleActiveStates) {
-      new_rule_set.active = (this.ruleActiveStates[new_rule_set.name] == "true");
+      new_rule_set.active = this.ruleActiveStates[new_rule_set.name];
```
cowlicks: This looks like another case we might have changed from bool to string.

chromium/rules.js
```diff
@@ -359,7 +359,7 @@ RuleSets.prototype = {
     var platform = ruletag.getAttribute("platform");
     if (platform) {
       default_state = false;
-      if (platform == "mixedcontent" && settings.enableMixedRulesets) {
+      if (platform === "mixedcontent" && settings.enableMixedRulesets) {
```
cowlicks: I'm guessing that `==` and `===` have the same operator precedence?  Could you add parentheses here to make it more explicit anyway

chromium/rules.js
```diff
@@ -493,7 +493,7 @@ RuleSets.prototype = {
 
     var potentiallyApplicable = this.potentiallyApplicableRulesets(hostname);
     for (let ruleset of potentiallyApplicable) {
-      if (ruleset.cookierules !== null && ruleset.active) {
+      if (ruleset.cookierules && ruleset.active) {
```
cowlicks: what if `ruleset.cookierules` is falsey but alse `!== null`?

chromium/rules.js
koops76: No, simply bad code.

chromium/rules.js
koops76: Thanks!

koops76: Why would `from` and `to` be falsey? They should not be empty.

chromium/rules.js
koops76: Can `exclusion` be false?

chromium/rules.js
koops76: It will always be either an array (always truthy), or `null` or `undefined`.

chromium/rules.js
koops76: They do.

chromium/rules.js
koops76: It can only be an array or `null` or `undefined`.

chromium/rules.js
cowlicks: I don't know, I'm asking you. What can `exclusion` be? Is it possible for it to be falsey, but not null?

chromium/rules.js
cowlicks: Well, is that a bug? What exactly do we want to catch? `null`, `undefined`, anything else?

koops76: Any falsey or non-array value I guess.

chromium/rules.js
koops76: I think it may be a string. In this case empty string is also falsey, but exclusions shouldn't be empty strings.

koops76:
ping @Hainish @cowlicks.

PR #12754, opened 2017-09-25
Update popup.js

koops76:
ping @Hainish @cowlicks.

### DIFF
--- a/chromium/popup.js
+++ b/chromium/popup.js
@@ -199,11 +199,6 @@ document.addEventListener("DOMContentLoaded", function () {
   e("add-rule-link").addEventListener("click", addManualRule);
 });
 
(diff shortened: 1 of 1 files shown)


koops76:

Hainish:
I like the existing prefill is better because for a user it is clearer what is being replaced.  Simply having "https:" in a replacement text makes sense for those who are familiar with trivialized rulesets and that this only applies to the target host, but it is not intuitive for most users.  It clearly gives the format for redirects, so that a user can simply change the prefill text, which is easier.  For instance, a user may forget about the "." being a metacharacter in regex, making the creation of a simple redirect from an insecure subdomain to a secure subdomain harder.

PR #12755, opened 2017-09-25
Create XeNTaX.com.xml

koops76:


J0WI:
Some sites have MCB issues.

koops76:
@J0WI Which ones?

J0WI:
Nearly any page, but after some research I think these are only Ad/Tracking Skripts. So I'm fine with this.

PR #12758, opened 2017-09-25
Unbreak WebAssign

Injust:
After logging in, the redirection from www.webassign.net to webassign.net breaks the site. Leaving the www in place fixes this.

### DIFF
--- a/src/chrome/content/rules/WebAssign.xml
+++ b/src/chrome/content/rules/WebAssign.xml
@@ -1,7 +1,11 @@
+<!--
+	Invalid certificate:
+	- feedback.webassign.net
+-->
+
 <ruleset name="WebAssign">
   <target host="api.webassign.net" />
   <target host="demo.webassign.net" />
-  <target host="feedback.webassign.net" />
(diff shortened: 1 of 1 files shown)


Bisaloo:
If you want to redirect all targets to their `https` equivalent, you can use

```xml
```

src/chrome/content/rules/WebAssign.xml
```diff
@@ -4,6 +4,5 @@
   <target host="webassign.net" />
```
J0WI: Please also add:
```
api.webassign.net
feedback.webassign.net
webassignwired.webassign.net

Injust: I've added the subdomains. Thanks!

J0WI: Sadly not all of them support HTTPS. Please use comments to note non-working hosts.

Injust: Fixed, sorry about that.

PR #12768, opened 2017-09-26
Fix blocked mixed content on Asexuality.org

gsinp:
Forum users on asexuality.org are having issues caused by this ruleset, such as broken page number links at the bottom of forum threads, search forms not submitting, broken like buttons on posts, etc.

There are lots of mixed content ajax requests getting blocked:
```
Mixed Content: The page at 'https://www.asexuality.org/en/search/' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://www.asexuality.org/en/index.php?app=core&module=system&controller=ajax&do=findMember&input=test&csrfKey=a4fd372821ea00cf4b5209116cd35183'. This request has been blocked; the content must be served over HTTPS.
```

Disabling the ruleset checkbox in the HTTPSE popup or altogether disabling HTTPSE makes the issues go away when using the site over http.

This PR makes the ruleset active only on Tor Browser with `platform="mixedcontent"`. There are users who access this site only with Tor Browser and they have few issues with the ruleset because TB doesn't block mixed content.

Additionally, the site's login form targets the insecure endpoint `http://www.asexuality.org/en/login/`. In the interest of securing Chrome and Firefox users' passwords, this PR also adds a new ruleset that's active in all browsers for the sole purpose of redirecting the login form target to https. This doesn't secure auth tokens after users have logged in, but it's a small improvement over logging in with plain http.

### DIFF
--- a/src/chrome/content/rules/asexuality.org_login.xml
+++ b/src/chrome/content/rules/asexuality.org_login.xml
@@ -12,12 +12,10 @@
 -->
 <ruleset name="Asexuality.org (partial)">
 	<target host="asexuality.org" />
-		<test url="http://asexuality.org/en/" />
 		<test url="http://asexuality.org/en/login/" />
 		<test url="http://asexuality.org/?q=about.html" />
 		<test url="http://asexuality.org/?q=contact.html" />
 	<target host="www.asexuality.org" />
-		<test url="http://www.asexuality.org/en/" />
 		<test url="http://www.asexuality.org/en/login/" />
 		<test url="http://www.asexuality.org/?q=about.html" />
 		<test url="http://www.asexuality.org/?q=contact.html" />
--- a/src/chrome/content/rules/asexuality.org_login.xml
+++ b/src/chrome/content/rules/asexuality.org_login.xml
@@ -22,9 +22,6 @@
 		<test url="http://www.asexuality.org/?q=about.html" />
 		<test url="http://www.asexuality.org/?q=contact.html" />
 
-	<exclusion pattern="^http://(www\.)?asexuality\.org/en/(?!login)" />
-		<!-- Exclude all areas of the forum except the login form target URL. -->
-
 	<rule from="^http://(www\.)?asexuality\.org/en/login"
 			to="https://www.asexuality.org/en/login/" />
 		<!-- Secure the login form target URL. -->
--- a/src/chrome/content/rules/asexuality.org_login.xml
+++ b/src/chrome/content/rules/asexuality.org_login.xml
@@ -12,8 +12,10 @@
 <ruleset name="Asexuality.org (login form)">
 	<target host="asexuality.org" />
 		<test url="http://asexuality.org/en/login/" />
+		<exclusion pattern="^http://asexuality\.org/$" />
 	<target host="www.asexuality.org" />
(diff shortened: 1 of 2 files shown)


Bisaloo:
Hi, thank you for your PR!

Could you please add a test url pointing to a page with mixed content? It helps a lot if someone comes back to this ruleset in the future. Right now, looking at the landing page https://asexuality.org/, it is possible to miss the mixed content issues.

Also, just for your information, Tor Browser has started to block active mixed content a couple of months ago. For this reasons, `platform="mixedcontent"`rulesets are off by default on Tor Browser as well. But they can still be manually enabled.

gsinp:
I've added test urls and comments explaining how to trigger mixed content blocking and Uncaught DOMExceptions caused by https.

Also, there is a difference between the forum part of the site located at http://www.asexuality.org/en/ and the non-forum part at https://www.asexuality.org/. They are essentially two different sites, with the non-forum one practically being a static site. The ruleset now redirects the non-forum area to https too.

Reports of issues with HTTPSE on the site have been happening over the last year. So reports of Tor Browser being free from issues may have happened prior to the browser's change to block mixed content.

Bisaloo:
Thank you! I just wish people would report the bugs here. We can't fix them otherwise :(

gsinp:
One more commit to add `(partial)` to the name of the main ruleset because it only covers a few parts of the site. I think it's ready to merge.

src/chrome/content/rules/asexuality.org_login.xml
```diff
@@ -0,0 +1,34 @@
...
+		<test url="http://asexuality.org/?q=about.html" />
+		<test url="http://asexuality.org/?q=contact.html" />
+	<target host="www.asexuality.org" />
+		<test url="http://www.asexuality.org/en/" />
+		<test url="http://www.asexuality.org/en/login/" />
+		<test url="http://www.asexuality.org/?q=about.html" />
+		<test url="http://www.asexuality.org/?q=contact.html" />
+
+	<exclusion pattern="^http://(www\.)?asexuality\.org/en/(?!login)" />
+		<!-- Exclude all areas of the forum except the login form target URL. -->
```
Bisaloo: The use of negative lookahead operator is discouraged in the [contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#regex-conventions) and I don't think it is needed anyways. Your rules are specific enough to avoid those cases.

gsinp:
I removed the exclusion and the corresponding test urls.

gsinp:
Just realized the ruleset needed a rule to redirect the account signup form target url to https, which is another password-related form that defaults to http (http://www.asexuality.org/en/register/)

I wanted to add a rule for the "change password" form too, but the build would fail because that url returns 403 Forbidden unless you're logged in.

gsinp:
I think this is ready to merge and I've tested it on the site. If someone could take a look at it when they get a chance, that'd be great. Thanks.

gsinp:
The site has been fixed and all the parts that were previously broken on https work fine now. Additionally, the server now redirects all requests to https by default, but there's still no hsts. Closing this request because it's no longer needed and the original ruleset is fine.

PR #12778, opened 2017-09-26
Added ruleset for kachelmannwetter.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/kachelmannwetter.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Content mismatch:
+		wetterkanal.kachelmannwetter.com
```
J0WI: I get the same content, but mixed content issues.

Jochen-A-Fuerbacher:
You're right. Changed the comment.

PR #12781, opened 2017-09-27
Delete Newgrounds.xml

mrjohnson22:
Newgrounds has migrated to HTTPS, and will be attentive to & will try to resolve any site breakages caused by it:
https://www.newgrounds.com/bbs/topic/1426552

This means HTTPS Everywhere is no longer necessary for Newgrounds.


koops76:
See #12783.

koops76:
Also see https://www.eff.org/https-everywhere/faq#why-does-https-everywhere-include-rules-for-sites-like-paypal-that-already-require-https-on-all-their-pages.

mrjohnson22:
Thanks for the advice, and for #12783. Closing this.

PR #12785, opened 2017-09-27
Added apply55gx.com rule

Apply55gx:


koops76:
Better use HSTS preloading.

koops76:
https://www.eff.org/https-everywhere/faq#how-do-i-add-my-own-site-to-https-everywhere

Apply55gx:
@koops76 I already did. Thought it would be still good to add a rule.

koops76:
@Apply55gx OK.

Apply55gx:
Latest Commit: Added a rule for ag.ch

koops76:
It would be removed when it will become preloaded in all browsers.

Apply55gx:
Oh OK

PR #12790, opened 2017-09-27
Update Dockerfile

koops76:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,12 @@ FROM electronicfrontierfoundation/https-everywhere-docker-base
 MAINTAINER William Budington "bill@eff.org"
 WORKDIR /opt
 
-ADD test/rules/requirements.txt test/rules/requirements.txt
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,12 +1,13 @@
...
 WORKDIR /opt
-ADD test/rules/requirements.txt test/rules/requirements.txt
-ADD test/chromium/requirements.txt test/chromium/requirements.txt
-RUN pip install -r test/rules/requirements.txt
-RUN pip install -r test/chromium/requirements.txt
 
-ENV FIREFOX /firefox-latest/firefox/firefox
+COPY test/rules/requirements.txt /tmp/
+RUN pip install -r /tmp/requirements.txt
+RUN rm /tmp/requirements.txt
```
gloomy-ghost: Maybe we should combine the `RUN`s so that there are less layers being produced?

PR #12797, opened 2017-09-27
Add "block all mixed content" option

koops76:


koops76:
ping @Hainish @cowlicks.

cowlicks:
Hi @koops76 this pull request looks good! And blocking mixed content is the right thing to do!

However I believe this feature is already enabled by default in both chrome and firefox. So unfortunately, I don't think we should include this. Since it may end up conflicting with the existing browser features. And it adds additional maintenance overhead.

Thank you for working on this! Let me know if I missed anything.

koops76:
@cowlicks This also blocks passive mixed content, Chrome doesn't have such option. Mozilla has it somewhere in about:config.

koops76:
Why would I add a feature that is already implemented in all browsers?

koops76:
@cowlicks Never mind.

PR #12798, opened 2017-09-27
Fix import_settings error

koops76:


koops76:
ping @Hainish @cowlicks.

PR #12805, opened 2017-09-28
no-var (let)

koops76:
ping @Hainish @cowlicks.


koops76:
ping @Hainish @cowlicks.

PR #12819, opened 2017-09-28
background.js: in this case we do not need async

koops76:


Hainish:
I'd rather ensure the import actually works before displaying a success dialogue in the options UI, which requires async.

koops76:
This code does the same thing, but OK.

PR #12828, opened 2017-09-29
Create WikiBound.info.xml

koops76:


src/chrome/content/rules/WikiBound.info.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="WikiBound.info">
+	<target host="www.wikibound.info" />
```
J0WI: Please also add `wikibound.info` (without `www`).

PR #12829, opened 2017-09-29
Create AMStat.org.xml

Bisaloo:


jeremyn:
- [ ] https://amstat.org works for me, make it a `target` and remove it from `Connection reset`
- [ ] https://jobs.amstat.org is `Time out`, not `Refused`
- [ ] I think the `ruleset` `name` should be `amstat.org`, not `American Statistical Association`. It's easier to understand what a domain-name ruleset does, and also I expect `amstat.org` would be easier for a non-English speaker to identify.

Bisaloo:
- https://jobs.amstat.org is still refused for me
- I renamed the ruleset as asked but I usually like using the website own title as ruleset name. "American Statistical Association is what you see on your browser tab and on the website top banner.

jeremyn:
I'm still getting a `Time out` for https://jobs.amstat.org/ instead of `Refused`, but it's not a big deal so that's fine.

I understand what you mean about the `ruleset` `name`. However even though `American Statistical Association` is easy for me personally to locate on the page and in the title, I might have trouble doing that if the name/title/page were in for example an alphabet I don't know, with grammar rules such as noun declension that change the text, or if the language followed an ordering convention I don't expect such as going right-to-left. In that case I would tell the contributor to change the `name` to the domain which I could unambiguously verify, and we shouldn't have a double standard here.

Another problem is that these descriptive-type ruleset names can encourage contributors to combine domains in a way that really does require a reviewer to be familiar with the language/subject. Maybe in this case a future contributor will want to add domains for suborganizations of the American Statistical Association. A really terrible example of this is [`OneBookShelf.xml`](https://github.com/EFForg/https-everywhere/blob/7e4193f45aeb4633e79cdfa56abe1178bf746b39/src/chrome/content/rules/OneBookShelf.xml) which I originally wrote.

PR #12840, opened 2017-09-29
[Racked.com] Update ruleset.

bardiharborow:
The CORS issue (#7383) seems to have been resolved.

### DIFF
--- a/src/chrome/content/rules/Racked.com.xml
+++ b/src/chrome/content/rules/Racked.com.xml
@@ -1,16 +1,46 @@
 <!--
 	For other Vox Media coverage, see Vox.com.xml.
 
-	Problematic subdomains:
+	Refused hosts in *racked.com:
 
-		- www *
+		- comments
+		- comments.boston
+		- www.chicago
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Racked.com.xml
```diff
@@ -1,16 +1,46 @@
...
+		- comments.miami (no https)
+		- comments (no https)
+		- www.ny (no https)
+		- m.ny (no https)
+		- comments.philly (no https)
+		- comments.ny (no https)
+		- www.philly (no https)
+		- comments.sf (no https)
+		- m.sf (no https)
+		- comments.vegas (no https)
```
J0WI: Please group the list by errors instead of writing `(no https)` to each line. Please be also more specific what "no https" means.

bardiharborow: To me `(no https)` means port 443 is refused at TCP level, but port 80 works, though I can see how that would be ambiguous.

PR #12853, opened 2017-09-29
[IEEE-mismatches] Move functional hosts to main ruleset.

bardiharborow:


J0WI:
The main ruleset is also `default_off`...

bardiharborow:
@J0WI I've made additional tweaks, and IEEE.xml now passes the fetch test. I'm happy to enable the ruleset, but do not have acess to a IEEE account, and hence I cannot log in to check if [trac #15653](https://trac.torproject.org/projects/tor/ticket/15653) still applies. This should be safe to merge now.

src/chrome/content/rules/IEEE.xml
```diff
@@ -130,14 +113,6 @@
 		<test url="http://www.ieee.org/go/shop" />
 		<test url="http://www.ieee.org/go/shop/" />
```
J0WI: I think we can also remove these extra rules

bardiharborow: Yeah those URLs seem to behave better now.

PR #12854, opened 2017-09-29
[CERN.ch-problematic] Move functional hosts to main ruleset.

bardiharborow:


J0WI:
Please have a look at Travis, there are some errors in the CERN rules.

bardiharborow:
@J0WI Travis is resolved.

src/chrome/content/rules/CERN.xml
```diff
@@ -71,6 +73,7 @@
 	ˢ Self-signed
 	ᵘ Untrusted root
 	ˣ Mixed css
+	ᵈ Redirects to login.cern.ch, valid cert
```
J0WI: So why you removed them as targets when they work?

bardiharborow: Those targets serve different content over HTTPS (a redirect to login.cern.ch rather than the relevant minisite), several targets are already disabled for that reason, see the block above this one, although having a closer look, those sites seem to be considered "nonfunctional" rather than "problematic", so I'll shuffle the comments around a bit.

PR #12859, opened 2017-09-29
Initialize the rulesets only after we are sure which storage engine w…

Hainish:
…e'll be using

This ensures that the rulesets are not initialized more than once from default.rulesets, and wraps all the initialization of rulesets into background.initialize.  Initialize is only called after we're sure what storage engine we'll be using.

This also ensures that the enableMixedRulesets is loaded from the correct storage engine before loading rulesets.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -572,7 +572,8 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse){
     ruleset.active = message.object.active;
     sendResponse(true);
   } else if (message.type == "add_new_rule") {
-    all_rules.addNewRuleAndStore(message.object).then(sendResponse);
+    all_rules.addNewRuleAndStore(message.object);
+    sendResponse(true);
     return true;
   } else if (message.type == "remove_rule") {
     all_rules.removeRuleAndStore(message.object);
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -24,30 +24,6 @@ function initialize(){
   });
 }
 
-/**
- * Load a file packaged with the extension
- *
- * @param url: a relative URL to local file
- */
-function loadExtensionFile(url, returnType) {
-  var xhr = new XMLHttpRequest();
-  // Use blocking XHR to ensure everything is loaded by the time
-  // we return.
-  xhr.open("GET", chrome.extension.getURL(url), false);
-  xhr.send(null);
-  // Get file contents
-  if (xhr.readyState != 4) {
-    return;
-  }
-  if (returnType === 'xml') {
-    return xhr.responseXML;
-  }
-  if (returnType === 'json') {
-    return JSON.parse(xhr.responseText);
-  }
-  return xhr.responseText;
-}
-
 
 // Load in the legacy custom rulesets, if any
 function load_legacy_custom_rulesets(legacy_custom_rulesets){
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -4,14 +4,15 @@
 
 let all_rules = new rules.RuleSets();
 
-function initialize(){
-  initializeStoredGlobals().then(() => {
-    all_rules.initialize();
+async function initialize(){
+  await store.initialize();
+  await initializeStoredGlobals();
+  all_rules.initialize();
 
-    // Send a message to the embedded webextension bootstrap.js to get settings to import
-    chrome.runtime.sendMessage("import-legacy-data", import_settings);
-  });
+  // Send a message to the embedded webextension bootstrap.js to get settings to import
+  chrome.runtime.sendMessage("import-legacy-data", import_settings);
 }
+initialize();
 
 
 // Load in the legacy custom rulesets, if any
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -4,7 +4,7 @@
 
 let all_rules = new rules.RuleSets();
 
-async function initialize(){
+async function initialize() {
   await store.initialize();
   await initializeStoredGlobals();
   all_rules.initialize();
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -2,25 +2,14 @@
 
 (function(exports) {
 
-var all_rules, ls;
-
-try{
-  ls = localStorage;
-} catch(e) {
-  ls = {setItem: () => {}, getItem: () => {}};
-}
+let all_rules = new rules.RuleSets();
 
 function initialize(){
-  // Rules are loaded here
-  all_rules = new rules.RuleSets(ls);
-
-  initializeStoredGlobals(() => {
-    all_rules.addFromJson(loadExtensionFile('rules/default.rulesets', 'json'));
+  initializeStoredGlobals().then(() => {
+    all_rules.initialize();
 
     // Send a message to the embedded webextension bootstrap.js to get settings to import
     chrome.runtime.sendMessage("import-legacy-data", import_settings);
-
-    loadStoredUserRules();
   });
 }
 
@@ -44,23 +33,25 @@ var httpNowhereOn = false;
 var showCounter = true;
 var isExtensionEnabled = true;
 
-function initializeStoredGlobals(cb){
-  store.get({
-    httpNowhere: false,
-    showCounter: true,
-    globalEnabled: true,
-    enableMixedRulesets: false,
-    legacy_custom_rulesets: []
-  }, function(item) {
-    httpNowhereOn = item.httpNowhere;
-    showCounter = item.showCounter;
-    isExtensionEnabled = item.globalEnabled;
-    updateState();
+function initializeStoredGlobals(){
+  return new Promise(resolve => {
+    store.get({
+      httpNowhere: false,
+      showCounter: true,
+      globalEnabled: true,
+      enableMixedRulesets: false,
+      legacy_custom_rulesets: []
+    }, function(item) {
+      httpNowhereOn = item.httpNowhere;
+      showCounter = item.showCounter;
+      isExtensionEnabled = item.globalEnabled;
+      updateState();
 
-    rules.settings.enableMixedRulesets = item.enableMixedRulesets;
-    load_legacy_custom_rulesets(item.legacy_custom_rulesets);
+      rules.settings.enableMixedRulesets = item.enableMixedRulesets;
+      load_legacy_custom_rulesets(item.legacy_custom_rulesets);
 
-    cb();
+      resolve();
+    });
   });
 }
 
@@ -95,7 +86,6 @@ chrome.webNavigation.onCompleted.addListener(function() {
   updateState();
 });
 
-var USER_RULE_KEY = 'userRules';
 // Records which tabId's are active in the HTTPS Switch Planner (see
 // devtools-panel.js).
 var switchPlannerEnabledFor = {};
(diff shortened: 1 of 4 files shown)


cowlicks:
This is great! I like the idea of separating the backgrounds' initialization step into its own function. This will let us orchestrate our complicated initialization process much more easily.

There are a few things about this, that I don't understand yet. But I think it could be re-organized to be a little easier to understand, and to make make future development easier. Here's how I would recommend doing that:

It seems there are ~3 sort of seperable things that need initialization in this order:
* storage (choosing `chrome.storage.local` or `sync`)
* rulesets, which depends on storage.
* background stuff from `initializedStoredGlobals`, which depends on storage, but does it depend on rulesets?

If that is roughly correct, then we could make the initialization flow much more clear. Currently initialization happens, like, everywhere. So it is hard to track down initialization issues. It would be nice to have one main initialization entrypoint. Like in a `background.initialize` function. Inside this function, We should be able to do something like:
```javascript
// in background.js
function initialize() {
  store.initialize().then(stored => {
    rules.initialize(stored);  // rules after
    initializeStoredGlobals(stored);
  });
}
```
This way, it is clear what order we want things initialized in. It is clear what parts depend on what other parts. And it is easier to debug because if  there is an initialization error, we should be able to find it by starting in this function, and tracing it.

I'll leave some comments about how I would move things around to accomodate this suggestion.

chromium/background.js
```diff
@@ -2,6 +2,28 @@
...
+} catch(e) {
+  ls = {setItem: () => {}, getItem: () => {}};
+}
+
+function initialize(){
+  // Rules are loaded here
+  all_rules = new rules.RuleSets(ls);
+
+  initializeStoredGlobals(() => {
+    all_rules.addFromJson(loadExtensionFile('rules/default.rulesets', 'json'));
```
cowlicks: Could this be moved out of this callback, and into the outer function?

chromium/store.js
```diff
@@ -9,14 +9,18 @@ function setStorage(store) {
...
+      setStorage(chrome.storage.local);
+    } else {
       setStorage(chrome.storage.sync);
-      background.initializeStoredGlobals();
     }
+    background.initialize();
   });
+} else {
+  setStorage(chrome.storage.local);
+  background.initialize();
```
cowlicks: Instead of calling `background.initialize` here, could export an initialization function for this module. Then in `background.initialize` call `store.initialize`?

cowlicks: Since `background.js` is the last thing to load, I think it makes more sense to orchestrate the initialization there.

chromium/background.js
cowlicks: If that is possible, maybe an API like this would be nice.
```javascript
// in background.initialize
all_rules = rules.initialize();
```

chromium/background.js
```diff
@@ -2,6 +2,28 @@
 
 (function(exports) {
 
+var all_rules, ls;
+
+try{
+  ls = localStorage;
+} catch(e) {
+  ls = {setItem: () => {}, getItem: () => {}};
```
cowlicks: Any idea what this `try`/`catch` is for? When is `localStorage` not defined?

chromium/background.js
```diff
@@ -118,32 +119,40 @@ chrome.webNavigation.onCompleted.addListener(function() {
...
 
+var USER_RULE_KEY = 'userRules';
+// Records which tabId's are active in the HTTPS Switch Planner (see
+// devtools-panel.js).
+var switchPlannerEnabledFor = {};
+// Detailed information recorded when the HTTPS Switch Planner is active.
+// Structure is:
+//   switchPlannerInfo[tabId]["rw"/"nrw"][resource_host][active_content][url];
+// rw / nrw stand for "rewritten" versus "not rewritten"
+var switchPlannerInfo = {};
```
cowlicks: Could this stuff (`USER_RULE_KEY`, `switchPlannerEnabled/Info`) be moved into `initializeStoredGlobals`?

chromium/background.js
```diff
@@ -118,32 +119,40 @@ chrome.webNavigation.onCompleted.addListener(function() {
...
+function getStoredUserRules() {
   var oldUserRuleString = ls.getItem(USER_RULE_KEY);
   var oldUserRules = [];
   if (oldUserRuleString) {
     oldUserRules = JSON.parse(oldUserRuleString);
   }
   return oldUserRules;
-};
+}
 var wr = chrome.webRequest;
```
cowlicks: If it isn't a big deal, can we  remove this, or move it to the top of the file? I think alias' are fine, but it is nice to have them declared upfront so we know to look for them.

cowlicks:
I left a few comments, I feel the most important is the one about exporting an init function from storage, and calling it in the background, instead of vice versa.

chromium/background.js
koops76: I think so. But I think there are cases when it is defined, but trying to access the contents results in an error or are ignored (if localStorage is disabled).

chromium/background.js
Hainish: It can't be moved to the outer function, since the `addFromJson` function requires us to know the stored value of `enableMixedRulesets` in https://github.com/EFForg/https-everywhere/blob/d9a4db4cc330a9f48dabb5b0f51e12918b5448c2/chromium/rules.js#L244.

chromium/background.js
Hainish: Attempting to access `localStorage` in Firefox when cookies are disabled results in an exception: `SecurityError: The operation is insecure.`

chromium/background.js
Hainish: `initializeStoredGlobals` is more for those stored via the storage API.

Hainish:
Ready for review again

chromium/util.js
```diff
@@ -28,13 +28,38 @@ function log(level, str) {
...
+ * @param url: a relative URL to local file
+ */
+function loadExtensionFile(url, returnType) {
+  var xhr = new XMLHttpRequest();
+  // Use blocking XHR to ensure everything is loaded by the time
+  // we return.
+  xhr.open("GET", chrome.extension.getURL(url), false);
+  xhr.send(null);
+  // Get file contents
+  if (xhr.readyState != 4) {
```
koops76: `!==`

chromium/util.js
```diff
@@ -28,13 +28,38 @@ function log(level, str) {
...
+    return;
```
koops76: `return null;`

chromium/store.js
```diff
@@ -2,21 +2,43 @@
...
-setStorage(chrome.storage.local);
-if (chrome.storage.sync) {
-  chrome.storage.sync.set({"sync-set-test": true}, () => {
-    if(!chrome.runtime.lastError){
-      setStorage(chrome.storage.sync);
-      background.initializeStoredGlobals();
-    }
-  });
-}
+Object.assign(exports, {
```
koops76: `exports.initialize = initialize;`

chromium/store.js
```diff
@@ -2,21 +2,43 @@
 
 (function(exports) {
 
+let ls;
+
+try{
```
koops76: `try {`

chromium/background.js
```diff
@@ -2,45 +2,18 @@
...
-    return xhr.responseXML;
-  }
-  if (returnType === 'json') {
-    return JSON.parse(xhr.responseText);
-  }
-  return xhr.responseText;
-}
+let all_rules = new rules.RuleSets();
 
+async function initialize(){
```
koops76: `() {`

chromium/store.js
cowlicks: I'd probably go with:
```javascript
Object.assign.(exports, {initialize})
```
Because you DRY writing the function name. NBD to me either way.

chromium/background.js
```diff
@@ -2,45 +2,18 @@
...
+  await store.initialize();
+  await initializeStoredGlobals();
+  all_rules.initialize();
```
cowlicks: Just to make sure I understand this:  `await store.init();` will resolve/finish before `await initializeStoredGlbs();` is run?

chromium/rules.js
```diff
@@ -340,6 +350,70 @@ RuleSets.prototype = {
     util.log(util.INFO, 'done removing rule');
     return true;
   },
+  
+  /**
+  * Load stored user rules
```
cowlicks: This comment makes figuring out how this function is different from `loadStoredUserRules`. Can you make the distinction between "`get`" and "`load`"  more clear for these `StoredUserRules` functions.

chromium/background.js
```diff
@@ -708,8 +612,7 @@ async function import_settings(settings) {
 
 Object.assign(exports, {
   all_rules,
-  initializeStoredGlobals,
-  urlBlacklist,
+  urlBlacklist
```
cowlicks: I usually leave a trailing `,` on the final entry because it makes diffs less noisy when you append an entry later. NBD tho.

Hainish: I like this way better, syntactically.

chromium/background.js
koops76: @cowlicks Indeed.

koops76: Better do this:
```js
store.initialize()
  .then(initializeStoredGlobals)
  .then(() => all_rules.initialize());
```

Hainish: @cowlicks that's right, in ES6 the `await` keyword will wait for the returned promise will resolve before running the next line.  It's a nice way to avoid callback hell in ES6.

To do parallel, you can do:

```js
await Promise.all([asyncFunction1(), asyncFunction2()]);
```

Hainish:
Fixed - ready for review again

chromium/background.js
```diff
@@ -118,32 +87,14 @@ chrome.webNavigation.onCompleted.addListener(function() {
...
-  for (let rule of rules) {
-    all_rules.addUserRule(rule);
-  }
-  util.log(util.INFO, 'loaded ' + i + ' stored user rules');
-};
-
-loadStoredUserRules();
+// Records which tabId's are active in the HTTPS Switch Planner (see
+// devtools-panel.js).
+var switchPlannerEnabledFor = {};
```
cowlicks: It looks like this `switchPlanner` stuff got moved down, but is unchanged?

chromium/rules.js
```diff
@@ -340,6 +350,70 @@ RuleSets.prototype = {
...
+        // reloaded.
+        var oldUserRules = this.getStoredUserRules();
+        // TODO: there's a race condition here, if this code is ever executed from multiple
+        // client windows in different event loops.
+        oldUserRules.push(params);
+        // TODO: can we exceed the max size for storage?
+        store.localStorage.setItem(this.USER_RULE_KEY, JSON.stringify(oldUserRules));
+        resolve(true);
+      } else {
+        resolve(false);
```
cowlicks: It looks like `addNewRule` (the function that this replaced), isn't even asynchronous, so it should have probably just returned a value, instead of using a promise. That's a little out of scope of this PR I think though.

Maybe the function used to be asynchronous because it used `chrome.storage.*`? Any idea?

Continuing this tangent, why do we use both `localStorage` AND `chrome.storage`? Is which data we store where documented somewhere?

chromium/background.js
Hainish: That's right.  I wanted to move it closer to where it is actually used, for clarity.

chromium/rules.js
Hainish: I think we should change to `chrome.storage` for everything, but I wasn't involved in the initial writing of the WebExtension so there might be some historical reason for this.  @semenko may know better.

In the meantime I've changed this to a synchronous method.

cowlicks:
Cursory tests with this on #12624 pass for FF and Chrome

cowlicks:
LGTM, merge when tests pass.

Issue #11794, opened 2017-08-09
In-extension ruleset XML editor


ghost:
We could use [CodeMirror](https://codemirror.net/).

koops76:
@J0WI Maybe you'll add this to the kanban?

jeremyn:
FYI, but as far as I know, the rulesets are stored in the extension itself in a compacted/modified form (see https://github.com/EFForg/https-everywhere/issues/9958#issuecomment-315202693). So, they are not available to edit within the extension itself the way that you see them on GitHub or the Atlas.

On the other hand this could be part of good alternative to the soon to be deprecated `HTTPSEverywhereUserRules/` subdirectory (issue https://github.com/EFForg/https-everywhere/issues/10822 and elsewhere).

koops76:
@jeremyn I mean the editor for the custom rulesets.

gloomy-ghost:
Does #12863 fulfill this?

koops76:
@gloomy-ghost No, we need real custom ruleset functionality.

jeremyn:
This issue is technically resolved by https://github.com/EFForg/https-everywhere/pull/12863, which does introduce an XML ruleset editor in the extension. I guess though that @koops76 wants more robust XML highlighting, which is reasonable.

There must be some compromise solution between nothing and something as full-featured as CodeMirror. In https://github.com/EFForg/https-everywhere/pull/12863#issuecomment-336958528 I asked @Hainish for some guidelines but so far he hasn't responded to that request:

> * We should have some kind of XML highlighting. I know we talked about this before in https://github.com/EFForg/https-everywhere/issues/11794. CodeMirror is probably too heavyweight but there appear to be many options for small JS/browser-based syntax highlighters. I can try to find something if you can give me some requirements, e.g. must be 10kB or less in size etc.

koops76:
@jeremyn CodeMirror can be built to only include XML highlighting for lesser size.

koops76:
@jeremyn CodeMirror with XML highlighting included, minified and gzipped is only 54.1 KB.

gloomy-ghost:
50KB sounds ok to me, given that we used to ship megabytes of tests in the ruleset file for years without anyone complaining

Hainish:
Is there any reason why this has to be shipped with the release version, and not simply passed as a build flag like `--include-debug-xml-highliting`?

jeremyn:
[...]

PR #12863, opened 2017-09-30
Debugging rulesets

Hainish:
Depends on https://github.com/EFForg/https-everywhere/pull/12859

Adds a special page where ruleset testers can enter debugging rulesets into a textarea.

When multiple rulesets are added, these have to all be wrapped `<root>` and `</root>` opening and closing tag in order to be valid XML.

To access this page, navigate to the extension options UI, then click "General Settings" and type `Ctrl-Z`.  This should open a new tab to enter debugging rules into.

When you click `Save`, these rulesets should immediately take effect.

The UX is pretty rough, but since this is only to be used by testers I figure that's alright.  The reason this page is so hidden away is that I'd rather have users enter custom rulesets in the standard way, through the popup.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -68,8 +68,9 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
       updateState();
     }
     if ('debugging_rulesets' in changes) {
-      Object.assign(all_rules, new rules.RuleSets());
-      await all_rules.initialize();
+      const r = new rules.RuleSets();
+      await r.initialize();
+      Object.assign(all_rules, r);
     }
   }
 });
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -69,7 +69,7 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
     }
     if ('debugging_rulesets' in changes) {
       const r = new rules.RuleSets();
-      await r.initialize();
+      await r.loadFromBrowserStorage(store);
       Object.assign(all_rules, r);
     }
   }
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -67,6 +67,10 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
       isExtensionEnabled = changes.globalEnabled.newValue;
       updateState();
     }
+    if ('debugging_rulesets' in changes) {
+      Object.assign(all_rules, new rules.RuleSets());
+      await all_rules.initialize();
+    }
   }
 });
 
--- a/chromium/debugging-rulesets.js
+++ b/chromium/debugging-rulesets.js
@@ -1,0 +1,20 @@
+/* global sendMessage */
+
+"use strict";
+
+window.onload = function() {
+  document.querySelector("form").addEventListener("submit", save_debugging_rulesets);
+
+  sendMessage("get_option", { debugging_rulesets: "" }, item => {
+    document.getElementById("debugging-rulesets").value = item.debugging_rulesets;
+  });
+}
+
+function save_debugging_rulesets(e){
+  e.preventDefault();
+  sendMessage("set_option", { debugging_rulesets: document.getElementById("debugging-rulesets").value }, () => {
+    const saved = document.getElementById("saved");
+    saved.style.display = "block";
+    setTimeout(() => { saved.style.display = "none" }, 1000);
+  });
+}
--- a/chromium/debugging-rulesets.js
+++ b/chromium/debugging-rulesets.js
@@ -3,6 +3,7 @@
 "use strict";
 
 const debugging_rulesets_textarea = document.getElementById("debugging-rulesets");
+const changed = document.getElementById("changed");
 const default_title = "Debugging Rulesets";
 const unsaved_title = "* Debugging Rulesets";
 
@@ -25,10 +26,12 @@ function save_debugging_rulesets(e){
 
     document.title = default_title;
     debugging_rulesets_textarea.className = "";
+    changed.style.visibility = "hidden";
   });
 }
 
 function debugging_rulesets_changed(){
   debugging_rulesets_textarea.className = "unsaved";
   document.title = unsaved_title;
+  changed.style.visibility = "visible";
 }
--- a/chromium/debugging-rulesets.js
+++ b/chromium/debugging-rulesets.js
@@ -2,13 +2,13 @@
 
 "use strict";
 
-window.onload = function() {
-  document.querySelector("form").addEventListener("submit", save_debugging_rulesets);
+document.getElementById("debugging-rulesets-form").addEventListener("submit", save_debugging_rulesets);
 
-  sendMessage("get_option", { debugging_rulesets: "" }, item => {
-    document.getElementById("debugging-rulesets").value = item.debugging_rulesets;
-  });
-}
+sendMessage("get_option", { debugging_rulesets: "" }, item => {
+  let debugging_rulesets_textarea = document.getElementById("debugging-rulesets");
+  debugging_rulesets_textarea.value = item.debugging_rulesets;
+  debugging_rulesets_textarea.style.display = "block";
+});
 
 function save_debugging_rulesets(e){
   e.preventDefault();
--- a/chromium/debugging-rulesets.js
+++ b/chromium/debugging-rulesets.js
@@ -2,19 +2,33 @@
 
 "use strict";
 
+const debugging_rulesets_textarea = document.getElementById("debugging-rulesets");
+const default_title = "Debugging Rulesets";
+const unsaved_title = "* Debugging Rulesets";
+
+debugging_rulesets_textarea.addEventListener("input", debugging_rulesets_changed);
 document.getElementById("debugging-rulesets-form").addEventListener("submit", save_debugging_rulesets);
 
+document.title = default_title;
+
 sendMessage("get_option", { debugging_rulesets: "" }, item => {
-  let debugging_rulesets_textarea = document.getElementById("debugging-rulesets");
   debugging_rulesets_textarea.value = item.debugging_rulesets;
   debugging_rulesets_textarea.style.display = "block";
 });
 
 function save_debugging_rulesets(e){
   e.preventDefault();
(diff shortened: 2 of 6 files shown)


Hainish:
cc @jeremyn

koops76:
@Hainish Better just make a link.

koops76:
Maybe simply only allow access to this page by typing the URL directly.

Hainish:
@cowlicks whenever you have the chance to review this

chromium/debugging-rulesets.js
```diff
@@ -0,0 +1,20 @@
+/* global sendMessage */
+
+"use strict";
+
+window.onload = function() {
+  document.querySelector("form").addEventListener("submit", save_debugging_rulesets);
+
+  sendMessage("get_option", { debugging_rulesets: "" }, item => {
+    document.getElementById("debugging-rulesets").value = item.debugging_rulesets;
```
cowlicks: This is a race condition in the UI. The `<textarea id="debugging-rulesets">` element will display before it is populated with the data.

This is a little bit jarring visually, but the big problem is that it makes things harder to test with selenium. Since selenium  might get the textarea's value before it is populated.

One way to fix this, is to add `style="display:none"` to the appropriate element in the HTML. Then in this script, change the style to `"block"` when you've populated the data. There are helpers for this in `popup.js` [here](https://github.com/EFForg/https-everywhere/blob/master/chromium/popup.js#L207-L213).

An example of fixes  like this I've made are [here](https://github.com/EFForg/https-everywhere/pull/12624/commits/74763c83120686e9e1b8513878a440a35cbf8f85).

chromium/debugging-rulesets.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+}
+
+function save_debugging_rulesets(e){
+  e.preventDefault();
+  sendMessage("set_option", { debugging_rulesets: document.getElementById("debugging-rulesets").value }, () => {
+    const saved = document.getElementById("saved");
+    saved.style.display = "block";
+    setTimeout(() => { saved.style.display = "none" }, 1000);
```
cowlicks: Why the delay? Should you also erase `doc.getElementById(#debugging-rulesets).value`?

chromium/debugging-rulesets.js
```diff
@@ -0,0 +1,20 @@
+/* global sendMessage */
+
+"use strict";
+
+window.onload = function() {
+  document.querySelector("form").addEventListener("submit", save_debugging_rulesets);
```
cowlicks: This works fine, but it might be better to give the form an id, instead of with this queryselector, just to be defensive in-case another form is added some day.

chromium/debugging-rulesets.js
```diff
@@ -0,0 +1,20 @@
+/* global sendMessage */
+
+"use strict";
+
+window.onload = function() {
```
cowlicks: Why is this triggered by `window.onload`? Why not just have it execute as it loads?

chromium/background.js
```diff
@@ -60,6 +60,10 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
       isExtensionEnabled = changes.globalEnabled.newValue;
       updateState();
     }
+    if ('debugging_rulesets' in changes) {
+      Object.assign(all_rules, new rules.RuleSets());
+      await all_rules.initialize();
```
cowlicks: It might be better to do:
```javascript
r = new rules.RuleSets();
await r.initialize();
Object.assign(all_rules, r);
```

Otherwise, `all_rules` will temporarily be populated with empty data.

chromium/debugging-rulesets.js
Hainish: This introduces a delay before removing the "Saved" dialogue, since this should only display for a second.

Hainish: We don't want to clear the value of the `textarea`, since this should reflect what's stored in `debugging_rulesets` in the storage API.

chromium/debugging-rulesets.js
Hainish: I think it's a bit clearer, but I can remove this.

chromium/background.js
Hainish: Good call. Changing now.

Hainish:
Ready for review again.

jeremyn:
Thanks for adding this @Hainish. The basic functionality works okay for me.

Some usability improvements would be:

* Make it obvious if the user has unsaved changes pending, including the counterintuitive case where a user has restarted the browser and has prior text in the textbox. For example you could have red bold text at the bottom "There are unsaved changes".

* Don't require the user to add the `<root></root>` tag themselves. This is obscure. Silently autowrap the user's input with this tag as needed.

* Disable all spellcheck/autocomplete/etc in the textarea, for example see https://stackoverflow.com/a/254716.

* We should have some kind of XML highlighting. I know we talked about this before in https://github.com/EFForg/https-everywhere/issues/11794. CodeMirror is probably too heavyweight but there appear to be many options for small JS/browser-based syntax highlighters. I can try to find something if you can give me some requirements, e.g. must be 10kB or less in size etc.

* The `Ctrl-Z`-after-clicking-in-a-non-clickable-area activation method is too hidden. I think a button in `about:addons` labeled "Add custom rulesets" with an extra text warning like "for advanced users" would be fine.

* Add minimal documentation (a few sentences) in the debug tab.

As far as I know this is the only way to add custom rulesets, so it's not just a debugging tool. It should (eventually) be labeled and documented for general technical users, not just people involved with HTTPS Everywhere development.

jeremyn:
Despite my all my criticisms in the previous comment, it would be nice to get this feature included in the release versions as soon as possible even in its current minimal state.

chromium/background.js
```diff
@@ -60,6 +60,11 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
       isExtensionEnabled = changes.globalEnabled.newValue;
       updateState();
     }
+    if ('debugging_rulesets' in changes) {
+      const r = new rules.RuleSets();
```
cowlicks: This API has changed to requiring `store` gets passed in explicitly. This changed in #12938 

Also, #13213 might land before this, it changes the name of `initialize` to `loadFromBrowserStorage`.

chromium/debugging-rulesets.js
```diff
@@ -0,0 +1,20 @@
...
+  debugging_rulesets_textarea.value = item.debugging_rulesets;
+  debugging_rulesets_textarea.style.display = "block";
+});
+
+function save_debugging_rulesets(e){
+  e.preventDefault();
+  sendMessage("set_option", { debugging_rulesets: document.getElementById("debugging-rulesets").value }, () => {
+    const saved = document.getElementById("saved");
+    saved.style.display = "block";
+    setTimeout(() => { saved.style.display = "none" }, 1000);
```
cowlicks: This code block is not getting called for me. I think it is because we are calling `sendResponse` asynchronously [here](https://github.com/Hainish/https-everywhere/blob/0ecb5b4f66a2b94adb8250f3eaf8084b5c41ef67/chromium/background.js#L558). But we are not returning `true` in that listener. See the note about "`return true`" [here](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onMessage).

This is not a bug with this PR specifically, but it seems like a pretty bad thing. It is probably causing other problems. I'll fix it tomorrow and determine what the implications are.

Hainish:
Ready for review again @cowlicks.  I think we should address https://github.com/EFForg/https-everywhere/pull/12863#discussion_r145580273 in a separate PR.

koops76:
@Hainish There's no need for a secret keyboard shortcut. You can access the page directly by its extension URL.

Hainish:
https://github.com/EFForg/https-everywhere/pull/12863#issue-261807149

> The reason this page is so hidden away is that I'd rather have users enter custom rulesets in the standard way, through the popup.

koops76:
@Hainish You do not need to add it since it's possible to access it without any shortcuts: chrome-extension://gcbommkclmclpchllfjekcdonpmejbdp/debugging-rulesets.html

Hainish:
It's not about being possible to access it.  I don't want users to stumble upon it and rely on this for custom rulesets, rather than the normal UI custom ruleset flow.

koops76:
@Hainish I mean stumbling upon Ctrl-Z is easier than stumbling upon a URL.

koops76:
Ctrl-Z is a common keyboard shortcut.

jeremyn:
@Hainish Thanks for the update. Some further minor points:

I don't think the red border around the textarea and asterisk in the title are enough to indicated unsaved changes. For example there might be accessibility issues. There should be actual text indicating unsaved changes.

"instate" is a weird verb, to me. How about "activated"?

It would be nice if the Chromium-specific message only displayed if the browser is Chrome/Chromium.

I still disagree with hiding this option away. I guess you want to make sure this feature stays focused on testing and doesn't become a complex parallel path to creating rulesets. In that case you can just add a warning like "For testing only, not for production. Use at your own risk!" somewhere, and remove the hidden `Ctrl-Z` activator and add a link or button in `about:addons`.

Hainish:
@jeremyn how does the indication of change look now?  This feature is mainly because you requested it, and yes the reasoning is so that, as you mention, it "doesn't become a complex parallel path to creating rulesets."  I think hiding it away makes more sense, lowering the exposed options contributes to the simplicity of the extension.

jeremyn:
@Hainish The indication of change is much more obvious now, thanks.

The extension is equally complicated whether the option is hidden or not, because the feature is there either way. If you want to discourage people from relying on it, or from hassling you for usability improvements, or whatever, the best way to do that is to just say it like with the warning text I gave. Hiding the option sends an unclear message.

Hainish:
I'm not concerned about the extension being complicated.  I'm concerned with regular users stumbling upon an options page unexpectedly that they shouldn't have to be exposed to.

Regardless, I've added the warning and kept the `Ctrl-Z` workflow.  Hopefully this doesn't hinder the UX for anyone intending to actually do ruleset debugging.

Travis is experiencing problems right now, but this passed previously and I've only added a single warning text line, so I felt it safe to merge.

jeremyn:
The added warning text looks fine.

PR #12867, opened 2017-09-30
Create Kolibri.is.xml

koops76:


src/chrome/content/rules/Kolibri.is.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	SSL protocol error:
+		- nectar-ci
+-->
+
+<ruleset name="Kolibri.is">
+	<target host="kolibri.is" />
+	<target host="www.kolibri.is" />
+	<target host="img.kolibri.is" />
+	<target host="blog.kolibri.is" />
```
J0WI: `kort.kolibri.is`

src/chrome/content/rules/Kolibri.is.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	SSL protocol error:
+		- nectar-ci
```
J0WI: wildcard DNS

PR #12881, opened 2017-09-30
Update and rename coin-hive.com.xml to coinhive.com.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


src/chrome/content/rules/coinhive.com.xml
```diff
@@ -1,8 +1,13 @@
 <!--
+	Insecure Connection:
+		www.coinhive.com
```
J0WI: wildcard DNS

PR #12888, opened 2017-10-01
Re-activate and Update Statistics.gov.uk.xml

cschanaj:


src/chrome/content/rules/statistics.gov.uk.xml
```diff
@@ -1,60 +0,0 @@
...
-			<test url="http://www.statistics.gov.uk/index.php" />
-
-
-	<securecookie host="^\w" name=".+" />
-
-
-	<!--	Redirect drops forward slash and args:
-							-->
-	<rule from="^http://www\.statistics\.gov\.uk/.*"
-		to="https://www.gov.uk/government/statistics/announcements" />
```
J0WI: Shouldn't we keep that one?

src/chrome/content/rules/Statistics.gov.uk.xml
```diff
@@ -29,8 +29,14 @@
 		- www2.secure-transfer.statistics.gov.uk
 -->
 <ruleset name="Statistics.gov.uk (partial)">
+	<target host="www.statistics.gov.uk" />
+		<test url="http://www.statistics.gov.uk/about/national_statistics" />
 	<target host="neighbourhood.statistics.gov.uk" />
 	<target host="www.neighbourhood.statistics.gov.uk" />
+	
+	<rule from="^http://www\.statistics\.gov\.uk/.*"
+		  	to="https://www.gov.uk/government/statistics/announcements" />
```
J0WI: That behavior seems wrong: `http://www.statistics.gov.uk/invalid`

cschanaj: Do you mean that I should also include the exclusion in the old ruleset?

J0WI: Yes, or you change the rule to only match specific urls.
And please remove is the `.*` in the expression.

J0WI:
Please check the test coverage exposed by Travis.

cschanaj:
@J0WI Travis passing now. thanks!

PR #12894, opened 2017-10-01
Update MacWorld.com.xml

koops76:


src/chrome/content/rules/MacWorld.com.xml
```diff
@@ -1,42 +1,19 @@
-<!--
-	For other IDG coverage, see IDG.se.xml.
-
-
-	Nonfunctional subdomains:
-
-		- (www.)? ¹
-		- forums ²
-		- images	(404, CN: *.https.internapcdn.net)
```
J0WI: WFM, please add it as target.

koops76:
@J0WI All subdomains are working over HTTPS and I think the certificate is wildcard, maybe simply add `*.macworld.com`?

PR #12895, opened 2017-10-01
Delete ssl-observatory.dtd

koops76:
SSL Observatory is dead.


koops76:
ping @Hainish @cowlicks.

Hainish:
This is coming back at some point.  We're working on certificate introspection in WebExtensions.  We don't want to remove this, as it will be removed in transifex as well if we do.

PR #12896, opened 2017-10-01
Delete https-everywhere.properties

koops76:
Unused?


koops76:
ping @Hainish @cowlicks.

PR #12910, opened 2017-10-02
Delete simple.py

koops76:


cowlicks:
Do you think it is possible that we should still be using this for rule validation? Would you mind figuring out the last place this was used, and why we stopped using it? This might require some spelunking with git.

koops76:
ping @Hainish @cowlicks.

cowlicks:
@koops76 I've looked at the PR. But would you mind responding to https://github.com/EFForg/https-everywhere/pull/12910#issuecomment-334554947 so I know I can merge safely?

koops76:
@cowlicks I'm not sure.

Hainish:
Looks good, thanks @koops76

PR #12925, opened 2017-10-03
Added ruleset for bundeswehrkarriere.de

Jochen-A-Fuerbacher:


jeremyn:
- [ ] Add `target` for https://bewerbung.bundeswehrkarriere.de
- [ ] Remove the `securecookie` or limit it to the non-`^` `target`s. There is another domain https://statistik.bundeswehrkarriere.de/ that returns a 403 for both HTTPS and plain HTTP, so we don't need to include it here, but a general `securecookie` might break behavior there.
- [ ] Change the `ruleset` `name` to `BundeswehrKarriere.de`
- [ ] Change the filename to `BundeswehrKarriere.de.xml`

Also it looks like there is another group of domains for `bundeswehr-karriere.de`, with a dash. If you want, you can make a new ruleset for those and include it in this pull request, and have the two files refer to each other.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/12925/commits/0358e1303210beb365e76f0f8f90a81219a1e0da to add some new subdomains for `bundeswehr-karriere.de`. Also, I agree with you adding https://statistik.bundeswehrkarriere.de , though because it gets a `403` for the root URL, I've asked you to add a `test` for it.

jeremyn:
Thanks, merged.

PR #12927, opened 2017-10-03
Fixed mediamarkt.de ruleset.

Jochen-A-Fuerbacher:


src/chrome/content/rules/mediamarkt.de.xml
```diff
@@ -39,6 +36,7 @@ software-download.mediamarkt.de/: (52, 'Empty reply from server')
 stagenew.mediamarkt.de ³
 swdltest.mediamarkt.de ³
 tarife.mediamarkt.de ⁶
+tracking.mediamarkt.de/: (60: 'SSL certificate problem')
```
J0WI: That's because of an incomplete cert chain.

PR #12933, opened 2017-10-03
add rule for Ze.tt

numismatika:
Youth magazine from Zeit Online

### DIFF
--- a/src/chrome/content/rules/Ze.tt.xml
+++ b/src/chrome/content/rules/Ze.tt.xml
@@ -1,0 +1,14 @@
+<!-- for more Zeit-Online rules  see Zeit.de.xml -->
+<ruleset name="Ze.tt">
+
+	<target host="ze.tt" />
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Ze.tt.xml
```diff
@@ -0,0 +1,14 @@
+<!-- for more Zeit-Online rules  see Zeit.de.xml -->
```
J0WI: Please also update the reference list there

J0WI:
`Zeit.de.xml` has no test coverage, so tests are failing.

numismatika:
just pushed a new iteration to see how much work for that general rewrite of the Zeit.de rule is required.
maybe we are lucky with a few more additional target hosts.

numismatika:
@J0WI  we have coverage onthe Zeit.de domain again

J0WI:
Thanks!

PR #12938, opened 2017-10-03
Add a require function

cowlicks:
NB: This is blocked until #12859 is merged, since that removes the `background` dependency in `store.js`.

This PR adds a "`require`" function. This is nice because:
* We have no more mystery globals floating around.
* We can import files like `rules.js` using node, this lets us unittest them our code without  a browser.

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -7,7 +7,7 @@ let all_rules = new rules.RuleSets();
 async function initialize() {
   await store.initialize();
   await initializeStoredGlobals();
-  await all_rules.initialize();
+  await all_rules.initialize(store);
 
   // Send a message to the embedded webextension bootstrap.js to get settings to import
   chrome.runtime.sendMessage("import-legacy-data", import_settings);
@@ -597,7 +597,7 @@ async function import_settings(settings) {
     });
 
     Object.assign(all_rules, new rules.RuleSets());
-    await all_rules.initialize();
+    await all_rules.initialize(store);
 
   }
 }
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -2,6 +2,11 @@
 
 (function(exports) {
 
+const rules = require('./rules'),
+  store = require('./store'),
+  util = require('./util');
+
+
 let all_rules = new rules.RuleSets();
 
 async function initialize() {
@@ -607,4 +612,4 @@ Object.assign(exports, {
   urlBlacklist
 });
 
-})(typeof exports == 'undefined' ? window.background = {} : exports);
+})(typeof exports == 'undefined' ? require.scopes.background = {} : exports);
--- a/chromium/bootstrap.js
+++ b/chromium/bootstrap.js
@@ -1,0 +1,9 @@
+"use strict";
+
+function require(module) {
+  if (module.startsWith('./') && !module.hasOwnProperty(module.slice(2))) {
+    return require.scopes[module.slice(2)];
+  }
+  throw new Error('module: ' + module + ' not found.');
+}
+require.scopes = {};
--- a/chromium/incognito.js
+++ b/chromium/incognito.js
@@ -2,6 +2,10 @@
 
 (function(exports) {
 
+const util = require('./util'),
+  background = require('./background'),
+  rules = require('./rules');
+
 // This file keeps track of incognito sessions, and clears any caches after
 // an entire incognito session is closed (i.e. all incognito windows are closed).
 
@@ -67,4 +71,4 @@ if (chrome.windows) {
 
 Object.assign(exports, {});
 
-})(typeof exports == 'undefined' ? window.incognito = {} : exports);
+})(typeof exports == 'undefined' ? require.scopes.incognito = {} : exports);
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -3,6 +3,8 @@
 
 (function(exports) {
 
+const util = require('./util');
+
 // Stubs so this runs under nodejs. They get overwritten later by util.js
 if (typeof util == 'undefined' || typeof global != 'undefined') {
   Object.assign(global, {
@@ -680,4 +682,4 @@ Object.assign(exports, {
   RuleSets
 });
 
-})(typeof exports == 'undefined' ? window.rules = {} : exports);
+})(typeof exports == 'undefined' ? require.scopes.rules = {} : exports);
--- a/chromium/rules.js
+++ b/chromium/rules.js
@@ -206,7 +206,8 @@ function RuleSets() {
 
 RuleSets.prototype = {
 
-  initialize: async function() {
+  initialize: async function(store) {
+    this.store = store;
     this.ruleActiveStates = store.localStorage;
     this.addFromJson(util.loadExtensionFile('rules/default.rulesets', 'json'));
     this.loadStoredUserRules();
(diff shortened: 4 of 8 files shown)


koops76:
Keep in mind that this is a temporary measure until we will start to use ES6 modules. I would prefer doing it after WebExtensions transition.

chromium/bootstrap.js
```diff
@@ -0,0 +1,7 @@
+function require(module) {
```
koops76: `'use strict';`

chromium/bootstrap.js
```diff
@@ -0,0 +1,7 @@
...
+  if (module.startsWith('./') && !module.hasOwnProperty(module.slice(2))) {
+    return require.scopes[module.slice(2)];
+  }
+  throw 'module: ' + module + ' not found';
```
koops76: ```throw new Error(`module: ${module} not found`);```

chromium/incognito.js
```diff
@@ -2,6 +2,9 @@
 
 (function(exports) {
 
+let util = require('./util'),
```
koops76: `const`

chromium/rules.js
```diff
@@ -3,6 +3,8 @@
 
 (function(exports) {
 
+let util = require('./util');
```
koops76: `const`

chromium/store.js
```diff
@@ -2,6 +2,8 @@
 
 (function(exports) {
 
+let background = require('./background');
```
koops76: `const`

chromium/background.js
```diff
@@ -712,4 +716,4 @@ Object.assign(exports, {
   urlBlacklist,
 });
 
-})(typeof exports == 'undefined' ? window.background = {} : exports);
+})(typeof exports == 'undefined' ? require.scopes.background = {} : exports);
```
koops76: Use `module.exports`. `exports` doesn't always work.

chromium/bootstrap.js
koops76: `/* exported require */`

chromium/bootstrap.js
```diff
@@ -0,0 +1,7 @@
+function require(module) {
+  if (module.startsWith('./') && !module.hasOwnProperty(module.slice(2))) {
```
koops76: What does `!module.hasOwnProperty(module.slice(2))` do?

cowlicks: it strips off the leading `./` of the module name.  Since we do ` require('./background');` but the property on the object is called `background`. This is to be compatible with node's require, which uses relative paths.

chromium/bootstrap.js
cowlicks: I do that in the `.eslintrc`, not here

cowlicks:
@koops76 ES6 modules are not compatible with node, but this is.

chromium/background.js
cowlicks: `typeof module.exports` would throw a `ReferenceError` because `module` is not defined.

cowlicks:
I'm going to revert the incognito.js changes and move them into a seperate PR where I refactor the file to add unittests.

chromium/rules.js
```diff
@@ -1,19 +1,8 @@
...
-if (typeof util == 'undefined' || typeof global != 'undefined') {
-  Object.assign(global, {
-    util: {
-      DBUG: 2,
-      INFO: 3,
-      WARN: 5,
-      log: ()=>{},
-    }
-  });
-}
```
cowlicks: @RReverser ping

cowlicks:
@RReverser would you like to review this PR? It makes the code much easier to use with node.

RReverser:
I can take a look after the weekend. In the meanwhile, any chance to get reviews on some of my PRs?

cowlicks:
Merging so #12940 can proceed.

RReverser:
This API is changing so frequently... It's understandable given that it's seen by HTTPS Everywhere as internal implementation detail, but it's really hard to keep Cloudflare Auto HTTPS Rewrites up-to-date and properly tested while it continues changing *sigh*

Could we consider moving rules rewriter to its own independent package with a fixed API for non-extension consumers to use?

koops76:
Better standardize our rules format.

RReverser:
Rules format is already specified and I'm parsing it separately, but for testing it's important to use HTTPS Everywhere own rewriter implementation to compare results.

koops76:
@cowlicks Do ytou think we should try making a Webpack build of the extension now (using CommonJS).

cowlicks:
@RReverser I think that is a great idea. I'm sorry the API has been shifting a lot recently. However I believe we are in a much better place to maintain a stable API now, since we can now work with pieces of HE more independently. And we are better able to test them, so that we are actually able to now verify that we have a stable API that is providing deterministic results.

With recent changes, you should be able to add unittests that verify we have not broken the API's you are using. I'd welcome a pull request like this.

cowlicks:
@koops76 once we have transitioned to WebExtensions, I don't think there will be a need for webpack because the build process will be extremely simple. We will just need to link the locales, and rulesets into the `chromium/` directory (which should be renamed). Then the zipped version of tho this directory will be the crx file.

koops76:
@cowlicks We will no longer need to use `require.scopes` since Webpack supports standard CommonJS. It also may allow to stop using the current build system completely, replacing it with a more flexible one.

koops76:
@cowlicks See https://github.com/salsita/chrome-extension-skeleton for an example.

koops76:
We could have used Rollup, but we can't use ES6 modules yet.

cowlicks:
@koops76 I think webpack is a lot more complicated than what we need. Our "build" is very simple, we just copy locales and rulesets into the extension directory. We could probably even skip copying the locales by just relocating them into the `chromium/` directory.

Webpack/rollup would actually complicate development because we'd have to "rebuild" every time we edit js. When we are able to load the unpacked extension from `chromium/`, like I'm talking about in #13190, we won't have to do that.

We currently have *no* third party dependencies at runtime so we don't need a bundler to handle those.

Our own module level dependencies are not something we should rely on a bundler resolve for us. We need to understand what the actual dependencies and requirements of our code are, in order to reason about it, and improve it, and fix bugs.

RReverser:
> With recent changes, you should be able to add unittests that verify we have not broken the API's you are using. I'd welcome a pull request like this.

Problem is I don't understand the implicit dependencies each module (or at least `rules.js`) requires very well myself yet as it was changed again.

Given that you worked on this PR, could you perhaps document the expected objects in each integration point? (e.g. what the newly added `store` is expected to be for external consumers to pass in).

koops76:
@cowlicks We have complicated JS which should be broken into *more* modules, not less in my opinion. We will need to merge the modules into one bundle file while building if we want to keep the extension modular. We could use AMD and Require.js instead of CommonJS and Webpack but *we need a module loader now* and the need for it will only increase with time.

koops76:
@cowlicks Can we at least try a few options in separate branches before deciding which one to keep?

cowlicks:
@RReverser I should definitely add some documentation. I'll do that ASAP. And change the name of the `initialize` function, it initializes `RuleSets` from browser storage. Maybe `loadFromBrowserStorage` would be a better name. 

`rewriter.js` does not use browser storage, so it should not need to use `initialize` or whatever it is renamed to. Again, sorry for the confusion. I'll stub out some unit tests for `rewriter.js` so we'll notice any API breakages. And the tests can be built on later.

@koops76 You are absolutely correct. We need to break things into more modules. I'm trying to make that easier. The codebase, as of this pull request, now has all dependencies managed explicitly. So it is ready to be broken up into more modules, and we won't have to worry about side effects.

The `require` function that we now have, allows us work in both node and browser environments *without* a bundling step. It allows us to be more modular, because we now have all implicit dependencies managed explicitly.

PR #12940, opened 2017-10-03
Add unit tests

cowlicks:
NB: this depends on #12938 

This lays out the basics for adding unittests for code in `chromium/`.

To run them, cd into the `chromium/`. Run `npm install`. Then run `npm test`.

Todo: put tests in a better place. Hook in to travis.

### DIFF
--- a/chromium/package.json
+++ b/chromium/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "https-everywhere",
+  "version": "1.0.0",
+  "description": "",
+  "main": "utils.js",
+  "dependencies": {},
+  "devDependencies": {
+    "chai": "^4.1.2",
+    "coveralls": "^3.0.0",
+    "mocha": "^4.0.0",
+    "nyc": "^11.2.1"
+  },
+  "scripts": {
+    "test": "mocha",
+    "cover": "nyc --reporter=html --reporter=text mocha",
+    "report": "nyc report --reporter=text-lcov | coveralls"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "GPL-2.0+"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ tokenkeys.py*
 Vagrantfile
 .cache/
 geckodriver.log
+node_modules/
+coverage/
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,13 @@ matrix:
   fast_finish: true
   include:
     - env: INFO="lint"
+      language: node_js
+      node_js:
+        - "node"
+    - env: INFO="unittests"
+      language: node_js
       node_js: node
+        - "node"
     - env: INFO="rules"
     - env: INFO="fetch"
     - env: INFO="preloaded"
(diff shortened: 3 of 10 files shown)


koops76:
Also move ESLint there.

chromium/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "chai": "^4.1.2",
+    "mocha": "^4.0.0"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "mocha"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
```
koops76: `"license": "GPLv2+"`

chromium/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "background.js",
+  "dependencies": {
+    "chai": "^4.1.2",
+    "mocha": "^4.0.0"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "mocha"
```
koops76: "test": "eslint . && mocha"

chromium/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "chromium",
+  "version": "1.0.0",
+  "description": "",
+  "main": "background.js",
+  "dependencies": {
+    "chai": "^4.1.2",
+    "mocha": "^4.0.0"
```
koops76: This should go into `"devDependencies"`.

chromium/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "chromium",
+  "version": "1.0.0",
```
koops76: `"version": "2017.9.12"`

chromium/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "",
+  "main": "background.js",
+  "dependencies": {
+    "chai": "^4.1.2",
+    "mocha": "^4.0.0"
+  },
+  "devDependencies": {},
```
koops76: Also add ESLint to `"devDependencies"`.

chromium/package.json
cowlicks: eslint tests are run separately by travis

chromium/package.json
cowlicks: Thank you. Don't worry about reviewing this too hard yet, it is still a WIP

chromium/package.json
koops76: No, I mean move it here.

cowlicks: This is the main entry point for unit testing. So I don't think it should also run the linter.

This is a problem if, for example, I want to use `npm test` to run the unit tests on travis. It would require that eslint be installed, and it would run the linter twice. So if there is a lint failure both the travis linting job, and the unit test job would fail, regardless of whether the unit tests actually passed.

koops76: Remove the old ESLint then.

koops76: Also you can make two scripts, `lint` and `unittest` and run both from `test` script, allowing to test either ESLint, unit tests or both.

cowlicks: linting is not specific to the either the node package context, nor the extension context. So I don't think it belongs here, in a node specific context.

koops76: @cowlicks It's specific to Node context. It can only be done with Node.js.

cowlicks:
It's working, with coverage. Here is an example of the badge in my repo:
coverag-bagge

cowlicks:
Merging this. It should not effect https-everywhere at all.

PR #12942, opened 2017-10-04
Simplify import_settings function in background.js

koops76:


koops76:
ping @Hainish @cowlicks.

cowlicks:
@koops76 I prefer async function expressions  over promises when possible.  believe that @Hainish does too. No need to worry about "simplifying" them. Thank you.

PR #12943, opened 2017-10-04
Simplify initialize function in background.js

koops76:


koops76:
ping @Hainish @cowlicks.

cowlicks:
I don't think this simplifies the function, it adds more line noise, and more lines of code.

PR #12958, opened 2017-10-04
[Spring Cleaning] Remove domains with no assigned nameservers.

bardiharborow:
This PR removes all `default_off` rulesets that contain only hosts which do not have a valid authoritative nameserver (i.e. are expired or very broken). The list is [here](https://gist.github.com/bardiharborow/cf2ae6b0e15af171fb39504a338e20f1).

To verify, run the following command on a clean workspace, and ensure the rulesets deleted matches this PR:

```bash
curl https://gist.githubusercontent.com/bardiharborow/cf2ae6b0e15af171fb39504a338e20f1/raw/1ad6c751f6c3b2cd36f788f3804685649179524b/dead.txt | while read domain; do host -t ns $domain > /dev/null || egrep -l -m 1 "<target host=\"[0-9a-zA-Z.-]*$domain\" ?\/>" *.xml | xargs rm; done
```


koops76:
@bardiharborow Do this to all rulesets, not only the disabled ones.

gloomy-ghost:
For the record. disk42.com, waeplus.co.uk, pandoblog.com have valid DNS records, but since they are all not working in https and have been disabled for a long time I am ok to remove them.

bardiharborow:
@gloomy-ghost Hmm. My list was computer generated, so I'm really not sure how that slipped through, but anyway.

PR #12960, opened 2017-10-05
Create Ace-book.net.xml

gsinp:
The site has active mixed content blocking caused by http stylesheet requests (eg. on the homepage) however this doesn't interfere with usage of the site or cause any noticeable breakage.


jeremyn:
What's tricky about this PR is that most of the site's content is behind a login, so we can't easily test it. In particular we can't easily test whether the mixed content blocking or `securecookie` cause problems or not after (or during) logging in.

I'm inclined to say we shouldn't make a PR for this site. @gloomy-ghost @J0WI what do you think?

pipboy96:
Closing abandoned PR, replacement coming soon.

PR #12993, opened 2017-10-06
Added ruleset for offenesparlament.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/offenesparlament.de.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="OffenesParlament.de">
```
J0WI: Please add comments for:
```
beta.offenesparlament.de
www.offenesparlament.de

Issue #9993, opened 2017-05-22
Rewrite Python/bash tools in Node.js


strugee:
As CONTRIBUTING.md notes, newer utilities are written to be run under Node.js, but most of them are written in Python or bash. Various comments in this issue tracker have pointed out that it'd be really good if we could rewrite all these in JavaScript/Node instead.

In particular this would be a huge boon for contributor experience if we could rewrite enough such that ruleset contributions required little to no implicit global dependencies that need to be installed via `install-dev-dependencies.sh`.

jeremyn:
I'm not totally on board with switching to Node, but I've discussed it with @Hainish and I agree there is value in making the utils codebase match the (JavaScript) extension codebase.

If we're going to rewrite existing utils for Node, then a first step would be to install Node/npm in `install-dev-dependencies.sh` and document its basic use in `CONTRIBUTING.md`.

PR #7136, opened 2016-09-26
Specify the exact python version, don't just assume it's python 2.7

Hainish:

Issue #6749, opened 2016-09-04
Move from Python2 to Python3


Foorack:

capt2101akash:
Can you please describe the whole issue regarding this bug

Foorack:
@capt2101akash Will explain more in detail later today. Essentially you are unable to push if you uninstall python2.7. During push git runs the tests which currently require python 2.7.

Foorack:
@capt2101akash Hello again. Sorry for the delay.

First off I would like to clarify this is not a issue but rather an improvement request.
For more information what I am referring to:
https://github.com/EFForg/https-everywhere/blob/master/test/rules.sh#L30
As one can see it is currently hard-coded to fail if python2.7 is not installed on the system.

jeremyn:
It's possible that the `python2.7` requirement in `rules.sh` is unnecessary, for example `check_rules.py` in `fetch.sh` [here](https://github.com/EFForg/https-everywhere/blob/master/test/fetch.sh#L22) just executes with `python`. On the other hand, `fetch.sh` is used mainly for Travis testing and it's possible for simplicity in the script that we just assume the environment uses 2.7, since we control that environment.

@fuglede @Hainish @J0WI Can any of you give us any information you have about Python 2 vs Python 3 support for this project?

Hainish:
This definitely needs to happen at some point, since Python 2.7 is being [EOLed in 2020](https://pythonclock.org/).  However this is not a priority for us currently, as other development requirements (such as the transition to WebExtensions for Firefox) take precedent.

jeremyn:
If `python2.7` is actually required then I agree that it's a bug anywhere that we refer to `python` or `/usr/bin/python`.

renyouguo:
Python3 support unicode by default, may fix the encoder errors just like: https://travis-ci.org/EFForg/https-everywhere/jobs/161551245#L358

cowlicks:
Partial fixes in #12996

Hainish:
# Python 2 to 3 Migration Work

I've been working on modifying [`zipfile_determinstic.py`](https://github.com/Hainish/https-everywhere/blob/ec42ff8e0ceefbc78807b0d40c9e45e528c5fbaa/utils/zipfile_deterministic.py) (which is sourced from the python standard library) https://github.com/EFForg/https-everywhere/pull/14515 and have it successfully building the extension in python 3.6 deterministically.  I also think we should move forward with #12996 and other efforts which are more specific about the python version supported.

[...]

PR #12996, opened 2017-10-06
Make some buildtools py2/3 compatible

cowlicks:


utils/make-json.py
```diff
@@ -57,13 +57,13 @@
...
     # FF version can find it.
-    xpath_ruleset(tree)[0].attrib["f"] = os.path.basename(fi).decode(encoding="UTF-8")
+    xpath_ruleset(tree)[0].attrib["f"] = os.path.basename(fi)
 
     for target in targets:
         # id is the current length of the rules list - i.e. the offset at which
         # this rule will be added in the list.
         json_output["targets"][target].append(len(json_output["rulesetStrings"]))
-    json_output["rulesetStrings"].append(etree.tostring(tree))
+    json_output["rulesetStrings"].append(etree.tounicode(tree))
```
cowlicks: This file is where behavior changes would be if there are any.

cowlicks:
It looks like this directly reverts some changes in #7136

Hmm I think the correct path forward is probably to adapt our python scripts to be 2/3 compatible, and update their shebangs as we do so.

The hardest change would be to `utils/zipfile_deterministic.py`, instead of updating that file it might be better to try to just use a linux utility. We could also just bundle it with `create_xpi.py` and use that directly as a script, since there are no external dependencies beyond that.

koops76:
@cowlicks I don't think we need the builds to be deterministic since Tor Project makes their own builds.

Bisaloo:
Related: https://github.com/EFForg/https-everywhere/issues/9993

Hainish:
Superseded by more recent py3.6 rewrite.

Hainish:
https://github.com/EFForg/https-everywhere/issues/6749

PR #12999, opened 2017-10-07
Remove excessive returns in background.js

koops76:
Returns are excessive here because of `else if`s.


koops76:
ping @Hainish @cowlicks.

cowlicks:
These are to signal that we want to call `sendResponse` asynchronously.  See the docs here https://developer.chrome.com/extensions/messaging

PR #13007, opened 2017-10-07
Create NoiseTrade.xml

wesdawg:


J0WI:
The tests are expected to fail, because the server sends an incomplete certificate chain: `firefox --new-instance --profile $(mktemp -d) https://noisetrade.com/`. Sadly we cannot mitigate this issue, so please `default_off` the ruleset.

PR #13013, opened 2017-10-07
Added ruleset for patientenbeauftragter.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/patientenbeauftragter.de.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Invalid certificate:
+		staging.patientenbeauftragter.de
+-->
+<ruleset name="Die Beauftragte der Bundesregierung für die Belange der Patientinnen und Patienten sowie Bevollmächtigte für Pflege">
```
J0WI: Please use a short ruleset name.

PR #13026, opened 2017-10-09
Update NASA.gov.xml

wesinator:
* Re-enable and cleanup
* Enable `sservi` sub


youdly:
Looks like false error. apd440.gsfc.nasa.gov and gracefo.jpl.nasa.gov work for me.
`curl -I -L https://gracefo.jpl.nasa.gov` gives error but firefox is fine.

wesinator:
#15974

PR #13027, opened 2017-10-09
Update Tutorialspoint.xml

wesdawg:


src/chrome/content/rules/Tutorialspoint.xml
```diff
@@ -1,13 +1,6 @@
-<!--
-	
-	Mismatch: ^tutorialspoint.com
```
J0WI: `codingground.tutorialspoint.com`

wesdawg: Fixed.

J0WI:
Thanks!

PR #13053, opened 2017-10-12
Update Chromium.org.xml

wesdawg:
Simplify ruleset

### DIFF
--- a/src/chrome/content/rules/Chromium.org.xml
+++ b/src/chrome/content/rules/Chromium.org.xml
@@ -1,14 +1,6 @@
 <!--
 	For other Google coverage, see GoogleServices.xml.
 
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Chromium.org.xml
```diff
@@ -31,20 +25,12 @@
...
-	<!--	Complications:
-				-->
-	<target host="chromium.org" />
-
-
-	<securecookie host="^codereview\.chromium\.org$" name=".+" />
-
+	
+	<target host="crbug.com" />
+	<target host="www.crbug.com" />
```
gloomy-ghost: crbug.com is already HSTS preloaded so there is no need to add it.

PR #13054, opened 2017-10-12
Directly set "checked" attribute in popup.js

koops76:


koops76:
@cowlicks Review please.

PR #13056, opened 2017-10-12
Replace parseInt with type cast in popup.js

koops76:


cowlicks:
@koops76 in the future please use more descriptive commit messages, and include some motivation for the change in the description.

Issue #7978, opened 2016-12-26
Clarify license so GitHub can detect it


jeremyn:
GitHub can detect and display a project license if the included LICENSE file matches "a short list of known licenses", according to https://github.com/blog/2252-license-now-displayed-on-repository-overview.

HTTPS Everywhere's license file is [here](https://github.com/EFForg/https-everywhere/blob/c8c4ae35a2f27745480cd52d5f042b5c8bd32b19/LICENSE.txt), and it's complicated so GitHub can't figure it out. I think [this](https://github.com/EFForg/https-everywhere/blob/c8c4ae35a2f27745480cd52d5f042b5c8bd32b19/LICENSE.txt#L27-L28) is saying the overall license is GPLv3.

I propose we change `LICENSE.txt` to the standard GPLv3 text so GitHub can detect it, and move the rest of the license information to another file.

jeremyn:
Pinging @Hainish .

jeremyn:
@Hainish Is the license for HTTPS Everywhere actually GPLv3, and we just need to change the text around to make that clear, or is there some actual question about the overall license?

koops76:
#13062

brainwane:
(Some discussion ongoing in #13062.)

jeremyn:
@Hainish I would like to have at least one open issue or pull request reflecting an ongoing discussion, or at least reflecting that my concerns have not been fully addressed.

brainwane:
@jeremyn Per https://github.com/EFForg/https-everywhere/pull/13062#issuecomment-347375367 and your response there: it sounds as though you do not think an EFF lawyer's word would be authoritative here in responding to your concerns, since as you see it this is less a matter of legal compliance than of the developer experience of other potential downstreams and contributors and your take on open source citizenship norms, which are social and technical issues more than legal issues. Am I right? I want to understand so I can potentially get people with the right expertise into any future discussion (whether it's in a GitHub issue, a mailing list thread, or elsewhere).

jeremyn:
@brainwane I'm not sure what you are asking. Staff lawyers generally oppose open source licensing and support the status quo if their client isn't facing significant legal trouble. I don't expect an EFF staff lawyer to lead an effort to clarify licensing problems for the benefit of contributors and/or downstreams.

brainwane:
@jeremyn let's unpack what you've just said or implied.

[...]

PR #13062, opened 2017-10-13
Modify license and readme files to let GitHub detect the license (#7978)

koops76:


koops76:
@Hainish This needs legal review.

jeremyn:
We should use GPL v3+, not GPL v2+, right? In any case GPL v2+ "reduces" to GPL v3+ because anyone can immediately re-release GPL v2+ stuff as v3+.

koops76:
@jeremyn 
```
HTTPS Everwyhere:
Copyright © 2010-2017 Electronic Frontier Foundation and others
Licensed GPL v2+
```

jeremyn:
@koops76 Oh, hm, I guess `Net license` is referring to some `Net` code, rather than saying `The net license for this entire project`. Confusing. So I agree the current license is GPL v2+.

Also, contrary to my earlier argument, there is an advantage for GPL v2+ instead of v3+: v2 and v3 are incompatible, so v3+ can't be released with v2-only code. See https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility. However we should probably use GPL v3+ anyway unless the EFF has some specific reason not to.

Hainish:
The language in the current `LICENSE.txt` suffices and has been put through legal review.

jeremyn:
@Hainish [`LICENSE.txt`](https://github.com/EFForg/https-everywhere/blob/c68d1b52a11fc9996030575f6ba87fd14725356b/LICENSE.txt) says GPL v2+ but there's no copy of GPL v2 anywhere in the repository. Ordinarily this would be in `LICENSE.txt`, or `COPYING` or similar, which are the sort of places where GitHub looks for the license (see the [`Licensee`](https://github.com/benbalter/licensee/blob/325de73c4bfc59ab981904a33f324e8ddf9b2249/docs/what-we-look-at.md#detecting-the-license-file) documentation, which [GitHub uses](https://help.github.com/articles/licensing-a-repository/#detecting-a-license) to detect the license).

What's going on with that?

jeremyn:
There's also a typo at [the top](https://github.com/EFForg/https-everywhere/blob/c68d1b52a11fc9996030575f6ba87fd14725356b/LICENSE.txt#L1) of the current, very short `LICENSE.txt` file:

> HTTPS Everwyhere:

This will sound snarky, but how much time did the legal team spend reviewing this if they missed such a simple thing?

Hainish:
You're right, it does sound snarky :laughing:

They reviewed new content, in person, with me.  That doesn't mean they're making edits in the files themselves.  That line was committed on 2010-06-22, so just as the project started.

Hainish:
Despite the FSF recommendations I don't think it's necessary to include the full text of the GPL in every repo that is under that license.  It just adds clutter.

Hainish:
There's no ambiguity about "which GPLv2" we're talking about, it's already clear, adding the full body of the document is unnecessary.

jeremyn:
@Hainish The stated purpose of this PR and of my earlier issue https://github.com/EFForg/https-everywhere/issues/7978 is to allow GitHub to automatically detect the license. GitHub still can't do that so the issue is still not solved.

Are you saying that the current `LICENSE.txt` is functionally identical to the text of the GPL v2 and that the only reason you don't want to update it is to save a few thousand bytes? Or is the current `LICENSE.txt` not functionally identical?

Hainish:
Modifying the contents of `LICENSE.txt` to fit some GitHub display requirement is silly.  We should state our `LICENSE.txt` with as much nuance as is descriptive and necessary.  Having a badge that is reductive is not helpful.

jeremyn:
GitHub only fails detection if you're doing something really unusual, like we're currently doing with the non-standard `LICENSE.txt`. I'm not sure why doing licensing the standard way is such a sticking point, unless there is actually some ambiguity about the overall license. I'm also slightly annoyed that you refer to my request to standardize licensing for GitHub as "silly" and hope you will refrain from that framing going forward.

Could you please answer my earlier questions:

> Are you saying that the current `LICENSE.txt` is functionally identical to the text of the GPL v2 and that the only reason you don't want to update it is to save a few thousand bytes? Or is the current `LICENSE.txt` not functionally identical?

Hainish:
Someone contributing to the project seeing a reductive license in the GitHub display may mistakenly assume licensing restrictions that aren't true if we shape the `LICENSE.txt` file to suit GitHub's requirements.  I didn't say your request was silly, so please don't interpret it that way.  I do think it's unwise to be overly reductive when being specific is preferable.

jeremyn:
I'm still confused here. You haven't directly answered my earlier questions, and now you seem to be saying that you're happy we're breaking the GitHub license detection functionality because GitHub's "reductive" display might cause some problem. Could you walk me through an example licensing problem your hypothetical contributor might have?

Again, GitHub doesn't have any weird criteria for its detection functionality. If you put the standard license text in any one of a number of standard places, GitHub will find it. There's very little "shaping" required.

Hainish:
Sure.  So someone, seeing that the repo is GPLv2+, goes in and modifies code in https://github.com/EFForg/https-everywhere/blob/457dc980f2f36e6fc6a4675fbd78f23413cb9595/utils/create_xpi.py assuming this will be under the GPL as well.  In fact, it will not be.  This will be under the Python Software Foundation License Version 2.

Another example: someone does a mass search-and-replace on files in `utils/issue-format-bot/` using `sed`, assuming these changes will be GPLv2+ because of the reductive label.  In fact, these will be under the AGPL v3+.

Stating that we're a GPLv2 project *a dicto simpliciter* amounts to misleading our contributors as to the licensing specifics of various aspects of our project.  So my question to you is this: why would you want to leave these details out?  If the answer is just for the sake of a GitHub badge, which is itself contributing to the misleading dynamic, then that's not an acceptable qualification.

jeremyn:
@Hainish You're describing a bunch of problems here.

Most basically, one integral software application or library has one license. Saying that different integral parts of an application have different licenses makes no sense. It's like if I gave you a tic-tac-toe program and told you that the game is GPLv2 but the part that draws the "X"s is MIT. It's not a matter of argument, it's a matter of dictionary definition of words.

Normally what you want to do when using third-party code is to, if possible, relicense that code under your license so you can release everything together under your chosen license. If you can't relicense the third-party code but you still want to use it, then generally you need to enforce some strong separation between your code and the third-party code, and even then you have to consider the third-party code's license.

***

So let's look at the Python stuff. Recent Python 2 software licenses [are compatible with GPLv2](https://docs.python.org/2/license.html), so the simple thing to do is to relicense the Python stuff under GPLv2 and call the whole thing a *derived work*. (Note the Python license might have other requirements to satisfy, for example the 2.7.14 license at my link has the clause:

> 3. In the event Licensee prepares a derivative work that is based on or
   incorporates Python 2.7.14 or any part thereof, and wants to make the
   derivative work available to others as provided herein, then Licensee hereby
   agrees to include in any such work a brief summary of the changes made to Python
   2.7.14.

I don't know what version of the Python license applies to the code HTTPS Everywhere uses, so I don't know if it has a similar clause.)

The other option is to claim that HTTPS Everywhere is a *combined work* and argue that the Python stuff and the other stuff is sufficiently separate. That would be a tough argument given how integrated and customized the build process is with the rest of the main code. There have been a lot of [local modifications to `utils/create_xpi.py`](https://github.com/EFForg/https-everywhere/commits/master/utils/create_xpi.py). I also don't know why you would make this argument given the compatibility of the GPLv2 and recent Python 2 licenses.

Those are the only options: it's a derived work (which seems clear to me) and all under GPLv2, or it's a combined work with strong separations (a tough sell) and therefore can be separately licensed. Having one derived work with different parts licensed separately isn't, logically or semantically, an option.

(And, side note, you may want to investigate whether you're violating the Python license by not including the "brief summary of the changes" that that license requires, if that clause in the relevant Python license, unless this summary exists somewhere and I just haven't seen it.)

***

The issue bot code has two different, separate problems. One is that it isn't actually integrated with the main code and there's no reason it should be in the same repository except for programmer convenience. The other is that it uses the Affero GPLv3 license which is [not compatible with GPLv2](https://www.gnu.org/licenses/license-list.en.html#AGPLv3.0).

***

In the end what I'm saying is that the licensing isn't clear, in fact I would say it's incoherent. Your last comment seems to argue that it's some kind of "GPLv2-but-here-are-some-necessary-details" license, which, maybe (?), but, at least, that's not the same as GPLv2.

I'm not sure what to propose as a single to-do item here. Best case, the EFF would go through all the code and split everything up into separate repositories and relicense the necessary third-party code so that each of these repositories can be put under an appropriate license in the usual simple way. I'm not sure what else to suggest beyond that.

brainwane:
Hey @jeremyn -- I appreciate your diligence in wanting to make sure the project's 100% in compliance with relevant licenses. (I'm not a lawyer, and I'm writing this assuming you are not a lawyer; if you are a lawyer then please let me know so I can set my expectations better!)

I'd like to make sure I understand the situation. You've suggested that staying with a single Git repository and a single `LICENSE` file (which I'll call the "monorepo" approach) is less optimal than splitting up different utilities/projects into different repositories, each of which has its own `LICENSE` file (which I'll call the "multirepo" approach). @Hainish is pretty sure that the stipulations of the GPL are such that the restrictiveness only applies to code which is included in the same program. That is to say, since the utilities are separated out, this is okay - they are not directly included in the compiled HTTPS Everywhere extension delivered to end-users. You, on the other hand, see the application as a unified whole, and -- well, I wouldn't want to put words in your mouth, but you're concerned about respecting the licenses while combining and deriving works.

My personal take is that the basic "this repo/application also includes some other work under other licenses" approach is something I'm used to seeing, e.g., in [Zulip](https://github.com/zulip/zulip/blob/master/LICENSE#L15), [a MediaWiki installation](https://www.mediawiki.org/wiki/Special:Version), and [GitLab](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/LICENSE). But I could be wrong on the details regarding what makes combined/derived works ok.

In order to make sure #13600 (the external dependencies policy) is on a good foundation, we'll need to have a clear architectural decision on whether we're going multirepo or staying with a monorepo. @Hainish is in favor of staying with a monorepo for multiple reasons -- it's easier to manage a single repository in various ways, for instance. I don't really have an opinion yet. I think it'd be useful to have that particular architectural discussion in a separate issue or on the mailing list, outside of the discussion of this particular PR, if you're amenable to that -- let me know.

Regardless, as I see it, the most important question in the discussion above is: is the status quo ok, legally speaking? At least, that's how I'm reading your most recent comment above. After all, if we're not legally in the clear, then we have to make a change to come into compliance.

So, if you're ok with @Hainish's and my assessment, and the existing legal review and sign-off @Hainish mentioned above, then I figure we're good. If you'd like for one of the EFF staff attorneys to take a fresh look and review the current licensing situation and check whether it needs clarifying, with especial attention to the issues you have brought up, I can ask for that. :) Please let me know.

Thanks.

jeremyn:
@brainwane Thanks for your detailed response. I'm not a lawyer. If I am directly in error on some point, please correct me.

As I said, there are several problems here.

***

The most serious problem is the apparent confusion about the difference between code licensed to the project and the project's license to others, for example with `utils/create_xpi.py`. The EFF has a license from the Python Software Foundation (PSF) to use this code. The EFF modifies and incorporates this code into the larger HTTPS Everywhere work. The confusion is that @Hainish argues (https://github.com/EFForg/https-everywhere/pull/13062#issuecomment-341248522) that this interferes with the EFF's ability to redistribute all of HTTPS Everywhere under a single license. Instead he seems to think that the EFF is forced to distribute this particular, integral part of the project under a PSF license while the rest goes out under GPL. I argue that this belief is wrong, because I think that everything including the modified PSF code can be redistributed under a single GPL license. In fact it is not clear what the sort of "hybrid" license he argues for, would look like.

(Think of a simpler example. You write a function `foo` and release it under MIT. I incorporate that into a program called `foobar`. Can I release `foobar` under the GPL? Yes.)

***

The current "multirepo" situation is confusing. The utilities are not well separated, for example `utils/create_xpi.py` is integral to HTTPS Everywhere while `utils/issue-format-bot/` is not, but they are both in `utils`.

I've read that there are some organizations that are so uncomfortable with the AGPL that they won't even allow programmers to have AGPL code on their systems. So including the issue bot with the rest of the code would be a blocker for them.

(From a technical standpoint, git submodules are a good way to make "monorepos" look like a "multirepo", if needed. I do that in https://github.com/jeremyn/https-everywhere-dev-environment/tree/master/roles, for example. And some things like the issue bot simply don't need to be attached at all.)

***

For the external dependencies issue https://github.com/EFForg/https-everywhere/issues/13600, note that the current dependency CodeMirror is licensed [under MIT](https://codemirror.net/LICENSE). The MIT license specifically requires including the license text in full, so the current "shorthand" approach in our `LICENSE.txt` would not be sufficient.

***

For the PSF license, in [`LICENSE.txt`](https://github.com/EFForg/https-everywhere/blob/38f55dc9018509b0c1788193b4d8e0d71341e4b0/LICENSE.txt) we say:

> The build system incorporates code from Python 2.6,
> Copyright © 2001-2006 Python Software Foundation
> Python Software Foundation License Version 2

It seems however that PSF licenses are by minor version, for example the 2.7.14 license is at https://docs.python.org/2/license.html#psf-license-agreement-for-python-release. This matters because of this clause

> 3. In the event Licensee prepares a derivative work that is based on or
   incorporates Python 2.7.14 or any part thereof, and wants to make the
   derivative work available to others as provided herein, then Licensee hereby
   agrees to include in any such work a brief summary of the changes made to Python
   2.7.14.

We can't describe the changes if we don't know what version we started from. Also I don't think this summary of changes is included anywhere in the project. 

***

You, @brainwane, wrote:

> Regardless, as I see it, the most important question in the discussion above is: is the status quo ok, legally speaking? At least, that's how I'm reading your most recent comment above. After all, if we're not legally in the clear, then we have to make a change to come into compliance.

This isn't how we should look at it. The question "is the status quo ok, legally speaking?" is more about, is the EFF likely to be sued as things are now? I mean legally, other than the PSF and CodeMirror issues mentioned above, the EFF is probably in the clear, and even then, I guess the odds that the PSF would sue the EFF for failing to honor some minor licensing requirement are really low.

But, part of being a good open source citizen is making your licensing friendly to potential downstream redistributors. (For example, would Google feel comfortable including HTTPS Everywhere in Chrome?) The project represents many thousands of volunteer hours, contributed under open source terms, and I think it's a matter of respect to the volunteers to go above the bare legal minimum.

jeremyn:
What's the current status of this issue at the EFF?

Unfortunately, I can't reopen this PR because the original contributor has deleted their GitHub account, but the problem is still active.

PR #13071, opened 2017-10-14
[dnscrypt.org] New ruleset

youdly:


gloomy-ghost:
Thanks for the ruleset, but dnscrypt.org is already [preloaded](https://hstspreload.com/api/v1/status/dnscrypt.org).

PR #13107, opened 2017-10-14
Create Al-Mostafa.info.xml

ihave100openPRs:


jeremyn:
- [ ] Remove the entire top comment. This site has a wildcard DNS (see the response from `dig *.al-mostafa.info`) so any subdomain will match. I checked some of the top comment URLs and I guess that everything except `^` and `www` is actually a mirror of `al-mostafa.com`, not `al-mostafa.info`. So, let's just keep `^` and `www`.
- [ ] Delete the `securecookie` or restrict it to `www`, because of the failing subdomains.
- [ ] Change the filename to `Al-Mostafa.info.xml`. I understand that lower-case `al-` is valid capitalization for Arabic transliteration, but one purpose of capitalization in this project is to make ruleset filenames ASCII-sorted.

ihave100openPRs:
@jeremyn Done!

PR #13127, opened 2017-10-14
Create SageMath.org.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


Bisaloo:
The tests should warn that we already have rules for some of those targets. I can't find anything about it. Can you, @I-have-50-open-PRs-to-HTTPS-EVERYWHERE ?

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@Bisaloo Yes https://travis-ci.org/EFForg/https-everywhere/jobs/287891828 

['src/chrome/content/rules/SageMath.org.xml', 'src/chrome/content/rules/Sagemath.org.xml'] failed case-insensitivity testing.

Will make a new PR

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@Bisaloo Here's the new PR https://github.com/EFForg/https-everywhere/pull/13142

Bisaloo:
> ['src/chrome/content/rules/SageMath.org.xml', 'src/chrome/content/rules/Sagemath.org.xml'] failed case-insensitivity testing.
>
> Rules exist with identical case-insensitive names, which breaks some filesystems.

Hum, that's not what I was talking about. We are supposed to have a test that prevents duplicated targets (see `utils/duplicate-whitelist.txt`). In this case, if you named your ruleset differently, it could have gone through. This shouldn't happen.

Issue #13091, opened 2017-10-14
Audit default_off="self-signed" rulesets.


bardiharborow:
Type: other

The following `default_off="self-signed"` rulesets pass a `curl -I https://host/` scan for at least one host ([full host list](https://gist.github.com/bardiharborow/8cfb01eb478d7123eb0face869929b36)).

Fails fetch test (needs manual review):

- [x] [Channelx.biz.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/Channelx.biz.xml) (#13126)
- [x] [Creative_Virtual.com.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/Creative_Virtual.com.xml) (#13135)
- [x] [Digital-Transactions.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/Digital-Transactions.xml) (#13136)
- [x] [Files.poulsander.com.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/Files.poulsander.com.xml) (#13138)
- [x] [How_to_Box.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/How_to_Box.xml) (#13137)
- [x] [Macmillan.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/Macmillan.xml) (#13091)
- [x] [MegaStock.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/MegaStock.xml) (#13174)
- [x] [PCS.com.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/PCS.com.xml) (#13139)
- [x] [PL-Casual-Dating.xml](https://github.com/EFForg/https-everywhere/blob/4f045a2644e7281bd28d8641511a37fd13c7b695/src/chrome/content/rules/PL-Casual-Dating.xml) (#13175)
[...]

PR #13145, opened 2017-10-14
Bulk re-enabling of default_off="self-signed" rulesets.

bardiharborow:
This PR re-enables all the rulesets from #13091 which require either only minimal simplification or straight re-enabling.


J0WI:
Thanks!

PR #13148, opened 2017-10-14
Create tnstimes.org.xml

katzrkool:


J0WI:
Thanks for your contribution! The rule LGTM, but our code style requires to use tabs to indent lines. Could you please change this?

katzrkool:
Should be fixed now!

J0WI:
Nope, now you are using two spaces.

katzrkool:
I think I finally got it, my text editor was being weird and not doing indentations right. But this should work.

J0WI:
Thanks!

PR #13158, opened 2017-10-15
Added ruleset for bundesregierung.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/bundesregierung.de.xml
```diff
@@ -0,0 +1,45 @@
+<!--
+	Invalid certificate:
+		akkreditierung.bundesregierung.de
+		www.cvd.bundesregierung.de
+		www.styleguide.bundesregierung.de
+
+	Not found:
```
J0WI: Sites without DNS records can be removed.

src/chrome/content/rules/bundesregierung.de.xml
```diff
@@ -0,0 +1,45 @@
+<!--
+	Invalid certificate:
```
J0WI: They can be redirected to their working alias.

J0WI:
Thanks!

PR #13160, opened 2017-10-15
Added ruleset for bnitm.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/bnitm.de.xml
```diff
@@ -2,6 +2,8 @@
 
 	<target host="bnitm.de" />
 	<target host="www.bnitm.de" />
+		<exclusion pattern="^http://(?:www\.)bnitm\.de/SEEIS201\d" />
```
J0WI: Please avoid non-capturing groups.

PR #13165, opened 2017-10-15
Added ruleset for bghm.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/bghm.de.xml
```diff
@@ -0,0 +1,44 @@
...
+
+	Not found:
+		entwicklung.extranet.bghm.de
+		rollin.extranet.bghm.de
+		www.receiver.bghm.de
+-->
+<ruleset name="BGHM.de">
+
+	<target host="bghm.de" />
+	<target host="*.bghm.de" />
```
J0WI: It's bad to have so many exclusions. I would prefer a list of valid targets to simplify this rule.

PR #13170, opened 2017-10-16
[Spring Cleaning] Remove more parked domains.

bardiharborow:
These domains are pointed at one of the following parking services and have been manually verified as nonfunctional.

* cimweb.com.au
* sedoparking.com
* parkingcrew.net
* ibspark.com


J0WI:
Thanks!

PR #13173, opened 2017-10-16
[Macmillan] Update and rename to Tor.com.xml.

bardiharborow:
#13091


src/chrome/content/rules/Tor.com.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+
+	Probmelatic hosts in *tor.com:
```
Bisaloo: Problematic

bardiharborow: Indeed. 😆

Issue #13176, opened 2017-10-16
Audit default_off="expired" rulesets.


bardiharborow:
Type: other

The following `default_off="expired"` rulesets pass a `curl -I https://host/` scan for at least one host:

Fails fetch test (needs manual review):

- [ ] [AboutUs-problematic.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/AboutUs-problematic.xml)
- [ ] [CableTV.com.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/CableTV.com.xml)
- [x] [CloudFlare_Challenge.com.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/CloudFlare_Challenge.com.xml)
- [ ] [Etre_Touchy.com.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/Etre_Touchy.com.xml)
- [ ] [EuroSmartz.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/EuroSmartz.xml)
- [ ] [Extabit.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/Extabit.xml)
- [ ] [Media-Proweb.de.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/Media-Proweb.de.xml)
- [ ] [Motigo.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/Motigo.xml)
- [ ] [MyFreeCopyright.com-problematic.xml](https://github.com/EFForg/https-everywhere/blob/0279990170739dc8c3e9d8cace9e2d4065898a85/src/chrome/content/rules/MyFreeCopyright.com-problematic.xml)
[...]

PR #13177, opened 2017-10-16
Bulk re-enabling of default_off="expired" rulesets.

bardiharborow:
This PR re-enables all the rulesets from #13176 which require either only minimal simplification or straight re-enabling.


src/chrome/content/rules/Adsurve.com.xml
```diff
@@ -3,20 +3,17 @@
...
 -->
-<ruleset name="adsurve.com (partial)" default_off="expired">
+<ruleset name="Adsurve.com (partial)">
 
 	<target host="adsurve.com" />
 	<target host="www.adsurve.com" />
 
 
-	<securecookie host="^adsurve\.com$" name=".+" />
+	<securecookie host="^\w" name=".+" />
```
Bisaloo: why `"^\w"` instead of `".+"` ?

bardiharborow: To be quite honest, I don't really understand what I'm doing with `securecookie`s, but as I understand it `^\w` secures cookies specific to a particular host, and `.+` secures domain-wide cookies that might be accessed by HTTP-only subdomains. This could be very wrong though. In this instance I haven't actually done any cookie testing and therefore I've tried to match the existing behaviour as closely as possible.

Bisaloo: As I understand it, `.+` secures cookies for all targets and `^\w` secures cookies for all targets with only alphabetic characters.

So they should be functionally equivalent (as long as you have have digits in your targets) but `^\w` won't be read as a trivial securecookie rule by the rewriter.

Only subdomains listed as targets can be affected by securecookie and rewrite rules.

But that's just my take on it, feel free to wait for another opinion.

koops76: @bisaloo Only the first character should be alphabetic.

koops76: In this case, better use `.+`.

bardiharborow: @Bisaloo [`^\w` matches any alpha**numeric** character including the underscore](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-word). This means that all cookies will be accepted, except those starting with dots – however it appears that my understanding of the difference between `example.com`and `.example.com` [no longer applies](https://stackoverflow.com/a/20884869/1414639). Perhaps @J0WI can clarify.

I've made the change to this ruleset though.

Bisaloo: About the leading dot, see https://www.eff.org/https-everywhere/rulesets#secure-cookies

> Non-HTTPS subdomains that are not listed as targets may depend on cookies set on the root domain, that we secure and stop them from accessing.

True but that doesn't change the fact that `^\w` and `.+` should be functionally equivalent.

src/chrome/content/rules/Digital_News_Agency.com.xml
```diff
@@ -1,41 +1,11 @@
...
-	<!--securecookie host="^\.digitalnewsagency\.com$" name="^_dna_session$" /-->
-
 	<securecookie host="^\.digitalnewsagency\.com$" name=".+" />
 
-
-	<rule from="^http://digitalnewsagency\.com/"
-		to="https://www.digitalnewsagency.com/" />
-
 	<rule from="^http://dna-bucket-dna\.cf\.rawcdn\.com/"
 		to="https://df7r7v78mcbtc.cloudfront.net/" />
```
J0WI: That looks a bit dangerous. Can you provide a test url for it? Otherwise you should remove this target.

koops76: @J0WI ```
> nslookup
Default Server:  UnKnown
Address:  192.168.0.1

> dna-bucket-dna.cf.rawcdn.com
Server:  UnKnown
Address:  192.168.0.1

Non-authoritative answer:
Name:    df7r7v78mcbtc.cloudfront.net
Addresses:  13.33.23.17
          13.33.23.146
          13.33.23.85
          13.33.23.82
          13.33.23.240
          13.33.23.121
          13.33.23.71
          13.33.23.35
Aliases:  dna-bucket-dna.cf.rawcdn.com
```

bardiharborow: @J0WI yeah, I'd actually initially removed it, but I added it back because I realised it's valid and it's a little beyond the scope of this PR to mess with individual rulesets too much. I've added a test URL, and I don't think there are any issues with this rewrite (possibly it's for third-parties, as I haven't seen any mixedcontent issues causing a need for the rewrite on the main site).

bardiharborow:
`craftysyntax.com` started timing out, so I've had to disable it again.

J0WI:
@bardiharborow it would be easier to review for me if you keep you old commit, so I just have to review the new changes.

PR #12452, opened 2017-09-03
Update chromium/incognito-cache-clearing.js

cschanaj:
ping @Hainish @cowlicks To clear domainBlacklist and urlBlacklist when all incognito sessions are destroyed.

## Step to reproduce the leak

1. From chrome://extensions, open the "Inspect views" for HTTPS Everywhere in a new session.
2. Toggle "Block all unencrypted requests" and open an Incognito Session
3. In the console, type `domainBlacklist`. the console should output something like

```
Set(0) {} // Empty Set
```

4. Visit http://http.badssl.com/ (connection blocked because of too many redirects as expected)
5. Close all Incognito Sessions, repeat Step 3 and observe

```
Set(1) {"http.badssl.com"}
```

The same also hold true for urlBlacklist. 

P.S. IMHO, just a minor issue because this will require physical access to the browser console and specific types of connections to work. Thanks.

PR #13189, opened 2017-10-17
Add 100% unit test coverage for incognito.js

cowlicks:
This  makes background.js depend on incognito.js, instead of incognito.js depending on background.js. This lets us test incognito.js, and it makes more sense for background.js to orchestrate imports.

This also adds unit tests that cover incognito.js 100%.

### DIFF
--- a/chromium/test/incognito_test.js
+++ b/chromium/test/incognito_test.js
@@ -1,0 +1,59 @@
+'use strict'
+
+const expect = require('chai').expect,
+  tu = require('./testing_utils'),
+  incognito = require('../incognito');
+
+describe('incognito.js', function() {
+  beforeEach(function() {
+    tu.stubber([
+      ['chrome.windows.onCreated.addListener', tu.Mock()],
+      ['chrome.windows.onRemoved.addListener', tu.Mock()],
+      ['chrome.windows.getAll', tu.Mock()],
+    ]);
+  });
+
+  describe('onIncognitoDestruction', function() {
+    beforeEach(function() {
+      incognito.state.incognito_session_exists = false;
+      this.callbackCalled = false;
+      this.callback = () => this.callbackCalled = true;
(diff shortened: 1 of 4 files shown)


koops76:
@cowlicks Make Incognito an ES6 class.

chromium/test/incognito_test.js
```diff
@@ -0,0 +1,59 @@
...
+    it('callback called when last incognito closed', async function() {
+      incognito.state.incognito_session_exists = true;
+      chrome.windows.getAll = func => func([]);
+
+      await this.instance.detect_incognito_destruction();
+      expect(incognito.state.incognito_session_exists, 'constant changed').to.be.false;
+      expect(this.callbackCalled).to.be.true;
+    });
+
+    it('detcts when an incognito window is created', function() {
```
koops76: detects

cowlicks:
@koops76 I think it'd be better to save coverting to classes for a separate PR.

cschanaj:
@cowlicks it seems that this re-introduce the issue of #12452 in `Version 2017.10.30` using Chromium on a debian platform. I got the following error message from the extension console

```nodejs
Uncaught (in promise) TypeError: Cannot read property 'onIncognitoDestruction' of null
    at detect_incognito_destruction (incognito.js:42)
    at <anonymous>
detect_incognito_destruction @ incognito.js:42
async function (async)
detect_incognito_destruction @ incognito.js:38
EventImpl.dispatchToListener @ extensions::event_bindings:403
publicClassPrototype.(anonymous function) @ extensions::utils:140
EventImpl.dispatch_ @ extensions::event_bindings:387
dispatchArgs @ extensions::event_bindings:260
dispatchEvent @ extensions::event_bindings:269
```

and the caches are not destroy after all incognito tabs are removed. Could you please take a took at the issue?

PR #13196, opened 2017-10-17
Update Dar-us-Salam-Texas-USA.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


jeremyn:
- [ ] Remove the `securecookie` tag. The problem is that I assume the cookie covers their store, and we can't test it without buying stuff from their store, and I don't want to break their store.

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@jeremyn Done

jeremyn:
Thanks, merged.

PR #13197, opened 2017-10-17
Update Yahoo.xml

wesdawg:
* Add `store` subdomain


youdly:
I suggest to list all the subdomains in `<target host="" />` instead of appending to the `<rule from="" to="" />` for readability.

jeremyn:
@wesdawg Your change to the `rule` looks fine. Please:

- [ ] Update the hash for `Yahoo.xml` in `ruleset-whitelist.csv` [here](https://github.com/EFForg/https-everywhere/blob/cc7792f5c598c674877418477eba6186225e6340/utils/ruleset-whitelist.csv#L3839). I ran `sha256sum` against the updated `Yahoo.xml` from your branch and it looks like the new hash is `da2fd160f5803ed73336a0fcd8df659b18857c624b22cd38b286edef13e6c73e`.

(@youdly I agree that we prefer not to use wildcards, but in this case I sympathize with @wesdawg wanting to make a quick change without rewriting the ruleset.)

jeremyn:
Thanks, merged.

PR #13213, opened 2017-10-18
Fix rewriter.js and test the API it uses.

cowlicks:
cc @RReverser 

I looked at `rewriter.js`. It was broken :( It happened when the packaged rulesets switched from xml to json in #12273.

I updated `rewriter.js` and did a quick test on a file that just contained `<a href="http://freerangekitten.com/">foo</a>` and it rewrote it as expected.

I also added tests for the API's that rewriter js uses: `addFromJson` (previously `addFromXml`) and `rewriteURI`.

### DIFF
--- a/chromium/test/rules_test.js
+++ b/chromium/test/rules_test.js
@@ -126,6 +126,10 @@ describe('rules.js', function() {
 
         assert.strictEqual(newuri, 'https://' + host + '/', 'protocol changed to https')
       })
+
+      it('does not rewrite unknown hosts', function() {
+        assert.isNull(this.rsets.rewriteURI('http://unknown.com/', 'unknown.com'));
+      })
     });
 
     describe('#potentiallyApplicableRulesets', function() {
--- a/chromium/test/rules_test.js
+++ b/chromium/test/rules_test.js
@@ -96,7 +96,14 @@ describe('rules.js', function() {
   })
 
   describe('RuleSets', function() {
-    let rules_json = JSON.parse('[{"name":"Freerangekitten.com","rule":[{"to":"https:","from":"^http:"}],"target":["freerangekitten.com","www.freerangekitten.com"]}]');
+    let rules_json = [{
+      name: "Freerangekitten.com",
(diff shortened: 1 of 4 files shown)


chromium/test/rules_test.js
```diff
@@ -96,14 +96,35 @@ describe('rules.js', function() {
...
+    });
+
+    describe('#rewriteURI', function() {
+      it('rewrites host added from json', function() {
+        let host = 'freerangekitten.com';
+        this.rsets.addFromJson(rules_json);
+
+        let newuri = this.rsets.rewriteURI('http://' + host + '/', host);
+
+        assert.strictEqual(newuri, 'https://' + host + '/', 'protocol changed to https')
```
RReverser: Could you also add a test for the no-match case, please? (it returns `null` as far as I remember)

chromium/test/rules_test.js
```diff
@@ -96,14 +96,35 @@ describe('rules.js', function() {
   })
 
   describe('RuleSets', function() {
+    let rules_json = JSON.parse('[{"name":"Freerangekitten.com","rule":[{"to":"https:","from":"^http:"}],"target":["freerangekitten.com","www.freerangekitten.com"]}]');
```
RReverser: Perhaps better to inline the object? Result would be the same but would look slightly more readable in the source (and with proper syntax highlighting)

chromium/test/rules_test.js
```diff
@@ -96,14 +96,46 @@ describe('rules.js', function() {
...
+      it('rewrites host added from json', function() {
+        let host = 'freerangekitten.com';
+        this.rsets.addFromJson(rules_json);
+
+        let newuri = this.rsets.rewriteURI('http://' + host + '/', host);
+
+        assert.strictEqual(newuri, 'https://' + host + '/', 'protocol changed to https')
+      })
+
+      it("does'nt rewrite unknown hosts", function() {
```
RReverser: a typo

PR #13223, opened 2017-10-19
Create Browserprint.info.xml

ihave100openPRs:


src/chrome/content/rules/Browserprint.info.xml
```diff
@@ -0,0 +1,43 @@
+<ruleset name="Browserprint.info">
```
J0WI: Wildcard DNS record.

ihave100openPRs:
@J0WI Added

PR #13227, opened 2017-10-19
Create ruleset for axios.com

thisisparker:


src/chrome/content/rules/Axios.com.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	Nonfunctional hosts in *.axios.com:
+
+	ads.axios.com t
+	links.axios.com m
```
J0WI: This should be `link`, not `links`.

J0WI:
@thisisparker there was an issue on Travis, please re-revert your change :)

thisisparker:
Instead of re-reverting I created a new commit, which corrects the subdomain but (correctly) leaves it in the nonfunctional category.

PR #13233, opened 2017-10-19
Fix: extension crash on Fennec due to missing icon APIs in chrome.bro…

### LINKED ISSUE #13035: 2017.10.4 doesnt work on the last Firefox for android
2017.10.4 doesnt work on the last Firefox for android

Hainish:
…wserAction (fixes #13035)

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -130,11 +130,13 @@ function updateState () {
     iconState = 'blocking';
   }
 
-  chrome.browserAction.setIcon({
-    path: {
-      38: 'icons/icon-' + iconState + '-38.png'
-    }
-  });
+  if ('setIcon' in chrome.browserAction) {
+    chrome.browserAction.setIcon({
+      path: {
+        38: 'icons/icon-' + iconState + '-38.png'
+      }
+    });
+  }
 
   chrome.browserAction.setTitle({
     title: 'HTTPS Everywhere' + ((iconState === 'active') ? '' : ' (' + iconState + ')')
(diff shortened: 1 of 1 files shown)


Hainish:
Ready for review @cowlicks

chromium/background.js
```diff
@@ -130,11 +130,13 @@ function updateState () {
     iconState = 'blocking';
   }
 
-  chrome.browserAction.setIcon({
-    path: {
-      38: 'icons/icon-' + iconState + '-38.png'
-    }
-  });
+  if ('setIcon' in chrome.browserAction) {
```
koops76: `if (chrome.browserAction.setIcon) {` would work just as well and is more readable.

chromium/background.js
```diff
@@ -147,11 +149,15 @@ function updateState () {
     const tabId = tabs[0].id;
     const activeCount = getActiveRulesetCount(tabId);
 
-    chrome.browserAction.setBadgeBackgroundColor({ color: '#666666', tabId });
+    if ('setBadgeBackgroundColor' in chrome.browserAction) {
```
koops76: `if (chrome.browserAction.setBadgeBackgroundColor) {`

chromium/background.js
```diff
@@ -147,11 +149,15 @@ function updateState () {
...
+      chrome.browserAction.setBadgeBackgroundColor({ color: '#666666', tabId });
+    }
 
     const showBadge = activeCount > 0 && isExtensionEnabled && showCounter;
 
-    chrome.browserAction.setBadgeText({ text: showBadge ? String(activeCount) : '', tabId });
+    if ('setBadgeText' in chrome.browserAction) {
```
koops76: `if (chrome.browserAction.setBadgeText) {`

chromium/background.js
Hainish: I actually think this is more readable.

koops76: It's a bit harder to read in my opinion. It's common in JS to use such property existence checks.

Hainish: Even if so, I like the explicit method you're searching for separated, as a syntactic nicety.

Issue #12677, opened 2017-09-22
Re-activating rulesets disabled by https-everywhere-checker with only one problematic domain


I-have-50-open-PRs-to-HTTPS-EVERYWHERE:

cschanaj:
a dirty way to find such ruleset could be

```bash
$ grep -l 'failed ruleset test' *.xml | while read filename; do if [ $(egrep -c '^Fetch error:' $filename) -eq 1 ]; then echo $filename; fi; done
```

with some false result for fetch-error only.

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
Thanks @cschanaj

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
Here are the results if anyone is interested:

0p.no.xml
1984.is.xml
1and1-Internet.xml
1d4.us.xml
1_Deg.org.xml
1fichier.xml
1nsk.ru.xml
1ShoppingCart.com.xml
247exchange.com.xml
24img.com.xml
2_Ton.com.au.xml
360_Yield.xml
3conline.com.xml
3D_Robotics.com.xml
42Floors.com.xml
4Tulemar.xml
5NINES.xml
7-Eleven.com.xml
8chan.co.xml
8tracks.xml
95516.com.xml
A14_Electronics.com.xml
Aalto.fi.xml
Aastatus.net.xml
AA.xml
Aberdeenshire.gov.uk.xml
About-Ads.xml
Absolute.xml
Abysmal.nl.xml
Academics.de.xml
Accessible-Information-Management.xml
Accuvant.com.xml
ACPICA.org.xml
active24.co.uk.xml
active24.cz.xml
active24.de.xml
active24.es.xml
active24.nl.xml
act-on.co.uk.xml
Adaptive_Computing.xml
Adbuyer.com.xml
Adlink.net.xml
AdMaster.com.cn.xml
Adobe_Connect.xml
Adobe_forms_central.com.xml
Adperium.com.xml
adreclaim.com.xml
adr.org.xml
Adscale.xml
Adtech.de.xml
Advertisers-OpenX.com.xml
Adzerk.net.xml
AEMI.xml
afisha.ru.xml
Agra-net.com.xml
Ahmia.fi.xml
aircrack-ng.org.xml
AirsoftGI.com.xml
akbars.ru.xml
akixi.com.xml
Aktivix.org.xml
Albennet.com.xml
Alien.net.au.xml
[...]

PR #13269, opened 2017-10-21
Re-activate Tapad.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
#12677

### DIFF
--- a/src/chrome/content/rules/Tapad.xml
+++ b/src/chrome/content/rules/Tapad.xml
@@ -1,10 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Tapad.xml
```diff
@@ -25,11 +19,10 @@ Fetch error: http://tapad.app10.hubspot.com/ => https://tapad.app10.hubspot.com/
 		- .tapad.com
 
 -->
-<ruleset name="Tapad (partial)" default_off='failed ruleset test'>
+<ruleset name="Tapad (partial)" >
```
J0WI: We can now secure some more hosts here.

I-have-50-open-PRs-to-HTTPS-EVERYWHERE: @J0WI Done

PR #13277, opened 2017-10-21
[Halcyon.sh] Add target

Bisaloo:


Bisaloo:
I got two different results on different computers.

I am okay with changing it for `Refused`.

jeremyn:
Merged, thanks.

PR #13285, opened 2017-10-22
Require lowercase targets as per #13036

### LINKED ISSUE #13036: Make tests case insensitive
Type: other

In #12955 it turns out, that our ruleset tests are checking the targets case sensitive. But they shouldn't.
So `FOO.BAR` and `foo.bar` should be treated as duplicated target host.

cschanaj:
ping @J0WI @cowlicks @Hainish this fix #13036


Hainish:
lgtm, thanks!

PR #13289, opened 2017-10-22
Create pushshift.io.xml

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:


jeremyn:
Mark as `Invalid certificate` in a top comment:
- [ ] https://tracker.pushshift.io

Mark as `No working URL known` in a top comment:
- [ ] https://api.pushshift.io (and remove the `target`)

Also:
- [ ] Remove the `securecookie` tag (since the only cookie set is for `^` and not all subdomains support HTTPS)

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@jeremyn Why is it necessary to give a working URL for it when it works just fine with HTTPS and doesn't drop a certificate error or something like that? I do know about a working URL with it:

> https://api.pushshift.io/reddit/search/comment?limit=50000&link_id=78fycp

Do you want me to mention it in a comment under the `target`?

jeremyn:
There's some historical discussion about `No working URL known` in issue https://github.com/EFForg/https-everywhere/issues/7662. The idea is to have at least one working URL for each target as a sort of minimal check. Note that subdomains can be bad `target`s even if there is no problem making the HTTPS connection, for example there could be too much mixed content blocking, or a redirect loop back to HTTP.

Where the root URL fails but you know of a working URL, the common pattern is to add it as a `test`. The `test` `url` should be something that will be valid indefinitely without any time-limited or personal-account-related items. Can `link_id` be safely removed? In that case you might make things look like this:

```xml
	<target host="api.pushshift.io" />
		<test url="http://api.pushshift.io/reddit/search/comment?limit=50000" />
	<target host="elasticsearch.pushshift.io" />
```

I-have-50-open-PRs-to-HTTPS-EVERYWHERE:
@jeremyn Thanks for the details, I added the custom test url as you proposed

jeremyn:
The PR is ready to merge but I'm not sure why the tests are failing. The tests are failing on the `preloaded` step but https://pushshift.io is not preloaded. The problem might be on Travis' end, so let's wait a little while and see if it fixes itself.

jeremyn:
Closing and reopening a PR is the best way I know of to rerun Travis, instead of pushing empty commits.

PR #13319, opened 2017-10-23
Run hsts-prune 

Bisaloo:


Hainish:
I see the same results.  Thank you @Bisaloo

PR #13324, opened 2017-10-24
Create btcgpu.org.xml

ivysrono:


jeremyn:
I get a `403` for both `^` and `www`. Are there any working URLs for these domains?

ivysrono:
The site seems under attack

jeremyn:
Looks okay now, thanks.

PR #13326, opened 2017-10-24
Update bootstrap.js

koops76:
@cowlicks What was the original code supposed to do? Why check the module?


koops76:
ping @Hainish.

PR #13330, opened 2017-10-24
[CAE.edu.au] Add ruleset.

bardiharborow:
I get a strange error (`need more than 1 value to unpack`) on two of the hosts in this PR, but they are both internal use things that we probably don't need to mess with anyway.


J0WI:
Travis had some issues, let's try again.

bardiharborow:
@J0WI Travis passed.

PR #13340, opened 2017-10-25
Update qdaily.com.xml

ivysrono:


J0WI:
Looks like the cert has just been renewed. Thanks anyway!

PR #13343, opened 2017-10-25
remove the wildcard entry for openrightsgroup.org

gwire:
The wildcard entry for openrightsgroup.org is effectively breaking the unsubscribe links in some of the mails sent via a third-party service.

(If it were safe, "preload" would have to the openrightsgroup.org Strict-Transport-Security header.)

### DIFF
--- a/src/chrome/content/rules/Open_Rights_Group.xml
+++ b/src/chrome/content/rules/Open_Rights_Group.xml
@@ -4,32 +4,23 @@
 		- 451_Unavailable.org.xml
 		- Blocked.org.uk.xml
 
-
-	Insecure cookies are set for these domains:
-
-		- .openrightsgroup.org
+	Widcard matching cannot be used since some subdomains point at third-party services
+	(without matching https certificates) such as news.openrightsgroup.org
 
 -->
 <ruleset name="Open Rights Group.org">
 
 	<target host="openrightsgroup.org" />
-	<target host="*.openrightsgroup.org" />
-
-		<test url="http://bug.openrightsgroup.org/" />
-		<test url="http://cardiff.openrightsgroup.org/" />
-		<test url="http://lists.openrightsgroup.org/" />
-		<test url="http://northeast.openrightsgroup.org/" />
-		<test url="http://scotland.openrightsgroup.org/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Open_Rights_Group.xml
```diff
@@ -4,32 +4,23 @@
 		- 451_Unavailable.org.xml
 		- Blocked.org.uk.xml
 
-
-	Insecure cookies are set for these domains:
-
-		- .openrightsgroup.org
+	Widcard matching cannot be used since some subdomains point at third-party services
+	(without matching https certificates) such as news.openrightsgroup.org
```
J0WI: Can you provide an example?

gwire: e.g. Mails sent via CleverReach have unsubscribe links that look something like

```
List-Unsubscribe: <http://news.openrightsgroup.org/rmftlp.php?cid=8888888&mid=88888888&h=888888-888888888888&liu=1>
```

ORG has received feedback from https-everywhere users that its unsubscribe links are broken, and I have traced the back to this file switching to a wildcard match.

J0WI: Could you document this issue in the ruleset?

gwire: There is currently no intention for the all subdomains to be https, but some subdomains are https.  That's what should be expressed, not enumerating current exceptions (including exceptions for internal services that only would affect a handful of users) since there may be further exceptions in the future.

(And we are contacting third-party suppliers to request the adoption of https for custom domains, but that's not the sole criteria we might have.)

I would like for the domain to be safely http-only, but at the point ORG chooses to make that commitment, it will be expressed in the HSTS header.

J0WI: Please document this in the ruleset. Otherwise your change may get reverted in future maintenance and we aren't aware of any issues.

gwire: We're still receiving reports that https-everywhere users are unable to use http://news.openrightsgroup.org unsubscribe links. Can someone let me know what it is I need to do to resolve this?

PR #13349, opened 2017-10-26
Update LeaseWeb.com.xml

wesdawg:


src/chrome/content/rules/LeaseWeb.xml
```diff
@@ -9,23 +9,27 @@
...
 	<!--	Direct rewrites:
 				-->
 	<target host="leaseweb.com" />
-	<target host="secure.leaseweb.com" />
 	<target host="www.leaseweb.com" />
+	<target host="auth.leaseweb.com" />
+	<target host="mirror.leaseweb.com" />
+	<target host="secure.leaseweb.com" />
+	
+	<target host="*.leaseweb.net" />
```
J0WI: This wildcard breaks various hosts.

wesdawg: How so ?

J0WI: Here are some:
```
ns7.leaseweb.net
ns8.leaseweb.net
www.leaseweb.net (same for ^ btw.)
www.mirror.de.leaseweb.net

J0WI: Can you add a comment for them on the top?

J0WI:
Thanks!

PR #13353, opened 2017-10-26
Update Xiami music file subdomains

lqf96:


src/chrome/content/rules/Xiami.xml
```diff
@@ -17,9 +17,11 @@
     <!-- Xiami.net rules -->
     <target host="img.xiami.net" />
         <test url="http://img.xiami.net/favicon.ico" />
+	<target host="m128.xiami.net" />
```
gloomy-ghost: Can you add a test here? http://m128.xiami.net/260/1260/6641/376016_14750_l.mp3 would be good.

lqf96: While I'd like to add a test here, the reality is that without a random generated token from the Xiami main site I will get a 403 Forbidden error, and this kind of approach is quite common in Chinese music sites.

PR #13368, opened 2017-10-27
Webextensions

Hainish:


Hainish:
These are the builds I've been using (with minor modifications) for the dev version.  I plan on merging these before a new WebExtensions release early next week.

Hainish:
Note: once this is merged remember to inform the Tor Project since they have `./makexpi` hard-coded in their build scripts.

koops76:
@Hainish I think it's a good idea to delete everything but `src/chrome/content/rules` from `src` in this PR since we are not using this code base anymore.

Hainish:
@koops76 we also need `src/META-INF` for our release process and `src/chrome/locale` for translations.  We may change the locations of these paths in the future, but it's not necessary for WebExtensions launch.

Hainish:
Creating a merge commit for granularity's sake, since so many files have changed.

koops76:
@Hainish You may delete the `webextensions` branch now.

PR #13372, opened 2017-10-28
Add Calculator.net.xml

andschwa:
I ran the tests, built the XPI, installed it (in Developer edition with legacy extensions enabled, can't figure out how to build the Firefox WebExtension I guess?), and tested. It's redirecting correctly. The site is fairly simple (but great), no assets fail to load.


andschwa:
Note: I highly recommend adding this website, since it has a mortgage calculator, which some people (_ahem_ me) use with what they consider rather personal information.

andschwa:
Also, I generated the rule with `make-trivial-rule calculator.net` :)

J0WI:
Thanks!

PR #13398, opened 2017-10-29
[hercrentals.com] New ruleset

youdly:


src/chrome/content/rules/Hercrentals.com.xml
```diff
@@ -0,0 +1,48 @@
+<!--
+	Non-functional subdomains:
+		- airwatch	(expired cert)
+		- autodiscover	(connection refused)
+		- css		(SSL handshake failed)
+		- g		(redirects to invalid www.g)
```
J0WI: WFM

youdly: it still redirects to `www.g` which refuse connection. connected via US location.

J0WI: It works for me in browser, but not with `curl`.
The DNS resolves to multiple endpoints.

youdly: I see. I added a test url. should I redirect `www.g` (which is not resolvable) to `g`?

youdly: Redirect doesn't help. Removed `g`.

PR #13406, opened 2017-10-29
Added ruleset for torservers.net

Jochen-A-Fuerbacher:


gloomy-ghost:
Thank you for your contribution but this domain has already been preloaded in all supported browsers and thus is unnecessary to add a ruleset.

PR #13408, opened 2017-10-29
Added ruleset for as250.net

Jochen-A-Fuerbacher:


zoracon:
Reran the build and this popped up
`ERROR src/chrome/content/rules/as250.net.xml: Fetch error: http://eris.as250.net/ => https://eris.as250.net/: (28, 'Connection timed out after 20001 milliseconds')`

Once amended I can merge this in

PR #13411, opened 2017-10-29
Added ruleset for bushcraft-germany.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/bushcraft-germany.com.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Bushcraft-Germany.com">
+
+	<target host="bushcraft-germany.com" />
```
J0WI: + `www.bushcraft-germany.com`

PR #13413, opened 2017-10-29
Add initial ruleset for randaris-anime.net

pointhi:
http://randaris-anime.net is a anime page which is quite complicated to get https onto it, because of Mixed Content everywhere.

The first version of the ruleset tries to get as much functionality as possible, while at least encrypting stuff like login and basic browsing. Chat as well as Comunity Feed are not visible on https sites, due to Mixed Content Problems.

Probably someone can also tell me if it's possible to get AJAX calls rewritten to https, to get some api's working.

### DIFF
--- a/src/chrome/content/rules/Randaris-anime.net.xml
+++ b/src/chrome/content/rules/Randaris-anime.net.xml
@@ -1,0 +1,29 @@
+<ruleset name="Randaris-anime.net">
+	<target host="randaris-anime.net" />
+	<target host="www.randaris-anime.net" />
+
+	<!-- This pages contain Mixed Content, and is rewritten to http by default -->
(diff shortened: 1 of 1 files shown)


jeremyn:
@pointhi Are you still interested in working on this?

pointhi:
It would be interest what's you though is about the current implementation, so I can fix outstanding issues

jeremyn:
It's tricky to handle domains that partly work and partly don't. You can't be sure that you handled all the problems.

A better approach here is instead of "blacklisting" stuff by adding `exclusion`s, you instead "whitelist" stuff by using focused `rule`s. For example we can probably safety rewrite URLs for static elements like images, CSS, and JS files. In particular this means you use the same `target`s you have now, but the `rule` or (`rule`s) will have more complicated `from` and `to` values. You can find many examples of these `rule`s in existing files.

pipboy96:
wip

PR #13421, opened 2017-10-30
Add Pietsmiet.de.xml, fix #13369

### LINKED ISSUE #13369: Pietsmiet.de no https redirect
The German Website "www.pietsmiet.de" should be redirected to https://www.pietsmiet.de 

Type: new ruleset
Domain: www.pietsmiet.de

J0WI:


jeremyn:
https://www.pietsmiet.de is an HTTP redirect back to http://www.pietsmiet.de so it's not a valid target. https://pietsmiet.de redirects to http://www.pietsmiet.de , so `^` is technically a valid target, but not really useful. I don't think this should be a ruleset at all, unfortunately.

gloomy-ghost:
> https://www.pietsmiet.de is an HTTP redirect back to http://www.pietsmiet.de

I am getting different result, also I don't think the ci will pass if it's a loop.

jeremyn:
I checked again and for whatever reason, https://www.pietsmiet.de works fine now.

@J0WI Do you want to add any of the domains @numismatika listed above in https://github.com/EFForg/https-everywhere/pull/13421#commitcomment-25338719 ? Otherwise this is ready to merge.

J0WI:
Most of them are non public hosts, so I only added `cdn.pietsmiet.de`.

PR #13425, opened 2017-10-31
Exclude test directory while generating package

gloomy-ghost:
The 2017.10.30 release includes a test directory which is not used within the extension

### DIFF
--- a/.build_exclusions
+++ b/.build_exclusions
@@ -18,3 +18,4 @@ webextension/.eslintrc.json
 .eslintignore
 node_modules
 package-lock.json
+test


jeremyn:
This looks okay to me.

@Hainish Can you please double-check this and merge it? It saves about 2-3KB off the final extension size.

Hainish:
Good catch @gloomy-ghost.  We already exclude files via `do_not_ship`:

https://github.com/EFForg/https-everywhere/blob/5db399307a014ff75d84d6e49be8895aee5d7646/make.sh#L56-L57

These are files that can't be in `--exclude` due to the wildcard, even though `test` can be.  For the sake of consistency, though, we should add `test` to this env var and change the `rm -f` to a `rm -rf` on the line under it.

Hainish:
Instead of `rsync` then `rm`, this `rsync` option should prove useful:

```
       --exclude-from=FILE
              This option is related to the --exclude option, but it specifies a FILE that contains exclude patterns (one per line).  Blank lines in the file and lines starting with ’;’ or ’#’ are ignored.
              If FILE is -, the list will be read from standard input.
```

jeremyn:
I don't know about `--exclude-from=FILE`. I don't want to have an extra config file hanging around, and I guess the standard input option would require some `xargs` piping.

Hainish:
Seems like there may be some overlap here: we already have the https://github.com/EFForg/https-everywhere/blob/master/.build_exclusions file.  Perhaps we should just move the `do_not_ship` files into this, instead of having two separate places which serve the same purpose.

cschanaj:
What is the status of this PR? it has been a while since this is on the "Before the Next Release" milestone. 

Related #16122

Hainish:
Thanks @gloomy-ghost!

PR #13428, opened 2017-10-31
Create tabac-info-service.fr.xml

Bisaloo:


J0WI:
I get some mixed content warnings on these sites.

Bisaloo:
Only active mixed content I could find was on `www` and it was from an adserver. Do you see something else?

J0WI:
I'm fine if only ads are affected.

PR #13429, opened 2017-10-31
Fix "Please specify whether you want browser_style or not in your browser_action options." warning

koops76:


Bisaloo:
This has a merge conflict as well and the chances of it getting fixed are lower than #12776 given the submitter account has been deleted. Closing.

PR #13430, opened 2017-10-31
Added ruleset for bremen.freifunk.net

Jochen-A-Fuerbacher:


J0WI:
We already have a rule `Freifunk.xml`, so you should add a reference for it.

PR #13433, opened 2017-10-31
[Archives-ouvertes.fr] Simplify and update ruleset

Bisaloo:


src/chrome/content/rules/Archives-ouvertes.fr.xml
```diff
@@ -1,48 +1,34 @@
...
 
 	<target host="archives-ouvertes.fr" />
 	<target host="*.archives-ouvertes.fr" />
-
-
-	<!--	Not secured by server:
-					-->
-	<!--securecookie host="^tel\.archives-ouvertes\.fr$" name="^PHPSESSID$" /-->
+		<test url="http://www.archives-ouvertes.fr/" />
+		<test url="http://archives-ouvertes.fr/" />
```
J0WI: test url already covered by target

src/chrome/content/rules/Archives-ouvertes.fr.xml
```diff
@@ -1,48 +1,34 @@
...
+		to="https://archives-ouvertes.fr/" />
 
-	<!--	Redirect keeps path and args:
-						-->
-	<rule from="^http://(?:www\.)?archives-ouvertes\.fr/+"
-		to="https://hal.archives-ouvertes.fr/" />
-
-	<rule from="^http://(api|aurehal|aurehal-preprod|hal|hal-\w+|halshs|pastel|tel)\.archives-ouvertes\.fr/"
+	<rule from="^http://(api|aurehal|aurehal-preprod|cel|cv|doc|hal|hal-\w+|halshs|medihal|pastel|tel|telearn|www)\.archives-ouvertes\.fr/"
 		to="https://$1.archives-ouvertes.fr/" />
```
J0WI: Why aren't you using a generic http to https rule here?

Bisaloo: I'm not sure. I simply kept the old rule. I guess it makes sense if some targets don't support `https`.

src/chrome/content/rules/Archives-ouvertes.fr.xml
```diff
@@ -1,48 +1,34 @@
...
-		- aurehal
-		- aurehal-preprod
-		- hal
-		- hal-\w+ *
-		- halshs
-		- medihal
-		- pastel
-		- tel
-
-	* Per-institution domains
```
J0WI: We should keep a comment about the wildcard.

PR #13434, opened 2017-10-31
Added ruleset for ffnw.de

Jochen-A-Fuerbacher:


J0WI:
Their cert expired today, so let's wait a bit.

PR #13443, opened 2017-10-31
Added ruleset for nuand.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/nuand.com.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="Nuand.com">
+
+	<target host="nuand.com" />
+	<target host="*.nuand.com" />
```
J0WI: Is there a specific use case for a wildcard here?

PR #13448, opened 2017-11-01
Add openbudget and news subdomains to City_of_Seattle.gov

andschwa:
Two more sub domains for `seattle.gov`.

### DIFF
--- a/src/chrome/content/rules/City_of_Seattle.xml
+++ b/src/chrome/content/rules/City_of_Seattle.xml
@@ -2,54 +2,22 @@
 Disabled by https-everywhere-checker because:
 Non-2xx HTTP code: http://seattle.gov/ (200) => https://www.seattle.gov/ (303)
 	For other U.S. government coverage, see US-government.xml.
-
-
-	Nonfunctional subdomains:
-
-		- host1		(times out)
-
-
-	Problematic subdomains:
-
-		- ^ *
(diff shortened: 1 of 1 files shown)


cschanaj:
You can avoid adding the test URLs by listing the `target` explicitly instead of using a wildcard. Besides, please also have a look on the Travis error.

jeremyn:
@andschwa Are you still interested in working on this pull request?

andschwa:
Taking a look, the original ruleset looks pretty old. That Travis CI failure is because the `wald1` subdomain is down. Maybe I'll just rewrite some of this...

andschwa:
I'm just going to close this, there are so many weird things about their website :( Like, `https://seattle.gov` just 404s.

jeremyn:
@andschwa Thanks for the update. If you like, you're welcome to rewrite the existing ruleset to cover at least the most important subdomains. It doesn't need to provide exhaustive coverage, especially since the ruleset is `(partial)` anyway. In particular it would be nice to cover https://www.seattle.gov . But it is up to you.

jeremyn:
Actually on closer inspection it looks like `www` is already covered by the wildcard target. We generally don't like wildcards so it would be nice for this to get rewritten, but in any case my specific complaint about `www` is not an issue, just to clarify.

PR #13454, opened 2017-11-01
add rule for ORF.at

numismatika:
Work in Progress,
some "subfolders" of the websites that relink to plain can be secured.
Needs a few more exceptions therefore

### DIFF
--- a/src/chrome/content/rules/ORF.at.xml
+++ b/src/chrome/content/rules/ORF.at.xml
@@ -1,0 +1,253 @@
+<!--
+
+	Non-2xx HTTP code:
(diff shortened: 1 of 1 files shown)


J0WI:
Is this still work in progress?

numismatika:
Seems like a very small subset of fonts and files can be secured on the main sites, certificate situation didn't change yet. I would consider this best effort for the time being and finished for now

J0WI:
Please make sure that all tests on Travis pass.

numismatika:
Looks like the content wasn't static after all. Removed the rewrites

PR #13473, opened 2017-11-03
Create gorod.dp.ua.xml

Folant:


jeremyn:
I'll look at this but please first go through and clean up some of the basic stuff we've discussed (indentation, names).

jeremyn:
Almost all of the domains seem to be mirrors of each other, except https://http://m.gorod.dp.ua , and all of the main URLs (except `m`) have mixed content blocking. Most of it is ads but there is one thing that looks like it might be a currency converter ("Наличная валюта") (?) that is blocked. However I don't see anything other non-ad stuff broken on other pages in these sites, so overall I think it's okay not to make this `mixedcontent` or to add some kind of exception for the main pages.

So, I've merged it.

PR #13493, opened 2017-11-05
Add status.aws.amazon.com to AmazonWebServices.xml

jeremyn:


src/chrome/content/rules/AmazonWebServices.xml
```diff
@@ -291,6 +292,7 @@
 	<test url="http://fls-eu.amazon.co.uk/" />
 	<test url="http://docs.aws.amazon.com/" />
 	<test url="http://forums.aws.amazon.com/" />
+	<test url="http://status.aws.amazon.com/" />
```
gloomy-ghost: Isn't it implied? Actually I think we can remove all [these](https://github.com/jeremyn/https-everywhere/blob/c32ec2d06fd7fa3ee7af9eb00c20df6a87123252/src/chrome/content/rules/AmazonWebServices.xml#L292-L300) without breaking things

jeremyn:
I've removed the tests.

gloomy-ghost:
merged, thanks!

PR #13500, opened 2017-11-06
Re-enable Bakerlab.org.xml

gloomy-ghost:


src/chrome/content/rules/Bakerlab.org.xml
```diff
@@ -16,11 +9,10 @@ Fetch error: http://ws2.bakerlab.org/ => https://ws2.bakerlab.org/: (28, 'Connec
 		- robetta.bakerlab.org
 		- rosettaserver.bakerlab.org
 -->
-<ruleset name="bakerlab.org" default_off='failed ruleset test'>
+<ruleset name="bakerlab.org">
 	<target host="www.bakerlab.org" />
```
J0WI: Could you also add `^bakerlab.org`?

PR #13501, opened 2017-11-06
Fix #4174

### LINKED ISSUE #4174: gfblip.appspot.com does not work with https
This site uses HTTP to send single-packet payloads to two different web servers in order to measure latency. This does not work with HTTPS but multiple rules for Google hit since this is on appspot.com, which should probably not be turned off.

### LINKED ISSUE #4656: gfblip.appspot.com kills browser (freeze)
Visiting gfblip.appspot.com with "HTTPS Everywhere" enabled causes my browser to freeze, at least for long enough to trigger ~"this app is unresponsive, kill it?"

Odd that this wasn't seen in report #4174

Firefox 45.0.2, running under Fedora Linux.
Other enabled content extensions: NoScript, Disconnect (tracker blocker).

gloomy-ghost:
Should close #4656 as well

### DIFF
--- a/src/chrome/content/rules/Google_App_Engine.xml
+++ b/src/chrome/content/rules/Google_App_Engine.xml
@@ -23,6 +23,9 @@
 		<test url="http://sbux-portal.appspot.com/" />
 		<exclusion pattern="^http://www\.appspot\.com/" />
 		<test url="http://www.appspot.com/" />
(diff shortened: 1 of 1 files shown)


gloomy-ghost:
Oops, ci requires some more test urls to pass.

PR #13507, opened 2017-11-06
Add Eskimi.com.xml

J0WI:
Site contains some trivial mixed content issues.


src/chrome/content/rules/Eskimi.com.xml
```diff
@@ -0,0 +1,33 @@
+<!--
+	Nonfunctional hosts in *.eskimi.com:
+		- business.eskimi.com (r)
+		- games.eskimi.com (r)
```
gloomy-ghost: mismatch for me

J0WI: Thanks, changed

Issue #2323, opened 2015-07-26
"Block all HTTP requests" blocks requests even to localhost


adammw:
Localhost refers to the 127.0.0.1 loopback interface, therefore no data is ever leaving the network. Including it in the "block all HTTP requests" makes it needlessly harder to interact with local webservers.

semenko:
Going to close as wontfix for now.

Feel free to send a PR to fix, but this adds a fair amount of complexity if we try to deal with localhost 127.0.0.0/8 resolution, etc.

mweitzel:
What about an HTTP-OK whitelist? That would resolve this without it being a special case fix.

semenko:
Ok, rethinking this, I like it. I'll try to integrate it, at least in Chrome, with the overhaul of the extension.

Chrome trusts these origins, which we should whitelist (if we see them in webRequest):
```
(https, *, *)
(wss, *, *)
(*, localhost, *)
(*, 127/8, *)
(*, ::1/128, *)
(file, *, —)
(chrome-extension, *, —)
```

per https://www.chromium.org/Home/chromium-security/security-faq#TOC-Which-origins-are-secure-

Bisaloo:
@Hainish, what's the status of this after https://github.com/EFForg/https-everywhere/commit/1dafcc9578ab3c9287dc4f811878269fdd57f9ae?

jeremyn:
I've closed issues #9142 and #13511 as duplicates of this issue, though those issues are about allowing unencrypted requests to [RFC1918](https://tools.ietf.org/html/rfc1918) private addresses and not localhost 127.0.0.1/8. If a fix for this issue is limited to just localhost, then one of those issues should be reopened or a new issue should be created to cover other private address ranges.

Hainish:
This is no longer the case, closing.  Thanks!

PR #13515, opened 2017-11-07
Add IPv6 to block unencrypted connections whitelist

koops76:
https://github.com/EFForg/https-everywhere/issues/2323#issuecomment-167377814

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -226,14 +226,14 @@ function onBeforeRequest(details) {
   const uri = new URL(details.url);
 
   // Should the request be canceled?
-  var shouldCancel = (
-    httpNowhereOn &&
-    uri.protocol === 'http:' &&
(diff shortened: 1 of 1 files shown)


J0WI:
Sadly the repo is gone and needs to be rebased, but lthe missing IPv6 link local address is still an issue.
Let's ask @Hainish if he has any use for this patch.

Hainish:
The removal of regex is preferred, so I'm inclined to keep these changes.

I find the regex change for hostnames starting with `127.` overly hard to read for no added benefit (other than it being stricter regex, but I don't really see an instance where that would be helpful.)  I'll keep the existing regex in place.

The IPv6 conditional looks good.

Hainish:
Fixed in https://github.com/EFForg/https-everywhere/commit/5db399307a014ff75d84d6e49be8895aee5d7646

PR #13523, opened 2017-11-08
[Onion.to] Add target and test urls

Bisaloo:


J0WI:
The service seems to be down/broken.

J0WI:
WFM now.

PR #11804, opened 2017-08-10
Update Bitly Branded Short Domain list

SeanOC:
This is a periodic update to the list of Branded Short Domains powered by Bitly with current HTTPS support. (The list was last updated in #5611)

It's expected that there is some natural churn on this list due to domains being added or removed from the Bitly platform, or SSL certificates becoming available or unavailable for a given domain.

Methodology:

This is the list of domains on the Bitly platform that were verified with valid DNS settings that point to Bitly in the last 12 hours, and for which Bitly has a valid SSL certificate with >20 days remaining. (20 days is important as our current threshold for renewing LetsEncrypt certificates is 21 days.)

The ruleset directory was checked for potential duplicates via the https-everywhere test suite and 3 were removed.

cc: @jsha @Hainish and #11439

PR #13541, opened 2017-11-10
Update Bitly Branded Short Domain list.

SeanOC:
This is a periodic update to the list of Branded Short Domains powered by Bitly with current HTTPS support. (The list was last updated in #11804)

It's expected that there is some natural churn on this list due to domains being added or removed from the Bitly platform, or SSL certificates becoming available or unavailable for a given domain.

Methodology:

This is the list of domains on the Bitly platform that were verified with valid DNS settings that point to Bitly in the last 12 hours, and for which Bitly has a valid SSL certificate with >20 days remaining. (20 days is important as our current threshold for renewing LetsEncrypt certificates is 21 days.)

In this update there were: 
Domains Added 2667
Domains Removed 2451
Domains Kept 32295

The test suite runs successfully locally and this update has been scrubbed of domains on the HSTS preload list.

cc: @jsha @Hainish


Hainish:
Thanks again for this!

SeanOC:
Travis CI Tests found 3 duplicate domains. Will shortly push up an update that removes the dups.

For reference the duplicate domains are:

go.dmoruzzi.com
s.nikkei.com
st.aastocks.com

koops76:
@SeanOC Better use HSTS preloading.

SeanOC:
Looks like the only failure is the "fetch" test timed out which I think is somewhat expected.

@koops76 could you please expand your comment? The file has had the HSTS preload filter script run against it and it passed that test in Travis.

Hainish:
@SeanOC, that's right, fetch is expected to fail.  Otherwise looks good.  Merging.

koops76:
@SeanOC Bit.ly should use HSTS preloading instead of creating the ruleset since Bit.ly controls the domains and may set HSTS headers for them.

Hainish:
@koops76 I believe Bit.ly has already been in talks with the HSTS Preload list maintainers.

J0WI:
@SeanOC would be great to get another update of the list 😍

Hainish:
cc @zoracon

SeanOC:
Unfortunately I don't work at Bitly anymore, but @jehiah might be able to help you out.

zoracon:
Thank you @SeanOC , @jehiah It would be a great favor if you could update this if you can

PR #13542, opened 2017-11-10
Update Arduino.cc.xml

Rotzbua:
more domains offer https

### DIFF
--- a/src/chrome/content/rules/Arduino.cc.xml
+++ b/src/chrome/content/rules/Arduino.cc.xml
@@ -1,26 +1,20 @@
 <!--
 	Nonfunctional subdomains:
-
-		- (www.)? ¹
-		- blog ¹
 		- www.id ²
-		- playground ¹
-		- store ¹
 
 	¹ Refused
 	² Differs from http
-
-
-	Fully covered subdomains:
-
-		- id
-
+    
+    Latest test: 10.11.2017
 -->
 <ruleset name="Arduino.cc (partial)">
-
+	<target host="arduino.cc" />
+	<target host="www.arduino.cc" />
+	<target host="blog.arduino.cc" />
+	<target host="cdn.arduino.cc" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Arduino.cc.xml
```diff
@@ -1,26 +1,20 @@
...
 <ruleset name="Arduino.cc (partial)">
-
+	<target host="arduino.cc" />
+	<target host="www.arduino.cc" />
+	<target host="blog.arduino.cc" />
+	<target host="cdn.arduino.cc" />
 	<target host="id.arduino.cc" />
-
+	<target host="playground.arduino.cc" />
+	<target host="store.arduino.cc" />
```
J0WI: See also:
```
auth.arduino.cc
casajasmina.arduino.cc (refused)
create.arduino.cc
cloud.arduino.cc
day.arduino.cc
downloads.arduino.cc
forum.arduino.cc

Rotzbua: thanks, sub domains added

J0WI:
`ERROR src/chrome/content/rules/Arduino.cc.xml: Fetch error: http://auth.arduino.cc/ => https://auth.arduino.cc/: need more than 1 value to unpack`

@Hainish we have some more PRs with this error. Do you have any clue what this means?

Rotzbua:
@J0WI @Hainish The reason seems to be a broken http response. There is a redirect but the `location` is wrong.
```HTTP Headers for http://auth.arduino.cc

Date: Sat, 11 Nov 2017 18:13:40 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Set-Cookie: AWSALExpires=Sat, 18 Nov 2017 18:13:40 GMT; Path=/
Server: nginx/1.13.6
Location: https:///

PR #13552, opened 2017-11-11
Create LaTribune.fr.xml

Bisaloo:


J0WI:
There are much more hosts, e.g. `evenement.latribune.fr`. Would be great if you could add them too.

PR #13555, opened 2017-11-11
Add Smallpdf.com.xml

J0WI:


jeremyn:
- [ ] https://extract.smallpdf.com has `Different HTTP/HTTPS content`: http://extract.smallpdf.com redirects to https://smallpdf.com/pdf-to-jpg but https://extract.smallpdf.com doesn't redirect and shows what looks like a mirror of `^`. So `extract` should be removed as a `target` and added to a top comment.

J0WI:
Thanks, done.

PR #13557, opened 2017-11-11
[Sci-hub] Add target

Bisaloo:


src/chrome/content/rules/Sci-hub.xml
```diff
@@ -1,13 +1,9 @@
 <!--
 	Invalid certificate:
 		www.sci-hub.bz
-		sci-hub.io (broken chain)
-		www.sci-hub.io (broken chain)
-
-Mirrors of the documents live on their own subdomain where https is refused. Examples:
```
J0WI: This comment is still true

Bisaloo: The refused part is still true. But those subdomains are not used anymore from what I understand. See my full commit message for more info.

PR #13558, opened 2017-11-11
[Spotify.com] Simplify ruleset

Bisaloo:


src/chrome/content/rules/Spotify.xml
```diff
@@ -49,31 +41,23 @@
 -->
 <ruleset name="Spotify.com (partial)">
```
J0WI: We are no longer `partial`.

Issue #13512, opened 2017-11-07
[Discussion] New build system


ghost:
Type: other

Important features:

- [ ] Builds are reproducible.
- [ ] Runs at least on latest version of Windows 10, Ubuntu and macOS.
- [ ] Runs faster than the current build system.

Desired features:

- [ ] Based on Node.
- [ ] Runs everywhere Node does.

koops76:
ping @Hainish @cowlicks @jeremyn @gloomy-ghost.

jeremyn:
@koops76 I'm wondering what the motivation is here. Are you looking for a way to personally contribute something big to the project and want to focus on this? Or have there been recent build problems you want to fix?

The build process seems mostly okay to me. It's not the highest priority problem we have. It could definitely be improved though.

koops76:
@jeremyn The current process is tied to Linux, not very fast and the builds are not reproducible.

koops76:
@jeremyn Recently we have switched to WebExtensions too.

Hainish:
@koops76 the builds are reproducible from one run to another locally - this is one of the reasons for the inclusion of `utils/zipfile_deterministic.py` in the repo.  If they are not reproducible from one build environment to another, this should be investigated as some source of entropy is leaking into the project somewhere (perhaps a `.gitignore`'d file is sneaking into the build)

koops76:
@Hainish I would like to replace the current build system with a Gulp or Webpack based one, this is what this issue is about. Either of these supports [building CRX files](https://github.com/oncletom/crx) and [reproducible builds](https://github.com/bitgenics/deterministic-zip).

koops76:
Also I would like proper support for NPM dependencies.

Hainish:
This set of requirements is particular to this contributor.  They have removed their account.  Closing.

jeremyn:
@Hainish I agree with closing this issue since the contributor has closed their account. However, I think we agree with some of their suggestions as wishlist items, like converting the build system to Node, and it might be worth adding them to whatever wishlist roadmap exists, or making such a roadmap. But there's no need to discuss that further here.

PR #13560, opened 2017-11-11
Use CodeMirror in ruleset debugger (#11794)

jeremyn:
@Hainish @cowlicks This should resolve issue #11794 to add CodeMirror to the ruleset debugger.

The commit message describes how I uglified (minified) the CodeMirror files. For security purposes I encourage someone at the EFF to verify these minifications.

Also, as the commit message says, this almost certainly requires another update to satisfy CodeMirror's MIT license. In https://github.com/EFForg/https-everywhere/pull/13062#issuecomment-341272152 I discuss some concerns I have about HTTPS Everywhere's approach to licensing, so for the time being I will let someone at the EFF figure out how to handle licensing here. Please feel free to push any needed changes directly to my branch.


koops76:
Instead of adding CodeMirror as a binary blob in your PR, use NPM dependency instead.

jeremyn:
It doesn't look like we use npm in the build process, for example we don't require it in [`install-dev-dependencies.sh`](https://github.com/EFForg/https-everywhere/blob/28f49a128070da231c919131a28c9294e9b60bbf/install-dev-dependencies.sh). Could you be more specific?

koops76:
@jeremyn Including a binary blob means your PR is hard to review since the reviewer would have to check if it is identical to an official distibution. Instead we need to add proper dependency management and possibly Webpack. I will make a PR to your branch with a temporary solution.

koops76:
@jeremyn You used `uglify-es`, not `uglifyjs`, right?

koops76:
I cannot reproduce the same JS file.

koops76:
I can't reproduce the CSS file either, my file is 1 byte shorter.

koops76:
I can reproduce a CSS file but I had to delete an empty line at the end.

jeremyn:
I used https://www.npmjs.com/package/uglifyjs as mentioned in the git commt message.

koops76:
@jeremyn https://www.npmjs.com/package/uglify-js

chromium/debugging-rulesets.js
```diff
@@ -2,36 +2,44 @@
```
koops76: Remove these blank lines please.

jeremyn:
I'm not sure what the relationship is between https://www.npmjs.com/package/uglifyjs and https://www.npmjs.com/package/uglify-js -- I see they share a collaborator -- but https://www.npmjs.com/package/uglify-js has many more downloads so I agree that's the one to use.

I've (force-)pushed an update that uses `uglify-js@3.1.9` for the two JS files. I also removed an extra blank line from the middle of `debugging-rulesets.js`.

jeremyn:
@koops76 I'm not clear from your comments whether you could reproduce the CSS minification. Try the following:

```
$ md5sum codemirror.css  # should be 4b107dce67663f59927a270fa63f3832
$ npm install uglifycss@0.0.27
$ node_modules/uglifycss/uglifycss codemirror.css > codemirror.min.css
$ md5sum codemirror.min.css  # should be 4169bd1daef71f34419197aeaa4b32b1
```

Do you get the same hashes?

koops76:
@jeremyn Use `sha256sum`.

koops76:
I reproduced everything.

Hainish:
I do think if we're adding dependencies (which I'm not necessarily against, as the `sign-rulesets` branch also adds dependencies and so there are multiple reasons why we're moving in that direction) we should include some kind of JS package management.

We already have `npm` as a requirement for many of the utilities we use.  Although `npm` isn't exclusively tasked with front-end dependency graphs, it is nonetheless useful for this purpose.  In combination with `grunt` or some other task-runner, we could include steps which minify the resources and place them properly in the extension path.

@koops76 I see you deleted comments that suggested similar.  I'm curious why?

jeremyn:
@Hainish I agree that listing a dependency in a `package.json` or equivalent has some advantages compared with checking the dependency into the repository. On the other hand, adding a new download source introduces another point of failure during builds, including test builds.

However assuming we prefer the `package.json` approach, is that available at all in the current build process? That was my question to @koops76. I don't think it is.

By the way, @koops76 recently opened an issue (https://github.com/EFForg/https-everywhere/issues/13512) to discuss remaking the build system.

Hainish:
@jeremyn my typical workflow would be to create a `package.json` with any dependencies, using that to do version tracking, and of course putting `node_modules` into `.gitignore`.  I'd then create a `grunt` task to minify and copy the minified JavaScript or CSS to the extension paths, but actually _check those files in_.  This way, we could periodically update those dependencies but wouldn't have to deal with update failures at build time.  We also wouldn't need to add `npm` as a hard build dependency in that case, just as a dev dependency for those that are updating resources.

jeremyn:
I think it's too early to create a project-wide installation architecture, just to make CodeMirror available to the ruleset debugger. It raises questions like, where should the package file go? How should we document the installation process? Also for security we should include checksums somewhere, and I don't think `npm shrinkwrap` supports that, so we'd need to use something like https://github.com/mozilla/npm-lockdown or maybe [Yarn](https://yarnpkg.com/). We should wait for one or two more dependency/feature use cases before trying to figure these questions out.

Maybe you're concerned that the only documentation about how to create the minified CodeMirror files is in the git commit message, which isn't great. I can add a small `README.md` to `chromium/codemirror/` if you want.

koops76:
@jeremyn NPM is better than Yarn security-wise. It creates a `package-lock.json` file with SHA512 sums instead of SHA1 sums Yarn currently uses.

jeremyn:
@koops76 According to https://docs.npmjs.com/files/package-lock.json#integrity : "For registry sources, this is the `integrity` that the registry provided, or if one wasn't provided the SHA1 in `shasum`." `uglifycss` has a SHA1 hash, for example you see this in `package-lock.json` after installing the package:

```json
  "dependencies": {
    "uglifycss": {
      "version": "0.0.27",
      "resolved": "https://registry.npmjs.org/uglifycss/-/uglifycss-0.0.27.tgz",
      "integrity": "sha1-U1U7gRXeJtzle0K6CzTojdgNDN4="
    }
  }
```

If you change the `sha1` value, for example from `sha1-U1U7...` to `sha1-11U7...` it will fail as expected. However if you change the `integrity` field to try to use a SHA512 hash, for example to `sha512-1111...` it will unexpectedly work. (The correct SHA512 hash for https://registry.npmjs.org/uglifycss/-/uglifycss-0.0.27.tgz is `0hA7g3cXvHdhetBkc2bGbA+Mu/C23jncs6kkmZLsTJCRloT72InV5XZs08t+m8AcFqBAsms+ZRK7cuY+eAXI6w==` which you can find with `openssl sha512 -binary uglifycss-0.0.27.tgz | base64`.)

This is not all to say "you're wrong", but instead just that npm integrity checking looks complicated. I'm also concerned that it silently continues to work even with a bad hash. I don't know how Yarn would compare in similar tests.

jeremyn:
And also, it would be nice if we could just merge this PR without having to first dig deep into comparing the integrity mechanisms of npm vs Yarn vs anything else.

koops76:
@jeremyn This is because uglifycss is too old and doesn't have a SHA512 hash yet.

Hainish:
@jeremyn I think it's sensible to merge in these dependencies without creating a dependency management system in general.  I haven't had a chance to review this yet (I'm in meetings all day) but after just a cursory look, can I suggest we put any external deps in a separate path, say `chromium/external/` and have `codemirror` included under that?  Just so we're clear that this is an external library, and that way we could also perhaps add a licensing note under that path explaining how these libraries are covered under their respective licenses (or add that note, mentioning the path, under the top-level `LICENSE.txt` file).

Hainish:
> I think it's sensible to merge in these dependencies without creating a dependency management system in general.

I mean this is sensible _for now_, given the relatively confined scope of the usage of this library.  I *do* think it will be necessary to have a dependency management system in the future, as we build out the extension.  I also think if we're not to go full-blown dependency system right now, having this under a `chromium/external/` path (with the steps to recreate included) will ensure that when we do put a dependency management system in place, this won't fall through the cracks.

koops76:
@Hainish I think the directory should be called `vendor`.

jeremyn:
I moved the minified files from `codemirror/` to `external/` and added an `external/README.md`. I also rebuilt the three minified files, because my original approach used redirection instead of the `--output` option to create the output files and apparently redirection introduces a newline at the end. I think the `--output` option makes a better build process.

I used a force push instead of a regular push. I'm sorry for the trouble. This should keep unnecessary versions of the minified files out of the repository.

I prefer `external` to `vendor`. `external` is less obscure, and it avoids odd naming situations in the future such as if HTTPS Everywhere uses a dependency from another EFF project.

@Hainish For licensing, as I describe in issue https://github.com/EFForg/https-everywhere/pull/13062#issuecomment-341272152 (still waiting for a response there), I have concerns about the overall licensing of this project. That being the case, you/the EFF can handle the CodeMirror licensing. In fact I don't want your licensing changes to be done under my name, so I've unchecked the "Allow edits from maintainers" option to make sure you don't push licensing changes and then accidentally squash-and-merge them. (I know you would not do that intentionally if I asked you not to, but you might forget if this issue drags on.) So that will need to be done in a separate PR.

Hainish:
@jeremyn thanks for those changes.  As I mentioned I'm in meetings all day so I'll have to wait until at least tomorrow to review, but I definitely plan to have this merged by next release (which I'm thinking should be early next week.)

chromium/external/README.md
```diff
@@ -0,0 +1,42 @@
+# External dependencies
+
+This directory contains files that are external dependencies for HTTPS Everywhere, with instructions on how to get or remake them.
+
+## CodeMirror 5.31.0
+
+```
+$ npm install uglify-js@3.1.9
```
Hainish: Since this is not a package dependency and is intended to be run as a command line app, let's add a `-g` flag for global install as the installation instructions specify: https://github.com/mishoo/UglifyJS2#install

chromium/external/README.md
```diff
@@ -0,0 +1,42 @@
...
+$ npm install uglifycss@0.0.27
```
Hainish: Same here: https://github.com/fmarcia/UglifyCSS#installation

chromium/external/README.md
```diff
@@ -0,0 +1,42 @@
...
+  --comments '/copyright|license/' \
+  --output codemirror-5.31.0.xml.min.js \
+  codemirror-5.31.0/mode/xml/xml.js
+```
+
+### codemirror-5.31.0.min.css
+
+```bash
+uglifycss \
+  --output codemirror-5.31.0.min.css
```
Hainish: Missing a trailing `\` here

chromium/.eslintignore
```diff
@@ -0,0 +1 @@
+external/*
```
Hainish: We'll want to add this file to the top-level `.build_exclusions` as well.

chromium/debugging-rulesets.html
```diff
@@ -4,20 +4,21 @@
     <meta charset="utf-8">
     <title></title>
     <link href="debugging-rulesets.css" rel="stylesheet">
+    <script src="external/codemirror-5.31.0.min.js"></script>
+    <link href="external/codemirror-5.31.0.min.css" rel="stylesheet">
+    <script src="external/codemirror-5.31.0.xml.min.js"></script>
```
Hainish: I may have not been specific enough in my suggestion.  Can we move each of these under `external/codemirror/`, just so each dependency moving forward is in its own neat path?

chromium/external/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+### codemirror-5.31.0.min.js
+
+```bash
+uglifyjs \
+  --compress \
+  --mangle \
+  --comments '/copyright|license/' \
+  --output codemirror-5.31.0.min.js \
```
Hainish: Relating to the suggestion above of adding each external dependency to its own unique path, we'll want to change these output paths to `external/codemirror/`...

chromium/external/README.md
mgol: What's the drawback of defining all the dependencies in `package.json` in desired versions and define npm scripts that would execute proper commands?

Global packages should be avoided when possible. Someone may skip those steps as they've already performed them before and not notice they've since installed different versions since some other project required different ones. Using local packages gets rid of that problem.

Hainish: Good point, but the above command will cause `uglify-js` to be added to the `dependencies` hash, rather than the `devDependencies` hash.  Thus we should preference running this with the `-D` flag, and the below reproducibility instructions should be modified to be:
`../node_modules/uglify-js/bin/uglifyjs`

mgol: Is there any reason to specify those packages directly instead of just having `package.json` checked into the repo and running `npm install`?

BTW, if you're referring to the binary by path, it's usually better to use `../node_modules/.bin/uglifyjs` as that depends less on the internal package structure. `node_modules/.bin` contains links to binaries exported by top-level packages.

Hainish: @mgol creating a dependency management system is on the table, but I don't want to block merging until we have one in place.  See https://github.com/EFForg/https-everywhere/pull/13560#issuecomment-344083403 and resulting discussion.

Hainish:
Thinking it over, I don't really see the use for minification here.  In web development, it's useful to minify in order to reduce the size of the downloaded resources.  Here, we're already compressing the extension at build time, since `xpi` and `crx` files are effectively `zip`s.  And while there may be a memory overhead win, I can't imagine it would be very much.

If there isn't a definitive benefit to minification here, we could remove the necessity for `uglify-js` and `uglifycss`.

koops76:
@Hainish It's a common courtesy to minify any distributed files to save some bits, I think the extension's code should also be minified.

Hainish:
I do see some savings in download size between the minified

`-rw-r--r-- 1 user user 1972105 Nov 15 14:51 pkg/https-everywhere-2017.10.30~pre-amo.xpi`

and non-minified:

`-rw-r--r-- 1 user user 2019772 Nov 15 14:54 pkg/https-everywhere-2017.10.30~pre-amo.xpi`

so it appears that minification does save us some space.

koops76:
@jeremyn We will add a minification step later, you may completely remove minification for now.

Hainish:
@koops76 let's leave minification in if it's a win on download size.

koops76:
@Hainish I mean minification for the whole extension.

chromium/.eslintignore
jeremyn: Done.

chromium/debugging-rulesets.html
jeremyn: Done.

chromium/external/README.md
jeremyn: Done.

chromium/external/README.md
jeremyn: These instructions are supposed to give someone who is knowledgeable about `npm` the key information needed to reproduce the files: package names, version numbers, options used. It's not supposed to completely walk a beginner through the process of making these files. The user can decide for themselves whether to install the minifier executables globally or not and how to get the files into whatever directories they want.

chromium/external/README.md
jeremyn: See https://github.com/EFForg/https-everywhere/pull/13560#discussion_r151280705.

chromium/external/README.md
jeremyn: See https://github.com/EFForg/https-everywhere/pull/13560#discussion_r151280705.

jeremyn:
I kept the minified files. Someone who wants to see the originals can refer to the new `README.md` for instructions on where to get them. If they need runtime debugging with the originals, they can rebuild the extension for themselves with those files.

PR #13562, opened 2017-11-11
fix duped target

J0WI:


J0WI:
@jeremyn @gloomy-ghost @Hainish this builds are currently broken because of this bug.

PR #13567, opened 2017-11-12
Revert "[Lemde.fr] Remove default_off (#13318)"

Bisaloo:
It was working in https://github.com/EFForg/https-everywhere/pull/13318 so maybe this is just temporary. Maybe we should wait before the release to merge this.


Bisaloo:
It was working in https://github.com/EFForg/https-everywhere/pull/13318 so maybe this is just temporary. Maybe we should wait before the release to merge this.

Bisaloo:
It works again... They're driving me crazy.

PR #13570, opened 2017-11-12
Split landalskilife.* from Wyndham.xml

cschanaj:
cc @J0WI this is ready to be merged. please have a look when you have the time. thanks!!


J0WI:
Since they all have the same configuration and even the same cert I would also be fine with having them in one file. ;)

cschanaj:
Thanks!! :smile:

PR #13571, opened 2017-11-12
Create NoFap.com.xml

koops76:


Bisaloo:
Created in the wrong folder. This is an easy ruleset so we're not losing anything of value here. Anyone willing to work on this can easily write it from scratch. Closing.

PR #13572, opened 2017-11-12
Create GameStorage.xyz.xml

koops76:


Bisaloo:
Website is unreachable over both HTTP and HTTPS. The submitter also deleted their account so closing.

PR #13576, opened 2017-11-12
Create EOL.org.xml

Bisaloo:


src/chrome/content/rules/EOL.org.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+	Invalid certificate:
+		blog.eol.org
+	
+	Refused:
```
J0WI: `cephbase.eol.org`

J0WI:
I have some MCB issues with this site.

src/chrome/content/rules/EOL.org.xml
Bisaloo: This hosts times out on both `http` and `https` for me. Is it working for you?

J0WI: I can reproduce the timeout, but this could also be a temporary issue. I think it worked on http and was refused on https when I tested it.

Bisaloo:
@J0WI, you were right.

- http://cephbase.eol.org works fine.
- https://cephbase.eol.org is refused.

PR #13578, opened 2017-11-12
Create Catbox.moe.xml

ihave100openPRs:


src/chrome/content/rules/Catbox.moe.xml
```diff
@@ -0,0 +1,18 @@
...
+<ruleset name="Catbox.moe">
+	<target host="catbox.moe" />
+	<target host="www.catbox.moe" />
+	<target host="files.catbox.moe" />
+	<target host="nofun.catbox.moe" />
+	<target host="qt.catbox.moe" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://www\.catbox\.moe/" to="https://catbox.moe/" />
```
Hainish: This constitutes a snapping redirect, which we try to avoid.  See https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects

src/chrome/content/rules/Catbox.moe.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+	Invalid Certificate:
+		www.catbox.moe
```
Hainish: I don't see an invalid certificate here.

ihave100openPRs:
@Hainish I get an invalid certificate with both Firefox ESR and Firefox 56, you can try it here: https://www.browserling.com/browse/win/7/firefox/56/https%3A%2F%2Fwww.catbox.moe%2F

I suspect you tested it with Chromium where it doesn't happen. https://www.browserling.com/browse/win/7/chrome/61/https%3A%2F%2Fwww.catbox.moe%2F

src/chrome/content/rules/Catbox.moe.xml
jeremyn: Snapping redirects are generally okay if they redirect "around" a broken URL, in this case avoiding `www` which has an invalid certificate. (This is implied by your link: "if https://foo.fm serves HTTPS correctly".) This is especially true for redirects between `^` and `www`, which should be very low-risk.

ihave100openPRs: @jeremyn Can you merge this one as well? (I ran travis again and fixed some problems FWIW)

jeremyn:
@Hainish This looks ready to merge, can you confirm?

Hainish:
@jeremyn Looks good to me.

PR #13593, opened 2017-11-14
Homotopy type theory

MilanKral:


src/chrome/content/rules/HomotopyTypeTheory.org.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+    Homotopy type theory
+-->
+<ruleset name="HomotopyTypeTheory.org">
+	<target host="homotopytypetheory.org" />
+	<target host="www.homotopytypetheory.org" />
+
```
ihave100openPRs: You can add `<securecookie host=".+" name=".+" />` as well for setting secure cookies :)

PR #13601, opened 2017-11-15
update transifex instructions in contribution doc

wonderchook:
Transifex's interface has changed a bit so I updated the instructions in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -358,7 +358,13 @@ Several of our utilities and our full test suite is written in Python.  Eventual
 
 # Contributing Translations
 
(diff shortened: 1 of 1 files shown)


Bisaloo:
Is there a way HTTPS-Everywhere could work with both Github and Transifex for translations?

In my opinion, the current process has a high entry barrier. For example, I tried to join the translation team on the Tor project yesterday (specifically to check HTTPS-Everywhere translations) but they have some kind of review process and what looks like a quite complex team structure.

I think the [workflow used by Privacy Badger](https://github.com/EFForg/privacybadger/blob/master/doc/Translation.md) is way easier and may help attract more contributors.

jeremyn:
@wonderchook Extremely minor issue, but your markdown has `4.` and `4.` right next to each other. I know GitHub rendering replaces the second `4.` with a `5.`, but since you sequentially numbered the other items, it looks off.

pipboy96:
See https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#contributing-translations:
> We are reviewing our process around translations and currently discussing ways to improve. Translations are still processed under the same entity and those who have an account already, do not need to take action at this time. Thank you for your contributions.

PR #13614, opened 2017-11-17
Run hsts-prune and fix affected rulesets

Bisaloo:


ihave100openPRs:
The number of new preloaded domains in this period is kinda impressive considering that last time IIRC only 4 domains were removed...

Bisaloo:
Yeah, I'm guessing this is due to the release of a new major FF version. So we are using a list to check if some domains need to be removed.

jeremyn:
I merged this but there are some small issues that maybe someone wants to fix in one or more new PRs:

The following rulesets should have their ruleset names and filenames updated (including in the whitelist for some):
* `FlokiNET.is.xml`
* `Floor64.xml`
* `Grok_Learning.com.xml`
* `jobs.ch.xml`
* `Yelp.xml`

The following comments should be deleted:
* [`Grok_Learning.com.xml`](https://github.com/Bisaloo/https-everywhere/blob/0e40d6b5bbf183132e52eaca010620d261cc4af9/src/chrome/content/rules/Grok_Learning.com.xml#L1-L4)
* [`Unblocked.xml`](https://github.com/Bisaloo/https-everywhere/blob/0e40d6b5bbf183132e52eaca010620d261cc4af9/src/chrome/content/rules/Unblocked.xml#L3) (really the entire top comment there should be deleted)

Also:
* A comment for `gmx.it` was removed [here](https://github.com/EFForg/https-everywhere/pull/13614/files#diff-4a70c8385d3691af0f4518f6ca2826a6L26), but it's not in the preload list, see https://chromium.googlesource.com/chromium/src.git/+/62.0.3202.94/net/http/transport_security_state_static.json. According to https://hstspreload.org/?domain=gmx.it it has an invalid certificate chain even though it serves the correct STS header. Maybe there is some problem with the hsts-prune process.

Bisaloo:
> A comment for gmx.it was removed here, but it's not in the preload list, see https://chromium.googlesource.com/chromium/src.git/+/62.0.3202.94/net/http/transport_security_state_static.json. According to https://hstspreload.org/?domain=gmx.it it has an invalid certificate chain even though it serves the correct STS header. Maybe there is some problem with the hsts-prune process.

I am not sure what you mean. I manually removed the outdated comments. I must have made a mistake here. What makes you suggest there is a problem with `hsts-prune.js`?

jeremyn:
I assumed that it was the automated process that incorrectly removed the comment, not you.

PR #13615, opened 2017-11-17
 Create p9e.de.xml 

ihave100openPRs:


src/chrome/content/rules/p9e.de.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="p9e.de">
+	<target host="p9e.de" />
+	<target host="www.p9e.de" />
```
J0WI: `git.p9e.de`

src/chrome/content/rules/searx.fi.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="searx.fi">
+	<target host="searx.fi" />
+	<target host="www.searx.fi" />
```
J0WI: I get a `401` on both targets.

ihave100openPRs: @J0WI It was working before without authentication, I asked its operator and he said that he put it behind a password: https://github.com/asciimoo/searx/issues/1072#issuecomment-347153737

Do you think I should remove it nevertheless?

J0WI: I wounder why the tests are passing since we are not able to test this host.
I would remove it, yes.

ihave100openPRs:
@J0WI Ok, I removed it now

PR #13626, opened 2017-11-17
[LinuxTag.org] Add CAcert platform

Bisaloo:


jeremyn:
The PR itself looks okay. The Travis error seems to be because of an unrelated timeout. Maybe if we wait a while, the error will go away on its own and then we can merge this.

PR #13629, opened 2017-11-18
Update apprcn.com.xml

ivysrono:


jeremyn:
This is not a good rule because all it does is make some specific and (in some cases) cross-domain rewrites. You wrote this rule in October 2017 and asked to delete it in November 2017, and it's still broken in January 2017.

Let's just delete this rule, what do you think?

PR #13631, opened 2017-11-18
Create MSF.org.au.xml

Bisaloo:


src/chrome/content/rules/MSF.org.au.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Invalid certificate:
+		ssociativehistory.msf.org.au
```
J0WI: Typo? `associativehistory.msf.org.au`

PR #13634, opened 2017-11-18
Create SheldrickWildlifeTrust.org.xml

Bisaloo:


src/chrome/content/rules/SheldrickWildlifeTrust.org.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Secure connection failed:
+		sheldrickwildlifetrust.org
+
+-->
+<ruleset name="Sheldrick Wildlife Trust">
```
J0WI: See also `giftshop.sheldrickwildlifetrust.org`

PR #13636, opened 2017-11-18
[nyaa.si] New ruleset

youdly:


jeremyn:
- [ ] Your `securecookie` tag doesn't cover what I think you wanted it to cover, instead you might use `host="^\.nyaa\.si$"`. But in this case you should just use `<securecookie host=".+" name=".+" />` as suggested in [`CONTRIBUTING.md`](https://github.com/EFForg/https-everywhere/blob/45cf7c85151609862175b09c8ee682d204846465/CONTRIBUTING.md#regex-conventions).
- [ ] Also, the indentation for your `securecookie` is mixed spaces and tabs. It should just be one tab.

youdly:
Added wildcard `securecookie`. Fixed tab.

jeremyn:
Thanks, merged.

PR #13639, opened 2017-11-18
[catch.com.au] New ruleset

youdly:


src/chrome/content/rules/Catch.com.au.xml
```diff
@@ -0,0 +1,46 @@
...
+			- image	(m)
+			- link	(m)
+			- rdir	(m)
+			
+	m: certificate mismatch
+	t: timeout on https
+-->
+<ruleset name="Catch.com.au">
+
+	<target host="catch.com.au" />
```
J0WI: `www.catch.com.au`

youdly:
Added `www`.

J0WI:
Thanks!
Please avoid force push for small changes.

PR #13643, opened 2017-11-18
changed to use wildcard for all surge sites

beenotung:


src/chrome/content/rules/Surge.sh.xml
```diff
@@ -1,8 +1,7 @@
 <ruleset name="Surge.sh">
 
 	<target host="surge.sh" />
-	<target host="laughable-jam.surge.sh" />
-	<target host="www.surge.sh" />
+	<target host="*.surge.sh" />
```
youdly: wildcard rule must have at least 3 `<test url>`. See https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/TheFreeDictionary.com.xml . My [understanding](http://surge.sh/help/using-https-by-default) is that each user has its own subdomain. would be nice to have a comment on top to explain the necessity of wildcard rule.

youdly: you can use [Sublist3r](https://github.com/aboul3la/Sublist3r) to query the subdomains.

beenotung:
@youdly I just added some tests.

If your concern is some subdomain users may not want their sites to be served under HTTPS intentionally, I would accept that and explicitly add my sites to the list instead.

J0WI:
Thanks!

youdly:
@beenotung 
My suggestion was only to meet the ruleset requirements (the test url and wilcard comment). The tool I suggested is just for you (in this case) to query some sample subdomains so that you can put down some as `test url`. 
I don't think any user would want to be exempted from HTTPS.

PR #13657, opened 2017-11-19
Remove obsolete reference to non-existent rulesets

cschanaj:
Currently, we lack a way to to manage the reference list nicely. Should we add the below script into the rules-test???

TODO

- [x]  src/chrome/content/rules/Evidon.xml
- [x] src/chrome/content/rules/WorldWildlife.org.xml

```bash
#!/bin/bash

# https://stackoverflow.com/questions/5544322/sed-get-only-the-replaced-string-from-a-multiline-input-omit-unmatched-lines

egrep -l '^\s*[-|+]\s*([^ ]*\.xml)$' *.xml | while read rname; do
  sed -n 's/^\s*[-|+]\s*\([^ ]*\.xml\)$/\1/gp' "$rname"| while read fname; do
    if [ ! -f "$fname" ]; then
      sed -i "/^\s*[-|+]\s*$fname$/d" "$rname"
    fi
  done
done
```

Getting a list of rulesets failing rules-test from Travis log
```bash
sed -n 's/^ERROR src\/chrome\/content\/rules\/\([^ :]*\).*$/\1/gp' rules-log.txt | sort -u
```

cc @Hainish @cowlicks blocked by #13614 because of failing travis,


cschanaj:
@Hainish Thanks for merging #13614, Travis is passing now. Could you please have a look at this when you have the time?

jeremyn:
(This is a general comment, not about @cschanaj 's specific efforts here.)

(Pinging @brainwane @gloomy-ghost @J0WI @wonderchook , FYI, and comments welcome.)

Like a lot of our comments, these ruleset references are a bunch of work for little gain. Rulesets should stand on their own. And, the references aren't useful. What does it matter that [all these rulesets](https://github.com/EFForg/https-everywhere/blob/715404903dda53eb76788daecb79f74a6fdc2b3f/src/chrome/content/rules/AOL.xml#L6-L29) are associated with `AOL.xml`? That should go in a Wikipedia trivia section, not here. Is there any valid use case for us? So I would be okay with deleting all of these comments entirely and never worrying about it again.

However, if there is a need or if people just like keeping this information for some reason, then I propose that we instead add a `group` attribute to the `ruleset` tag with a unique identifier that is shared by all the members of that group. So for example the [`AOL.xml` `ruleset` tag](https://github.com/EFForg/https-everywhere/blob/715404903dda53eb76788daecb79f74a6fdc2b3f/src/chrome/content/rules/AOL.xml#L160) might have this:

```xml
<ruleset name="AOL.com (partial)" group="aol">
...
</ruleset>
```

and all of the rulesets I linked above would also have `group="aol"`. A small script could be written to generate a list of all related rulesets based on this. This would avoid the entire "obsolete reference" problem discussed here. It would also avoid the need to choose a "primary" ruleset which may not be clear. Yet another advantage is being able to assign multiple `group`s, if for some reason someone wanted to do that.

Hainish:
Thank you.  This is superseded by https://github.com/EFForg/https-everywhere/commit/f6cbf575de698d165873889dd0c149b203346a16

PR #13665, opened 2017-11-19
add rule for NowYouKnow.eu

numismatika:


src/chrome/content/rules/NYK.eu.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="NowYouKnow">
+
+	<target host="nowyouknow.eu" />
+	<target host="www.nowyouknow.eu" />
+
+	<target host="cdn.nowyouknow.eu" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://(www\.)?nowyouknow\.eu/" to="https://www.nowyouknow.eu/" />
```
J0WI: Could you add a comment for the mismatch on `nowyouknow.eu`?

PR #13666, opened 2017-11-19
[PSSL.com] Update and simplify ruleset

Bisaloo:


src/chrome/content/rules/PSSL.com.xml
```diff
@@ -1,22 +1,20 @@
...
-		- ^	(mismatched)
+	No working URL known:
+		cdn.pssl.com
 
 -->
 <ruleset name="PSSL.com">
 
 	<target host="pssl.com" />
-	<target host="*.pssl.com" />
+	<target host="www.pssl.com" />
```
J0WI:
```
livechat2.pssl.com
m.pssl.com
```

J0WI: @Bisaloo would you like to finish this rule?

PR #13671, opened 2017-11-20
[geoffg.net] Add ruleset.

bardiharborow:


J0WI:
MCB affects only tracking, so let's remove the platform.

bardiharborow:
@J0WI done.

PR #13686, opened 2017-11-21
Fixing some dev dependency requirements for Debian 9 as well as Fedor…

### LINKED ISSUE #13597: Dev install: "ChromeDriver isn't installed" error on Debian 9.2
Type: other

I'm on Debian 9.2 and successfully ran the `install-dev-dependencies.sh` script. But running the test script gave me this error:

`ChromeDriver isn't installed. Check test/chromium/README.md for instructions on how to install ChromeDriver`

I had installed `chromedriver` successfully:

~~~
$ which chromedriver
/usr/bin/chromedriver
$ /usr/bin/chromedriver --version
ChromeDriver 2.32 (undefined)
$ dpkg -l | grep chromedriver
ii  chromedriver                          62.0.3202.89-1~deb9u1                       amd64        web browser - WebDriver support transitional package
~~~

I dug around in the commit log and found that we used to have clearer instructions about ChromeDriver in that `README.md` file, in b3965a122e907e551d26a33bbbea828641e1e149. Based on that, I created a symbolic link:

`$ sudo ln --symbolic /usr/bin/chromedriver  /usr/lib/chromium/chromedriver`

and after that I was able to successfully run the tests.

So I think we need to improve `test/chromium/README.md` to put those troubleshooting directions back, and potentially to fix `install-dev-dependencies.sh` to put that dependency in the right place in the filesystem.

Heads-up @Hainish.

Hainish:
…a.  Only prompt once at the beginning, with a warning.  Fixes #13597

### DIFF
--- a/install-dev-dependencies.sh
+++ b/install-dev-dependencies.sh
@@ -1,13 +1,30 @@
-#!/bin/bash -ex
+#!/bin/bash
 # Install packages that are necessary and/or useful to build and debug
 # HTTPS Everywhere
-set -o errexit -o xtrace
+set -o errexit
+
+if [ "$1" != "--no-prompt" ]; then
+  echo
+  echo "Warning: Installing the development dependencies for HTTPS Everywhere"
+  echo "may alter your system, installing requirements both within the package"
+  echo "management system and also external binaries."
+  echo
+  echo -n "Are you sure you want to continue? [Y/n]: "
+  read CONTINUE
+  CONTINUE=`echo $CONTINUE | xargs | head -c 1 | awk '{print tolower($0)}'`
+  if [ "$CONTINUE" != "y" ]; then
+    exit
+  fi
+  echo
+fi
 
 if [ $UID != 0 ]; then
   SUDO_SHIM=sudo
 fi
 
 if type apt-get >/dev/null ; then
+  $SUDO_SHIM apt-get update
+  $SUDO_SHIM apt-get install -y lsb-release
   BROWSERS="firefox chromium-browser"
   CHROMEDRIVER="chromium-chromedriver"
   if [[ "$(lsb_release -is)" == "Debian" ]]; then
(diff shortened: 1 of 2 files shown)


install-dev-dependencies.sh
```diff
@@ -65,7 +85,7 @@ elif type dnf >/dev/null ; then
   if [ ! -f /var/lib/dbus/machine-id ]; then
     $SUDO_SHIM sh -c 'dbus-uuidgen > /var/lib/dbus/machine-id'
   fi
-  export PYCURL_SSL_LIBRARY=nss
```
Hainish: nss used to be the default fedora libcurl library, no longer

jeremyn: In that case, also consider removing the NSS cipherlist at https://github.com/EFForg/https-everywhere/blob/e2f93d02fe08ef461c1d7260c9290eedbcd63907/test/rules/src/https_everywhere_checker/http_client.py#L48-L49

install-dev-dependencies.sh
```diff
@@ -1,13 +1,30 @@
...
+#!/bin/bash
 # Install packages that are necessary and/or useful to build and debug
 # HTTPS Everywhere
-set -o errexit -o xtrace
+set -o errexit
+
+if [ "$1" != "--no-prompt" ]; then
+  echo
+  echo "Warning: Installing the development dependencies for HTTPS Everywhere"
+  echo "may alter your system, installing requrements both within the package"
```
Bisaloo: requirements

install-dev-dependencies.sh
Hainish: We should update the OpenSSL ciphers as well, since they are currently pinned at those offered by Firefox 51.

Hainish: Probably a task for another day, though.

Hainish: We should also ensure that the macOS development environment doesn't use NSS.

jeremyn: I added that `_DEFAULT_CIPHERLIST_NSS` line specifically when I was working on Fedora last year. It seems unlikely that macOS needs it. I understand wanting to test it, though.

Hainish: Note that this also doesn't solve all the problems with Fedora, I'm still running into errors when I try to run the test in docker. This should at least get Fedora systems to a point where the dependencies are installed, though.

PR #13687, opened 2017-11-21
[RubyGems.org] Remove wildcard

Bisaloo:


src/chrome/content/rules/GemCutter.org.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Invalid certificate:
+		gemcutter.org
+
+-->
+<ruleset name="GemCutter.org">
+
+	<target host="gemcutter.org" />
```
J0WI: Same for `www`.
The file should be referenced in `RubyGems.xml`.

Bisaloo: When you go through the invalid certificate warning for `www`:

> Fastly error: unknown domain: www.gemcutter.org. Please check that this domain has been added to a service.

src/chrome/content/rules/GemCutter.org.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Invalid certificate:
+		gemcutter.org
```
J0WI: Could you also add a reference back to `RubyGems.xml`?

J0WI:
Thanks!

PR #13690, opened 2017-11-21
Create relink.to.xml

heuser443:


src/chrome/content/rules/relink.to.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="relink.to">
+	<target host="relink.to" />
+  	<target host="relink.us" />
```
J0WI: Please also add the `www` subdomain for both targets.

src/chrome/content/rules/relink.to.xml
```diff
@@ -1,9 +1,14 @@
...
+  	<target host="www.relink.to" />
   	<target host="relink.us" />
+  	<target host="www.relink.us" />
 
 	<!-- http://relink.us/ redirects to http://relink.to/, but https://relink.us/ shows error 502 -->
-	<rule from="^http://relink\.us/"
+	<!-- http://www.relink.us/ redirects to http://relink.to/, but https://www.relink.us/ is a redirect loop -->
+	<rule from="^http://(www\.)?relink\.us/"
+		to="https://relink.to/" />
+	<rule from="^http://www\.relink\.to/"
```
J0WI: This rule can be omitted, `https://www.relink.to/` WFM.

PR #13691, opened 2017-11-21
Update ImageShack.xml

heuser443:
imagizer.imageshack.com is an alias for imagizer.imageshack.us.

### DIFF
--- a/src/chrome/content/rules/ImageShack.xml
+++ b/src/chrome/content/rules/ImageShack.xml
@@ -15,6 +15,10 @@
 	<target host="imagizer.imageshack.com" />
 
 	<target host="imageshack.us" />
+	<target host="www.imageshack.us" />
+	<target host="a.imageshack.us" />
+	<target host="imagizer.imageshack.us" />
+	<target host="post.imageshack.us" />
 
 	<!-- Show 404 page: -->
 		<test url="http://a.imageshack.us/img254/8219/40789751.png" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ImageShack.xml
```diff
@@ -12,19 +13,22 @@
...
 		<test url="http://img5.imageshack.us/img5/133/homeworld.jpg" />
 		<test url="http://img46.imageshack.us/img46/4804/intel1jk8.gif" />
 		<test url="http://img530.imageshack.us/img530/133/homeworld.jpg" />
 
 	<securecookie host=".+" name=".+" />
 
-	<rule from="^http://(www\.|api\.|blog\.)?imageshack\.com/" to="https://$1imageshack.com/" />
+	<rule from="^http://(www\.|api\.|blog\.|imagizer\.)?imageshack\.com/" to="https://$1imageshack.com/" />
 
 	<rule from="^http://(www\.|a\.|imagizer\.|post\.)?imageshack\.us/" to="https://$1imageshack.us/" />
```
J0WI: Can we replace all redirect rules with one simple `<rule from="^http:" to="https:" />`?

heuser443: Only if we can make sure that all broken imageshack.us subdomains are known and there are exclude patterns matching all of them. Otherwise the *.imageshack.us target will cause problems.

J0WI: I'm not aware of any other domains than the listed ones, but I'm also fine with keeping it for now.

PR #13692, opened 2017-11-21
Create abload.de.xml

heuser443:


ihave100openPRs:
Please document its subdomains that have invalid certificates such as h13.abload.de

heuser443:
I do not see any use of mentioning a probably incomplete list of internal subdomains, of which some have likely just accidentially found a way into search engine indexes without actually being intended/used for serving web pages, but there you go.

heuser443:
Hm, after failing without an obvious reason on my previous commit, now it looks like my last commit did not trigger a Travis CI build at all. I think something is wrong with the build/checking process.

ihave100openPRs:
@heuser443 You can close and then re-open this PR so that Travis gets triggered.

heuser443:
@ihave100openPRs, thanks for the advice. Check was successful now, as expected. Must have been a temporary issue with Travis CI in my two attempts before.

src/chrome/content/rules/abload.de.xml
```diff
@@ -0,0 +1,36 @@
...
+	Invalid certificate:
+		h1.abload.de
+		h2.abload.de
+		...
+		h19.abload.de
+		pool.abload.de
+
+-->
+
+<ruleset name="abload.de">
```
J0WI: Please rename the rule to `abload.de (partial)`.

J0WI:
Thanks!

PR #13694, opened 2017-11-21
Create ncrypt.in.xml

heuser443:


heuser443:
I need some advice how to fix the failing check here. The error message is unfortunately completely meaningless to me and I failed to find any documentation on it.

heuser443:
Alright, the check now passed after an arbitrary change to the file. So it looks like the checking process is not deterministic here. #13706 likely has the same problem.

src/chrome/content/rules/ncrypt.in.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="ncrypt.in">
+	<target host="ncrypt.in" />
```
J0WI: Please also add `www.ncrypt.in`

J0WI:
Please have a look at our policy for snapping redirects: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects

J0WI:
Thanks!

PR #13701, opened 2017-11-22
Fix path to .eslintignore in build exclusions

Bisaloo:
`.eslintignore` is added to the packaged extension (see [here](https://github.com/EFForg/https-everywhere/pull/13693/files#diff-ac9aed2a391263bbc3c8236f8e674155R3)). It was added to `build_exclusions` in #13560 but it looks like the path was wrong.

@Hainish @jeremyn

### DIFF
--- a/.build_exclusions
+++ b/.build_exclusions
@@ -15,4 +15,4 @@ chrome-resources/update-from-chrome-svn.sh
 webextension/chrome-resources/update-from-chrome-svn.sh
 .eslintrc.json
 webextension/.eslintrc.json
-chromium/.eslintignore
+.eslintignore


jeremyn:
I did that, sorry. Thanks for fixing it.

Based on a closer reading of the code -- [`make.sh`](https://github.com/EFForg/https-everywhere/blob/master/make.sh), [`utils/create_xpi.py`](https://github.com/EFForg/https-everywhere/blob/master/utils/create_xpi.py), and [`utils/zipfile_deterministic.write_from_directory`](https://github.com/EFForg/https-everywhere/blob/master/utils/zipfile_deterministic.py#L1141-L1154) -- it looks like `pkg/crx/` is the root directory for `.build_exclusions`, and `pkg/crx/` is first rsync'd from `chromium/`, so really `chromium/` is the root for `.build_exclusions`. Is that right?

Bisaloo:
Yes, I think so. I also built the `xpi` from this branch and `.eslintignore` was indeed gone.

PR #13707, opened 2017-11-22
NetCologne:

numismatika:
- check subdomains and re-enable the ruleset

Webtv is called nettv now so we can enable the ruleset again

### DIFF
--- a/src/chrome/content/rules/NetCologne.xml
+++ b/src/chrome/content/rules/NetCologne.xml
@@ -1,60 +1,103 @@
-
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://webtv.netcologne.de/ => https://webtv.netcologne.de/: (6, 'Could not resolve host: webtv.netcologne.de')
-
-	NetCologne Gesellschaft für Telekommunikation mbH
+<!--NetCologne Gesellschaft für Telekommunikation mbH
 
 	Other NetCologne rulesets:
 
 		- DotCologne.de.xml
 
-	Insecure cookies are set for these hosts:
+	Cert expired:
+		- gk-shop
+		- hosting
 
-		- tracking.netcologne.de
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/NetCologne.xml
```diff
@@ -1,60 +1,90 @@
...
-
-	<!--
-	timed out:
-	netty.netcologne.de
-	speedtest.netcologne.de
-	plg.netcologne.de
-	merkur.netcologne.de
-	websitemakertest.netcologne.de
-	news.netcologne.de
-	sitestest[1-3].netcologne.de
```
J0WI: These hosts do still exist. Can we keep the comments?

src/chrome/content/rules/NetCologne.xml
```diff
@@ -1,60 +1,90 @@
...
-	<target host="musicbox.netcologne.de"/>
-	<target host="netcologne.de"/>
-	<target host="netkompakt-fuer-mitarbeiter.netcologne.de"/>
-	<target host="onlineservice.netcologne.de"/>
-	<target host="planauskunft.netcologne.de"/>
-	<target host="shop.netcologne.de"/>
-	<target host="statistik.netcologne.de"/>
-	<target host="tracking.netcologne.de"/>
-	<target host="webtv.netcologne.de"/>
-	<target host="www.netcologne.de"/>
```
J0WI: The following hosts have been removed:
```
netcologne.de
www.netcologne.de
aktion.netcologne.de
```
Please keep the rule for them or add a comment why you removed them.

src/chrome/content/rules/NetCologne.xml
numismatika: sure, just a collateral damage of sorting and deduplication of the subdomain test results.

PR #13712, opened 2017-11-23
[Kogan.com] New ruleset

youdly:


J0WI:
Rule LGTM, but I think it's better to split it in multiple files.

J0WI:
Please define one ruleset as "parent" and list all references there. All other files should just refer to the list in this file. Otherwise we have to update the list in all files if something needs to be changes. See `Amazon.xml` for an example.

youdly:
Updated comment

PR #13714, opened 2017-11-23
Reactivate transfer.sh.xml

ihave100openPRs:


src/chrome/content/rules/transfer.sh.xml
```diff
@@ -1,14 +1,8 @@
-
-<!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://www.transfer.sh/ => https://www.transfer.sh/: (35, 'error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure')
-
-
-
--->
-<ruleset name="transfer.sh" default_off='failed ruleset test'>
+<ruleset name="transfer.sh" >
```
J0WI: Could you also add a hint about the wildcard DNS cert/record?

ihave100openPRs: @J0WI Done

PR #13715, opened 2017-11-23
Create AlJazeera.com.xml

Bisaloo:


ihave100openPRs:
No `<securecookie host=".+" name=".+" />` in this case?

Bisaloo:
@ihave100openPRs, we've had bad cases of ruleset bugs in the past due to securecookie so I don't really feel confident adding it for complex domains when I haven't done extensive tests.

src/chrome/content/rules/AlJazeera.com.xml
```diff
@@ -0,0 +1,59 @@
...
+		projects.aljazeera.com
+		login.stream.aljazeera.com
+		staff.stream.aljazeera.com
+		studio.aljazeera.com
+
+	No working URL known:
+		ajnsearch.aljazeera.com
+		login.america.aljazeera.com
+		bc05.aljazeera.com
+		interactive.aljazeera.com
```
J0WI: Search for `site:interactive.aljazeera.com`

PR #13721, opened 2017-11-23
Remove old icons

Bisaloo:
Those files have the old version of HTTPS Everywhere icon. It is not used anymore.


jeremyn:
The appropriate entries should be removed from https://github.com/EFForg/https-everywhere/blob/568d2c58f7ec1c000a3f243afe0d97ed899b2d50/src/META-INF/manifest.mf or that file should be regenerated. (I'm not sure how to regenerate it.)

The reference to `icon.jpg` should also be removed from https://github.com/EFForg/https-everywhere/blob/992009f5db5442b53e2ba233583a21489f6c883f/make.sh#L56

Bisaloo:
@jeremyn, I'm not sure how to regenerate `manifest.mf` either. I thought it was done automatically on release but I can't do it using `make.sh`.

Hainish:
@Bisaloo you don't have to worry about regenerating `manifest.mf`, this is delivered to us by AMO during the release process.

Hainish:
Thank you

PR #13729, opened 2017-11-24
Update EFF.xml

koops76:


J0WI:
The change LGTM, but having the list of related files in every ruleset makes it hard to maintain.
Closing because the repo/user has been deleted. There is no functional change anyway.

PR #13732, opened 2017-11-25
 Add Travis test for obsolete ruleset references 

cschanaj:
Failing Travis is expected, blocked by #13657. This will help to keep the ruleset references up-to-date when we have some bulk ruleset updates, e.g. when running `hsts-prune` etc. The effort required by contributors should be minimal in each PR as they are only responsible for ruleset they have changed.

ping @Hainish @cowlicks @J0WI @jeremyn for opinions and reviews. Thanks!!

EDIT: ~It seem that when the script is ran in debug mode, the output doesn't fit Travis console....~

Where to `chmod +x` ???

### DIFF
--- a/test/travis.sh
+++ b/test/travis.sh
@@ -39,6 +39,7 @@ set -e
 if [ "$TEST" == "rules" ]; then
   echo >&2 "Performing comprehensive coverage test."
   docker run --rm -ti -v $(pwd):/opt httpse python utils/ruleset_filenames_validate.py
+  docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/remove-obsolete-references.sh"
   docker run --rm -ti -v $(pwd):/opt httpse bash -c "utils/validate.sh"
   docker run --rm -ti -v $(pwd):/opt httpse bash -c "test/rules.sh"
   docker run --rm -ti -v $(pwd):/opt node bash -c "cd /opt && node utils/normalize-securecookie.js"
--- a/utils/remove-obsolete-references.sh
+++ b/utils/remove-obsolete-references.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# Remove obsolete references to child rulesets which
+# has been renamed/ deleted
+
+
+# Change directory to git root; taken from ../test/test.sh
+if [ -n "$GIT_DIR" ]
+then
+    # $GIT_DIR is set, so we're running as a hook.
+    cd $GIT_DIR
+else
+    # Git command exists? Cool, let's CD to the right place.
+    git rev-parse && cd "$(git rev-parse --show-toplevel)"
+fi
+
+# Run from ruleset folder to simplify the output
+cd src/chrome/content/rules
+
+# Default exit status
+EXIT_CODE=0
+
+# List of file(s) which contain at least one reference
+FILES=`egrep -l '^\s*[-|+]\s*([^ ]*\.xml)\s*$' *.xml`
+
+while read FILE; do
+    # List of referenced rulesets
+    REFS=`sed -n 's/^\s*[-|+]\s*\([^ ]*\.xml\)\s*$/\1/gp' "$FILE"`
+    
+    while read REF; do
+        if [ ! -f "$REF" ]; then
+            echo >&2 "ERROR src/chrome/content/rules/$FILE: Dangling reference to $REF"
+            sed -i "/^\s*[-|+]\s*$REF$/d" "$FILE"
+            EXIT_CODE=1
+        fi
+    done <<< "$REFS"
+done <<< "$FILES"
+
+# Exit with errors, if any
+exit "$EXIT_CODE"


jeremyn:
My comment at https://github.com/EFForg/https-everywhere/pull/13657#issuecomment-346944585 applies here too.

Hainish:
I'll take a look at this (as well as #13657) next week.

Hainish:
Thanks, this looks good to me.  A future improvement would be to remove the entire comment, or at least the `Other [Ruleset] rulesets:` text and blank lines, surrounding the list if all entries in the list are removed.

cschanaj:
@Bisaloo This is expected since a generic way to cover all the variants introduce false positive results which needed to be cleanup before merging. Personally, I prefer the regex to be strict such that any contributor who want the references to be tracked can _opt-in_ this test (since the inclusions of the references are more or less optional anyway). 

You might be interested in the following command (slow) to find all the dangling references. There is 300 such dangling references and I suspect a lot of them are actually false positive. 

```bash
for f in *.xml; do 
  egrep -o "[a-zA-Z0-9][-a-zA-Z0-9._]*\.xml" "$f" | while read x; do 
    if [ ! -f "$x" ]; then
       echo "$x $f"; 
    fi; 
  done; 
done | sort -u
```

Bisaloo:
>  I prefer the regex to be strict such that any contributor who want the references to be tracked can opt-in this test (since the inclusions of the references are more or less optional anyway).

Maybe we should add something about this in `CONTRIBUTING.md` then.

jeremyn:
I don't like the idea of programmatically checking comments in general. If we want tests to fail then we should move the references out of the comments.

Also this test makes it awkward to work on files that refer to each other, in separate PRs. See PR https://github.com/EFForg/https-everywhere/pull/14258#issuecomment-357244015 for an example.

Probably we should move this into a periodic clean-up task, like we do with `hsts-prune`, rather than a Travis test.

jeremyn:
`hsts-prune` isn't a great example since we run it in Travis, but I hope my point was clear.

Bisaloo:
I agree with @jeremyn's comment in https://github.com/EFForg/https-everywhere/pull/13657#issuecomment-346944585 this is starting to  become too much of a hassle.

I don't see any clear benefit for those comments. IMO, we should only keep them when `target`s *from the same domain* are scattered over more than one ruleset. Otherwise, I am voting for removing them completely.

Should we open a new issue to discuss this?

Hainish:
@Bisaloo I'd rather not create the overhead of a new issue if we're in agreement here, which I suspect we are for the purposes of a stop-gap solution.

I agree with @jeremyn's assessment that this doesn't need to be a test built into our CI, and it's already coded in a way that will allow us to run it as a periodic clean-up task.  If we're all in agreement, I can remove this line from `travis.sh`.

Bisaloo:
@Hainish, we are not only talking about removing this task from the tests but removing/not adding "useless" references in general.

@jeremyn probably explained it better than me in https://github.com/EFForg/https-everywhere/pull/13657#issuecomment-346944585.

Bisaloo:
See for example [`Conde-Nast.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Conde-Nast.xml). I am not sure why we care about the list of the companies they own.

Hainish:
@Bisaloo I've seen that proposal, but that would take a lot more work and is a more comprehensive overhaul.  I'm fine with opening up a discussion for how to deal with related ruleset references if it's becoming an increasing maintenance headache.  I suspect when the ruleset list was still small, years ago, the informal process of adding the references in comments made more sense than it does today - that warrants discussion.  But we should deal with the immediate problem first.

jeremyn:
@Hainish What is the immediate problem?

Hainish:
@jeremyn this:

> I don't like the idea of programmatically checking comments in general. If we want tests to fail then we should move the references out of the comments.
> 
> Also this test makes it awkward to work on files that refer to each other, in separate PRs. See PR #14258 (comment) for an example.

Hainish:
The above can be solved very easily, by just removing this line.  At the same time, we should create a list of periodic clean-up tasks (probably in the wiki).

Bisaloo:
As a first step, we should at least stop adding this kind of comments then and remove them when we get the chance.

jeremyn:
The questions of what to do with this test and what to do with references are related. If we want to remove references entirely then we should remove the test, but if we want to keep references then we can think about improving the test.

Hainish:
@Bisaloo I think there is a use to having ruleset references included.  It makes it clearer to discern what file targets should be added to, when it's ambiguous.  For instance, if someone discovers a new subdomain for `fbcdn.net` which should be covered, they may not know that a ruleset file exists for that specific scope.  If they open `Facebook.xml` and it does not reference `FbCDN.net.xml`, they're likely to miss it and simply add that subdomain to the wrong file.

@jeremyn I think there is a simple solution to the immediate problem: removing the test from CI.  If you're running into problems with this test, we should simply remove it for now.  Then we can consider the long-term questions in a separate issue: if we want to improve it, how to improve it, and what to do with ruleset references in general.

jeremyn:
@Hainish The `fbcdn.net` example could be solved by encouraging contributors to grep for domains before adding them.

I think the current test is better than nothing if we want to keep references. We can keep it in place while we have a larger discussion.

PR #13744, opened 2017-11-25
add rule for Urban-Media

numismatika:


src/chrome/content/rules/Urban-Media.xml
```diff
@@ -0,0 +1,45 @@
...
+			- pdf
+			- sip
+			- sportsdata
+			- static
+			- talkabout
+			- tracking
+			- www
+			- zh
+-->
+<ruleset name="Urban Media">
```
J0WI: Could you add `(partial)` to the name? Their HTTPS coverage is very poor.

PR #13754, opened 2017-11-25
Create Conrad_Electronic.xml

heuser443:


src/chrome/content/rules/Conrad_Electronic.xml
```diff
@@ -0,0 +1,42 @@
+<!--
+	Invalid certificate:
+		produktinfo.conrad.com (expired)
+		www.produktinfo.conrad.com (expired, mismatch)
+		
+	Redirect to HTTP:
+		conrad.com
+		www.conrad.com
```
J0WI: On `conrad.com` you find a list of some more TLD used by this company.
It would be great if we can secure/comment all of them. See also our section about [enumerating subdomains](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains).
You may split the TLD into different files to avoid huge, complex rules that are hard to maintain. See `Amazon.xml` for an example.

J0WI:
Please double check your syntax, there are a lot of missing quotes.

src/chrome/content/rules/Conrad_Electronic.xml
```diff
@@ -0,0 +1,119 @@
...
+		to="https://produktinfo.conrad.com/"/>-->
+
+	<!-- fix sites which refuse connection without www subdomain-->
+	<rule from="^http://conrad\.at/"
+		to="https://www.conrad.at/" />
+	
+	<rule from="^http://conrad\.ch/"
+		to="https://www.conrad.ch/" />	
+		
+	<rule from="^http://biz-conrad\.ch/"
```
J0WI: `biz-conrad.ch` needs to be added as target

heuser443:
What would be the correct way to handle this `Non-2xx HTTP code: http://api.conrad.com/ (200) => https://api.conrad.com/ (404)` error? Also note: Requesting HTTPS ressources that are in fact used on the site (like https://api.conrad.com/search/1/campaigns/redirects/de/de/b2c?apikey=eB6SoveLt6wWbRQuIeYtZnBAL6lUo38b) via HTTP swaps the status codes. While HTTPS now results in 200, HTTP requests result in 404. So adding a exclude pattern and/or test urls will probably not help here.

The Domain is used for XHR requests (already via HTTPS) on e.g. conrad.de and server includes a `Strict-Transport-Security: max-age=31536000` so it should be safe to include. But this header (and being a target a user will not likely call directly) also would make it much less problematic to leave out this target.

src/chrome/content/rules/Conrad_Electronic.xml
```diff
@@ -0,0 +1,119 @@
...
+	<target host="conrad.pl" />
+	<target host="www.conrad.pl" />
+	<target host="conrad.se" />
+	<target host="www.conrad.se" />
+	<target host="m.conrad.se" />
+	<target host="conrad.si" />
+	<target host="www.conrad.si" />
+	<target host="m.conrad.si" />
+		
+	<target host="api.conrad.com" />
```
J0WI: This target returns a 404 on https. You need to find a valid test url. For the tests you need to add an exclusion for `^http://api\.conrad\.com/$`.

heuser443: Please see my comment above, I expect it to fail also. Or is my assumption wrong?

J0WI: Sure, you can comment it as `different content`, `status code mismatch` or similar.

J0WI:
Thanks!

PR #13757, opened 2017-11-26
Improve CONTRIBUTING.md

brainwane:
Suggest RegExr, and fix minor stylistic issues.

I'm totally open to resplitting this into multiple commits, revising it, and so on. :)


jeremyn:
@brainwave Thanks for the update.

I would prefer we not recommend RegExr, or anything else like that. It can come across like an endorsement, both us endorsing them and vice versa, which may cause problems. It might also obligate us as maintainers to help contributors troubleshoot it.

(Also, I personally tend to use https://regex101.com/ for regex checking. I don't know what the other maintainers use.)

Just as an FYI, you might find issue https://github.com/EFForg/https-everywhere/issues/7277 interesting, where we discussed mentioning Sublist3r, another third-party tool.

brainwane:
Ready for re-review.

jeremyn:
Thanks, merged.

PR #13769, opened 2017-11-27
Update MegaGlest.org.xml

### LINKED ISSUE #13765: redirection does not work properly with HTTPS Everywhere enabled
Type: other

Domain: translate.megaglest.org

The URL:
http://translate.megaglest.org/
should redirect to:
https://docs.megaglest.org/MG/Translations

but with "HTTPS Everywhere" enabled it does not redirect properly.
On Parabola with GNU IceCat 52.3.0 (64-bit)
HTTPS Everywhere 5.2.21
it redirects to:
https://docs.megaglest.org/MGTranslations

On Mac OS X with Mozilla Firefox 57.0 (64-bit)
HTTPS Everywhere 2017.11.21
it redirects to:
https://docs.megaglest.org/MegaGlestTranslations

Is this a bug in "HTTPS Everywhere"?

Disabling the extension makes redirection work properly in both cases.

I talked to Tom Reynolds from MegaGlest and we came to the conclusion that I file this bug report.
Tom:
This is the series of redirections I get to see here (indeed a lot):
> $ curl --head --location http://translate.megaglest.org/
> HTTP/1.1 301 Moved Permanently
> Server: nginx/1.6.2
> Date: Fri, 17 Nov 2017 18:53:26 GMT
> Location: http://wiki.megaglest.org/Translations
> X-IPLB-Instance: 10216
>
> HTTP/1.1 301 Moved Permanently
> Server: nginx/1.6.2
> Date: Fri, 17 Nov 2017 18:53:26 GMT
> Location: http://docs.megaglest.org/MG/Translations
> X-IPLB-Instance: 10215
>
> HTTP/1.1 301 Moved Permanently
> Date: Fri, 17 Nov 2017 18:53:31 GMT
> Content-Type: text/html
> Content-Length: 178
> Connection: keep-alive
> Keep-Alive: timeout=20
> Location: https://docs.megaglest.org/MG/Translations
> Server: MegaGlest 61.357
>
> HTTP/1.1 200 OK
[where the actual HTML page is then rendered]

Me: 
https://wiki.megaglest.org
is not reachable.
Could this be the problem?
Shall I file a bug report to the extension?

Tom:
I cannot remotely disagnose which part of the client software is the problem here - but I agree that "HTTPS Everywhere" is the most suspect based on your report. https://wiki.megaglest.org will indeed not work, there is no HTTPS configured there at this time. We could change this, but not to work around a bug in client software (but we could do it to enable more use of HTTPS and the general move towards replacing HTTP by HTTPS).

heuser443:
Fix #13765, also remove snapping redirect skipping https://docs.megaglest.org/MG

### DIFF
--- a/src/chrome/content/rules/MegaGlest.org.xml
+++ b/src/chrome/content/rules/MegaGlest.org.xml
@@ -2,7 +2,8 @@
 	Refused:
 		- faq		(→ docs)
 		- readme	(→ raw.github.com)
-		- wiki		(→ docs)
+		- translate	(→ docs)
+  		- wiki		(→ docs)
 -->
 <ruleset name="MegaGlest.org">
 	<target host="megaglest.org" />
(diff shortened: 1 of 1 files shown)


jeremyn:
@heuser443 I get what you're doing here, but I think this is the wrong approach. These sort of deep redirects, like http://translate.megaglest.org to https://docs.megaglest.org/MG/Translations , are generally bad practice. We've also caused a specific problem for this site.

Let's just remove all `target`s and `rule`s that don't support a direct rewrite. We'll lose some security, but I think that's fine here. The site's users and owners can work toward supporting HTTPS better if they want to.

jeremyn:
Please:

- [ ] Make the [top comment](https://github.com/heuser443/https-everywhere/blob/c7512a239096d40df4584a13a3d10ae61768d456/src/chrome/content/rules/MegaGlest.org.xml#L1-L7) look like this:

```xml
<!--
	Secure connection failed:
		faq.megaglest.org
		readme.megaglest.org
		translate.megaglest.org
		wiki.megaglest.org

-->
```

- [ ] Completely delete [the redirect comments](https://github.com/heuser443/https-everywhere/blob/c7512a239096d40df4584a13a3d10ae61768d456/src/chrome/content/rules/MegaGlest.org.xml#L14-L35)

jeremyn:
@heuser443 Can we move forward on this one?

jeremyn:
I'm closing this pull request in favor of my pull request https://github.com/EFForg/https-everywhere/pull/14104.

Issue #13603, opened 2017-11-15
eBay Ruleset


BK1233:
Type: ruleset issue

Domain: ebay.at
Domain: ebay.be
Domain: ebay.ca
Domain: ebay.ch
Domain: ebay.co.uk
Domain: ebay.com
Domain: ebay.com.au
Domain: ebay.com.hk
Domain: ebay.com.my
Domain: ebay.com.sg
Domain: ebay.de
Domain: ebay.es
Domain: ebay.fr
Domain: ebay.ie
Domain: ebay.in
Domain: ebay.it
Domain: ebay.nl
Domain: ebay.ph
Domain: ebay.pl
Domain: ebay.ru

Most of eBay is now HTTPS by default except for the stores.ebay domains.

youdly:
- [x] ebayimg.com
- [x] ebayrtm.com
- [x] ebaystatic.com
- [ ] ebay.at
- [ ] ebay.be
- [ ] ebay.ca
- [ ] ebay.ch
- [ ] ebay.co.uk
- [ ] ebay.com
- [ ] ebay.com.au
- [ ] ebay.com.hk
- [ ] ebay.com.my
- [ ] ebay.com.sg
- [ ] ebay.de
- [ ] ebay.es
- [ ] ebay.fr
- [ ] ebay.ie
- [ ] ebay.in
- [ ] ebay.it
- [ ] ebay.nl
- [ ] ebay.ph
- [ ] ebay.pl
- [ ] ebay.ru

J0WI:
See also the list on `ebayclassifiedsgroup.com`

youdly:
So that's where `mobile.de` comes from. Maybe I'll work on them later.
As for the comments in the rulesets, I wouldn't group them under `ebay.com`, I think it's more suitable to group them under `ebayclassifiedsgroup.com`.
e.g. `mobile.de.xml` -> `ebayclassifiedsgroup.com.xml` -> `ebay.com.xml`.

J0WI:
See also `ebayinc.com`

youdly:
`ebayinc.com` is already HSTS-preloaded

youdly:
All PRs have been merged, this issue can be closed.

PR #13774, opened 2017-11-27
[ebay.at] New ruleset

youdly:


youdly:
`k2b-bulk` WFM on `curl` and `openssl`.

PR #13777, opened 2017-11-27
[ebay.ca] New ruleset

youdly:


src/chrome/content/rules/ebay.ca.xml
```diff
@@ -0,0 +1,345 @@
...
+
+	For other eBay coverage, see ebay.com.xml	
+	
+	
+	Non-functional subdomains:
+	
+		- ebay.ca	(h)
+		- arribada	(t)
+		- buyvrfn	(m)
+		- cafr	(t)
```
J0WI: WFM

youdly:
Added `cafr` subdomain. WFM too.

J0WI:
Looks like we need to redirect them to `www`, just like `^`.

youdly:
Added redirect

PR #13778, opened 2017-11-27
[ebay.com.au] New ruleset

youdly:


src/chrome/content/rules/ebay.com.au.xml
```diff
@@ -0,0 +1,295 @@
+<!--
+
+	For other eBay coverage, see ebay.com.xml	
+	
+	
+	Non-functional subdomains:
+	
+		- ebay.com.au	(h)
```
J0WI: Can be redirected to secure www.

youdly: Added redirect

PR #13791, opened 2017-11-28
Fixed and updated cloudbees.com ruleset.

Jochen-A-Fuerbacher:


src/chrome/content/rules/cloudbees.com.xml
```diff
@@ -1,32 +1,67 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://wiki.cloudbees.com/ => https://wiki.cloudbees.com/: (6, 'Could not resolve host: wiki.cloudbees.com')
+	Invalid certificate:
+		info.cloudbees.com
+		join.cloudbees.com
+		resource.cloudbees.com
+
+	Not found:
```
J0WI: There are no DNS records for these hosts, so you can remove them.

PR #11419, opened 2017-07-27
New ruleset for Kiwix

galenide:
At the current time, for the Kiwix site, it appears that only the `download` subdomain supports HTTPS. Even so, this may be useful if, for example, a user on an external site accesses a link such as http://download.kiwix.org/zim/wikipedia_hi_medicine.zim.

In addition, it may be that the `mirror.download` subdomain is intended to identically mirror the contents of the `download` subdomain. If that is the case, then there is the question of whether it would be useful to have the ruleset redirect requests from `http://mirror.download.kiwix.org/` to `https://download.kiwix.org/`. At the same time, there is the possibility as to whether redirecting such requests could lead to redirect loops or, for that matter, if the content under the `mirror.download` and `download` subdomains is not quite identical.

PR #13793, opened 2017-11-28
Add Kiwix.org.xml

jeremyn:
This replaces PR #11419.

### DIFF
--- a/src/chrome/content/rules/Kiwix.org.xml
+++ b/src/chrome/content/rules/Kiwix.org.xml
@@ -1,0 +1,32 @@
+<!--
+	Connection refused:
+		kiwix.org
+		input.kiwix.org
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Kiwix.org.xml
```diff
@@ -0,0 +1,32 @@
...
+		bugs.kiwix.org
+		code.kiwix.org
+		compile.kiwix.org
+		forum.kiwix.org
+		ios.kiwix.org
+		microblog.kiwix.org
+		reportabug.kiwix.org
+
+-->
+<ruleset name="Kiwix.org">
```
J0WI: How about `Kiwix.org (partial)`?

PR #13798, opened 2017-11-29
[AFS] Add ruleset.

bardiharborow:


J0WI:
The "Find AFS in your country" on afs. org lists even more ccTLD related to AFS.
You should create one ruleset for each ccTLD.

bardiharborow:
@J0WI I've added the rest of the ccTLDs. Almost all of them appear to be served through a CNAME pointing at `*.afssite.afs.org` (though Cloudflare proxies traffic, so it shows as an A record), try for example `australia.afssite.afs.org`. As they are mostly just localised versions of the same website it made more sense to me to keep them in the same ruleset from maintenance reasons (i.e. it's far easier to notice where the gaps are, just as you did before, when it it's not spread across multiple files). I can still split it if you'd like though.

J0WI:
Since the behavior is consistent I'm fine with having them in one file. Thanks!

PR #13809, opened 2017-11-29
Added ruleset for bamf.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/bamf.de.xml
```diff
@@ -0,0 +1,36 @@
+<!--
+	Connection failed:
+		oet.bamf.de
+		zirf.bamf.de
```
J0WI: `webgis.bamf.de`

PR #13819, opened 2017-12-01
[ebay.fr] New ruleset

youdly:


src/chrome/content/rules/ebay.fr.xml
```diff
@@ -0,0 +1,242 @@
...
+	i: invalid certificate chain
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="eBay.fr">
+
+	<target host="www.ebay.fr" />
+	<target host="actualites.ebay.fr" />
```
J0WI: I get some MCB issues here

src/chrome/content/rules/ebay.fr.xml
```diff
@@ -0,0 +1,242 @@
+<!--
+
+	For other eBay coverage, see ebay.com.xml
+	
+	
+	Non-functional subdomain:
+	
+		- ebay.fr	(h)
```
J0WI: Please add this host and redirect to secured www.
This also affects other PRs in #13603.

youdly: Added redirect

PR #13844, opened 2017-12-02
Create Yagg.com.xml

Bisaloo:
My understanding is that this website nearly shut down about a year ago. So all the other (mismatched) subdomains listed by `Sublist3r` will slowly stop being used.


jeremyn:
- [ ] Add `(partial)` to the `ruleset` `name`

jeremyn:
We should have some kind of automated test to detect ruleset files outside of the rules directory.

PR #7376, opened 2016-10-19
Fix math.uwaterloo.ca policy

tomhukins:
http://www.math.uwaterloo.ca/tsp/pubs/ loads successfully but
https://www.math.uwaterloo.ca/tsp/pubs/ does not.

PR #13845, opened 2017-12-02
Update University-of-Waterloo.xml

jeremyn:
This pull request updates `University-of-Waterloo.xml` mainly to fix the problem with http://www.math.uwaterloo.ca/tsp/pubs/ reported in PR https://github.com/EFForg/https-everywhere/pull/7376 and to get the tests to pass, which was blocking the earlier PR.

Regarding the complex rewrite rules I deleted for `campaign.uwaterloo.ca` etc: it's not obvious what some of the regexes do, some of the rewrites are incorrect (for example check the search results for `site:hr.uwaterloo.ca -site:www.hr.uwaterloo.ca`), and they are generally bad practice.

### DIFF
--- a/src/chrome/content/rules/University-of-Waterloo.xml
+++ b/src/chrome/content/rules/University-of-Waterloo.xml
@@ -1,64 +1,23 @@
 <!--
-	Nonfunctional subdomains:
-
-		- cacr		("Gone")
-
-
-	Problematic subdomains:
-
-		- campaign	(refused)
-		- development	(refused)
-		- cecs		(shows info; mismatched, CN: uwaterloo.ca)
-		- hr		(redirects to ^.../$)
-		- info		(redirects to ^.../$, valid cert)
-
-
-	Partially covered subdomains:
-
-		- campaign	(→ ^)
-		- info		(→ ^)
-
-
-	Fully covered subdomains:
-
-		- (www.)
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/University-of-Waterloo.xml
```diff
@@ -1,64 +1,23 @@
...
-		- cecspilot.cecssys
-		- crysp
-		- (www.)cs
-		- csclub
-		- development		( → ^)
-		- hr			(→ ^)
-		- (www.)math
-		- ripple
+	Invalid certificate and different HTTP/HTTPS content:
+		cacr.uwaterloo.ca
```
J0WI: Can we also keep the previous comments for problematic hosts?

jeremyn: I don't think we should. The old comments are not even accurate, for example https://campaign.uwaterloo.ca isn't `refused`, and there are also over a thousand other subdomains we're not mentioning.

J0WI: Since they were in the ruleset before, I would not remove them without any comment. Of course the reason in comments can be updated.

jeremyn: I've updated the top comment.

PR #13851, opened 2017-12-03
Split Landal.* from Wyndham.xml

cschanaj:


src/chrome/content/rules/Wyndham.xml
```diff
@@ -126,26 +123,13 @@ Fetch error: http://www.wyndhamrentals.com/ => https://www.wyndhamrentals.com/:
...
 
 	<rule from="^http://www\.landacampings\.(be|nl)/(css|favicons|img)/"
 		to="https://www.landalcampings.$1/$2/"/>
 
-	<rule from="^http://secure\.landal\.com/"
-		to="https://secure.landal.com/"/>
-
 	<!--	redirects as so	-->
 	<rule from="^http://(?:www\.)?landalgreenparks\.com/"
 		to="https://www.landal.com/"/>
```
J0WI: What about that one?

cschanaj: I'm sorry. Can you be more specific?

J0WI: The `landalgreenparks.com` hosts are redirected to `landal.com`. If you plan to keep this rule it should be in the `Landal.xml` file.

cschanaj:
@J0WI I have moved `landalgreenparks.com` to `Landal.xml`. Thanks!!

PR #13855, opened 2017-12-03
Add more subdomains for Serving-sys.com

numismatika:


src/chrome/content/rules/Serving-sys.com.xml
```diff
@@ -1,70 +1,109 @@
...
-	<!--securecookie host="^bs\.serving-sys\.com$" name="(?:S_\d+|r1)$" /-->
+	<target host="secure-msntest.serving-sys.com" />
+	<target host="services.serving-sys.com" />
 
 	<securecookie host="^bs\.serving-sys\.com$" name=".+" />
 
 
-	<rule from="^http://activity\.serving-sys\.com/"
+	<!--<rule from="^http://activity\.serving-sys\.com/"
 		to="https://bs.serving-sys.com/" />
```
J0WI: Do we have any issues with this redirect?

numismatika: I commented it out to see if the url test works with a simple http to https redirect without changing the subdomain and left it this way because the check was happy.

J0WI: But there is no redirect for `activity.serving-sys.com` because you also removed the target.

numismatika: indeed, it is back now

PR #13857, opened 2017-12-03
Split Potager.org.xml into more manageable rulesets

Bisaloo:
@potager, could you fix the mismatch issues that occur when you are just redirecting to another domain?

In that case, we could bring back the wildcards. You may also want to consider [HSTS preloading](https://scotthelme.co.uk/hsts-preloading/).

### DIFF
--- a/src/chrome/content/rules/Pimienta.org.xml
+++ b/src/chrome/content/rules/Pimienta.org.xml
@@ -1,0 +1,20 @@
+<ruleset name="Pimienta.org">
+
+	<target host="pimienta.org" />
+	<target host="www.pimienta.org" />
+	<target host="alcachofa.pimienta.org" />
+	<target host="assembleasocialpoblenou.pimienta.org" />
+	<target host="bibliotecadelaevasion.pimienta.org" />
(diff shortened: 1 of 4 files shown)


src/chrome/content/rules/Pimienta.org.xml
```diff
@@ -0,0 +1,20 @@
...
+	<target host="frustros.pimienta.org"/>
+	<target host="gofistfoundation.pimienta.org" /><!-- NSFW -->
+	<target host="mambo.pimienta.org" />
+	<target host="oreja.pimienta.org" />
+	<target host="quematumovil.pimienta.org" />
+	<target host="terracremada.pimienta.org" />
+
+	<rule from="^http:" to="https:" />
+
+<ruleset>
```
pipboy96: `</ruleset>`

J0WI: :ping_pong: @Bisaloo

J0WI:
Some fetch tests are now failing, but syntax LGTM now.

PR #13862, opened 2017-12-04
[ebay.nl] New ruleset

youdly:

### DIFF
--- a/src/chrome/content/rules/ebay.nl.xml
+++ b/src/chrome/content/rules/ebay.nl.xml
@@ -1,0 +1,205 @@
+<!--
+
+	For other eBay coverage, see ebay.com.xml
+	
+	
+	Non-functional subdomain:
+
+		- ebay.nl	(h)
+		- affiliates	(m)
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ebay.nl.xml
```diff
@@ -0,0 +1,205 @@
...
+	<target host="viv.ebay.nl" />
+	<target host="wantitnow.ebay.nl" />
+
+	<!-- fix http redirect -->
+	<rule from="^http://ebay\.nl/"
+		to="https://www.ebay.nl/" />
+
+	<rule from="^http:"
+		to="https:" />
+</ruleset>
```
J0WI: I miss some hosts here:
```
ocsnext.ebay.nl
pages.ebay.nl

youdly: what do you mean? they're non-functional?

J0WI: Sorry, I just found them in your commit.

PR #13840, opened 2017-12-02
Create CookMinute.com.xml

Bisaloo:
@jeremyn

Issue #13830, opened 2017-12-02
Release 2017-12-6 forthcoming


Hainish:
Type: other

cc @jsha @pde @J0WI @diracdeltas @jeremyn @gloomy-ghost

J0WI:
#13721 removes some obsolete icons

J0WI:
And #13515 might be worth to look at.

jeremyn:
https://github.com/EFForg/https-everywhere/pull/13425 would be nice to have.

Hainish:
https://github.com/EFForg/https-everywhere/issues/13876 seems pretty serious (enough for delaying the release, if necessary), looking into this now

Hainish:
Release made

PR #13864, opened 2017-12-04
Create utils/normalize-securecookie.py

cschanaj:
@Hainish This fixes an issue with Travis discussed in #13840. I have rewritten the utility in python because we no longer have `securecookie` which is not normalized since #12493. It suffices to check that no such rulesets are created in newly opened PRs. This will also make `rules-test` faster as per https://github.com/EFForg/https-everywhere/pull/12493#issue-256083902. 

Any chance before the next release #13830 ?


Hainish:
@cschanaj I'm confused - why rewrite the utility in Python rather than fix the bug within `normalize-securecookie.js`?

Hainish:
Also, why not in this PR remove `normalize-securecookie.js`?

cschanaj:
@Hainish `normalize-securecookie.js` use complex RegExp to match wildcard securecookies. To resolve the bug in #13840, it needs to capture the spaces and newlines as well. IMHO, the resultant RegExp is too complicated for effective maintenance. It is more efficient and preferable to parse the XML ruleset and check the securecookie tags.

Besides, we do not have NodeJS built into the docker image, `rules-test` is slow because we need to download the image each run. I do not want to make the `rules-test` much slower by downloading additional dependencies to parse the XMLs.

P.S. I have remove the `normalize-securecookie.js` in c2fc2b7. Thanks!

Hainish:
Thank you!

PR #13865, opened 2017-12-04
Create Vim.org.xml

Bisaloo:
reopen of https://github.com/EFForg/https-everywhere/pull/8868

### DIFF
--- a/src/chrome/content/rules/Vim.org.xml
+++ b/src/chrome/content/rules/Vim.org.xml
@@ -1,0 +1,19 @@
+<!--
+	http://www.vim.org/ advises to visit https://vim.sourceforge.io/
+	for the HTTPS version of this page.
+
+	Refused:
+		www.vim.org
+
+	Invalid certificate:
+		www.[lang].vim.org
+
+-->
+<ruleset name="Vim.org">
+
+	<target host="www.vim.org" />
+
+	<rule from="^http://www\.vim\.org/"
+		to="https://vim.sourceforge.io/" />
+
+</ruleset>
--- a/src/chrome/content/rules/Vim.org.xml
+++ b/src/chrome/content/rules/Vim.org.xml
@@ -1,4 +1,4 @@
-<ruleset name="Vim.org">
+<ruleset name="Vim.org (partial)">
 
 	<target host="www.vim.org" />
 
--- a/src/chrome/content/rules/Vim.org.xml
+++ b/src/chrome/content/rules/Vim.org.xml
@@ -1,19 +1,8 @@
-<!--
-	http://www.vim.org/ advises to visit https://vim.sourceforge.io/
-	for the HTTPS version of this page.
-
-	Refused:
-		www.vim.org
-
-	Invalid certificate:
-		www.[lang].vim.org
-
--->
 <ruleset name="Vim.org">
 
 	<target host="www.vim.org" />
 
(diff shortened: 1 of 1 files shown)


jeremyn:
I disagree on principle with having a ruleset whose entire purpose is to redirect to another site.

Bisaloo:
@jeremyn, did you try visiting http://www.vim.org. On the top of the page, they link to https://vim.sourceforge.io/. We are simply following their instructions.

I think this is a very different scenario than other "complex" redirects that we have.

jeremyn:
Yes, I see the message at http://www.vim.org. I still think this is not something we should be doing as a general rule. Maybe http://www.vim.org was hacked to say that the new site is https://vim.sourceforge.io (I doubt it, but still). Maybe the vim.org owners will want to change to a different provider besides SourceForge, or they figure out how to get HTTPS at the vim.org domain. Maybe there are a few files hidden on the site that aren't on SourceForge yet.

Pretty much the worst attack possible through HTTPS Everywhere is redirecting http://good.example.com to https://evil.example.org. I don't like having a ruleset where (the non-evil version of) this redirect is all we are doing.

jeremyn:
Also, vim.org doesn't redirect itself to https://vim.sourceforge.io. The site owners could signal a permanent shift with a `301 Moved Permanently`. I don't know if that would convince me on its own, but it would be a good sign.

Bisaloo:
@jeremyn, I understand your concerns and I agree they are valid in the general case. I also agree that we should avoid redirects to another domain as much as possible. 

On the other hand, [use of "smart" rules has always been one of HTTPS Everywhere major selling point](https://www.eff.org/https-everywhere/faq#why-use-a-whitelist-of-sites-that-support-https-why-cant-you-try-to-use-https-for-every-last-site-and-only-fall-back-to-http-if-it-isnt-available). If we completely stop using rules other than `rule from="^http:" to="https:"`, I am not sure why all of us are spending so much time working on these rulesets because we only do slightly better than simple extensions like [Smart HTTPS](https://addons.mozilla.org/en-US/firefox/addon/smart-https-revived/).

Sure, we have to draw the line somewhere but this specific PR is one case where it is okay in my opinion.

jeremyn:
The rewrite proposed in this PR is more like a CDN redirect, like `images.example.com` -> `something.cloudfront.net`, which is uncommon and discouraged.

More common, and better, uses of non-trivial `rule`s are redirects from one subdomain to another (example: `^` -> `www`), or narrowing down the scope of a wildcard (example: `from="host\d+\.example\.com"`), or limiting coverage to subsections of a domain (example: only `/politics`; only `*.jpg`).

Rewrites like the one in this PR are dangerous and sort of hostile to the site owner. HTTPS Everywhere can get embedded in places by downstream redistributors that can be hard to remove, like Debian packaging. We might lock this site into SourceForge for longer than it wants to be there.

Also an HTTPS Everywhere ruleset might lend some authority to the relationship between two sites that shouldn't be there. If we redirect `example.com` to `example.net` then someone might think these sites are related when maybe we made a mistake, or the relationship was temporary, or something.

Bisaloo:
https://www.vim.org now works fine with HTTPS. I updated this PR accordingly.

jeremyn:
There are several other `vim.org` subdomains, but you can just mark this `(partial)` if you want.

jeremyn:
Thanks, merged.

PR #13867, opened 2017-12-04
 Remove utils/merge-rulesets.sh 

cschanaj:
ping @Hainish #14336


Hainish:
I like it.  Nice and clean, removes cruft!

PR #13875, opened 2017-12-04
[FiveThirtyEight.com] Fix redirect loop (#13873)

### LINKED ISSUE #13873: Broken redirect for fivethirtyeight.com
Type: ruleset issue
Domain: fivethirtyeight.com

https://fivethirtyeight.com/ redirects to http://fivethirtyeight.com which causes a broken redirect loop.

Bisaloo:
fix #13873

### DIFF
--- a/src/chrome/content/rules/FiveThirtyEight.com.xml
+++ b/src/chrome/content/rules/FiveThirtyEight.com.xml
@@ -1,4 +1,6 @@
 <!--
+	This domain has a wildcard DNS.
+
 	Redirect to http:
 		fivethirtyeight.com
 		www.fivethirtyeight.com
--- a/src/chrome/content/rules/FiveThirtyEight.com.xml
+++ b/src/chrome/content/rules/FiveThirtyEight.com.xml
@@ -1,8 +1,4 @@
-<!--
-	This domain has a wildcard DNS.
-
--->
-<ruleset name="FiveThirtyEight.com (partial)">
+<ruleset name="FiveThirtyEight.com">
 
 	<target host="fivethirtyeight.com" />
 		<!-- Redirect to http -->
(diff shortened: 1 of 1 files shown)


jeremyn:
It looks like the redirect loop only happens for the root URL of `^`. For example, https://fivethirtyeight.com/politics/ seems fine.

Also, https://www.fivethirtyeight.com seems to redirect successfully to https://fivethirtyeight.com .

So, the only change needed is to add an `exclusion` for the root URL for `^`, and an extra `test`, like this:

```xml
	<target host="fivethirtyeight.com" />
		<exclusion pattern="http://fivethirtyeight\.com/$" />
		<test url="http://fivethirtyeight.com/politics/" />
	<target host="www.fivethirtyeight.com" />
	<target host="projects.fivethirtyeight.com" />
```

jeremyn:
We also don't need the wildcard DNS comment, or at least I'm not sure why we need it.

Bisaloo:
> We also don't need the wildcard DNS comment, or at least I'm not sure why we need it.

I am not entirely sure. J0WI usually asks for this comment. I think the main point is to warn that `Sublist3r` should not be trusted for this domain.

jeremyn:
I wouldn't add either the wildcard DNS comment or `(partial)` to the `ruleset` `name`. Since we didn't start with those, can you please remove them? Otherwise this looks fine to me.

jeremyn:
@Bisaloo Thanks, merged.

@J0WI FYI the above discussion since you were mentioned but not tagged.

J0WI:
I tend to add such comments to avoid a infinite list of (non) working targets that makes no sense. (e.g. common typos like `ww.`, `wwww.`etc.)

jeremyn:
@J0WI That makes sense, but the comment "Wildcard DNS" is not clear in itself, like if you took a random person and had them look at a random ruleset with that in it, if they would know what it means for the ruleset.

I personally favor web/Google searches over Sublist3r, because Sublist3r finds a lot of these nonsense domains like you said.

Issue #13212, opened 2017-10-18
Document and unit test API's used by downstream projects


cowlicks:
Type: other

To do this we need:
- [ ] list of downstream projects
- [ ] list of api's used by those projects

Currently I know of:

Downstream projects:
* rewriter.js

apis:
* `RuleSets`, `RuleSets.addFromXml`, and `RuleSets.rewriteURI`

jeremyn:
See also issue https://github.com/EFForg/https-everywhere/issues/6937.

koops76:
@cowlicks CloudFlare HTTPS rewrites and Brave.

PR #13877, opened 2017-12-05
Replace utils/validate.sh with utils/validate.py, Fix #1097

### LINKED ISSUE #1097: Echo line numbers when ruleset coverage checker fails
If libxml allows us to get at the line number for a given tag.

cschanaj:
This avoid wrapping test script unnecessarily ... hopefully made it easier to document all test in #13212

I guess that we shall remove `libxml2-utils` from `https-everywhere-docker-base` as well.

P.S. Fix #1097 :smile:


cschanaj:
ping @Hainish would you mind to take a look?

PR #13886, opened 2017-12-05
Watson.ch:

numismatika:
- add more subdomains

### DIFF
--- a/src/chrome/content/rules/watson.ch.xml
+++ b/src/chrome/content/rules/watson.ch.xml
@@ -1,6 +1,44 @@
+<!--
+	Cert mismatch:
+		- won.chww.watson.ch
+		- rts.chwww.watson.ch
+		- www.sorrynopenadanixzuberichten.watson.ch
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/watson.ch.xml
```diff
@@ -1,6 +1,44 @@
...
+	<target host="w.watson.ch"/>
+	<target host="w10ziy8aft.watson.ch"/>
+	<target host="ww.watson.ch"/>
+	<target host="wwww.watson.ch"/>
+	<target host="yakbs46r0p.watson.ch"/>
+
 	<rule from="^http:" to="https:" />
 </ruleset>
+
+
```
J0WI: They use a wildcard DNS and cert, so listing subdomains makes not really sense here.

numismatika: not all of them , i will recheck.
CDN  domains are used though.

PR #13906, opened 2017-12-07
Turn.com:

numismatika:
- re-enable rule


src/chrome/content/rules/Turn.com.xml
```diff
@@ -1,97 +1,81 @@
...
-			- t3.trackalyzer.com ²
-
-	¹ Unsecurable <= 404
-	² Secured by us
+	Cert expired:
+		- autodiscover.turn.com
+		- d-audienceiq-atl1.turn.com
+		- webmail.turn.com
+
+	Chain issues:
```
J0WI: It seems that some of them have been fixed.

J0WI:
@numismatika please don't overwrite previous commits.

src/chrome/content/rules/Turn.com.xml
J0WI: ping @numismatika

numismatika: Should be correct now

PR #13908, opened 2017-12-07
BR:

numismatika:
- add ticket domain

### DIFF
--- a/src/chrome/content/rules/BR.xml
+++ b/src/chrome/content/rules/BR.xml
@@ -5,7 +5,20 @@
 	<target host="story.br.de"/>
 	<target host="www.br.de"/>
 	<target host="www.brreisen.de"/>
+	<target host="br-ticket.de" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/BR.xml
```diff
@@ -5,6 +5,8 @@
 	<target host="story.br.de"/>
 	<target host="www.br.de"/>
 	<target host="www.brreisen.de"/>
+	<target host="br-ticket.de" />
+	<target host="www.br-ticket.de" />
```
J0WI: Error on `(www.\)?shop.br-ticket\.de`

J0WI: ping @numismatika

numismatika: indeed they messed up the correct chain there, will do the redirect to the ticket company myself then

PR #13909, opened 2017-12-07
+ search.stinpriza.org

xinomilo:
add search.stinpriza.org (searx public instance)
https://github.com/asciimoo/searx/issues/1072


J0WI:
Thanks for adding this rule!
We use to add rules for the whole `*.stinpriza.org` domain. So could you please add the remaining hosts? Our [docs](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains) contain some useful tips.

src/chrome/content/rules/stinpriza.org.xml
```diff
@@ -0,0 +1,18 @@
+<ruleset name="*.stinpriza.org">
```
J0WI: You can just name it `stinpriza.org`, otherwise it will look a bit odd.

Please add a comment on the top for non-working domains:
```
Mismatching cert:
ecoefficiency.stinpriza.org

Self signed:
intefix.stinpriza.org
eepf.intefix.stinpriza.org

xinomilo: you're right, sorry for the fuzz. 
i will come up with a complete ruleset, after doing a bit of cleanup on unused subdomains.

J0WI:
Thanks!

PR #13911, opened 2017-12-08
[Sci-hub] Update TLDs

Bisaloo:


Balgden:
There is also dabamirror.sci-hub.XX. I'd provide a test link, but the only one I can find right now is ~300MB.

Bisaloo:
sci-hub.mn and sci-hub.name don't seem to have a DNS record at the moment

Balgden:
Both domains load fine for me. This checker agrees as well:
http://viewdns.info/propagation/?domain=sci-hub.mn

Bisaloo:
Could you add them in another PR once this one is merged? I can't test them.

Balgden:
Maybe try again tomorrow or so, in case the DNS got updated for you by then? If it still doesn't work for you by the time this gets merged, I'll submit a follow up PR.

Bisaloo:
Actually, I just tested with Tor. It looks like all the new domains have a broken chain cert at the moment.

Balgden:
For .la, .tv, and .tw:
A Cloudflare certificate gets served for me. No chain problem.

For .hk, .mn, and .name:
My Tor Browser agrees with you, my main Firefox does not. Not entirely sure why but maybe because my ca-certificates-mozilla package got updated just last week to 3.34.1.

Btw, maybe rename the ruleset "Sci-hub" -> "Sci-Hub"?

Bisaloo:
https://whatsmychaincert.com/?www.sci-hub.tv

You can use this site to check if the certificate is serving the correct chain.

Balgden:
Ah, you tested www.sci-hub.tv. I tested plain sci-hub.tv.
https://whatsmychaincert.com/?sci-hub.tv

src/chrome/content/rules/Sci-Hub.xml
```diff
@@ -0,0 +1,36 @@
+<!--
+	Broken chain:
+		sci-hub.name
+		*.sci-hub.name
+		sci-hub.hk
+		*.sci-hub.hk
+		sci-hub.mn
+		*.sci-hub.mn
+		*.sci-hub.tv
+		*.sci-hub.tw
```
J0WI: tw and tv seems to be fixed.

Bisaloo: I already added `^` for these 2 domains. `www` is still misconfigured though.

Bisaloo:
The `moscow` host is still down but I think it's important to include the new targets in the next release so I have removed `moscow.sci-hub.*` for now and will had them later if that's okay with you.

PR #13934, opened 2017-12-11
[benj.cloud] New Ruleset

Apply55gx:


Apply55gx:
For some reason the CI didn't kick in....
Is there something wrong with my commit?

jeremyn:
I'm not sure what happened with Travis the first time, but it seemed to work after opening/closing the PR.

Apply55gx:
OK, weird...

Apply55gx:
Can this be merged or is there still something missing?

jeremyn:
This isn't a great site to include in HTTPS Everywhere because the entire real functionality of the site is behind a login, so I can't confirm that HTTPS Everywhere doesn't break the site. There isn't even an introduction page, so I'm not sure what the site does, which makes it hard to guess whether the site will work behind the login. There's also a strange "You are accessing the server from an untrusted domain." `400` response at https://www.benj.cloud which, again, makes it unclear whether I can trust the site to work behind the login.

There are also some minor stylistic problems, such as the filename starting with a lower-case letter, but you've deleted your PR repository so I don't think you can update the PR for these changes.

Probably the best thing to do here, if you are in charge of the site or can contact the owner, is to prepare the site for HSTS preloading and then get the site added to the list at https://hstspreload.org .

Apply55gx:
OK, thank you for your explanation. 

The 400 response though only happens when accessing the site through https://www.benj.cloud (doesn't happens with https://benj.cloud). 

But I unterstand your point that you only wanna merge changes where you can be sure that they work.

Gonna try now to make it hsts preload compatible, as you suggested. :)

jeremyn:
Thanks for following up.

HSTS preloading is really a better solution if you can make the site support it. We recommend HSTS preloading over extensions in the FAQ: https://www.eff.org/https-everywhere/faq#how-do-i-add-my-own-site-to-https-everywhere .

PR #13940, opened 2017-12-11
Added ruleset for in-form.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/in-form.de.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Not found:
```
J0WI: They are not in any public DNS zone, so they can be removed.

PR #13954, opened 2017-12-12
Updated avm.de ruleset.

Jochen-A-Fuerbacher:


src/chrome/content/rules/Avm.de.xml
```diff
@@ -1,68 +1,96 @@
...
+		wizzard.avm.de
+
+	Content mismatch:
+		blog.avm.de
+
+	Invalid certificate:
+		cmrelay.avm.de
+		shop.avm.de
+
+	Not found:
```
J0WI: Hosts without DNS record can be removed.

PR #13955, opened 2017-12-12
Added duden.de ruleset.

Jochen-A-Fuerbacher:


src/chrome/content/rules/Duden.de.xml
```diff
@@ -0,0 +1,26 @@
+<!--
+	Not found:
```
J0WI: Nonexistent hosts can be removed.

PR #13977, opened 2017-12-15
[trvl-media.com] New ruleset

youdly:


J0WI:
This seems related too `Expedia.xml`. Could you please add the references?

youdly:
Added in https://github.com/EFForg/https-everywhere/pull/13978

PR #13978, opened 2017-12-15
[trvl-px.com] New ruleset

youdly:


J0WI:
This seems related too `Expedia.xml`. Could you please add the references?

youdly:
Added references.

src/chrome/content/rules/trvl-px.com.xml
```diff
@@ -0,0 +1,23 @@
+<!--
```
J0WI: We use to add a reference back to the 'parent' ruleset too. (Same for pending #13976 and #13977.)

youdly: Added reference to `Expedia.com.xml` which I'll rename from `Expedia.xml` in separate PR. Current working on its ccTLD.

PR #13981, opened 2017-12-16
Replace utils/validate.sh with utils/validate.py, Fix #1097

### LINKED ISSUE #1097: Echo line numbers when ruleset coverage checker fails

cschanaj:
Hopefully before #14021


utils/validate.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python2.7
```
J0WI: Would be great to have Python 3 support for new scripts.

cschanaj: it seems to me that https://github.com/EFForg/https-everywhere-docker-base does not support Python 3, @Hainish can you confirm this?

cschanaj:
@Hainish this PR increase the speed of `rules-test` and make the error message more readable. would you mind to take a look on it?

Hainish:
@cschanaj I have this on my short-list for review.  Thanks for pinging me on this.

cschanaj:
@Hainish never mind 😄 thank you!!

PR #13983, opened 2017-12-16
Create SONGLYRICS.com.xml

christianheinrichs:


src/chrome/content/rules/SONGLYRICS.com.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Bad gateway:
+		- m.songlyrics.com
+
+	MCB:
+		- http://l.yimg.com/pb/webplayer/0.9.30/js/player-min.js
+
+	Server not found:
```
J0WI: There are no DNS records for them, so you can remove the comments.

src/chrome/content/rules/SONGLYRICS.com.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Bad gateway:
+		- m.songlyrics.com
+
+	MCB:
+		- http://l.yimg.com/pb/webplayer/0.9.30/js/player-min.js
```
J0WI: Does this break any functionality?

src/chrome/content/rules/SONGLYRICS.com.xml
christianheinrichs: Done.

J0WI: Thanks!

PR #13987, opened 2017-12-16
Create Freetz.org.xml

christianheinrichs:


J0WI:
Host is currently down from here.

J0WI:
@christianheinrichs host WFM now.
Can we redirect `trac.freetz.org` to `freetz.org` or `www.freetz.org` to avoid the cert mismatch?

christianheinrichs:
@J0WI Done.

However, visiting `freetz.org` or any of its subdomains now show a `502 Bad Gateway` error.

J0WI:
It worked a few hours ago...

J0WI:
It's currently up, so let's try again.

src/chrome/content/rules/Freetz.org.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Invalid certificate:
+		- svn.freetz.org
+		- trac.freetz.org
```
J0WI: Please add `trac.freetz.org` as target host, but keep the comment.

christianheinrichs: Done.

J0WI:
Thanks!

PR #13988, opened 2017-12-16
* Certificate chain for maths has been fixed, removed comment, and ad…

zafos:
…ded www2 back to rules

* Added maths.ox.ac.uk subdomains that support https


J0WI:
This ruleset is still `default_off` because of the errors listed on top.

zafos:
Is it worth removing the urls that fail from the ruleset for now, in order to have at least a working one?

J0WI:
They need to be checked again and moved to the list of nonfunctional hosts if they still fail (and exist).

J0WI:
@zafos would you like to continue here?

zafos:
Hello,

I will pick it up at some point but not in the near future.

Thanks

PR #13996, opened 2017-12-17
Split WelcomeCottages.com from Wyndham.xml

cschanaj:
`Access Denied` for all pages, seemingly no longer in use.


J0WI:
WFM, but some MCB from Google Maps.

cschanaj:
@J0WI the site also work for me over `tor`, so it might be a region specific site. Thanks!!

PR #13997, opened 2017-12-17
add rule for Yusp

numismatika:


J0WI:
Can we split this into multiple files?

numismatika:
I can do that, 2 or 3 files?
It is the same owner though.

pipboy96:

PR #14000, opened 2017-12-17
Create Focloir.ie.xml

Folant:


jeremyn:
- [ ] Mark https://breis.focloir.ie as `Time out`

PR #14015, opened 2017-12-18
add tripbaa.com ruleset

wonderchook:
I hope I remembered everything. I reviewed a bunch of already merged rulesets after I made this to try to make sure.

### DIFF
--- a/src/chrome/content/rules/Tripbaa.com.xml
+++ b/src/chrome/content/rules/Tripbaa.com.xml
@@ -1,5 +1,6 @@
 <!--
 	Invalid certificate:
+		ec2.tripbaa.com
 		hotel.tripbaa.com
 		transport.tripbaa.com
 		tripbaa2015.tripbaa.com
--- a/src/chrome/content/rules/Tripbaa.com.xml
+++ b/src/chrome/content/rules/Tripbaa.com.xml
@@ -4,7 +4,6 @@
 		transport.tripbaa.com
 		tripbaa2015.tripbaa.com
 -->
-
 <ruleset name="Tripbaa.com">
 	<target host="tripbaa.com" />
 	<target host="www.tripbaa.com" />
--- a/src/chrome/content/rules/tripbaa.com.xml
+++ b/src/chrome/content/rules/tripbaa.com.xml
@@ -1,0 +1,24 @@
+
+<!--
+	Invalid certificate:
+		ec2.tripba..com
+		hotel.tripbaa.com
+		ns.tripbaa.com
+		transport.tripbaa.com
+	Redirect to a different Domain:
+		tripbaa2015.tripbaa.com
+-->
+
+<ruleset name="tripbaa.com">
+	<target host="tripbaa.com" />
+	<target host="www.tripbaa.com" />
+	<target host="blog.tripbaa.com" />
+		<test url="https://blog.tripbaa.com/top10-taiwan-newyear-sunrise/" />
+	<target host="cn.tripbaa.com" />
+	<target host="en.tripbaa.com" />
+	<target host="hk.tripbaa.com" />
+
+	<rule from="^http:"
+		to="https:" />
+
+</ruleset>
--- a/src/chrome/content/rules/tripbaa.com.xml
+++ b/src/chrome/content/rules/tripbaa.com.xml
@@ -1,19 +1,14 @@
-
 <!--
 	Invalid certificate:
-		ec2.tripba..com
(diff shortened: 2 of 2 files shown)


jeremyn:
- [ ] https://tripbaa2015.tripbaa.com and remove it from `Redirect to a different Domain`

Sometimes domains have multiple problems and in that case it's best to only mark it with the earliest problem in the request. The browser will encounter an invalid certificate before it is able to do any redirection, so `Invalid certificate` is what you should use. However in this case redirecting is not actually a problem, instead it is common, valid, and requires no comment.

Also:
- [ ] Remove `ec2` entirely (Note you incorrectly wrote `ec2.tripba..com` instead of `ec2.tripbaa.com`. You will want to copy/paste as much as possible when you work with rulesets.)
- [ ] Remove `ns` entirely

This is a judgment call, and it's not strictly wrong to include these, but I think `ec2` and `ns` are too "internal" to include. (`ec2` probably refers to an AWS EC2 instance, and `ns` probably means "nameserver".) Sites often have lots of these internal subdomains, which you'll find especially if you use Sublist3r. They can clutter up a ruleset. Also maybe it is a little inappropriate to document a site's internal structure, even if you are using public information to do it.

If a contributor includes a couple of these sorts of domains, especially semi-public domains like an organizational webmail portal, I generally let it pass. If it's a lot of these domains, or they are obscure (like a nameserver), I'll probably ask them to remove those domains.

- [ ] Remove the `test` for `blog`

`target`s come with implicit `test`s for the main/root URL, in this case `http://blog.tripbaa.com` . We only need to add an explicit `test` in exceptional cases, such as if the main URL has some problem or behavior that is not typical for the domain.

Also, your `test` `url` incorrectly has `https`, not `http`. This is why the Travis `rules` step failed.

- [ ] Change the `ruleset` `name` to `Tripbaa.com`
- [ ] Change the filename to `Tripbaa.com.xml`

What you have isn't wrong but I like capitalizing this way as a matter of style. Most ruleset filenames have a leading capital letter so it's good for new ruleset filenames to have one too. The `ruleset` `name` should, if possible, both match the filename and also closely resemble the domain. Internal capitalization for the `ruleset` `name` should match the way the site itself capitalizes it, if you can easily determine that, or otherwise you can use title/grammatically-appropriate casing (without spaces) if the domain is made of recognizable words. This site clearly prefers "Tripbaa" so `Tripbaa.com` is appropriate but `TripBaa.com`, for example, would be incorrect. `tripbaa.com` would be fine except that we want to match the filename and filenames should have a leading capital letter.

- [ ] Remove the blank lines surrounding the top comment
- [ ] Add a newline at the end of the file (I would not normally ask someone to do this, but I'm including it for completeness and because you have other changes to make too)

jeremyn:
Further note: I prefer to use checklists to organize things when I ask a contributor to make changes. That way it's clear what's blocking the merge, and it's easier to review after the changes are made.

I also like to assign a pull request to myself once I've asked a contributor to make specific changes. I feel like if I'm asking a contributor to do a specific thing, I owe it to them to review the changes afterwards, and self-assigning makes that obligation explicit and helps me not lose the pull request.

wonderchook:
@jeremyn thanks for all the detailed instructions and information. I think it is now good to go.

jeremyn:
@wonderchook Please don't check off items in the checklist yourself or otherwise edit other people's comments.

I've unchecked all the boxes and then rechecked as appropriate through https://github.com/EFForg/https-everywhere/pull/14015/commits/672eda045505a7efc17076de32338d1d5bfc6f7b.

jeremyn:
I've also re-added `ec2` back to the list after I asked you to remove it before. It looks there are actually many results for `ec2` in Google, and the page content for `ec2` doesn't match the equivalent `^`/`www` URL. For example http://ec2.tripbaa.com/theme_index.php and https://www.tripbaa.com/theme_index.php are not the same. So we should include `ec2` even though it is/might be "internal" as I discussed before. I'm sorry for the confusion.

wonderchook:
okay I hope this is good to go. Funny thing on the renaming the file, at least in my workflow on Windows it would recognize capitalization in the file name. I ended up renaming through the Github web interface.

jeremyn:
Thanks, merged.

WSL keeps the Linux file metadata "alongside" the actual files (somehow), and it wants you to only use the Linux tools to work with the files. If you changed the filename using the regular Windows file explorer, you may have confused it. Otherwise I'm not sure could have happened.

PR #14010, opened 2017-12-18
Create Snopes.com.xml, Fix #14001

cschanaj:

PR #14016, opened 2017-12-19
Create utils/check-ruleset-path.sh

cschanaj:
Add a simple script to detect the inclusion of ruleset files outside of the rules directory. Relate #14010 https://github.com/EFForg/https-everywhere/pull/13844#issuecomment-348794745

ping @jeremyn @Hainish cc @J0WI @gloomy-ghost

### DIFF
--- a/utils/check-ruleset-path.sh
+++ b/utils/check-ruleset-path.sh
@@ -32,7 +32,7 @@ if [ -z "$COMMON_BASE_COMMIT" ]; then
   COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
 fi
 
-CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER`
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER | grep '.xml'`
 
 EXIT_CODE=0
 
--- a/utils/check-ruleset-path.sh
+++ b/utils/check-ruleset-path.sh
@@ -28,7 +28,10 @@ else
   git fetch upstream-for-travis master
 fi
 
-COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
+if [ -z "$COMMON_BASE_COMMIT" ]; then
+  COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
+fi
+
 CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER`
 
 EXIT_CODE=0
--- a/utils/check-ruleset-path.sh
+++ b/utils/check-ruleset-path.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER | grep '.xml'`
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESETFOLDER | grep '.xml'`
 
 EXIT_CODE=0
 
@@ -11,7 +11,7 @@ if [ "$CHANGED_FILES" != "" ]; then
 
     # if file matched the RegExp
     if [ $? -eq 0 ]; then
-      echo >&2 "ERROR: $FILE Inclusion of ruleset outside of $RULESET_FOLDER"
+      echo >&2 "ERROR: $FILE Inclusion of ruleset outside of $RULESETFOLDER"
       EXIT_CODE=1
     fi
   done <<< "$CHANGED_FILES"
--- a/utils/check-ruleset-path.sh
+++ b/utils/check-ruleset-path.sh
@@ -17,8 +17,10 @@ else
 fi 
 
 # Fetch the current GitHub version of HTTPS-E to compare to its master
-git remote add upstream-for-travis https://github.com/EFForg/https-everywhere.git
-trap 'git remote remove upstream-for-travis' EXIT
+if [ git ls-remote upstream-for-travis ]; then
+  git remote add upstream-for-travis https://github.com/EFForg/https-everywhere.git
+  trap 'git remote remove upstream-for-travis' EXIT
+fi
 
 if [ "$TRAVIS" ]; then
   git fetch --depth=50 upstream-for-travis master
--- a/utils/check-ruleset-path.sh
+++ b/utils/check-ruleset-path.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+if [ -z "$RULESET_FOLDER" ]; then
+  RULESET_FOLDER="src/chrome/content/rules"
+fi
+
+# Go to git repo root; note that $GIT_DIR is .git in this case.
+if [ -n "$GIT_DIR" ]; then
+  # GIT_DIR is set, so we're running as a hook.
+  cd "$GIT_DIR" && cd ..
+elif [ ! `which git > /dev/null` ]; then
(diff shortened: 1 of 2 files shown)


jeremyn:
@cschanaj Thank you for looking at this!

I don't like hardcoding the number of non-ruleset XML files. That will surprise people in the future.

How about either:

* adding a extra check that says "error if any of the XML files outside of `src/chrome/content/rules` resemble ruleset files", either with a proper XML parser (ideal, sort of the inverse of what `validate.sh` does) or with some crude regex (`^<ruleset`); or

* do something similar but move the check into the logic flow in `validate.sh`

Also, isn't there some part of the testing that identifies what files have changed based on the git diff? If so, we can just check whether the files that changed are rulesets and are not in `src/chrome/content/rules`.

Another note, only vaguely related: I think we should start "variable-izing" the location of the ruleset directory. It's annoying that some part of the code uses the `rules/` symlink because this is actively hostile to Windows. Probably the right way to do this is to use a environment variable for this directory. Anyway we don't need to refactor that here but it would be nice to not add more places where the ruleset directory is hardcoded.

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,29 @@
...
+    # $GIT_DIR is set, so we're running as a hook.
+    cd $GIT_DIR
+else
+    # Git command exists? Cool, let's CD to the right place.
+    git rev-parse && cd `git rev-parse --show-toplevel`
+fi
+
+XML_COUNT=`find . -path "./src/chrome/content/rules" -prune -o -name "*.xml" -print | wc -l`
+
+if [ "$XML_COUNT" -ne 2 ]; then
```
Hainish: This seems rather brittle, and (in my working directory) currently innacurate:

```shell
user@https-everywhere ~/workspace/https-everywhere (master) $ find . -path "./src/chrome/content/rules" -prune -o -name "*.xml" -print
./utils/relaxng.xml
./chromium/updates-master.xml
./chromium/updates.xml
```

We should generalize this somehow, either by taking @jeremyn's suggestion in https://github.com/EFForg/https-everywhere/pull/14016#issuecomment-352792000 or by some other method.

It would probably suit us to add some additional labels or re-purpose existing ones as well, and auto-label PRs based on which directories they've placed files in.  It would make detection of problems like we saw in  https://github.com/EFForg/https-everywhere/pull/14010 easier to detect.  For instance: `ruleset`, `utils`, `tests`, `non-ruleset` etc.

jeremyn: There's no `chromium/updates.xml` , see https://github.com/EFForg/https-everywhere/blob/master/chromium/updates.xml (404).

Hainish: @jeremyn It is included in https://github.com/EFForg/https-everywhere/blob/master/chromium/.gitignore and generated by https://github.com/EFForg/https-everywhere/blob/fe9f98c88688752cb88bbb328e29e7866683d743/make.sh#L46.  Therefore this test will fail for anyone running it after `./make.sh`

cschanaj:
@Hainish Travis is passing now. Could you please review this? thanks.

J0WI:
Tagging for the next release.

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,49 @@
...
+
+COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER`
+
+EXIT_CODE=0
+
+if [ "$CHANGED_FILES" != "" ]; then
+  while read FILE; do
+    # check if changed file is actually a ruleset
+    egrep -q "^<ruleset[^>]+>" "$FILE"
```
Hainish: One additional condition we may want to add is to ensure that `$FILE` has an `.xml` filetype.  I can see the `<ruleset>` tag being used in documentation.

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+if [ "$TRAVIS" ]; then
+  git fetch --depth=50 upstream-for-travis master
+else
+  git fetch upstream-for-travis master
+fi
+
+COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER`
```
Hainish: This doesn't seem very dry, since in most cases we've done the work of the previous lines in https://github.com/EFForg/https-everywhere/blob/02e07e3e90878b47b3b44f7494742496f995d638/test/travis.sh#L7-L34

We may want to instead save the changed files to an env var, and pass them to this script as an argument, kind of like what we do with `RULESETS_CHANGED` in https://github.com/EFForg/https-everywhere/blob/02e07e3e90878b47b3b44f7494742496f995d638/test/travis.sh#L51

cschanaj:
@Hainish done. Please have a look on this again. thanks!!

Hainish:
@cschanaj reviewing now

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [ "$TRAVIS" ]; then
+  git fetch --depth=50 upstream-for-travis master
+else
+  git fetch upstream-for-travis master
+fi
+
+if [ -z "$COMMON_BASE_COMMIT" ]; then
+  COMMON_BASE_COMMIT=`git merge-base upstream-for-travis/master HEAD`
+fi
```
Hainish: Maybe I wasn't so clear in the previous comment.  I mean that I think we can get rid of most of the lines above, since the work of cloning the repo and finding what files have been changed has already been done in https://github.com/EFForg/https-everywhere/blob/02e07e3e90878b47b3b44f7494742496f995d638/test/travis.sh#L7-L34.  A lot of the work above is a copy of what we already have there.

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,52 @@
...
+
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER | grep '.xml'`
```
Hainish: :+1:

utils/check-ruleset-path.sh
Hainish: So I think we can remove lines 3 through 33, because it's costly to do the `git fetch` and we've already done it in `travis.sh`

cschanaj:
@Hainish I have removed those lines. thanks!

Hainish:
@cschanaj reviewing now

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+CHANGED_FILES=`git diff --name-only $COMMON_BASE_COMMIT | grep -v $RULESET_FOLDER | grep '.xml'`
```
Hainish: Change `$RULESET_FOLDER` to `$RULESETFOLDER`.

utils/check-ruleset-path.sh
```diff
@@ -0,0 +1,20 @@
...
+EXIT_CODE=0
+
+if [ "$CHANGED_FILES" != "" ]; then
+  while read FILE; do
+    # check if changed file is actually a ruleset
+    egrep -q "^<ruleset[^>]+>" "$FILE"
+
+    # if file matched the RegExp
+    if [ $? -eq 0 ]; then
+      echo >&2 "ERROR: $FILE Inclusion of ruleset outside of $RULESET_FOLDER"
```
Hainish: Change `$RULESET_FOLDER` to `$RULESETFOLDER`.

cschanaj:
@Hainish done. thanks!!

Hainish:
Thank you!

PR #14017, opened 2017-12-19
[apply55gx.com] Add rule

Apply55gx:


J0WI:
Thanks!
You're also welcome to submit your domain to https://hstspreload.org/, which makes a rule in this Add-On obsolete :)

Apply55gx:
Thanks for merging :)
Unfortunately I can't set the required headers to preloade it, because I'm using shared hosting.

PR #14018, opened 2017-12-19
[raiffeisen.ch] Add rule

Apply55gx:


Apply55gx:
Didn't saw that there was an already existing rule...

PR #14020, opened 2017-12-19
Add narsol.org

VoidPhantom:


VoidPhantom:
There are also a number of other subdomains: `cpanel.nationalrsol.org`, `webdisk.nationalrsol.org`, `webmail.nationalrsol.org`, `cpanel.rsolconference.org`, `webdisk.rsolconference.org`, `webmail.rsolconference.org`. They seem to be accessible over HTTPS, but they seem to be some sort of non-publicly-accessible administrative interfaces, so I can’t test that everything there work well over HTTPS. What should I do with them? Should I add them to a top comment?

PR #14030, opened 2017-12-21
Update Weibo.com.xml

ShiningRay:
Add a subdomain which weibo.com uses for sharing content from 3rdparty tools/bookmarklets.

### DIFF
--- a/src/chrome/content/rules/Weibo.com.xml
+++ b/src/chrome/content/rules/Weibo.com.xml
@@ -53,7 +53,7 @@
 		<test url="http://security.weibo.com/app/js/v1/apps/account/jQuery1.7.1.js" />
 	<target host="yahoo.tw.weibo.com" />
 	<target host="widget.weibo.com" />
-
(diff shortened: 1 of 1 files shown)


jeremyn:
@ShiningRay Thanks for your pull request. Unfortunately this domain can't be simply added because it causes various problems, for example mixed content blocking at https://service.weibo.com/siteshow/siteshow.php and different HTTP/HTTPS content between http://service.weibo.com/z/guyuanrenzheng?fr=leftnlq and https://service.weibo.com/z/guyuanrenzheng?fr=leftnlq .

If you can identify subsections of this domain that can be safely rewritten, we might be able to figure out a way to include those.

Also for future reference, if we could add this domain, it should go directly below the `test` for `security.weibo.com`, and you would also need to remove it from `Invalid certificate`.

J0WI:
@ShiningRay would you like to continue here?

zoracon:
Closing for now due to inactivity, but if interest resumes we can reopen

PR #14035, opened 2017-12-22
[cimb.com] Add subdomains and rename ruleset

youdly:

PR #14049, opened 2017-12-23
[cimbbank.com.my] Add subdomains

youdly:


youdly:
`cimbbank.com.my` has been removed from the original ruleset from https://github.com/EFForg/https-everywhere/pull/14035

PR #14050, opened 2017-12-23
[cimbclicks.com.my] Add subdomains

youdly:


youdly:
`cimbclicks.com.my` has been removed from the original ruleset in https://github.com/EFForg/https-everywhere/pull/14035

PR #14052, opened 2017-12-23
[cimbpreferred.com] Move to a separate ruleset

youdly:


youdly:
`cimbpreferred.com` has been removed from the original ruleset in https://github.com/EFForg/https-everywhere/pull/14035

PR #14061, opened 2017-12-23
 Create University_of_California_San_Diego.xml

cschanaj:
I have performed a complete rewrite of the respective ruleset since the comment (dated May 2012) are likely obsolete for a long time. Besides, this ruleset has been disabled since Feb 2015, so I don't really care about the coverage loss here. 

P.S. If there is some important domains I have missed, please let me know. Thanks.


youdly:
Dashes are preferred over underscore for ruleset name. suggest to `squash` those `Update` commits.

PR #14062, opened 2017-12-23
 Simplify and Update mvd.ru.xml 

cschanaj:


src/chrome/content/rules/mvd.ru.xml
```diff
@@ -1,675 +1,12 @@
...
-	<target host="veteran.mvd.ru" />
-	<target host="vi.mvd.ru" />
-	<target host="vipk.mvd.ru" />
-	<target host="vnii.mvd.ru" />
-	<target host="vsut.mvd.ru" />
-	<target host="zbut.mvd.ru" />
+	<target host="*.mvd.ru" />
+		<test url="http://www.mvd.ru/" />
+		<test url="http://static.mvd.ru/upload/site1/05IWFr3NZU.png" />
+		<test url="http://static.mvd.ru/media/mvd-2015/css/flexlist.css" />
```
J0WI: So from >500 testable targets we end up with two hosts...😐
Furthermore, some of the certs in the comments are currently still expired or mismatch.

PR #14066, opened 2017-12-24
[Maybank.com] New ruleset

youdly:

PR #14070, opened 2017-12-24
[Maybank-KE.co.th] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/14066

### DIFF
--- a/src/chrome/content/rules/Maybank-KE.co.th.xml
+++ b/src/chrome/content/rules/Maybank-KE.co.th.xml
@@ -1,0 +1,35 @@
+<!--
+	For other Maybank coverage, see Maybank.com.xml
+
+
+	Non-functional subdomains:
+
+		- event	(t)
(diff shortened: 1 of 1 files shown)


youdly:
Rebased to resolve build issue

src/chrome/content/rules/Maybank-KE.co.th.xml
```diff
@@ -0,0 +1,35 @@
+<!--
+	For other Maybank coverage, see Maybank.com.xml
+
+
+	Non-functional subdomains:
+
+		- www.maybank-ke.co.th	(r)
```
J0WI: WFM, but has mixed content issues

youdly: Those HTTP resources are available on HTTPS in my test, but doesn't load due to MCB. Anyhow, I updated the comment.

PR #14072, opened 2017-12-24
[Maybank.co.id] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/14066

### DIFF
--- a/src/chrome/content/rules/Maybank.co.id.xml
+++ b/src/chrome/content/rules/Maybank.co.id.xml
@@ -1,0 +1,45 @@
+<!--
+	For other Maybank coverage, see Maybank.com.xml
+
+
+	Non-functional subdomains:
+
+		- mail	(m)
+		- maingw2	(t)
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Maybank.co.id.xml
```diff
@@ -0,0 +1,45 @@
+<!--
+	For other Maybank coverage, see Maybank.com.xml
+
+
+	Non-functional subdomains:
+
+		- mail	(m)
+		- maingw2	(t)
+		- www.mbiwfh	(i)
+		- microsite	(m)
```
J0WI: Can be redirected to `www.micorsite`
See also `form.maybank.co.id` (requires exclusion and custom test url)

youdly: Added redirect. All the pages I found on `form.maybank.co.id` (via google) simply timeout.

PR #14086, opened 2017-12-25
Create Pokaa.fr.xml

Bisaloo:


J0WI:
Please revert the change in the translations submodule.

Bisaloo:
@J0WI, in case you missed it, I have reverted the change as requested.

PR #14101, opened 2017-12-26
The New York Times has a new SecureDrop URL

runasand:
Updated the .xml files to indicate that The New York Times has a new SecureDrop URL. This can be verified on [https://nytimes.com/tips](https://nytimes.com/tips). The old URL is serving a static page over HTTP.

### DIFF
--- a/src/chrome/content/rules/NYTimes.xml
+++ b/src/chrome/content/rules/NYTimes.xml
@@ -6,7 +6,7 @@
 		- NYT_cn.me.xml
 		- NYT_eXaminer.com.xml
 		- INYT.com.xml
-		- nytimes2tsqtnxek.onion.xml
+		- nyttips4bmquxfzw.onion
 
 	CDN buckets:
 		- pimage.timespeople.nytimes.com.amazonaws.com
--- a/src/chrome/content/rules/NYTimes.xml
+++ b/src/chrome/content/rules/NYTimes.xml
@@ -6,7 +6,7 @@
 		- NYT_cn.me.xml
 		- NYT_eXaminer.com.xml
 		- INYT.com.xml
-		- nyttips4bmquxfzw.onion
+		- nyttips4bmquxfzw.onion.xml
 
 	CDN buckets:
 		- pimage.timespeople.nytimes.com.amazonaws.com
--- a/src/chrome/content/rules/NYTimes.xml
+++ b/src/chrome/content/rules/NYTimes.xml
@@ -45,6 +45,7 @@
 
 	Too many redirects:
 		regilite.nytimes.com
+		homedelivery.nytimes.com
 
 	Problematic hosts in *nytimesathome.com:
 		- ^ ᵐ
@@ -71,15 +72,12 @@
 		<test url="http://graphics8.nytimes.com/css/common/global.css" />
 		<test url="http://graphics8.nytimes.com/images/blogs_v5/artsbeat/artsbeat_main.png" />
 		<test url="http://graphics8.nytimes.com/images/misc/nytlogo153x23.gif" />
-	<target host="homedelivery.nytimes.com" />
 	<target host="international.nytimes.com" />
 	<target host="meter-svc.nytimes.com" />
 	<target host="myaccount.nytimes.com" />
 	<target host="markets.on.nytimes.com" />
 		<test url="http://markets.on.nytimes.com/research/Net/SectionFrontAPI/MarketModuleData/jsonp" />
 	<target host="static.nytimes.com" />
-	<target host="store-media.nytimes.com" />
-	<target host="store-skin.nytimes.com" />
 	<target host="typeface.nytimes.com" />
 	<target host="up.nytimes.com" />
 	<target host="tagx.nytimes.com" />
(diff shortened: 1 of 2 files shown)


jeremyn:
@runasand Thanks for your pull request.

When I try the old URL at http://nytimes2tsqtnxek.onion with the Tor Browser, I get an `Unable to connect` error, not a static page, and the same with the `https` version of that URL.

I see from your Twitter that you appear to be in charge of technical security at the NY Times. Is there any way you can make it so that both the `http` and `https` versions of the old URL redirect to the `https` version of the new URL? This would be ideal for users since the old URL is already out there, and in that case we would want to provide coverage for both the old and the new URLs.

runasand:
Thanks for your comments, @jeremyn.

You can leverage [tor-curl](https://github.com/nytm/nytimes-onion/blob/master/tor-curl.sh) to verify that nytimes2tsqtnxek.onion is serving a static page over HTTP. You're seeing a connection error because HTTPS Everywhere is causing the Tor Browser to default to HTTPS. The goal of this PR is to correct that behavior.

```
$ ./tor-curl.sh nytimes2tsqtnxek.onion
Looking for The New York Times? See <a href="https://nytimes.com/tips">https://nytimes.com/tips</a>.
```

I am the Senior Director of Information Security at The New York Times. I helped launch our [Tips](https://nytimes.com/tips) page a year ago and am responsible for our SecureDrop instance. We believe that the right behavior in this case is to refer users back to our Tips page for the correct URL, not have them rely on redirects.

jeremyn:
After your last comment, I was able to see the static page at the old `http` address with the Tor Browser. Thanks.

I understand if you want to redirect obsolete pages to a centralized `/tips` page instead of managing individual redirects. However it would be good if you could enable HTTPS on the old page, in which case we could keep the existing `target` for the old URL.

Also you'll need to fix the problems in the existing `NYTimes.xml` ruleset so Travis will pass. You can delete the two missing hosts `store-media` and `store-skin` by deleting [these two lines](https://github.com/EFForg/https-everywhere/blob/2f3bf2bf0a0bdcfe103662788363140bc77fa55e/src/chrome/content/rules/NYTimes.xml#L81-L82). For https://homedelivery.nytimes.com you'll need to remove the `target` [here](https://github.com/EFForg/https-everywhere/blob/2f3bf2bf0a0bdcfe103662788363140bc77fa55e/src/chrome/content/rules/NYTimes.xml#L74) and list `homedelivery.nytimes.com` directly under `Too many redirects` [here](https://github.com/EFForg/https-everywhere/blob/2f3bf2bf0a0bdcfe103662788363140bc77fa55e/src/chrome/content/rules/NYTimes.xml#L46).

runasand:
Enabling HTTPS on the old page would require us to pay DigiCert to renew the SSL certificate for the .onion name. We don't want to do that, nor do we want to serve a self-signed certificate. For those two reasons, we decided to serve the page over HTTP and refer users to our Tips page where they can find the new URL.

jeremyn:
@runasand That's okay, thanks for explaining the situation to me.

Please:
- [ ] Add `.xml` to the reference in `NYTimes.xml`

- [ ] Change `nyttips4bmquxfzw.onion.xml` to:
```xml
<!--
    For other NY Times rulesets, see NYTimes.xml

    See https://nytimes.com/tips for confirmation that this site is associated
        with The NY Times.

    Invalid certificate:
        www.nyttips4bmquxfzw.onion

-->
<ruleset name="nyttips4bmquxfzw.onion">
    <target host="nyttips4bmquxfzw.onion" />

</ruleset>
```
These are mostly small style changes, but I would like to have a link to the `/tips` URL to confirm this domain's relationship to The NY Times. We have something similar in a ProPublica `.onion` ruleset [here](https://github.com/EFForg/https-everywhere/blob/a6d338c1db742f768569c1a8c748a04cf408bcf8/src/chrome/content/rules/propub3r6espa33w.onion.xml#L4-L9). Please feel free to change the wording if you want.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/14101/commits/38711c4de8d14b74eb431e77ce9b18d2c041c59b.

`homedelivery` should go above `regilite` alphabetically in `NYTimes.xml`. I've now added that to the checklist. I hinted at that in https://github.com/EFForg/https-everywhere/pull/14101#issuecomment-354008303 but I'm sorry that I didn't catch that in https://github.com/EFForg/https-everywhere/pull/14101/commits/3917639e9139a521f9336b20387df6410e19f4fe when I first made the checklist.

We prefer tabs instead of spaces. This was incorrect in my first version in the checklist. Again, I'm sorry about that. Your current version of the `.onion.xml` file has spaces [here](https://github.com/runasand/https-everywhere/blob/38711c4de8d14b74eb431e77ce9b18d2c041c59b/src/chrome/content/rules/nyttips4bmquxfzw.onion.xml#L8).

The `ruleset` `name` in the `.onion.xml` file should in this case be the same as the domain, as in my suggested version, not the current wordy description. See https://github.com/EFForg/https-everywhere/pull/12829#issuecomment-338031036 if you'd like some justification.

jeremyn:
Thanks, merged.

PR #14104, opened 2017-12-26
Update MegaGlest.org.xml

jeremyn:
This pull request replaces pull request https://github.com/EFForg/https-everywhere/pull/13769.

### DIFF
--- a/src/chrome/content/rules/MegaGlest.org.xml
+++ b/src/chrome/content/rules/MegaGlest.org.xml
@@ -1,33 +1,27 @@
 <!--
-	Refused:
-		- faq		(→ docs)
-		- readme	(→ raw.github.com)
-		- wiki		(→ docs)
+	Invalid certificate:
+		getdeb.megaglest.org
+		snapshots.megaglest.org
+		softcoder.megaglest.org
+
+	Secure connection failed:
+		changelog.megaglest.org
+		faq.megaglest.org
+		paste.megaglest.org
+		readme.megaglest.org
+		transifex.megaglest.org
+		wiki.megaglest.org
+
 -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/MegaGlest.org.xml
```diff
@@ -1,33 +1,27 @@
...
 	<target host="www.megaglest.org" />
+	<target host="archer.megaglest.org" />
+	<target host="ci.megaglest.org" />
 	<target host="docs.megaglest.org" />
+	<target host="filux.megaglest.org" />
 	<target host="forum.megaglest.org" />
-
-	<target host="faq.megaglest.org" />
-	<target host="readme.megaglest.org" />
-	<target host="wiki.megaglest.org" />
```
J0WI: Why are you removing these redirects?

jeremyn: I wrote that in https://github.com/EFForg/https-everywhere/pull/13769#issuecomment-347211172:

>@heuser443 I get what you're doing here, but I think this is the wrong approach. These sort of deep redirects, like http://translate.megaglest.org to https://docs.megaglest.org/MG/Translations , are generally bad practice. We've also caused a specific problem for this site.
>
>Let's just remove all `target`s and `rule`s that don't support a direct rewrite. We'll lose some security, but I think that's fine here. The site's users and owners can work toward supporting HTTPS better if they want to.

J0WI: Okay, that sounds reasonable to me.

cschanaj:
this should close #13765

jeremyn:
Thanks @cschanaj , I've closed that issue.

PR #14119, opened 2017-12-27
Gramps project.org

MilanKral:


jeremyn:
- [ ] Remove the `target` and `rule` for `bugs` and note it as `Invalid certificate` at the top.

The problem with this `rule` is that we have no examples of real `bugs` URLs other than the root/main URL, so we don't know if this rewrite always works. For example maybe there is some URL like `http://bugs.gramps-project.org/example` that is not supposed to be rewritten to `https://gramps-project.org/bugs/example`.

jeremyn:
@MilanKral Thanks, but the indentation at the top is not good though, in particular you have mixed tabs and spaces. Try making it exactly like this:

```xml
<!--
	Invalid certificate:
		bugs.gramps-project-.org

-->
<ruleset ...>
```

jeremyn:
Thanks, merged.

PR #14126, opened 2017-12-28
Update University-of-Waterloo.xml

jtracey:
Added some new rules for the University of Waterloo, mostly relating to the library.

I also found a bunch of subdomains that could go in the leading comment, but I don't know what's considered sufficient cause to add them there, so I didn't. Some examples (all linked from library pages) include:

subjectguides.uwaterloo.ca
digital.library.uwaterloo.ca
ereference.uwaterloo.ca
journal-indexes.uwaterloo.ca
testtube.uwaterloo.ca

### DIFF
--- a/src/chrome/content/rules/University-of-Waterloo.xml
+++ b/src/chrome/content/rules/University-of-Waterloo.xml
@@ -7,6 +7,7 @@
 		cacr.uwaterloo.ca
 		cecs.uwaterloo.ca
 		hr.uwaterloo.ca
+		bookings.lib.uwaterloo.ca (incomplete certificate chain)
 
 -->
 <ruleset name="University of Waterloo (partial)">
@@ -19,7 +20,6 @@
 	<target host="www.cs.uwaterloo.ca" />
 	<target host="csclub.uwaterloo.ca" />
 	<target host="library.uwaterloo.ca" />
-	<target host="bookings.lib.uwaterloo.ca" />
 	<target host="www.lib.uwaterloo.ca" />
 	<target host="math.uwaterloo.ca" />
 	<target host="cemc.math.uwaterloo.ca" />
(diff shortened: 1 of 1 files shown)


jtracey:
Hmm not sure what that test failure is, https://bookings.lib.uwaterloo.ca/ looks fine from Firefox and Chrome on my end.

Also, I should probably mention: There's one more pair of pages to add. nexusmail.uwaterloo.ca and www.nexusmail.uwaterloo.ca just had a cert expire yesterday afternoon. Given that it's the webmail portal for the university it will probably be fixed soon, but since it isn't yet I didn't add it. If it's easier we can hold off on this PR until it is fixed, or just merge this now and do a separate PR for it later, I'm fine with either.

jeremyn:
https://bookings.lib.uwaterloo.ca/ serves a bad intermediate certificate chain. You can check this by loading the page in a new Firefox profile (`about:profiles`). It might work for you in your normal profile because you've previously visited a site that correctly served the necessary certificate chain and your browser has cached it.

jtracey:
Ah, thanks. I'll remove it and add to the leading comment.

jeremyn:
Please:

Mark as `Different HTTP/HTTPS content`:
- [ ] https://library.uwaterloo.ca and remove its `target`

The different content in this case is that https://library.uwaterloo.ca returns a `200` but http://library.uwaterloo.ca returns a `301` redirect to `http://www.lib.uwaterloo.ca/` . The two end pages look very similar but it is still technically a different result, and also the overall `uwaterloo.ca` environment feels -- unstable? -- enough to me (see PR https://github.com/EFForg/https-everywhere/pull/13845) that I don't trust anything unusual here.

For the other domains you mentioned in https://github.com/EFForg/https-everywhere/pull/14126#issue-284982953, we don't need to exhaustively document them here, since this is a `(partial)` ruleset anyway and `uwaterloo.ca` has so many subdomains.

jeremyn:
Thanks, merged.

PR #14129, opened 2017-12-30
Create RARLab.xml and Update WinRAR.xml

cschanaj:
Replace #14128

### DIFF
--- a/src/chrome/content/rules/RARLab.xml
+++ b/src/chrome/content/rules/RARLab.xml
@@ -1,0 +1,29 @@
+<!--
+	For other WinRAR.com related rulesets, see WinRAR.xml
+
+	Non-functional hosts
+		Couldn't connect to server:
+		- imap.rarlab.com
+		- mx1.rarlab.com
+		- ns1.rarlab.com
+		- ns2.rarlab.com
+		- pop3.rarlab.com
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/WinRAR.xml
```diff
@@ -1,16 +1,35 @@
...
-		- (www\.)?win-rar.com
+	Other WinRAR.com related rulesets:
+		+ RARLab.xml
 
-	Mismatch:
-		- mailing.win-rar.com
--->
-<ruleset name="WinRAR (partial)">
+	Non-functional hosts
+		Couldn't connect to server:
```
J0WI: There is no HTTP service running, so you can remove them.

src/chrome/content/rules/RARLab.xml
```diff
@@ -0,0 +1,29 @@
+<!--
+	For other WinRAR.com related rulesets, see WinRAR.xml
+
+	Non-functional hosts
```
J0WI: Hosts without any HTTP service running can be removed.

cschanaj:
@J0WI I have removed the hosts without HTTP service. Thanks!!

J0WI:
Thanks!

PR #14130, opened 2017-12-30
Advertising.com:

numismatika:
- add adserver-us subdomain and more

### DIFF
--- a/src/chrome/content/rules/Advertising.com.xml
+++ b/src/chrome/content/rules/Advertising.com.xml
@@ -1,80 +1,215 @@
 <!--
 	For problematic rules, see Advertising.com-mismatches.xml.
-
-
 	For other AOL coverage, see AOL.xml.
 
+		Cert expired:
+			- jp.advertising.com
 
-	Problematic subdomains:
+		Cert mismatch:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Advertising.com.xml
```diff
@@ -1,80 +1,215 @@
...
 	For problematic rules, see Advertising.com-mismatches.xml.
-
-
 	For other AOL coverage, see AOL.xml.
 
-
-	Problematic subdomains:
-
-		- ace-tag *
-		- cte.ace.beta		(mismatched, CN: servedby.advertising.com)
```
J0WI: This host does still exist, so please keep a comment about it.

numismatika:
I thought having multiple "test this" "revert the test for this" commits is ugly and in the past the commits weren't always properly squashed so it became a , probably bad, habit to always squash and rebase for each iteration.

src/chrome/content/rules/Advertising.com.xml
numismatika: cte.ace.beta is part of the -mismatched file so it is already covered there.

PR #14148, opened 2017-12-31
add rule for Aniview.com

numismatika:


src/chrome/content/rules/Aniview.com.xml
```diff
@@ -0,0 +1,43 @@
+<!--
+	Cert mismatch:
+		- d41751.aniview.com
+		- data1.aniview.com/
+		- data2.aniview.com
+
+	TLS error:
+		- aniview.com
```
J0WI: You can redirect it to www.

src/chrome/content/rules/Aniview.com.xml
```diff
@@ -0,0 +1,48 @@
...
+	<target host="urlhandler.aniview.com" />
+	<target host="wl.aniview.com" />
+	<target host="wlplay.aniview.com" />
+	<target host="vtrack.aniview.com" />
+	<target host="wlplayer.aniview.com" />
+
+	<rule from="^http://aniview\.com/"
+		to="https://www.aniview.com/" />
+
+		<test url="http://aniview.com/"/>
```
J0WI: You need to add it as target.

J0WI: Please remove this test url since it's redundant to the target.

PR #14155, opened 2018-01-01
Added an additional rule for img bato.to subdomains.

Alhashasrardi:
Was seeing requests to img4.bato.to over port 80 but wouldn't accept https. Changed it to img.bato.to which did accept and also tried the base bato.to domain which happily serves images as well.

### DIFF
--- a/src/chrome/content/rules/bato.to.xml
+++ b/src/chrome/content/rules/bato.to.xml
@@ -6,8 +6,8 @@
 	<target host="img2.bato.to"/>
 	<target host="img.bato.to"/>
 
-	<rule from="^http://img\d?\.bato\.to"
-		to="https://bato\.to"/>	
+	<rule from="^http://img\d?\.bato\.to/"
+		to="https://bato\.to/"/>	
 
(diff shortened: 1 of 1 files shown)


jeremyn:
There's an announcement on the main page https://bato.to/ that "Batoto will be closing down permanently. Jan 18, 2018". https://img2.bato.to/ , https://img3.bato.to/ , and https://img4.bato.to/ are missing from DNS. https://img.bato.to/ has an invalid certificate error and the plain HTTP version http://img.bato.to/ eventually redirects to a 403.

I think this pull request should be closed without merging. Maybe some months from now we should remove the ruleset entirely. Please let us know if you disagree. We'll close this pull request if we don't get a response from you soon.

jeremyn:
Currently, https://img.bato.to/ , https://img2.bato.to/ , https://img3.bato.to/ , and https://img4.bato.to/ are all `Refused` with `https` and require a login for `http`, so they are not good domains to include in the ruleset. I'm closing this pull request.

PR #14162, opened 2018-01-02
[abs.ch] Add rule

coffeemakr:
Add rules for Alternative Bank Switzerland (ABS)

### DIFF
--- a/src/chrome/content/rules/Abs.ch.xml
+++ b/src/chrome/content/rules/Abs.ch.xml
@@ -24,6 +24,7 @@
 	<target host="alternativebankschweiz.ch" />
 	<target host="www.alternativebankschweiz.ch" />
 	<target host="bancaalternativasvizzera.ch" />
+	<target host="www.bancaalternativasvizzera.ch" />
 
 	<rule from="^http:"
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Abs.ch.xml
```diff
@@ -0,0 +1,32 @@
...
+	<target host="wwwsec.abs.ch" />
+
+	<!-- Redirection domains -->
+	<target host="alternativebankswitzerland.ch" />
+	<target host="www.alternativebankswitzerland.ch" />
+	<target host="banquealternativesuisse.ch" />
+	<target host="www.banquealternativesuisse.ch" />
+	<target host="alternativebankschweiz.ch" />
+	<target host="www.alternativebankschweiz.ch" />
+	<target host="bancaalternativasvizzera.ch" />
```
J0WI: And last but not least `www.bancaalternativasvizzera.ch` ;)

coffeemakr: Good catch. Fixed

J0WI:
Thanks!

PR #14165, opened 2018-01-03
Added rulesets for three websites

Bischnu:
These websites have passive mixed-content. Also I failed to test with the script but there should not be any problem with the rules (also it is the first time I really use git, I am not yet at ease with the different commands).

### DIFF
--- a/src/chrome/content/rules/Idleman.fr.xml
+++ b/src/chrome/content/rules/Idleman.fr.xml
@@ -1,0 +1,20 @@
+<!--
+	Nonfunctional hosts in *.idleman.fr:
+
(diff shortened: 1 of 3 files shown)


jeremyn:
Thanks for your contribution, however it's best if you submit one ruleset per pull request instead of grouping unrelated pull requests together. Can you please close this and then submit three pull requests for what you have here?

jeremyn:
I've added the domains `idleman.fr` , `techradar.com` , and `tldp.org` to https://github.com/EFForg/https-everywhere/issues/3069#issuecomment-361998993.

PR #14169, opened 2018-01-03
Re-enable MIT.xml (continue from #12002)

J0WI:


jeremyn:
@J0WI For what it's worth, I don't agree with how you handled this. As I wrote in https://github.com/EFForg/https-everywhere/pull/12002#issuecomment-353680911, I think you should have resubmitted the old PR with your extra changes as a new PR and then waited for someone else to review it. Instead you effectively self-committed many changes to a complicated `top-1k` ruleset.

I'm not accusing you of anything or whatever, but I think we should discuss this as a group first if you plan on taking this approach to future abandoned / @ghost PRs.

cpitclaudel:
Could this PR be causing #14397?

jeremyn:
I've added a comment at https://github.com/EFForg/https-everywhere/issues/14397#issuecomment-359926293.

PR #14174, opened 2018-01-04
Create Meltdownattack.com.xml

cschanaj:
...


jeremyn:
We don't need `securecookie`s for sites like this that don't serve cookies. I merged this anyway because it's a minor point and including it isn't technically wrong anyway.

Issue #8833, opened 2017-02-26
downgrade rules need periodic inspection


ivysrono:
- [x] #8551
- [x] #8832 
- [x] #8834 
- [x] #8899 
- [x] #8906 
- [x] #9190 
- [x] #9271 
- [x] #9305 
- [x] #9311 
- [x] #9318 
- [x] #9319 
- [x] #9320 
- [x] #9334 
- [x] #9335 
- [x] #9339 
- [x] #9340 
- [x] #9349 
- [x] #9350 
- [x] #9381

J0WI:
See also #7717

ivysrono:
Apple: #8906

J0WI:
#8551

ivysrono:
#9271

ivysrono:
Oak Ridge National Laboratory (partial) ?

I can't find downgrade rule on https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Oak-Ridge-National-Laboratory.xml

J0WI:
The ruleset name has also changed, so this entry is obsolete.

ivysrono:
#9304

ivysrono:
#9311

ivysrono:
#9318 
#9319 
#9320

ivysrono:
#9334 
#9335 
#9339

ivysrono:
#9349 
#9350

ivysrono:
#9381 
End.

PR #14175, opened 2018-01-04
Forbid the creation of fallback rules

cschanaj:
#14336

### DIFF
--- a/utils/relaxng.xml
+++ b/utils/relaxng.xml
@@ -72,7 +72,7 @@
             <!-- The to attribute of a rule must not contain spaces or
                  backslashes. It also must contain at least one slash after the
                  hostname. Alternatively it can be the literal string "https:"-->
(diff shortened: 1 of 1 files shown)


jeremyn:
Was there a clean-up issue where we went through and rewrote all the rules that had fallbacks? I feel like there was an issue like this but I can't find it now.

cschanaj:
I think you are referring to #8833. We don't have any downgrade rules since `5.2.16`, as suggested in https://github.com/EFForg/https-everywhere/issues/9621#issuecomment-299241902.

jeremyn:
@cschanaj Yes, #8833 looks like the right issue, thanks.

J0WI:
@Hainish this change looks trivial, isn't it?

Hainish:
:+1:

PR #14182, opened 2018-01-04
Create kiro7.com.xml

Bisaloo:


src/chrome/content/rules/kiro7.com.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Invalid certificate:
+		a135.kiro7.com
+		amp.kiro7.com
+		click.e.kiro7.com
+		image.e.kiro7.com
+		li.kiro7.com
+		sa135.kiro7.com
```
J0WI: The page tries to load some scripts from there. I don't know if they are used somewhere. Do we care?

Bisaloo: Hum, I don't see this.

J0WI: @jeremyn could you double check?

jeremyn: I'm not sure what I'm checking. https://sa135.kiro7.com has an invalid certificate, and I don't see any network requests to `sa135.kiro7.com` from https://www.kiro7.com/.

PR #14187, opened 2018-01-05
Create HDME.eu.xml

christianheinrichs:


jeremyn:
This is not a good domain to cover because https://hdme.eu is almost entirely behind a login so it's hard to know that HTTPS is safe to use there. Also the plain HTTP version of the page doesn't redirect itself to HTTPS which is also not a good sign that we can trust that HTTPS works behind the login.

As it is, I think this pull request should be closed without merging. You might suggest to the site owners for them to look into HSTS preloading at https://hstspreload.org .

PR #14192, opened 2018-01-05
Create CommitStrip.com.xml

Bisaloo:


J0WI:
MCB seems to be trivial.

PR #14193, opened 2018-01-05
[TPG.com] Update, rename and reactivate ruleset

Bisaloo:


J0WI:
There is also `TPG-Capital.xml` but that one should be split anyway.

PR #14194, opened 2018-01-05
[PressLabs.com] Simplify and rename ruleset

Bisaloo:


jeremyn:
Add `target`

- [ ] https://billing.presslabs.com and remove from `Login required`. It has an STS header so I assume it works behind the login.
- [ ] https://dev.presslabs.com and remove from `Login required`, for the same reason.
- [ ] https://cdn.dev.presslabs.com and remove from `Login required`, because we are making `dev` a `target`
- [ ] https://o.presslabs.com

Also:
- [ ] Remove the `test` for https://cdn.presslabs.com, the 301 for the implicit test is fine.

Bisaloo:
Can we keep the test url for cdn.presslabs.com? I like having additional test urls when `$` doesn't return the same HTTP code than the other paths of the subdomain.

jeremyn:
Looking at just this ruleset, I understand why you want the test. The point of `cdn` is probably to serve static assets, and the redirect behavior for the root URL is atypical.

The problem I have is that, for me -- maybe some of the other maintainers disagree -- I don't like the idea of optional extra tests. Each test should be there for some specific reason, and in any situation either a test is required to be there or to be absent. So going forward I would either need to ask other contributors to add these tests, or else remember that you personally have a particular preference, and I don't really like either option.

Anyway what we are really testing in most cases is whether there is HTTPS misconfiguration at the server, which any 2XX/3XX response rules out.

jeremyn:
I merged PR https://github.com/EFForg/https-everywhere/pull/14359 which has a similar situation as the one here, an extra `test` for an `img` subdomain. We can go ahead and merge this one if you want as it is, just let me know. Still, I would like to keep the number of tests down.

Bisaloo:
> Still, I would like to keep the number of tests down.

I will try to keep that in mind. I would be great if you can merge it as it is.

jeremyn:
We should also update the filename reference in `PressLabs_SSL.xml`. I've added a new item to the checklist. I'm sorry to add this at the last minute.

jeremyn:
Thanks, merged.

PR #14195, opened 2018-01-05
[ST.org] Trivialize ruleset

Bisaloo:


jeremyn:
- [ ] Add `target` for https://eutbildning.st.org and remove it from `Login required`. The plain HTTP version of the page http://eutbildning.st.org redirects to HTTPS so in this case rewriting should be safe.
- [ ] Move `wiki.st.org (broken chain)` from `Login required` to `Invalid certificate`.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/14195/commits/52122b608153bf96208a9c23469e507c7a7966cc, oops.

jeremyn:
Thanks, merged.

PR #14204, opened 2018-01-06
[Kodi.tv] Add targets

Bisaloo:


src/chrome/content/rules/Kodi.tv.xml
```diff
@@ -1,34 +1,28 @@
 <!--
-	Nonfunctional hosts in *kodi.tv:
-
-		- addons ¹
-		- chatlog ¹
-		- forum ¹
-		- ftp ¹
```
J0WI: That one is missed

Issue #12901, opened 2017-10-01
www.nationalgeographic.com


szhorvat:
Type: ruleset issue

Domain: www.nationalgeographic.com

With HTTP Everywhere, redirects to www-s.nationalgeographic.com, which shows an error page.

bardiharborow:
Confirmed. `www` seems to redirect to HTTP now, and `www-s` is defunct.

Bisaloo:
@J0WI, can you close this now please? It was fixed in #14211

PR #14164, opened 2018-01-03
 Update National_Geographic.xml

episkorowski:
#12901
Just removed the www-s redirect, website loads now.

I had trouble with the testing, since some of the subdomains in the rule don't seem to exist anymore, namely donate and natgeotv. Can't create test URLs for URLs that aren't valid anymore, so I'm not sure if I can proceed without restructuring everything.

This is my first pull request for the project, so let me know if I messed something up.

PR #14211, opened 2018-01-06
Update National_Geographic.xml

episkorowski:
#12901
@Bisaloo New PR. Added a barebones list of subdomains and merged `National_Geographic.com-problematic.xml` into the primary ruleset in order to use the simplified regex.

Ruleset no longer uses www-s, so the website should load now.


Bisaloo:
Great! It is passing tests now. Some things to change:

- [ ] `archive.nationalgeographic.com` has mixed content issues. See for example https://archive.nationalgeographic.com/?iid=133596#folio=CV1. So it should not be included in this ruleset. You can add it to the top comment and explain there is active mixed content.
- [ ] Update the top comment. Right now, it includes a lot of unnecessary info so if you want, you can stick to the format in our [contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts).
- [ ] Remove the trailing tab line 108
- [ ] Remove line 126. `donate.nationalgeographic.com` is not listed in your targets so we shouldn't try to secure the cookie for this subdomain.
- [ ] http://media-members.nationalgeographic.com, http://media-mmdb.nationalgeographic.com, http://media-yourshot.nationalgeographic.com all return `HTTP 403`. In this case, we like having a `test url` that returns `HTTP 200`.
  For example
```xml
	<target host="media-mmdb.nationalgeographic.com" />
		<test url="http://media-mmdb.nationalgeographic.com/static-media/img/favicon.ico" />
```

- [ ] `press.nationalgeographic.com` has mixed content as well but as far as I can tell, it is only fonts and ads so it is okay to keep it.
- [ ] Rename the file to `NationalGeographic.com.xml`

episkorowski:
I've implemented most of the changes, but I'm running into some trouble with getting a test url for `media-yourshot.nationalgeographic.com`. The one you provided works fine for http://media-members.nationalgeographic.com and http://media-mmdb.nationalgeographic.com, but results in a `404` for `media-yourshot`, which means it fails the testing. 

This might just be lack of experience on my part, but I'm not sure how to get a test URL for it either, since the URL by itself results in the `403`. Do you have any suggestions? For now I just left it as is.

Bisaloo:
Yes, finding test urls can be tricky.

Usually, you can find them with your search engine. `site:media-yourshot.nationalgeographic.com` or `"media-yourshort.nationalgeographic.com`. But it doesn't work in this case.

According to the name, it must be a cdn for `yourshort.nationalgeographic.com` so I tried visiting this subdomain but still nothing.

A last thing that can sometimes help for cdn is visiting [reddit](https://www.reddit.com/domain/media-yourshot.nationalgeographic.com/). Still nothing.

If we can't find a working test url, it is safer to remove the target and add it to the top comment under `No working URL known`.

Bisaloo:
As you have noticed, it is easy to miss the mixed content on `archive.nationalgeographic.com`. Could you add the URL I provided to the top comment as example? It will greatly help the next person who works on this ruleset.

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,44 @@
...
+	<target host="media-yourshot.nationalgeographic.com" />
+	<target host="members.nationalgeographic.com" />
+	<target host="mmdb.nationalgeographic.com" />
+	<target host="news.nationalgeographic.com" />
+	<target host="press.nationalgeographic.com" />
+	<target host="shop.nationalgeographic.com" />
+
+	<!--	Tracking cookies:
+					-->
+	<securecookie host="^\.nationalgeographic\.com$" name="^(?:gpv_p\d+|s_\w+|__utm)\w$" />
```
Bisaloo: While we are at it, could you use a [capturing group](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#regex-conventions) in the regex?

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,44 @@
...
+		- archive
+		- intelligenttravel		
+		- news
+		- press
+		- video
+
+-->
+<ruleset name="National Geographic">
+
+	<target host="www.nationalgeographic.com" />
```
Bisaloo: Please also add

```xml
	<target host="nationalgeographic.com" />

	<rule from="^http://nationalgeographic.com/"
		to="https://www.nationalgeographic.com/" />
```

with a note in the top comment explaining that `^` times out over `https`.

episkorowski:
I'm not totally sure if what I did with the regex is correct, but I've implemented the other changes.

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
+<!--
+
+	Redirect to HTTP:
+
+		- adventureblog
```
Bisaloo: I have `Invalid certificate` for this one.

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
...
+		- channel
+		- help
```
Bisaloo: `Invalid certificate` for me.

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
...
+		- yourshot
```
Bisaloo: works for me over `https`

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
...
+		- yourshotblog
```
Bisaloo: invalid certificate

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
...
+	
+	Mixed content:
+
+		- archive (Example: https://archive.nationalgeographic.com/?iid=133596#folio=CV1)
+		- intelligenttravel		
+		- news
```
Bisaloo: I can't find any page with mixed content for `news`. Do you have an example?

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,55 @@
...
+		- press
+		- video
```
Bisaloo: I can't find any mixed content, do you have an example?

(Note that we only care about **active** mixed content)

Bisaloo:
I don't have the same results as you for some subdomains listed in the top comment, can you double check?

episkorowski:
Sorry for the delay. I ran through those subdomains again and most of it was consistent with what you said, so I made the changes. Also, I wasn't sure of the difference between active and passive mixed content at first, so I fixed that as well.

I couldn't get `channel` to work over HTTPS, I was getting the "too many redirects" error in the tests, so I made a note of that in the comment.

Bisaloo:
@episkorowski, thank you and sorry for the hassle. As I said, this is a very tricky ruleset. To be honest, I have been wanting to update it myself for a long time but hadn't found the time yet. I knew it was going to be tedious.

@jeremyn @J0WI, feel free to add some comments but this should be about ready to merge (follow up of https://github.com/EFForg/https-everywhere/pull/14164).

Bisaloo:
> I couldn't get channel to work over HTTPS, I was getting the "too many redirects" error in the tests, so I made a note of that in the comment.

This means that the target redirects to plain `http`.

episkorowski:
Ah, okay. I'll correct that real quick.

J0WI:
Thanks @Bisaloo for the comments.
@episkorowski please make sure that all previous comments/rules are present in the new ruleset (as long as the host does still exist). E. g. `mshop` does still exist, but is no longer present in the new ruleset.

episkorowski:
I added the rules I missed. I also added the sections for CDN buckets and the mixed content on `news` from the old ruleset. I'm not totally sure how relevant the info still is, though.

`images` doesn't appear to exist anymore, so I removed mentions of it.

I wasn't sure what to do with `metric`, since the link in the rule (`https://nationalgeographic-com.112.2o7.net/`) leads to a 404 now.

Bisaloo:
For `metric`, you are right we don't want to keep the old rule but you can still list the subdomain under `Invalid certificate`.

src/chrome/content/rules/National_Geographic.com.xml
```diff
@@ -0,0 +1,85 @@
...
+	Time out:
+		- nationalgeographic.com *
+
+	* Standard rule times out over HTTPS, new rule redirects to https://www.nationalgeographic.com/
+
+-->
+<ruleset name="National Geographic">
+
+	<target host="www.nationalgeographic.com" />
+	<target host="nationalgeographic.com" />
```
J0WI: Should be the first target.

episkorowski:
I've added `media-channel` to the "Redirects to HTTP" section. 

I'm not totally sure what to do with `images`, though. `https://images.nationalgeographic.com/` gives me a 502 Error, same as `https://www-s.nationalgeographic.com/.` Should I make a category for it or just put it into one of the existing ones?

J0WI:
> I've added `media-channel` to the "Redirects to HTTP" section.

There is only `channel`

> I'm not totally sure what to do with `images`, though.

Since http and https doesn't behave the same way I would add a new category for it.

episkorowski:
Added a category for `images` and put `media-channel` in the "Redirects to HTTP" section.

Reordered targets and subdomains in the `Invalid certificate` section to meet standards.

PR #14212, opened 2018-01-06
[Zhaw.ch] New ruleset for Zurich University of Applied Sciences

coffeemakr:
Create a ruleset for _Zurich University of Applied Sciences (ZHAW)_.

There are a few non-functional subdomains. Should I add a wildcard for all the subdomains with some exceptions instead of adding all those targets?


J0WI:
I'm totally fine with this, wildcards are more difficult to test.

PR #14223, opened 2018-01-08
[Mudah.my] New ruleset

youdly:


src/chrome/content/rules/Mudah.my.xml
```diff
@@ -0,0 +1,70 @@
...
+		- ooglewww	(t)
+		- panel	(m)
+		- stompinsabah	(t)
+		- survey	(t)
+		- ww	(t)
+		- www.ww	(t)
+		- ww43	(t)
+		- wwa	(t)
+		- www1	(t)
+		- www2	(t)
```
J0WI: WFM

src/chrome/content/rules/Mudah.my.xml
```diff
@@ -0,0 +1,70 @@
...
+		- www9	(t)
+		- wwww	(t)
+		- www.wwww	(t)
+		- wwwww	(t)
```
J0WI: Wildcard DNS (all `(t)`)

src/chrome/content/rules/Mudah.my.xml
youdly: Added

src/chrome/content/rules/Mudah.my.xml
J0WI: A comment about the wildcard should be added.

youdly: Done

PR #14229, opened 2018-01-08
[FreshFlowers.com.au] New ruleset

youdly:


coffeemakr:
Maybe you also want to add `cdn.freshflowers.com.au`

youdly:
Added

PR #14231, opened 2018-01-08
[DuckDuckGo] Update targets

coffeemakr:
Many certificates have been fixed. Also Include `favicons.` and `icons.` with the correct exclusion rule and tests.
`va-l1.duckduckgo.com` doesn't seem to exist anymore.

### DIFF
--- a/src/chrome/content/rules/DuckDuckGo.xml
+++ b/src/chrome/content/rules/DuckDuckGo.xml
@@ -40,14 +40,12 @@
 	<target host="dub.duckduckgo.com" />
 	<target host="favicons.duckduckgo.com" />
 		<test url="http://favicons.duckduckgo.com/ip2/twitter.com.ico" />
-		<exclusion pattern="^http://favicons\.duckduckgo\.com/$" />
 	<target host="ff.duckduckgo.com" />
 	<target host="help.duckduckgo.com" />
 	<target host="i.duckduckgo.com" />
 	<target host="images.duckduckgo.com" />
 	<target host="icons.duckduckgo.com" />
 		<test url="http://icons.duckduckgo.com/ip2/twitter.com.ico" />
-		<exclusion pattern="^http://icons\.duckduckgo\.com/$" />
 	<target host="karma.duckduckgo.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/DuckDuckGo.xml
```diff
@@ -41,24 +34,35 @@
...
+	<target host="favicons.duckduckgo.com" />
+		<test url="http://favicons.duckduckgo.com/ip2/twitter.com.ico" />
+		<exclusion pattern="^http://favicons\.duckduckgo\.com/$" />
 	<target host="ff.duckduckgo.com" />
+	<target host="help.duckduckgo.com" />
 	<target host="i.duckduckgo.com" />
 	<target host="images.duckduckgo.com" />
+	<target host="icons.duckduckgo.com" />
+		<test url="http://icons.duckduckgo.com/ip2/twitter.com.ico" />
+		<exclusion pattern="^http://icons\.duckduckgo\.com/$" />
```
J0WI: Why is the exclusion here required?

src/chrome/content/rules/DuckDuckGo.xml
```diff
@@ -20,17 +20,10 @@
...
-			 - blog.duckduckgo.com
-			 - help.duckduckgo.com
-			 - meme.duckduckgo.com
-			 - settings.duckduckgo.com
-			 - shop.duckduckgo.com
-			 - wiki.duckduckgo.com
 			 - www.ddg.gg
 
-		5xx server error:
-			 - favicons.duckduckgo.com
```
J0WI: You may keep this somewhere around, that we know why the exclusion is there.

src/chrome/content/rules/DuckDuckGo.xml
coffeemakr: Strange. When I tried the two domains yesterday it resulted in a 500 error. Know it shows an search interface which results in a 502 when used to search. :/

coffeemakr:
I have removed the comment and the exception because the index pages for `favions` and `icons` don't deliver a 5xx error anymore.

PR #14233, opened 2018-01-08
Added ruleset for baua.de

Jochen-A-Fuerbacher:


jeremyn:
- [ ] Remove https://webmail.baua.de entirely, because it times out with both HTTPS and plain HTTP
- [ ] Remove the `securecookie` tag or restrict it to `www` and `shop`

jeremyn:
Thanks, merged.

PR #14241, opened 2018-01-09
Added ruleset for ktxtr.com

Jochen-A-Fuerbacher:


Jochen-A-Fuerbacher:
Seems like a domain for malware, so closed this PR.

PR #13773, opened 2017-11-27
[ebay.com] Clean-up ruleset

youdly:
Remove country domains for https://github.com/EFForg/https-everywhere/issues/13603
There are >3k subdomains listed by Sublist3r, which is impractical to test.
I rewrite the ruleset based on the subdomains listed in the comment of previous ruleset, which I assume are the most popular/relevant subdomains.

PR #14258, opened 2018-01-11
Delete CSdata1.com.xml

youdly:
`dsa` could not be resolved. Other subdomains simply timeout.


jeremyn:
- [ ] Remove the reference to this file from [`EBay.xml`](https://github.com/EFForg/https-everywhere/blob/29f984ea472021978cb9c619c6de6050494cde33/src/chrome/content/rules/EBay.xml#L7)

youdly:
Pending https://github.com/EFForg/https-everywhere/pull/13773

jeremyn:
I merged this despite the Travis error `ERROR src/chrome/content/rules/EBay.xml: Dangling reference to CSdata1.com.xml` since that should be fixed in #13773 .

jeremyn:
I'm reverting this change with pull request #14277 . The "dangling reference" error will break all Travis runs until https://github.com/EFForg/https-everywhere/pull/13773 is merged.

I recommend moving this intended deletion directly into https://github.com/EFForg/https-everywhere/pull/13773.

PR #10663, opened 2017-06-24
Update and reactivate Hollywood-Reporter.xml.

bardiharborow:
`^` and `www` now redirect to HTTP.

Pinging @jazwiecki (original author of ruleset). Closes #10268.

PR #14263, opened 2018-01-11
[Hollywood-Reporter] reactived rule for bare and www domains

thisisparker:
There's been some trouble with this rule previously (most recently with a redirect to HTTP documented in #10663) but it looks like things are working a little better now.


J0WI:
Thanks for the update!

PR #14277, opened 2018-01-12
Revert "Delete CSdata1.com.xml"

jeremyn:
Reverts EFForg/https-everywhere#14258.

@gloomy-ghost @Hainish @J0WI Can one of you please merge this? It's breaking Travis, sorry.

### DIFF
--- a/src/chrome/content/rules/CSdata1.com.xml
+++ b/src/chrome/content/rules/CSdata1.com.xml
@@ -1,0 +1,14 @@
+<!--
+	Web bugs.
+
+	For other eBay coverage, see EBay.xml.
+
+-->
+<ruleset name="CSdata1.com">
+
+	<target host="dsa.csdata1.com" />
+
+
+	<rule from="^http:" to="https:" />
+
+</ruleset>


jeremyn:
The fetch error

> ERROR src/chrome/content/rules/CSdata1.com.xml: Fetch error: http://dsa.csdata1.com/ => https://dsa.csdata1.com/: (6, 'Could not resolve host: dsa.csdata1.com')

can be ignored, since this is what the original deletion was supposed to fix.

J0WI:
I think the rule needs to be disabled, otherwise we still have a broken build.

jeremyn:
Travis is currently broken because of the dangling reference. If we re-enable the rule, the dangling reference error should disappear, and we'll be left with one ruleset with a bad `target`, and we have many of those already without breaking the builds.

Will Travis somehow sense that `dsa.csdata1.com` has been changed and keep trying to check it? I don't know how that works. If that's what will happen, then we should leave this deleted and instead delete the dangling reference from `EBay.xml`. In that case @youdly will have to deal with rebasing their changes, which is not ideal for them but may be necessary.

jeremyn:
@J0WI I propose that you merge this reversion, and then we rerun the tests on some unrelated PR to see if Travis keeps complaining about `dsa.csdata1.com`. If it does, then I can make another PR that re-deletes this ruleset and also deletes the dangling reference.

cschanaj:
> Travis is currently broken because of the dangling reference. If we re-enable the rule, the dangling reference error should disappear, and we'll be left with one ruleset with a bad target, and we have many of those already without breaking the builds.

@jeremyn @J0WI I can confirm this. AFAIK, Merging this will not break Travis for future build. This is because `fetch-test` works on changed rulesets only.

PR #14279, opened 2018-01-12
[Infobae.com] new ruleset

thisisparker:


thisisparker:
I think the CI build failed due to external factors, but Infobae has been added to the HSTS preload list anyway and so is no longer a good candidate for a rule.

PR #14281, opened 2018-01-12
[SFGate.com] new ruleset

thisisparker:


zoracon:
These errors occurred with this host:
```
ERROR src/chrome/content/rules/SFGate.com.xml: Fetch error: http://login.sfgate.com/ => https://login.sfgate.com/: (60, 'SSL certificate problem: certificate has expired')

ERROR src/chrome/content/rules/SFGate.com.xml: Fetch error: http://www.s.sfgate.com/ => https://www.s.sfgate.com/: (6, 'Could not resolve host: www.s.sfgate.com')
```

PR #14282, opened 2018-01-13
[Ticketmaster] Move eticketing.co.uk to its own ruleset

Bisaloo:


J0WI:
Is there still a relationship to `Live-Nation.xml` or `Ticketmaster.xml`? If so, it should be commented.

Bisaloo:
I'm not sure. It redirects to ticketmaster.co.uk but that's all I know.

J0WI:
So let's keep it.

PR #14286, opened 2018-01-14
Bild.de:

numismatika:
- add subdomains


numismatika:
Code has version numbers so it may fail later, for now added 3 files there

jeremyn:
I don't like having extra tests, it can make it unclear what exactly is being tested. Can you please remove two of them? Also please move the `test` directly under the `target` with an extra indent.

jeremyn:
This is ready except that the `test` for `code.bildstatic.de` should be moved directly under the matching `target`, indented as you have it.

PR #14291, opened 2018-01-14
Smart-AdServer:

numismatika:
- re-enable ruleset

### DIFF
--- a/src/chrome/content/rules/Smart-AdServer.xml
+++ b/src/chrome/content/rules/Smart-AdServer.xml
@@ -1,10 +1,5 @@
-
 <!--
-Disabled by https-everywhere-checker because:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Smart-AdServer.xml
```diff
@@ -157,7 +161,6 @@ Fetch error: http://www84.smartadserver.com/ => https://www84.smartadserver.com/
 	<target host="manage-v5.smartadserver.com" />
 	<target host="manage-v53.smartadserver.com" />
 	<target host="manage.smartadserver.com" />
-	<target host="managebeta.smartadserver.com" />
```
J0WI: This host does still exist

PR #14306, opened 2018-01-16
Create LeContrepied.org.xml

Bisaloo:


jeremyn:
- [ ] Add `target` for https://crm.lecontrepied.org and remove from `Login required`, because it redirects itself to HTTP so it should be okay for us to cover.

jeremyn:
Thanks, merged.

PR #14309, opened 2018-01-17
Update OpenNIC.org.xml

cschanaj:


J0WI:
Please double check the removed targets, they WFM.

PR #14317, opened 2018-01-17
Added ruleset for unpkg.com

Jochen-A-Fuerbacher:


Jochen-A-Fuerbacher:
Targets already in HSTS.

PR #14327, opened 2018-01-18
 Remove preloaded test URLs from Amazon rulesets 

cschanaj:


J0WI:
@cschanaj I'll also accept one PR for updating all amazon rules, since the change is the same everywhere.

J0WI:
Looks like `pda-bes.amazon.com` is no longer active. LGTM otherwise.

cschanaj:
@J0WI I have moved the domain to the top comment. Thanks!!

PR #14293, opened 2018-01-15
Update https_everywhere_checker/rules.py, Fix #14290

cschanaj:
Make Travis to assert that each test url is covered by at least one target

- [x] Amazon-Adsystem.com.xml (#14357)
- [x] Amazon.xml (#14327)
- [ ] AmazonWebServices.xml (#14457)
- [x] Cloudinary.xml (#14430)
- [x] CrossRef.org.xml (#14294)
- [x] General-anzeiger-bonn.de.xml (#14300)
- [x] Gtimg.cn.xml (#14358)
- [ ] Hatena.xml
- [x] LiveJournal.com.xml (#14301)
- [x] ManageEngine.jp.xml (#14402)
- [x] Massengeschmack.tv.xml (#14302)
- [x] Migros-Bank.xml (#14340)
- [x] OpenNIC.org.xml (#14309)
- [x] Qpic.cn.xml (#14337)
- [x] Sikur.xml (#14338)
- [x] Sohu.com.xml (#14339)
- [x] Verizon.xml (#14429)
- [ ] Wemfbox.ch.xml (#14330)
- [x] XVideos.com.xml (#14341)
- [x] Zwame.pt.xml (#14342)
- [x] amazon.ca.xml (#14327)
- [x] amazon.co.jp.xml (#14327)
- [x] amazon.co.uk.xml (#14327)
- [x] amazon.com.au.xml (#14327)
- [x] amazon.com.br.xml (#14327)
- [x] amazon.com.mx.xml (~#14328~, #14327)
- [x] amazon.de.xml (#14327)
- [x] amazon.es.xml (#14327)
- [x] amazon.fr.xml (#14327)
- [x] amazon.it.xml (#14327)
- [x] amazon.nl.xml (#14327)
- [x] centrum.cz.xml (#14299)
- [ ] freetalklive.com.xml
- [x] lofter.xml (#14297)
- [x] naukimg.com.xml (#14403)
- [x] polyvoreimg.com.xml (#14343)
- [x] pornBB.xml (#14359)
- [x] thehubpeople.com.xml (#14360)
- [x] uptodown.net.xml (#14296)
- [x] webcaster.pro.xml (#14428)

PR #14330, opened 2018-01-18
Add Net-metrix.ch.xml and update Wemfbox.ch.xml

J0WI:
`Wemfbox.ch.xml` is only tracking anyway.
See also: #14293

### DIFF
--- a/src/chrome/content/rules/Net-metrix.ch.xml
+++ b/src/chrome/content/rules/Net-metrix.ch.xml
@@ -1,0 +1,24 @@
+<!--
+	For other NET-Metrix coverage, see Wemfbox.ch.xml.
+
+	Nonfunctional hosts in *.net-metrix.ch:
+		- sehstoff.net-metrix.ch (m)
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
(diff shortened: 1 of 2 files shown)


jeremyn:
Where are you getting information about the redirects? For example how would I verify this change is correct?
```xml
		<rule from="^http://wetter\.wemfbox\.ch/"
		to="https://we-ssl.wemfbox.ch/" />
```

to
```xml
		<rule from="^http://wirelter\.wemfbox\.ch/"
		to="https://we-ssl.wemfbox.ch/" />
```

Neither `wetter.wemfbox.ch` nor `wirelter.wemfbox.ch` redirect themselves to `we-ssl.wemfbox.ch`.

J0WI:
I was searching for the source of `wirelter.wemfbox.ch` (which I found in an archived version of the site) and found `we-ssl.wemfbox.ch` in the current source code of `https://www.wireltern.ch/`.

cschanaj:
related to #14660

Bisaloo:
Related: #12297

J0WI:
I just rebased this. Would be great to get a review here to fix the mentioned issues.

PR #14331, opened 2018-01-18
Add PCtipp.ch.xml

J0WI:


jeremyn:
I haven't completed reviewed this, but:

http://pctipp.ch/typo3_src/typo3/sysext/install/composer.json gets three 3XX responses and ends up at http://www.pctipp.ch/seite-nicht-gefunden/ . Is that expected?

Why do you have parentheses around `(aktivierung|benutzerdaten|login)`? It's a third capturing group but you only have `$1` and `$2` in your rewrite target.

J0WI:
> I haven't completed reviewed this, but:
> `http://pctipp.ch/typo3_src/typo3/sysext/install/composer.json` gets three 3XX responses and ends up at `http://www.pctipp.ch/seite-nicht-gefunden/` . Is that expected?

Nope. The TYPO3 version is far older than expected, so I updated the test urls.

> Why do you have parentheses around `(aktivierung|benutzerdaten|login)`? It's a third capturing group but you only have $1 and $2 in your rewrite target.

They are all prefixed by `mein-pctipp/`. Since this is all inside `$2` it should work. Do you have any issues with this pattern?

jeremyn:
> Do you have any issues with this pattern?

Sorry, I missed that it was a slash `/` before that group, not a pipe `|`. The pattern is fine.

jeremyn:
- [ ] It looks like `^` generally redirects to `www`, and it's `www` that has the HTTP redirect problem, so I think we can safely cover all of `^`. We can add a `rule` after the existing `rule` specifically for `^`, delete all the `^` `test`s, and simplify the existing `rule`.
- [ ] Mark `go`, `m`, and `redi` with an `(m)`
- [ ] Add `www` to the top comment with an `(h)`
- [ ] Add a `www` `test` for `forum`
- [ ] Sort the `test`s in the same order that they show up in the `rule`, for example put `mein-pctipp/aktivierung/` above `mein-pctipp/benutzerdaten/` and `mein-pctipp/*` above `typo3`

J0WI:
Done.

jeremyn:
There's an MCB error at https://www.pctipp.ch/forum/forum.php for http://www.pctipp.ch/forum/clientscript/vbulletin-core.js?v=422 and some errors "ReferenceError: vbphrase is not defined" and "ReferenceError: vBulletin_init is not defined". I don't think we should include all of `forum`. I think `forum/images/` and `forum/clientscript/` should be okay. What do you think?

J0WI:
I didn't notice any issues. However, I adjusted the rule.

jeremyn:
Thanks, merged.

I didn't see any actual problems using the site, however maybe the errors only cause problems once you're logged in.

PR #14335, opened 2018-01-18
[Cvent.com] Add warning and test url

Bisaloo:


jeremyn:
I don't like vague warnings like this because they will effectively never be removed.

It looks like the problem is that pre-existing `test`s like http://www.cvent.com/events/csh-summit-2016/event-summary-b05896613e504f549b12ef1d73909e59.aspx?ct=48832732-1e0d-43b1-8af7-01e4e9cd58aa now show a `The page you are requesting from this event is no longer available.` message, and the page works fine for HTTPS, but problematic URLs like the new one you added http://www.cvent.com/events/isbe-2018/custom-118-8b8c6b90d61e4194b0893e21e3ef0553.aspx redirect to HTTP. Is that right? If so then I think the error should be more specific about what's happening: "these URLs are a constantly changing series of events and" etc.

jeremyn:
Thanks, merged.

Hainish:
For some reason this caused the `chrome beta` test to fail: https://travis-ci.org/EFForg/https-everywhere/jobs/331031530.  Re-running on master.

Hainish:
Re-running fixed the issue.

jeremyn:
@Hainish The link to Travis you gave shows passing so I'm not sure what the specific error was, though I did get an error email from Travis. However `chrome beta` regularly fails with a time out error, to the point where I mostly ignore it as long as the tests pass in the PR before the merge.

Hainish:
@jeremyn I've been keeping my eye on builds recently.  I've noticed slow build spin-up times, and sometimes hanging builds.  Travis has had caused some significant trouble for other projects here at EFF in the last few days, to the point of starting to consider alternatives.  We're just starting to look into Circle CI to see if it might be a viable alternative.

PR #14340, opened 2018-01-19
Update Migros-Bank.xml

cschanaj:


jeremyn:
Removing `banking` is fine.

The `blog` behavior isn't right though. Without browser history, and not counting intermediate redirects, the redirects work like this:

http://blog.migrosbank.ch > https://blog.migrosbank.ch/de/
http://blog.bancamigros.ch > https://blog.migrosbank.ch/it/
http://blog.banquemigros.ch > https://blog.migrosbank.ch/fr/

Also it looks like once you visit one language site, it can affect what happens with other sites. For example if you first visit http://blog.banquemigros.ch and get redirected to the `fr/` site, and then visit http://blog.migrosbank.ch , you'll be redirected to `fr/` instead of `de/`. However http://blog.bancamigros.ch will still redirect you to https://blog.migrosbank.ch/it/ .

I think we should remove the `target`s for both http://blog.bancamigros.ch and http://blog.banquemigros.ch and their `rule`s.

cschanaj:
@jeremyn travis is passing now. thanks!!

jeremyn:
Thanks, merged.

PR #14342, opened 2018-01-19
Update Zwame.pt.xml

cschanaj:


src/chrome/content/rules/Zwame.pt.xml
```diff
@@ -1,94 +1,28 @@
...
-
-	* Secured by us
+	Non-functional hosts
+		Couldn't connect to server:
+		- mailsrv.zwame.pt
 
+		SSL peer certificate was not OK:
+		- go.zwame.pt
 -->
 <ruleset name="Zwame.pt (partial)">
```
J0WI: `(partial)` can be removed.

cschanaj:
@J0WI fixed. thanks!

PR #14345, opened 2018-01-19
Fix #13953

### LINKED ISSUE #13953: Breaks tutti.ch
[//]: #
[//]: #
[//]: # (Delete all but one of the following "Type" lines.)
[//]: # (Leave in the line that best describes the issue you are reporting.)

Type: ruleset issue

[//]: # (If you are reporting a ruleset/website problem, include the top-level)
[//]: # (domain below. For example, if you want to report an issue about)
[//]: # ("one.example.com" and "two.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: example.com)
[//]: #
[//]: # (Be sure to remove the parenthesis and comment marker.  If you are only)
[//]: # (reporting an issue about "one.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: one.example.com) 
[//]: #
[//]: # (Only include one top-level domain. If you have more than one top-level)
[//]: # (domain to report, such as both "example.com" and "example.org", open a)
[//]: # (new issue for each top-level domain.)
[//]: #
[//]: # (If you are not reporting a ruleset/website problem, you can delete the)
[//]: # ("Domain" line or leave it blank.)

Domain: tutti.ch

Issues with rules in https://github.com/EFForg/https-everywhere/pull/6489/files

It seems, the main page redirects straight back to http. From there on the ruleset will cause cross domain issues.

As an example open any ad, like http://www.tutti.ch/bern/fahrzeuge/motorraeder/angebote/piaggio-vespa-gtr_17284472.htm and click on "Nummer anzeigen" to uncover the phone number. This will cause the following error in the latest Firefox:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.tutti.ch/ajax/phoneview. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
```

I contacted the operators and they are not willing to fix the issue and do not have a timeline for https support (ads ftw).

J0WI:
This should fix the issue described in #13953.

### DIFF
--- a/src/chrome/content/rules/tutti.ch.xml
+++ b/src/chrome/content/rules/tutti.ch.xml
@@ -13,10 +13,14 @@
 	<target host="m.tutti.ch" />
 
 	<!-- product pages on desktop redirect to http -->
-	<exclusion pattern="http://www\.tutti\.ch/[\w/-]+\d{3,}\.htm$" />
+	<exclusion pattern="^http://www\.tutti\.ch/[\w/-]+\d{3,}\.htm$" />
 	<test url="http://www.tutti.ch/bern/kleidung-accessoires/schuhe-fuer-damen/angebote/nike-schuhe_11448617.htm" />
 	<test url="http://www.tutti.ch/neuchatel/livres/romans/offres/juste-avant-l-orage_3288278.htm" />
 	<test url="http://www.tutti.ch/tessin-fr/maison/electromenager-ustensiles/offres/climatizzatore_11448580.htm" />
+
+	<!-- CORS issue https://github.com/EFForg/https-everywhere/issues/13953 -->
+	<exclusion pattern="^http://www\.tutti\.ch/ajax/" />
+	<test url="https://www.tutti.ch/ajax/phoneview" />
 
 	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


jeremyn:
After including the fix you have and clicking "Nummer anzeigen" at http://www.tutti.ch/bern/fahrzeuge/motorraeder/angebote/piaggio-vespa-gtr_17284472.htm as described in issue https://github.com/EFForg/https-everywhere/issues/13953, there's another CORS block for https://www.tutti.ch/get_marked. I can't tell what this is doing but maybe we should exclude `get_marked` while we're looking at this.

J0WI:
Referring to it's output it has something to do with advertisement. That's why I didn't add it.

jeremyn:
Sure, but in this case the whole page is an ad, and the original CORS problem happens when the user interacts with it, so we want the ad to work. Maybe if you're logged in, `/get_marked` marks this ad as viewed in some list the user cares about.

J0WI:
Okay, I added this as well.

jeremyn:
Thanks, merged.

PR #13904, opened 2017-12-07
Create alaalam.org.xml

99n45i:

PR #14346, opened 2018-01-19
Add Alaalam.org.xml

jeremyn:
This pull request replaces pull request https://github.com/EFForg/https-everywhere/pull/13904.

### DIFF
--- a/src/chrome/content/rules/Alaalam.org.xml
+++ b/src/chrome/content/rules/Alaalam.org.xml
@@ -1,0 +1,14 @@
+<!--
+	Login required:
+		cpanel.alaalam.org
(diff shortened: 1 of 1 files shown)


Bisaloo:
I get `Unable to connect` with both HTTP and HTTPS for:

- cpanel.alaalam.org
- webdisk.alaalam.org
- webmail.alaalam.org

Can you double check and remove them from the top comment if you can reproduce?

jeremyn:
In my case those three subdomains are missing from DNS entirely. Maybe that's what you meant, but I usually think of `unable to connect` as `connection refused`, which is something different. Anyway I've deleted the top comment entirely.

PR #14347, opened 2018-01-19
add rule for Clickmatic.pl

numismatika:


jeremyn:
- [ ] Remove https://mdb.clickmatic.pl entirely, it fails with both HTTPS and plain HTTP
- [ ] Move https://beta.clickmatic.pl and https://poczta.clickmatic.pl from `Self-signed cert` to `Cert mismatch`, since they are also mismatches and we don't need two groups. Optionally you can just relabel `Cert mismatch` as `Invalid certificate`.
- [ ] Add a `test` for https://cdn.clickmatic.pl or remove the `target` and move it to the top under `No working URL known`
- [ ] Remove the `target` for https://dmp.clickmatic.pl and move it to the top under `Login required`

https://i.clickmatic.pl should also get a `test`, but in this case I could find a working URL through a web search. I only found one URL though, and it looks like it has a long number in it (it ends in `/cookie-match`) so I don't want to include it in the ruleset. If you can find a better `test` `url` for https://i.clickmatic.pl feel free to include it as a `test`.

numismatika:
So far i can't find an url for the i-subdomain who at least looks somewhat stable.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/commit/66530ea4012a58bc903c77499621a6ae19db0076 and added some extra items.

jeremyn:
@numismatika I'd like to move forward on this pull request.

jeremyn:
Thanks, merged.

PR #14349, opened 2018-01-19
Add hs-mannheim.de

DavidLiedke:


src/chrome/content/rules/hs-mannheim.de.xml
```diff
@@ -0,0 +1,26 @@
...
+		www.fsw.hs-mannheim.de
+		www.fotovoltaik.hs-mannheim.de
+	
+	Time Out:
+		www.gestaltung.hs-mannheim.de
+-->		
+<ruleset name="Hochschule Mannheim (partial)">
+
+	<!-- hs-mannheim.de -->
+	<target host="www.hs-mannheim.de"/>
```
J0WI: `www` is already covered by the wildcard, but the wildcard would also break all non-working hosts listed above.

DavidLiedke:
@J0WI

J0WI:
Can you add a comment to document the behavior that the non www version of the listed targets are serving a mismatching cert?

DavidLiedke:
@J0WI 
I add a comment for the cert mismatch.

J0WI:
LGTM, thanks!

PR #14350, opened 2018-01-19
Add cacti.net

DavidLiedke:


src/chrome/content/rules/cacti.net.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="cacti.net">
+	<target host="cacti.net" />
+	<target host="*.cacti.net" />
+		<test url="http://docs.cacti.net/" />
+		<test url="http://forums.cacti.net/" />
```
J0WI: Please prefer a complete list of targets (also add `www`) over wildcards.

DavidLiedke:
@J0WI

J0WI:
Thanks!

PR #14352, opened 2018-01-20
Update Greek-various.xml

cschanaj:
Note:
- `ellak.gr` is preloaded
- `barikat.gr` is covered in its own ruleset 
https://github.com/EFForg/https-everywhere/blob/368eeb82b0d0d7528173aeeda46e8d21a2c50118/src/chrome/content/rules/barikat.gr.xml#L1-L8

### DIFF
--- a/src/chrome/content/rules/dei.gr.xml
+++ b/src/chrome/content/rules/dei.gr.xml
@@ -1,4 +1,6 @@
 <ruleset name="dei.gr">
+	<target host="elearning.dei.gr" />
+	<target host="sapwasrmprd.dei.gr" />
 	<target host="www.dei.gr" />
 	
 	<securecookie host=".+" name=".+" />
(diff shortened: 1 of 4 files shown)


J0WI:
It looks quite easy to spit this ruleset into one file per domain. Would you like to do this?

src/chrome/content/rules/dei.gr.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="dei.gr (partial)">
+	<target host="www.dei.gr" />
```
J0WI:
```
elearning.dei.gr
sapwasrmprd.dei.gr

J0WI: @cschanaj not sure if you also noticed [this](https://github.com/EFForg/https-everywhere/pull/14352#discussion_r162782630) comment.

J0WI:
Thanks!

PR #14359, opened 2018-01-20
Rename and Update PornBB.org.xml

cschanaj:


jeremyn:
Can you please remove the wildcard? There aren't many domains here, maybe ten or so.

jeremyn:
- [ ] Delete https://mail.pornbb.org entirely, it is `unable to connect` for both HTTPS and plain HTTP
- [ ] Remove `test` for https://pornbb.org , we don't need it
- [ ] Remove `test` for https://www.pornbb.org , we don't need it
- [ ] Mark https://static05.pornbb.org as `No working URL known` and remove the `target` (or delete it entirely)
- [ ] Mark https://static5.pornbb.org as `No working URL known` and remove the `target` (or delete it entirely)
- [ ] Change the `ruleset` `name` to `PornBB.org`
- [ ] Change the filename to `PornBB.org.xml`

jeremyn:
Thanks, merged.

PR #14362, opened 2018-01-20
Create u51.com

ivysrono:


cschanaj:
@ivysrono the filename should be `u51.com.xml` instead of `u51.com`

ivysrono:
@cschanaj thanks

src/chrome/content/rules/u51.com.xml
```diff
@@ -0,0 +1,15 @@
+<ruleset name="u51.com">
+	<target host="u51.com" />
+	<target host="www.u51.com" />
+	<target host="credit.u51.com" />
```
J0WI: `h5.u51.com`

PR #14364, opened 2018-01-20
Add Haproxy.org.xml

J0WI:


J0WI:
@Hainish do we have IPv6 problems again?

Hainish:
@J0WI for IPv6 support, we start the `miredo` IPv6-over-4 service, which is run by a third party.  Testing this now, it appears as though this service is down.

Hainish:
This seems to have more to do with the Travis host environment than with `miredo` itself.  I've installed `miredo` on a VPS, and I'm able to `curl -6 http://ipv6.haproxy.org/` and `curl -6 https://ipv6.haproxy.org/` fine.  It's strange that when I try to do the same within docker on a local VM, I get the same error as the Travis environment.  I suspect this has to do with the fact that `miredo` is accessing kernel-level networking (this is why docker needs the `--privileged` flag) and that for some reason the Travis kernel has changed to make ipv6 networking more difficult.

J0WI:
Why is `miredo` used inside the docker container and not outside?

Hainish:
@J0WI Travis lacks ipv6 network access, and you can't arbitrarily install packages such as `miredo` within the Travis env.

Hainish:
This may have changed since last we checked.  Looking into running the fetch tests within the Travis env.

J0WI:
Otherwise we could use `torsocks` for IPv6 😕

Hainish:
I'm testing out the solution detailed https://github.com/travis-ci/travis-ci/issues/8891#issuecomment-353403729 in https://github.com/EFForg/https-everywhere/pull/14386, let's see if tests pass.

jeremyn:
I'm trying to get Miredo working in Docker, and I'm getting this with the current base image, using a command based on [`test/travis.sh`](https://github.com/EFForg/https-everywhere/blob/45dd7e40d3116dadfb9bc70658535b497c1f4256/test/travis.sh#L51):

```
$ sudo docker run --rm -ti --privileged --userns=host electronicfrontierfoundation/https-everywhere-docker-base bash -c "service miredo start"
 * Starting Teredo IPv6 tunneling daemon miredo                                 Cannot create PID file /var/run/miredo.pid:
 Permission denied
Please make sure another instance of the program is not already running.
                                                                         [fail]
```

This is on an Ubuntu 16.04 host system. My host system uses user namespaces, and I'm disabling those with `--userns=host` because those are incompatible with `--privileged`. The original command has `httpse` instead of `electronicfrontierfoundation/https-everywhere-docker-base`, and I'm not sure why but I guess there's a special tag for `httpse` set up in Travis somewhere.

Does anyone know what the permission problem is from? Can anyone get this running locally?

J0WI:
Thanks for investigating! All tests are passing now.
It's quite frustrating that @travis-ci still lacks native IPv6 support. This hacky downgrade is definitely not a long term solution.

Hainish:
@J0WI I'm hoping that they will resolve this issue with the latest trusty build soon.  We are also investigating whether it's worth it to switch over to CircleCI, given the trouble and slowness we've had with Travis build environments to both run tests and to initiate them.

PR #14366, opened 2018-01-20
Iichan.hk.xml

ekeelore:


jeremyn:
Thanks for your pull request. Please:

- [ ] Add a comment to the top saying that this domain has a wildcard DNS entry. For example, make the ruleset look like this:
```xml
<!--
	Wildcard DNS.
-->
<ruleset ...>
...
</ruleset>
```

jeremyn:
Also, I notice that some of the URLs in the left frame use a double slash relative URL like `//example.com` so that your browser will visit `http` or `https` depending on what you are using for the current site. In particular it uses `//` for https://410chan.org which seems to have a bad certificate chain. (You should notice an HTTPS error if you visit this site using a new browser profile.) So, the `410chan.org` links are broken by default for HTTPS. You might want to advise either the site owner for `410chan.org` to fix their HTTPS implementation, or the site owner for `iichan.hk` to provide `http://` links instead of `//` links for `410chan.org`.

(This is just FYI, you don't need to change anything in the ruleset.)

jeremyn:
Thanks, merged.

Issue #14365, opened 2018-01-20
fetch-test.sh incorrectly reports an error because it doesn't follow a non-trivial rewrite rule


jeremyn:
Type: code issue

There are three domains that behave like this:

| URL | result |
|-----|--------|
| http://armytimes.com | 301 http://www.armytimes.com/ |
| https://armytimes.com | Connection refused |
| http://www.armytimes.com | 301 https://www.armytimes.com/ |
| https://www.armytimes.com | 200 |
| http://link.armytimes.com | 302 http://armytimes.com |
| https://link.armytimes.com | 302 http://armytimes.com |

If you run `fetch-test.sh` (d5c54c4ecb9dcab1789903990b57c0cfc53795ba) against the following ruleset, which is a minimal version of [`ArmyTimes.com.xml`](https://github.com/EFForg/https-everywhere/blob/d5c54c4ecb9dcab1789903990b57c0cfc53795ba/src/chrome/content/rules/ArmyTimes.com.xml):
```xml
<ruleset name="ArmyTimes.com (minimal)">
	<target host="armytimes.com" />
	<target host="www.armytimes.com" />
	<target host="link.armytimes.com" />

	<rule from="^http://armytimes\.com/"
		to="https://www.armytimes.com/" />

	<rule from="^http:" to="https:" />
</ruleset>
```

and change the logging level to `debug` in `test/rules/manual.checker.config`, you get:
```
DEBUG Loaded certificate platforms: cacert,chromium,default,firefox
DEBUG Parsing ArmyTimes-minimal.com.xml
DEBUG Checking coverage for 'ArmyTimes.com (minimal)'.
DEBUG Checking target validity for 'ArmyTimes.com (minimal)'.
DEBUG =**= Start http://armytimes.com/ => https://www.armytimes.com/ ****
DEBUG Fetching transformed page https://www.armytimes.com/
DEBUG Fetching plain page http://armytimes.com/
DEBUG Following redirect http://armytimes.com/ => http://www.armytimes.com/
DEBUG Following redirect http://www.armytimes.com/ => https://www.armytimes.com/
DEBUG ==== D: 0.0000; http://armytimes.com/ (140074) -> https://www.armytimes.com/ (140074) =====
INFO Finished comparing http://armytimes.com/ -> https://www.armytimes.com/. Rulefile: ArmyTimes-minimal.com.xml.
DEBUG =**= Start http://www.armytimes.com/ => https://www.armytimes.com/ ****
[...]

PR #14368, opened 2018-01-21
Test

cschanaj:
insertion order matters. but fetch-test ignore order and retrieve applicable rulesets differently... I'm afraid this bug has been around for some time and it is not easy to fix without a major rewrite.


jeremyn:
@cschanaj I guess this is for my issue `https://github.com/EFForg/https-everywhere/issues/14365` .  Thanks for looking at it. If you're done testing, could you please add any comments to the issue itself?

PR #14370, opened 2018-01-21
Add Pururin.us.xml

lievedido:


jeremyn:
This domain is a little tricky because https://forum.pururin.us is broken from mixed content blocking (notice how broken the styling is), but we want to safely rewrite static content from this domain like images, scripts etc.

- [ ] Try this:
```xml
<!--
	Mixed content blocking:
		forum.pururin.us

-->
<ruleset name="Pururin.us">
	<target host="pururin.us" />
	<target host="www.pururin.us" />
	<target host="forum.pururin.us" />
  
  	<rule from="^http://forum\.pururin\.us/(css.php|data/|js/|styles/)"
		to="https://forum.pururin.us/$1" />
	<rule from="^http://(www\.)?pururin\.us/"
		to="https://$1pururin.us/" />
</ruleset>
```

Note that you can do in-browser ruleset debugging by going to `about:addons`in Firefox (or the Chrome equivalent), clicking under `General Settings` and pressing <kbd>Ctrl-Z</kbd>. You can find the various URLs by loading a page with the developer tools' (<kbd>F12</kbd>) `Network` tab open.

Also note that I indent with tabs, instead of spaces which you used.

Please let me know if you have any questions.

jeremyn:
My first version of the ruleset in the checklist had test problems, sorry. I've updated it.

I noticed a request to http://forum.pururin.us/deferred.php , and I'm not sure what caused it, but I've added that to the list. I've also removed the items for `css.php` and `data/` because I couldn't find test URLs that I liked. Ideal test URLs have no identification numbers/version numbers/codes, are safe-for-work, and the response size is small. If you can find some good test URLs, please feel free to add those back in.

jeremyn:
Thanks, merged.

PR #14373, opened 2018-01-22
Update Wiz.cn.xml

ivysrono:


jeremyn:
- [ ] Add `test` for https://get.wiz.cn or move it to `No working URL known`

jeremyn:
https://get.wiz.cn/wiznote-windows-x86-2018-01-22.exe is almost 40MB. Is there any smaller test you can find? Where did you find this?

ivysrono:
@jeremyn It's download url. I have updated comment about where to find them.

jeremyn:
I don't want to surprise someone with a 40MB test. I think it's okay in this case to use just the implicit test, even though it returns a 403. I've updated the checklist with a suggested replacement.

jeremyn:
Thanks, merged.

PR #14378, opened 2018-01-22
[Levo.com] Update targets

Bisaloo:


jeremyn:
- [ ] https://pages.levo.com/jetset/ has an HTTP redirect caused by a script. https://pages.levo.com/guides/ also has an HTTP redirect. Other `pages` URLs, for example https://pages.levo.com/for-companies/ and https://pages.levo.com/women-in-politics/ , seem okay. I think we should remove the `pages` `target` and the `securecookie` and move `pages` to the top under `HTTP redirect`. Maybe note it like this:
```xml
    HTTP redirect:
        pages.levo.com ( https://pages.levo.com/guides/ , https://pages.levo.com/jetset/ )
```

Bisaloo:
Should I leave the securecookie for `^` ?

jeremyn:
No, the `^` `securecookie` should be removed too. The `^` `securecookie` should be removed if there are some subdomains that don't support HTTPS.

This looks ready to me. Is it okay for me to merge?

Bisaloo:
yes, you can merge it.

Issue #14375, opened 2018-01-22
Using HTTPSEverywhereUserRules with WebExtensions


irykoon:
Type: ruleset issue

Dear HTTPSEverywhere Developers,

I tried to use HTTPSEverywhereUserRules in the latest Tor Browser according to [the instructions](https://github.com/EFForg/https-everywhere/issues/9958#issuecomment-321452120) by @Hainish : 

> To add custom rules, you can still add them to your HTTPSEverywhereUserRules path, and subsequently change the extensions.https_everywhere.webextensions-migrated pref to false so that the extension knows to run another migration.

However, it seems there is no such a preference in about:config in TBB by default.

Therefore, I opened the about:config in Firefox 57, finding a preference called `extension.https_everywhere.webextension-migrated` (which is different from what @Hainish said: `extensions.https_everywhere.webextensions-migrated`). 

To trigger a re-import of my custom rules, I did the following steps:
- put my custom ruleset to HTTPSEverywhereUserRules direcotry
- toggled the value `extension.https_everywhere.webextension-migrated` to false
- created a new pref called `extension.https_everywhere.webextensions-migrated` and set the value to false
- restarted the Firefox

However, it seems the rules I set in HTTPSEverywhereUserRules does not work.

For your information, the ruleset I used is Browser/profile.default/HTTPSEverywhereUserRules/WhonixOnion.xml . The content is as follows:

```
<ruleset name="Whonix Onion">
  <target host="whonix.org" />
  <target host="www.whonix.org" />
  <target host="phabricator.whonix.org" />
  <target host="forums.whonix.org" />
  <target host="download.whonix.org" />
  <target host="deb.whonix.org" />
  <rule from="^https?://(.*\.?)whonix\.org/" to="http://$1kkkkkkkkkk63ava6.onion/"/>
</ruleset>
```
I did notice that later, @Hainish said in #11977 that:
> Adding new rules to the HTTPSEverywhereUserRules directory will not work after the one-time migration in the Embedded WebExtension (>= 2017.8.15) is performed. If you had old rules, they were migrated to the new extension.
> To add new custom rules, you'll have to use the popup by visiting the https version of the site you want to add a rule for, clicking the extension icon, and clicking "Add a rule for this site".

Therefore, would you tell me if HTTPSEverywhereUserRules is still supported by WebExtensions version? If it is supported, would you please educate me on:
1. How I can use HTTPSEverywhereUserRules in Firefox 57?
[...]

PR #14385, opened 2018-01-22
Update document since HTTPSEverywhereUserRules is no longer supported

irykoon:
For details, see the issue here: https://github.com/EFForg/https-everywhere/issues/14375

Please review and improve it as needed.


jeremyn:
@irykoon Your pull request removes most of the content of the file @Hainish linked. Most of it is still useful though, since we still use ruleset files and they have the same format.

I'm not really clear what documentation update you want to do. Do you just want to tell people that `HTTPSEverywhereUserRules/` is no longer supported? You can just remove the paragraph that starts with "You should also manually test your ruleset by placing it in the `HTTPSEverywhereUserRules/` subdirectory...".

You could also replace that paragraph with a similar paragraph guiding people to the <kbd>Ctrl-Z</kbd> debugging page, although when the feature was first released, @Hainish argued against documenting that feature at all. I think it should be documented somewhere though, so we don't have to type out a walkthrough like I did in https://github.com/EFForg/https-everywhere/issues/14375#issuecomment-359449102 every time someone seems like they might want to know about it. Perhaps @Hainish could share his thoughts here.

Hainish:
@jeremyn I think it's worth documenting the `Ctrl-Z` debugging page _if_ it is clear that this should only be used for debugging purposes (just as we've added the warning at the top of the debugging page) instead of for normal usage.

irykoon:
Hi @jeremyn and @Hainish !

Thank you so much for offering me the feedback!

I have made the changes according to it.

jeremyn:
- [ ] I'd like to keep the diff size small and remove extraneous changes. Please re-add the space back to the end of the "We use an [automated checker]" paragraph, and the space after `</ruleset>` for the `Amazon (buggy)` `ruleset`. (To be clear, removing these spaces is better styling, I just want to simplify the diff.)

- [ ] There is no need to say "Previously you could use this approach that no longer exists." Just don't mention the other approach. I would replace your `Previously...Currently...` paragraphs with this:
```
You can test rulesets in the browser using a hidden debugging page, but please be aware that this approach should only be used for debugging purposes and should not be used for setting up personal custom rules. You can access the hidden debugging page this way:

* Firefox: `about:addons` > HTTPS Everywhere preferences > click under `General Settings` > press <kbd>Ctrl-Z</kbd>
* Chromium/Chrome: `chrome://extensions/` > HTTPS Everywhere options > click under `General Settings` > press <kbd>Ctrl-Z</kbd>

You might need to disable popup blocking for the page to appear. Once you have you loaded the page, you might find it convenient to bookmark it for later use.
```

I think the debug page is unavailable in the Tor Browser right now because the Tor people are making their own pre-WebExtensions version of the extension. Once Tor switches to use Firefox 59 ESR as their base, the debug page will probably be accessible. So we don't need to mention Tor here specifically for now, because the problem will fix itself in a few months. @Hainish what do you think?

- [ ] Delete the `#### [Rulesets](#rulesets)` section at the end, we don't need to talk about the UI in a document that is about writing rulesets.

irykoon:
Thank you, @jeremyn !

They are all done now!

jeremyn:
It looks like you added an extra space after the `RabbitMQ` `ruleset` instead of after the `Amazon (buggy)` `ruleset`. Could you please fix that? The idea is that there should only be changes in the `Testing` section.

Also I incorrectly had `you have you loaded` in my suggested text, which you put into your changes. It should just be `you have loaded`. I've fixed my suggested text, please update your changes too. I'm sorry for the trouble.

irykoon:
Hi @jeremyn !

Thank you for your careful review! I have fixed the typos according to your suggestions now!

jeremyn:
@irykoon Thanks, this looks good to me.

@Hainish Can you please review this and merge it if you agree with the changes? Note my question to you about Tor earlier at https://github.com/EFForg/https-everywhere/pull/14385#issuecomment-359855914.

Hainish:
@jeremyn thanks for reviewing, this looks good to me.

Tor Browser is actually building from the signed tags of our releases, and updates are delivered through the EFF self-hosted release channel.  The `Ctrl-Z` keystroke *does* work in Tor Browser.  Whether it's worth mentioning here or not is a toss-up.  I tend to think for the sake of brevity, it's nice to just list Firefox and Chrome, rather than Firefox, Chrome, Opera, and Tor Browser.

Thanks for this update @irykoon, I'll update the website from this ASAP.

jeremyn:
The debug page doesn't open for me in Tor Browser after pressing <kbd>Ctrl-Z</kbd> as of tor-browser-linux64-7.5_en-US.tar.xz which is the most current version available at https://www.torproject.org/projects/torbrowser.html.en#downloads . There's actually no `General Settings` area in the HTTPS Everywhere preferences section, but there is a checkbox with no label, and I've tried pressing <kbd>Ctrl-Z</kbd> after clicking the checkbox and around the checkbox.

Hainish:
@jeremyn odd, it opens for me, and I do see a `General Settings` area for `tor-browser-linux64-7.5_en-US.tar.xz`.  I'm confused as to why you're seeing a different result.

Shoot, I merged instead of squash-merge, my apologies.

jeremyn:
@Hainish After some testing, it seems related to the Torbutton security settings. If you open the HTTPS Everywhere preferences page while Torbutton is at the least secure/"Standard" level, there is a `General Settings` area and the debug page opens with <kbd>Ctrl-Z</kbd>. If Torbutton is at the middle ("Safer") or highest ("Safest") security levels, you get the broken behavior I described earlier.

This makes sense since the <kbd>Ctrl-Z</kbd> functionality is added by [`chromium/options.js`](https://github.com/EFForg/https-everywhere/blob/b70a51dea7a5e7c1bd3d916f6c690a43e8fd9ea6/chromium/options.js) when the preferences page is opened, and the higher security settings block JavaScript in some/all cases.

Hainish:
Interestingly, this same behavior is exhibited in Firefox 52 ESR.  Apparently, there is a fix for this as of Firefox 57, but they haven't bothered to backport it to ESR.

This reminds me of the popup fields not populating in Firefox 52 ESR but having been fixed in later versions.  The Tor Browser had to add an extra Firefox patch for this to work: https://trac.torproject.org/projects/tor/ticket/23258#comment:21

Since Mozilla hasn't backported this fix, I'll open a Tor Browser trac issue to take care of the options ui.

Hainish:
Ticket created: https://trac.torproject.org/projects/tor/ticket/25014

jeremyn:
Thanks @Hainish .

PR #14392, opened 2018-01-23
[Disqus.com] Increase coverage

Bisaloo:


src/chrome/content/rules/Disqus.xml
```diff
@@ -1,36 +1,42 @@
 <!--
 	Commenting system used by several large sites.
-		
-	Subdomains like SITENAME.disqus.com aren't covered in this ruleset.
-	Probably some other subdomains aren't covered either.
+
 -->
 <ruleset name="Disqus (partial)">
```
J0WI: I think we aren't `partial` anymore.

PR #14394, opened 2018-01-23
Fix #14382: Squarespace.com scripts not correctly excluded.

### LINKED ISSUE #14382: HTTPs everywhere breaks squarespace.com custom-domain sites
[//]: #
[//]: #
[//]: # (Delete all but one of the following "Type" lines.)
[//]: # (Leave in the line that best describes the issue you are reporting.)

Type: ruleset issue

[//]: # (If you are reporting a ruleset/website problem, include the top-level)
[//]: # (domain below. For example, if you want to report an issue about)
[//]: # ("one.example.com" and "two.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: example.com)
[//]: #
[//]: # (Be sure to remove the parenthesis and comment marker.  If you are only)
[//]: # (reporting an issue about "one.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: one.example.com) 
[//]: #
[//]: # (Only include one top-level domain. If you have more than one top-level)
[//]: # (domain to report, such as both "example.com" and "example.org", open a)
[//]: # (new issue for each top-level domain.)
[//]: #
[//]: # (If you are not reporting a ruleset/website problem, you can delete the)
[//]: # ("Domain" line or leave it blank.)

Domain: http://laicale.com/ (or any others which have http enabled)

The problem is badly configured CORS policy on the squarespace.com side which disallows customer's HTTP origins to fetch some javascript from squarespace.com HTTPs urls (which do work when directly visiting the HTTPs sites).

I have tried reporting this, but they said having this work is a "feature" and didn't promise anything.

image

hsanjuan:
See #14382


jeremyn:
Try `scripts-compressed/.+\.js$` for the regex instead.

hsanjuan:
@jeremyn done, i think

jeremyn:
Please make the indentation of your new `test`s even with the existing `test`s, using a single tab instead of spaces, and put them before the existing `test`s.

hsanjuan:
sorry @jeremyn. how about now?

Issue #13194, opened 2017-10-17
New ruleset for fujitsu.com


BK1233:
Type: new ruleset

Domain: fujitsu.com

pipboy96:
See https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Fujitsu.xml.

PR #14398, opened 2018-01-24
Update Fujitsu.xml

episkorowski:
#13194
Updated the existing Fujitsu ruleset and added `fujitsu.com` and `www.fujitsu.com` which weren't there for some reason.

- Renamed ruleset from `Fujitsu (partial)` to `Fujitsu`
- Modified old rules to use capturing groups in the regex.
- Removed targets with left wildcards.
- Added trivial rule, added new rule to cover `jp.fujitsu.com` and `www.jp.fujitsu.com` since old rule didn't work properly.
- Removed `img.jp.fujitsu.com`, added to `Invalid certificates`.
- Removed rule for `azby.fmworld.net`, added to `Redirects to HTTP` section

### DIFF
--- a/src/chrome/content/rules/Fujitsu.xml
+++ b/src/chrome/content/rules/Fujitsu.xml
@@ -24,7 +24,7 @@
 
 	<!-- Page normally redirects to https://www.fujitsu.com/jp/, does not
 	redirect properly with standard rule -->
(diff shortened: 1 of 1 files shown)


zoracon:
Re-triggering Travis

cschanaj:
Replaced #17970

PR #14401, opened 2018-01-24
Create SpringerNature.com.xml

Bisaloo:


src/chrome/content/rules/SpringerNature.com.xml
```diff
@@ -0,0 +1,59 @@
...
+		owa.springernature.com
+
+	Invalid certificate:
+		click.springernature.com
+		editorial-jobs.springernature.com
+		www.search.springernature.com
+		sndigital.springernature.com
+
+	Mixed content:
+		group.springernature.com
```
J0WI: Is there a specific issue? I wasn't able to find one and the site defaults to https for me.

Bisaloo: If I don't specify a test url, it's usually on `$`. It seems indeed gone in that case.

Bisaloo:
`id.springernature.com` fails over both HTTP and HTTPS for me.

J0WI:
Try on `http://id.springernature.com/adfs/ls/`

Bisaloo:
Still timing out. Should I add it as a target?

J0WI:
Seems to be HTTPS only.

J0WI:
Sorry, I didn't check the cert chain of `id.springernature.com`

PR #14409, opened 2018-01-24
Added ruleset for deutsche-rohstoffagentur.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/Deutsche-Rohstoffagentur.de.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Deutsche-Rohstoffagentur.de">
+
+	<target host="deutsche-rohstoffagentur.de" />
+	<target host="www.deutsche-rohstoffagentur.de" />
+
+	<!-- Invalid certificate on https://deutsche-rohstoffagentur.de/,
+	http://deutsche-rohstoffagentur.de/ redirects to http://www.deutsche-rohstoffagentur.de/ -->
```
Bisaloo: Just put this in the top comment, using the usual format. We also don't need to say that `^` redirects to `www` over HTTP. We would not add a rule if that was not the case.

PR #14415, opened 2018-01-24
Added ruleset for gpdn.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/GPDN.de.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="GPDN.de">
+
+	<target host="gpdn.de" />
+	<target host="www.gpdn.de" />
+	<target host="cardo.gpdn.de" />
```
Bisaloo: This subdomain does not seem to be working anymore.

Issue #14290, opened 2018-01-14
Test if test url is covered by any target


J0WI:
Test urls with hosts not related to any target are currently allowed in rulesets.
The following scenario should should fail in ruleset tests:
```
<ruleset name="example">
	<target host="target.example" />

	<test url="http://no-target.example/" />

	<rule from="^http:" to="https:" />
</ruleset>
```

Type: other

jeremyn:
This might be challenging to check completely, for example for a wildcard `target` with a very complex `rule`.

Have you seen rulesets like this in the wild?

J0WI:
One example has been fixed here: #14177

cschanaj:
Submitted a PR in #14293 (not really sure if I have the implementation correct.. though)

@Hainish there is something was with `chrome-beta` and `chrome-stable` since https://github.com/EFForg/https-everywhere/commit/d36ea408818d865966b8cd632b065a1f1b5a88fc. Would you mind to take a look?

Hainish:
@cschanaj looking now

Hainish:
@cschanaj this should be fixed with 7f4fe7c8547aa7dabb0762c268b8b67857b91182, thanks for alerting me to this.

PR #14417, opened 2018-01-24
[MIT.xml] fix #14397

### LINKED ISSUE #14397: MIT personal websites are redirected to an incorrect URL
Type: ruleset issue
Domain: mit.edu

The ruleset in MIT.xml says this:

```xml
<rule from="^http://(www\.)?mit\.edu/" to="https://web.mit.edu/" />
```

This translates e.g. http://www.mit.edu/~cpitcla/test (which works) into https://web.mit.edu/~cpitcla/test (404).  Working HTTPS URLs for that page include https://www.mit.edu/~cpitcla/test (no need to rewrite the `www` part into `web`) and https://web.mit.edu/cpitcla/www/test (no `~`, extra `www`).

This is documented here: http://kb.mit.edu/confluence/pages/viewpage.action?pageId=3907090 .

It looks like the issue was introduced by https://github.com/EFForg/https-everywhere/pull/14169 or https://github.com/EFForg/https-everywhere/pull/12002 maybe?

```diff
+	<rule from="^http://(www\.)?mit\.edu/" to="https://web.mit.edu/" />
```

J0WI:


jeremyn:
I guess you can just wrap the three `test`s in a comment, like this:

```xml
		<!-- should not get secured
		<test url="http://mit.edu/~puzzle/" />
		<test url="http://www.mit.edu/~puzzle/" />
		<test url="http://web.mit.edu/puzzle/www/" />
		-->
```

We don't really have anything like a "negative" test though.

jeremyn:
https://github.com/EFForg/https-everywhere/pull/14417/commits/7c10f9b77c4926bc35b7643db7f69d436b1a468d is fine with me.

J0WI:
Sure, we fixed this in #14290...

varenc:
@J0WI @jeremyn this update seems to introduce a new problem.  Now http://web.mit.edu is rewritten to https://www.mit.edu which will fail for most people.  (but confusingly not for MIT students/staff)

MIT has a rather unique situation where the HTTPS version of its main web domains (web.mit.edu, www.mit.edu) can only be accessed by MIT affiliates with an MIT issued client cert.

Looking at this rule, it really seems like it should only match URLs that that start their path with "/files/" but from the debug logs I can confirm that this new rule is somehow matching `http://web.mit.edu`.  Logs:
image

even so, I think this rule should be deleted entirely because https on www.mit.edu and web.mit.edu is not working at all for the general public.  If you're an MIT affiliate trying to debug this, be sure to delete you client certs before testing.  (they're easy to get again at ca.mit.edu)

J0WI:
Thanks, I've filled #16236

PR #14419, opened 2018-01-24
add rule for Interia.pl

numismatika:


J0WI:
Please add some test urls.

PR #14424, opened 2018-01-25
Delete du00.cc.xml

ivysrono:


jeremyn:
- [ ] Let's just mark this disabled for now, and wait to delete it entirely in case they fix the problem.

```xml
<ruleset name="du00.cc" default_off="refused">
...
</ruleset>
```

ivysrono:
#14458

PR #14430, opened 2018-01-25
Simplify and Update Cloudinary.xml

cschanaj:


src/chrome/content/rules/Cloudinary.xml
```diff
@@ -1,101 +1,35 @@
...
-
+	Non-functional hosts:
+		SSL peer certificate was not OK
+		- a1.res.cloudinary.com
+		- a2.res.cloudinary.com
+		- a3.res.cloudinary.com
+		- a4.res.cloudinary.com
+		- a5.res.cloudinary.com
 -->
 <ruleset name="Cloudinary.com (partial)">
```
J0WI: Let's also get rid of the `partial`

cschanaj:
@J0WI done. thanks!

PR #14432, opened 2018-01-25
requesting merge of 4 new rulessets

ohdns:


Folant:
@ohdns 
All of these sites are [HSTS-preloaded](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#hsts-preloaded-rules).

ohdns:
@Folant yes I was adding them for any case where HSTS may not be used but your addon is used.  Perhaps not required.  Feel free to discard changes.

PR #14436, opened 2018-01-25
[BilletNet.dk] Move targets to their own ruleset

Bisaloo:


src/chrome/content/rules/BilletNet.dk.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Related ruleset: Ticketmaster.xml
+
+	Refused:
+		billetnet.dk
+
+	Invalid certificate:
+		ams3pxy.billetnet.dk
+		kundeanmeldelser.billetnet.dk
+		lca4pxy.billetnet.dk
```
J0WI: `wmr.billetnet.dk`

PR #14443, opened 2018-01-26
[2o7.net] Add test urls

Bisaloo:


src/chrome/content/rules/2o7.net.xml
```diff
@@ -1,22 +1,18 @@
...
+		<test url="http://cewecolor.112.2o7.net/optout.html" />
+		<test url="http://truckstore.122.2o7.net/optout.html" />
 
 
-	<securecookie host="^(?:.*\.)?2o7\.net$" name=".+" />
-
+	<securecookie host=".+" name=".+" />
 
 	<rule from="^http://([^/:@\.]+)(\.1[12]2|\.sbx1)?\.2o7\.net/"
 		to="https://$1$2.2o7.net/" />
```
J0WI: Can we simplify this somehow or is there a known issue?

Bisaloo: I'm not sure. See my edit for a better but still safe solution.

Issue #14336, opened 2018-01-19
Release 2018.1.24 forthcoming


Hainish:
Type: other

cc @jsha @pde @J0WI @diracdeltas @jeremyn @gloomy-ghost @brainwane @wonderchook

Hainish:
I've been unable to make a release this week, deferring to early next week.

Hainish:
Done

PR #14446, opened 2018-01-26
Update TIP.net.au.xml

cschanaj:
#14336 @J0WI Would you mind to merge this to fix the mistake from ae37e50571d?


J0WI:
Thanks for the correction!

PR #14454, opened 2018-01-27
[Stack-Exchange] Exclude problematic path

### LINKED ISSUE #14452: HTTPS Everywhere still breaks area51.stackexchange.com login
Type: ruleset issue

Domain: area51.stackexchange.com

See https://area51.meta.stackexchange.com/questions/13597/area-51-openid-login-is-broken-with-https/27726#27726 for more details, but it looks like the [Stack Exchange ruleset](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Stack-Exchange.xml) should still have an exclusion for `http://area51.stackexchange.com/users/authenticate/`, something like this:

    
    <exclusion pattern="^http://area51\.stackexchange\.com/users/authenticate/" />

Area 51 runs a distinct and somewhat obsolescent fork of the Stack Exchange codebase, and apparently this problem (which used to be more general, see issue #58) still occurs there.

Ps. Also, while investigating this, I noticed that [the ruleset syntax documentation](https://github.com/EFForg/https-everywhere/blob/19468262c7a01d776371eaa7086c9f4df04099b3/docs/en_US/rulesets.md#exclusions) happens to feature as an example a very similar but broader exclusion that would've caught this. However, that exclusion was apparently removed from the actual ruleset in commit 783d040.

Pps. This may have been previously reported as issue #7275, but that report was a bit lacking in details.

Bisaloo:
fix #14452

### DIFF
--- a/src/chrome/content/rules/Stack-Exchange.xml
+++ b/src/chrome/content/rules/Stack-Exchange.xml
@@ -54,6 +54,7 @@
 		-->
 		<exclusion pattern="^http://area51\.stackexchange\.com/users/authenticate/\?s=" />
 		<test url="http://area51.stackexchange.com/users/authenticate/?s=" />
+		<test url="http://area51.stackexchange.com/users/authenticate/?s=11111111-1111-1111-1111-111111111111" />
 
 	<!-- Stackoverflow -->
 	<target host="stackoverflow.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Stack-Exchange.xml
```diff
@@ -48,6 +48,14 @@
 		<exclusion pattern="^http://blog\.gaming\.stackexchange\.com"/>
 		<test url="http://blog.gaming.stackexchange.com/" />
 
+		<!-- this causes login issues on area51. See:
+			- https://area51.meta.stackexchange.com/questions/13597/area-51-openid-login-is-broken-with-https/27726
```
J0WI: Could you also add https://meta.stackexchange.com/questions/292058/network-wide-https-its-time/306308#306308?

PR #14456, opened 2018-01-27
Delete Eloqua-clients.xml 

Bisaloo:


src/chrome/content/rules/Brightcove.xml
```diff
@@ -103,6 +103,7 @@
 <ruleset name="Brightcove (partial)">
 
 	<target host="*.brightcove.com" />
+	<target host="go.brightcove.com" />
```
J0WI: That's already covered by the wildcard. But this rule can be reqritten without using a wildcard anyway (if you want to create a new PR for this ;)).

Bisaloo: I intended that as a first step to remove the wildcard but perhaps that's not so useful. Should I remove this target?

J0WI: Yes, duplication is worse.

PR #14457, opened 2018-01-27
Clean up AmazonWebServices.xml

Bisaloo:
This is going to be a difficult review but I broke it down into multiple simple commits. It should be okay if you review commit by commit.

I can create a separate PR for each commit if you prefer.

#14293


src/chrome/content/rules/Amazon.xml
```diff
@@ -98,6 +110,11 @@
 	<securecookie host=".+" name="^(?:aps-trtmnt$|s_v)" />
 	<securecookie host="^\w" name=".+" />
 
+	<rule from="^http://g-images\.amazon\.com/"
+		to="https://d1ge0kk1l5kms0.cloudfront.net/" />
```
J0WI: We should add a test url for it.

utils/ruleset-whitelist.csv
```diff
@@ -3844,3 +3844,4 @@ f76e810891c9301354123b801ace36ba307fff5ae323fb9bb608d8c2757efb36,1,0,Zissousecur
 0a43f73ea001dcef9e970232e3d8d3123f0a4eb163e1d69b23498bff60a735c9,1,1,Zopim.xml
 6750e29d302751af08e3cbe82647eb8a652fba4be1efbed4bb63473ec6ed40c2,1,1,Zorpia.xml
 54f30ba3843ce2be5355b1259c28f6ffe7b3c9feddcbe0de95025bc506e913cf,1,0,Zulius.com.xml
+fbda0e5c6d6b44ff948bcd20faad126f55d5f34e89212d74b56e1af1758ae725,1,0,Images-Amazon.com.xml
```
J0WI: This shouldn't be necessary. What was the error there?

src/chrome/content/rules/Images-Amazon.com.xml
```diff
@@ -0,0 +1,58 @@
...
+	<target host="ec3.images-amazon.com" />
+	<target host="ec4.images-amazon.com" />
+	<target host="ec5.images-amazon.com" />
+	<target host="ec6.images-amazon.com" />
+	<target host="ec8.images-amazon.com" />
+	<target host="g-ecx.images-amazon.com" />
+	<target host="z-ak.images-amazon.com" />
+
+	<!-- Zeitgeist MP3 player https://trac.torproject.org/projects/tor/ticket/7594 -->
+	<exclusion pattern="^http://g-ecx\.images-amazon\.com/.*(?:[mM][pP]3|zeitgeist|digital/music)" />
```
J0WI: Can we get rid of the non-capturing group here?

J0WI: It seems that this regex will match more or less everything :/
Are there any up to date examples to verify this bug? Most Amazon services are now HTTPS by default.

src/chrome/content/rules/Images-Amazon.com.xml
```diff
@@ -0,0 +1,58 @@
...
+									-->
+		<exclusion pattern="^http://([\w-]+).images-amazon.com/images/G/01/zeitgeist/mp3player" />
+			<test url="http://ec2.images-amazon.com/images/G/01/zeitgeist/mp3player/swf/zgMp3Player-1.0._V212274098_.swf" />
+			<test url="http://ec1.images-amazon.com/images/G/01/zeitgeist/mp3player/swf/zgMp3Player-1.0._V212274098_.swf" />
+
+	<rule from="^http://g-ecx\.images-amazon\.com/"
+		to="https://d1ge0kk1l5kms0.cloudfront.net/" />
+
+	<rule from="^http://(ec[1-8]|z-ak)\.images-amazon\.com/"
+		to="https://images-na.ssl-images-amazon.com/" />
```
J0WI: I'm not happy with forcing everything to the North America CDN, but I have not yet found a better solution.

Bisaloo:
"conflicting" was a poor choice of words but the rule removed in https://github.com/EFForg/https-everywhere/commit/62903de78cf4d4ddab10af022de3f3d1ba0b7296 was redundant with 
https://github.com/EFForg/https-everywhere/blob/62903de78cf4d4ddab10af022de3f3d1ba0b7296/src/chrome/content/rules/AmazonAWS.com.xml#L159-L160

J0WI:
LGTM apart from https://github.com/EFForg/https-everywhere/pull/14457#discussion_r165058206

src/chrome/content/rules/Images-Amazon.com.xml
Bisaloo: After reviewing the Tor tracker ticket, it looks that the previously broken functionality no longer exists.

src/chrome/content/rules/Images-Amazon.com.xml
```diff
@@ -0,0 +1,37 @@
...
+	<target host="ec3.images-amazon.com" />
+	<target host="ec4.images-amazon.com" />
+	<target host="ec5.images-amazon.com" />
+	<target host="ec6.images-amazon.com" />
+	<target host="ec8.images-amazon.com" />
+	<target host="g-ecx.images-amazon.com" />
+	<target host="z-ak.images-amazon.com" />
+
+	<rule from="^http://g-ecx\.images-amazon\.com/"
+		to="https://d1ge0kk1l5kms0.cloudfront.net/" />
```
J0WI: Sorry, I forgot that one: Could you add test url for it?

Bisaloo:
Fixed conflicts from https://github.com/EFForg/https-everywhere/pull/14521.

J0WI:
I think `ws.amazon.co.uk` was a temporary issue. Can you retry?

PR #14460, opened 2018-01-28
Create laibafile.cn.xml

ivysrono:


src/chrome/content/rules/laibafile.cn.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	images for bbs.tianya.cn , one of the biggest Chinese bbs.
```
J0WI: Please add a comment for `^` and `www`.

src/chrome/content/rules/laibafile.cn.xml
```diff
@@ -0,0 +1,17 @@
...
+-->
+<ruleset name="laibafile.cn">
+	<target host="img1.laibafile.cn" />
+	<target host="img2.laibafile.cn" />
+	<target host="img3.laibafile.cn" />
+
+	<rule from="^http://img(1|2)\.laibafile\.cn/"
+		to="https://img3.laibafile.cn/" />
```
J0WI: Please document why this special rule is required.

PR #14468, opened 2018-01-29
Added ruleset for myfritz.net

Jochen-A-Fuerbacher:


src/chrome/content/rules/MyFRITZ.net.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="MyFRITZ.net (partial)">
```
J0WI: Add a hint about customer controlled subdomains.

PR #14496, opened 2018-01-31
Add ruleset for gocontigo.com

Injust:


src/chrome/content/rules/Gocontigo.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Gocontigo.com">
```
J0WI: See also non-functional `sip.gocontigo.com`.

Injust: Added, thank you.

J0WI:
Thanks!

PR #14500, opened 2018-01-31
Added ruleset for artfiles.org

Jochen-A-Fuerbacher:


src/chrome/content/rules/artfiles.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="artfiles.org (partial)">
```
J0WI: It should be noted that they use a wildcard DNS record.

J0WI:
Thanks!

PR #14507, opened 2018-02-01
Split RCI.com.xml from Wyndham.xml

cschanaj:


src/chrome/content/rules/RCI.com.xml
```diff
@@ -0,0 +1,30 @@
...
+	<target host="rci.com" />
+	<target host="www.rci.com" />
+	<target host="ai.rci.com" />
+	<target host="api.rci.com" />
+	<target host="arrivalguides.rci.com" />
+	<target host="b2b.rci.com" />
+	<target host="discover.rci.com" />
+	<target host="m.rci.com" />
+	<exclusion pattern="^http://m\.rci\.com/$" />
+		<test url="http://m.rci.com/rcimobile/" />
```
J0WI: I think `$` can be redirected to `/rcimobile`. Please also add document the issue.

PR #10875, opened 2017-07-05
Add ruleset for Hive.co.uk.xml

sampablokuper:

PR #14520, opened 2018-02-02
Add note to ruleset creators

sampablokuper:
per invitation from @jeremyn at
https://github.com/EFForg/https-everywhere/pull/10875#issuecomment-362609979


zoracon:
Thank you!

PR #14528, opened 2018-02-04
Ruleset for Pushexp.com, a digital music dist platform

aladar42:
Pushexp.com is used by many major labels (like Universal and Warner Music) to distribute digital music from vouchers in physical releases. It requires the user to sign up for an account, but currently the site is http by default. vinylcollector.pushexp.com is the most important domain, as it's the one probably hundreds of releases refer to. This may be a messy one, but I wasn't sure how else to make the rule, yet I think it's important to be added.

There are also several domains with mixed content:
https://metals.pushexp.com/welcome/
https://achtung.pushexp.com/welcome/
https://decembersong.pushexp.com/welcome/


pipboy96:
Sorry, this can't be merged as is. I will create a corrected PR in a moment.

PR #14539, opened 2018-02-06
[ScientificLinux.org] Simplify ruleset

Bisaloo:


src/chrome/content/rules/Scientificlinux.org.xml
```diff
@@ -1,20 +1,15 @@
 <!--
-	^scientificlinux.org: Mismatched
+	Handshake failure:
+		content.scientificlinux.org
+
+	Invalid certificate:
+		ww.scientificlinux.org
```
J0WI: See also:
```
ftp
ftp1
ftp2

PR #14560, opened 2018-02-07
[RStudio.com] Add targets

Bisaloo:


J0WI:
This seems to be related to `RStudio.com.xml`. I think you can directly add it there.

J0WI:
Thanks!

PR #14561, opened 2018-02-07
Delete Thing.net-falsemixed.xml

Bisaloo:


src/chrome/content/rules/Thing.net.xml
```diff
@@ -1,37 +1,6 @@
...
-			- post from $self ¹
-			- post from www.jamesfuentes.com ²
-
-	¹ Secured by us
-	² Unsecurable <= interrupted
+	Invalid certificate:
+		thingist.thing.net
 
 -->
 <ruleset name="thing.net (partial)">
```
J0WI: `thing.net` lists some more hosts that can be added.

PR #14563, opened 2018-02-07
[CDON] Remove cdon.se subdomains

Bisaloo:
cdon.se subdomains are handled in [`cdon.se.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/cdon.se.xml).


J0WI:
Could you please add a comment about the relationship?

PR #14569, opened 2018-02-08
Added rulesets for sonarqube.org, sonarsource.com, sonarlint.org, son…

Jochen-A-Fuerbacher:
…arcloud.io


src/chrome/content/rules/SonarSource.com.xml
```diff
@@ -0,0 +1,36 @@
+<!--
+	See also:
+		SonarCloud.io.xml
+		SonarLint.org.xml
+		SonarQube.org.xml
```
J0WI: Please only include a complete list in one file and always refer to the same list.

PR #14570, opened 2018-02-08
Clean up test execution flow

Hainish:


Hainish:
cc @jeremyn @J0WI @Bisaloo @cschanaj Just an FYI: I've cleaned up the tests here.  Mainly, I've renamed the `rules` test to the `validations` test, and separated out `test/validations.sh` so that the `validations` test in `test/run_travis.sh` just calls out to that.  It is now called from `test.sh`, so contributors will be running all the validations and not just the test coverage validation when they run `./test.sh`.

I've also moved much of the testing infrastructure into `tests/` and created a `tests/validations/` path, where all the validation tests live.  These were previously under `utils/` which I thought was inappropriate since they're used mainly for tests.  `utils/` I consider something that, when run, creates commitable output.  This is why I've retained `remove-obsolete-references.sh` in `utils/` - it serves both purposes, but it can be run separate from the testing infrastructure.

jeremyn:
FYI to anyone wondering: `fetch-test.sh` is now `test/manual.sh`. So instead of:
```
$ ./fetch-test.sh rules/Freerangekitten.com.xml
```

You can run:
```
$ test/manual.sh rules/Freerangekitten.com.xml
```

`fetch-test.sh` is still mentioned at [`docs/en_US/development.md#testing-and-contributing-changes-to-the-source-code`](https://github.com/EFForg/https-everywhere/blob/0c7b699d0ca1bed52737c6b0b0df2c3b5b75773a/docs/en_US/development.md#testing-and-contributing-changes-to-the-source-code).

PR #14576, opened 2018-02-09
Create CarbonDoomsday.com.xml

Bisaloo:


src/chrome/content/rules/CarbonDoomsday.com.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Time out:
+		carbondoomsday.com
```
J0WI: WFM

J0WI:
Timeout on `www`.

Bisaloo:
Contacted them. Let's see if they want to fix it.

PR #13887, opened 2017-12-05
Add section #pull-requests-from-deleted-accounts

jeremyn:
Ping @brainwane @gloomy-ghost @Hainish @J0WI @wonderchook .

PR #14580, opened 2018-02-10
Add section about ghost pull requests.

sampablokuper:
Fixes #13887.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -375,3 +375,8 @@ and choose "Translate now" to enter the translation interface.
 
 * * *
 
+# Ghost pull requests
+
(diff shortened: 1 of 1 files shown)


Hainish:
@sampablokuper my text in https://github.com/EFForg/https-everywhere/pull/13887#issuecomment-363931406 was pretty carefully crafted.  Any reason for the change here?

sampablokuper:
@hainish, yes. Your text is OK, but IMO not ideal:

- Your text implicitly raised, but *did not answer*, the question, "Don't maintainers reserve the right to close PRs regardless?"
- The first sentence of your second paragraph seemed superfluous. Having already explained, by that point, what @ghost PRs are and how they arise, you should not need to also tell readers how to avoid creating them, because it is implicit in what you have already said.

My wording attempts to address those shortcomings.

Hainish:
@sampablokuper I do prefer my wording, making specific why it is inconvenient for maintainers.  I've committed this in 376c84a68df2d41023976dfa12e0fdfafec1eeac.  Thanks for this though!

PR #14585, opened 2018-02-10
Update test/run_travis.sh, Fix Travis fetch-test

cschanaj:
#14586 should not pass the fetch-test. IMHO, we should assume all tests are running from `${toplevel}` and use relative path everywhere.

cc @Hainish @jeremyn @J0WI `preload` and `fetch-test` is not running.


J0WI:
@Hainish is this a regression from #14570?

Hainish:
@cschanaj good catch!  Yes, looks like a regression.  Merging now.

PR #14587, opened 2018-02-10
Extend duplicates check to rulesets with different platforms

### LINKED ISSUE #11440: Include rulesets with different platforms in duplicate checks

Bisaloo:
See https://github.com/EFForg/https-everywhere/issues/11440 for context (fix #11440).

I would like to get this merged when possible because @cschanaj [statistics](https://github.com/EFForg/https-everywhere/wiki/Ruleset-Statistics) show that duplicate targets count increases (even though I regularly remove entries) and I don't think this is really intended. It is probably more of an oversight from contributors.

I am not super happy with the current clean up script but after giving it some thought, I don't think it is worth fixing it because I would like to see this whitelist totally removed at some point.

### DIFF
--- a/test/validations/special/duplicate-whitelist-cleanup.sh
+++ b/test/validations/special/duplicate-whitelist-cleanup.sh
@@ -4,7 +4,7 @@ TMPFILE=`mktemp /tmp/buffer.XXXXXXXX`
 trap 'rm "$TMPFILE"' EXIT
 
 for host in `cat test/validations/special/duplicate-whitelist.txt`; do
-    REPEATS=`grep -F "target host=\"$host\"" src/chrome/content/rules/*.xml | wc -l`
+    REPEATS=`egrep -l "<target\s+host=\s*\"$host\"\s*/>" src/chrome/content/rules/*.xml | wc -l`
     if [ $REPEATS -gt 1 ]; then
         echo $host
     fi
--- a/test/validations/special/duplicate-whitelist.txt
+++ b/test/validations/special/duplicate-whitelist.txt
@@ -9,7 +9,6 @@ www1.arun.gov.uk
 mailing.channel4.com
 training.citrix.com
 consumerreports.org
-c-spanvideo.org
 dell.com
 www.dell.com
 investors.demandware.com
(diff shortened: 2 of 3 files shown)


Bisaloo:
@Hainish, I can take care of python3.6 transition for https://github.com/EFForg/https-everywhere/blob/master/test/validations/special/run.py in this PR if you wish. It would save us a git conflict.

Hainish:
@Bisaloo I think this would be useful.  Converted the validations to python 3.6 in https://github.com/EFForg/https-everywhere/pull/14574, if you want to rebase from that.

Bisaloo:
@Hainish, this should be ready for review

Bisaloo:
@Hainish, it would be great if you could review this when you have some time. It's a pretty simple modification to the current process.

Hainish:
@Bisaloo thanks - superseded by https://github.com/EFForg/https-everywhere/pull/14861, just to pass CI with extra duplicates in the whitelist.

PR #14598, opened 2018-02-11
Update https_everywhere_checker/http_client.py, Fix #14365

### LINKED ISSUE #14365: fetch-test.sh incorrectly reports an error because it doesn't follow a non-trivial rewrite rule

cschanaj:
Travis failed to follow a non-trivial rewrite because the redirect location lack a trailing `/`. A testing PR is available at #14599. Close #14365. Close #14599 

P.S. This bug should not affect the extension itself.

cc @Hainish @jeremyn

### DIFF
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -195,37 +195,36 @@ class HTTPFetcher(object):
 		return urlparse.urlunparse(parts)
 		
 	def absolutizeUrl(self, base, url):
-		"""Returns absolutized URL in respect to base URL as per
-		RFC 3986. If url is already absolute (with scheme), return url.
+		"""
+		Construct a full ("absolute") URL by combining a "base URL" (base)
+		with another URL (url) as per RFC 3986.
 		
 		@param base: base URL of original document
 		@param url: URL to be resolved against base URL
 		"""
-		#urljoin fails for some of the abnormal examples in section 5.4.2
-		#of RFC 3986 if there are too many ./ or ../
-		#See http://bugs.python.org/issue3647
-		resolved = urlparse.urljoin(base, url)
-		resolvedParsed = urlparse.urlparse(resolved)
-		path = resolvedParsed.path
-		
-		#covers corner cases like "g:h" relative URL
-		if path == "" or not path.startswith("/"):
-			return resolved
-		
-		#strip any leading ./ or ../
-		pathParts = path[1:].split("/")
-		while len(pathParts) > 0 and pathParts[0] in (".", ".."):
-			pathParts = pathParts[1:]
-		
-		if len(pathParts) > 0:
-			newPath = "/" + "/".join(pathParts)
-		else:
-			newPath = "/"
-			
-		#replace old path and unparse into URL
-		urlParts = resolvedParsed[0:2] + (newPath,) + resolvedParsed[3:6]
-		newUrl = urlparse.urlunparse(urlParts)
-		
+
+		# urljoin fails for some of the abnormal examples in section 5.4.2
+		# of RFC 3986 if there are too many ./ or ../
+		# See https://bugs.python.org/issue3647
+		joinedUrl = urlparse.urljoin(base, url)
+		joinedUrlParts = urlparse.urlparse(joinedUrl)
+
+		# Strip any leading ./ and ../
+		path = joinedUrlParts.path
+		if path[:1] == '/':
+			segments = path.split('/')
+			while '.' in segments:
+				segments.remove('.')
+			while '..' in segments:
+				segments.remove('..')
+			joinedUrlParts = joinedUrlParts._replace(path='/'.join(segments))
+
+		# Non-trivial rewrites do not work without a trailing '/'
+		# See https://github.com/EFForg/https-everywhere/issues/14365
+		if path == "":
+			joinedUrlParts = joinedUrlParts._replace(path='/')
+
+		newUrl = urlparse.urlunparse(joinedUrlParts)
 		return newUrl
 		
 	@staticmethod
--- a/test/rules/src/https_everywhere_checker/http_client.py
+++ b/test/rules/src/https_everywhere_checker/http_client.py
@@ -217,15 +217,22 @@ class HTTPFetcher(object):
 				segments.remove('.')
 			while '..' in segments:
 				segments.remove('..')
-			joinedUrlParts = joinedUrlParts._replace(path='/'.join(segments))
+			path = '/'.join(segments)
 
 		# Non-trivial rewrites do not work without a trailing '/'
 		# See https://github.com/EFForg/https-everywhere/issues/14365
-		if path == "":
-			joinedUrlParts = joinedUrlParts._replace(path='/')
+		if path == '':
+			path = '/'
 
-		newUrl = urlparse.urlunparse(joinedUrlParts)
-		return newUrl
+		joinedUrlParts = urlparse.ParseResult(
+			joinedUrlParts.scheme,
+			joinedUrlParts.netloc,
+			path,
+			joinedUrlParts.params,
+			joinedUrlParts.query,
+			joinedUrlParts.fragment,
+		)
+		return joinedUrlParts.geturl()
 		
 	@staticmethod
 	def _doFetch(url, options, platformPath):


jeremyn:
@cschanaj Thank you for looking my issue https://github.com/EFForg/https-everywhere/issues/14365. I can confirm that your changes let my minimal test case pass.

Your code chnages the logic in ways that are not entirely obvious, for example I'm not sure whether your updated version handles the "g:h" "corner case" that the original code mentions in a comment. Also your `joinedUrlParts._replace(...)` call replies on a non-public `_replace` method, which is itself associated with the `namedtuple` class that is also an implementation detail.

Have you identified exactly where the redirect logic fails without a trailing `/`? I suppose it is somewhere in the `fetchUrl` function. It might be better to fix the problem there, if possible. Alternatively, how about not modifying `absolutizeUrl`, but just conditionally appending a trailing `/` after the call to `absolutizeUrl`? This change added [here](https://github.com/EFForg/https-everywhere/blob/af5837376b0c8efc8b89c0006fd0c633038bbf85/test/rules/src/https_everywhere_checker/http_client.py#L387-L388) makes my test case pass:

```python
location = self.absolutizeUrl(newUrl, location) 
if location[-1] != '/':                         
    location += '/'
logging.debug("Following redirect %s => %s", newUrl, location)
```

jeremyn:
My previous comment may also not account for the "g:h" "corner case". I guess that imagines redirecting to something like `mailto:fake@example.com`. Anyway to complete my suggestion, we could modify the return value of `absolutizeUrl` to add a boolean to say whether to (possibly) append the `/` or not, for example replace `return newUrl` with `return (newUrl, True)` and similarly for `return resolved`.

Hainish:
@jeremyn I believe the `g:h` corner case refers to the first example here: https://www.ics.uci.edu/~fielding/url/test1.html.  This edge case doesn't seem to be accounted for in either case, though:

```python
>>> def absolutizeUrl(base, url):
...    """Returns absolutized URL in respect to base URL as per
...    RFC 3986. If url is already absolute (with scheme), return url.
...    
...    @param base: base URL of original document
...    @param url: URL to be resolved against base URL
...    """
...    #urljoin fails for some of the abnormal examples in section 5.4.2
...    #of RFC 3986 if there are too many ./ or ../
...    #See http://bugs.python.org/issue3647
...    resolved = urlparse.urljoin(base, url)
...    resolvedParsed = urlparse.urlparse(resolved)
...    path = resolvedParsed.path
...    
...    #covers corner cases like "g:h" relative URL
...    if path == "" or not path.startswith("/"):
...            return resolved
...    
...    #strip any leading ./ or ../
...    pathParts = path[1:].split("/")
...    while len(pathParts) > 0 and pathParts[0] in (".", ".."):
...            pathParts = pathParts[1:]
...    
...    if len(pathParts) > 0:
...            newPath = "/" + "/".join(pathParts)
...    else:
...            newPath = "/"
...            
...    #replace old path and unparse into URL
...    urlParts = resolvedParsed[0:2] + (newPath,) + resolvedParsed[3:6]
...    newUrl = urlparse.urlunparse(urlParts)
... 
>>> import urlparse
>>> absolutizeUrl('http://example.com', 'g:h')
'g:h'
```

```python
>>> def absolutizeUrl(base, url):                                                                                                                     
...     """                                                                  
...     Construct a full ("absolute") URL by combining a "base URL" (base)   
...     with another URL (url) as per RFC 3986.                              
...                                                                          
...     @param base: base URL of original document                           
...     @param url: URL to be resolved against base URL                      
...     """                                                                  
...                                                                          
...     # urljoin fails for some of the abnormal examples in section 5.4.2   
...     # of RFC 3986 if there are too many ./ or ../                        
...     # See https://bugs.python.org/issue3647                              
...     joinedUrl = urlparse.urljoin(base, url)                              
...     joinedUrlParts = urlparse.urlparse(joinedUrl)                        
...                                                                          
...     # Strip any leading ./ and ../                                       
...     path = joinedUrlParts.path                                           
...     if path[:1] == '/':                                                  
...         segments = path.split('/')                                       
...         while '.' in segments:                                           
...             segments.remove('.')                                         
...         while '..' in segments:                                          
...             segments.remove('..')                                        
...         joinedUrlParts = joinedUrlParts._replace(path='/'.join(segments))
...                                                                          
...     # Non-trivial rewrites do not work without a trailing '/'            
...     # See https://github.com/EFForg/https-everywhere/issues/14365        
...     if path == "":                                                       
...         joinedUrlParts = joinedUrlParts._replace(path='/')               
...                                                                          
...     newUrl = urlparse.urlunparse(joinedUrlParts)                         
...     return newUrl                                                        
... 
>>> absolutizeUrl('http://example.com', 'g:h')                                                                                                        
'g:h'
```

The logic for URL modifications belongs in the `absolutizeUrl` method, not outside of it.  It seems like this more closely mirrors the redirect logic of the browser.  I originally shared your aversion to `joinedUrlParts._replace(...)` as well, but the documentation is littered with examples using this: https://docs.python.org/2/library/collections.html#collections.somenamedtuple._replace.

jeremyn:
It's an implementation detail that the return value from `urlparse` is a `namedtuple` with a `_replace` method. It requires some digging in the source (`urlparse.py`) to discover this. `_replace` also makes it more difficult to figure out what is going on, that `joinedUrlParts._replace(path='/')` means the following `urlunparse` call returns `http://example.com/` instead of `http://example.com`. If simple string manipulation can do this, we should use that instead. 

I'm not clear on `Non-trivial rewrites do not work without a trailing '/'`. The earlier `joinedUrlParts._replace(path='/'.join(segments))` doesn't append a trailing `/`, and it probably shouldn't, so the problem is likely more specific than just missing a trailing `/`.

The failure to follow a redirect must happen somewhere outside of `absolutizeUrl`. Ideally we should either fix that failure, or, if it's in code we don't control, work around the problem as close to the failure point as we can.

Hainish:
@jeremyn 

> It's an implementation detail that the return value from `urlparse` is a `namedtuple` with a `_replace` method. It requires some digging in the source (`urlparse.py`) to discover this. `_replace` also makes it more difficult to figure out what is going on, that `joinedUrlParts._replace(path='/')` means the following `urlunparse` call returns `http://example.com/` instead of `http://example.com`. If simple string manipulation can do this, we should use that instead.

`joinedUrlParts._replace` is used twice in the above, and it's not obvious to me that string manipulation would do the job.  Your [above code example](https://github.com/EFForg/https-everywhere/pull/14598#issuecomment-364777407) simply adds a `/` wherever the URL does not end in a `/`, which is not desirable.  The previous method used the following line:

```python
urlParts = resolvedParsed[0:2] + (newPath,) + resolvedParsed[3:6]
```

which is functionally the same as 

```
urlParts = resolvedParsed._replace(path=newPath)
```

and avoids the `_replace` private method but is very unreadable. 

> I'm not clear on `Non-trivial rewrites do not work without a trailing '/'`. The earlier `joinedUrlParts._replace(path='/'.join(segments))` doesn't append a trailing `/`, and it probably shouldn't, so the problem is likely more specific than just missing a trailing `/`.

> The failure to follow a redirect must happen somewhere outside of `absolutizeUrl`. Ideally we should either fix that failure, or, if it's in code we don't control, work around the problem as close to the failure point as we can.

This is due to the way the rule is written, I'd think.  Your redirect rule example in #14365 includes a trailing `/`, which matches the `302` redirect in the browser, and the equivalent functionality in `absolutizeUrl` for this PR.

cschanaj:
> I'm not clear on Non-trivial rewrites do not work without a trailing '/'. The earlier joinedUrlParts._replace(path='/'.join(segments)) doesn't append a trailing /

This is because of the way we create non-trivial rewrites using regex. For example,

```xml
<rule from="^http://example\.com/" to="https://www.example.com/" />
                                ^
```

The _slash trailing the hostname_ must present such that the regex match, i.e. the path must begin with a slash for the non-trivial rewrite to be effective.

Ideally, we should modify `rule.apply` or change the way we create non-trivial rewrites. However, I am not sure about the purpose of the trialing slash requirement so I am not sure if it will do any harm.

jeremyn:
@Hainish I agree that my earlier naive string concatentation will not work.

`ParseResult` with parameters [is public](https://docs.python.org/2/library/urlparse.html#results-of-urlparse-and-urlsplit) and can be directly instantiated. This or similar should work:

```python
joinedUrlParts = ParseResult(join
	joinedUrlParts.scheme,
	joinedUrlParts.netloc,
	'/' if joinedUrlParts.path == '' else joinedUrlParts.path,
	joinedUrlParts.params,
	joinedUrlParts.query,
	joinedUrlParts.fragment,
)
```

with a later `geturl` call. We could also try some string manipulation based on the `ParseResult` members directly, like `'/'.join(joinedUrlParts.scheme, ...) + ...` or something like that, but that would be awkward and error-prone.

`_replace` should be avoided entirely because it is a private method and it relies on using `namedtuple` which is a private implementation detail. I'm not interested in arguing this particular point further, so if you still want to keep `_replace`, then I concede.

@Hainish and @cschanaj I think I understand the need for this change now, thanks.

After some digging ( https://webmasters.stackexchange.com/a/35646 , https://tools.ietf.org/html/rfc2616#section-5.1.2 ), it seems the relevant RFC requires a trailing `/` if the path is blank. Browsers add this themselves, and HTTPS Everywhere matches the URL after the `/` is added. So, adding a `/` manually, like we're trying to do in this pull request, is actually the correct thing for us to do. 

To make it totally clear, for my issue https://github.com/EFForg/https-everywhere/issues/14365 the full chain of events is this:

* HTTPS Everywhere sees `http://link.armytimes.com/` 
* it rewrites this to `https://link.armytimes.com/`
* the server for `https://link.armytimes.com/` returns a `302` response with a `Location` of `http://armytimes.com` (without a trailing `/`)
* (crucially) the ruleset checker doesn't match `http://armytimes.com` to `^http://armytimes\.com/` because it is missing a trailing `/`, so it doesn't rewrite it to `www`, and instead it hits the trivial rewrite to become `https://armytimes.com`
* curl/PycURL makes a request to `https://armytimes.com/`
* `https://armytimes.com/` gets `Failed to connect`

Please correct me if this is wrong.

Hainish:
@jeremyn that's right.  I'm fine with using the `ParseResult` constructor instead of `_replace` - seems pretty straightforward and readable, and avoids issues we've covered.  @cschanaj please make the change and I'll merge.

lbpasu:
@Hainish @jeremyn modifications made. please review and merge. thanks.

cschanaj:
@Hainish @jeremyn modifications made. please help to review and merge. thanks.

jeremyn:
@cschanaj The changes look fine to me, thanks again.

To tie up a loose end about `g:h` URLs, I get:

```
>>> joinedUrl = urlparse.urljoin('http://example.com', 'mailto:fake@example.org')
>>> urlparse.urlparse(joinedUrl)
ParseResult(scheme='mailto', netloc='', path='fake@example.org', params='', query='', fragment='')
>>>
```

The `path` of `fake@example.org` isn't modified by the later logic in the method, so the final `geturl()` call should return `mailto:fake@example.org`, which correctly matches the behavior of the original version of the method. So, it's not a problem.

Hainish:
Thank you @cschanaj

PR #14600, opened 2018-02-11
Add upgrade-insecure-requests to responseHeaders in HTTP Nowhere Mode

### LINKED ISSUE #8506: Use upgrade-insecure-requests to resolve MCB issues
From https://github.com/EFForg/https-everywhere/pull/7435#issuecomment-279627950:

> Hey guys, @ivysrono just talked to me about [UpgradeMixedContent](https://addons.mozilla.org/zh-CN/android/addon/upgrademixedcontent/). It's an extension that inserts upgrade-insecure-requests to every https page so browsers can rewrite http requests before start mixed content blocking.
> 
> The flaw is inserting such a header let browsers also rewrite display content while it's not blocked by default. If the display content isn't available for HTTPS, it fails. It's quite aggressive to do so for normal users who think http images don't really hurt, but it might be a good way for http-nowhere mode. Any thoughts?

Furthermore, we might be able to do something on mixedcontent rulesets.

cschanaj:
cc @Hainish @gloomy-ghost It would be really great to see this in the next release.

TEST CASES
1. CSP Exists, Contains `upgrade-insecure-requests` directive
  - https://www.w3.org/TR/upgrade-insecure-requests/

2. CSP Exists, Not Contains `upgrade-insecure-requests` directive
  - https://www.theguardian.com/international

3. CSP Not Exists, With mixed content
  - https://www.gsu.edu/about/

4. CSP Not Exists, Without mixed content
  - https://www.mozilla.org/

Follow up: Should we modify the CORS headers as well??

Close #8506 Close cschanaj/https-everywhere/pull/13

### DIFF
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -503,6 +503,20 @@ function onErrorOccurred(details) {
  */
 function onHeadersReceived(details) {
   if (isExtensionEnabled && httpNowhereOn) {
+    for (const idx in details.responseHeaders) {
+      if (details.responseHeaders[idx].name.match(/Content-Security-Policy/i)) {
+        // Existing CSP headers found
+        const value = details.responseHeaders[idx].value;
+
+        // Prepend if no upgrade-insecure-requests directive exists
+        if (!value.match(/upgrade-insecure-requests/i)) {
+          details.responseHeaders[idx].value = "upgrade-insecure-requests; " + value;
+        }
+        return {responseHeaders: details.responseHeaders};
+      }
+    }
+
+    // CSP headers not found
     const upgradeInsecureRequests = {
       name: 'Content-Security-Policy',
       value: 'upgrade-insecure-requests'
--- a/chromium/background.js
+++ b/chromium/background.js
@@ -497,10 +497,24 @@ function onErrorOccurred(details) {
   }
 }
 
+/**
+ * handle webrequest.onHeadersReceived, insert upgrade-insecure-requests directive
+ * @param details details for the chrome.webRequest (see chrome doc)
+ */
+function onHeadersReceived(details) {
+  if (httpNowhereOn) {
+    const upgradeInsecureRequests = {
+      name: 'Content-Security-Policy',
+      value: 'upgrade-insecure-requests'
+    }
+    details.responseHeaders.push(upgradeInsecureRequests);
+  }
+  return {responseHeaders: details.responseHeaders};
+}
+
 // Registers the handler for requests
 // See: https://github.com/EFForg/https-everywhere/issues/10039
 chrome.webRequest.onBeforeRequest.addListener(onBeforeRequest, {urls: ["*://*/*"]}, ["blocking"]);
-
 
 // Try to catch redirect loops on URLs we've redirected to HTTPS.
 chrome.webRequest.onBeforeRedirect.addListener(onBeforeRedirect, {urls: ["https://*/*"]});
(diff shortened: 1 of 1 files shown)


Bisaloo:
Built and tested with http://www.gsu.edu/about/. Good job! 👍

epicminecrafting:
@cschanaj If Content-Security-Policy header already exists, you should modify it instead of adding a new one.

epicminecrafting:
@cschanaj I've created a PR in your repo. Check it out.

cschanaj:
@epicminecrafting is there any new feature in the code you proposed? since the code is relatively short, I am not going to bother with the coding style here.

epicminecrafting:
@cschanaj Just minor fixes.

Hainish:
@cschanaj I can review as-is or wait for you to merge @epicminecrafting's changes.  Please advise.

cschanaj:
@Hainish Please review this as-is. thanks!

chromium/background.js
```diff
@@ -497,11 +497,39 @@ function onErrorOccurred(details) {
   }
 }
 
+/**
+ * handle webrequest.onHeadersReceived, insert upgrade-insecure-requests directive
+ * @param details details for the chrome.webRequest (see chrome doc)
+ */
+function onHeadersReceived(details) {
+  if (isExtensionEnabled && httpNowhereOn) {
```
Hainish: There may be a wild edge case here.  In https://github.com/EFForg/https-everywhere/blob/28087477257523c65d14bd5b4e1e79ad84056c08/chromium/background.js#L247-L253, we specifically allow `.onion` URLs that do not have HTTPS in HTTP Nowhere mode.  If a `.onion` URL serves only one page in HTTPS, and the rest are over HTTP, landing on that page with this logic will cause the client to make unreachable upgraded requests over HTTPS.

As I said, it's a pretty far-fetched edge case (why would a `.onion` site have a TLS configuration like that?) but it is possible, and probably worth accounting for.

cschanaj: I agree. Since `.onion` requests are encrypted anyway, IMHO we shall exclude them from the HTTPS upgrades.

chromium/background.js
```diff
@@ -497,11 +497,46 @@ function onErrorOccurred(details) {
...
+    // Do not upgrade the .onion requests in HTTP Nowhere Mode,
+    // See https://github.com/EFForg/https-everywhere/pull/14600#discussion_r168072480
+    const uri = new URL(details.url);
+    if (uri.hostname.slice(-6) == '.onion') {
+      return {responseHeaders: details.responseHeaders};
```
epicminecrafting:
```js
return {};
```

chromium/background.js
```diff
@@ -497,11 +497,46 @@ function onErrorOccurred(details) {
...
+    }
+
+    for (const idx in details.responseHeaders) {
```
epicminecrafting: Don't use a `for` loop, use `find` function.

chromium/background.js
epicminecrafting: I think requests to non-`.onion` websites should still be upgraded if they were made from `.onion` website.

cschanaj:
As per the Chrome's docs in https://developer.chrome.com/extensions/webRequest#type-HttpHeaders

> ... Only return responseHeaders if you really want to modify the headers in order to limit the number of conflicts (only one extension may modify responseHeaders for each request).

@epicminecrafting fixed in f08b9f4; as for `loop` vs `find`, I don't think there is much difference. Thanks anyway.

Hainish:
Thank you!

PR #14609, opened 2018-02-12
Update make.sh to support packaging on OSX, Fix #13794

### LINKED ISSUE #13794: Tests Fail on OSX due to sha1sum not found
(this is a low priority for me)
Type: code issue

I tested installing for dev on OSX. I was able to get install-dev-dependencies.sh to work without too much issue. When I go to run the tests though I run in to `./make.sh: line 150: sha1sum: command not found`

After a little research is appears shasum could be used instead on OSX, so perhaps a minor issue. I was able to get further by [linking to shasum](https://superuser.com/questions/734411/how-to-get-sha1sum-to-work-on-mac-os-x), but then ran into other errors. I'm not really a Mac person, so I stopped there.

cschanaj:
cc @Hainish

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -98,7 +98,7 @@ if ! [ -f "$key" ] ; then
 fi
 
 
-## Based on https://code.google.com/chrome/extensions/crx.html
+## Based on https://developer.chrome.com/apps/crx
 
 dir=pkg/crx
 name=pkg/crx
@@ -110,7 +110,7 @@ trap 'rm -f "$pub" "$sig" "$zip"' EXIT
 # zip up the crx dir
 cwd=$(pwd -P)
 (cd "$dir" && ../../utils/create_zip.py -n "$cwd/$zip" -x "../../.build_exclusions" .)
-echo >&2 "CWS crx package has sha1sum: `sha1sum "$cwd/$zip"`"
+echo >&2 "CWS crx package has sha256sum: `openssl dgst -sha256 -r "$cwd/$zip"`"
 
 # signature
 openssl sha1 -sha1 -binary -sign "$key" < "$zip" > "$sig"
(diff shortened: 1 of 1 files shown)


cschanaj:
AFAIK, OSX uses BSD `sed` which does not have a `--version` flags and `gsed` is not available on the machine I was testing on.

make.sh
```diff
@@ -98,7 +98,7 @@ if ! [ -f "$key" ] ; then
 fi
 
 
-## Based on https://code.google.com/chrome/extensions/crx.html
+## Based on https://developer.chrome.com/apps/crx
```
cschanaj: It seems that Google has modified the site not to includes the sample packaging script here...

jeremyn: The original URL https://code.google.com/chrome/extensions/crx.html redirects to https://developer.chrome.com/extensions/hosting , but the new URL https://developer.chrome.com/apps/crx redirects to https://developer.chrome.com/extensions/linux_hosting . I'm not clear what you intended, but it should probably be the final URL, after redirects.

jeremyn:
I'm not sure what process was described at https://code.google.com/chrome/extensions/crx.html , since it redirects now to https://developer.chrome.com/extensions/hosting , but I guess the magic numbers and `sed`/`xxd` stuff in `make.sh` are an old way of building Chrome extensions. At https://developer.chrome.com/extensions/linux_hosting#packaging the instructions suggest building the extension directly with Chrome, like this: `chrome.exe --pack-extension=C:\myext --pack-extension-key=C:\myext.pem`. If this is a new process, then we should switch that, and it would avoid the `sed`/`xxd` discussion above.

make.sh
```diff
@@ -110,7 +110,7 @@ trap 'rm -f "$pub" "$sig" "$zip"' EXIT
 # zip up the crx dir
 cwd=$(pwd -P)
 (cd "$dir" && ../../utils/create_zip.py -n "$cwd/$zip" -x "../../.build_exclusions" .)
-echo >&2 "CWS crx package has sha1sum: `sha1sum "$cwd/$zip"`"
+echo >&2 "CWS crx package has sha256sum: `openssl dgst -sha256 -r "$cwd/$zip"`"
```
jeremyn: There's nothing obvious here that says that this unusual `openssl dgst -sha256 -r` syntax is needed for OS X support. As an alternative, you could add something to the top like this:

```
if platform is OS X:
	pkg_hash_cmd=openssl dgst -sha256 -r
else:
	pkg_hash_cmd=sha256sum
```

and then use `$pkg_hash_cmd` as needed.

J0WI:
Isn't `shasum -a 256` working on both systems?

jeremyn:
I don't have a macOS system available and can't check it personally, but according to https://apple.stackexchange.com/a/230920 `shasum` is not included by default but is instead installed with Xcode Command Line Tools.

By the way, this all illustrates a general problem, suggested by my earlier comment at https://github.com/EFForg/https-everywhere/pull/14609#discussion_r170344903, which is that requiring `shasum` or `openssl dgst -sha256` instead of `sha256sum`, to support macOS, introduces a silent restriction that future maintainers might not know about or be able to check. For example someone might want to modify this script later, and might either break macOS support, or not have a macOS system available and so be unable to check their changes.

cschanaj:
@jeremyn the url was pointing to a script using `xxd` instead of the lengthy `sed` command; since it has been redirected, I guess Google now recommend developers to use the `chrome.exe --pack-extension=C:\myext --pack-extension-key=C:\myext.pem` to package extension. 

@J0WI I am using `openssl` to hash the files since it is also being used to sign the extension.  Without `openssl` the extension will not be signed and thus unusable. It seem reasonable to assume `openssl` is always available. (However, we might want to switch to the new way packaging HTTPSE). 

As @Hainish mentioned that we are standardizing the scripts for Debian platforms, it might not worth modifying a working script as it might introduce future maintenance burden as suggested by @jeremyn. I am closing this.

PR #14622, opened 2018-02-13
[Slides.com] Default_off ruleset

Bisaloo:
I can't think of a way to keep partial coverage.


J0WI:
I think this have been fixed, can you confirm?

Bisaloo:
Yeah, it looks like it's been fixed.

PR #14625, opened 2018-02-13
[ebayrtm.com] New ruleset

youdly:

### DIFF
--- a/src/chrome/content/rules/ebayrtm.com.xml
+++ b/src/chrome/content/rules/ebayrtm.com.xml
@@ -1,0 +1,83 @@
+<!--
+
+	For other eBay coverage, see ebay.com.xml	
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ebayrtm.com.xml
```diff
@@ -0,0 +1,83 @@
+<!--
+
+	For other eBay coverage, see ebay.com.xml
```
J0WI: Please also update the list in `ebay.com.xml`

youdly: Added in https://github.com/EFForg/https-everywhere/pull/14626/commits/009a46792aa559d2146894613c875cff3bc78fd0

PR #14641, opened 2018-02-15
added subdomains of cs.byu.edu that should be https only

bryceml:


J0WI:
@bryceml please have a look at the failing tests.

bryceml:
It doesn't appear that any of the sites I added caused the failing tests.  I can try to get around to fixing existing ones if you want, but it might be a while to get around to it.

J0WI:
Yes, please fix them.

bryceml:
All fixed.  There are probably more in the comments that work now that could be moved to the effective rules though.

J0WI:
LGTM, Thanks!

Issue #14467, opened 2018-01-29
Discussion: use Google styleguide for Markdown documents


jeremyn:
Type: other

(Ping @brainwane @gloomy-ghost @Hainish @J0WI @wonderchook but this discussion is open to anyone.)

I'm proposing that we adopt the [Google styleguide](https://github.com/google/styleguide/blob/7969290bacb1965d09677a79d523b4871c9d039c/docguide/style.md) for Markdown documents such as [`CONTRIBUTING.md`](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md) and [`README.md`](https://github.com/EFForg/https-everywhere/blob/master/README.md).

The motivation is that in pull request https://github.com/EFForg/https-everywhere/pull/14423 , @sampablokuper proposed hard wrapping our Markdown documents to 80 characters per line. This resulted in an unproductive discussion debating the merits of soft- vs hard-wrapping, particularly related to [computer accessibility](https://en.wikipedia.org/wiki/Computer_accessibility). I would like to prevent these discussions in the future, and take maintainers out of the role of judging Markdown style, by fully adopting a styleguide from some large organization that has presumably considered accessibility and other factors.

If we go with this, we don't need to update all our documents right away. The idea is simply to arrange things so that style updates that align with the styleguide can be merged without discussion, and similarly style updates that go against the styleguide can be rejected without discussion.

What I want from this issue is:

* For EFF staff ( @Hainish ) to specifically agree to support this proposal, and
* To give people in the community a chance to discuss it

Assuming the EFF supports this proposal, and no one argues against it for a week or so, we should consider it adopted.

I'm open to using the styleguide from another organization if people want to avoid Google for ethical/political reasons.

In this discussion, please avoid arguing based on aesthetic or trivial preferences. Also, I want to avoid a piecemeal styleguide, such as starting with Google's and making modifications. The point here is to remove discretion to debate trivial differences, but if we have some custom styleguide then I think it will make things worse in that respect.

jeremyn:
@sampablokuper made these points in https://github.com/EFForg/https-everywhere/pull/14423#issuecomment-361125026:

> At a glance, Google's Markdown style guide looks broadly reasonable. I think it would be a good starting point. I wouldn't necessarily adopt it wholesale.
> 
[...]

Issue #5613, opened 2016-07-06
Unable to right-click zip repository on Windows because of characters in filenames


jeremyn:
Filenames in the HTTPS Everywhere repository containing certain characters cause problems when using the "Send to > Compressed (zipped) folder" feature in Windows 10 when the system locale is "English (United States)". The files that cause the problem are:
- `cert-validity\mozilla\builtin-certs\NetLock_Arany_=Class_Gold=_Főtanúsítvány.crt`
- `test\rules\platform_certs\cacert\Cert-010_EBG Elektronik Sertifika Hizmet Sağlayıcısı_EBG Bilişim Teknolojileri ve Hizmetleri A.Ş.pem`
- `test\rules\platform_certs\cacert\Cert-126_NetLock Arany (Class Gold) Főtanúsítvány_NetLock Kft.pem`

Please consider renaming these files, wrapping them in other files, or something like that.
##### Workaround attempts
- If I change the system locale through "Control Panel > Clock, Language, and Region > Region > Administrative > Change system locale..." to "Hungarian (Hungary)", the first and third files work but the second still causes an error. If I change the system locale to "Turkish (Turkey)", the second works and first and third still cause errors.
- The [Compress-Archive](https://technet.microsoft.com/en-us/library/dn841358.aspx) PowerShell cmdlet compresses the file without error, but unfortunately it [quietly excludes hidden folders](https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/14754069-compress-archive-does-not-include-hidden-files-fol) like the `.git` folders.
##### How to reproduce the error
1. Download the zip file for commit 09414188005c2819691953daf2505d9ef92c9caf from GitHub
2. Unzip the file.
3. Right click the `https-everywhere-master` directory and select "Send to > Compressed (zipped) folder".
4. Observe the error: "[filename] cannot be compressed because it includes characters that cannot be used in a compressed folder, such as [character]. You should rename this file or directory."

jeremyn:
@Hainish Is there any reasonable chance we would rename or wrap the certificate files I named above to prevent this problem?

Hainish:
Renaming them shouldn't pose any problem AFAIK.

I wonder if we even need the `cert-validity` top-level path for anything?  I don't see it referenced anywhere.  @jsha, do you know what this path is for?  Is it still being used?

jsha:
Let's delete that dir. AFAIK not used for anything.

Hainish:
[...]

PR #14643, opened 2018-02-15
Update certificate bundle to latest firefox cert bundle

### LINKED ISSUE #1055: Update platform certs in https-everywhere-checker
We should pull the latest platform certs trusted by Firefox.

### LINKED ISSUE #14636: Update certificates to latest NSS
Type: other

Following from discussion in https://github.com/EFForg/https-everywhere/issues/5613#issuecomment-365798279 with @jeremyn, let's update our root certificate store to the latest NSS.

From that issue:

> https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/ seems to indicate that we can use [this file](https://hg.mozilla.org/mozilla-central/raw-file/tip/security/nss/lib/ckfw/builtins/certdata.txt) to determine the NSS root certificates.  Getting it into a format `curl` can handle is a different story.  https://gist.github.com/jjb/996292 provides some additional context here, including that we can use http://curl.haxx.se/ca/cacert.pem for the root store.

Bisaloo:
fix #14636, fix #1055

@Hainish


jeremyn:
Thanks @Bisaloo . Some comments:

* the downloaded `certdata.txt` file should be added to the repository, or added to the top level `.gitignore`

* the first line "This file comes from..." in the README is exactly 80 characters, one character too long. (See https://github.com/EFForg/https-everywhere/issues/14467 "Discussion: use Google styleguide for Markdown documents".) It should be wrapped like this:

```
This file comes from the [curl source                                       
repository](https://github.com/curl/curl).
```

* I'm surprised this works without code changes. PycURL must be able to read the combined `.crt` file in `platform_certs/default/` the same way as it currently reads the separate `.crt` files. However, maybe we should get rid of the `chromium/`, `default/` and `firefox/` directories and just point to the one combined `.crt` file, as @Hainish suggested in https://github.com/EFForg/https-everywhere/issues/14636#issuecomment-366019567.

Bisaloo:
From [curl man page](https://curl.haxx.se/docs/manpage.html):

> Using `--capath` can allow OpenSSL-powered curl to make SSL-connections much more efficiently than using `--cacert` if the `--cacert` file contains many CA certificates. 

I am okay with removing `firefox` and `chromium` directories to just keep `default` but we should disable those platforms in the same PR by editing the relaxNG file in this case.

jeremyn:
@Bisaloo From your quote I guess you are saying we should use `--capath` instead of `--cacert` for performance reasons, but I wonder if the curl documentation is considering the case when `capath` contains just one file with many certificates, like we are considering.

In any case we should probably use a directory in case we want to drop in extra certificates later on, permanently or just for a temporary test. You can clean up the platforms if you want but for this PR I would be okay with leaving all the directories and platforms as they are.

The changes in https://github.com/EFForg/https-everywhere/pull/14643/commits/f87bcb7370d36ea41554ec220f748f1cc6a4e34e and https://github.com/EFForg/https-everywhere/pull/14643/commits/11093f4b46422da428698f41d8faf7eaf21192f6 look fine to me and address the other two bullet items in https://github.com/EFForg/https-everywhere/pull/14643#issuecomment-366088299, thanks.

Hainish:
@Bisaloo I would avoid adding `utils/mk-ca-bundle/*`, since these files are already contained in a separately maintained project.  I would include the `cacert.pem` certificate authority bundle as well as instructions on how to reproduce it (similar to what we have in https://github.com/EFForg/https-everywhere/blob/5d19d26/chromium/external/README.md) in a separate file.

> From curl man page:
> 
> > Using --capath can allow OpenSSL-powered curl to make SSL-connections much more efficiently than using --cacert if the --cacert file contains many CA certificates.
> 
> I am okay with removing firefox and chromium directories to just keep default but we should disable those platforms in the same PR by editing the relaxNG file in this case.

You can use the following script to split the `cacert.pem` file into a directory specified by `--capath`:

```bash
#!/bin/bash                                          
                                                     
rm -rf default                                       
mkdir default                                        
cd default                                           
curl -O https://curl.haxx.se/ca/cacert.pem # or use  
cat cacert.pem|awk 'split_after==1{n++;split_after=0}
   /-----END CERTIFICATE-----/ {split_after=1}       
   {if(length($0) > 0) print > "cert" n ".pem"}'     
c_rehash .                                           
```

jeremyn:
11093f4 (the branch) takes about 125%-150% as long to run as f6937b4 (recent master), tested using `time test/manual.sh rules/<ruleset.xml>` with `Freerangekitten.xml` and `MIT.xml`. Using 11093f4 (the branch) but breaking up the combined file with @Hainish 's method, it takes about the same as f6937b4 (recent master) for `MIT.xml` and is faster for `Freerangekitten.xml`.

So, we should definitely avoid a combined file. First combining and then breaking up the certificate seems a little pointless at first, but it avoids the Unicode problem from issue https://github.com/EFForg/https-everywhere/issues/5613, and the process can be easily rerun. So let's do that.

@Hainish 's small script is mostly fine but a little buggy, particularly because:

* it leaves the combined file in `default/`, and `c_rehash` creates a link to it, and
* `c_rehash` reports `WARNING: Skipping duplicate certificate cert.pem` twice, which seems to be caused by the leading comment from the combined file being kept in `cert.pem`

This script should probably be rewritten in bash or Python so it can handle these little details better.

jeremyn:
By "this script" in my previous comment, I mean the awk part.

Hainish:
@jeremyn definitely.  I didn't mean for that to be a polished product, just as a useful reference  for the commands necessary to split and "rehash" the `.pem` file.

Bisaloo:
Is 96156a3 what you had in mind? If so, I can remove previous commits.

jeremyn:
@Bisaloo What you have works, however it has some small issues. How about what I have below? It has some improvements: it is compatible with Python 2 and 3, and it avoids putting Unicode in filenames, which avoids the problem in issue https://github.com/EFForg/https-everywhere/issues/5613.

The bash script uses the `cacert.pem` from the curl site, however I think checking in `mk-ca-bundle.pl` and using that is better. @Hainish seems to disagree though, at least with checking in `mk-ca-bundle.pl`. In any case the script can be easily modified for either approach. Also feel free to move things around a little, for example you can put these both into `utils/` somewhere and change the filenames or paths.

```bash
#!/usr/bin/env bash
set -e

COMBINED_FILE=cacert.pem
OUTPUT_DIR=./default/
PEM_URL=https://curl.haxx.se/ca/cacert.pem

rm -rf "${OUTPUT_DIR}"
mkdir -p "${OUTPUT_DIR}"

curl -o "${COMBINED_FILE}" "${PEM_URL}"

python split_combined_file.py "${COMBINED_FILE}" "${OUTPUT_DIR}"

rm "${COMBINED_FILE}"

c_rehash "${OUTPUT_DIR}"
```

```python
#!/usr/bin/env python
import argparse
import codecs
import os
import re

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument(
        'input_filename',
        help='combined crt/pem file to split',
    )
    parser.add_argument(
        'output_dir',
        help='output directory for split files',
    )
    args = parser.parse_args()
    with codecs.open(args.input_filename, 'r', encoding='utf-8') as input_file:
        bundle = input_file.read()
        certs = re.compile('\n{2,}').split(bundle)

        # First element of certs is a comment
        certs = certs[1:]
        cert_filename_num_width = len(str(len(certs)))
        for index, cert in enumerate(certs):
            cert_name, cert_content = re.compile("\n=+\n").split(cert)
            cert_filename = os.path.join(
                args.output_dir,
                "cert%s.pem" % str(index+1).zfill(cert_filename_num_width),
            )
            with codecs.open(
                    cert_filename, 'w', encoding='utf-8') as cert_file:
                cert_file.write("%s\n" % cert_name)
                cert_file.write(cert_content)
                cert_file.write('\n')
```

Hainish:
@jeremyn

> The bash script uses the `cacert.pem` from the curl site, however I think checking in `mk-ca-bundle.pl` and using that is better. @Hainish seems to disagree though, at least with checking in `mk-ca-bundle.pl`. In any case the script can be easily modified for either approach. Also feel free to move things around a little, for example you can put these both into `utils/` somewhere and change the filenames or paths.

I think it's safer to generate `cacert.pem` from the perl script instead of just curling it.  We should then check `cacert.pem` in, instead of checking in `mk-ca-bundle`, which is separately maintained.  We should then include an instruction on how to update the `cacert.pem` file in a separate markdown document.

Hainish:
Alternatively, instead of checking in `cacert.pem`, we could just check in the result of the split certificate path.  It's probably not necessary to check in `cacert.pem` if the contents of the split directory can be built reproducibly.

jeremyn:
If we are concerned about exact reproducibility, then we don't want `mk-ca-bundle.pl` to be separately maintained, because it can go out of sync.

The ideal approach to cover all needs seems to be:

* check in the split certificate directory; this is for convenience, so people don't need to generate it to run tests

* check in the original input, all programs used to create the split certificate directory, and instructions; this means the `certdata.txt` file that `mk-ca-bundle.pl` downloads (but not `ca-bundle.crt` (`cacert.pem`) which it creates), `mk-ca-bundle.pl` itself, the programs that run after it like the ones @Bisaloo and I have proposed, and a README; this is for complete reproducibility

* provide instructions on how to update this process, meaning where to get a new `mk-ca-bundle.pl` (which will itself download a new `certdata.txt`); this can be a couple lines in the README from the previous step; this is for future maintenance.

The only hitch is licensing. `certdata.txt` is MPL v2.0. `mk-ca-bundle.pl` is kind of weird: the blurb at the top of it is not the same as what is found at the URL https://curl.haxx.se/docs/copyright.html included in the blurb, and the license at https://curl.haxx.se/docs/copyright.html has a non-standard "no advertising" clause. We might just provide instructions on where to get these files, and their hashes/versions, just to avoid licensing issues.

Hainish:
> If we are concerned about exact reproducibility, then we don't want mk-ca-bundle.pl to be separately maintained, because it can go out of sync.

I'm not as concerned with asynchronous reproducibility as I am with synchronous reproducibility.  The output of `mk-ca-bundle.pl` will change over time anyway, even if the script is checked in and remains unchanged, due to the fact that it references the certificate store from Mozilla's hg `HEAD`.  If we're concerned with bit-by-bit reproducibility, we can ensure this by having multiple parties run the script synchronously.  Frankly, I don't think bit-by-bit reproducibility is that important, since this affects the tests alone and this code is not delivered to our millions of users.  I don't consider false negatives or positives in our fetch test a significant risk to our users.

You can see from https://curl.haxx.se/docs/caextract.html that the CA store is only changed rarely.  I consider it safe for a trusted maintainer to update the certificates manually when these changes occur.

Hainish:
So ideally, we'd check in:

* the split certificate directory
* a script which takes the bundled `cacert.pem` as input and outputs the split directory
* instructions on how to generate or download `cacert.pem`

jeremyn:
The output from `mk-ca-bundle.pl` should be bit-by-bit identical given the same `certdata.txt` input, which is the certificate store from Mozilla you mention. So bit-by-bit reproducibility is achievable if we want it.

Bad certificates could cause some trouble if they cause some rulesets to get incorrectly rejected or accepted. They are not totally harmless.

Without bit-by-bit reproducibility, if the certificates are later found to have a problem, you might have no way of knowing what happened, and can only discover that (for example) `@some_contributor` and `@some_maintainer` signed off on them. That's not a good position to be in, not for the contributor or the maintainer or the person looking back trying to figure out what happened.

Anyway, regardless of what's checked in or not, etc, the key issue for reproducibility is whether we include enough information such that the split certificates can be bit-by-bit regenerated months or years later using integrity-checkable external inputs, with integrity-checkable internal or external code. I hope that we do include enough information but I'm not super picky on exactly what it is.

jeremyn:
@Bisaloo If you're not sure about moving forward here because @Hainish and I are talking about this, I can take this over for you if you want, meaning I can push changes to your branch and make revisions as needed until this is merged. Just let me know.

Bisaloo:
@jeremyn, I am slightly under the water with work atm. You are totally welcome to edit stuff on this branch if you wish.

jeremyn:
@Bisaloo @Hainish My changes are ready for review.

jeremyn:
I've remove the `c_rehash` step and output. It doesn't seem to be required for `test/manual.sh` or meaningfully improve its performance.

Hainish:
@jeremyn I'll review this later today or tomorrow.

Hainish:
Superseded by https://github.com/EFForg/https-everywhere/pull/14825

PR #14649, opened 2018-02-16
[Where2GetIt.com] Remove broken rules and wildcard

Bisaloo:


src/chrome/content/rules/Where2GetIt.com.xml
```diff
@@ -17,7 +18,10 @@
 <ruleset name="Where2GetIt.com (partial)">
 
 	<target host="ignite.where2getit.com" />
+	<target host="info.where2getit.com" />
 	<target host="mobile.where2getit.com" />
+	<target host="social.where2getit.com" />
+		<test url="http://social.where2getit.com/fiveguys/OK/EDMOND/BRYAN-410SBASAEO/" />
```
J0WI: Fetch tests did not pass on this host.

Bisaloo: This host seems to randomly fail over both `http` and `https`. Should I restart tests until it works?

J0WI: LGTM

PR #14652, opened 2018-02-16
[humankode.com] new ruleset

thecarlo:


J0WI:
Thanks!
You might be also interested to submit your domain on https://hstspreload.org/

thecarlo:
Thanks for the quick merge.
My domain is already on the hsts preload list. The hsts header wasn't loaded - I made a correction and everything is fine now.

PR #14660, opened 2018-02-18
Add Test for duplicated test URLs, Fix #14659

### LINKED ISSUE #14659: Add test for duped test urls
Duped test urls should not be allowed:

```
<test url="http://example.com/" />
<test url="http://example.com/" />
```
Type: feature request

cschanaj:
**UPDATE 2018.03.20** Blocked by #14944 Blocked by #14924 

Close #14659 Close #14661 Hopefully before #14907 

c18d192 seems to be faster than 2e3e2c1 over Travis... 

We might want to go through some Alexa top sites before performing automatic removals.

... TODO LIST REMOVED ...


Bisaloo:
Any idea why you didn't pick up `GoogleAPIs.xml`?

see https://github.com/EFForg/https-everywhere/pull/14740/files (`storage.googleapis.com`)

Bisaloo:
Ah, I'm guessing it's because it was whitelisted for coverage.

cschanaj:
> Ah, I'm guessing it's because it was whitelisted for coverage.

@Bisaloo yes, whitelisted rulesets won't be showing up here.

cschanaj:
See #14924 for automated removal of the duplicated test URLs. thanks.

cschanaj:
@Hainish @J0WI Travis is passing now. I guess this is ready to be merge. thanks!!

J0WI:
ping @Hainish again

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -80,6 +80,12 @@ def __init__(self, url):
         """
         self.url = url
 
+    def __eq__(test1, test2):
+        return test1.__hash__() == test2.__hash__()
```
Hainish: Why can't you use `return test1.url == test2.url` here, and remove the `__hash__` method below?  This would also remove the internal string `__hash__` method, which is preferable.

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -177,6 +183,14 @@ def _determineTestApplication(self):
                  off that rule or exclusion.  Return any coverage problems."""
         if not self.determine_test_application_run:
             self.test_application_problems = []
+
+            # check for duplicated test urls
+            if len(self.tests) != len(set(self.tests)):
+                for test in self.tests:
```
Hainish: Currently, if you run this test it will print one error message for each time the duplicate test URL appears.  You can avoid this by just making this line `for test in set(self.tests):`

cschanaj:
@Hainish done. thanks!!

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -80,6 +80,12 @@ def __init__(self, url):
         """
         self.url = url
 
+    def __eq__(test1, test2):
+        return test1.url == test2.url
+
+    def __hash__(self):
```
Hainish: We can get rid of this now, no?

test/rules/src/https_everywhere_checker/rules.py
```diff
@@ -177,6 +183,15 @@ def _determineTestApplication(self):
                  off that rule or exclusion.  Return any coverage problems."""
         if not self.determine_test_application_run:
             self.test_application_problems = []
+
+            # check for duplicated test urls
+            if len(self.tests) != len(set(self.tests)):
+                uniqueTests = set(self.tests)
```
Hainish: We're not using this anywhere.  We should change the below line to be `for test in set(self.tests):`

cschanaj:
@Hainish In 4c52886 it seems that `Test` has to be hashable to construct set, so I have kept the method. thanks!!

Hainish:
Thank you!

PR #14668, opened 2018-02-18
Remove duplicated test urls from JS.org.xml

cschanaj:
Honestly, I don't really like this ruleset given the overwhelming amount of `exclusion` which cannot be tested automatically...


J0WI:
This ruleset has been generated from a public list, see https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/JS.org.xml#L2

PR #14670, opened 2018-02-18
Remove duplicated test urls from Flickr.xml

cschanaj:


src/chrome/content/rules/Flickr.xml
```diff
@@ -45,9 +46,7 @@
 	<target host="api.flickr.com" />
 		<test url="http://api.flickr.com/services/rest/" />
 		<exclusion pattern="^http://api\.flickr\.com/$" />
-			<test url="http://api.flickr.com/" /><!-- 404 -->
```
J0WI: Please keep the `404` comment for the exclusion.

PR #14685, opened 2018-02-18
Remove duplicated test urls from Sina.com.cn.xml

cschanaj:


src/chrome/content/rules/Sina.com.cn.xml
```diff
@@ -60,6 +60,7 @@
 		roll.news.sina.com.cn
 		sky.news.sina.com.cn
 		slide.news.sina.com.cn
+		help.pay.sina.com.cn
```
J0WI: MCB is the wrong category, I get an expired cert.

PR #14689, opened 2018-02-19
Add byond.com ruleset

MrStonedOne:


src/chrome/content/rules/Byond.com.xml
```diff
@@ -0,0 +1,23 @@
...
+<ruleset name="byond.com">
+	<target host="byond.com"/>
+	<target host="www.byond.com"/>
+	<target host="secure.byond.com"/>
+	
+	<exclusion pattern="^http://(www\.)?byond\.com/play"/>
+		<test url="http://byond.com/play"/>
+		<test url="http://www.byond.com/play"/>
+	
+	<rule from="^http://([\w.-]*)byond\.com/" to="https://secure.byond.com/"/>
```
J0WI: Please try to avoid complex regex for simple cases. Just use `(www\.)?` here and add a `<rule from="^http:" to="https:" />` to handle all other cases (here it's just `secure.byond.com`).

src/chrome/content/rules/Byond.com.xml
```diff
@@ -0,0 +1,23 @@
...
+		<test url="http://byond.com/"/>
```
J0WI: This test url is already covered by the target host.

src/chrome/content/rules/Byond.com.xml
```diff
@@ -0,0 +1,23 @@
...
+		Most links specify http in the url.
+-->
+<ruleset name="byond.com">
+	<target host="byond.com"/>
+	<target host="www.byond.com"/>
+	<target host="secure.byond.com"/>
+	
+	<exclusion pattern="^http://(www\.)?byond\.com/play"/>
+		<test url="http://byond.com/play"/>
+		<test url="http://www.byond.com/play"/>
```
J0WI: Please add trailing slashes.

src/chrome/content/rules/Byond.com.xml
```diff
@@ -0,0 +1,24 @@
...
+<ruleset name="byond.com">
+	<target host="byond.com" />
+	<target host="www.byond.com" />
+	<target host="secure.byond.com" />
+	
+	<exclusion pattern="^http://(www\.)?byond\.com/play/" />
+		<test url="http://byond.com/play/" />
+		<test url="http://www.byond.com/play/" />
+	
+	<rule from="^http://(www\.)?byond\.com/" to="https://secure.byond.com/" />
```
J0WI: Please see https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects
So `^` should just redirect to `https://www`. All other cases are handled below or server side.

MrStonedOne: I think it might be better if i just removed the `(www\.)?` regex and redirected byond.com to either `www.` or `secure.` and allow the generic rule below handle www. and secure.

J0WI:
Thanks!

PR #14704, opened 2018-02-20
Create QLCrew.com.xml

stordoff:


src/chrome/content/rules/QLCrew.com.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="QLCrew">
+
+	<target host="qlcrew.com" />
+	<target host="www.qlcrew.com" />
```
J0WI: Please also add `login.qlcrew.com`.

J0WI:
Thanks!

PR #14706, opened 2018-02-20
[AmazonAWS.com] Add exclusion (fix #11640)

### LINKED ISSUE #11640: brackets.io - Download button does not work with HTTPS Everywhere enabled
On http://brackets.io the CTA is broken (href value is "#") when HTTPS Everywhere is enabled.

See the comment on my original report which lead me to test the add-on:/
https://github.com/webcompat/web-bugs/issues/8632#issuecomment-319865463

I have now also tested it in current stable FF 54.0.1 with a pretty clean profile and only HTTPS Everywhere installed. Same issue.

Please let me know if more details are required. Thanks

Bisaloo:


J0WI:
Can we limit the exclusion to the `json`requests?

Bisaloo:
Yes, I tested with the debugging tool and it works.

PR #14708, opened 2018-02-21
Add ruleset for vertx.com

injust:


Bisaloo:
Please remove `autodiscover.vertx.com`. It's an internal subdomain and there is no public accessible url for it.

injust:
@Bisaloo I've reverted that commit, sorry for the long wait.

PR #14709, opened 2018-02-21
Add ruleset for cheatbook.de

Injust:


src/chrome/content/rules/Cheatbook.de.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Invalid certificate:
+	- blog.cheatbook.de
+-->
+
+<ruleset name="Cheatbook.de">
```
J0WI: This rule needs the mixedcontent platform attribute. Some pages load CSS over http.

Issue #9263, opened 2017-03-26
HTTPS Everywhere breaks Seriouseats.com videos


vidiat:

Bisaloo:
> 08:29:37.526 Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://hlsak-a.akamaihd.net/4841592010001/5361782679001/4841592010001_5361782679001_s-1.ts?pubId=4841592010001&videoId=5361756079001. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). 1 (unknown)

The culprit is the akamai ruleset.

Bisaloo:
@J0WI, this is fixed since you merged #14710

PR #14710, opened 2018-02-21
Create ruleset for seriouseats.com

injust:


Bisaloo:
@Injust, can you add `Fix #9263` to your PR description please? It will automatically close #9263 once this PR is merged.

More info: https://help.github.com/articles/closing-issues-using-keywords/

Bisaloo:
Oh, I didn't see it was `platform="mixedcontent"`. Do we have a way to fix #9263 for everybody?

Bisaloo:
As far as I can tell, only active mixed content is from http://www.assoc-amazon.com/. I think it's okay to break ads so I would remove the `platform`.

Injust:
@Bisaloo The site also uses mixed content for comments.

J0WI:
I wasn't able to reproduce MCB issues with comments.

Injust:
@J0WI Seems that between the time when I opened the PR and when you checked, SSL was deployed on the site.
Removing `platform="mixedcontent"` from the ruleset.

Issue #8265, opened 2017-01-29
Rule for hartwork.org 


joshtriplett:
The rules in src/chrome/content/rules/Hartwork-Project.xml redirect http://hartwork.org to https://hartwork.org ; however, the latter gives SSL errors, because the certificate only works for www.hartwork.org.  The site itself redirects to https://www.hartwork.org/

cc @hartwork

hartwork:
I have migrating to Let's Encrypt on my todo list. I need more time though.
If there is something quick that I could do before new certificates, please let me know.

joshtriplett:
@hartwork I don't think you need to do anything here.  https-everywhere should stop sending people to hartwork.org with https.

When you do migrate to Let's Encrypt, I'd recommend choosing a canonical domain name (for instance, hartwork.org), and redirecting everything else (such as www.hartwork.org) to that canonical name, rather than having multiple domains that serve the same content.  But you'll still need SNI and SANs to allow https connections for the non-canonical domains in order to send such a redirect.

hartwork:
> @hartwork I don't think you need to do anything here. https-everywhere should stop sending people to hartwork.org with https.

Okay.

> When you do migrate to Let's Encrypt, I'd recommend choosing a canonical domain name (for instance, hartwork.org), and redirecting everything else (such as www.hartwork.org) to that canonical name, rather than having multiple domains that serve the same content. But you'll still need SNI and SANs to allow https connections for the non-canonical domains in order to send such a redirect.

Sounds good, thanks!

hartwork:
> The rules in src/chrome/content/rules/Hartwork-Project.xml redirect http://hartwork.org to https://hartwork.org ; however, the latter gives SSL errors, because the certificate only works for www.hartwork.org.

I fixed the SSL errors on https://hartwork.org now using Let's encrypt. I would like to keep [src/chrome/content/rules/Hartwork-Project.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Hartwork-Project.xml) in place personally, so that people are not MITM'ed on the initial (otherwise plain HTTP) request.

> When you do migrate to Let's Encrypt, I'd recommend choosing a canonical domain name (for instance, hartwork.org), and redirecting everything else (such as www.hartwork.org) to that canonical name, rather than having multiple domains that serve the same content.

I have put that to action now.

Bisaloo:
This seems fixed. Can we close this issue?

hartwork:
[...]

PR #14724, opened 2018-02-22
rules: Hartwork-Project: Remove forward for binera.de

hartwork:
Somewhat related to #8265. binera.de has a dedicated certificate of its own by now and could change to serving its own content at some point.


hartwork:
Thanks!

PR #14752, opened 2018-02-25
[InternetSociety.org] Simplify ruleset 

### LINKED ISSUE #4035: Review Internet-Society
Is the [Internet-Society ruleset](https://github.com/EFForg/https-everywhere/blob/d1f8c60458fe05e95df71ce451cecaf0f5de5b9b/src/chrome/content/rules/Internet-Society.xml) still up-to-date?

There are many exclusions, but when I tried to access some exclusions over HTTPS, they all worked. So possibly this ruleset can be simplified.

/cc @internetsociety

Bisaloo:
Fix #4035


src/chrome/content/rules/Internet-Society.xml
```diff
@@ -3,67 +3,31 @@
...
-
+	<target host="apps.internetsociety.org" />
+	<target host="assets.internetsociety.org" />
+	<target host="connect.internetsociety.org" />
+	<target host="dev.internetsociety.org" />
+	<target host="future.internetsociety.org" />
+	<target host="inforum.internetsociety.org" />
+	<target host="intersect.internetsociety.org" />
+	<target host="staging.internetsociety.org" />
+	<target host="cdn.staging.internetsociety.org" />
```
J0WI: `cdn.prod.internetsociety.org`

src/chrome/content/rules/Internet-Society.xml
```diff
@@ -3,67 +3,31 @@
...
 
 	Mixed content:
-
-		- Images on www from $self *
-
-	* Secured by us
+		wp.internetsociety.org
 
 -->
 <ruleset name="Internet Society.org (partial)">
```
J0WI: It's no longer `(partial)`

PR #14761, opened 2018-02-25
Add electroneum.com

DavidLiedke:


src/chrome/content/rules/electroneum.com.xml
```diff
@@ -0,0 +1,14 @@
+
+<ruleset name="electroneum.com">
+
+	<target host="electroneum.com" />
+	<target host="*.electroneum.com" />
```
J0WI: Is there a specific reason for using a wildcard?

DavidLiedke: Because they have a wildcard certificate. If you want i can replace the wildcard.

J0WI: We try to avoid wildcards, because they are harder to maintain.

src/chrome/content/rules/electroneum.com.xml
```diff
@@ -2,10 +2,10 @@
 <ruleset name="electroneum.com">
 
 	<target host="electroneum.com" />
```
J0WI: `www.electroneum.com`

DavidLiedke:
Done

PR #14765, opened 2018-02-25
Add spacepools.org

DavidLiedke:


src/chrome/content/rules/spacepools.org.xml
```diff
@@ -0,0 +1,13 @@
+
+<ruleset name="spacepools.org">
+
+	<target host="*.spacepools.org" />
```
J0WI: Is there a specific reason for using a wildcard?

DavidLiedke:
@J0WI Wildcard removed.

PR #14767, opened 2018-02-25
Removed hosts ics.imensa.de and rss.imensa.de

arneschuldt:
* Removed host ics.imensa.de which is no longer available
* Removed host rss.imensa.de which has reached its EOL and will be no longer available after March (all feeds already show notice about new URLs on main host www.imensa.de)

### DIFF
--- a/src/chrome/content/rules/Imensa.de.xml
+++ b/src/chrome/content/rules/Imensa.de.xml
@@ -1,8 +1,6 @@
 <ruleset name="Imensa.de">
   <target host="imensa.de" />
   <target host="www.imensa.de" />
-  <target host="ics.imensa.de" />
-  <target host="rss.imensa.de" />
 
   <rule from="^http:" to="https:" />
 </ruleset>


arneschuldt:
* Host rss.imensa.de is now also no longer available (as announced before and already reflected in change set)

Bisaloo:
Can you also add:

Invalid certificate:
- `firefox.app.imensa.de`
- `austria.firefox.app.imensa.de`

Target:
- `css.imensa.de` with test url http://css.imensa.de/1.0.38/css/imensa.css
- `js.imensa.de` with test url http://js.imensa.de/1.0.38/js/javascript.js

arneschuldt:
Thanks for your great work and for proposing extensions to the ruleset.

Please let me explain why we would prefer not to make these changes.

* firefox.app.imensa.de
* austria.firefox.app.imensa.de

Actually, these host names are simply not configured for HTTPS. So, an invalid certificate error might rather occur from calling them with HTTPS and the webserver serving the default website (which has no matching certificate for obvious reasons).

These host names serve web apps formerly distributed through Firefox Marketplace. They are no longer advertised and are simply kept for very few remaining users. It would be easy to add SSL support to these hosts. However, that meant no benefit because the original URL of a Firefox OS web app cannot be changed once it has been installed (and installation happened long time ago).

We will likely deactivate these host names soon. So adding rules for them now meant useless effort now and soon again to remove them.

* css.imensa.de
* js.imensa.de

These are purely technical host names on which we serve static assets through a content delivery network (CDN). We can guarantee that we reference resources on these host names always with HTTPS (because otherwise HTTPS for our website would be broken). So adding rules for these host names would only bloat the ruleset (there is no use case for opening these URLs manually in the browser).

Furthermore, we change these purely technical CDN host names from time to time. We do not intend to keep the rulesets up to date for such changes (because it simply means no benefit). For the same reason, we do not want volunteers to unnecessarily spent time on reviewing pull requests for such purely technical updates. :)

I hope you agree that not adding the proposed rules has no drawbacks, but hopefully makes life simpler for everyone.

Bisaloo:
You should not worry about bloating the ruleset or using our time. The impact of two additional targets is completely negligible.

I still would like to add the subdomains I listed in https://github.com/EFForg/https-everywhere/pull/14767#issuecomment-374621333.

The main reason is to follow the same format as the other rulesets. We always add all subdomains, either as targets or in the top comment if they don't support HTTPS (see [the related section in our contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts)).

As opposed to what you think, I believe it would make things more complicated to make an exception here (we would then have to remember it, etc.) than adding two simple targets and a short comment about subdomains that are not configured to work with HTTPS.

arneschuldt:
Thanks for the clarification. Please feel free to add any further host names you like.

While we do not consider it useful to add these purely technical hosts (and thus won't spent our time), we have no objection to adding them. All our certificates are subject to certificate transparency. So, adding them to rulesets should not reveal any secrets.

However, please note that some of the hosts you mentioned in your comment have already changed in the meantime.

arneschuldt:
Closed due to inactivity.

PR #14769, opened 2018-02-25
Remove IP address in targets

Bisaloo:
I'm not interested in fixing/updating the rest of the ruleset


epicminecrafting:
Good idea since IP addresses frequently change, with some exceptions (1.1.1.1 and 1.0.0.1).

PR #14780, opened 2018-02-26
Update freeboard.io.xml

J0WI:
#14660

### DIFF
--- a/src/chrome/content/rules/freeboard.io.xml
+++ b/src/chrome/content/rules/freeboard.io.xml
@@ -1,15 +1,17 @@
 <!--
-www.freeboard.io mismatch
+	Mismatch:
+		- www.freeboard.io
 -->
 <ruleset name="freeboard.io">
 	<target host="freeboard.io" />
 	<target host="www.freeboard.io" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/freeboard.io.xml
```diff
@@ -1,15 +1,17 @@
 <!--
-www.freeboard.io mismatch
+	Mismatch:
+		- www.freeboard.io
```
Bisaloo: see also vz.freeboard.io

src/chrome/content/rules/freeboard.io.xml
```diff
@@ -1,15 +1,17 @@
...
 -->
 <ruleset name="freeboard.io">
 	<target host="freeboard.io" />
 	<target host="www.freeboard.io" />
 	<target host="alpha.freeboard.io" />
+	<target host="openxc.freeboard.io" />
+	<target host="renesas.freeboard.io" />
```
Bisaloo: I get `HTTP ERROR 503`

J0WI: Removed for now.

PR #14786, opened 2018-02-27
Restore a missing line break

sampablokuper:
that seems to have been accidentally deleted at some point since
https://github.com/EFForg/https-docs/commit/236ea7c5f0404a2d0f4b7ac1cf614b555ae734d4
.


J0WI:
@Hainish this LGTM

PR #14788, opened 2018-02-27
Restore missing section from rulesets.md

sampablokuper:
which seems to have been left behind, in the migration of documentation from https://github.com/EFForg/https-docs to https://github.com/EFForg/https-everywhere/tree/master/docs/en_US .

If, for some reason, that omission was deliberate, please feel free to give the reason here, and to decline this pull request :)

### DIFF
--- a/docs/en_US/rulesets.md
+++ b/docs/en_US/rulesets.md
@@ -205,14 +205,17 @@ bug](https://trac.torproject.org/projects/tor/ticket/5196) caused us to
 temporarily disable the Google Translate rules on Chromium and Chrome.
 This can be achieved with the "platform" attribute:
 
-    <ruleset name="Google Translate" platform="firefox">
-    <target host="translate.googleapis.com"/>
-    <target host="translate.google.com"/>
-    <rule from="^http://translate\.googleapis\.com/"
-            to="https://translate.googleapis.com/"/>
-    <rule from="^http://translate\.google\.com/translate_a/element\.js"
-            to="https://translate.google.com/translate_a/element.js"/>
-    </ruleset>
+```xml
+<ruleset name="Google Translate" platform="firefox">
+	<target host="translate.googleapis.com"/>
+	<target host="translate.google.com"/>
+
+	<rule from="^http://translate\.googleapis\.com/"
+	    to="https://translate.googleapis.com/"/>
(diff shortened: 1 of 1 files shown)


Bisaloo:
The code block you added doesn't look very well: [preview](https://github.com/sampablokuper/https-everywhere/blob/6f7b54bdaf2b1bd6913f9019afbde8c3e2c4e17a/docs/en_US/rulesets.md).

You should fix the indents and mark the code block as `xml` to get syntax highlighting.

cschanaj:
AFAIK, some of the `platform` are not being used anymore in the rulesets database. 

Both `firefox`, `chromium` and `ipsca` has no special meaning to the extension core behaviors (all of them regards as `default_off`) and thus it might not worth including them anymore. 

/cc @Hainish

sampablokuper:
@cschanaj thanks, that's exactly the sort of reason (alluded to in my opening comment above) why this block *might* have been deliberately omitted. Whether it *was* deliberately omitted or not, and which `platform` options are still available (if any), I do not know. Thank you for pinging Hainish; with any luck we will get clarity from one of the core contributors/maintainers on this :)

Hainish:
@sampablokuper these docs are intended to generate the HTML that is posted on the website at https://www.eff.org/https-everywhere/rulesets.  I was actually unaware of the https://github.com/EFForg/https-docs repo until relatively recently, when @brainwane brought them to my attention.  I don't know if at some point this section was removed or if it was never added.  At any rate, I think it's not longer relevant.  The only `platform` used is `mixedcontent`, in https://github.com/EFForg/https-everywhere/blob/99b8f9ceeeca4d3ad7eb5ec88e6e30e703588620/chromium/rules.js#L257-L266 and https://github.com/EFForg/https-everywhere/blob/99b8f9ceeeca4d3ad7eb5ec88e6e30e703588620/chromium/rules.js#L462-L471 - any other `platform` specification results in a `default_off` state, as @cschanaj points out.  As @J0WI has stated elsewhere, we probably want to make the option to enable `mixedcontent` rules more accessible, but these are the only places where there is logic around `platform` at all.  It may be worth including a small section on the `mixedcontent` platform, but this will have to be edited from the text in this PR - all browsers we support now block mixed content by default.

PR #14801, opened 2018-02-27
Add ruleset for radio-utopie.de

JulianHi:


DavidLiedke:
radio-utopie.de has HSTS already enabled. 
https://hstspreload.org/?domain=radio-utopie.de

This is the reason for the failed check. 

PR #14802, opened 2018-02-27
Add .mailmap file

sampablokuper:
In addition to making `git shortlog` output more comprehensible, a `.mailmap` file should make it easier, in future, to do things like create a list of contributors, or contact contributors if relicensing is ever needed.

This is just a start; there is probably more deduplication that could be added.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,3 +1,5 @@
+Aaron Swartz <me@aaronsw.com>                               Aaron Swartz <me@aaronsw.com>
+Aaron Swartz <me@aaronsw.com>                               aaronsw <me@aaronsw.com>
 Dan Auerbach <dan@eff.org>                                  Dan Auerbach <dan@eff.org>
 Dan Auerbach <dan@eff.org>                                  Dan Auerbach <dan@EFFs-MacBook-Pro.local>
 Dan Auerbach <dan@eff.org>                                  Dan Auerbach <dan@EFFs-MacBook-Pro-2.local>
--- a/.mailmap
+++ b/.mailmap
@@ -9,6 +9,9 @@ Jeremy Nation <jeremy@jeremynation.me>                      Jeremy Nation <jerem
 J0WI <J0WI@users.noreply.github.com>                        J0WI <J0WI@users.noreply.github.com>
 J0WI <J0WI@users.noreply.github.com>                        J0WI <sjw@gmx.ch>
 J0WI <J0WI@users.noreply.github.com>                        jowi <sjw@gmx.ch>
+MB <kunjurcca@gmail.com>                                    MB <kunjurcca@gmail.com>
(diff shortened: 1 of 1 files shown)


jeremyn:
I recognize the effort here but we don't need this at the moment, so it just introduces something that requires maintenance and can get out of date. If we do need it later, we can generate it with `git shortlog`, and we should also (at that time, not now) contact contributors to find out which address they would prefer if there are multiple choices.

So I think this should be closed.

sampablokuper:
@jeremyn wrote:
> If we do need it later, we can generate it with `git shortlog`

AFAIK there is no way to automatically generate a correct `.mailmap` from `git shortlog`. If you are aware of a way to do that, I would be glad to know it (and so might anyone who finds themselves needing to generate it in the future), so please could you state it in a comment here? Thanks :)

Alternatively, if I am correct in thinking that there is no automated way to do that, then please could we merge this after all, so that the effort required to bring the `.mailmap` up to date in future will be smaller (i.e. because it would build on my commits above, rather than requiring someone in the future to duplicate my effort).

Thanks again for another considered pull request review :)

jeremyn:
@sampablokuper We could start with a base like the output of `git shortlog -se` and then merge it all together manually or semi-manually, with the default primary identity being the one with the most commits. This wouldn't take long to do, if we needed it, so we don't need a partial version now.

Also, as I said, we should check with contributors about what identity they consider primary, especially if we are going to check that into the repository.

jeremyn:
And I should add, it's not worth contacting contributors for that information now, since we don't really need it.

sampablokuper:
@jeremyn wrote:
> We could start with a base like the output of `git shortlog -se` and then merge it all together manually or semi-manually, with the default primary identity being the one with the most commits.

That's essentially what I did: I selectively copied parts of the output from `git shortlog -nse` to `.mailmap`, one contributor at a time, ignoring contributors with only one name/email combo, and generating one commit per contributor added to `.mailmap`.

> This wouldn't take long to do, if we needed it, so we don't need a partial version now.

It takes longer than one might expect ;)

Although we don't *need* it now, it is nice to have. (It avoids the need to rely on GitHub for commit stats, for example.)

Moreover, avoiding duplication of effort seems worthwhile at any stage in a project: to act otherwise would seem to be a clear anti-pattern. So, do please reconsider. Thanks again :)

jeremyn:
@sampablokuper I'm not going to merge this for the reasons I've already said. Maybe one of the other maintainers will.

jeremyn:
That is, I disagree with this pull request but I'm not interested in discussing it further. If one of the other maintainers wants to work with you on it, I won't argue with them about it.

zoracon:
Closing due to changes and updates in Github since 2018 has made it easier and more clear for points made above.

PR #14807, opened 2018-02-27
[OSGeo.org] Add targets and simplify ruleset

Bisaloo:


src/chrome/content/rules/OSGeo.org.xml
```diff
@@ -1,106 +1,79 @@
...
+	Redirect to HTTP:
+		www2.osgeo.org
 
 -->
-<ruleset name="OSGeo.org (partial)">
+<ruleset name="OSGeo.org">
 
-	<!--	Direct rewrites:
-				-->
 	<target host="osgeo.org" />
```
ivysrono: https://travis-ci.org/EFForg/https-everywhere/jobs/586563465#L361
Lots expired.

PR #14817, opened 2018-02-28
Create Bison-Fute.gouv.fr.xml

Bisaloo:


src/chrome/content/rules/Bison-Fute.gouv.fr.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Mixed content:
+		m.bison-fute.gouv.fr
```
J0WI: I only found mixed passive content.

Bisaloo: I just checked and there is active mixed content as soon as you click on the map. It prevents many icons from loading which is an issue for an interactive map.

PR #14825, opened 2018-03-01
Bisaloo cert store

### LINKED ISSUE #1055: Update platform certs in https-everywhere-checker

### LINKED ISSUE #14636: Update certificates to latest NSS

Hainish:
fix #14636, fix #1055


Hainish:
Supersedes https://github.com/EFForg/https-everywhere/pull/14643.  This is rebased from `master`, which was necessary due to the indentation changes to `http_client.py`.

Hainish:
The output of `c_rehash` is necessary, as stated in the curl manpage:

       --capath <CA certificate directory>
              (SSL)  Tells  curl  to use the specified certificate directory to verify the peer. Multiple paths can be provided by separating them
              with ":" (e.g.  "path1:path2:path3"). The certificates must be in PEM format, and if curl is built against  OpenSSL,  the  directory
              must  have  been  processed  using the c_rehash utility supplied with OpenSSL. Using --capath can allow OpenSSL-powered curl to make
              SSL-connections much more efficiently than using --cacert if the --cacert file contains many CA certificates.

              If this option is set, the default capath value will be ignored, and if it is used several times, the last one will be used.

Otherwise you'll receive an error:

    SSL certificate problem: unable to get local issuer certificate

In order to ignore the system-level `ca-certificates.crt` when testing certs and *only* rely on the browser CAs (specified with `CURLOPT_CAPATH`), we need to set `CURLOPT_CAINFO` to `null`.  Unfortunately, `curl` throws an error if this is set to anything but a valid certificate.  This explains the need for a new `null.pem` file, which just the result of generating a self-signed certificate.

I've also cleaned up some scripts and added the Firefox version tag directly into the URL in the `SHA256SUMS` file.

Hainish:
Ordinarily I would squash and merge, but since we have multiple authors here we would lose attribution.  Since this is unkind, I could simply make an exception and create a merge commit this time.

Bisaloo:
Could you also please re-add `fix #14636, fix #1055` to the OP to auto-close relevant issues?

jeremyn:
I suppose my earlier testing without `c_rehash` (https://github.com/EFForg/https-everywhere/pull/14643#issuecomment-368265680) worked because it was falling back to the system certificates.

Suggestions as of https://github.com/EFForg/https-everywhere/pull/14825/commits/ad6ced4eb1f5746d5b088a8ce20246c3499f4d1c (latest commit):

* The `curl -I` command in `SHA256SUMS` needs quotes around `https://download.mozilla.org/?product=firefox-latest&os=linux64&lang=en_US` because of the `&`s.

* Instead of `null.pem`, you can use `platformPath` and `cert001.pem`. For example:

```python
import pathlib
...
            c.setopt(c.CAINFO, str(pathlib.Path(platformPath, 'cert001.pem')))
```

This looks fine to me otherwise.

It's okay if my contributions are squash-and-merged out of the log.

Hainish:
@Bisaloo 

> What would happen if you squash and merge? All my work has been overwritten and I don't mind being removed from the list of authors here.

I believe that it just lists the author of the last commit.

> Could you also please re-add fix #14636, fix #1055 to the OP to auto-close relevant issues?

Done.

@jeremyn 

> * The `curl -I` command in `SHA256SUMS` needs quotes around `https://download.mozilla.org/?product=firefox-latest&os=linux64&lang=en_US` because of the `&`s.

Done.


> * Instead of `null.pem`, you can use `platformPath` and `cert001.pem`. For example:
>
> ```python
> import pathlib
> ...
>             c.setopt(c.CAINFO, str(pathlib.Path(platformPath, 'cert001.pem')))
> ```

I originally did this and then switched it out for a generic placeholder, in order to be relatively agnostic to what's in the `platform_certs/default` path.  But reconsidering it, it's probably better to set it to `cert001.pem`, since it already exists.  The above will be invalid for testing `cacert` rules, though.  Try:

    $ test/manual.sh rules/LinuxTag.org.xml

So we have to specify the full path:

```python
c.setopt(c.CAINFO, str(pathlib.Path('test', 'rules', 'platform_certs', 'default', 'cert001.pem')))
```

Hainish:
Annoyingly, I pressed the wrong button and created a merge commit.

PR #14832, opened 2018-03-02
Update chromium/util.js

cschanaj:
Allow developers to see more verbose logging in the console (again). 

cc @Hainish

### DIFF
--- a/chromium/util.js
+++ b/chromium/util.js
@@ -14,7 +14,7 @@ var WARN = 5;
 // (Developers can just type DEFAULT_LOG_LEVEL=VERB in the console)
 var DEFAULT_LOG_LEVEL = NOTE;
 console.log("Hey developer! Want to see more verbose logging?");
-console.log("Type this into the console: DEFAULT_LOG_LEVEL=VERB");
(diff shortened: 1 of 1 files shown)


chromium/util.js
```diff
@@ -14,7 +14,7 @@ var WARN = 5;
 // (Developers can just type DEFAULT_LOG_LEVEL=VERB in the console)
 var DEFAULT_LOG_LEVEL = NOTE;
 console.log("Hey developer! Want to see more verbose logging?");
-console.log("Type this into the console: DEFAULT_LOG_LEVEL=VERB");
+console.log("Type this into the console: require.scopes.util.setDefaultLogLevel(require.scopes.util.VERB);");
```
Hainish: `scopes` shouldn't be exposed to the user - this is for internal `require` usage.  I think it better to use the following code in the message:

```js
let util = require('./util'); util.setDefaultLogLevel(util.VERB);
```

cschanaj:
@Hainish I agree. fixed in 4ce1d50. thanks

Hainish:
Thanks!  Btw @cschanaj I sent an email to you, if you haven't seen it.

Issue #7375, opened 2016-10-19
Ruleset test shouldn't fail for cacert if the website is marked with the cacert platform


numismatika:
I was editing the CCC-Cacert ruleset in  #6316 , which is marked with a platform tag of "CaCert".
So far so good, not everyone has imported their root cert.
The results of the checkin test script is confusing though.
https://travis-ci.org/EFForg/https-everywhere/builds/168084327 marks all domains as "error: self-signed"
although the Domains are listed in the cert properly.
In my humbe opinion the test should pass as the cert is valid (if you trust the root) for the domains in question.

koops76:
Add `default_off="cacert"` to the ruleset and it will work.

pipboy96:
`cacert` platform is deprecated now.

PR #14833, opened 2018-03-02
Restructure platform certs

Hainish:
Looking at it now, the entire way we test with `platform_certs` is strange.  It looks like the `cacert` platform is inconsistently tested.  Only `default` and the last platform loaded in `check_rules.py` is actually tested, leading to differing results on subsequent runs of the same ruleset.

This PR fixes this inconsistent behavior by restructuring the way `platform_certs` works.  If a ruleset specifies a `platform` attribute and a platform by that name exists under `platform_certs`, the certs in this directory are tested *in addition to* the default certificates.  I've updated the `cacert` certificates directory with the latest `cacert` root as well.  This seems to fix problems with `platform="cacert"` certificates such as `LinuxTag.org.xml`.

Originally I was going to add this to #14825 but these changes are significant enough to warrant another PR.


Hainish:
cc @jeremyn @Bisaloo

jeremyn:
I think we should re-evaluate whether to have a `cacert` `platform`.

The practical effect for setting a `platform` is that the ruleset is disabled by default, including `cacert`. For example, http://linuxtag.org from `LinuxTag.org.xml`, which has `platform="cacert"` and is not `default_off`, does not rewrite to HTTPS.

There are only twelve ruleset files with `platform="cacert"`. Only two of these, `Flaska.net.xml` and `LinuxTag.org.xml`, are not `default_off`. I propose we remove the `platform` from these two ruleset files, and instead add `default_off="cacert"`, or even a more generic `default_off="missing certificate chain"`. We could then remove all the `platform` stuff from the code, which would theoretically affect only twelve rulesets, but in practice only affect two, and that only when testing, not when using the extension normally.

Bisaloo:
Fix https://github.com/EFForg/https-everywhere/issues/7375

As @jeremyn pointed out, we only have 12 `cacert` rulesets (most of them concerning low traffic websites) so I don't really mind what happens with them. Many of them are also expired and can't be enabled even for people who installed the `cacert` root in their browser.

> We could then remove all the platform stuff from the code, which would theoretically affect only twelve rulesets, but in practice only affect two, and that only when testing, not when using the extension normally.

I'm not sure what you mean by "*all* the platform stuff" but we still have many `mixedcontent` rulesets.

jeremyn:
@Bisaloo Right, the `mixedcontent` stuff needs to stay, but it can be simplified into a simple boolean. In fact we could relabel the attribute from `platform` to `mixedcontent` and leave the existing values as `mixedcontent="mixedcontent"`.

Hainish:
Regardless of whether we change the way `platform` works, this restructures the `platform` logic to a sensible state right now.  The logic in `check_rules.py` is really convoluted.  I'm going to merge this as an incremental change - we can revisit the removal of `platform` later.

jeremyn:
@Hainish Does that mean you agree with removing the `cacert` `platform` and would merge a PR that did that?

Hainish:
@jeremyn I would be fine with the removal of the `cacert` platform, as long as the `platform` logic in `check_rules.py` is kept in place.

jeremyn:
PR https://github.com/EFForg/https-everywhere/pull/14875 removes the `cacert` `platform`.

PR #14840, opened 2018-03-02
SZ:

numismatika:
- re-enable ruleset

### DIFF
--- a/src/chrome/content/rules/Suddeutsche_Zeitung.xml
+++ b/src/chrome/content/rules/Suddeutsche_Zeitung.xml
@@ -1,10 +1,4 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://jetzt.sueddeutsche.de/ => https://jetzt.sueddeutsche.de/: (28, 'Connection timed out after 20000 milliseconds')
-Fetch error: http://stellenmarkt.sueddeutsche.de/ => https://stellenmarkt.sueddeutsche.de/: Too many redirects while fetching 'https://stellenmarkt.sueddeutsche.de/'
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Suddeutsche_Zeitung.xml
```diff
@@ -72,7 +66,6 @@ Fetch error: http://www.sueddeutsche.de/favicon.ico => https://www.sueddeutsche.
 		- maerkte
 		- motormarkt
 		- service
-		- stellenmarkt
```
J0WI: Please move it to the `Nonfunctional domains` section.

src/chrome/content/rules/Suddeutsche_Zeitung.xml
```diff
@@ -153,7 +147,6 @@ Fetch error: http://www.sueddeutsche.de/favicon.ico => https://www.sueddeutsche.
 	<test url="http://geoservice.sueddeutsche.de/" />
 	<test url="http://global.sueddeutsche.de/assets/img/header/sprite_header.png" />
 	<test url="http://id.sueddeutsche.de/" />
-	<test url="http://jetzt.sueddeutsche.de/" />
```
J0WI: Please move it to the `Nonfunctional` or `Problematic` domains section.

numismatika: it is gone. The youth magazine lives on www.jetzt.de

J0WI: The old host is still alive from here, but times out on https.

PR #14850, opened 2018-03-05
[Zerodayinitiative.com] supports HTTPS completely

xiaoyinl:
The ZDI (https://zerodayinitiative.com) no longer redirects HTTPS to HTTP, and supports HSTS, but not preloaded yet. This PR updates the rulesets accordingly.

https://www.ssllabs.com/ssltest/analyze.html?d=zerodayinitiative.com

Also the ZDI is now operated by Trend Micro Security. HP sold it out 2 years ago.

### DIFF
--- a/src/chrome/content/rules/Zero_Day_Initiative.com.xml
+++ b/src/chrome/content/rules/Zero_Day_Initiative.com.xml
@@ -1,3 +1,7 @@
+<!--
+		For other Trend Micro coverage, see Trend_Micro.xml. 
+-->
+
 <ruleset name="Zero Day Initiative.com">
 
 	<target host="zerodayinitiative.com" />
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/Zero_Day_Initiative.com.xml
```diff
@@ -1,25 +1,12 @@
 <!--
-	For other Hewlett-Packard coverage, see Hewlett-Packard.xml.
-
-
-	Some pages redirect to http.
-
+	For other Trend Micro coverage, see Trend_Micro.xml.
```
J0WI: Please also update the references in `Trend_Micro.xml`.

xiaoyinl: @J0WI Done. Thanks!

PR #14852, opened 2018-03-05
Update Chefkoch.de

DavidLiedke:


DavidLiedke:
@J0WI @Bisaloo Can someone tell me my mistake with the tests?

J0WI:
You can add an exclusion for `^http://static\.chefkoch-cdn\.de/$` to avoid the `Non-2xx HTTP` errors. Test urls must be `http`, not `https`.

DavidLiedke:
Thanks :)

PR #14853, opened 2018-03-05
Add Transfermarkt.de

DavidLiedke:


src/chrome/content/rules/transfermarkt.de.xml
```diff
@@ -0,0 +1,17 @@
...
+			transfermarkt.de
+-->
+<ruleset name="Transfermarkt.de">
+
+	<target host="www.transfermarkt.de" />
+	<target host="bfd.transfermarkt.de" />
+	<target host="jari.transfermarkt.de" />
+	<target host="lb01.transfermarkt.de" />
+	<target host="wap.transfermarkt.de" />
+	<target host="wwrw.transfermarkt.de" />
```
J0WI: Wildcard DNS

src/chrome/content/rules/transfermarkt.de.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+		404:
+			transfermarkt.de
```
J0WI: Can be redirected to `www` (but keep the comment to document the issue).

src/chrome/content/rules/transfermarkt.de.xml
```diff
@@ -0,0 +1,20 @@
...
+
+	<target host="transfermarkt.de" />
+	<target host="*.transfermarkt.de" />
+		<test url="http://www.transfermarkt.de/" />
+		<test url="http://www.transfermarkt.de/live/counter" />
+		<test url="http://www.transfermarkt.de/navigation/hauptnavi" />
+	
+	<rule from="^http://transfermarkt\.de/"
+			to="https://www.transfermarkt.de/" />
+			<test url="http://transfermarkt.de/" />
```
J0WI: This test url is already covered by the target.

src/chrome/content/rules/transfermarkt.de.xml
```diff
@@ -0,0 +1,20 @@
...
+		404:
+			transfermarkt.de
+-->
+<ruleset name="Transfermarkt.de">
+
+	<target host="transfermarkt.de" />
+	<target host="*.transfermarkt.de" />
+		<test url="http://www.transfermarkt.de/" />
+		<test url="http://www.transfermarkt.de/live/counter" />
+		<test url="http://www.transfermarkt.de/navigation/hauptnavi" />
```
J0WI: The idea of the 3 test urls requirement for wildcards is to have a bunch of subdomains covered. If `www` is the only subdomain you should avoid the wildcard target.

J0WI: Please add a hint about the wildcard DNS record in the comments.

src/chrome/content/rules/transfermarkt.de.xml
DavidLiedke: https://transfermarkt.de has no content and no redirect to the www-version.  Only http://transfermarkt.de redirects to the www-version. 

What should i do? Remove the target?

J0WI: The target and the rule are fine. Just remove the redundant test url.

DavidLiedke:
@J0WI done.

J0WI:
LGTM, Thanks!

PR #14858, opened 2018-03-06
[Wikimedia.xml] Remove blog.wikimedia.de from exclusion

xiaoyinl:
https://blog.wikimedia.de/ probably no longer has mixed content. It uses HTTPS links to blog pages on its home page, e.g. https://blog.wikimedia.de/2018/03/01/neues-urheberrecht-fuer-bildung-eine-erleichterung-fuer-die-praxis/.


J0WI:
This ruleset is `default_off` due some failing tests. Would you like to re-enable it?

J0WI:
Thanks!

PR #14864, opened 2018-03-07
Rename and update whisky.de.xml

christianheinrichs:


src/chrome/content/rules/Whisky.de.xml
```diff
@@ -0,0 +1,23 @@
...
+		- remote.whisky.de
+-->
+<ruleset name="Whisky.de">
+	<target host="whisky.de" />
+	<target host="www.whisky.de" />
+	<target host="news.whisky.de" />
+	<target host="proxy.whisky.de"/>
+	<target host="shophe.whisky.de"/>
+	<target host="typo3.whisky.de"/>
+	<target host="vanilla.whisky.de"/>
```
J0WI: This host uses a wildcard DNS record and cert.

christianheinrichs: Done.

PR #14865, opened 2018-03-07
Update FileFormat.info.xml

christianheinrichs:


src/chrome/content/rules/FileFormat.info.xml
```diff
@@ -1,17 +1,21 @@
...
+	403 Forbidden:
+		- samples.fileformat.info
 
-	Problematic domains:
+	MCB:
+		- blog.fileformat.info
 
-		fileformat.info		(connection refused)
+	Mixed content:
+		- blog.fileformat.info
```
J0WI: Duplicated comment

src/chrome/content/rules/FileFormat.info.xml
```diff
@@ -1,17 +1,21 @@
...
 
+	Unable to connect:
+		- fileformat.info
+		- fonts.fileformat.info
 -->
-
-<ruleset name="FileFormat.info (MCB)" platform="mixedcontent">
-
-	<target host="fileformat.info" />
```
J0WI: Target and redirect to `www` should be kept, because `^` is still broken on https.

src/chrome/content/rules/FileFormat.info.xml
christianheinrichs: I wanted to differentiate between the active and passive mixed content.

But since it appears on the same subdomain, the comments have been merged.

src/chrome/content/rules/FileFormat.info.xml
christianheinrichs: Done.

PR #14866, opened 2018-03-07
Add ruleset for findlawimages.com

roanhawkeye:


src/chrome/content/rules/Findlawimages.com.xml
```diff
@@ -1,5 +1,4 @@
 <ruleset name="Findlaw_Images">
-	<target host="findlawimages.com" />
```
J0WI: Please add a hint about the mismatch and the mixed content issues on the related `findlaw.com` page.

src/chrome/content/rules/Findlawimages.com.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="Findlaw_Images">
```
J0WI: Please avoid underscores in rulset names.

roanhawkeye: Thanks, I will change that.

J0WI: @roanhawkeye please see also my [previous comment](https://github.com/EFForg/https-everywhere/pull/14866#discussion_r178438665) and the related [chapter in our docs](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts).

src/chrome/content/rules/Findlawimages.com.xml
roanhawkeye: Hi , I added as a comment non working host, it is that ok?

PR #14868, opened 2018-03-08
Add ruleset for Lawyermarketing.com

roanhawkeye:


src/chrome/content/rules/Lawyermarketing.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="FindLaw_LawyerMarketing">
```
J0WI: Please avoid underscores in ruleset names.
LGTM otherwise.

J0WI: ping @roanhawkeye

roanhawkeye: Thanks, I will change that :)

PR #14871, opened 2018-03-08
Added ruleset for diakonie.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/Diakonie.de.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Invalid certificate:
+		www.info.diakonie.de
```
Bisaloo: I get `Secure connection failed`

PR #14876, opened 2018-03-09
Modify the platform check in test/script.py

jeremyn:
This pull request changes how `test/script.py` identifies the platform when setting `chromedriver_path` on a Linux system. There are two reasons for this change:

* The technical reason is that `platform.linux_distribution` [is deprecated](https://docs.python.org/3.6/library/platform.html#platform.linux_distribution) as of Python 3.5.

* The actual reason is that `platform.linux_distribution` [doesn't correctly identify Ubuntu](https://stackoverflow.com/a/33997262) if you have built Python 3.6 from source, because you are, for example, using Ubuntu 16.04. As the link mentions, the Ubuntu developers patch the `platform` module that Ubuntu provides so it works as expected on Ubuntu. With the `platform` module built from source, the HTTPS Everywhere Chromium test process looks for `chromedriver` in the wrong place, and fails.

### DIFF
--- a/test/script.py
+++ b/test/script.py
@@ -35,11 +35,9 @@ if sys.argv[1] == "Chrome":
     # Find the path to chromedriver
     chromedriver_path = "chromedriver"
     if sys.platform.startswith("linux"):
-        if 'Ubuntu' in platform.linux_distribution():
+        if 'Ubuntu' in platform.uname().version:
             chromedriver_path = "/usr/lib/chromium-browser/chromedriver"
-        elif 'debian' in platform.linux_distribution():
-            #Debian is lowercase when platform.linux_distribution() is used.
-            #This is not a mistake.
+        elif 'Debian' in platform.uname().version:
             chromedriver_path = "/usr/lib/chromium/chromedriver"
 
     try:
--- a/test/script.py
+++ b/test/script.py
@@ -10,7 +10,7 @@
 # of linux is required for the script to run correctly as well.
 # Otherwise, use pyvirtualdisplay.
 
-import sys, os, platform, time
+import sys, os, time
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
--- a/test/script.py
+++ b/test/script.py
@@ -35,10 +35,14 @@ if sys.argv[1] == "Chrome":
     # Find the path to chromedriver
     chromedriver_path = "chromedriver"
     if sys.platform.startswith("linux"):
-        if 'Ubuntu' in platform.uname().version:
-            chromedriver_path = "/usr/lib/chromium-browser/chromedriver"
-        elif 'Debian' in platform.uname().version:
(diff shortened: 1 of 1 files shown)


jeremyn:
Ping @Hainish .

Hainish:
@jeremyn I'm using Debian 8 with python3.6 compiled from source and getting the following:

```
user@https-everywhere ~/workspace/https-everywhere/chromium (master) $ python3.6
Python 3.6.4 (default, Feb  9 2018, 18:09:54) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.uname().version
'#1 SMP Tue Oct 17 23:58:50 UTC 2017'
>>> 
```

Apparently there's a [new module](http://miroslav.suchy.cz/blog/archives/2016/10/27/how_to_replace_deprecated_platform_distro/index.html) that can be used as a drop-in replacement:

```
(.venv3) user@https-everywhere ~/workspace/https-everywhere/chromium (master) $ pip install distro
Collecting distro
  Downloading distro-1.2.0-py2.py3-none-any.whl
Installing collected packages: distro
Successfully installed distro-1.2.0
(.venv3) user@https-everywhere ~/workspace/https-everywhere/chromium (master) $ python3.6
Python 3.6.4 (default, Feb  9 2018, 18:09:54) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import distro
>>> distro.linux_distribution()
('Debian GNU/Linux', '8', 'jessie')
>>> distro.linux_distribution(full_distribution_name=False)
('debian', '8', 'jessie')
```

jeremyn:
For me, `uname -a` on both Debian 8 and 9 has `Debian` in the output. I'm not sure what to say beyond that.

Instead of checking for the platform, let's instead check for the existence of `chromedriver` in various locations. So, if `/usr/lib/chromium-browser/chromedriver` exists, we use that, otherwise if if `/usr/lib/chromium/chromedriver` exists, we use that, otherwise we just use `chromedriver`. What do you think?

Hainish:
Yes, that sounds like a good solution.

On March 10, 2018 9:08:57 AM PST, Jeremy Nation <notifications@github.com> wrote:
>For me, `uname -a` on both Debian 8 and 9 has `Debian` in the output.
>I'm not sure what to say beyond that.
>
>Instead of checking for the platform, let's instead check for the
>existence of `chromedriver` in various locations. So, if
>`/usr/lib/chromium-browser/chromedriver` exists, we use that, otherwise
>if if `/usr/lib/chromium/chromedriver` exists, we use that, otherwise
>we just use `chromedriver`. What do you think?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/EFForg/https-everywhere/pull/14876#issuecomment-372045539

-- 

jeremyn:
Done.

jeremyn:
Done (again).

Hainish:
lgtm, thanks!

PR #14883, opened 2018-03-11
add ruleset dragongoserver.net

gimpf:
This PR adds a rule for https://www.dragongoserver.net. To be honest, given the lack of a suitable dev-environment at the moment, currently untested, though I hope the eight lines are mostly acceptable as they are.

BR


src/chrome/content/rules/Dragongoserver.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Dragongoserver.net">
```
J0WI: Please also add `dragongoserver.net` (without `www`)

PR #14885, opened 2018-03-11
fix orf.at rules for TEST=fetch; add rule for currently-broken teletext.orf.at

girst:
the rules already contain a redirect for appfeeds.orf.at, which is
requested from teletext.orf.at. If one naviates to
`http://teletext.orf.at`, HTTPS-Everywhere upgrades the API request from
`http://appfeeds.orf.at` to `https://appfeeds.orf.at`, which results in the
following CORS-error:

    Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://appfeeds.orf.at/teletext/api/v2/mobile/channels/orf1/pages/100. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

redirecting teletext.orf.at as well fixes that.

~~**EDIT:** `test=FETCH` fails, because the rule contains redirects for other subdomains that are currently unavailable. I do not want to delete them, since I don't know why they are not available at this time (might be just the `/` path that has not content) or if they will be publicly accessible in the future again.~~ should be fixed now. I went through all of them; see commit message for details

### DIFF
--- a/src/chrome/content/rules/ORF.at.xml
+++ b/src/chrome/content/rules/ORF.at.xml
@@ -26,6 +26,7 @@
 		
 	Cert expired:
 		- oe3verkehr.orf.at
+		- content.orf.at
 
 	Cert mismatch:
 		- burgenlandkalender.orf.at
--- a/src/chrome/content/rules/ORF.at.xml
+++ b/src/chrome/content/rules/ORF.at.xml
@@ -212,6 +212,7 @@
 	<target host="sms.orf.at" />
 	<target host="speedy.orf.at" />
 	<target host="sts.orf.at" />
+	<target host="teletext.orf.at" />
 	<target host="thredbo.orf.at" />
 	<target host="tickets.orf.at" />
 	<target host="tobago.orf.at" />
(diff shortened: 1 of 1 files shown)


girst:
I've got no idea why `TEST="chrome beta" BROWSER=google-chrome-beta` fails :|

src/chrome/content/rules/ORF.at.xml
```diff
@@ -118,33 +131,25 @@
...
 	<target host="boa.orf.at" />
 	<target host="bolek.orf.at" />
 	<target host="book.orf.at" />
 	<target host="booking.orf.at" />
 	<target host="bpo.orf.at" />
-	<target host="brain.orf.at" />
 	<target host="buergerforumdb.orf.at" />
 	<target host="cloud.orf.at" />
-	<target host="clyde.orf.at" />
-	<target host="content.orf.at" />
```
J0WI: Please add a comment for it

J0WI:
Just keep trying if chrome fails...

src/chrome/content/rules/ORF.at.xml
girst: content.orf.at redirects to https, then to a microsoft-cloud-login (which I presume only orf employees have access). what kind of comment should I add for that? the commit message mentions already that it redirects itself.

J0WI: You can add it to the `Cert expired` section

PR #14886, opened 2018-03-11
Fix settings on new versions of Chrome.

s4y:
On Chrome Canary, the checkboxes in the HTTPS Everywhere popup are missing and this error shows up in the background page's console every time the popup loads: 

> `Error in event handler: TypeError: Error processing argument at index -1, conversion failure from undefined`

It looks like this is because `chrome.storage`'s methods no longer work unbound. The fix is to bind them to the `StorageArea ` when they get pulled out.

### DIFF
--- a/chromium/background-scripts/store.js
+++ b/chromium/background-scripts/store.js
@@ -58,8 +58,8 @@ async function performMigrations() {
 
 function setStorage(store) {
   Object.assign(exports, {
-    get: store.get,
-    set: store.set,
+    get: store.get.bind(store),
+    set: store.set.bind(store),
     get_promise,
     set_promise
   });


jeremyn:
@s4y This is a good catch, and your fix makes intuitive sense. However I think it would be better to rewrite `store.js` to avoid pass-through calls to `get`/`set` entirely. Your fix depends on the internals of `get`/`set` , and it's not clear why this change is needed now, or how it would affect Firefox. Calling `get`/`set` directly on the storage object would be safer.

For example, you could set some variable `storage` to either `chrome.storage.local` or `chrome.storage.sync` before defining the functions, and then call `storage.get`/`storage.get` in the two `_promise` functions. I don't think the current `exports.get`/`exports.set` are used outside of this file.

However, @Hainish is likely going to be the one to review this pull request, so please feel free to wait for his feedback before making any changes.

Hainish:
Since `bind` is a general method of the function prototype, I think it's fine to use in this context.  This may be a problem which Chrome fixes on its own as Canary is considered experimental.  Let's see if this is fixed by the time 67 hits dev.

s4y:
Sounds reasonable. I just filed a crbug to find out:

https://bugs.chromium.org/p/chromium/issues/detail?id=821654

s4y:
@Hainish Looks like we've got an [answer](https://bugs.chromium.org/p/chromium/issues/detail?id=821654#c4) on the crbug:

> For the HTTPS Everywhere case, I'd recommend that they update to bind.  And note that updating to bind should work whether we keep the new behavior or decide it's worth the added investment to maintain strict backwards compatibility with the unspec'd behavior.

Hainish:
Good enough for me.  Thank you!

s4y:
\o/

PR #14888, opened 2018-03-12
[SublimeGit.net] Default_off ruleset

Bisaloo:


jeremyn:
https://release.sublimegit.net/ and https://www.sublimegit.net/ are `secure connection failed`, not `expired`. I'm not sure how to mark this, maybe `default_off="expired, secure connection failed"` ?

Bisaloo:
I would still keep `expired`. It is the problem for most of the subdomains and the `SSL_ERROR_NO_CYPHER_OVERLAP` is somewhat related (in that in all cases, the site owner hasn't updated the SSL config in a long time). It's also not so rare for rulesets to be `default_off` with a reason that does not apply to every subdomain.

If you don't want to use `default_off="expired"`, we should probably use `other` (in accordance with #9906) and explain the specific reason for each subdomain in the top comment.

jeremyn:
There are four domains, two of which are `expired`, and the other two `ssl-error` (`SSL_ERROR_NO_CYPHER_OVERLAP`) so we can't say most of the subdomains have one problem or the other. Also `SSL_ERROR_NO_CYPHER_OVERLAP` is distinct from an expired certificate, we should not group them together if we are going to have the two categories at all.

(I do think we have too many failure categorizations generally, but that is another discussion. If we're going to have the different categories, then we should use them appropriately.)

Let's mark it as `default_off="other"` and note the domains as failed for whatever reason in the top comment, like we would normally do for a non-`default_off` ruleset.

jeremyn:
Please remove `(expired)` and `(SSL_ERROR_NO_CYPHER_OVERLAP)` from the top comment. These are just extra things that have to be checked and can go out of date, and don't provide useful information.

PR #14893, opened 2018-03-12
Fix openssl sha256 command

christianheinrichs:
When testing via Tor browser and then closing it normally, the following message appears:

```bash
++ openssl sha -sha256 pkg/https-everywhere-2018.2.26~pre-eff.xpi
Invalid command 'sha'; type "help" for a list.
+ shasum=
```

It does not occur on a keyboard interrupt with <kbd>CTRL</kbd>+<kbd>C</kbd>.

This PR fixes the issue.

### DIFF
--- a/test/tor-browser.sh
+++ b/test/tor-browser.sh
@@ -5,13 +5,12 @@
 
 if [ -n "$GIT_DIR" ]
 then
-    # $GIT_DIR is set, so we're running as a hook.
-    cd $GIT_DIR
+  # $GIT_DIR is set, so we're running as a hook.
+  cd $GIT_DIR
 else
-    # Git command exists? Cool, let's CD to the right place.
-    git rev-parse && cd "$(git rev-parse --show-toplevel)"
+  # Git command exists? Cool, let's CD to the right place.
+  git rev-parse && cd "$(git rev-parse --show-toplevel)"
 fi
-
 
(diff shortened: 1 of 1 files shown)


J0WI:
@Hainish I think this is ready to merge.

jeremyn:
According to [this](https://apple.stackexchange.com/questions/230917/need-a-cli-to-check-the-sha256-hash-of-a-file#comment281981_230919), `openssl sha -sha256` (the previous way) works on Mac but `openssl sha256` (the new way) doesn't. I don't know what OpenSSL version the commenter is using.

Also, both ways seems to work for me on Ubuntu 16.04 with 1.0.2g.

Can someone link to a changelog entry or something between 1.0.2g and 1.1.0g that describes why this change is needed?

christianheinrichs:
As I don't have access to a macOS system, I can't confirm whether or not the syntax in the linked comment works.

This PR has been tested on Debian Stretch with `openssl` version [1.1.0f](https://packages.debian.org/stretch/openssl) and `openssl sha -sha256` continues to fail.

A slightly longer substitute for `openssl sha256` would be the command `openssl dgst -sha256`, which does work on Debian Stretch. Not sure if it is cross-compatible.

Hainish:
Thank you!

PR #14919, opened 2018-03-17
Fix ruleset for Hatena

itochan:
I fixed replace pattern.
Currently, broken CSS loading. http://www.hatena.ne.jp/

* For example, from http://www.hatena.ne.jp/css/hatenaportal/portal2012.css?t=1487038420 redirect to...
  * Before: https://www.hatena.ne.jp/hatenaportal/portal2012.css?t=1487038420 (404)
  * After: https://www.hatena.ne.jp/css/hatenaportal/portal2012.css?t=1487038420


J0WI:
Thanks!

PR #14929, opened 2018-03-18
Add Iemoji.com.xml

J0WI:


Bisaloo:
Invalid certificate for pix.iemoji.com. See for example http://pix.iemoji.com/images/emoji/apple/ios-9/256/thinking-face.png

J0WI:
Added.

PR #14931, opened 2018-03-18
Add Getemoji.com.xml

J0WI:


Bisaloo:
I get MCB on `blog.getemoji.com` but it looks like it's only ads and social trackers so it should be fine.

PR #14940, opened 2018-03-19
Added ruleset for javaland.eu

Jochen-A-Fuerbacher:


src/chrome/content/rules/JavaLand.eu.xml
```diff
@@ -0,0 +1,18 @@
...
+		test.javaland.eu
+-->
+<ruleset name="JavaLand.eu">
+
+	<target host="javaland.eu" />
+	<target host="www.javaland.eu" />
+	<target host="programm.javaland.eu" />
+
+	<!-- Invalid certificate on https://javaland.eu/,
+	http://javaland.eu/ redirects to http://www.javaland.eu/ -->
```
Bisaloo: I get `Connection closed` / `Secure connection closed` (which is the same error but presented differently depending on your browser).
It would also be better if you could put this in the top comment for consistency.

Bisaloo:
I'm sorry I wasn't clear. Can you use the same format as your other PR? ie:

```xml
	Connection closed:
		javaland.eu
```

We don't need the part about `http://javaland.eu` redirecting to `www`

PR #14943, opened 2018-03-20
added glowingbear.org

bryceml:


Bisaloo:
Please also add `latest.glowingbear.org`

PR #14924, opened 2018-03-17
Batch removal of duplicated test urls

cschanaj:
@Hainish A dirty command is used to remove duplicated test urls found; `log.txt` is the Travis output of the failing test in #14660 

Block #14660

```bash
TRAVIS_REPLACE='s/^ERROR \([^:]*\): Duplicated test URL found \(.*\)$/\1 \2/g'

grep ^ERROR ~/log.txt  | tr -d '\r' | sed "$TRAVIS_REPLACE" | sort -u | while read FILE HOST; 
do 
    REGEX_ESCAPED_HOST=`python3.6 -c "import re; print(re.escape('$HOST'))"`;  
    sed -i "0,/<test.*\"$REGEX_ESCAPED_HOST\"/{//d;}" $FILE; 
done
```

See https://stackoverflow.com/questions/23696871/how-to-remove-only-the-first-occurrence-of-a-line-in-a-file-using-sed

### TODO

16 rulesets failed `fetch-test`

- [x] Destructoid.xml (3672bc6)
- [x] Hdfcbank.com.xml (59c6493)
- [x] Komoona.com.xml (115b6d7)
- [x] PropellerAds.xml (8888416) 
- [x] WP-Engine.xml (fddbf28)
- [x] Wemfbox.ch.xml (f0ad991)
- [x] cloud66.com.xml (f8e49bf)
- [x] greenhouseci.com.xml (fc55268)
- [ ] JD.com-Problematic.xml
- [ ] espivblogs.net.xml
- [ ] Twenga.com.xml
- [ ] Fenland.gov.uk.xml
- [ ] Wandoujia.xml
- [ ] luosimao.com.xml
- [ ] tanx.com.xml

6 rulesets failed `validations-test`
- [x] Imgs_Mail.ru.xml (d74a705)
- [x] Threema.xml (895c094)
- [x] Twenga.com.xml (c79806c)
- [x] Wemfbox.ch.xml (9f2573e)
- [x] Fenland.gov.uk.xml (f1075ff)
- [x] Komoona.com.xml (870e2d7)

2 rulesets having more than 1 duplicates
- [x] CloudAtCost.com.xml (628b7f5)
- [x] westberks.gov.uk.xml (e59ff8b)

PR #14944, opened 2018-03-20
Rename and Update Squat.gr.xml

cschanaj:
Subdomains are found using the following command:

`$ python sublist3r.py -e Google -d squat.gr`


cschanaj:
@J0WI would you also have a look on this PR? this is a follow up PR of #14924 adding back `squat.gr` to the ruleset base.

PR #14950, opened 2018-03-20
Official Enjin Ruleset

Bradez:
Primary Domain: enjin.com

Additional Domains:
- enjincoin.io
- enjinwallet.io
- enjinx.io
- efinity.io

### DIFF
--- a/src/chrome/content/rules/Enjin.xml
+++ b/src/chrome/content/rules/Enjin.xml
@@ -1,47 +1,64 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://dracokraft.enjin.com/ => https://dracokraft.enjin.com/: Too many redirects while fetching 'https://dracokraft.enjin.com/'
+	IMPORTANT INFORMATION:
 
-	CDN buckets:
-		- s3.amazonaws.com/files.enjin.com/
+	Sites, developed by the public, created on the Enjin Network have the ability to force
+	HTTP or HTTPS themselves. If they enable HTTPS, the site will automatically force a
+	HTTPS connection. You cannot add a rule to force HTTPS for sites that specifically
+	request to force the HTTP protocol, doing so will cause a redirect loop
+	(between the HTTP and HTTPS protocols) to occur.
 
-		- d2umfjgc9kwn9b.cloudfront.net
-			- assets-cloud.enjin.com
+	This functionality may be altered in the near future to ensure all sites use HTTPS and proxy
+	insecure material (such as scripts, stylesheets and images) through a HTTPS-enabled proxy.
 -->
-<ruleset name="Enjin" platform="mixedcontent" default_off='failed ruleset test'>
+
+<ruleset name="Enjin" platform="mixedcontent">
+
+	<!-- Enjin Network -->
 	<target host="enjin.com" />
-	<target host="*.enjin.com" />
-	<target host="resources.guild-hosting.net" />
-	
-	<test url="http://www.enjin.com/" />
-	<test url="http://assets-cloud.enjin.com/" />
-	<test url="http://support.enjin.com/" />
-	<test url="http://files.enjin.com/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Enjin.xml
```diff
@@ -1,47 +1,64 @@
...
+
+	<rule from="^http://([\w-]+\.)?enjinx\.io/" to="https://$1enjinx.io/" />
+
+	<!-- Efinity -->
+	<target host="efinity.io" />
+	<target host="www.efinity.io" />
+
+	<rule from="^http://([\w-]+\.)?efinity\.io/" to="https://$1efinity.io/" />
+
+	<!-- CDN Rules -->
```
J0WI: Please add a short comment, that they serve a mismatching cert.
Please also try to provide some `<test url="" />` for all of them.

src/chrome/content/rules/Enjin.xml
```diff
@@ -1,47 +1,64 @@
...
+	(between the HTTP and HTTPS protocols) to occur.
 
-		- d2umfjgc9kwn9b.cloudfront.net
-			- assets-cloud.enjin.com
+	This functionality may be altered in the near future to ensure all sites use HTTPS and proxy
+	insecure material (such as scripts, stylesheets and images) through a HTTPS-enabled proxy.
 -->
-<ruleset name="Enjin" platform="mixedcontent" default_off='failed ruleset test'>
+
+<ruleset name="Enjin" platform="mixedcontent">
```
J0WI: Are there any serious mixedcontent issues on the sites? I didn't found any so far.

J0WI:
Thanks for providing this update! I hope you're able to switch to HTTPS soon, at best you also manage to get directly [preloaded](https://hstspreload.org/) in browsers.

src/chrome/content/rules/Enjin.xml
Bradez: The user's are capable of adding their own content which may cause mixed content, the official website has a flag to automatically upgrade all of our resources to HTTPS so you won't see mixed content on our main website.

src/chrome/content/rules/Enjin.xml
Bradez: This is a new website, owned by Enjin, it has no additional rules but the point is that the website will fully utilise HTTPS.

Bradez:
Thank you for your comments, this is definitely something we're looking into although - given it's 9 years of work and we have many paying customers - it's not as simple to enforce HTTPS on all of the sites although we are very much seeking to do so in the upcoming future in which case we will definitely be utilising HSTS and preloading.

src/chrome/content/rules/Enjin.xml
```diff
@@ -1,47 +1,64 @@
...
+	<target host="www.enjinwallet.io" />
+
+	<rule from="^http://([\w-]+\.)?enjinwallet\.io/" to="https://$1enjinwallet.io/" />
+
+	<!-- EnjinX -->
+	<target host="enjinx.io" />
+	<target host="www.enjinx.io" />
+	<target host="api.enjinx.io" />
+
+	<test url="http://api.enjinx.io/" />
```
Bradez: I will remove this test URL for the time being as the target is intended to be a wildcard yet we only have 2 test URLs (www and api) so can't utilise the wildcard target. Should we ever change that, will be certain to ensure the target host is changed and test URLs added.

src/chrome/content/rules/Enjin.xml
J0WI: Just to clarify: this comment was referring to the CDN hosts below.
If your S3 bucket changed for some reason, test urls will ensure that we are aware that the location has changed.

Bradez: I apologise, I missed that in my initial review of your comments, I'll add some test URLs for the CDN along with a comment. Thank you.

PR #14954, opened 2018-03-21
[Indiana.edu] Add exclusion

Bisaloo:


jeremyn:
I'd like to avoid using a personal URL like `~kruschke` as a `test`. How about `http://www.indiana.edu/~oah/` ?

Also, some of the `~` domains do support HTTPS, such as https://www.indiana.edu/~bands/ . I think that an `exclusion` for all of them is better than breaking some of them.

Bisaloo:
Yes, I noticed that some `~` paths work fine with HTTPS but I don't think it is possible to have an exclusion that differentiate them from broken paths.

jeremyn:
Thanks, merged.

PR #14960, opened 2018-03-21
Added ruleset for baf.bund.de

Jochen-A-Fuerbacher:


Bisaloo:
Does it make sense to keep a list of related rulesets in `Verwaltung_Online.xml`? 

See [this comment](https://github.com/EFForg/https-everywhere/pull/11510#issuecomment-318866310) from @jeremyn about `*.gov` targets. In the same idea, the domain `gouv.fr` for the French government for example should actually be considered as a TLD.

What do you think?

Jochen-A-Fuerbacher:
Well, I just continued the list at `Verwaltung_Online.xml` someone else started. In other pull requests I was told to include a reference at `Verwaltung_Online.xml` so I did it (see [this comment](https://github.com/EFForg/https-everywhere/pull/7475#issuecomment-280641959) by @J0WI ). I am fine to remove the list at `Verwaltung_Online.xml` but I think, this should be done in another pull request.

@J0WI What to you think about this?

jeremyn:
@J0WI and I have disagreed about comments like this in the past. See PR https://github.com/EFForg/https-everywhere/pull/10468 "separate ag.ch" where we debated whether `swissgov.xml`, for domains related to the national Swiss government, should have a reference to `ag.ch.xml`, which is for domains related to a particular local Swiss political entity that is, as far as I know, unrelated to the national government. We didn't reach an agreement on the issue, but instead I let it go to keep the PR moving forward.

In this case I'm okay with @Jochen-A-Fuerbacher updating an existing list, not because I agree with it, but because I don't want to put @Jochen-A-Fuerbacher in the middle of this disagreement.

Bisaloo:
I agree this belongs to another PR anyways so let's move forward with this one.

Can you also add targets for https://www.anlagenschutz.baf.bund.de and http://test.anlagenschutz.baf.bund.de please?

PR #14967, opened 2018-03-22
[Livestatserver.com] Remove wildcard

Bisaloo:


src/chrome/content/rules/Livestatserver.com.xml
```diff
@@ -1,19 +1,20 @@
...
-		- staging *
-		- w1 *
-
-	* Works; mismatched, CN: ssl.livestatserver.com
+	No working URL known:
+		cdn.livestatserver.com
+		images.livestatserver.com
+		ssl.livestatserver.com
+		staging.livestatserver.com
+		w1.livestatserver.com
```
J0WI: Ah, I just found `/login.php`

J0WI: And `/view.php` and `http://luckyorange.livestatserver.com/in-the-news.php`

PR #14968, opened 2018-03-22
Create TablesGenerator.com.xml

Bisaloo:


jeremyn:
I disagree with adding new rulesets that are `default_off`. I think this PR should be closed.

Bisaloo:
I tried reaching out to the site owner but their contact form captcha is broken.

PR #14970, opened 2018-03-22
Create TeXblog.org.xml

Bisaloo:


jeremyn:
- [ ] Change the `ruleset` `name` and filename to `Texblog` instead of `TeXblog`. I know "TeX" is written with the capital "X", but the site owner seems to style the name of their site as "texblog", which for us is "Texblog" since we capitalize the first letter.

PR #14990, opened 2018-03-25
Added ruleset for gnunet.org

Jochen-A-Fuerbacher:


numismatika:
Seems to be completely preloaded in the browsers and can probably be closed

PR #14994, opened 2018-03-26
Create chuapp.com.xml

ivysrono:


Bisaloo:
`Sublist3r` finds a small number of additional subdomains. Could you add them as well please?

Bisaloo:
Sorry I wasn't clear. If MCB does not affect functionality, we can keep it as a target. I was just asking you if you could confirm it did not have any visible effect in your testing.

PR #14995, opened 2018-03-26
Remove duplicates from Cherwell.gov.uk.xml

cschanaj:


src/chrome/content/rules/Cherwell.gov.uk.xml
```diff
@@ -1,91 +1,25 @@
...
-	Problematic hosts in *cherwell.gov.uk:
-
-		- consult ᵐ
-
-	ᵐ Mismatched
-
+	Non-functional hosts
+		Timeout was reached:
+		- icm.cherwell.gov.uk
+		- motbookings.cherwell.gov.uk
```
Bisaloo: This target times out over HTTP as well. I think we can remove it.

Bisaloo:
Although I don't really with this kind, you should also add a reference to `GOV.UK.xml` (this ruleset is already referenced there).

PR #14996, opened 2018-03-26
Remove duplicates from Inc.com.xml

cschanaj:


src/chrome/content/rules/Inc.com-falsemixed.xml
```diff
@@ -1,20 +0,0 @@
-<!--
-	For rules not causing MCB, see Inc.com.xml.
-
--->
-<ruleset name="Inc.com (mixed content)" platform="mixedcontent">
-
-	<target host="inc.com" />
-	<target host="conference.inc.com" />
-	<target host="mediakit.inc.com" />
```
Bisaloo: This target still has MCB. Can we keep it in this ruleset?

src/chrome/content/rules/Inc.com.xml
```diff
@@ -1,68 +1,19 @@
...
 	<target host="inc.com" />
-	<target host="*.inc.com" />
-		<!--
-			Avoid broken MCB:
-						-->
-		<exclusion pattern="^http://(?:conference|mediakit)\.inc\.com/+(?!wp-content/|wp-includes/)" />
-		<exclusion pattern="^http://(?:www\.)?inc\.com/+(?!favicon\.ico|lib/|views/css/)" />
-
+	<target host="www.inc.com" />
+	<target host="conference.inc.com" />
```
Bisaloo: For the record, this target has MCB but the server redirects to HTTPS so I'm fine with keeping it.

src/chrome/content/rules/Inc.com-falsemixed.xml
cschanaj: I would like to avoid creating new `mixedcontent` rulesets since they are disabled by default anyway.

PR #14997, opened 2018-03-26
Remove duplicates from Tripwire.com.xml

cschanaj:


Bisaloo:
I still have MCB but the website itself redirects to HTTPS so I'm fine with fine. I take it you are not interested in adding more subdomains?

cschanaj:
@Bisaloo yes, I would like this merged as-is if this is okay.

PR #15005, opened 2018-03-27
[ChromeStatus.com] Update regex style

Bisaloo:
The main goal of this PR is to remove this ruleset from the whitelist.


jeremyn:
`Secure connection failed` is a more typical description than `Connection closed`, though what you have is fine if you want to keep it.

This appears to be a Google site, for example see https://github.com/GoogleChrome/chromium-dashboard/blob/master/README.md , so it might as well stay in the list of Google sites at `GoogleServices.xml`. You should remove the `_` though.

You did not actually remove this from `ruleset-whitelist.csv`.

Bisaloo:
Yes, I know I didn't remove it from the whitelist. But I changed the file hash so it will be automatically removed in the next release.

PR #15006, opened 2018-03-27
Update DENX.de.xml

christianheinrichs:


src/chrome/content/rules/DENX.de.xml
```diff
@@ -1,13 +1,34 @@
...
+		- denx.de
+		- www.denx.de
+		- theia.denx.de
 
-	<rule from="^http:"
-		to="https:" />
+	Service unavailable:
+		- jabber.denx.de
+-->
+<ruleset name="DENX.de" platform="mixedcontent">
```
J0WI: Is there any active mixed content? Otherwise the `mixedcontent` can be removed.

christianheinrichs: There are two CSS files being loaded via HTTP on the `theia` subdomain, but since it's not listed as target anymore, `mixedcontent` has been removed.

J0WI: Thanks!

Issue #14936, opened 2018-03-19
Standard for Non-working Hosts?


wonderchook:
Type: feature request

Currently, there is not a standard for the comments regarding non-working hosts. This inhibits automation of ruleset testing. By creating a parsable standard it would enable automation and ease some of the manual role in ruleset testing. 

[Documentation on nonworking hosts
](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts)

It seems a little weird for this to remain as a comment and maybe it should have a standard elsewhere i the XML.

numismatika:
First problem i see is that sometimes people just cut of the common part of the domain
Instead of some.subdomain.domain.com and another.subdomain.domain.com 
the domain.com is left out and it becomes
```
Cert mismatch on:
- some.subdomain
- another.subdomain
```

wonderchook:
I definitely don't think this is an easy fix. I think if I was going to implement it I would suggest coming up with a standard and applying for new hosts to start. Then as hosts are modified add them to the standard. It would be a slow path, but at least would make verifying new hosts easier.

jeremyn:
There are open questions about what value there is marking non-working hosts, and what comments should be allowed in rulesets generally. Until those questions are answered and affirmatively agreed to by all the active ruleset maintainers, this sort of documentation effort, while a good idea and meaning no disrespect to @wonderchook , is unlikely to be successful.

PR #15014, opened 2018-03-28
[StockTwits.com] Simplify ruleset

### LINKED ISSUE #2586: stocktwits.com does not work when https-everywhere is enabled
Only an uncomplete, unstyled version of stocktwits.com loads when httpseverywhere is anabled

Bisaloo:
fix #2586


jeremyn:
This looks fine.

One thing I like to do is organize the top comment categories alphabetically, so for example `Time out` would go after `Redirect to HTTP`. If you want to make that change then go ahead, but if not just let me know and I'll merge this as it is.

Bisaloo:
> One thing I like to do is organize the top comment categories alphabetically, so for example Time out would go after Redirect to HTTP. If you want to make that change then go ahead, but if not just let me know and I'll merge this as it is.

I'm fine with that but there is no point in doing for some rulesets only. If you want to do this from now on, you should mention it in #14936 so that we start enforcing it on all new PR.

jeremyn:
Merged.

This has been my own informal guideline for a while. In practice, for a while now there have only been two active maintainers, me and @J0WI, and (no disrespect intended) @J0WI has a much more relaxed attitude than I do toward comment requirements. So I'm not interested in having a discussion to attempt to compel @J0WI to do this a certain way.

There has also been resistance to over-documenting contributing guidelines and I doubt whether adding some sorting rule would actually make contributors change their behavior. Also I probably wouldn't comment on a PR that was otherwise correct, if this was the only thing wrong with it. (In this case it's worth mentioning since you're a maintainer and you can guide other contributors in your own reviews if you agree.)

What you can do as a maintainer if you think this is a good rule is to give to-do items in sorted order, and also sort your own examples and rulesets. Contributors will naturally do things in the order you ask them to do.

PR #15031, opened 2018-03-30
Update Samsung.xml

cschanaj:

PR #15032, opened 2018-03-30
Update Samsung.xml

### LINKED ISSUE #10514: Samsung.com (partial) breaks font
http://www.samsung.com/us/

`Samsung.com (partial)` prevents font from loading, so it defaults to plain, ugly font instead of custom font.

### LINKED ISSUE #14942: Fix ruleset for samsung.com
Type: ruleset issue

Domain: samsung.com

Is HTTPS enabled but not by default.

cschanaj:
Fix #10514, Fix #14942, Re-open #15031 

P.S. I cannot reproduce the font breakage in #10514. Simplified the complicated regex used.

### DIFF
--- a/src/chrome/content/rules/Samsung.xml
+++ b/src/chrome/content/rules/Samsung.xml
@@ -20,7 +20,6 @@
 			- help.content *
 			- m		(503, akamai)
 			- origin *
-			- us *
 
 		- (www.)samsungmobilepress.com *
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Samsung.xml
```diff
@@ -29,6 +20,7 @@ Non-2xx HTTP code: http://shop.us.samsung.com/ (200) => https://shop.us.samsung.
 			- help.content *
 			- m		(503, akamai)
 			- origin *
+			- us *
```
J0WI: WFM

terrorist96:
> P.S. I cannot reproduce the font breakage in #10514.

Me neither, but only because the rule is off by default right now. Enabling it, I see this:

Rule off:
image

Rule on:
image

cschanaj:
@terrorist96 noted with thanks!

PR #15034, opened 2018-03-31
Update and rename Zopim.xml to Zopim.com.xml

cschanaj:


src/chrome/content/rules/Zopim.xml
```diff
@@ -1,44 +0,0 @@
-<!--
-	zopim.zendesk.com
-
-
-	Fully covered subdomains:
-
-		- (www.)
-		- api
-		- blog
-		- cdn
```
J0WI: That one is missing in the new version

src/chrome/content/rules/Zopim.xml
```diff
@@ -1,44 +0,0 @@
...
-		- classic
```
J0WI: That one is missing in the new version

src/chrome/content/rules/Zopim.xml
```diff
@@ -1,44 +0,0 @@
...
-		- dashboard
-
-		- de\d\d:
```
J0WI: That one is missing in the new version

src/chrome/content/rules/Zopim.xml
```diff
@@ -1,44 +0,0 @@
...
-
-			- de04
-
-		- status
-		- status
-		- status0[12]
-		- tickets
-		- translate
```
J0WI: That one is missing in the new version

PR #15035, opened 2018-03-31
Remove unnecessary rule from UK_Ministry_of_Defence.xml

cschanaj:


src/chrome/content/rules/UK_Ministry_of_Defence.xml
```diff
@@ -3,11 +3,9 @@
...
-    <target host="defenceimagery.mod.uk" />
-    <target host="www.defenceimagery.mod.uk" />
+	<target host="defenceimagery.mod.uk" />
+	<target host="www.defenceimagery.mod.uk" />
 
-    <rule from="^http://defenceimagery\.mod\.uk/"
-        to="https://www.defenceimagery.mod.uk/"/>
-    <rule from="^http://([^/:@]+)?\.defenceimagery\.mod\.uk/"
-        to="https://$1.defenceimagery.mod.uk/" />
+	<rule from="^http://(www\.)?defenceimagery\.mod\.uk/"
```
J0WI: Could you also add a hint about the mismatch?

PR #15036, opened 2018-03-31
Delete 7Headlines.com.xml

Bisaloo:
Only two valid subdomains are ^ and www and both are refused over HTTPS.

#10843


jeremyn:
Normally rulesets should be marked `default_off` for a while before we delete them, but in this case http://www.7headlines.com redirects to https://goodbye7hl.events.pixnet.net which, according to Google Translate, indicates the service is being ended.

PR #15038, opened 2018-03-31
[Sparkasse-ansbach.de] New ruleset

n-se:


src/chrome/content/rules/Sparkasse-ansbach.de.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="Sparkasse-ansbach.de">
```
J0WI: Please also add `sparkasse-ansbach.de` as target and redirect rule to `https://www.sparkasse-ansbach.de/`. The issue on `https://sparkasse-ansbach.de` should be documented for future maintenance.

SeefriedNoah:
How can I set an exception for this domain? Currently, he's still requesting it.

src/chrome/content/rules/Sparkasse-ansbach.de.xml
```diff
@@ -1,5 +1,15 @@
...
+Fetch error: http://sparkasse-ansbach.de/ => https://sparkasse-ansbach.de/: (7, 'Failed to connect to sparkasse-ansbach.de port 443: Connection refused')
+
+-->
 <ruleset name="Sparkasse-ansbach.de">
+	<target host="sparkasse-ansbach.de" />
 	<target host="www.sparkasse-ansbach.de" />
 
-	<rule from="^http:" to="https:" />
+	<rule from="^http://sparkasse-ansbach\.de/" to="https://www.sparkasse-ansbach.de/" />
+	<rule from="^http://www.sparkasse-ansbach\.de/" to="https://www.sparkasse-ansbach.de/" />
```
J0WI: You can keep the `<rule from="^http:" to="https:" />` here. Otherwise you must escape the dot in `www\.`.

src/chrome/content/rules/Sparkasse-ansbach.de.xml
```diff
@@ -1,5 +1,15 @@
+<!--
+
+Connection closed:
+
+sparkasse-ansbach.de	
+Fetch error: http://sparkasse-ansbach.de/ => https://sparkasse-ansbach.de/: (7, 'Failed to connect to sparkasse-ansbach.de port 443: Connection refused')
```
J0WI: Please just write something less wordy like:
```
Secure connection refused:
	- sparkasse-ansbach.de
```

You can find various examples in existing rules.

src/chrome/content/rules/Sparkasse-ansbach.de.xml
J0WI: `<rule from="^http:" to="https:" />` should be the very last rule. This does the same as `<rule from="^http://www\.sparkasse-ansbach\.de/" to="https://www.sparkasse-ansbach.de/" />`

PR #15043, opened 2018-03-31
[Protonvpn.com] New ruleset

Perflyst:


Bisaloo:
Thank you for your contribution but this domain is [HSTS preloaded](https://scotthelme.co.uk/hsts-preloading/). HSTS preloading offers the same level of protection as HTTPS Everywhere. For this reason, we don't include preloaded domains in our rulesets. You can check the HSTS preload status of a given domain on [this page](https://hstspreload.org/).

PR #15048, opened 2018-04-01
Update background.js, Fix #15045

### LINKED ISSUE #15045: Incompatibility with First Party Isolation
Type: code issue

Found on [reddit](https://www.reddit.com/r/firefox/comments/875j1n/https_everywhere_with_firstparty_isolation_error/). I'm reporting myself because OP has not done it yet and they might have forgotten about it.

> Unchecked lastError value: Error: First-Party Isolation is enabled, but the required 'firstPartyDomain' attribute was not set. background.js:481

[Link to L481 of `background.js`](https://github.com/EFForg/https-everywhere/blob/master/chromium/background-scripts/background.js#L481)

I could not reproduce myself but I'm still reporting because after reading the MDN documentation, it does look like we are not handling FPI as we should:

> The cookies.get(), cookies.getAll(), cookies.set() and cookies.remove() APIs all accept a firstPartyDomain option. When first-party isolation is on, you must provide this option or the API calls will fail.

Source: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/cookies

@Hainish

cschanaj:
Since I am unable to reproduce this issues, I _assumed_ this PR works. Any step/ URLs to reproduce this issue is welcomed.

* Preserve sameSite property on Chromium

* Preserve firstPartyDomain property on Firefox

Relate #15045 

/cc @Hainish

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -455,17 +455,29 @@ function sortSwitchPlanner(tab_id, rewritten) {
 function onCookieChanged(changeInfo) {
   if (!changeInfo.removed && !changeInfo.cookie.secure && isExtensionEnabled) {
     if (all_rules.shouldSecureCookie(changeInfo.cookie)) {
-      var cookie = {name:changeInfo.cookie.name,
+      let cookie = {
+        name:changeInfo.cookie.name,
         value:changeInfo.cookie.value,
         path:changeInfo.cookie.path,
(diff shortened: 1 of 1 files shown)


cschanaj:
@Hainish this PR will preserve the `sameSite` and `firstPartyDomain` property of the coookies we are modifying. Any chance this is merged before #15090?

Hainish:
Great, thank you.  In the future please use more descriptive PR names.

PR #15053, opened 2018-04-01
[B-static.net] Simplify and reactivate ruleset 

Bisaloo:


jeremyn:
https://boards.ie seems okay. I think we can move `b-static.net` and its subdomains to a new ruleset and remove it from `Boards.ie.xml`, but otherwise keep `Boards.ie.xml`. What do you think?

Bisaloo:
It's hidden in my commit message but `boards.ie` is preloaded.

jeremyn:
Oh okay, that's fine then. In that case:

- [ ] https://b-static.net returns an HTTP 200 so you can make it a `target` and remove it from the top comment.
- [ ] Change the `ruleset` `name` to `B-static.net` to match the filename.
- [ ] https://avatar.b-static.net returns an HTTP 301 so the extra `test` isn't needed. (Also from the URL is looks like this might be a particular person's avatar image, which is also not good for a `test`.)

jeremyn:
Thanks, merged.

PR #15061, opened 2018-04-02
Propose lint chromium-translations

jeis2497052:
are these updates helpful?

### DIFF
--- a/utils/chromium-translations.py
+++ b/utils/chromium-translations.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3.6
 """
-Given two directories, copy Firefox-style https-everywhere.dtd translations from
-the first directory into appropriately-named Chromium-style messages.json
+Given two directories, copy Firefox-style https-everywhere.dtd translations
+from the first directory into appropriately-named Chromium-style messages.json
 translations in the second directory.
 """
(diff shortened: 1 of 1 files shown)


jeremyn:
@jeis2497052 Thanks for your interest in the project.

For your specific changes, it looks like you're not changing the functionality of the code, which is good, and some of the style changes are fine, such as using all caps for constants and keeping line lengths under 80 characters. However some of the changes I don't really agree with, for example the unnecessary `"""convert"""` comment, and changing (file label) `f` to `f_l` (I'm not sure what `f_l` stands for). Also there are other improvements to make, such as alphabetizing the imports and putting the `for` loop at the end under an `if __name__ == '__main__'` conditional.

My feeling here is to close the PR and defer these changes until someone needs to revisit the actual functionality of this script, and to also soft-discourage lint/style-only changes (from anyone, not just you) in other files.

I don't mean to discourage you as a contributor. If you are interested in making changes other than for rulesets and aren't sure where to begin, you can let us know here.

@Hainish What do you think?

jeremyn:
(I suppose `f_l` stands for "file label".)

jeis2497052:
thanks for the feedback

PR #15063, opened 2018-04-03
Sign rulesets

Hainish:


chromium/background-scripts/update.js
```diff
@@ -31,7 +31,7 @@ let store,
   background_callback;
 
 // how often we should check for new rulesets
-const periodicity = 10;
+const periodicity = 7200;
```
cschanaj: @Hainish isn't this just 2 hours instead of 2 days in seconds?

Hainish: You're definitely right!  Thank you.

Hainish: It's those little bugs which are killers.  Glad you caught that.

cschanaj: 033f82d looks good to me, thanks! 😄

PR #15067, opened 2018-04-03
Create Free-OCR.com.xml

### LINKED ISSUE #15050: Rule for www.free-ocr.com
Type: new ruleset
Domain: www.free-ocr.com

Bisaloo:
Fix #15050

### DIFF
--- a/src/chrome/content/rules/Free-OCR.com.xml
+++ b/src/chrome/content/rules/Free-OCR.com.xml
@@ -1,0 +1,17 @@
+<!--
+	Invalid certificate:
+		free-ocr.com
+
+-->
+<ruleset name="Free-OCR.com">
+
+	<target host="free-ocr.com" />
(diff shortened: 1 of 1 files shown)


Bisaloo:
I'm not sure why tests aren't passing...

jeremyn:
I'm getting an `Unable to connect` (what we sometimes call `Connection refused`) error at https://www.free-ocr.com/ , which is what the tests report. Are you seeing something different?

jeremyn:
I get the error from my previous comment in both Firefox and curl.

Bisaloo:
Everything is working fine over both HTTP and HTTPS with Firefox or curl. Maybe this is location dependent. Should I close this then?

jeremyn:
The problem could be location-dependent, though that would be unusual.

I think we should close this. I'm not sure what else to do with it.

PR #15072, opened 2018-04-04
Update MIUI.com.xml

ivysrono:


jeremyn:
- [ ] Mark https://www.miui.com as `Mixed content` and remove the `target`.
- [ ] The root URL for https://api.bbs.miui.com is `Mixed content`, but other URLs like https://api.bbs.miui.com/url/xiaomi816app seem okay. You could add an `exclusion` for the top URL, with a comment, and add some other URLs as `test`s.
- [ ] Mark https://avatar.bbs.miui.com as `No working URL known` and remove the `target`.

ivysrono:
https://www.miui.com MCB？ Where?

jeremyn:
I'm not sure why you're not seeing MCB on https://www.miui.com and https://api.bbs.miui.com . It is very obvious to me, and the console has a lot of CSS and JS files from http://static.bbs.miui.com blocked.

Do you have another `test` URL for https://avatar.bbs.miui.com ? The existing URL looks like it's for some specific person's avatar image, though the image itself seems generic.

ivysrono:
Test them in IE11：
default
default
They are OK.

you can get lots avatar from forum like: https://www.miui.com/forum-37-1.html

jeremyn:
Both URLs are definitely broken for me in Firefox and Chrome. Maybe the problem is geographic-location-dependent?

Can you find an https://avatar.bbs.miui.com `test` URL that is not an image, maybe a CSS or JS file? I don't like including `test`s that may be tied to a particular person, such as a particular person's avatar image.

ivysrono:
I get MCB on Tor browser.

jeremyn:
I've updated the checklist through https://github.com/EFForg/https-everywhere/pull/15072/commits/919c29c8ad7e8c242e7cf7b912f827b6b0283703. For `www`, please spell out the domain completely (`www.miui.com`) in the top comment.

Also instead of `( Test by Tor browser )`, how about `(might need to test with Tor to see MCB)`?

PR #15092, opened 2018-04-07
Use upgradeToSecure when possible

### LINKED ISSUE #49: Causes issue with Access-Control-Allow-Origin:*

bcyphers:
Closes #49. 

In Firefox, cross-origin resource sharing (CORS) requests that are approved for http urls are sometimes 
blocked after being rewritten to https. Firefox has recently added a new option to [webRequest.BlockingResponse](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/BlockingResponse), upgradeToSecure, which performs a simple http --> https rewrite. This branch tries to use that flag instead of a standard rewrite whenever (1) the rewrite is trivial and (2) the browser is Firefox 59+.

Can be tested on this website: http://shop.decatorevista.ro/

With the current version of the extension in Firefox, a few calls to maxcdn.bootstrapcdn.com will be blocked, breaking styles on the page. This branch allows them to be rewritten successfully.

### DIFF
--- a/chromium/background-scripts/util.js
+++ b/chromium/background-scripts/util.js
@@ -106,3 +106,48 @@ Object.assign(exports, {
 });
 
 })(typeof exports == 'undefined' ? require.scopes.util = {} : exports);
+
+// from
+// https://www.gregoryvarghese.com/how-to-get-browser-name-and-version-via-javascript/
+function getBrowserInfo() {
+  var agent = navigator.userAgent,
+    match = agent.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i);
+
+  if (!match) {
+    return {
+      name: null,
+      version: null
+    };
+  }
+
+  var name = match[1],
+    version = match[2];
+
+  if (/trident/i.test(name)) {
+    var ieMatch = /\brv[ :]+(\d+)/g.exec(agent) || []; 
+    return {
+      name: 'IE ', 
+      version: (ieMatch[1] || '')
+    };
+  }   
+
+  if (name === 'Chrome') {
+    var opMatch = agent.match(/\bOPR\/(\d+)/);
+    if (tem != null) {
+      return {
+        name: 'Opera', 
+        version: match[1]
+      };
+    }
+  }   
+
+  var vMatch = agent.match(/version\/(\d+)/i);
+  if (vMatch != null) {
+    version = vMatch[1];
+  }
+
+  return {
+    name: name,
+    version: version
+  };
+}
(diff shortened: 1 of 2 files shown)


bcyphers:
Still need to work out the kinks in Chrome -- in the meantime, @Hainish do you have any other feedback?

chromium/background-scripts/util.js
```diff
@@ -106,3 +106,48 @@ Object.assign(exports, {
 });
 
 })(typeof exports == 'undefined' ? require.scopes.util = {} : exports);
+
+// from
+// https://www.gregoryvarghese.com/how-to-get-browser-name-and-version-via-javascript/
+function getBrowserInfo() {
```
cschanaj: why don't you define `getBrowserInfo` inside `exports`?

epicminecrafting:
May I re-write this PR with better code and either create a new one or create a PR on your branch?

epicminecrafting:
**Warning:** The browser detection code doesn't have any explicit license and is most likely "all rights reserved" and can't be used in open source projects.

epicminecrafting:
See #15127.

bcyphers:
Oh thanks @epicminecrafting! Still new to the project, and that wouldn't have occurred to me. I'll take a closer look at your PR and try to merge it in.

Hainish:
Hey @bcyphers - also, please make pull requests from your own HTTPS Everywhere repository.  This is to keep the number of branches listed in the public repo relatively manageable.

Thanks!

cschanaj:
I guess the browser detection code can be replaced with compile time boolean variable since all we need to know is that whether the browser support the upgradeToSecure property. Chromium and Firefox extensions uses different fIle extensions anyway.

epicminecrafting:
@cschanaj We would still need to detect the Firefox version.

Hainish:
@cschanaj 
 > @epicminecrafting in the case it is known that the browser is Firefox, we could use browser.runtime.getBrowserInfo() instead , I presume? See https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/getBrowserInfo

@bcyphers and I chatted about this, and this is the solution we came up with as well.

bcyphers:
I've amended the pull request enough that I'm going to close this one and re-open it from a branch in my own repo.

PR #15087, opened 2018-04-06
Update and combine Steam related rulesets

xPaw:
Fixes #15086.

Valve just recently stopped redirecting https to http, as they're preparing for full https roll out. Steam community already forces https.

PR #15113, opened 2018-04-08
Allow only from=^http: in relaxng

Bisaloo:
We still have some rulesets that rewrite HTTPS to HTTPS (#10843) but I'd like to merge this early and whitelist the few rulesets that don't pass the tests to prevent new HTTPS->HTTPS rewrites such as in #15087.


Bisaloo:
Looks like the whitelist is not used for relaxng validation. This will probably have to wait.

Issue #13762, opened 2017-11-26
Discussion: securecookie tags


jeremyn:
Type: other

I want to revisit how we use `securecookie` tags. I generally dislike these tags. They are hard to reason about, and the most common use with a general `.+`/`.+` can break website functionality. See my comments at https://github.com/EFForg/https-everywhere/pull/7527#issuecomment-262405979 and https://github.com/EFForg/https-everywhere/pull/11475#issuecomment-340901080. 

I think `securecookie`s should rarely be used. From my comment at https://github.com/EFForg/https-everywhere/issues/2907#issuecomment-245113096:

> 1. `securecookie` tags should only be used in targeted situations where we are trying to fix a identified vulnerability, instead of used as a default catch-all, but
> 2. we don't need to make special efforts to remove existing `securecookie`s, and
> 3. casual contributors should not be encouraged to create them.

The underlying theme here is that I prefer to be cautious in rulesets, because it is generally difficult for a regular user to effectively report problems and get them fixed.

However, I've noticed an increase in the number of PRs that include `securecookie`s. I sometimes avoid these PRs because I don't want to have the `securecookie` argument again. However if the community generally feels that we should be more free with `securecookie`s then I will stop worrying about it as much.

Ping @brainwane @gloomy-ghost @Hainish @J0WI @wonderchook .

bardiharborow:
I agree with your interpretation in https://github.com/EFForg/https-everywhere/pull/7527#issuecomment-262405979, but per https://github.com/EFForg/https-everywhere/pull/13177#discussion_r144802578 I'm still very unclear how `^\w` rules are treated. If you are able to clarify this, it would help inform my ruleset work. I generally only create `securecookie` rules if all the subdomains seem to support HTTPS but maybe I need to be even more cautious.

Bisaloo:
> However, I've noticed an increase in the number of PRs that include securecookies.

My guess is that this is due to the arrival of new contributors. When I first started out, I read that most cookies were missing the `secure` flag and that we should add it whenever possible.

I don't recall any warning about the issues associated with it. It's only later, when I saw bug reports about it, that I understood it was a dangerous tool.

My take on this is that we should add a warning in `CONTRIBUTING.md` using your comments.

ihave100openPRs:
One question: With HSTS preloaded domains, do cookies get only accepted from secure domains? If that's the case then emulating their behavior shouldn't be problematic it appears.

jeremyn:
[...]

PR #15116, opened 2018-04-08
Add bash.im ruleset

VoidPhantom:
Sublist3r also listed the `s.bash.im` domain name, but it does not seem to resolve to an IP address.

### DIFF
--- a/src/chrome/content/rules/Bash.im.xml
+++ b/src/chrome/content/rules/Bash.im.xml
@@ -1,0 +1,10 @@
+<!--
+  Does not connect over HTTPS:
+    lol.bash.im
+-->
+<ruleset name="Bash.im">
+	<target host="bash.im" />
+	<target host="www.bash.im" />
+
+	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


epicminecrafting:
You need to add this line:
```xml
```

VoidPhantom:
Are you saying that I should add this line simply because it is a good idea, or because it has something to do with the reason why the test is failing?

I didn’t add a `securecookie` line after reading the discussion in #13762, which seems to suggest that:
> `securecookie` tags should only be used in targeted situations where we are trying to fix a identified vulnerability, instead of used as a default catch-all

Should I instead add a `securecookie`, just in case?

If you are saying that this is the reason why the test failed, why is that? If not, what is causing this test to fail? (I can’t seem to figure this out…)

epicminecrafting:
@VoidPhantom The current project policy is to use `securecookie` unless it causes issues. `securecookie` is not required to pass tests.

VoidPhantom:
OK, I added the `securecookie` line.

Issue #14152, opened 2017-12-31
Fastly.net (partial) breaks functionality on spanishdict.com


mileskrell:
Type: ruleset issue

Domain: spanishdict.com

Steps to reproduce:

1. Visit http://www.spanishdict.com/translate/video with HTTPS Everywhere enabled
2. Click the button labelled "Spanish to English"
3. Observe that the button is nonfunctional
3. Disable the "Fastly.net (partial)" rule
4. Click the button again
5. Observe that the button now functions as expected

patrickdark:
I'm having similar issues in Firefox Developer Edition 60.0b1 with HTTPS Everywhere 2018.02.26:
* The "Menu" button at the top-left does nothing when clicked.
* The "á" button in the "Translate" button text field does nothing when clicked.
* The "Spanish to English" tab does nothing when clicked.

The issues seem to be mitigated by disabling the "Fastly.net (partial)" rule.

patrickdark:
For the record, I got in touch with a SpanishDict representative shortly after I posted here and they informed me that the site uses HTTPS for login pages, but deliberately does not do so for the rest of their website because they rely on advertising for revenue and at least one of their advertisers is incompatible with HTTPS.

That advertiser doesn't seem to be fastly.net though (as one might guess) since https://weather.com/, which also invokes that rule, seems to work fine with the "Fastly.net (partial)" rule enabled.

Bisaloo:
@nanek, even if you are not ready to switch the whole website to HTTPS, could you use HTTPS for resources from `fastly.net`. It will likely solve this issue.

nanek:
@bisaloo Agreed, this would fix the issue. Unfortunately last we tested HTTPS for assets it increased our site load time. HTTPS is something we plan to revisit later this year with Fastly's http/2 endpoints, however in the meantime I'm making a PR for consideration to exclude our fastly domain from the rule to avoid breaking our site in the meantime.

gloomy-ghost:
massive close per #15157

PR #15117, opened 2018-04-09
Exclude n1.global.ssl.fastly.net

nanek:
Fixes https://github.com/EFForg/https-everywhere/issues/14152


gloomy-ghost:
massive close per #15157

PR #15119, opened 2018-04-09
[Fundinfo.com] Merge and simplify rulesets

Bisaloo:


src/chrome/content/rules/Fundinfo.com.xml
```diff
@@ -1,59 +1,37 @@
...
+	Invalid certificate:
+		blob.fundinfo.com
 
+	No working URL known:
+		meet.fundinfo.com
+		ucweb.fundinfo.com
 
-	Nonfunctional subdomains:
+	Login required:
+		microsites.fundinfo.com
```
gloomy-ghost: This looks fine to me, otherwise we'll need to list https://digital.fundinfo.com as well.

Bisaloo: digital.fundinfo.com redirects to HTTPS. It is not the case for microsites.fundinfo.com. It may mean that HTTPS is not supported behind to login page (mixed content or redirect to HTTP).

PR #15123, opened 2018-04-10
Delete apply55gx.com.xml

bn4t:
This Domain is now preloaded.


J0WI:
This will be handled by [HSTS Prune](https://github.com/EFForg/https-everywhere/blob/master/utils/hsts-prune/README.md). Thank you anyway!

PR #15139, opened 2018-04-11
Add ruleset: 1-800courier.com

injust:


src/chrome/content/rules/1-800courier.com.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Invalid certificate:
+		- blog.1-800courier.com
+		- webmail.1-800courier.com
```
J0WI: Wildcard DNS

injust: Noted, thanks!

PR #15141, opened 2018-04-11
Add ruleset: connectify.me

injust:


src/chrome/content/rules/Connectify.me.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Invalid certificate:
+		- cloud.connectify.me
+
+	Invalid response:
+		- downloads.connectify.me
```
J0WI: WFM

injust: Fixed, thank you

PR #15142, opened 2018-04-11
Add ruleset: reebok.ca

injust:


src/chrome/content/rules/Reebok.ca.xml
```diff
@@ -7,6 +7,9 @@
 	<target host="reebok.ca" />
 	<target host="www.reebok.ca" />
 	<target host="cp.reebok.ca" />
+		<exclusion pattern="^http://cp\.reebok\.ca" />
```
J0WI: Should be `^http://cp\.reebok\.ca/$`

injust: Oops, fixed. Thanks!

PR #15143, opened 2018-04-11
Add ruleset: adidas.ca

injust:


J0WI:
You may try an exclusion and a custom test url (same error on #15142).

src/chrome/content/rules/Adidas.ca.xml
```diff
@@ -8,6 +8,9 @@
 	<target host="www.adidas.ca" />
 	<target host="cfg.adidas.ca" />
 	<target host="cp.adidas.ca" />
+		<exclusion pattern="^http://cp\.adidas\.ca" />
```
J0WI: Should be `^http://cp\.adidas\.ca/$`

injust: Oops, fixed. Thanks!

injust:
@J0WI Looks like any URL on `cp.adidas.ca` times out during the CI build. I'll go ahead and add a whitelist entry.

PR #15150, opened 2018-04-12
Relaxng whitelist

### LINKED ISSUE #15148: Add relaxng whitelist
Type: feature request

See #15113 for a use case.

YegorIevlev:
Closes #15148, closes #15113.


cschanaj:
see https://github.com/EFForg/https-everywhere/issues/15148#issuecomment-380676814

test/validations/relaxng/run.py
```diff
@@ -20,14 +21,26 @@
...
+
 exit_code = 0
 
 print("Validation of rulesets against relaxng schema.xml begins...")
 
 for filename in sorted(files):
+    basename = os.path.basename(filename)
+    skip_relaxng = False
+    for entry in ruleset_whitelist:
+        if entry[2] == '1' and entry[3] == basename:
```
cschanaj: We can have a small performance gain if you do `if entry[3] == basename and entry[2] == '1':` insteads

test/validations/relaxng/run.py
```diff
@@ -20,14 +21,26 @@
...
+            skip_relaxng = True
+            break
     tree = etree.parse(filename)
```
cschanaj: You can also check `skip_relaxng` and `continue` here to avoid unnecessary parsing `filename` into a XML tree. This will futher improve the performance.

YegorIevlev:
@cschanaj Done.

cschanaj:
Could you also avoid the changes to the `translations` submodule?

YegorIevlev:
@cschanaj I removed this change, which was accidentally introduced while rebasing.

cschanaj:
@Hainish Would you please have a look on this PR? I think this is ready to be merged.

Hainish:
We should consider using an extra column to [`ruleset-whitelist.csv`](https://github.com/efforg/https-everywhere/blob/master/utils/ruleset-whitelist.csv).  This would whitelist many rulesets that pass the `relaxng` test.

Hainish:
I'm open to being persuaded either way.

cschanaj:
Apart from https://github.com/EFForg/https-everywhere/issues/15148#issuecomment-380676814. I would like to consider this PR as an transitional step to phrase out the HTTPS->HTTPS rewrites instead of improving the whitelisting features. This is because all rulesets in the code base pass `relaxng` without any modification, it is only after #15113 we will have the need to whitelist some rulesets from `relaxng`. IMHO, it happens only when there is a policy change in the _ruleset standards_ and in general we should not expect contributors to whitelist for the `relaxng` test.

YegorIevlev:
I think the changes brought by this PR should be reverted after HTTPS->HTTPS rewrites are removed.

Hainish:
In my opinion, `relaxng` should reflect the fundamental structural requirements for any given ruleset.  Excluding certain rulesets entirely from the `relaxng` strikes me as potentially dangerous.  We might instead consider changing the validation logic or mirroring the current `relaxng` with a new one which includes the HTTP->HTTPS requirement, then using a third binary column on the witelist to determine which validation file to use.  At some point in the future, we can remove the old `relaxng` file.

Hainish:
Instead of a third column, which will require that we change the logic in other places in the code which touch that same file, a separate file may be warranted here.  Since this is only a small set of rulesets which need to be whitelisted, I anticipate the longevity of this file to be relatively small, compared to the rest of the rulesets whitelist.

YegorIevlev:
@Hainish Adding a new column only requires changing the `ruleset_whitelist_cleanup.sh` file.

cschanaj:
> Instead of a third column, which will require that we change the logic in other places in the code which touch that same file, a separate file may be warranted here.

I agree this is a better approach, the changes required for this PR will be limited to

- updating the path to the new whitelist
- updating the indices of the columns

And we shell revert this PR after the whitelist is gone (and relatively easily).

cschanaj:
@Hainish @YegorIevlev  FYI, I have submitted a proof-of-concept PR in #15173 for https://github.com/EFForg/https-everywhere/pull/15150#issuecomment-381299905.

YegorIevlev:
@cschanaj I guess it will be closed since it duplicates mine.

Hainish:
I did decide that it makes more sense to include the `https2https_whitelist` in the existing whitelist file, since there is a nice sanitation script for that file.  We probably want to keep a single, canonical whitelist file rather than many scattered about.

I've made these changes in https://github.com/EFForg/https-everywhere/pull/15194

Hainish:
Merged in https://github.com/EFForg/https-everywhere/pull/15194.  Thank you!

PR #15153, opened 2018-04-13
Delete Google.xml

YegorIevlev:
All targets are preloaded.


YegorIevlev:
@J0WI I just looked into Chromium's preload list and found out that they are not even preloaded, but only pinned. For example:
```json
{ "name": "google.ru", "policy": "google", "include_subdomains": true, "pins": "google" }
```

Issue #13682, opened 2017-11-20
Update hsts-prune to use preload policies


Bisaloo:
Type: feature request

The HSTS preload list has added a tag to indicate under which policy a given domain was added (see https://github.com/chromium/hstspreload.org/issues/111 and [this diff from the Chromium codebase](https://chromium.googlesource.com/chromium/src/+/4bf96c30121ed23f8025e0a2ee6c68831fd0786f%5E!/)).

It would be great if `hsts-prune.js` could use this new info to remove more domains. More specifically, domains with a special policy, such as `google`, `custom` or `public-suffix-requested` should be removed automatically, without the need to check for the presence of the required headers.

Related: https://github.com/EFForg/https-everywhere/issues/13235

cschanaj:
cc @Hainish

Hainish:
See https://github.com/chromium/hstspreload.org/issues/111#issuecomment-363957772

J0WI:
I don't like the "special policies" used by some browser vendors for their own domains. There is no guarantee that they're all handled well across all browsers. It's also not clear what policies are added or kept in forks of these browsers.

pipboy96:
@J0WI Same here, especially since Chrome devs have deprecated static pinning of CA keys for every website except Google-owned ones.

PR #15156, opened 2018-04-13
Remove right-side wildcard in Google.tld_Subdomains.xml

YegorIevlev:
This file was generated by [a script](https://gist.github.com/YegorIevlev/bcb58102aed71d85f6fe9420a4934a6e) and manually edited from Chromium's HSTS preload list.

See also #11203.


Hainish:
Closing temporarily to fix #15213

J0WI:
I'm not sure if it's worth to change this in the case of Google. They own nearly all `*.google.*` domains and the ruleset gets rather complicated by listing them all.

YegorIevlev:
@J0WI We're currently in the process of removing wildcards at the right to improve performance, see #11203.

J0WI:
@cschanaj @Hainish is the tradeoff worth it?

cschanaj:
I tried rebase this PR locally and run `memusage.js`, below is the result

Before this change,

```
Hey developer! Want to see more verbose logging?
Type this into the console: let util = require('./util'); util.setDefaultLogLevel(util.VERB);
Accepted levels are VERB, DBUG, INFO, NOTE and WARN, default is NOTE
Initial usage: 42.6 MB
Maximum usage: 79.7 MB
```

After this change,

```
Hey developer! Want to see more verbose logging?
Type this into the console: let util = require('./util'); util.setDefaultLogLevel(util.VERB);
Accepted levels are VERB, DBUG, INFO, NOTE and WARN, default is NOTE
Initial usage: 41.2 MB
Maximum usage: 80.5 MB
```

IMHO, having a big ruleset like `Bitly_branded_short_domains.xml` is not really ideal, since they never go through the `fetch-test`, solutions like #13682 avoided the maintenance issues is more preferable. Given that the ruleset database already have 145k+ `target`, I am okay with this merged or not. Would be great to have @Hainish 's opinions on this.

Hainish:
As for the appropriateness of replacing right-wildcards, I'll refer to my comment in https://github.com/EFForg/https-everywhere/issues/11203#issuecomment-316560800

> As it stands, we highly discourage right-wildcards, but allow them in some cases. I don't think we should remove them completely if they are useful in removing potentially thousands of targets, such as is the case with Google. Judging where they are appropriate and where they can be removed is a call of the ruleset maintainers.

I think Google is one of those cases where replacing 117 lines with 3.2k+ lines doesn't seem worth the trade-off.  I'm open to arguments here, but unless we have a concrete example of HTTPS Everywhere actively doing the wrong thing by including a right-wildcarded `google.*` domain, this change seems overly unwieldy.

PR #15161, opened 2018-04-14
Create steemh.org.xml

ivysrono:


jeremyn:
- [ ] Change the `ruleset` `name` to `Steemh.org`
- [ ] Change the filename to `Steemh.org.xml`

PR #15165, opened 2018-04-14
Add ruleset: dblp.uni-trier.de

nhynes:


J0WI:
Thanks for your submission to HTTPS Everywhere!
The current ruleset covers just a small fraction of `uni-trier.de`. It would be great if you could add as many hosts as you can find. We provide a simple [guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains) to find more subdomains.

PR #15167, opened 2018-04-15
Fix relaxng fail in yWorks.com.xml

YegorIevlev:


src/chrome/content/rules/yWorks.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="yWorks">
+	<target host="www.yworks.com" />
+	<target host="yworks.com" />
+	<target host="www.yworks.de" />
+	<target host="yworks.de" />
```
J0WI: See also:
```
docs.yworks.com
kb.yworks.com
live.yworks.com
yed.yworks.com

Comment (J0WI) [reply src/chrome/content/rules/yWorks.xml]: Please sort `www` right after `^`.

File: src/chrome/content/rules/yWorks.xml [hunk 2]
```diff
@@ -0,0 +1,11 @@
...
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://(www\.)?yworks\.de/" to="https://www.yworks.com/" />
```
J0WI: Please document the issue.

injust: The rule is necessary, both `yworks.de` and `www.yworks.de` fail due to a TLS handshake error.

J0WI:
Have you considered adding the hosts from https://github.com/EFForg/https-everywhere/pull/15167#discussion_r195903251?

YegorIevlev:
@J0WI Done.

PR #15168, opened 2018-04-15
Add ruleset: chapter1.io

injust:


src/chrome/content/rules/Chapter1.io.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="chapter1.io">
```
Bisaloo: Please use the same case for the filename and the ruleset name. In this example, the ruleset name should be `Chapter1.io`

injust: Is this standard practice?
[CONTRIBUTING.md](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md) gives [SeattleAquarium.org.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/SeattleAquarium.org.xml) as an example for ruleset naming conventions, and the ruleset name doesn't match the filename.

Bisaloo: You are right that the contributing guide is not clear enough on that point IMO. The next section states: 

> Filenames should vaguely resemble the name so that someone looking for the file based on the name can find it easily. 

Following that logic, ruleset name and filename should actually be exactly the same because it makes it easier to find the file when you only know the ruleset name.

> Is this standard practice?

I know at least @jeremyn is also following this rule in his reviews.

PR #15178, opened 2018-04-15
Update ruleset: dailymotion.com

injust:


src/chrome/content/rules/Dailymotion.xml
```diff
@@ -1,29 +1,14 @@
 <!--
 	Nonfunctional domains:
-
+		- advertising.dailymotion.com
 		- blog.dailymotion.com
-		- press.dailymotion.com		(shows steaw.com, CN: www.steaw.com)
-		- proxy-46.dailymotion.com
-		- publicite.dailymotion.com
```
J0WI: That one should be kept.

src/chrome/content/rules/Dailymotion.xml
```diff
@@ -1,29 +1,14 @@
...
 	Partially covered domains:
-
 		- (www.)dailymotion.com
-
 			- cdn/manifest/video/\w+.mnft 403s
 			- crossdomain.xml breaks videos
-
 -->
+
 <ruleset name="Dailymotion (default off)" default_off="breaks some embedded videos">
```
J0WI: The site seems to be HTTPS by default, can we enable it now?

src/chrome/content/rules/Dailymotion.xml
injust: Added, thanks!

src/chrome/content/rules/Dailymotion.xml
injust: Removing the `default_off`. Thanks!

J0WI:
This needs more test urls (or less wildcards).

injust:
@J0WI We still need the wildcards for cookie coverage. The CI build looks good now after adding some test URLs (and fixing up the regex).

Issue #13685, opened 2017-11-21
Firefox dev tools tab off option not sticking


Hlsgs:
Type: other

As per title. It reverts to checked after closing and reopening the dev tools.

Related: Any way to disable the dev tools tab in Chrome?

jeremyn:
@Hlsgs Thanks for your report, however I'm closing this as a duplicate of issue https://github.com/EFForg/https-everywhere/issues/13611.

I'm not sure if there is a way to disable any dev tool tab in Chrome.

Issue #13611, opened 2017-11-16
Developer-tools "switch planner pane" visibility preference not stored


wavexx:
Type: code issue

The "HTTPS Everywhere" pane which is added to the developer tools is enabled by default.

By going to the developers tools preference, you can switch it off, but the setting is not remembered/stored. As soon as the developer tool pane is closed/reopened, the pane re-appears.

This is happening with both FF56 and 57, with the latest available web-extension.

If the setting is not remembered due to a bug in FF, it should be hidden by default, as it's a nice feature to have, but I'd rather have more space for the other panes.

Hainish:
This is a bug in WebExtensions, not HTTPS Everywhere.  It is fixed in Firefox 61: https://bugzilla.mozilla.org/show_bug.cgi?id=1394750

PR #15192, opened 2018-04-18
Add "Show Devtools tab" option to hide CDT tab

gavinhungry:
Adds "Show Devtools tab" to the extension options, defaulting to `true`.  Firefox itself lets users disable/hide devtools tabs out of the box, but Chromium does not.

Tested in Chromium 66 and Firefox 60b13 (Linux).

Rendered in Chromium 66:
phobos-20180418_111540_pdt

Also see #13611, #13685.

/cc @Hainish

### DIFF
--- a/chromium/pages/devtools/index.html
+++ b/chromium/pages/devtools/index.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title></title>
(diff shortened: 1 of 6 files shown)


Hainish:
Looks good - thanks!

PR #15193, opened 2018-04-18
Add custom rule for root domain: Seattle Aquarium

injust:


src/chrome/content/rules/SeattleAquarium.org.xml
```diff
@@ -1,6 +1,5 @@
 <!--
 	Invalid certificate:
-		- seattleaquarium.org
```
J0WI: The comment should be kept

J0WI:
Thanks!

PR #15197, opened 2018-04-19
Add ruleset: Freenom.com

injust:


src/chrome/content/rules/Freenom.com.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+	Timed out:
+		- resellers.freenom.com
+	
+	Expired, weak certificate:
+		- partners.freenom.com
+-->
+
+<ruleset name="Freenom.com" platform="mixedcontent">
```
J0WI: Do you have an example for mixed content issues?

injust: I believe that it was for `whois.freenom.com`.
Going to comment that one out specifically instead.

PR #15128, opened 2018-04-10
Add ruleset: nvmexpress.org

injust:

PR #15201, opened 2018-04-19
nvmexpress.org: Keep comment for www domain

injust:
Follow-up of #15128


Bisaloo:
Thank you! I missed that.

PR #15203, opened 2018-04-20
Remove ArrayBufferToString function

YegorIevlev:


Hainish:
Nice catch, thank you

Hainish:
Looks like this caused a reversion, see https://github.com/EFForg/https-everywhere/issues/15482

YegorIevlev:
@Hainish I will look into this. Thanks for noticing.

PR #15206, opened 2018-04-21
Use SHA256 in make.sh

YegorIevlev:


J0WI:
`openssl dgst -sha256` should be used for better x-platform compatibility.
@Hainish do we need this hash somewhere?

Hainish:
@J0WI This is mainly is used for manually checking cross-platform reproducibility, as well as reproducibility upon subsequent runs in the same platform.

YegorIevlev:
@Hainish Done.

PR #15214, opened 2018-04-22
Create Nova.fr.xml

Bisaloo:


jeremyn:
- [ ] https://podcast.nova.fr/ gets a `403` for me, like https://audio.nova.fr/ does, so please move `podcast` into the top comment and remove its `target`

Bisaloo:
Strange. The landing pages mirrors `^` and `www` but it freezes as soon as I try to click on a link. So I think I should indeed remove it.

jeremyn:
It seems like if I visit https://podcast.nova.fr/ first, before visiting other `nova.fr` pages, it will load, but if I visit https://podcast.nova.fr/ after visiting another `nova.fr` page (or even after reloading `podcast`), I get a `403`. Directly visiting a non-root URL like https://podcast.nova.fr/frequences returns a `404` immediately even if it's the first URL I visit.

Also, even on the first visit to https://podcast.nova.fr/ when the page loads , the audio player (top left corner) doesn't work (the sound doesn't start), but it does work on https://nova.fr/ .

I get all of these same problems with `http` too, but I don't think we should rewrite this page. The problem is more complicated than just a `403` though. How about changing the comment to something like `podcast.nova.fr ( various problems, see https://github.com/EFForg/https-everywhere/pull/15214 )` ?

jeremyn:
Thanks, merged.

PR #15237, opened 2018-04-23
Add chromium/package-lock.json

YegorIevlev:
[`package-lock.json`](https://docs.npmjs.com/files/package-lock.json) is an important security feature of `npm` which allows to check integrity of downloaded packages to protect against repository compromise or man-in-the-middle attacks.

### DIFF
--- a/.build_exclusions
+++ b/.build_exclusions
@@ -17,3 +17,4 @@ webextension/chrome-resources/update-from-chrome-svn.sh
 webextension/.eslintrc.json
 .eslintignore
 node_modules
+package-lock.json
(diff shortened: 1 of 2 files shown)


J0WI:
On the other other hand this locks packages to a specific version without having automated (security) updates.
@Hainish what do you think?

YegorIevlev:
@J0WI Not doing that makes package versions you get completely unpredictable, also `package-lock.json` contains SHA-512 hashes of packages, ensuring that you get correct, unmodified versions of packages.

Hainish:
I'm fine with this, but we should also add it to `.build_exceptions`.  Once this is done I'll be ready to merge.

Hainish:
Thank you!

PR #15240, opened 2018-04-23
typescriptlang.org

theel0ja:


src/chrome/content/rules/typescriptlang.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="typescriptlang.org">
```
J0WI: Please also add `typescriptlang.org` which can be redirected to `www`.

J0WI: Just adding the target won't work here, since `typescriptlang.org` does not support https. It needs to be redirected to `https://www.typescriptlang.org/` explicitly.

J0WI: Please add `<rule from="^http://typescriptlang\.org/" to="https://www.typescriptlang.org/" />` before `<rule from="^http:" to="https:" />`.

J0WI: Please also add a short comment about the issue on `https://typescriptlang.org/`

J0WI:
Thanks!

PR #15243, opened 2018-04-24
Add releases.llvm.org to rules

dschwoerer:
Downloading software should always be encrypted.


J0WI:
triggering Travis again

PR #15245, opened 2018-04-24
Update Saaid.net.xml [top-100k]

Giltyhub:
DDG and Bing now redirect to the default https.


J0WI:
I can still see non-trivial MCB on the site.

PR #15246, opened 2018-04-24
Create LyinComey.com.xml

Giltyhub:


zoracon:
@Giltyhub Been noticing significant contributions from you the past year. Wanted to reach out to you if you were interested in becoming a Ruleset Maintainer? It would be along the lines of what you are [already doing](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#contributing-rulesets) and it would be great to add another person to the HTTPSE crew. If you aren't interested that's perfectly okay, but if you are then we can have an offline discussion. Or maybe we can have it here if that's more comfortable for you.

Thanks for all the contributions regardless!

PR #15249, opened 2018-04-25
masterpasswordapp.com

theel0ja:


injust:
Please add `ssl.masterpasswordapp.com` as a target.

theel0ja:
Done.

J0WI:
Could you add a comment about the wildcard DNS record?

theel0ja:
So something like this?

https://github.com/EFForg/https-everywhere/blob/620ad5f2ca607a3c45c6648b5781616aff000975/src/chrome/content/rules/Openshot.org.xml#L1-L3

J0WI:
Exactly.

J0WI:
Thanks!

PR #15251, opened 2018-04-26
PyPy: Add custom rule for doc.pypy.org

injust:


src/chrome/content/rules/PyPy.xml
```diff
@@ -12,5 +12,9 @@
 <ruleset name="PyPy">
 	<target host="pypy.org" />
 	<target host="www.pypy.org" />
+	<target host="doc.pypy.org" />
```
J0WI: Please also add a hint about the mismatch.

J0WI:
Thanks!

PR #15264, opened 2018-04-28
Re-enable trezor.io

karel-3d:
We probably had the domain disabled at the time of the check.  All the domains should be working now.


J0WI:
Thanks!

J0WI:
Btw: Adding this host to https://hstspreload.org/ would make this rule here obsolete.

PR #15265, opened 2018-04-28
iltalehti.fi

theel0ja:


injust:
This site contains active mixed content, both CSS and JS.
Please add a platform="mixedcontent" attribute to the ruleset.

PR #15266, opened 2018-04-28
ilcdn.fi

theel0ja:


injust:
This site contains active mixed content, both CSS and JS.

theel0ja:
Done, what should I do with `static.ilcdn.fi`?

src/chrome/content/rules/ilcdn.fi.xml
```diff
@@ -1,5 +1,7 @@
 <ruleset name="ilcdn.fi">
         <target host="static.ilcdn.fi" />
```
injust: Move this to a comment above the header, such as:

```
<!--
	Active mixed content:
		- static.ilcdn.fi
-->
```

theel0ja: Done. https://github.com/EFForg/https-everywhere/pull/15266/commits/700e0ffb6135b719741147cacb958e79a4201b28

src/chrome/content/rules/ilcdn.fi.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	Active mixed content:
+		- static.ilcdn.fi
+-->
+
+<ruleset name="ilcdn.fi">
+        <target host="ad.ilcdn.fi" />
+        <target host="img.ilcdn.fi" />
```
J0WI: `assets.ilcdn.fi`
`frontpage-s3.ilcdn.fi`

src/chrome/content/rules/ilcdn.fi.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	Active mixed content:
+		- static.ilcdn.fi
```
J0WI: Since only `http://static\.ilcdn\.fi/$` seems to be affected an exclusion would be the better choice here.

J0WI:
Please add a comment here and in #15265 to show the relation.

J0WI:
> Please add a comment here and in #15265 to show the relation.
An example of this can be found in the Amazon rulesets.

PR #15268, opened 2018-04-28
typora.io

theel0ja:


injust:
There already exists a ruleset for the target `typora.io`. Please edit https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Typora.xml instead.

theel0ja:
@injust done here: https://github.com/EFForg/https-everywhere/pull/15278

PR #15270, opened 2018-04-29
Add ruleset: Bgr.com

injust:


injust:
Ruleset already exists, updating that instead in #15272

PR #15283, opened 2018-05-01
[TransLink.ca] New ruleset

lucaspetter:
- The HSTS header is currently served on three domains, none of which are preloaded:
  - upassbc.translink.ca
  - translinklistens.ca 
  - translinkstore.ca
- Compasscard.ca.xml has been deleted because that domain is included in this new ruleset.


lucaspetter:
Two of the build jobs are failing and I'm not sure why. The jobs are "chrome beta" and "firefox".

lucaspetter:
@ivysrono Thanks!

PR #15290, opened 2018-05-01
Create Dorsten.de.xml

EliseeReclus:
ruleset for German town Dorsten


J0WI:
@EliseeReclus would be great if you could fix this minor code style issue

PR #15294, opened 2018-05-01
Add ruleset: Inmoment.com

injust:


src/chrome/content/rules/Inmoment.com.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+	Invalid certificate:
+		- info.inmoment.com
```
J0WI: `choose.inmoment.com`

injust: Added, thank you

PR #15297, opened 2018-05-02
[EverydayGiftCards.com.au] New ruleset

youdly:


src/chrome/content/rules/Woolworths.com.au.xml
```diff
@@ -5,6 +5,7 @@
 		- Cellarmasters.com.au.xml
 		- Countdown.co.nz.xml
 		- danmurphys.com.au.xml
+		- EverdayGiftCards.com.au.xml
```
J0WI: Should be Ever**y**day :wink:

J0WI: Please don't overwrite existing commits.

PR #15307, opened 2018-05-02
Hacklang.org

theel0ja:


src/chrome/content/rules/hacklang.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Hacklang.org">
```
J0WI: See also:
```
www.hacklang.org
docs.hacklang.org
```

Please use tabs to indent lines.

J0WI:
Yes, please see our examples on https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts

PR #15311, opened 2018-05-03
Add ruleset: wind.it

scop:


src/chrome/content/rules/Wind.it.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Wind.it">
+	<target host="wind.it" />
+	<target host="www.wind.it" />
```
J0WI: Please see also:
```
start.wind.it
telefonino.wind.it
startpre.wind.it
wep.wind.it
windcloud.wind.it
windsmartcontrol.wind.it
web.pos.wind.it
authproxy.wind.it
windecare.wind.it
areaclientiaziende.wind.it:8185
windecare.wind.it
partnerportal.wind.it:4445
weg.wind.it
mdm.wind.it

scop: More thorough investigation/listing takes more time than what I have available for this, so marked as partial instead.

J0WI: Since I already searched them for you I would be happy if this work wasn't a waste of time.

scop: Added mdm and start; for the others I'm not getting http results that I believe would warrant additions.

J0WI: Thanks

PR #15312, opened 2018-05-03
Add ruleset: tim.it

scop:


src/chrome/content/rules/Tim.it.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Nonfunctional hosts in *.tim.it:
+
+		- tim.it (m)
```
J0WI: Can be redirected to `www`

scop:
More thorough investigation/listing takes more time than what I have available for this, so marked as partial instead.

J0WI:
Some tests are failing here.

PR #15319, opened 2018-05-03
leafletjs.com

theel0ja:
https://github.com/Leaflet/Leaflet/issues/6147

### DIFF
--- a/src/chrome/content/rules/leafletjs.com.xml
+++ b/src/chrome/content/rules/leafletjs.com.xml
@@ -1,0 +1,6 @@
+<ruleset name="leafletjs.com">
+        <target host="leafletjs.com" />
+
+        <rule from="^http:"
+                to="https:" />
+</ruleset>


mourner:
@theel0ja do we need this rule if I enforce HTTPS in the GitHub Pages settings?

theel0ja:
@mourner No.

J0WI:
The rule is needed anyway: https://www.eff.org/https-everywhere/faq#why-does-https-everywhere-include-rules-for-sites-like-paypal-that-already-require-https-on-all-their-pages

src/chrome/content/rules/leafletjs.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="leafletjs.com">
+        <target host="leafletjs.com" />
```
J0WI: Please also add `www.leafletjs.com ` and a custom redirect to  `^`.

PR #15327, opened 2018-05-05
add rule for Podigee.com

numismatika:


src/chrome/content/rules/Podigee.xml
```diff
@@ -0,0 +1,22 @@
...
+
+	<target host="podigee.com" />
+	<target host="www.podigee.com" />
+
+	<target host="app.podigee.com" />
+	<target host="cdn.podigee.com" />
+	<target host="images.podigee.com" />
+	<target host="quickplay.podigee.com" />
+	<target host="staging-www.podigee.com" />
+	<target host="static.podigee.com" />
```
J0WI:
```
docs.podigee.com
help.podigee.com
hilfe.podigee.com

J0WI: @numismatika would be great if you could fix this. LGTM otherwise.

pipboy96:

PR #15335, opened 2018-05-05
Add ruleset: Mrlocksmith.com

injust:


src/chrome/content/rules/Mrlocksmith.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	520 Unknown Error:
+		- mrlocksmith.com, equivalent to www.mrlocksmith.com
```
Bisaloo: Can you check if that's still the case. I can't reproduce.

injust: Neither can I. Fixing.
Likely was a problem with their origin server behind Cloudflare.

PR #15338, opened 2018-05-06
Add ruleset: Matcc.company

injust:


src/chrome/content/rules/Matcc.company.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Server error:
+		- matcc.company, equivalent to www.matcc.company
```
J0WI: This should not be redirected to `www` to be consistent with the http variant.

injust: Fixed.

J0WI:
Thanks!

PR #15341, opened 2018-05-06
[Coles.com.au] New ruleset

youdly:

PR #15350, opened 2018-05-06
[Target.com.au] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/15341


youdly:
`target.com.au`, `www`, `shop` WFM when using `openssl` but not `curl`.

J0WI:
Akamai blocks Travis, could you please whitelist the fetch tests here?

PR #15354, opened 2018-05-06
Add ruleset: PetTravel.com

scop:


src/chrome/content/rules/PetTravel.com.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Nonfunctional hosts in *.pettravel.com:
+
+		- pettravel.com (h)
```
J0WI: Can be redirected to secured `www`.

scop: Done and rebased.

PR #15355, opened 2018-05-06
Add ruleset: AirMalta.com

scop:


J0WI:
Please have a look at https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains

scop:
More thorough investigation/listing takes more time than what I have available for this, so marked as partial instead.

PR #15376, opened 2018-05-08
[dongleauth.info] Add ruleset

bn4t:


src/chrome/content/rules/dongleauth.info.xml
```diff
@@ -1,9 +1,7 @@
-<!--
-	Invalid Certificate:
-		dongleauth.info
```
J0WI: Please keep this comment that we know the reason for the special redirect below.

PR #15379, opened 2018-05-08
Create ruleset: TheGuardian.pe.ca

injust:


Bisaloo:
- MCB does not seem to hinder navigation or appearance
- all other subdomains listed by `Sublist3r` redirect to apache default page

PR #15386, opened 2018-05-09
[tbz.ch] Add ruleset

bn4t:


src/chrome/content/rules/tbz.ch.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Invalid certificate: 
+		- webmail.tbz.ch
+    - vpn.tbz.ch
```
J0WI: Please only use tabs to indent lines and see also `intranet.tbz.ch`

bn4t: Thank you for the heads up. Fixed it now.

src/chrome/content/rules/tbz.ch.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Invalid certificate: 
+	- webmail.tbz.ch
+    - vpn.tbz.ch
```
J0WI: Please only use tabs to indent lines

Issue #15213, opened 2018-04-22
Alexa labeller bot crashed


Bisaloo:
Type: other

It looks like the labeller bot is offline. It's been a couple of weeks now. @Hainish could you try to reboot it?

Bisaloo:
@Hainish, this is really helpful to prioritize reviews. Did you get the chance to look at it?

Hainish:
It looks like there were a few large rulesets that the labeller bot was attempting to parse and check against the Alexa list.  These rulesets caused the script to increase CPU load to a critical state, resulting in the process being killed.  After closing these large rulesets and re-running the script, it appears to be behaving normally.

PR #15387, opened 2018-05-09
Run hsts-prune and fix impacted rulesets

Bisaloo:


Hainish:
Closing temporarily to fix https://github.com/EFForg/https-everywhere/issues/15213

src/chrome/content/rules/Financial_Times.xml
```diff
@@ -40,7 +40,6 @@
 -->
 <ruleset name="Financial Times (partial)">
 	<target host="ft.com" />
-	<target host="www.ft.com" />
```
Giltyhub: One question: Why only www.ft.com is removed when ft.com itself is preloaded?

src/chrome/content/rules/Python.org.xml
```diff
@@ -56,7 +56,6 @@
 	<target host="packaging.python.org" />
 	<target host="packages.python.org" />
 	<target host="pl.python.org" />
-	<target host="pypi.python.org" />
```
Giltyhub: same question for here and a couple of other places

src/chrome/content/rules/Financial_Times.xml
Bisaloo: It's missing the `includeSubDomains` directive.

See https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#hsts-preloaded-rules and https://securityheaders.com/?q=https%3A%2F%2Fft.com%2F

Hainish:
Thank you!

PR #15399, opened 2018-05-11
poliisi.fi

theel0ja:


src/chrome/content/rules/poliisi.fi.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="poliisi.fi">
+        <target host="poliisi.fi" />
+        <target host="www.poliisi.fi" />
+        <target host="blogi.poliisi.fi" />
+        <target host="asiointi.poliisi.fi" />
```
J0WI: `securemail.poliisi.fi`

pipboy96:
There is already a ruleset for `poliisi.fi`. Please make a new PR with amendments to it if you want to.

PR #15413, opened 2018-05-13
Create IRCICA.org.xml

Giltyhub:


src/chrome/content/rules/IRCICA.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="IRCICA.org">
+	<target host="ircica.org" />
+	<target host="www.ircica.org" />
+	<target host="erp.ircica.org" />
```
J0WI: See also `tr.ircica.org` (404)

Giltyhub:
@J0WI it's done

PR #15418, opened 2018-05-14
Create ruleset: ChemSpider.com

injust:


src/chrome/content/rules/ChemSpider.com.xml
```diff
@@ -0,0 +1,18 @@
...
+	Active mixed content:
+		- rdf.chemspider.com
+-->
+
+<ruleset name="ChemSpider.com">
+	<target host="chemspider.com" />
+	<target host="www.chemspider.com" />
+	<target host="chemsynthesis.chemspider.com" />
+	<target host="cssp.chemspider.com" />
+	<target host="cvsp.chemspider.com" />
```
J0WI:
```
onschallenge.chemspider.com

File: src/chrome/content/rules/ChemSpider.com.xml [hunk 2]
```diff
@@ -0,0 +1,18 @@
+<!--
+	Active mixed content:
+		- rdf.chemspider.com
```
J0WI: Do you have a specific example?

injust: Can't find any. Adding as a target.

PR #15419, opened 2018-05-14
Create ruleset: Toolbox.com

injust:


src/chrome/content/rules/Toolbox.com.xml
```diff
@@ -0,0 +1,20 @@
...
+-->
+
+<ruleset name="Toolbox.com">
+	<target host="toolbox.com" />
+	<target host="www.toolbox.com" />
+	<target host="assets.toolbox.com" />
+	<target host="finance.toolbox.com" />
+	<target host="hosteddocs.toolbox.com" />
+	<target host="hr.toolbox.com" />
+	<target host="it.toolbox.com" />
```
J0WI: How about `login.toolbox.com`?

injust: Added, thanks!

PR #15420, opened 2018-05-14
Create ruleset: ExtremeTech.com

injust:


src/chrome/content/rules/ExtremeTech.com.xml
```diff
@@ -0,0 +1,32 @@
...
+	<target host="product4.extremetech.com" />
+	<target host="static.extremetech.com" />
+
+	<securecookie host="^www\.extremetech\.com$" name=".+" />
+
+	<rule from="^http://extremetech\.com/"
+		  to="https://www.extremetech.com/" />
+
+	<rule from="^http://metrics\.extremetech\.com/"
+		  to="https://extremetech-com.122.2o7.net/" />
```
J0WI: Please add a test url for it.

injust: I can't find any requests for `metrics.extremetech.com`, so I can't add tests. Removing.

J0WI: See e.g. `http://metrics.extremetech.com/1x1.gif`

injust: Adding that, thanks.

injust: That actually 404s on `extremetech-com.122.2o7.net`, so removing the target entirely.

src/chrome/content/rules/ExtremeTech.com.xml
```diff
@@ -0,0 +1,32 @@
+<!--
+	Invalid certificate:
+		- discuss.extremetech.com
+		- metrics.extremetech.com, equivalent to extremetech-com.122.2o7.net
```
J0WI: Is this comment still true?

injust: ...not sure how I missed that. Removing now.

PR #15423, opened 2018-05-14
Create ruleset: DailyCaller.com

injust:


src/chrome/content/rules/DailyCaller.com.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Active mixed content:
```
J0WI: I can only find mixed trackers, so this can be ignored.

src/chrome/content/rules/DailyCaller.com.xml
```diff
@@ -0,0 +1,17 @@
...
+		- dailycaller.com
+		- www.dailycaller.com
+
+	Invalid certificate:
+		- jobs.dailycaller.com
+		- rambo.dailycaller.com
+-->
+
+<ruleset name="DailyCaller.com">
+	<target host="amp.dailycaller.com" />
```
J0WI: `cdn.dailycaller.com`

injust: Added, thanks!

PR #15424, opened 2018-05-14
Create ruleset: NYDailyNews.com

injust:


injust:
Ruleset for this domain already exists, will instead revise in #15429

PR #15425, opened 2018-05-14
Create ruleset: TryMagPop.ca

injust:


src/chrome/content/rules/TryMagPop.ca.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="TryMagPop.ca" platform="mixedcontent">
```
J0WI: I only see mixed fonts, but we don't consider them as "breaking".

injust: Fixed, thanks!

PR #15432, opened 2018-05-15
Update ruleset: Demandware

injust:
- `investors.demandware.com` redirects to http://investor.salesforce.com/


src/chrome/content/rules/Demandware.net.xml
```diff
@@ -7,16 +7,13 @@
 
 	Problematic hosts in *demandware.net:
 
-		- www *
-
-	* Shows blank page
-
+		- www: Invalid certificate
```
J0WI: You should add an exclusion for it.

Issue #14654, opened 2018-02-17
mathjax.org may need ruleset updated


noloader:
We are having a math formula [rendering problem on our wiki](https://github.com/jmnote/SimpleMathJax/issues/12). The math markup occurs with [Simple MathJax](https://github.com/jmnote/SimpleMathJax).

According to  [HTTPS Everywhere Atlas | mathjax.org](https://www.eff.org/https-everywhere/atlas/domains/mathjax.org.html):

> The release branch contains the following rules that are enabled by default:
> 
> ```
> <ruleset name="MathJax.org">
> 	<target host="mathjax.org"/>
> 	<target host="www.mathjax.org"/>
> 	<target host="beta.mathjax.org"/>
> 		<test url="http://beta.mathjax.org/mathjax/latest/test/localization.html"/>
> 	<target host="cdn.mathjax.org"/>
> 		<test url="http://cdn.mathjax.org/mathjax/latest/docs/html/platforms/index.html"/>
> 	<target host="docs.mathjax.org"/>
> 
> 	<securecookie host=".+" name=".+"/>
> 
> 	<rule from="^http:" to="https:"/>
> </ruleset>
> ```

MathJax's CDN allows HTTPS now.  From [Using the MathJax Content Delivery Network CDN](https://docs.mathjax.org/en/latest/start.html#using-the-mathjax-content-delivery-network-cdn):

> We retired our self-hosted CDN at cdn.mathjax.org in April, 2017. We recommend using cdnjs.com.
> 
> To jump start using cdnjs, you accomplish the first step by putting
> 
> ```
> <script type="text/javascript" async
>   src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_CHTML">
> </script>
> ```
> 
> into the <head> block of your document...

I don't quite understand the ruleset above, but it looks like it is using HTTP instead of the {newly|recently|?} supported HTTPS. It is also using the wrong domain for HTTPS.

Please consider enabling HTTPS for MathJax requests, and please consider using the new CDN.

[//]: # (Type: ruleset issue)

noloader:
@https-everywhere-bot,

Thanks, done.

noloader:
@https-everywhere-bot,

Thanks, done.

[...]

PR #14698, opened 2018-02-20
Have the bot label issues by type

strugee:
What the title says. Open questions:

* `ruleset enhancement` as an issue type seems like a bit of a mouthful - maybe it should be "ruleset feature" or something instead, I dunno
* There's a `ruleset` label but it doesn't seem to be used much. Should the bot always label ruleset issues with this label?
* There's no general `code` label so I used `bug` and `enhancement` which, conveniently, there are already types for (`code issue` and `feature request`, respectively). Does that seem right?

All tests passing locally.

As usual this contains an unrelated commit (removing an outdated TODO) that didn't seem worth bothering with a separate PR for.

Ref #14616

/cc @Hainish

Issue #14616, opened 2018-02-13
Tracking bug for issue bot improvements


strugee:
Type: code issue

This is a tracking bug collecting all the ideas that have been proposed for the issue format bot, in various issues. Feel free to comment below and I'll edit the description to add ideas to the list.

Issues with high impact and low work factor are prioritized higher.

- [x] [PR #14617] Automatically report code coverage when the test suite is run
- [ ] Strip domain information when it doesn't make sense for the type
- [x] [PR #15439] Link to the issue template when posting error messages
- [x] [PR #15439] Adjust wording (https://github.com/EFForg/https-everywhere/issues/14654#issuecomment-366416388)
- [x] [PR #14698] Have the bot autolabel ruleset issues, code issues, etc. (https://github.com/EFForg/https-everywhere/issues/13890#issuecomment-364301461)
- [ ] Didn't label #15594 correctly
- [ ] Allow community members to ask the bot to label an issue as potentially outdated (not sure this is a good idea but would help people with write access find issues to close, see https://github.com/EFForg/https-everywhere/issues/13890#issuecomment-364202481)
- [ ] Allow community members to ask the bot to add metadata to the issue description (https://github.com/EFForg/https-everywhere/issues/13890#issuecomment-364301461)
- [x] [PR #14635] Update dependencies
- [ ] Debug the webhook signature crash (https://github.com/EFForg/https-everywhere/pull/11615#issuecomment-3606688490)
- [ ] Merge the PR labelling bot (https://github.com/EFForg/https-everywhere/pull/11615#issuecomment-360898571)
- [ ] Have the bot manage incoming ruleset ideas in a separate repository (https://github.com/EFForg/https-everywhere/issues/13890#issuecomment-3617661260)
- [ ] Reference other issues/PRs with shared domains (https://github.com/EFForg/https-everywhere/issues/14616#issuecomment-365190936)
- [ ] Investigate Probot 4.0's improved logging - https://github.com/probot/probot/releases/tag/v4.0.0
- [ ] Clean up the test suite (#13466)
- [ ] Have Travis CI run the bot test suite

Bisaloo:
It would also be useful if the bot could warn the OP when there is another open issue/PR with the same domain.

It would:

- help prevent duplicates
- ensure contributors fix all issues related to a specific domain where they open a PR
- inform users that their issue may already be fixed in a pending PR
- ...

This may be tricky though because it would require the bot to keep a list of domains and to scan all issues/PR every single time.

[...]

PR #15439, opened 2018-05-16
Improve issue bot wording

strugee:
Suggested by @noloader in https://github.com/EFForg/https-everywhere/issues/14654#issuecomment-366416388. I also made it so that most interactions with the bot include a note to contact me in case of trouble, but I worry this could get repetitive in case the bot comments multiple times. See for example strugee/https-everywhere-test#49.

/cc @Hainish


strugee:
See also https://github.com/EFForg/https-everywhere/pull/14698#issuecomment-389639350 because I mixed up that PR and this one:

> One thing that wouldn't be too hard to do is make it so that the bot e.g. only commented this once in a given thread. It would be more difficult but we should also be able to do behaviors like "every other comment in a thread". I've included information so that we can do that now or later if desired.

strugee:
Ref #14616

Hainish:
Looks good, re-running the failing tests

PR #15444, opened 2018-05-16
Create SecureProxySite.com.xml

Giltyhub:


src/chrome/content/rules/SecureProxySite.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="SecureProxySite.com">
```
J0WI: `blog.secureproxysite.com` (which is currently down)

J0WI:
`blog.secureproxysite.com` looks down from here

PR #15458, opened 2018-05-18
Create ruleset: DailyHive.com

injust:
Not marking as `platform="mixedcontent"` because the only AMC script is [a tracker](http://tags.crwdcntrl.net/c/12218/cc_af.js).

### DIFF
--- a/src/chrome/content/rules/DailyHive.com.xml
+++ b/src/chrome/content/rules/DailyHive.com.xml
@@ -1,0 +1,13 @@
+<!--
+	Connection refused:
+		- advertise.dailyhive.com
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/DailyHive.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Connection refused:
+		- advertise.dailyhive.com
+		- wp.sandbox.dailyhive.com
```
J0WI: `sandbox.sfo1.dailyhive.com`

injust: Added, thank you.

PR #15459, opened 2018-05-18
Create ruleset: Lotame.com and Crwdcntrl.net

injust:


injust:
Ruleset already exists for both domains, revising instead in #15460

PR #15463, opened 2018-05-18
Update Ravenjs.com.xml

Giltyhub:


injust:
Would be best to add a comment explaining that specific rule, too.

Giltyhub:
@injust ok but it seems that it didn't fix it :( https://travis-ci.org/EFForg/https-everywhere/jobs/382079383

injust:
Oops, can you try changing the rule to `<rule from="^http://cdn\.ravenjs\.com/$" to="https://cdn.ravenjs.com/" />`

injust:
Okay, I don't know how to fix the validation error. The redirect rule seems to be working properly...

J0WI:
You can just use an exclusion for `"^http://cdn\.ravenjs\.com/$"`.

PR #15474, opened 2018-05-18
Install python3.6 instead of python3.6-dev on Ubuntu

s1l3ntcat:


s1l3ntcat:
Python3.6 has a stable version. Therefore, there is no need to install python3.6-dev

PR #15493, opened 2018-05-21
Create ruleset: SonnetTech.com

injust:


injust:
All targets fail fetch tests with chain issue: missing intermediate.

PR #15500, opened 2018-05-23
Update ruleset: Katwarn.de

injust:


injust:
I'm assuming that `ww.katwarn.de` is a misspelling of `www.katwarn.de` that was indexed somewhere.
Since this domain has wildcard coverage, would it be preferable to add a wildcard target instead?

J0WI:
The wild card won't work for something like `foo.bar.katwarn.de`.

injust:
@J0WI What's the proper regex for matching first-level wildcards?

J0WI:
There is an example [here](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#regex-conventions), but I'm not sure if its worth to use a wildcard here.

injust:
Right...not sure what I even used a wildcard anymore. Fixing.

PR #15515, opened 2018-05-24
Updates test sites for Solus-Project.com and makes it deafult to on

Jacalz:
The error that made the rule error was because of "bugs.solus-project.com" and that site does not exist anymore. This PR makes it default on and it updates the test sites to sites that exists now in 2018. I have been using the old rule for a long time and it does not cause any site breakage so it should be fine to make it default to on :+1:


J0WI:
Triggering Travis again

Jacalz:
Thanks 👍

J0WI:
Thanks for the update!

PR #15516, opened 2018-05-24
Fixed and added targets to n-tv.de ruleset.

Jochen-A-Fuerbacher:


src/chrome/content/rules/n-tv.de.xml
```diff
@@ -1,20 +1,38 @@
...
-Fetch error: http://www.n-tv.de/ => https://www.n-tv.de/: Too many redirects while fetching 'https://www.n-tv.de/'
+	This domain contains a wildcard DNS record.
 
 	Mismatch:
-		- bilder[1-4]
-		- tools
-
-	Expired:
-		- hbbtv
+		- n-tv.de
```
J0WI: Can be redirected to `www`.

PR #15520, opened 2018-05-24
Update ruleset: DigitalOcean.com

injust:


gloomy-ghost:
I can't find any URL that point to try.digitalocean.com, is this domain in use?

injust:
@gloomy-ghost Yes, see https://try.digitalocean.com/performance/ for example.

injust:
Added, thanks.

PR #15525, opened 2018-05-24
Update Tool Labs section to bring all of wmflabs to HTTPS

jonkolbert:


jonkolbert:
@Hainish any idea why this isn't working?

src/chrome/content/rules/Wikimedia.xml
```diff
@@ -64,10 +67,11 @@
 		<!-- mismatch on tools.wikimedia.de -->
 		<exclusion pattern="^http://tools\.wikimedia\.de/" />
 			<test url="http://tools.wikimedia.de/" />
+		<exclusion pattern="^http://civicrm\.wikimedia\.de/" />
```
J0WI: That's because of an incomplete cert chain. Could you document that?

jonkolbert: @J0WI Done, and the build finally passed.

J0WI:
Thanks!

PR #15536, opened 2018-05-27
Create Sci-Hub.love.xml

Giltyhub:


J0WI:
Please add this to `Sci-Hub.xml`

Giltyhub:
@J0WI i thought about that but the site isn't affiliated with Sci-Hub and it isn't managed by Sci-Hub, would that still be okay?

J0WI:
I didn't notice that, therefore I'm fine with it.

PR #15547, opened 2018-05-29
Create ruleset: EnricoZini.org

injust:


src/chrome/content/rules/EnricoZini.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="EnricoZini.org">
```
J0WI: `cal.enricozini.org`

injust: Added, thank you.

PR #15559, opened 2018-05-30
Added ruleset for bundesdruckerei.de

Jochen-A-Fuerbacher:


src/chrome/content/rules/Bundesdruckerei.de.xml
```diff
@@ -0,0 +1,37 @@
+<!--
+	Connection reset:
+		helpdesk.bundesdruckerei.de
+
+	Empty response:
+		mis.bundesdruckerei.de
+
+	Requires client certificate:
+		ems.bundesdruckerei.de
+		service.bundesdruckerei.de
```
YegorIevlev: Actually, you can redirect these too, since they don't work over HTTP.

J0WI: They won't pass fetch tests, so I'm okay with having them in comments.

PR #15579, opened 2018-05-31
Added ruleset for donttrack.us

Jochen-A-Fuerbacher:


src/chrome/content/rules/donttrack.us.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="donttrack.us">
```
J0WI: Please add a hint to the DDG ruleset.

J0WI:
DuckDuckGo.xml needs to be updated too.

PR #15585, opened 2018-05-31
Create Pravduh.com.xml

Giltyhub:


Bisaloo:
Is it expected to get a blank page when I visit http://www.pravduh.com/?

Giltyhub:
@Bisaloo yes since the website has just been bought off, it's going to be something about news with ratings by readers  and will thus have to support https https://www.cnbc.com/2018/05/25/elon-musk-may-actually-be-making-a-website-to-rate-journalists-for-credibility-and-core-truth.html (don't worry i will keep an eye on it in case of a no-https emergency ;) )

PR #15586, opened 2018-05-31
Update dlvr.it ruleset 

signe:
Update dlvr.it ruleset which included non-public development hosts, sites which don't exist, and incorrect redirects

### DIFF
--- a/src/chrome/content/rules/dlvr.it.xml
+++ b/src/chrome/content/rules/dlvr.it.xml
@@ -1,12 +1,7 @@
 <!--
 	Invalid certificate:
-		dev.dlvr.it
 		mailtrack.dlvr.it
 		support.dlvr.it
-		feeds.promotedstories.com
-
-	Too many redirects:
-		www.dlvrit.com
 
 -->
 <ruleset name="dlvr.it">
@@ -17,10 +12,8 @@
 		<test url="http://api.dlvr.it/1/help" />
 	<target host="app.dlvr.it" />
 	<target host="blog.dlvr.it" />
-	<target host="dev.dlvr.it" />
 	<target host="static.dlvr.it" />
 		<test url="http://static.dlvr.it/font-awesome/css/font-awesome.min.css" />
-	<target host="stories.dlvr.it" />
 	<target host="support.dlvr.it" />
 
 	<target host="dlvrit.com" />
@@ -29,20 +22,13 @@
 		<test url="http://api.dlvrit.com/1/help" />
 	<target host="app.dlvrit.com" />
 	<target host="blog.dlvrit.com" />
-	<target host="dev.dlvrit.com" />
 	<target host="static.dlvrit.com" />
 		<test url="http://static.dlvrit.com/font-awesome/css/font-awesome.min.css" />
 	<target host="support.dlvrit.com" />
 
-	<target host="promotedstories.com" />
-	<target host="www.promotedstories.com" />
+	<rule from="^http://support\.dlvr\.it/"
+		to="https://support.dlvrit.com/" />
 
-	<rule from="^http://(dev|support)\.dlvr\.it/"
-		to="https://$1.dlvrit.com/" />
-
-	<rule from="^http://www\.dlvrit\.com/"
-		to="https://dlvrit.com/" />
-	
 	<rule from="^http:" 
 		to="https:" />
 


src/chrome/content/rules/dlvr.it.xml
```diff
@@ -29,20 +23,13 @@
...
 	<target host="static.dlvrit.com" />
 		<test url="http://static.dlvrit.com/font-awesome/css/font-awesome.min.css" />
 	<target host="support.dlvrit.com" />
 
-	<target host="promotedstories.com" />
-	<target host="www.promotedstories.com" />
-
-	<rule from="^http://(dev|support)\.dlvr\.it/"
+	<rule from="^http://support\.dlvr\.it/"
 		to="https://$1.dlvrit.com/" />
```
injust: Please change this to redirect to `"https://support.dlvrit.com/"`.

src/chrome/content/rules/dlvr.it.xml
```diff
@@ -17,7 +12,6 @@
 		<test url="http://api.dlvr.it/1/help" />
```
YegorIevlev: Move tests after the rule they apply to.

src/chrome/content/rules/dlvr.it.xml
```diff
@@ -29,20 +23,13 @@
 		<test url="http://api.dlvrit.com/1/help" />
 	<target host="app.dlvrit.com" />
 	<target host="blog.dlvrit.com" />
-	<target host="dev.dlvrit.com" />
 	<target host="static.dlvrit.com" />
 		<test url="http://static.dlvrit.com/font-awesome/css/font-awesome.min.css" />
 	<target host="support.dlvrit.com" />
```
YegorIevlev:
```xml
```

src/chrome/content/rules/dlvr.it.xml
signe: I don't know what you're referencing as there are only four tests and they are all directly following their target.

src/chrome/content/rules/dlvr.it.xml
signe: No... do not mess with our site's cookies.  We very intentionally have portions of our site which are not secure and will not respond to https.  Modifying cookies will break things.

YegorIevlev: @signe See https://www.eff.org/https-everywhere/faq#how-do-i-add-my-own-site-to-https-everywhere.

signe: What does that have to do anything?  It's irrelevant.

Our site already runs in https, without any assistance, and has for many years.  Someone else added this ruleset over a year ago and I'm fixing it because it was broken.

As I stated, portions of our site are intentionally http.  Attempts to force them to https will break them.   We do not _want_ them to be https, so I do not want a rule added which will intentionally break the usage of any cookies.

YegorIevlev: Website owners should generally try to use HSTS preloading instead of creating HTTPS Everywhere rulesets.

Unencrypted HTTP is going away. Soon any unencrypted HTTP connection attempt will result in a security warning in all browsers. I recommend to upgrade your entire website to HTTPS and use HSTS preloading before this happens.

It is unlikely that this PR will be accepted unless you can demonstrate that upgrading the removed domains results in breakage.

If keeping a part of the website unencrypted is a hard requirement, I recommend using a different domain for unencrypted content, and preloading the encrypted domain.

Bisaloo:
Hi, thank you for your PR!

> We very intentionally have portions of our site which are not secure and will not respond to https.

It's not clear to me which portions you are referring to precisely? Is it something that should be documented in the top comment of this PR to avoid mistakes in future edits?

Also, I would be inclined to keep the `dev` subdomains. I completely understand that it's unlikely to be visited by the public. But does it break anything for you? If so, it should probably be documented. If not, it should stay in the ruleset.

Bisaloo:
@signe, there are also other subdomains that are not listed in this ruleset (such as `feeds.dlvr.it`). Are they the subdomains you don't want to secure on purpose you were referring to? If they are working with HTTPS, could you add them as well please?

src/chrome/content/rules/dlvr.it.xml
signe: As I stated, _someone else_ created this ruleset some year or more ago.  All I have done is remove rules for hosts which no longer exist or which break functionality.

I would much rather there be no ruleset at all, as our site operates exactly as intended and we automatically upgrade all connections that are supposed to be upgraded.

signe:
Inclusion of the dev subdomain absolutely breaks functionality which is why I began this PR in the first place.  Removal of the deprecated hosts/domains was simply an add-on.

The dlvr.it and dlvrit.com domains host separate content, and the dlvr.it domain hosts non-SSL, non-user-specific content. Having a wildcard redirect from one to the other is a non-starter.

Bisaloo:
@signe, as I said, can you document the non-working subdomains in this ruleset please? Otherwise, we're at risk of a future update breaking things again.

pipboy96:
@signe Please fix merging conflict.

pipboy96:
`support.dlvr.it` does not exist.

pipboy96:
@signe If you want this reopened, just ask. For now I will close this as an abandoned PR.

PR #15589, opened 2018-06-01
Create HotHardware.com.xml

Giltyhub:


src/chrome/content/rules/HotHardware.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Mismatched:
+		admin.hothardware.com, CN=hothardware.com, www.hothardware.com
+		beta.hothardware.com, CN=hothardware.com, www.hothardware.com
```
Bisaloo: I strongly disagree with the comment style. It adds no valuable info and puts a higher maintenance burden. Please revert to what you were doing before, it was just fine.

Giltyhub: @Bisaloo ok, you may want to inform our friend @YegorIevlev  as well since that was his suggestion :)

Bisaloo: Yes, I have seen that. I am not sure why they said this since it goes against this project global guidelines and what most ruleset maintainers are asking (some goes for other stuff they asked, about test urls for example, please disregard it).

PR #15596, opened 2018-06-02
Update Sublist3r command

injust:
No need to explicitly list search engines to use, Sublist3r defaults to using all of them.


youdly:
Defaults would also use PassiveDNS, DnsDumpster and ThreatCrowd that can results in invalid subdomains.
See https://github.com/EFForg/https-everywhere/issues/7277

injust:
Thanks for the catch, didn't see that issue.

PR #15602, opened 2018-06-02
Create ruleset: Radare.org

### LINKED ISSUE #15594: http://(.+\.)?radare\.org/ -> https://$1radare.org/
Type: new ruleset
Domain: radare.org

I guess the rule should be something like ```http://(.+\.)?radare\.org/``` -> ```https://$1radare.org/```

injust:
Fixes #15594

### DIFF
--- a/src/chrome/content/rules/Radare.org.xml
+++ b/src/chrome/content/rules/Radare.org.xml
@@ -10,5 +10,7 @@
 	<target host="radare.org" />
 	<target host="www.radare.org" />
 
+	<securecookie host="^(www)\.radare\.org$" name=".+" />
+
 	<rule from="^http:" to="https:" />
 </ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Radare.org.xml
```diff
@@ -0,0 +1,14 @@
...
+		- cloud.radare.org
+
+	Wrong server:
+		- cydia.radare.org
+-->
+
+<ruleset name="Radare.org">
+	<target host="radare.org" />
+	<target host="www.radare.org" />
+
```
Giltyhub: You can add `<securecookie host="^(www)\.radare\.org" name=".+" />` to exclude the problematic domains

injust: Added, thanks!

Issue #14275, opened 2018-01-12
Breaks aastock.com


KayLeung:
Type: ruleset issue
Domain: aastock.com

Some internal pages are broken, eg.:
https://www.aastocks.com/en/stocks/quote/dynamic-chart.aspx?symbol=00005  (charts doesn't load, logging showed on devtools)

I also contacted the site supports, they said the site do not support HTTPS.

numismatika:
Every object loaded there could be loaded over tls all they need to do is solve the mixed.content.
The foundation is already there.

PR #15605, opened 2018-06-03
Update aastocks.com.xml

cschanaj:
- Do not force HTTPS on the homepage as this break the "Real-time Latest Quote" functionality
- Regarding #14275, it seems to happen only on Chromium browser with CORS issues (partially resolved in #15606).

### DIFF
--- a/src/chrome/content/rules/aastocks.com.xml
+++ b/src/chrome/content/rules/aastocks.com.xml
@@ -94,8 +94,6 @@
 	<target host="fcadata.aastocks.com" />
 	<target host="fcsdata.aastocks.com" />
 	<target host="fctdata.aastocks.com" />
-	<!-- target host="fpdata.aastocks.com" / -->
-	<!-- target host="fxcharts.aastocks.com" / -->
 	<target host="chartdata1.internet.aastocks.com" />
 	
 	<target host="hkg.aastocks.com" />
@@ -139,7 +137,7 @@
 	<securecookie host="hkg8\.aastocks\.com" name=".+" />
 
 	<securecookie host="logon\.aastocks\.com" name=".+" />
-	<securecookie host="secure\.aastocks.com" name=".+" />
+	<securecookie host="secure\.aastocks\.com" name=".+" />
 
 	<rule from="^http://hkg1?\.aastocks\.com/"
 		  	to="https://hkg3.aastocks.com/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/aastocks.com.xml
```diff
@@ -80,7 +74,11 @@
 -->
 <ruleset name="AASTOCKS.com (partial)">
 	<target host="www.aastocks.com" />
+	<exclusion pattern="^http://www\.aastocks\.com/(tc|sc|en)/$" />
```
YegorIevlev: Put all exclusions after a rule they apply to.

src/chrome/content/rules/aastocks.com.xml
```diff
@@ -96,8 +94,8 @@
 	<target host="fcadata.aastocks.com" />
 	<target host="fcsdata.aastocks.com" />
 	<target host="fctdata.aastocks.com" />
-	<target host="fpdata.aastocks.com" />
-	<target host="fxcharts.aastocks.com" />
+	<!-- target host="fpdata.aastocks.com" / -->
```
YegorIevlev: Why do you comment targets out instead of simply removing them?

src/chrome/content/rules/aastocks.com.xml
```diff
@@ -143,5 +141,9 @@
 	<securecookie host="logon\.aastocks\.com" name=".+" />
 	<securecookie host="secure\.aastocks.com" name=".+" />
 
-	<rule from="^http:" to="https:" />
+	<rule from="^http://hkg1?\.aastocks\.com/"
+		  	to="https://hkg3.aastocks.com/" />
+	
+	<rule from="^http:" 
+		  	to="https:" />
```
YegorIevlev: Keep each rule as a single line.

src/chrome/content/rules/aastocks.com.xml
```diff
@@ -143,5 +141,9 @@
 	<securecookie host="logon\.aastocks\.com" name=".+" />
 	<securecookie host="secure\.aastocks.com" name=".+" />
```
YegorIevlev:
```xml
<securecookie host="(logon|secure)\.aastocks.com" name=".+" />
```

cschanaj:
@YegorIevlev AFAIK, the current practice is to put the `exclusion` after the `target` it is applying. I also prefer simpler `RegExp` for readability and maintainability rather than combining them for minimal/ dubious performance gain.

PR #15611, opened 2018-06-04
Create ruleset: SheldonBrown.com

injust:


src/chrome/content/rules/SheldonBrown.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="SheldonBrown.com">
+	<target host="sheldonbrown.com" />
+	<target host="www.sheldonbrown.com" />
```
Bisaloo: cdn.sheldonbrown.com
cdn-4.sheldonbrown.com
cdn-6.sheldonbrown.com

injust: Added, thank you!

PR #15613, opened 2018-06-04
Create ruleset: Furry.ch

Furikuda:


src/chrome/content/rules/Furry.ch.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Furry.ch">
+	<target host="furry.ch" />
+	<target host="www.furry.ch" />
+	<target host="glc.furry.ch" />
+
```
YegorIevlev: `<securecookie host=".+" name=".+" />`

Furikuda: done

PR #15624, opened 2018-06-05
wait for option page to load

gloomy-ghost:
it should migrate these failing:
https://travis-ci.org/EFForg/https-everywhere/jobs/388095434
https://travis-ci.org/EFForg/https-everywhere/jobs/388093906
https://travis-ci.org/EFForg/https-everywhere/jobs/388101456

### DIFF
--- a/test/selenium/test_options.py
+++ b/test/selenium/test_options.py
@@ -1,3 +1,4 @@
+from time import sleep
 from util import ExtensionTestCase
 
 class OptionsTest(ExtensionTestCase):
@@ -11,17 +12,19 @@ class OptionsTest(ExtensionTestCase):
     def test_show_counter(self):
         selector = '#showCounter'
         self.load_options()
+        sleep(1)
+
         el = self.query_selector(selector)
         self.assertTrue(el.is_selected())
         el.click()
 
         self.driver.refresh()
-
-        self.load_options()
+        sleep(1)
         el = self.query_selector(selector)
         self.assertFalse(el.is_selected())
         el.click()
 
         self.driver.refresh()
+        sleep(1)
         el = self.query_selector(selector)
         self.assertTrue(el.is_selected())


test/selenium/test_options.py
```diff
@@ -11,13 +12,13 @@ def test_options(self):
...
         self.load_options()
+        sleep(3)
+
         el = self.query_selector(selector)
         self.assertTrue(el.is_selected())
         el.click()
 
         self.driver.refresh()
-
-        self.load_options()
```
Hainish: I think we want to keep this, since we want to assure that the option is selected *after the options page is reloaded*, not just that it's checked on the same page after we check it.

Hainish: We'll need to add an additional sleep after this too, I suspect.

Giltyhub:
superseded by https://github.com/EFForg/https-everywhere/commit/06870de1d45d1039b949aabc9f78380b15403d4f

Hainish:
@Giltyhub it's not, this is for a different test.

test/selenium/test_options.py
gloomy-ghost: I'm not sure if it's necessary since the tests are running one by one. Otherwise do we need another load_option after the second reload?

Hainish:
@gloomy-ghost can you explain why these failures are being triggered in the first place, and how this fixes it?  I'm a bit confused what the sleep is fixing here.

gloomy-ghost:
I used a VM to do a throttle test and the second `assertTrue` also fails sometimes, but the `assertFalse` never fails. I guess it's because the option values don't have enough time to be loaded since it's async after the page is opened, and the default value for checkboxes are unchecked.

Here is how the values get loaded:
https://github.com/EFForg/https-everywhere/blob/2166ca3ca8d64af628d5b37b20cfa623d437ddf8/chromium/pages/options/ux.js#L45-L49

Hainish:
Thank you!

PR #15633, opened 2018-06-06
[RunRepeat.com] New ruleset

youdly:


src/chrome/content/rules/RunRepeat.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="RunRepeat.com">
+
+	<target host="runrepeat.com" />
+	<target host="www.runrepeat.com" />
+	<target host="cdn.runrepeat.com" />
+	<target host="p.runrepeat.com" />
+
```
YegorIevlev: `<securecookie host=".+" name=".+" />`

youdly: Added

PR #15641, opened 2018-06-06
Create GetOutline.org.xml

YegorIevlev:


src/chrome/content/rules/GetOutline.org.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	For other Google coverage, see GoogleServices.xml.
```
J0WI: The reference list in `GoogleServices.xml` needs to be updated too.

Issue #15263, opened 2018-04-28
Re-enable Symantec, Verisign rulesets


wesinator:
Type: ruleset issue
Domain: symantec.com

Disabled by https-everywhere-checker

J0WI:
This has been fixed.

PR #15642, opened 2018-06-07
[Verisign.com] Re-enable ruleset

youdly:
https://github.com/EFForg/https-everywhere/issues/15263


youdly:
To avoid dependency issue, all references are temporarily removed and will be re-added back (in this PR) once related rulesets have been merged.

PR #15646, opened 2018-06-07
Update ruleset: Adnxs.com

injust:


src/chrome/content/rules/Adnxs.com.xml
```diff
@@ -1,58 +1,34 @@
...
 		- .adnxs.com
 		- www.adnxs.com
-
-	ᶜ See https://owasp.org/index.php/SecureFlag
-
 -->
-<ruleset name="Adnxs.com">
 
+<ruleset name="Adnxs.com">
 	<target host="*.adnxs.com" />
```
J0WI: The wildcard seems to break some sites. Can we be more restrictive?

injust: Changed it to explicitly list targets instead of using a wildcard

PR #15653, opened 2018-06-08
Update ruleset: Cheezburger

injust:


src/chrome/content/rules/Cheezburger.xml
```diff
@@ -63,5 +52,4 @@ Non-2xx HTTP code: http://support.cheezburger.com/ (200) => https://cheezburger.
```
J0WI: Please replace both with a simple http-to-https rule.

injust: Fixed, thanks!

PR #15682, opened 2018-06-10
Create ruleset: CNNIC.net.cn

injust:


injust:
Is there any good way to redirect from `whois.cnnic.net.cn` to `whois.cnnic.cn`, while stripping the path and query? The CI tests don't like having `.*` in the `from` pattern.

PR #15695, opened 2018-06-12
Create ruleset: FoodRepublic.com

injust:


src/chrome/content/rules/FoodRepublic.com.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	Active mixed content:
```
J0WI: I only found mixed fonts and trackers so far.

injust: Fixed, thanks!

PR #15704, opened 2018-06-13
Update ruleset: Dell

injust:


injust:
Fetch tests seem to be timing out on URLs that work just fine...

src/chrome/content/rules/Dell.xml
```diff
@@ -87,7 +68,6 @@
 	<target host="dell.com" />
 	<target host="www.dell.com" />
 	<target host="accessories.dell.com" />
-	<target host="afcs.dell.com" />
```
J0WI: WFM

injust: A bunch of Google-indexed pages redirect to HTTP, e.g. http://afcs.dell.com/support/home/.

J0WI:
This ruleset was whitelisted before, you could just update the hash and the exclusion bits there.

J0WI:
Thanks!

PR #15709, opened 2018-06-13
Update ruleset: Taboola

injust:


src/chrome/content/rules/Taboola.xml
```diff
@@ -5,18 +5,21 @@
 	CDN buckets:
 		- orig-1000[23].taboola.cotcdn.net
 
-	Problematic subdomains:
-		- cannes2016	(invalid cert; mismatch)
-		- dev-www (timeout)
-		- events (invalid cert)
-		- pages (invalid cert)
-		- vrc	(expired)
+	Active mixed content:
```
J0WI: I only see mixed fonts, but we don't consider them as "breaking".

injust: Fixed, thanks!

PR #15714, opened 2018-06-13
Update Firefox.com.xml

Giltyhub:


src/chrome/content/rules/Firefox.com.xml
```diff
@@ -3,24 +3,24 @@
 
 	These altnames don't exist:
```
J0WI: These altnames have been removed.

src/chrome/content/rules/Firefox.com.xml
```diff
@@ -3,24 +3,24 @@
...
 		- scrypt.accounts.firefox.com
-
-	Insecure cookies are set for these hosts:
 		- marketplace.firefox.com
```
J0WI: These altname has been removed.

Giltyhub: @J0WI removed!

alexmuller:
I think this pull request has broken `detectportal.firefox.com` which has an invalid certificate. My guess is that it doesn't work with TLS because it's designed for captive portal detection.

PR #15726, opened 2018-06-14
Create ruleset: DecorAdventures.com

injust:


src/chrome/content/rules/DecorAdventures.com.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Active mixed content:
+		- www.decoradventures.com
```
J0WI: Do you have an example?

injust: Can't find any MCB, I guess I confused this ruleset with another one.
Thanks for the heads up, removing the `mixedcontent` attribute.

PR #15742, opened 2018-06-14
[ParcelPoint.com.au] New ruleset

youdly:


src/chrome/content/rules/ParcelPoint.com.au.xml
```diff
@@ -0,0 +1,47 @@
+<!--
+	Non-functional subdomains:
+		- blog	(m)
+		- wow-agent.staging	(m)
+		- store-portal	(m)
```
J0WI: `support.parcelpoint.com.au`

youdly: added

PR #15743, opened 2018-06-14
Create ruleset: ThinkingDiver.com

injust:


src/chrome/content/rules/ThinkingDiver.com.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Active mixed conntent:
+		- www.thinkingdiver.com
```
J0WI: I found only mixed fonts, so you can safely remove the platform.

PR #15746, opened 2018-06-14
Create passwordsgenerator.net.xml

jonkolbert:


src/chrome/content/rules/passwordsgenerator.net.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="passwordsgenerator.net">
+	<target host="passwordsgenerator.net" />
```
J0WI: Please also add `www.passwordsgenerator.net`

Issue #15750, opened 2018-06-15
OSS coverage in HTTPS Everywhere


J0WI:
Type: new ruleset

While I was creating https://github.com/alpinelinux/aports/pull/4516 I realized that this might be a good benchmark of HTTPS Everywhere's coverage on OSS projects. So here is the list of domains that have not been rewritten.
There are some false positives, like HSTS preloaded hosts, or osts that do not (yet) support HTTPS at all.

- [x] https://0pointer.de
- [x] https://25thandclement.com
- [ ] https://abcde.einval.com
- [x] https://ab-initio.mit.edu
- [x] https://abook.sourceforge.net
- [x] https://achurch.org
- [ ] https://acko.net
- [ ] https://acme.com
- [ ] https://acpica.org
- [x] https://advamacs.com
- [ ] https://agroman.net
- [x] https://ajaxsoundstudio.com
- [ ] https://alastairs-place.net
- [x] https://alexthepuffin.googlepages.com
- [x] https://alpine.freeiz.com
- [x] https://alpine.nethq.org
- [x] https://amavisd-milter.sourceforge.net
- [x] https://amule.sourceforge.net
- [ ] https://angband.pl
- [x] https://aoetools.sourceforge.net
- [x] https://archive.debian.org
- [x] https://archive.lbzip2.org
- [x] https://archive.monkeysphere.info
- [ ] https://archives.eyrie.org
- [x] https://archive.xfce.org
- [x] https://arpon.sourceforge.net
- [x] https://art.gnome.org
- [ ] https://arthurdejong.org
- [x] https://asbradbury.org
- [x] https://asciidoc.org
- [x] https://aspell.net
- [x] https://atheme.org
- [ ] https://atrey.karlin.mff.cuni.cz
- [ ] https://audacious-media-player.org
- [x] https://audacity.sourceforge.net
- [x] https://aufs.sourceforge.net
- [x] https://augeas.net
- [x] https://avf.sourceforge.net
- [x] https://awesome.naquadah.org
- [x] https://awstats.sourceforge.net
- [x] https://babel.pocoo.org
- [x] https://backuppc.sourceforge.net
- [x] https://bazaar.canonical.com
- [ ] https://bd808.com
- [x] https://beets.io
- [x] https://betterthangrep.com
- [x] https://binaries.openttd.org
- [x] https://binwalk.org
- [ ] https://bird.network.cz
- [x] https://bitbucket.org
- [x] https://bitmath.org
[...]

PR #15755, opened 2018-06-15
Add Lxde.org.xml

J0WI:
#15750


cschanaj:
It seems that `lists` is gone from DNS. Would you update this PR?

J0WI:
`lists.lxde.org` is IPv6 only, but redirects to `lists.lxqt.org`.

cschanaj:
thanks!

PR #15765, opened 2018-06-15
Spring cleaning #3

RReverser:
This is yet another follow-up to https://github.com/EFForg/https-everywhere/pull/4400 and https://github.com/EFForg/https-everywhere/pull/12294. This time with fewer changes, just added special case to recognise commonly used generic subdomain patterns in regexps and transform them to `*.example.org` targets as it has the same behaviour per guidelines.

As usual, aside from stylistic change, this slightly decreases memory usage as well since such rulesets become "trivial".

Before:
Initial usage: 41.3 MB
Maximum usage: 79 MB
After:
Initial usage: 41.3 MB
Maximum usage: 78.5 MB

Also, as usual with previous automated runs, this will require updating `ruleset-whitelist.csv` to avoid CI failures, but looks like the format has changed since I last checked it and I will need help with figuring out how to update it.

### DIFF
--- a/src/chrome/content/rules/33Across.xml
+++ b/src/chrome/content/rules/33Across.xml
@@ -35,7 +35,6 @@
 	<securecookie host="^(?:.*\.)?33across\.com$" name=".+" />
 
 
-	<rule from="^http://(\w+\.)?33across\.com/"
-		to="https://$133across.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/Acessoseguro.net.xml
+++ b/src/chrome/content/rules/Acessoseguro.net.xml
@@ -6,7 +6,6 @@
 	<securecookie host=".+\.acessoseguro\.net$" name=".+" />
 
 
-	<rule from="^http://([\w-]+)\.acessoseguro\.net/"
-		to="https://$1.acessoseguro.net/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/ActionKit.com.xml
+++ b/src/chrome/content/rules/ActionKit.com.xml
@@ -15,7 +15,6 @@
 	<securecookie host="^\w+\.actionkit\.com$" name=".+" />
 
 
-	<rule from="^http://(\w+)\.actionkit\.com/"
-		to="https://$1.actionkit.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/Active-Events.xml
+++ b/src/chrome/content/rules/Active-Events.xml
@@ -14,7 +14,6 @@
 	<securecookie host="^.+\.activenetwork\.com$" name=".+" />
 
 
-	<rule from="^http://([\w\-]+)\.activeevents\.com/"
-		to="https://$1.activeevents.com/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/AdSpirit.xml
+++ b/src/chrome/content/rules/AdSpirit.xml
@@ -22,7 +22,6 @@ Fetch error: http://cdn6.adspirit.de/banner/1x1.gif => https://cdn6.adspirit.de/
 		<test url="http://urban.adspirit.de/adnoclick.php?pid=" />
 
 
-	<rule from="^http://([\w-]+)\.adspirit\.de/"
-		to="https://$1.adspirit.de/" />
+	<rule from="^http:" to="https:" />
 
 </ruleset>
--- a/src/chrome/content/rules/Adverticum.xml
+++ b/src/chrome/content/rules/Adverticum.xml
@@ -1,5 +1,5 @@
 <ruleset name="Adverticum">
 	<target host="*.adverticum.net" />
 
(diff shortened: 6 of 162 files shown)


RReverser:
cc @Hainish as you reviewed previous PRs

Bisaloo:
Some changes here actually break some subdomains. You are rewriting some rules that weren't trivial on purpose. See for example [`GitLab.io.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/GitLab.io.xml) or [`Craigslist.org.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Craigslist.org.xml).

RReverser:
@bisaloo Ugh. I just realised that I confused rules for `*.` and `.*` thinking that the former will match only one word, while it's vice versa. Thanks for pointing out, I'll push a fix soon.

RReverser:
@Bisaloo Actually, I don't suppose it will be distinguish intentional ones from unintentional automatically for majority of cases e.g.:

```diff
 	<!--	Cert: www.astronomicaltours.net	-->
 	<target host="*.airshowjournal.com" />
 
 
 	<!--	Unique subdomains for each gallery.	-->
-	<rule from="^http://(\w+)\.airshowjournal\.com/"
-		to="https://$1.airshowjournal.com/" />
+	<rule from="^http:" to="https:" />
```

I suppose I can skip all of these, but it's a shame that all of them can't be optimised if we take 4th-level domains into account 😢 

Do you think these cases with 4th-level subdomains not supporting HTTPS are common or can be excluded from the diff on one-by-one basis?

Bisaloo:
> Do you think these cases with 4th-level subdomains not supporting HTTPS are common?

My guess is that it will be an issue in pretty much every case. Wildcard certs are only valid for one level  so we get mismatches.

However, I am not sure how common 4th-level subdomains are...

I don't know what the best course of action is here, sorry. We should probably wait for other inputs.

RReverser:
@Bisaloo All good points. Let's wait for more input I guess.

RReverser:
> Wildcard certs are only valid for one level so we get mismatches.

Actually, taking this into account, it seems that it's part of a reason why 4th-level subdomains are rather uncommon, and I'm thinking, maybe it would make sense to limit `*` in `*.host.com` to also mean just one-level subdomain by default (and require explicit wildcards for more levels to match certificate requirements).

Hainish:
I'm not aware of the initial reasoning for left wildcards matching > 1 domain label, or if it was just an oversight.  I wish we would have limited it to a single level subdomain initially, but I'm wary now that changing this functionality will break rulesets that have relied on the left-wildcard for 4th-level (plus) subdomains.  @J0WI, do you have any insight into this last part?

J0WI:
Yes, we are actively using them in huge rulesets related to Google, AmazonAWS, Wikimedia, Mozilla etc.
Some of them might be rewritten, but there are also cases like langugae/coutry specific 3rd-level domains and additional 4th-level sudomains. They are easier to catch with a regex in the rule to avoid huge lists of targets. I think the quality of our rulesets is not yet as good to differ them from trivial wildcards.

@RReverser for this PR you could try to determine if `imarandomhostthatshouldnotexist.$TARGETHOST` or `im.a.random.host.that.should.not.exist.$TARGETHOST` are existing HTTP hosts or not. `gitLab.io` or `craigslist.org` will resolve this address and should not be touched. `airshowjournal.com` will not resolve these and can be trivialized.

RReverser:
> They are easier to catch with a regex in the rule to avoid huge lists of targets.

I'd love to see an example of this, just wondering if complexity of regexp in such cases justifies benefits over a list of targets.

> for this PR you could try to determine if imarandomhostthatshouldnotexist.$TARGETHOST or im.a.random.host.that.should.not.exist.$TARGETHOST are existing HTTP hosts or not

That's an option, but I guess simply checking domain existence wouldn't be enough to know whether they support HTTPS on these levels and so such check would have false positives?

Hainish:
>> for this PR you could try to determine if imarandomhostthatshouldnotexist.$TARGETHOST or im.a.random.host.that.should.not.exist.$TARGETHOST are existing HTTP hosts or not
>
> That's an option, but I guess simply checking domain existence wouldn't be enough to know whether they support HTTPS on these levels and so such check would have false positives?

Perhaps you can monitor the certificate transparency logs to determine if HTTPS is supported on these levels?  https://github.com/SSLMate/certspotter may be helpful here.

J0WI:
> Perhaps you can monitor the certificate transparency logs to determine if HTTPS is supported on these levels?

https://crt.sh/ is very useful btw.


> I'd love to see an example of this, just wondering if complexity of regexp in such cases justifies benefits over a list of targets.

I's hard to find some good examples right now. Alibaba.xml, AmazonAWS.com.xml, Hewlett-Packard.xml, Salesforce.com.xml and Yandex.xml have some crazy regex/wildcard combinations. Though, they might not match our quality standards anymore. Newer rules tend to have more explicit targets. I think AWS is the most uptodate rule.

> That's an option, but I guess simply checking domain existence wouldn't be enough to know whether they support HTTPS on these levels and so such check would have false positives?

My idea was that nothing will be worse after this change for hosts without a wildcard DNS record. But for hosts that resolve any domain this could have a negative impact.

RReverser:
I think I'm going to close this for now and going to try some alternative, more precise, approaches due to potential issues outlined above.

PR #15766, opened 2018-06-15
Update FreeDesktop.xml

J0WI:
#15750

### DIFF
--- a/src/chrome/content/rules/FreeDesktop.xml
+++ b/src/chrome/content/rules/FreeDesktop.xml
@@ -47,6 +47,8 @@
 	<target host="xcb.freedesktop.org" />
 	<target host="xorg.freedesktop.org" />
 
+	<test url="http://people.freedesktop.org/~airlied/" />
+
 	<securecookie host=".+" name=".+" />
 
 	<rule from="^http://(?:download|sitewranglers)\.freedesktop\.org/"
--- a/src/chrome/content/rules/FreeDesktop.xml
+++ b/src/chrome/content/rules/FreeDesktop.xml
@@ -13,7 +13,6 @@
 	<target host="bugzilla.freedesktop.org" />
 	<target host="cgit.freedesktop.org" />
 	<target host="dbus.freedesktop.org" />
-	<target host="download.freedesktop.org" />
 	<target host="dri.freedesktop.org" />
 	<target host="fontconfig.freedesktop.org" />
 	<target host="freetype.freedesktop.org" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/FreeDesktop.xml
```diff
@@ -51,24 +47,10 @@
 	<target host="xcb.freedesktop.org" />
 	<target host="xorg.freedesktop.org" />
 
-	<!--	Complications:
-				-->
-	<target host="download.freedesktop.org" />
-	<target host="sitewranglers.freedesktop.org" />
-
-		<test url="http://people.freedesktop.org/~airlied/" />
```
J0WI: Please keep this test url because there is no content on `http://people.freedesktop.org/`.

J0WI:
@cschanaj could you review our changes?

src/chrome/content/rules/FreeDesktop.xml
```diff
@@ -1,96 +1,58 @@
...
 	<target host="piglit.freedesktop.org" />
+	<target host="planet.freedesktop.org" />
 	<target host="pm-utils.freedesktop.org" />
 	<target host="poppler.freedesktop.org" />
 	<target host="portland.freedesktop.org" />
 	<target host="secure.freedesktop.org" />
+	<target host="sitewranglers.freedesktop.org" />
 	<target host="specifications.freedesktop.org" />
 	<target host="standards.freedesktop.org" />
 	<target host="swfdec.freedesktop.org" />
```
cschanaj: I get mismatched certificate for `swfdec.freedesktop.org`

src/chrome/content/rules/FreeDesktop.xml
```diff
@@ -1,96 +1,58 @@
...
 
+	<securecookie host=".+" name=".+" />
 
-	<securecookie host="^bug" name=".+" />
-
-
-	<!--	Redirects as so:
-				-->
 	<rule from="^http://(?:download|sitewranglers)\.freedesktop\.org/"
 		to="https://fontconfig.freedesktop.org/" />
```
cschanaj: For both `download` and `sitewranglers` I get `302` to `https://www.freedesktop.org/wiki/` . The content is also different from `fontconfig`, this rule might need a rewrite.

J0WI: I'm getting a server side redirect to https. Removing this rule completely now.

Issue #15618, opened 2018-06-04
Audit at least top-1m mixedcontent rulesets


Giltyhub:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)
[//]: #
[//]: # (Delete all but one of the following "Type" lines.)
[//]: # (Leave in the line that best describes the issue you are reporting.)

Type: other

[//]: # (If you are reporting a ruleset/website problem, include the top-level)
[//]: # (domain below. For example, if you want to report an issue about)
[//]: # ("one.example.com" and "two.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: example.com)
[//]: #
[//]: # (Be sure to remove the parenthesis and comment marker.  If you are only)
[//]: # (reporting an issue about "one.example.com", then the line below should)
[//]: # (be:)
[//]: #
[//]: # (Domain: one.example.com) 
[//]: #
[//]: # (Only include one top-level domain. If you have more than one top-level)
[//]: # (domain to report, such as both "example.com" and "example.org", open a)
[//]: # (new issue for each top-level domain.)
[//]: #
[//]: # (If you are not reporting a ruleset/website problem, you can delete the)
[//]: # ("Domain" line or leave it blank.)

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

can anyone post a list of all at least top-100k mixed content rulesets? From my searches many rulesets that are mixedcontent are no longer the case, i can audit one ruleset per day to activate at least one per day and if others join in the effort i'm sure we can go though all the 1k mixedcontent rulesets. what do you think?

@cschanaj @Bisaloo @injust

J0WI:
See also #529

Giltyhub:
abandoned for the time being due to time restraint

pipboy96:
@Giltyhub Done: #17436

Issue #13290, opened 2017-10-22
Audit default_off="missing certificate chain" rulesets.


bardiharborow:
Type: other

The following `default_off="missing certificate chain"` rulesets pass the fetch test:

`-problematic` rulesets that need merging with their main rulesets:

- [ ] [Modernrepo.com-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/Modernrepo.com-problematic.xml)
- [ ] [Moodle.net-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/Moodle.net-problematic.xml)
- [ ] [North-Carolina-State-University-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/North-Carolina-State-University-problematic.xml)
- [ ] [Open_InfoSec_Foundation.org-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/Open_InfoSec_Foundation.org-problematic.xml)
- [ ] [Os-cillation-mismatches.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/Os-cillation-mismatches.xml)
- [ ] [ShareFile_support.com-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/ShareFile_support.com-problematic.xml)
- [ ] [UC_Hastings.edu-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/UC_Hastings.edu-problematic.xml)
- [ ] [VPS.net-problematic.xml](https://github.com/EFForg/https-everywhere/blob/6cb0f489addf4395f7a3d1d6ab8c996ec10b47c1/src/chrome/content/rules/VPS.net-problematic.xml)

Passes fetch test (possible bulk re-enabling):

[...]

PR #15768, opened 2018-06-15
Update SecurityWeek.com.xml

J0WI:
#13290 #15618

### DIFF
--- a/src/chrome/content/rules/SecurityWeek.com.xml
+++ b/src/chrome/content/rules/SecurityWeek.com.xml
@@ -1,8 +1,11 @@
+<!--
+	Cloudflare error:
+		- s1.securityweek.com
+-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/SecurityWeek.com.xml
```diff
@@ -1,57 +1,15 @@
...
-
-	<!--	Direct rewrites:
-				-->
-	<target host="www.securityweek.com" />
-
-	<!--	Complications:
-				-->
 	<target host="securityweek.com" />
+	<target host="www.securityweek.com" />
+	<target host="s1.securityweek.com" />
```
Bisaloo: I get a cloudflare error here.

J0WI: Confirmed

PR #15772, opened 2018-06-16
Create ruleset: Elgas.com.au

injust:


src/chrome/content/rules/Elgas.com.au.xml
```diff
@@ -0,0 +1,16 @@
...
+		- trade.elgas.com.au
+
+	Wrong server:
+		- testserver.elgas.com.au
+-->
+
+<ruleset name="Elgas.com.au">
+	<target host="elgas.com.au" />
+	<target host="www.elgas.com.au" />
+	<target host="emap.elgas.com.au" />
```
J0WI: `secure.elgas.com.au`

injust: That host fails the CI build, because accessing it on HTTP redirects to `www.secure.elgas.com.au`, which DNE.
Maybe the Travis tests should be modified to allow for sites that only work on HTTPS?

J0WI: The problem here is that HTTP servers a different response than HTTPS. Could you document the issue?

injust: I've added a comment for `secure.elgas.com.au`.
In this case, would it be better to leave the target out of the ruleset, or include it and whitelist the ruleset against fetch tests?

J0WI: I don't know what the expected behavior would be in this case. I'm fine with both variants.

PR #15712, opened 2018-06-13
Update ruleset: Eclipse.org

injust:

PR #15774, opened 2018-06-16
Create ruleset: EclipseCon.org

injust:


injust:
Adding in #15712 instead.

PR #15776, opened 2018-06-16
Update ruleset: PhotoShelter

injust:


src/chrome/content/rules/PhotoShelter.xml
```diff
@@ -1,43 +1,28 @@
...
+		- www
+		- [\w-]+ (unique subdomain per client)
 -->
-<ruleset name="PhotoShelter (partial)">
 
+<ruleset name="PhotoShelter (partial)">
 	<target host="photoshelter.com" />
 	<target host="*.photoshelter.com" />
-		<exclusion pattern="^http://blog\." />
-		<exclusion pattern="^http://(?:www\.)?photoshelter\.com/(?:about/|mkt/research/)?index(?:$|\?)" />
```
J0WI: Please add a hint why we need this exclusion.

src/chrome/content/rules/PhotoShelter.xml
```diff
@@ -1,43 +1,28 @@
...
-	Partially covered subdomains:
-
-		- (www.)	(some pages redirect to http)
-
+	Invalid certificate:
+		- cdn.c, equivalent to c
+		- css.c, equivalent to c
 
 	Fully covered subdomains:
-
```
J0WI: Please keep blog and add it as test url.

PR #15780, opened 2018-06-16
Add ruleset: Eustace.io

dnoetzel:


src/chrome/content/rules/Eustace.io.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Nonfunctional hosts in *.eustace.io:
+		- www
```
J0WI: Please also add `eustace.io` (without `www`).

src/chrome/content/rules/Eustace.io.xml
```diff
@@ -1,6 +1,7 @@
 <!--
 	Nonfunctional hosts in *.eustace.io:
-		- www
+        - eustace.io
```
J0WI: Please use tabs to indent lines.

dnoetzel: Whoops. I apologize for all of the silly mistakes. Thank you for your patience.

J0WI:
Thanks!

PR #15781, opened 2018-06-17
Fix incorrect paths specified in utils/memusage.js

cschanaj:


Hainish:
Looks good, thank you!

PR #15783, opened 2018-06-17
Update Verizon rulesets

injust:


J0WI:
The separated commits were easier to review than the whole PR, thanks!

PR #15784, opened 2018-06-17
Remove Dell.com-falsemixed.xml

injust:
No longer needed after #15703


J0WI:
This will be done by `utils/ruleset-whitelist-cleanup.sh`. Thanks anyway!

PR #15790, opened 2018-06-17
Remove libxml dependencies

gloomy-ghost:
It was used for xml linting but seems like it's no longer the case


Hainish:
Thank you!

PR #15803, opened 2018-06-18
Create ruleset: Marcusoft.net

injust:


src/chrome/content/rules/Marcusoft.net.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	Active mixed content:
+		- www.marcusoft.net
```
J0WI: I just found fonts and trackers, so we can safely ignore it.

injust: Fixed, thanks!

J0WI:
Thanks!

PR #15806, opened 2018-06-18
Create ruleset: BlackDiamondEquipment.com

injust:


injust:
Not sure why the CI build is failing the fetch test.
Travis shows that http://jp.blackdiamondequipment.com/ `200`s and https://jp.blackdiamondequipment.com/ `400`s, but they both 301 on SSL Labs and from my testing.

PR #15813, opened 2018-06-18
When 'Block all unencrypted requests' is checked, FTP is still allowed.

Hainish:
This is a fix for Chromium. Firefox still doesn't recognize the
<all_urls> special string for FTP urls.


Hainish:
@bcyphers care to look this over?

bcyphers:
Sure! I'll take a look later this week.

Bisaloo:
Related: https://github.com/EFForg/https-everywhere/issues/2311

chromium/background-scripts/background.js
```diff
@@ -636,7 +642,7 @@ function onHeadersReceived(details) {
 
 // Registers the handler for requests
 // See: https://github.com/EFForg/https-everywhere/issues/10039
-chrome.webRequest.onBeforeRequest.addListener(onBeforeRequest, {urls: ["*://*/*"]}, ["blocking"]);
+chrome.webRequest.onBeforeRequest.addListener(onBeforeRequest, {urls: ["<all_urls>"]}, ["blocking"]);
```
bcyphers: Is there a reason to use `<all_urls>` rather than enumerate `["*://*/*", "ftp://*/*"]`?

Hainish: Firefox doesn't seem to care either way, and never triggers this hook even with FTP urls.  I think you're right though, and enumerating both is a more constrained criteria than just `<all_urls>`.

PR #15822, opened 2018-06-19
Rename and Update NBCDigitalAdOps.com.xml

cschanaj:


Bisaloo:
- I think it's safe to drop the `(partial)` (although we don't have a clear rule on this so if you wish to keep it, it's okay)
- Could you also rename the file and the ruleset with the domain name please?

cschanaj:
@Bisaloo I have both renamed the ruleset and the filename. thanks!

PR #15846, opened 2018-06-21
Update ThePirateBay.xml

Giltyhub:


Giltyhub:
@J0WI thepiratebay.org was down at the time but it's up now so i reran successfully the ci

PR #15854, opened 2018-06-22
Add Schweikert.ch.xml

J0WI:
#15750

### DIFF
--- a/src/chrome/content/rules/Schweikert.ch.xml
+++ b/src/chrome/content/rules/Schweikert.ch.xml
@@ -1,10 +1,9 @@
 <!--
 	SSL error:
-		- gore
-		- ns1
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Schweikert.ch.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	SSL error:
+		- gore
+		- ns1
```
cschanaj: `ns1` seems to be gone from DNS, would you update the comment? Otherwise, this PR LGTM

J0WI: Done

cschanaj:
thanks!

PR #15863, opened 2018-06-22
Add Luadns.com.xml

J0WI:


src/chrome/content/rules/Luadns.com.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Nonfunctional hosts in *.luadns.com:
+		- status.luadns.com
+		- uptime.luadns.com
```
Bisaloo: @J0WI, I think you forgot to explain what the problem is for these hosts.

J0WI: Yes :laughing:

PR #15882, opened 2018-06-23
[STA Travel.com] New ruleset

youdly:


src/chrome/content/rules/STA-Travel.com.xml
```diff
@@ -0,0 +1,75 @@
...
+	<target host="api.partners.my.statravel.com" />
+	<target host="rail.statravel.com" />
+	<target host="tours.statravel.com" />
+		<test url="http://tours.statravel.com/content/images/icon-cal-small.png" />
+	<target host="usvpn.statravel.com" />
+
+	<securecookie host=".+" name=".+" />
+
+	<!-- http redirect -->
+	<exclusion pattern="^http://www\.statravel\.com/(blog|tours)" />
```
J0WI: Please add a trailing slash for blog

src/chrome/content/rules/STA-Travel.com.xml
```diff
@@ -0,0 +1,75 @@
...
+		<test url="http://www.statravel.com/blogs/" />
```
J0WI: Should be `blog`, not `blogs`

src/chrome/content/rules/STA-Travel.com.xml
```diff
@@ -0,0 +1,75 @@
...
+		<test url="http://www.statravel.com/tours-worldwide.htm" />
```
J0WI: Are there any other `tours*` or could you use a more restrictive regex?

youdly: `/tours.htm` is also valid, redirects to `/tours-worldwide.htm`.

PR #15883, opened 2018-06-23
[STA Travel.com.sg] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/15882

### DIFF
--- a/src/chrome/content/rules/STA-Travel.com.sg.xml
+++ b/src/chrome/content/rules/STA-Travel.com.sg.xml
@@ -4,7 +4,7 @@
 
 	Non-functional subdomains:
 		- $host		(t)
-		- www		(i)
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/STA-Travel.com.sg.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	For other coverage, see STA-Travel.com.xml
+
+
+	Non-functional subdomains:
+		- $host		(t)
+		- www		(i)
```
J0WI: The chain seems to be okay, but I get 404 and HTTP redirects.

youdly: I mislabeled, fixed.

PR #15905, opened 2018-06-24
Create ruleset: BrooksBrothers.com

injust:


src/chrome/content/rules/BrooksBrothers.com.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Chain issue, missing intermediate:
+		- vpn.brooksbrothers.com
+
+	Invalid certificate:
+		- brooksbrothers.com, equivalent to www.brooksbrothers.com
+		- membership.brooksbrothers.com
+
+	Timed out:
```
J0WI: They WFM

src/chrome/content/rules/BrooksBrothers.com.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+	Chain issue, missing intermediate:
```
J0WI: Seems to be fixed

PR #15910, opened 2018-06-24
Create ThrowAwayMail.com.xml

Giltyhub:


J0WI:
@Hainish this PR passed all fetch tests, but the hosts redirects all requests back to http.

Giltyhub:
@J0WI it worked with HTTPS at the time I tested it, but it doesn't for now, so there's no problem in the fetch tests

Giltyhub:
@J0WI all static resources seem to be loaded via www.throwawaymail

Giltyhub:
so i don't see what i can do here as i agree with you that "it makes no sense to add a trivial rule for a non-functional host" so closing and hoping that the website operator enables https once and for all

J0WI:
What I meant was a regex that just rewrites files like `http://www.throwawaymail.com/tam-logo-small.png`, but I'm okay with closing it.

PR #15916, opened 2018-06-25
Update ruleset: Showcase

injust:


J0WI:
Could you also add subdomains please?

pipboy96:
This should be split into several rulesets.

PR #15917, opened 2018-06-25
Add screenshotscdn and fix captive portal detection

J0WI:
https://github.com/EFForg/https-everywhere/pull/15714#issuecomment-399766165


alexmuller:
Thanks for fixing :)

PR #15920, opened 2018-06-25
Create ruleset: 680news.com

injust:


J0WI:
I also get MCB on the main site, but since they redirect to HTTPS themselves I'm okay with it.

PR #15922, opened 2018-06-25
Create ruleset: ICPDR.org

injust:


src/chrome/content/rules/ICPDR.org.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Invalid certificate:
+		- icpdr.org, equivalent to www.icpdr.org
+		- danubis.icpdr.org
```
J0WI: `danubis.icpdr.org` WFM

injust: Thanks for catching that, no idea what I was doing there 😛

PR #15929, opened 2018-06-26
[benj.cloud] Add ruleset

bn4t:


src/chrome/content/rules/benj.cloud.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="benj.cloud">
+	<target host="benj.cloud" />
+  <target host="office.benj.cloud" />
```
J0WI: Please make sure that you use only tabs to indent lines.

J0WI:
The test is still failing. @Hainish do you know how to interpret «`not enough values to unpack (expected 2, got 1)`»?

Hainish:
@J0WI the fetch script is trying to unpack a tuple which consists of an HTTP code as well as the HTML contents in https://github.com/EFForg/https-everywhere/blob/385e5e6d9a6e5bce13092fa6e1d99fa6a7ccf84b/test/rules/src/https_everywhere_checker/check_rules.py#L140

It looks like `office.benj.cloud` is open on port 80, but not responding to requests.  This is odd, and such an edge case is causing an exception.
```
(.venv3) user@https-everywhere ~/workspace/https-everywhere (master) $ curl -vv http://office.benj.cloud/
*   Trying 195.201.19.213...
* TCP_NODELAY set
* Connected to office.benj.cloud (195.201.19.213) port 80 (#0)
> GET / HTTP/1.1
> Host: office.benj.cloud
> User-Agent: curl/7.52.1
> Accept: */*
>
* Curl_http_done: called premature == 0
* Connection #0 to host office.benj.cloud left intact
```

J0WI:
@bn4t could you whitelist this rule from fetch test [here](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-whitelist.csv)? This seems to be a reasonable fix here.

bn4t:
@J0WI How is the hash in the file you mentioned generated. What's the input data?

J0WI:
It's just the SHA-256 of the file (in this case `benj.cloud.xml`). You can use `sha256sum` or any other tool to generate it.

bn4t:
Still fails. Did I do something wrong when I added it to the whitelist? @J0WI

J0WI:
I get the following result:
```
$ sha256sum benj.cloud.xml 
bd0efbba30261e92f11877fedb766e52efd5a9c7940b8db5db84c9096bb2ddcc  benj.cloud.xml

bn4t:
Hmm. Weird. I'm currently at work so I generated it using [this](https://emn178.github.io/online-tools/sha256_checksum.html) tool. Going to try again using your result.

J0WI:
Thanks!

PR #15962, opened 2018-06-29
Create MiKTEX.org.xml, fix #15960

### LINKED ISSUE #15960: Rule for docs.miktex.org
Type: new ruleset
Domain: docs.miktex.org

cschanaj:


J0WI:
Is the timeout always reproducibal in tests? It WFM on browsers.

cschanaj:
It seems that the timeout is not always reproducible on Travis, it also works for me on different browsers I have tried now.

J0WI:
Thanks!

PR #15974, opened 2018-06-30
Re-activate and update NASA.gov.xml

cschanaj:


J0WI:
Some of the timeouts are not reproducible anymore Would you like to double check them?
LGTM otherwise.

J0WI:
Why did you whitelist it from fetch tests? The only error was `Could not resolve host: testsite.gsfc.nasa.gov` which seems to be legit.

cschanaj:
This is because the error was caused by the server side redirect from `apd440.gsfc.nasa.gov` to `testsite.gsfc.nasa.gov`, but the domain itself serves HTTPS correctly. I _guess_ this is an internal domain, so it should be fine to remove it. will sybmit a new commit soon.

J0WI:
Thanks for clarify!

PR #15987, opened 2018-07-01
[Mit.gov.tr] New ruleset

agyild:


agyild:
The following line:
```xml
<rule from="^http://mit\.gov\.tr" to="https://www.mit.gov.tr" />
`````
Fails with the error: `Element rule failed to validate attributes`. Don't know exactly why it occurs, but I suspect it's due to redirect to www subdomain. Redirect is there because of SSL certificate only covering `*.mit.gov.tr` subdomain. Would appreciate feedback.

ivysrono:
@agyild You loase `/`
`<rule from="^http://mit\.gov\.tr/" to="https://www.mit.gov.tr/" />`

pipboy96:
Deleted repository.

PR #15995, opened 2018-07-01
[Seslisozluk.net] New ruleset

agyild:


agyild:
`test_show_counter` has failed, from my understanding I think this is an internal error, ruleset should be okay. Other than that even though I am able to secure the `__cfduid` cookie, I am unable to secure `PHPSESSID` and `seslisozluk` in this case. Any idea why?

pipboy96:
➡️#18111

PR #16000, opened 2018-07-02
 Batch trivialize unnecessary amazonaws.com rewrites

cschanaj:


src/chrome/content/rules/PolitiFact.xml
```diff
@@ -22,6 +22,6 @@
 		to="https://stpetersburgtimes.122.2o7.net/" />
```
Bisaloo: You may want to try running your script on `1[12]2.2o7.net` subdomains next.

PR #15998, opened 2018-07-02
Batch trivialize unnecessary cloudfront.net rewrites

cschanaj:
Attached the source code used to perform the batch rewrites

- https://gist.github.com/cschanaj/31abc5c1f08416683aeb54e74d3846ff

PR #16029, opened 2018-07-06
Detect static securecookies to trivialize more rules

cschanaj:
This PR modify `trivialize-rules.js` to detect _static_ `securecookie` to trivialize more rules. It also trivialize the `securecookie` tags if all the `securecookie` domain and `target` domain are strictly identical.  This will further trivializes about 100+ rulesets when ran against 8cfabe52e659fe0f0fb091f83e3b8004134d884f

cc @RReverser @Hainish @Bisaloo @J0WI

### DIFF
--- a/utils/trivialize-rules/explode-regexp.js
+++ b/utils/trivialize-rules/explode-regexp.js
@@ -4,8 +4,8 @@ const { parse } = require('regulex');
 
 class UnsupportedRegExp extends Error {}
 
-function explodeRegExp (re, callback) {
-  (function buildUrls (str, items) {
+function explodeRegExp(re, callback) {
+  (function buildUrls(str, items) {
     if (items.length === 0) {
       callback(str + '*');
       return;
@@ -79,7 +79,7 @@ function explodeRegExp (re, callback) {
 
     throw new UnsupportedRegExp(first.raw);
   })('*', parse(re).tree);
-}
+};
 
 module.exports = {
   UnsupportedRegExp,
--- a/utils/trivialize-rules/explode-regexp.js
+++ b/utils/trivialize-rules/explode-regexp.js
@@ -4,8 +4,8 @@ const { parse } = require('regulex');
 
 class UnsupportedRegExp extends Error {}
 
-function explodeRegExp(re, callback) {
-  (function buildUrls(str, items) {
+function explodeRegExp (re, callback) {
+  (function buildUrls (str, items) {
     if (items.length === 0) {
       callback(str + '*');
       return;
@@ -79,7 +79,7 @@ function explodeRegExp(re, callback) {
 
     throw new UnsupportedRegExp(first.raw);
   })('*', parse(re).tree);
-};
+}
 
 module.exports = {
   UnsupportedRegExp,
--- a/utils/trivialize-rules/trivialize-rules.js
+++ b/utils/trivialize-rules/trivialize-rules.js
@@ -258,9 +258,9 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) => 
 
   if (domains.slice().sort().join('\n') !== targets.sort().join('\n')) {
     // For each securecookie rule, check if it is a static securecookie.
-    // If it is non-static, remove the securecookie rule if it contains
-    // dangling rules. This removal is better done one by one to avoid
-    // unwanted side effects.
+    // If it is non-static, we do not trivialize the ruleset; Otherwise,
+    // we remove the securecookie if it contain only unsupported hosts.
+    // This removal is better done one by one to avoid side effects.
     // Else if ALL securecookie rules are static, trivialize the targets.
     for (const securecookie of securecookies) {
       let [isStatic, shouldRemove] = isStaticCookie(securecookie);
@@ -269,12 +269,7 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) => 
         if (shouldRemove) {
           let scReSrc = `\n([\t ]*)<securecookie\\s*host=\\s*"${escapeStringRegexp(securecookie.host)}"(\\s*)name=\\s*"${escapeStringRegexp(securecookie.name)}"\\s*?/>[\t ]*\n`;
           let scRe = new RegExp(scReSrc);
-          if (scRe && scRe.test(source)) {
-            source = source.replace(scRe, '');
-          } else {
-            fail`Failed to construct regexp which matches securecookie: ${JSON.stringify(securecookie)}`;
-            return;
-          }
+          source = source.replace(scRe, '');
         }
       } else {
         // Skip this ruleset as it contain non-static securecookies
--- a/utils/trivialize-rules/trivialize-rules.js
+++ b/utils/trivialize-rules/trivialize-rules.js
@@ -16,7 +16,7 @@ const rulesDir = `${__dirname}/../../src/chrome/content/rules`;
 
 const tagsRegExps = new Map();
 
-function createTagsRegexp (tag) {
+function createTagsRegexp(tag) {
   let re = tagsRegExps.get(tag);
   if (!re) {
     const tagRe = `<${tag}(?:\\s+\\w+=".*?")*\\s*\\/>`;
@@ -26,7 +26,7 @@ function createTagsRegexp (tag) {
   return re;
 }
 
-function replaceXML (source, tag, newXML) {
+function replaceXML(source, tag, newXML) {
   let pos, indent;
   let re = createTagsRegexp(tag);
 
@@ -75,12 +75,12 @@ const rules =
             }
           });
 
-function isTrivial (rule) {
+function isTrivial(rule) {
   return rule.from === '^http:' && rule.to === 'https:';
 }
 
 files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) => {
-  function createTag (tagName, colour, print) {
+  function createTag(tagName, colour, print) {
     return (strings, ...values) => {
       let result = `[${tagName}] ${chalk.bold(name)}: ${strings[0]}`;
       for (let i = 1; i < strings.length; i++) {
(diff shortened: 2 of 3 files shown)


RReverser:
@cschanaj Can you please explain in prose what rules you're using / when it's going to assume that cookies are ok? I believe this is quite tricky to detect statically but I might be wrong and hard to understand the nuances from code.

cschanaj:
@RReverser This PR aims to trivialize rulesets that would have been modified if securecookies are ignored. It is assumed that if all securecookies are static, they can be safely ignored.

A securecookie is called to be static either it is a trivial securecookie or **ALL** of the following conditions are satisfied: 

1. `securecookie.host` match `cookie.host` from the beginning `^` to the end `$`. Otherwise, it might match subdomains/ partial patterns, thus a non-trivial securecookie.
2. `securecookie.host` will not throw an error when passed to `explodeRegExp()`. Otherwise, it might match patterns too complicated for our interests. 
3. Each exploded `securecookie.host` should be included in `ruleset.target`/ exploded target. Otherwise, this ruleset is likely problematic itself. It is dangerous for a rewrite.

P.S. I have force-pushed 917e627 because there was an logic error in the previous implementation. Sadly we cannot trivialize the securecookies anyway.

P.S.2. When running this commit against the master, it is found that the script is problematic for `AllianceBernstein.xml`, namely a trivial domain is gone from `ruleset.target`; but I am guessing this a bug from the upstream...

Bisaloo:
@cschanaj, it would be useful to have this info somewhere else than a comment in this PR. It would be great if you could copy your text from https://github.com/EFForg/https-everywhere/pull/16029#issuecomment-403491197 in either the commit message or a standalone README.

cschanaj:
@Bisaloo I've added the comment in the commit message of 77f08cf

RReverser:
@cschanaj I see that you, for example, assume that `securecookie.host === '.+' && securecookie.name === '.+'` means that domain list can be trivialised; I think the opposite, it means that cookies might cover hosts not covered by `rule` expressions and so any current `target`s have to be preserved.

Also I wonder how you treat cookies on subdomains? In general, `subdomain1.cookie.com` can set cookies for `subdomain2.cookie.com` as well as `cookie.com` itself and so just the fact that cookie host doesn't appear in explicit targets, doesn't mean that targets can be trivialised.

RReverser:
> For cookies to be covered, there must be at least one rule covering the same domain. 

Ah, that I didn't know. It does change the perspective and optimisation opportunities. Thanks!

cschanaj:
> When running this commit against the master, it is found that the script is problematic for AllianceBernstein.xml, namely a trivial domain is gone from ruleset.target; but I am guessing this a bug from the upstream...

@RReverser could you please confirm this? It seem that not all covered domains are enumerated by `explodeRegExp()` for certain type of RegExp. You might remove the securecookie tag to run `trivialize-rules` in master to see the bug.

RReverser:
Okay, yes, I found the bug.

RReverser:
Fix is here: https://github.com/EFForg/https-everywhere/pull/16051

cschanaj:
@RReverser yes, thanks for filing a fix!!

utils/trivialize-rules/trivialize-rules.js
```diff
@@ -181,8 +182,43 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) =>
...
+    if (!securecookie.host.startsWith('^') || !securecookie.host.endsWith('$')) {
+      return false;
+    }
+
+    let localDomains = new Set();
+
+    try {
+      explodeRegExp(securecookie.host, domain => {
+        if (domain.startsWith('.')) {
+          domain = domain.slice(1);
```
Hainish: Is this sufficient munging to ensure it matches the target domains?
If not, at worst it will not match and result in a `false` result, which isn't terrible but not desirable.

cschanaj: @Hainish this should be good enough. As mentioned in https://github.com/EFForg/https-everywhere/pull/16029#issuecomment-404150837, securecookie.host only match  cookie.domain if there is a rule covers cookie.domain. Given the `target` are trivial, cannot be anything other than `domain.example.com` and `.domain.example.com` for a securecookie rule to works. With condition (1) effective, securecookie.host should explode to either one of them. Otherwise, the securecookie will never be applied (dangling rules?) under the current implementation. 

With the above logic in mind, I have also add 4bbd731to remove some of these securecookie rules that will never be effective.

cschanaj: IMHO, it would be great to have #15998 and #16000 merged before running this against the master.

cschanaj:
@Hainish would you merge this code before the next release if the changes I made after your first review look good to you?

Hainish:
@cschanaj yes - releases will be made every other month, so this should be deployed next month.  I'll have more time to review next week.  No promises if there is a security release, though.

Hainish:
Perhaps @bcyphers has some time to review?

bcyphers:
Looking at it now

Hainish:
@bcyphers?

cschanaj:
@RReverser it seems that `www.groupersocialclub.com` is missing in the following diff. would you take a look on this? 

```diff
diff --git a/src/chrome/content/rules/Grouper_Social_Club.com.xml b/src/chrome/content/rules/Grouper_Social_Club.com.xml
index 4e262e7fb0..8e9e2301a9 100644
--- a/src/chrome/content/rules/Grouper_Social_Club.com.xml
+++ b/src/chrome/content/rules/Grouper_Social_Club.com.xml
@@ -11,15 +11,13 @@ Fetch error: http://groupersocialclub.com.au/ => https://groupersocialclub.com.a
 <ruleset name="Grouper Social Club.com" default_off='failed ruleset test'>
 
        <target host="groupersocialclub.com" />
-       <target host="*.groupersocialclub.com" />
        <target host="groupersocialclub.com.au" />
-       <target host="*.groupersocialclub.com.au" />
+       <target host="www.groupersocialclub.com.au" />
 
 
        <securecookie host="^(?:www)?\.groupersocialclub\.com(?:\.au)?$" name=".+" />
 
 
-       <rule from="^http://(www\.)?groupersocialclub\.com(\.au)?/"
-               to="https://$1groupersocialclub.com$2/" />
+       <rule from="^http:" to="https:" />
 
 </ruleset>
```

cschanaj:
@RReverser my apologies, I have forgotten to rebase this PR before running the script. the issue is a duplicate of the fixed one. sorry for pinging you. please ignore my previous comment.

bcyphers:
Hi, sorry, still working on this -- had to catch up on the rest of the static simplifications first. I'll have some comments this weekend.

utils/trivialize-rules/trivialize-rules.js
```diff
@@ -181,11 +184,89 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) =>
...
+    // (possibly with more leading dots) for a securecookie rule ever to take
+    // place.
+    //
+    // With condition (1) effective, securecookie.host should explode to either
+    // one of the aforementioned patterns. Otherwise, the securecookie rules
+    // will never be applied. Such dangling securecookie rules can be removed
+    // safely.
+    if (unsupportedDomains.size > 0) {
+      if (unsupportedDomains.size === localDomains.size) {
+        shouldRemoveSecurecookies = true;
```
bcyphers: If any one of the securecookies is unsupported, this will cause all of them to be removed. I don't think that's the intended behavior. 

For example, for the following XML, the script will trivialize the file and remove all securecookies:
```
<target host="a.foo.com" />
<supercookie host="a.foo.com" />
<supercookie host="b.foo.com" />
...
```
But if the two securecookie entries are combined into one (`<securecookie host="^(a|b)\.foo\.com$" />`), the script will refuse to trivialize the XML.

It seems like we should remove one securecookie entry at a time, not drop them all at once.

utils/trivialize-rules/explode-regexp.js
```diff
@@ -4,8 +4,8 @@ const { parse } = require('regulex');
 
 class UnsupportedRegExp extends Error {}
 
-function explodeRegExp(re, callback) {
-  (function buildUrls(str, items) {
+function explodeRegExp (re, callback) {
```
RReverser: Hmm, this file was previously formatted with prettier. Did it add these whitespaces or did you manually? Looks quite odd.

cschanaj: May I know which prettier options did you used? I can revert the changes in a37df21

RReverser: I think just single-quotes plus semicolons.

cschanaj: It seems to me that prettier will create some subtle changes to `trivialize-rules.js`, e.g. 

```diff
+            let scReSrc = `\n([\t ]*)<securecookie\\s*host=\\s*"${escapeStringRegexp(
+              securecookie.host
+            )}"(\\s*)name=\\s*"${escapeStringRegexp(
+              securecookie.name
+            )}"\\s*?/>[\t ]*\n`;
``` 

So I guess it is better not to run prettier on it. Would you mind if I leave the file as-is in e8f27ee?

utils/trivialize-rules/trivialize-rules.js
```diff
@@ -181,9 +183,98 @@ files.fork().zipAll([ sources.fork(), rules ]).map(([name, source, ruleset]) =>
...
+    //
+    // With condition (1) effective, securecookie.host should explode to either
+    // one of the aforementioned patterns. Otherwise, the securecookie rules
+    // will never be applied. Such dangling securecookie rules can be removed
+    // safely.
+    if (unsupportedDomains.size > 0) {
+      if (unsupportedDomains.size === localDomains.size) {
+        return [true, true];
+      }
+      // TODO: Can we remove partial dangling securecookie rules???
```
bcyphers: Don't think so. This means that the author intended at least some of the target domains to use secure cookies, we shouldn't throw out that information

cschanaj:
@bcyphers Does e5ad3c5 looks good to you?

bcyphers:
Yeah, I can't find any other issues. I think this is good to go.

cschanaj:
@bcyphers @RReverser @Hainish Thanks for the effort reviewing this PR. linking #16122

cschanaj:
@bcyphers would you or @Hainish merge this PR? I would like to run this script before the next release and submit a PR to a downstream project in https://github.com/brave/https-everywhere-builder

Hainish:
@cschanaj sorry this didn't make it in to 2018.8.22.  This can be part of the ruleset release next week.

PR #16044, opened 2018-07-08
Deleted handy.de obsolete rule

dboecker:
this rule is obsolete, it was set by the origin maintainer of handy.de. These rules are redirecting to 404. Our page always redirects our traffic to the valid https file.


J0WI:
Ouch, this rule looks evil. But thankfully the site has proper HTTPS support now. Could you please create a new rule, that just redirects to HTTPS? A simple example can be found [here](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Ideenw3rk.de.xml).

PR #16053, opened 2018-07-12
Adding a warning to pages which 'Block all unencrypted requests' is u…

Hainish:
…nable to upgrade.


Hainish:
This will not work in Firefox ESR, since the `web_accessible_resources` requirement for extension URLs is not fully implemented there.  But it does not cause any dangerous behavior for ESR, and works with the latest Firefox as well as Chrome.

Hainish:
This can be tested on `http.badssl.com`.  @bcyphers can you review this?

jeremyn:
@Hainish You might consider linking to an FAQ from the cancel page to explain what "network-based downgrade attacks" are.

src/chrome/locale/en/https-everywhere.dtd
```diff
@@ -25,6 +25,9 @@
 <!ENTITY https-everywhere.prefs.reset_defaults "Reset to Defaults">
 <!ENTITY https-everywhere.prefs.reset_defaults_message "This will reset each ruleset to its default state. Continue?">
 
+<!ENTITY https-everywhere.cancel.he_blocking "HTTPS Everywhere is currently blocking all unencrypted requests.">
+<!ENTITY https-everywhere.cancel.he_blocking_explainer "This page could not be served over HTTPS, either because no HTTPS page is available, or because the HTTPS version of this website is being blocked. If you wish to view the unencrypted version of this page, you can still do so by disabling the 'Block all unencrypted requests' option in your HTTPS Everywhere extension. Be aware that disabling this option could make your browser vulnerable to network-based downgrade attacks on websites you visit.">
```
jsha: I think it would be helpful to also inform the user that they followed a link or entered a URL that was HTTP, and HTTPS Everywhere tried to rewrite it and failed. Something like:

"HTTPS noticed you were navigating to a non-HTTPS page, and tried to send you to the HTTPS version instead. The HTTPS version is unavailable. Most likely this site does not support HTTPS, but it is also possible that an attacker is blocking the HTTPS version. If you wish to view the unencrypted version of this page, you can still do so by disabling the 'Block all unencrypted requests' option in your HTTPS Everywhere extension."

I would leave out the "vulnerable to network-based downgrade attacks" since it's not strictly accurate. If you're clicking a non-HTTPS link, and HTTPS Everywhere tries a speculative upgrade, failing that upgrade is not a downgrade attack; it's just the status quo.

Hainish: > If you're clicking a non-HTTPS link, and HTTPS Everywhere tries a speculative upgrade, failing that upgrade is not a downgrade attack; it's just the status quo.

Explicit network downgrade attacks ala `sslstrip` will work when `Block all unencrypted requests` is disabled, but will not work when it's enabled.  Saying that disabling this option will make you vulnerable to these type of attacks is strictly accurate, or am I missing something?

I like the amended text otherwise.

jsha: Let's talk specifically about `sslstrip` attacks then (and we can handily link to sslstrip). I don't really consider sslstrip a downgrade attack: it just prevents you from receiving an (insecure) redirect.

Hainish:
@jeremyn it's rather difficult to add a link in the middle of translated text.  I've added another string translation for "network-based downgrade attacks," which is used to match the text in the explainer paragraph that we'd like to be linked.  As long as the link translation is character-by-character equivalent to the explainer-paragraph translation, the linked text will replace the unlinked text.

Hainish:
I could have added multple `<span>` tags for different places in the text, but I think that won't work for RTL scripts and is more confusing for translators.

Hainish:
I've also requested from our internal graphics team a red banner for this page, which seems more appropriate given the warning and circumstance.

Hainish:
Banner is updated

Hainish:
Good catch.  It should be changed to `{redirectUrl: cancelUrl}`. Fixed.

chromium/pages/cancel/ux.js
```diff
@@ -0,0 +1,21 @@
...
+  const explainer = document.querySelector("[data-i18n=cancel_he_blocking_explainer]");
+  observer = new MutationObserver(() => {replaceLink(explainer)});
+  if (explainer.innerText.length > 0) {
+    replaceLink(explainer);
+  } else {
+    observer.observe(explainer, {childList: true});
+  }
+});
+
+function replaceLink(explainer){
```
bcyphers: Neat trick!

PR #16064, opened 2018-07-16
Enhance explode regexp to support \d

cschanaj:
This will further trivialize 100+ additional rulesets when merged with #16029

cc @Hainish @RReverser it would be great if you could spare your time to help verify the changes.

### DIFF
--- a/utils/trivialize-rules/explode-regexp.js
+++ b/utils/trivialize-rules/explode-regexp.js
@@ -54,25 +54,37 @@ function explodeRegExp(re, callback) {
       }
 
       case 'charset': {
-        if (first.ranges.length === 1) {
-          let range = first.ranges[0];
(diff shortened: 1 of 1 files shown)


RReverser:
I intentionally excluded it, just like limited explicit number of ranges to exclude `0-9` and such. Adding it does trivialise more rules, but explodes target lists beyong reasonable limits.

cschanaj:
@RReverser I didn't notice that it was intentional. I agree that this will explode some `target` beyond reasonable limit, e.g. [sysprovide.xml](https://github.com/EFForg/https-everywhere/blob/812d5bb8aa51135f78f27368fe0edb36b6faadcb/src/chrome/content/rules/sysprovide.xml) will explode to 100 `target` which should be avoided. closing this PR accordingly.

PR #16088, opened 2018-07-23
Remove deprecated command line flags

jvasile:
These flags are no longer supported.  First they were deprecated, but now they're actually removed.  This patch deletes the flags because they were causing the pip install to fail.


J0WI:
LGTM, @Hainish does this need any follow up changes?

PR #15448, opened 2018-05-17
Measure and slightly improve memory usage

bcyphers:
Rebase of #12558 on (more) current master.

This branch also updates the unit tests to reflect the changes here.

PR #16092, opened 2018-07-25
Slightly reduce default.rulesets size and memory usage

cschanaj:
@Hainish @bcyphers This will merge ruleset exclusion into a single pattern as implemented in #15448. The memory improvement measure is as follow

Before:
Initial usage: 41.1 MB
Maximum usage: 80.3 MB

After:
Initial usage: 40.9 MB
Maximum usage: 80.1 MB

Related #12232


Hainish:
Thanks! Tested and verified it works as intended.

PR #16096, opened 2018-07-26
[thestar.com.my] Add redirect

youdly:
Redirect host with expired cert to valid host.


J0WI:
This has been fixed. Thanks anyway!

PR #16117, opened 2018-07-29
[foolcdn.com] Separate ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/16111

### DIFF
--- a/src/chrome/content/rules/Fool-CDN.com.xml
+++ b/src/chrome/content/rules/Fool-CDN.com.xml
@@ -1,0 +1,17 @@
+<!--
+	For other Motley Fool related rulesets, see Motley-Fool.xml
+-->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Fool-CDN.com.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+	For other Motley Fool related rulesets, see Motley-Fool.xml
+-->
+<ruleset name="Fool CDN.com">
+
```
J0WI: `foolcdn.com`

youdly: Added

J0WI:
Thanks!

PR #16126, opened 2018-08-01
A little more compact JSON representation in default.rulesets

cschanaj:
Before: 7390331 bytes
After: 6930115 bytes (93.77%) when uncompressed


Hainish:
Thank you!

PR #16127, opened 2018-08-01
Remove 'import settings' ux from options page

Hainish:
This is legacy code left over from the initial stages of planning the move from XPCOM to WebExtensions.  No longer used or needed.


Hainish:
I'd be interested in seeing how many people would be interested if we added an "export" functionality to match this import.  But we should probably simply remove this UX until a time where we have both.

PR #16129, opened 2018-08-04
fix web.mit.edu and related domains. They don't support HTTPS for the public

varenc:
web.mit.edu (and its identical www.mit.edu and mit.edu versions) only support HTTPS if the browser has an MIT issued certificate. This means HTTPS works fine if an MIT student/prof/affiliate is doing the testing but it does not work for the general public.  They get this error: 
image

The problem was introduced 6 months ago but only effects a narrow use case. The rule I’m deleting causes problems for non-MIT affiliates if they visit the domain roots “http://web.mit.edu“ or “http://mit.edu” or “http://www.mit.edu” directly, but not if they visit anything with a path like http://web.mit.edu/~jhawk (except for things under /files/).

This obscured the issue somewhat and perhaps took awhile for it to be noticed. 

That said, right now MIT’s base domains don’t support HTTPS for the public and HTTPS redirects for these shouldn’t be built into HTTPS Everywhere.

Cheers


Bisaloo:
Hi @varenc, thank you for your pull request. This looks good!

Can you have a look at travis' output and remove broken targets / rules please?

varenc:
update! MIT claims they're fixing this issue with anonymous HTTPS users on Friday this week: http://kb.mit.edu/confluence/pages/viewpage.action?pageId=152593506

I'll test after that goes out and if they fixed it, as they should, I'll happily close this PR.

varenc:
so MIT just rolled out HTTPS support and it's somewhat working.  The domain root website works well, but there's a very long tail of other mit.edu hosted sites with mixed-content that end up breaking over HTTPS.  https://web.mit.edu/puzzle/www/ vs http://web.mit.edu/puzzle/www/ for example relies on CSS requested over HTTP.

Given this, I'm inclined to abandon this diff.  MIT-complex.xml converts only the root urls and /files/* urls to HTTPS but leaves everything else untouched. Those places do work fine over HTTPS so this isn't breaking anyone anymore.  But I don't think these rules should be expanded to apply to all web.mit.edu URLs for the reasons cited above.  This makes sense since the rule was probably originally written by an MIT affiliate that had HTTPS access before it was open to the public.

Will close this PR soon unless anyone thinks otherwise!

Bisaloo:
That makes sense to me. I'm okay with closing this PR.

PR #16132, opened 2018-08-05
[Avis.com] Add subdomains

youdly:

PR #16148, opened 2018-08-05
[Avis.com.au] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/16132


youdly:
Updated in the comment.

PR #16169, opened 2018-08-05
[Avis.fr] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/16132

### DIFF
--- a/src/chrome/content/rules/Avis.fr.xml
+++ b/src/chrome/content/rules/Avis.fr.xml
@@ -5,6 +5,7 @@
 	Non-functional subdomains:
 		- $host		(m)
 		- pubs		(m)
+		- secure	(HTTP 403 error)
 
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Avis.fr.xml
```diff
@@ -0,0 +1,31 @@
...
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Avis.fr">
+
+	<target host="avis.fr" />
+	<target host="www.avis.fr" />
+	<target host="mirror.avis.fr" />
+	<target host="reporting.avis.fr" />
```
J0WI: `secure.avis.fr`

youdly: Added in the comment

PR #16255, opened 2018-08-12
[Budget.com.ar] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/16246

### DIFF
--- a/src/chrome/content/rules/Budget.com.ar.xml
+++ b/src/chrome/content/rules/Budget.com.ar.xml
@@ -1,0 +1,25 @@
+<!--
+	For other Budget coverage, see Budget.com.xml
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Budget.com.ar.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	For other Budget coverage, see Budget.com.xml
+
+
+	Non-functional subdomains:
+		- locations		(e)
```
J0WI: `mail.budget.com.ar`

youdly: Updated

PR #16279, opened 2018-08-12
[Budget-Jamaica.com] New ruleset

youdly:
https://github.com/EFForg/https-everywhere/pull/16246


youdly:
expired cert, might be fixed in a month.

PR #16287, opened 2018-08-13
Add Piwigo.com.xml

J0WI:


Bisaloo:
Unfortunately, this should probably have a `mixedcontent` `platform`. See for example https://historicalzg.piwigo.com/

J0WI:
Aren't these just blocked trackers?

Bisaloo:
In this case, yeah. However, I assumed that it meant users could add `http` resources on their pages. But maybe I'm wrong. Do you know more about this?

J0WI:
Let me consider a few more examples.

J0WI:
I checked now some more sites and only a minority of them have issues. None of them seemed to be critical, some of them are even enforcing HTTPS.

I think it's worth to enable this by default.

PR #16291, opened 2018-08-15
[PayPal.com] Add subdomains

youdly:
and remove obsolete rulesets.

PR #16292, opened 2018-08-15
[paypalobjects.com] Rename ruleset and add host

youdly:
https://github.com/EFForg/https-everywhere/pull/16291


youdly:
Resolve dangling reference error.

PR #16306, opened 2018-08-16
Update channels ux

Hainish:
Adding a UX to update channels, as well as functionality to add, remove, or edit update channels.

### DIFF
--- a/chromium/background-scripts/update.js
+++ b/chromium/background-scripts/update.js
@@ -25,19 +25,37 @@ let store,
 // how often we should check for new rulesets
 const periodicity = 86400;
 
-// jwk key loaded from keys.js
-let imported_keys = {};
-for(let update_channel of update_channels){
-  imported_keys[update_channel.name] = window.crypto.subtle.importKey(
-    "jwk",
-    update_channel.jwk,
-    {
-      name: "RSA-PSS",
-      hash: {name: "SHA-256"},
-    },
-    false,
-    ["verify"]
-  );
+let combined_update_channels;
+let imported_keys;
+
+// update channels are loaded from `background-scripts/update_channels.js` as well as the storage api
+async function loadUpdateChannelsKeys() {
+  util.log(util.NOTE, 'Loading update channels and importing associated public keys.');
+
+  const stored_update_channels = await store.get_promise('update_channels', []);
+  const combined_update_channels_preflight = update_channels.concat(stored_update_channels);
+
+  imported_keys = {};
+  combined_update_channels = [];
+
+  for(let update_channel of combined_update_channels_preflight){
+
+    try{
+      imported_keys[update_channel.name] = await window.crypto.subtle.importKey(
+        "jwk",
+        update_channel.jwk,
+        {
+          name: "RSA-PSS",
+          hash: {name: "SHA-256"},
+        },
+        false,
+        ["verify"]
+      );
+      combined_update_channels.push(update_channel);
+    } catch(err) {
+      util.log(util.WARN, update_channel.name + ': Could not import key.  Aborting.');
+    }
+  }
 }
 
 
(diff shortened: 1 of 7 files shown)


Hainish:
If anyone wants to test this functionality, you can use the following to add a `SecureDrop` update channel to Tor Browser:

JWK:
```json
{"kty":"RSA","n":"zv4NkDYgBL5n1LY_Pl2WWmiHV_IApqtXTXUL6STgUS3G_ZFORA5Qx_7x5TMA9OWeKsN9kgDkpiZWs_cUZCfWLkYeP6MRmtOe-50Wzca99qb8hQfBT81Pnir-hadlISgcuosZ
FWfEmaEPXHLI47ZgnelDvBcVD8XyOH5qc4umNCZIeFl_cQApv_t7KDo0HRr4fvWqWWhLwfJJU3HPNUr7VHAyEdU6whllOUl3uSEt7zXEfODnxFxUFILUmIXfqFS8c8VcI20l-8QXBijbHRXGTKpAj2
Sdu9v5pGL0-5hh3iXISlC3rRpNjwBhdlQSWK9V30Cfp163bOP4WekSt51J5w","e":"AQAB"}
```

Path Prefix:
`https://inputoutput.io/non-wp/securedrop/`

This will make it so that in any instance of Tor Browser, navigating to `theintercept.com.securedrop` (or a similar URL for any other SecureDrop) will redirect you to the SecureDrop instance.  This is a proof of concept and shouldn't be used yet for SecureDrop instances if privacy is required.

You can use `test/tor-browser.sh` to test this in Tor Browser.

Hainish:
You can also publish rulesets of your own by placing them in `rules/`, running `python3.6 utils/merge-rulesets.py`, generating an RSA key and using the `pem-jwk` npm package to get the JWK from it, and using the scripts in `utils/sign-rulesets/` to throw it up on an endpoint. Though that may be a lot of work.

Hainish:
cc @redshiftzero

chromium/background-scripts/update.js
```diff
@@ -170,7 +187,9 @@ async function applyStoredRulesets(rulesets_obj){
   const rulesets_jsons = await Promise.all(rulesets_promises);
   if(rulesets_jsons.join("").length > 0){
     for(let rulesets_json of rulesets_jsons){
```
cschanaj: It seems to me that EFF's update channels will always have the top priority in rewriting URLs, while other channels are prioritize by insertion order. If the coverage of different channels overlaps, it might not works as the users expected, e.g. some channels might never apply. If this is the case, it might be good to have a way to prioritize the channels (possibly in later PR/ release). Please correct me if I am wrong.

Hainish: Yes I think prioritization of update channels makes sense, but this will require some re-engineering.  This would be a good task for a future PR.

chromium/pages/options/ux.js
```diff
@@ -39,6 +39,208 @@ document.addEventListener("DOMContentLoaded", () => {
...
+    const update_channel_path_prefix_column_left = document.createElement('div');
+    update_channel_path_prefix_column_left.className = "update-channel-column-left";
+    update_channel_path_prefix_column_left.innerText = "Path Prefix:";
+    update_channel_row_path_prefix.appendChild(update_channel_path_prefix_column_left);
+    const update_channel_path_prefix_column_right = document.createElement('div');
+    update_channel_path_prefix_column_right.className = "update-channel-column-right";
+    update_channel_row_path_prefix.appendChild(update_channel_path_prefix_column_right);
+    const update_channel_path_prefix = document.createElement('input');
+    update_channel_path_prefix.setAttribute("type", "text");
+    update_channel_path_prefix.className = "update-channel-path-prefix";
```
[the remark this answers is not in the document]
Hainish: I can envision `http` update channels over `onion` services, or network administrators who wish to mirror update channels locally by DNS redirection (as is sometimes done with `apt`).  I'd like to allow update channel operators to choose their own transport layer.

PR #16307, opened 2018-08-17
Create ruleset: SPI Inc

franklinyu:
I’m not sure whether I should include the redirecting sites and/or HSTS (non-preload) sites. Also note that some sites only supports TLS 1.0.

### DIFF
--- a/src/chrome/content/rules/spi-inc.org.xml
+++ b/src/chrome/content/rules/spi-inc.org.xml
@@ -1,0 +1,17 @@
+<!--
+	Mismatch:
+		oldwww.spi-inc.org
+	443 port closed:
+		lists.spi-inc.org
+-->
+<ruleset name="SPI Inc">
+	<target host="spi-inc.org" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/spi-inc.org.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+	Mismatch:
+		oldwww.spi-inc.org
+	443 port closed:
+		list.spi-inc.org
```
J0WI: Should be `lists.spi-inc.org`

src/chrome/content/rules/spi-inc.org.xml
```diff
@@ -0,0 +1,18 @@
...
+	HSTS:
+		members.spi-inc.org
```
J0WI: Please add it as target anyway, since it's not HSTS preloaded.

franklinyu:
Any news? This is ready to review/merge.

J0WI:
Sorry, I don't get notified if you force push.
Thanks for the additions!

PR #16350, opened 2018-08-19
Update Gtimg.com.xml

ivysrono:


src/chrome/content/rules/Gtimg.com.xml
```diff
@@ -15,6 +15,10 @@
 		<!-- Find the test url in : https://888.qq.com/ -->
 		<test url="http://888.gtimg.com/new_js/info/marketing/statistics_fuc.js" />
 
+	<target host="ac.gtimg.com" />
```
J0WI: Please remove the obsolete comment on the top.

PR #16352, opened 2018-08-19
Update and rename 163.com-problematic.xml to 163.com-mixedcontent.xml

ivysrono:


src/chrome/content/rules/163.com-mixedcontent.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+	For rules not causing problems, see 163.com.xml.
```
J0WI: The reference in `163.com.xml` needs to be updated.

PR #16354, opened 2018-08-19
Update and rename lofter.com.xml to lofter.com-mixedcontent.xml

ivysrono:


src/chrome/content/rules/lofter.com-mixedcontent.xml
```diff
@@ -1,32 +1,23 @@
 <!--
-
-Mismatch:
-	^lofter.com
```
J0WI: Wrong content

PR #16355, opened 2018-08-19
Update Sina.com.cn.xml

ivysrono:


ivysrono:
@J0WI Passed. Thanks.

PR #16356, opened 2018-08-19
Update and rename Flyspray.org.xml

J0WI:


src/chrome/content/rules/Flyspray.org.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Nonfunctional hosts in *.flyspray.org:
+		- flyspray.org (m)
+		- scm.flyspray.org (m)
```
Bisaloo: snaps.flyspray.org

J0WI: Added, thanks!

Bisaloo:
Thank you!

PR #16359, opened 2018-08-20
Fix complex rewrites in Firefox HTTP Nowhere mode, close #16358

### LINKED ISSUE #16358: Complex rewrites do not work in httpNowhere mode on Firefox
Type: code issue

## Platform

Firefox 61.0.2 (64-bit) on debian 9; technically should affect tor browser if they supported `upgradeToSecure`

## Step to reproduce

1. Visit http://marketwatch.com/ with "Block all unencrypted requests" checked

2. Expect URL rewritten to https://www.marketwatch.com/ because of an existing rule

https://github.com/EFForg/https-everywhere/blob/1e529b38651b77aae9ec87d372799e8e3c6d84de/src/chrome/content/rules/Marketwatch.com.xml#L44-L45

3. Got expired certificate errors

Apparently. this does not affect chromium browsers, nor Firefox with HTTPSE with default settings, likely a regression from #15157. ~Will submit a fix soon~ #16122  .

cschanaj:


cschanaj:
ping @Hainish

Hainish:
Thank you!

PR #16362, opened 2018-08-20
Update Itc.cn.xml

ivysrono:


src/chrome/content/rules/Itc.cn.xml
```diff
@@ -48,5 +54,7 @@
 
 	<securecookie host="^\w" name=".+" />
 
+	<rule from="^http://i7\.itc\.cn/" to="https://i0.itc.cn/" />
```
J0WI: I don't get the same content on the provided test url.

ivysrono: Oh, it seems like server error.
I'm sure i7 is same to i[0-6]. What should I do?

J0WI: Can you provide other test urls?

PR #16363, opened 2018-08-20
Update Sohu.com.xml

ivysrono:


src/chrome/content/rules/Sohu.com-mixedcontent.xml
```diff
@@ -0,0 +1,28 @@
+<!--
+	Other Sohu rulesets:
+		- Sohu.com.xml
+		- Sohu.com.cn.xml
+		- Sohu.net.xml
+		- SohuCS.com.xml
+		- SohuSCE.com.xml
```
J0WI: Please just use `For other Sohu coverage, see Sohu.com.xml` to avoid additional maintenance.

J0WI:
Thanks!

PR #16365, opened 2018-08-20
Ruleset for www.nic.cl

huguei:
We're issuing a 302 redirect from plain http since a couple of weeks.

### DIFF
--- a/src/chrome/content/rules/nicchile.xml
+++ b/src/chrome/content/rules/nicchile.xml
@@ -1,9 +1,17 @@
+<!--
+   Other nic.cl websites should not be added to HTTPs everywhere
+   until infrastructure changes and testing validations.
+
+   For example
+     - blog.nic.cl and status.nic.cl are hosted in third-party servers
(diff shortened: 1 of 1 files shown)


huguei:
Thanks J0WI.
But our rules are just for nic.cl, www.nic.cl and clientes.nic.cl websites. The other are hosted in another infrastructure and aren't subject to the same policies.
I wonder if my securecookie is the problem? Is it too wide? I can change it to

`<securecookie` host="^(www\.|clientes\.)?nic\.cl$" name=".+" /> 

if it's more clear.

Thanks.

J0WI:
@huguei we try to keep our rulesets complete and the comments are helping us to maintain the rules.
If there is a strong reason to not add working hosts like `blog.nic.cl`, you should also document this in comments, otherwise somebody might add it accidentally.

As long as there are no exclusions, the securecookie rule should be fine.

huguei:
Updated pull with comments and new cookie rule, as requested. Thanks.

J0WI:
Thanks for documenting this!

PR #16368, opened 2018-08-21
Avoid unnecessary trivial rewrites in Chromium browsers

cschanaj:
On chromium browser, `upgradeToSecure` is not available, avoid the unnecessary trivial rewrites should improve performance.  Besides, since `ruleset.apply` might return `null` when the URL is excluded by a ruleset, the checking might be performed more than once in Firefox/Chromium. Move the check outside of the loop avoid unused variables and assignments.


Hainish:
Thanks!

PR #16388, opened 2018-08-24
Create ruleset: Klwines.com

mattrichards:
Tested using Firefox Developer Edition.

### DIFF
--- a/src/chrome/content/rules/Klwines.com.xml
+++ b/src/chrome/content/rules/Klwines.com.xml
@@ -1,0 +1,10 @@
+<ruleset name="Klwines.com">
+	<target host="klwines.com" />
+	<target host="www.klwines.com" />
+	<target host="m.klwines.com" />
+
+	<rule from="^http:" to="https:" />
+
+	<test url="http://www.klwines.com/" />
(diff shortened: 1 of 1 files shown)


Bisaloo:
Hi, thank you for your pull request. You will need to do some modifications before this is ready to merge:

- You are missing some subdomains such as `blog.klwines.com`, `spiritsjournal.klwines.com`, `onthetrail.klwines.com`. Subdomains that aren't functional over HTTPS still need to be listed in the top comment along the relevant error, as demonstrated in our [contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts).

- As you can see, the continuous integration tests failed because of "duplicated test url". Indeed, each target will create an [implicit test url](https://github.com/EFForg/https-everywhere/blob/57a41c2edd22e4ee60c61691dba3c210085ee1e8/ruleset-testing.md) so you don't need to duplicate them.

- Finally, https://klwines.com/ does not load and should be removed from the list of targets (and moved to the relevant section in the top comment).

Let me know if you have questions!

PR #16395, opened 2018-08-26
Update Android.xml, add messages.android.com

cschanaj:


src/chrome/content/rules/Android.xml
```diff
@@ -1,33 +1,23 @@
...
-	<target host="*.android.com" />
-
-	<test url="http://developer.android.com/" />
-	<test url="http://market.android.com/" />
-	<test url="http://source.android.com/" />
-	<test url="http://tools.android.com/" />
+	<target host="www.android.com" />
+	<target host="developer.android.com" />
+	<target host="market.android.com" />
+	<target host="messages.android.com" />
```
J0WI: `partner.android.com`

PR #16410, opened 2018-08-27
Create AlTafsir.com.xml

Giltyhub:


src/chrome/content/rules/AlTafsir.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+	Unable to Connect:
+		main.altafsir.com
+	Invalid Certificate:
+		m.altafsir.com
+		mobile.altafsir.com
```
J0WI: Wildcard DNS

Giltyhub: @J0WI thanks added

PR #16425, opened 2018-08-28
Update hsbc.com.hk, add more subdomains

cschanaj:


cschanaj:
I believe it was a mistake when I copy & paste the lines.. just pushed 5c2dc15 and added back `^`

PR #16443, opened 2018-08-31
adding several top-level-domains should lead to https://posteo.de/ we…

tscpd:
…bsite


src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://m.posteo.de/ => https://m.posteo.de/: (6, 'Could not resolve host: m.posteo.de')
+	Fetch error:
+
+		http://m.posteo.de/ => https://m.posteo.de/: (6, 'Could not resolve host: m.posteo.de')
```
J0WI: You can remove this comment since the host is gone.

src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
...
+
+	Fully covered subdomains:
```
J0WI: This comment is no longer needed, since all targets are listed below.

src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
...
-    <target host="cdn.posteo.de" />
-    <target host="m.posteo.de" />
-    <target host="lists.posteo.de" />
-    <target host="www.posteo.de" />
+	<target host="posteo.de" />
+	<target host="api.posteo.de" />
+	<target host="autodiscover.posteo.de" />
+	<target host="cdn.posteo.de" />
+	<target host="lists.posteo.de" />
+	<target host="www.posteo.de" />
```
J0WI: Please have a look at our [target ordering](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#target-ordering) guideline.

src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
...
+	<target host="www.posteo.si" />
+	<target host="posteo.tn" />
+	<target host="www.posteo.tn" />
+	<target host="posteo.uk" />
+	<target host="www.posteo.uk" />
+	<target host="posteo.us" />
+	<target host="www.posteo.us" />
 
-    <securecookie host="^(www\.|m\.|lists\.|autodiscover\.|api\.|cdn\.)?posteo\.de" name=".+" />
+	<securecookie host="^(www\.|lists\.|autodiscover\.|api\.|cdn\.)?posteo\.de" name=".+" />
```
J0WI: You can just use `<securecookie host=".+" name=".+" />`

src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
...
 
-    <rule from="^http:"
-            to="https:" />
+	<rule from="^http:" to="https://posteo.de/" />
```
J0WI: This is not a valid syntax

youdly:
On line [123](https://github.com/EFForg/https-everywhere/commit/690844fe9cc7233b483cf2dd3f54af4bac74ff28#diff-f85aa45f7f1f53ec8198f797bcfa6bf3R123), current `<rule>` redirects all above `<target host>` to posteo.de, which is not desirable in this domain. e.g. `api.posteo.de` will be redirected to `posteo.de` even though the subdomain is valid.
It should be `<rule from="^http:" to="https:" />`, so `http://www.posteo.de/` -> `https://www.posteo.de/`.
In this case, server will redirect `https://www.posteo.de/` to `https://posteo.de/`.

It might be faster if `http://www.posteo.de/` redirect straight to `https://posteo.de`, but since `https://www.posteo.de` is functional (has valid cert), https-e doesn't need to preempt it.
That is unless if `https://www.posteo.de/` is not functional, then only then https-e redirects it `https://posteo.de`, like this [example](https://github.com/EFForg/https-everywhere/pull/16148/files#diff-3285e29c969f251535c8c851aed70114R27).

src/chrome/content/rules/Posteo.xml
```diff
@@ -1,30 +1,124 @@
...
--->
-<ruleset name="Posteo.de" default_off='failed ruleset test'>
-    <target host="posteo.de" />
-    <target host="api.posteo.de" />
-    <target host="autodiscover.posteo.de" />
-    <target host="cdn.posteo.de" />
-    <target host="m.posteo.de" />
-    <target host="lists.posteo.de" />
-    <target host="www.posteo.de" />
+	<target host="posteo.de" />
```
pipboy96: `posteo.de` is preloaded.

PR #16445, opened 2018-08-31
[Microsoft.xml] Remove obsolete exclusion

xiaoyinl:
Since www.microsoft.com sends HSTS, excluding http://www.microsoft.com/en-us/software-recovery is unnecessary. Also this page (both http and https) now redirects to https://www.microsoft.com/software-download/home.


xiaoyinl:
I don't quite understand this error:
`ERROR src/chrome/content/rules/Microsoft.xml: Not enough tests (0 vs 1) for search.microsoft.com`.

There's `<test url="http://search.microsoft.com/default.aspx" /> ` at L509. What else is needed?

xiaoyinl:
There're a lot of fetch errors. I am sorry I don't want to spend too much time on fixing them all. So close this PR.

PR #16446, opened 2018-08-31
Documenting ruleset update channels

Hainish:


Hainish:
cc @cschanaj @J0WI FYI.

docs/en_US/ruleset-update-channels.md
```diff
@@ -0,0 +1,177 @@
...
+    git clone --depth=1 https://github.com/EFForg/https-everywhere.git
+
+or
+
+    curl -sLO https://github.com/EFForg/https-everywhere/archive/master.zip; unzip master.zip; mv https-everywhere-master https-everywhere
+
+Next,
+
+    cd https-everywhere
+    rm rules/*.xml
```
cschanaj: One can save the time `rm` the existing rulesets by pointing `utils/merge-rulesets.py` to the correct directory. 

https://github.com/EFForg/https-everywhere/blob/b87b31b1b1ff98b1902ea48e1d36ed682ef13f15/utils/merge-rulesets.py#L25-L27

Hainish: I want to avoid the possibility that a ruleset channel maintainer might forget this `--source_dir` flag in the future, and I don't see any harm in simply removing the `xml` files in `rules` for anyone maintaining a non-EFF ruleset channel.  I'm inclined to keep it, since adding a `mkdir` and pointing to it using `--source_dir` seems about as complicated as what we currently have here.

Hainish:
also cc @Bisaloo @bcyphers

PR #16453, opened 2018-09-01
Update ruleset: Omtrdc.net

injust:


J0WI:
Does this actually change/fix something?

src/chrome/content/rules/Omtrdc.net.xml
```diff
@@ -1,46 +1,34 @@
...
-		<test url="http://www.totalsystemservices.d1.sc.omtrdc.net/optout.html" />
-		<test url="http://www.wackerchemie.d1.sc.omtrdc.net/optout.html" />
 
-	<rule from="^http:"
-		to="https:" />
+	<rule from="^http://([\w-]+)\.(sc|tt)\.omtrdc\.net/"
+		to="https://$1.$2.omtrdc.net/" />
 
+	<rule from="^http://([\w-]+)\.d([1-3])\.sc\.omtrdc\.net/"
+		to="https://$1.d$2.sc.omtrdc.net/" />
```
zoracon: Found that regex groups become potentially problematic over time. Could you take the possible targets and list them above? Or perhaps they should rely on the wildcard above? Otherwise that would be preferred with the trivial rule `	<rule from="^http:"
		to="https:" />` still in place.

J0WI:
@injust would be great if you could have a look here.

injust:
The existing ruleset would redirect `http://a.b.d1.sc.omtrdc.net/optout.html` to `https://a-b.d1.sc.omtrdc.net/optout.html` (note the hyphen), which is a different page.
It would also force HTTPS on a URL like `http://a.b.c.d1.sc.omtrdc.net/`, which is not covered by the certificate.

src/chrome/content/rules/Omtrdc.net.xml
injust: The wildcard `<target>` is necessary because all subdomains exist.

I don't see a way to do the `<rule>`s without regex groups. We want to redirect `http://a.sc.omtrdc.net` but not `http://a.a.sc.omtrdc.net`.

ivysrono:
#17578

PR #16458, opened 2018-09-02
Update ruleset: WSJ.net

injust:


injust:
I find it quite interesting how the validations test for [this build](https://travis-ci.org/EFForg/https-everywhere/builds/423522907) managed to pass, especially when `<target host="s.wsj.net" />` is included twice in the same ruleset (which caused [the build before that](https://travis-ci.org/EFForg/https-everywhere/builds/423522278) to fail).

injust:
Done, opened #16468

PR #16467, opened 2018-09-04
Update ruleset: Electronic Arts

injust:

PR #16469, opened 2018-09-04
Create ruleset: TheSims3.com

injust:


injust:
Blocked on #16467 (for removal of targets from `Electronic-Arts.xml`)

PR #16470, opened 2018-09-04
Create ruleset: FBINAA.org

injust:


Slayer2469:
Thank you. Should I let poliiice know

J0WI:
@Slayer2469 it would be great if you manage to set up preloaded HSTS: https://hstspreload.org/

PR #16517, opened 2018-09-12
Add ruleset: Python-RQ.org

dnoetzel:


src/chrome/content/rules/Python-RQ.org.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+	Nonfunctional hosts in *.python-rq.org:
+		- www
```
Bisaloo: This subdomain does not exist so it can be completely omitted.

PR #16543, opened 2018-09-17
Fix an invalid securecookie pattern

cschanaj:
It seems to me that a wrong pattern was used... cc @Hainish 

```bash
$ grep -lF '(?:w*\.)' *.xml  | xargs sed -i 's/?:w/?:\\w/g'
```


cschanaj:
closing as `(?:w*\.)` seems to mean `(?:www\.)?` instead of a wildcard... this pattern is equal to a trivial securecookie rule if the ruleset covers only `^` and `www`. I might submit a PR later to trivialize them.

PR #16549, opened 2018-09-18
[myetherwallet.com] Remove preloaded rule 

ccfb3ee765a58cae:
Seems this was added to preload a while back.


Bisaloo:
Thank you for your PR! We have a dedicated automated tool to remove preloaded rulesets in `utils/hsts-prune`. Preloaded targets / rulesets should be removed using this tool.

PR #16579, opened 2018-09-22
Add Ossec.net.xml

J0WI:
#15750


src/chrome/content/rules/Ossec.net.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Nonfunctional hosts in *.ossec.net:
+		- ossec.net (s)
```
Bisaloo: I get a mismatch here

J0WI: Thanks, updated.

PR #16581, opened 2018-09-22
Add Gtk.org.xml

J0WI:
#15750


Bisaloo:
Mismatched: beast.gtk.org

J0WI:
Added, thanks!

PR #16584, opened 2018-09-22
Add Libspf2.org.xml

J0WI:
#15750


Bisaloo:
I get an error 500. Is it the same for you?

J0WI:
Yes, seems to be down right now. :confused:

PR #16595, opened 2018-09-23
Remove wildcard from Sunrise.xml (fix #15862)

### LINKED ISSUE #15862: Change ruleset for sunrise.ch
Type: ruleset issue
Domain: sunrise.ch

When the ruleset for "sunrise.ch" is active, it also affects the "idp.sunrise.ch" subdomain that (currently) does not support HTTPS (tested over multiple days). This stops the user from accessing most parts of the customer area. The only workaround is to disable the ruleset. 

My requested change is to exclude the subdomain from the ruleset, as it breaks the website.

J0WI:


src/chrome/content/rules/Sunrise.xml
```diff
@@ -1,14 +1,67 @@
...
+	<target host="suota.sunrise.ch" />
+	<target host="travel.sunrise.ch" />
+		<target host="serviceportal.vpbx.sunrise.ch" />
+	<target host="webmail.sunrise.ch" />
+	<target host="webmail1.sunrise.ch" />
+	<target host="webmail2.sunrise.ch" />
+	<target host="www1.sunrise.ch" />
 
-    <securecookie host="^(?:.*\.)?sunrise\.ch$" name=".+" />
+	<securecookie host="^(?:.*\.)?sunrise\.ch$" name=".+" />
```
cschanaj: AFIAK, this can be replaced with `.+` without affecting the ruleset behavior

J0WI: Updated

PR #16649, opened 2018-09-30
add spiegel-de subdomain

numismatika:
also re-enables the ruleset

### DIFF
--- a/src/chrome/content/rules/Spiegel.xml
+++ b/src/chrome/content/rules/Spiegel.xml
@@ -1,63 +1,31 @@
-
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://ophirum.spiegel.de/ => https://ophirum.spiegel.de/: (51, "SSL: no alternative certificate subject name matches target host name 'ophirum.spiegel.de'")
-
 	Other Spiegel Group rulesets:
 
 		- Spiegel-QC.xml
 
+	Cert mismatch:
+		 ophirum.spiegel.de
 
-	CDN buckets:
-
-		- cdn.spiegel.de.edgesuite.net
-
-			- cdn
-			- cdn[1-4]
-
-
-	Nonfunctional subdomains:
-
-		- orange *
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Spiegel.xml
```diff
@@ -110,27 +62,12 @@ Fetch error: http://ophirum.spiegel.de/ => https://ophirum.spiegel.de/: (51, "SS
...
-
-	<securecookie host="^(?:abo|count|magazin|ophirum)\.spiegel\.de$" name=".+" />
-
-
-	<!--	- Cert only matches www
-		- Unsupported paths redirect to http
-		- cdn\d?: Akamai
-				-->
-	<rule from="^http://cdn\d?\.spiegel\.de/"
-		to="https://www.spiegel.de/" />
```
J0WI: Why did you remove this?

src/chrome/content/rules/Spiegel.xml
```diff
@@ -110,27 +62,12 @@ Fetch error: http://ophirum.spiegel.de/ => https://ophirum.spiegel.de/: (51, "SS
 			<test url="http://www.spiegel.de/pics/43/0,1020,308043,00.jpg" />
 			<test url="http://www.spiegel.de/static/sys/dimensionspixel.gif" />
 
+	<securecookie host=".+" name=".+" />
```
J0WI: This is quite dangerous since we exclude major party of the site.

src/chrome/content/rules/Spiegel.xml
numismatika: Because the first step when i am cleaning up old rulesets is deleting all special rewrites to see if it works without them. Also there are no test urls to be found , neither in google nor on the spiegel website.

J0WI: I can find many examples when browsing on the site. With the rule they get secured, but without it you can see the requests to `cdn([1-4])?`.

PR #16658, opened 2018-10-01
Create ruleset: AdAstraArms.com

injust:


src/chrome/content/rules/AdAstraArms.com.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="AdAstraArms.com">
```
pipboy96:
```
<!--
	Mismatched:
		- ftp
-->
```

pipboy96:
@zoracon I'm not sure if adding this comment is critical (in case @injust won't add it) or this may be merged as it is.

PR #16660, opened 2018-10-01
Add ruleset for coletivos.org

Perflyst:


src/chrome/content/rules/Coletivos.org.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="Coletivos.org">
+	<target host="coletivos.org" />
```
Bisaloo: Could you also add `www.coletivos.org` as a target with a special redirect rule `<rule from="^http://www\.coletivos\.org/" to="https://coletivos.org/"> and a comment explaining why this is necessary (invalid certificate) please?

Perflyst: Done

Bisaloo:
Some things have changed for this domain since my last review:

- https://discourse.coletivos.org/ doesn't work anymore so you should remove the target
- `www.coletivos.org` doesn't work over HTTP either so you can remove it alongside the comment and the special rule.

FYI, even we could keep all targets, the ruleset wouldn't have work in its current form as rules are applied in the order they appear.

So http://www.coletivos.org would be rewritten as https://www.coletivos.org (L15) before it gets the chance to apply the rule from L16.

PR #16667, opened 2018-10-02
Update ruleset: Posthaven.com

injust:


src/chrome/content/rules/Posthaven.com.xml
```diff
@@ -1,31 +1,12 @@
...
-	* Works; mismatched, CN: posthaven.com
-
--->
-<ruleset name="Posthaven.com (partial)">
-
+<ruleset name="Posthaven.com">
 	<target host="posthaven.com" />
-	<target host="www.posthaven.com" />
-
+	<target host="*.posthaven.com" />
```
J0WI: Please add a comment why the wildcard is required.

J0WI:
Thanks!

PR #16675, opened 2018-10-04
Disable modifying headers on disabled pages

CrafterKina:
HTTPS Everywhere modifies access-control-allow-origin on HTTP nowhere mode. Even if you disabled HTTPS Everywhere with the "Disable HTTPS Everywhere on this site" feature, this was rewritten and the problem occurred. This PR disables modifying on the sites.


CrafterKina:
not work for chrome...

PR #16681, opened 2018-10-05
Create Alukah.net.xml

Giltyhub:


src/chrome/content/rules/Alukah.net.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Timeout:
+		alukah.net
+		api.alukah.net
+		en.alukah.net
+		fatawa.alukah.net
```
Bisaloo: I get `refused` here

PR #16702, opened 2018-10-09
archive.org fallback for 'block http'

arthuredelstein:
Hi! Here's my attempt at a patch to let users who have decided to "Block all unencrypted requests" to view a non-HTTPS page via an encrypted connection to archive.org.

I'm happy to revise as needed. :)

### DIFF
--- a/chromium/pages/cancel/ux.js
+++ b/chromium/pages/cancel/ux.js
@@ -11,7 +11,10 @@ document.addEventListener("DOMContentLoaded", () => {
   } else {
     observer.observe(explainer, {childList: true});
   }
-  displayURL();
+  const cancelURL = new URL(window.location.href);
+  const originURL = decodeURI(cancelURL.searchParams.get('originURL'));
+  displayURL(originURL);
+  displayWaybackMachineLink(originURL);
 });
 
 function replaceLink(explainer){
@@ -23,9 +26,7 @@ function replaceLink(explainer){
   explainer.innerHTML = explainer.innerHTML.replace(linkText, link.outerHTML);
 }
 
-function displayURL() {
-  const cancelURL = new URL(window.location.href);
-  const originURL = decodeURI(cancelURL.searchParams.get('originURL'));
+function displayURL(originURL) {
   const originURLLink = document.getElementById('originURL');
   originURLLink.innerText = originURL;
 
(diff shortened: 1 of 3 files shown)


Giltyhub:
i like the idea :)

chromium/pages/cancel/ux.js
```diff
@@ -38,3 +39,23 @@ function displayURL() {
     }
   });
 }
+
+function displayWaybackMachineLink(originURL) {
+  const waybackMachineURLLink = document.getElementById("waybackMachineURL");
+  waybackMachineURLLink.innerText = chrome.i18n.getMessage("cancel_he_blocking_internet_archive_link");
+  waybackMachineURLLink.addEventListener("click", async function () {
+    const response = await fetch("http://archive.org/wayback/available?url=" + encodeURI(originURL));
```
Giltyhub: `await fetch("http://archive.org/wayback/`

I'm probably misunderstanding this part, but why is `http://` instead of `https://`?

arthuredelstein: Good catch! (I didn't notice this mistake because the link was rewritten to https anyway.) I have pushed a fix. Thanks.

J0WI:
I'm not sure if this solution is suitable/in the scope of HTTPS Everywhere. The archive does not provide an up-to-date or interactive version. There are already many VPN or Proxy solutions providing this functionality. On the other hand it's a quite lightweight change. 

Let's ask @cschanaj @Bisaloo @Hainish @bcyphers for some more thoughts.

cschanaj:
I am more inclined to think this is beyond the scope of HTTPS Everywhere. This is because this option will redirect users to a third-party site archive.org with a less update-to-date/ possibly broken version. I believe HTTPSE users should be more adequately informed about possible breakage and privacy issues when they get prompted. 

P.S. Perhaps it is worth updating [Archive.org_Way_Back_Machine.xml](https://github.com/EFForg/https-everywhere/blob/a3473d2f0a6829fd603dba9279f733d674691f5e/src/chrome/content/rules/Archive.org_Way_Back_Machine.xml) after this is merged.

arthuredelstein:
> giving users multiple paths forward after reaching an error without a clear preference seems like an anti-pattern.

That’s an interesting point. I have been using this for the past couple of weeks and my usage flow is to always click the archive.org link. It’s quite satisfying to never, ever have to use HTTP! :) So maybe using archive.org for non-HTTPS domains should be a global setting rather than a repeated choice?

zoracon:
On one hand this provides a good option for encrypted connection. However, I would like the user to be in the know with the current state of the site at the point of the interstitial page.

I like the idea of maybe incorporating this in as a setting, but not so much at the point of the interstitial page when the user is in "Block all un-encrypted requests". Where the plugin displays that it is in fact doing so.

Hainish:
I'd be willing to merge this with the addendum of a global setting to add the link to the interstitial page, defaulted to off.

We are also discussing how to re-brand "Block all unencrypted requests" to something nicer.  This may affect the string translation bit of this PR.

Giltyhub:
only inconvenience here is when a page blocks the ia bot in robots.txt in which case the page won't be accessible w/t this method

Giltyhub2:
@arthuredelstein can you fix the conflicts now? thanks again for making this pull

pipboy96:
@Giltyhub2 I may create a fixed PR without a conflict. I don't feel this option will add any security, since Web Archive fetches the page over HTTP anyway, and if any malicious content was added, it would be in the archived page anyway. The copy is also non-interactive and outdated.

pipboy96:
@arthuredelstein If you want this reopened, just ask. For now I will close this as an abandoned PR since there were no updates for months.

PR #16716, opened 2018-10-10
Update ruleset: mheducation.com.xml

CoyoteCiphers:


src/chrome/content/rules/mheducation.com.xml
```diff
@@ -74,14 +48,15 @@ Fetch error: http://careers.mheducation.com/ => https://mhecareers.silkroad.com/
 	<target host="dev.mheducation.com" />
 	<target host="highered.mheducation.com" />
 	<target host="images-cf.mheducation.com" />
-	<target host="novellaqalive.mheducation.com" />
+	<!-- <target host="novellaqalive.mheducation.com" /> -->
```
zoracon: To clean this up some, just add the non-working hosts to list above and delete these lines

src/chrome/content/rules/mheducation.com.xml
```diff
@@ -96,7 +71,7 @@ Fetch error: http://careers.mheducation.com/ => https://mhecareers.silkroad.com/
 	<!--	Complications:
 				-->
 	<target host="mheducation.com" />
-	<target host="careers.mheducation.com" />
+	<!-- <target host="careers.mheducation.com" /> -->
```
zoracon: Same as above and for the commented out lines below as well.

src/chrome/content/rules/mheducation.com.xml
CoyoteCiphers: They actually happened to be hosts that no longer existed, so I just removed them.

PR #16729, opened 2018-10-13
Update W3C.xml

xiaoyinl:
https://people.w3.org/ is no longer mismatched.


xiaoyinl:
The connection timed out. They can be connected from my computer.

xiaoyinl:
It seems https://people.w3.org/ redirects to https://people.w3.org/webmail/, which differs from http://people.w3.org/. Other pages show the same content, e.g. https://people.w3.org/janet/. But since these are user pages, it's likely to have mixed content. I will remove this rule, but change the comment to "mixed content".

PR #16730, opened 2018-10-13
Delete AKA.ms.xml

xiaoyinl:
aka.ms is preloaded. https://hstspreload.org/?domain=aka.ms


xiaoyinl:
Is there a way I can remove the dangling reference from Microsoft.xml without triggering a revalidation of Microsoft.xml? That file is so complicated to fix.

J0WI:
We use a tool to periodically remove preloaded targets: https://github.com/EFForg/https-everywhere/tree/master/utils/hsts-prune
Thanks anyway!

PR #16731, opened 2018-10-13
Delete Bing.xml

xiaoyinl:
Bing is preloaded with subdomain included.
https://www.ssllabs.com/ssltest/analyze.html?d=bing.com
https://hstspreload.org/?domain=bing.com


J0WI:
We use a tool to periodically remove preloaded targets: https://github.com/EFForg/https-everywhere/tree/master/utils/hsts-prune
Thanks anyway!

PR #16774, opened 2018-10-16
Update Cheshire Constabulary ruleset

tomodwyer:
Update Cheshire Constabulary ruleset to match other PoliceUK rulesets ([List of UK police forces](https://www.police.uk/forces/)).


ivysrono:
https://travis-ci.org/EFForg/https-everywhere/jobs/586541595#L324

pipboy96:
I think there is no point in this (purely cosmetic) change right now, especially since it fails tests.

PR #16775, opened 2018-10-16
Add ruleset for Dorset Police

tomodwyer:
Add ruleset for Dorset Police ([List of UK police forces](https://www.police.uk/forces/)).

### DIFF
--- a/src/chrome/content/rules/PoliceUK.xml
+++ b/src/chrome/content/rules/PoliceUK.xml
@@ -29,7 +29,6 @@
       - WestMidlandsPolice.xml
 -->
 <ruleset name="Police UK">
-	<target host="police.uk" />
 	<target host="www.police.uk" />
 
 	<rule from="^http:" to="https:" />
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/PoliceUK.xml
```diff
@@ -31,5 +32,8 @@
 	<target host="police.uk" />
 	<target host="www.police.uk" />
 
-  <rule from="^http://(www\.)?police\.uk/" to="https://www.police.uk/" />
+	<test url="http://police.uk/" />
```
zoracon: You can delete this since there is an implicit test when you add the target URL.

src/chrome/content/rules/PoliceUK.xml
```diff
@@ -31,5 +32,8 @@
...
+	<test url="http://www.police.uk/" />
```
zoracon: Same as above

src/chrome/content/rules/PoliceUK.xml
```diff
@@ -31,5 +32,8 @@
...
+
+	<rule from="^http://(www\.)?police\.uk/" to="https://www.police.uk/" />
```
zoracon: Please convert to trivial rule
```suggestion
	<rule from="^http:" to="https:" />
```

pipboy96:
@zoracon I think there should be an option for disabling "Target containing entire eTLD" check.

PR #16776, opened 2018-10-16
Add distro fallback on install-dev-dependencies.sh

andrealmeid:
This PR is inspired by the moment when the script failed to work on Arch Linux.

Little enhancements were made:

- Comments specify which distro is the command related
- Output of `type` command doesn't appears  for user (either in success or fail)
- If any of install commands doesn't work (`apt`, `dnf`, ...), it displays an warning message

Thanks!

### DIFF
--- a/install-dev-dependencies.sh
+++ b/install-dev-dependencies.sh
@@ -22,6 +22,7 @@ if [ $UID != 0 ]; then
   SUDO_SHIM=sudo
 fi
 
+# debian based installation
 if type apt-get >/dev/null ; then
   $SUDO_SHIM apt-get update
   $SUDO_SHIM apt-get install -y lsb-release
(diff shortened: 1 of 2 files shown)


install-dev-dependencies.sh
```diff
@@ -37,7 +38,7 @@ if type apt-get >/dev/null ; then
   $SUDO_SHIM apt-get install -y libxml2-dev libxml2-utils libxslt1-dev \
     python3.6-dev $BROWSERS zip sqlite3 python3-pip libcurl4-openssl-dev xvfb \
     libssl-dev git curl $CHROMEDRIVER
-  if ! type geckodriver >/dev/null; then
+  if ! type geckodriver >/dev/null 2>&1;  then
     curl -LO "https://github.com/mozilla/geckodriver/releases/download/v0.17.0/geckodriver-v0.17.0-linux64.tar.gz"
```
J0WI: For the `chromedriver` we detect the `$ARCH`. Would be great if we are consistent.

Hainish:
Thank you!

PR #16924, opened 2018-10-18
HTTPS to HTTPS 40 Rules Refactor

zoracon:
Background:
There were quite a few older rules supporting HTTPS to HTTPS redirects that we generally would like to clean and move towards a safer and more reliable service as an extension.

The Concerns:
- There were some rulesets that had HTTPS to HTTPS redirects, causing the potential shady realm of valid urls being pointed to places they potentially should not
- Some of these rulesets were originally acting as a redirect service for site devs/admins who may not have properly routed their CDNs
- Many of these rulesets come from more older conventions of the encrypted web 
- For example:
1. When site devs forgot to add www and proper redirects to the host from www urls
2. secure.example.com urls

Reasoning:
1. The main reason is to truly return to the initial nature of the plugin. And that is to safely port the user from un-encrypted to encrypted connections. 
2. Many of the original concerns around bad SSL errors have been addressed with: #16546
3. Also, there is a future for the plugin to serve in other capacities. And for that to happen as a reliable source in these plans, we need to work together to make sure the plugin does exactly as intended.
4. This cleaning also reaffirms our promise to the user that the plugin does exactly what it says

Plan
1. I will tag in and reference the original contributors for certain complex rulesets before any changes get merged in
2. As review occurs, I will keep up tests for the rulesets as any changes that may come in
3. After testing, we can move forward to merge


zoracon:
@Hainish @J0WI @Bisaloo 
Above I have explained this PR and the reasoning behind it.

 A lot of the original contributors to these rulesets are no longer on Github. So when any of you have a chance, possibly take a look at some of the refactored rulesets that had some further nuance, complicated regex, and rules; I'd greatly appreciate any feedback. I am just making sure i follow typical conventions and documentation

In particular: These files

- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-b039aaa98455f279c1b350b1eccebeac
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-268dc2564aab70688b761faca97d395b
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-abee326905ddc5ab2f7b421776f38b97
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-17be465a3d153dfee09ab4324e038196
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-d3125a87952a8ae0aea5f37d4eabe66d
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-cbcc53dfb03a5dbe46ba02b0774f2b7b
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-65f2ab106fb0c611f539482e0fc1dc0d
- https://github.com/EFForg/https-everywhere/pull/16924/files#diff-1e5c530d14291bc6176f372f318dc5d9

src/chrome/content/rules/Aeriagames.xml
```diff
@@ -1,6 +1,13 @@
 <!--
 	CDN buckets:
 
+		New
+
+		- Updated CDN url
+			- cms-content.s.aeriastatic.co
+
+		Out of date
```
Bisaloo: Hum, if that's out of date, shouldn't you simply remove it?

src/chrome/content/rules/Aeriagames.xml
```diff
@@ -10,28 +17,25 @@
...
-<target host="aeriagames.com" />
-	<target host="*.aeriastatic.com" />
 
-<rule from="^http://(?:www\.)?aeriagames\.com/" to="https://www.aeriagames.com/"/>
-
-	<rule from="^http://c\.aeriastatic\.com/"
-		to="https://c.aeriastatic.com/" />
-
-	<!--	We must rewrite from https://s to c due to protocol-relative
+		We must rewrite from https://s to c due to protocol-relative
```
Bisaloo: But we don't as far as I can tell (?). So either this comment is outdated and can be removed or an additional rule should be added to this ruleset.

src/chrome/content/rules/BME.hu_incomplete.xml
```diff
@@ -27,10 +27,6 @@
 	<target host="cs.bme.hu" />
 	<target host="www.cs.bme.hu" />
 
-	<!-- Workaround https://mit.bme.hu cert error -->
```
Bisaloo: Can you also remove the `<!-- !!! -->` next to the target? It's hard to understand why it was put there now and it might discourage other contributors to edit this file because they could feel like there is a hidden issue they don't understand.

src/chrome/content/rules/Blip.xml
```diff
@@ -68,15 +68,9 @@ Fetch error: http://a.blip.tv/ => https://as.blip.tv/: (7, 'Failed to connect to
 
   <securecookie host="^(?:.*\.)?blip\.tv$" name=".+" />
 
-
 	<rule from="^http://a\.blip\.tv/"
 		to="https://as.blip.tv/" />
 
-	<!--	s? for protocol-relative links:
-						-->
-	<rule from="^https?://\d\.i\.blip\.tv/"
```
Bisaloo: As far as I can tell, `https://d.i.blip.tv/` still times out so this ruleset will cause issues in its current state.

You could either add an exclusion for `d.i.blip.tv` and remove the wildcard for `i.blip.tv` and only list working subdomains.

src/chrome/content/rules/CafePress.xml
```diff
@@ -34,8 +34,8 @@ Fetch error: http://server.iad.liveperson.net/ => http://server.iad.liveperson.n
 		to="https://content.cafepress.com/"/>
 
 	<!--	rewriting help to liveperson breaks links under "design your own"
-		in the footer.  Fix this by rewriting back.	-->
-	<rule from="^https://server\.iad\.liveperson\.net/make/"
-		to="https://www.cafepress.com/make/"/>
```
Bisaloo: I think you can completely remove this rule and the related target because it's already handled in the (currently `default_off`) [`LivePerson.net.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/LivePerson.net.xml) ruleset. It will also fix the issue where you have a target without a related rule.

src/chrome/content/rules/Caltech.xml
```diff
@@ -150,19 +150,6 @@ Fetch error: http://web.caltech.edu/ => https://web.caltech.edu/: (6, 'Could not
 
 	<securecookie host=".\.caltech\.edu$" name=".+" />
 
-
```
Bisaloo: Don't forget to remove the related targets :wink:. Right now, they will be caught by the trivial rule at the end of this rule and cause `Invalid certificate` errors.

src/chrome/content/rules/Blip.xml
zoracon: It looks like this ruleset overall has been turned off buy the comprehensive test. Will review this ruleset for deletion.

src/chrome/content/rules/CafePress.xml
zoracon: Thank you for pointing out the related ruleset

src/chrome/content/rules/Aeriagames.xml
```diff
@@ -1,37 +1,16 @@
...
-			- https://bugzilla.mozilla.org/show_bug.cgi?id=879072#c1
-							-->
-	<rule from="^https?://s\.aeriastatic\.com/"
-		to="https://c.aeriastatic.com/" />
-
+		<test url="http://cms-content.s.aeriastatic.com/82de179009b2505fcc419f60f4d98ab7/files/aurakingdom/image/F/Featured_Shinobi.jpg" />
+		<test url="http://c.aeriastatic.com/modules/js/AG/version1/assets/signupVerifier.css" />
+		<test url="http://cms-content.s.aeriastatic.com/cb69c2e1a7010ab58379cbef75909305/files/eos/image/e/episode_2.5_EOS_featured_2.jpg" />
+	<rule from="^http:"
+		to="https:" />
```
J0WI: The generic redirect is a bit dangerous in combination with the wildcard target above.

zoracon:
> I did not yet review all files, but here are some things I noticed:
> 
>     * Is it worth to apply these updates to `default_off` rules?
> 
>     * Applying a generic redirect to a rule that contains a wildcard target can have negative side effects (breaking not yet listed subdomains).

I can definitely go back and leave the ones that were already off. I must admit, I started these changes when I didn't fully get the default off feature, but now I do. So I can condense the PR some for that. I can go back over the rulesets that had wildcards and see what can be amended there as well. Thank you for the feedback.

Issue #16950, opened 2018-10-20
Travis CI not triggered by PRs


J0WI:
Type: other

Travis CI is no longer triggered by PRs.

@Hainish could you have a look please?

J0WI:
Actually the checks run in background, but the API to expose the results is somehow broken.

Hainish:
@J0WI I'll take a look at this tomorrow.

I've noticed a lot of new issues for sites that don't yet have rulesets.  Are you finding these issues useful, or should I encourage users who are automating these issue creations to stop?

J0WI:
If the issue creation is automated it's not very helpful. Or is the bot open source, so that we can see the criteria (might be interesting to also auto generate rulesets)?
For manual created issues we know at least, that someone noticed that the site actually works over HTTPS.

J0WI:
@Hainish did you already found something about the Travis issue?

Hainish:
I've switched Travis from being a project service to GitHub app, and re-enabled the integration on the Travis CI side.  It looks like it's working now, but I think you'll have to close and re-open PRs in order to initiate the CI in existing PRs.

PR #16951, opened 2018-10-21
Use ES6 generator functions to reduce run-time overhead in search & lookup

cschanaj:
This will make `potentiallyApplicableRulesets()` an ES6 generator function and save us some run-time overhead in the searching of rulesets. As a trade-off, we can no longer cache all search results and see an extensive logging.

Remark: Travis is not triggered for this PR #16950

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -540,26 +540,42 @@ RuleSets.prototype = {
    * @param host The host to check
    * @returns {*} (empty) list
    */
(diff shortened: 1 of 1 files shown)


cschanaj:
It seems that we cannot further optimize the use of `potentiallyApplicableRulesets` without breaking change in the UI... 

https://github.com/EFForg/https-everywhere/blob/cf7fc81f50ca22cc17957a794a8b084e634514e8/chromium/background-scripts/background.js#L401-L408

PR #16956, opened 2018-10-22
Create lczero.org.xml

ivysrono:


src/chrome/content/rules/lczero.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="lczero.org">
+	<target host="lczero.org" />
+	<target host="www.lczero.org" />
+	<target host="blog.lczero.org" />
```
Bisaloo: Please add a target for `oldmain.lczero.org`

src/chrome/content/rules/lczero.org.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="lczero.org">
```
Bisaloo: Please add a comment about `play.lczero.org`

Bisaloo:
Thank you!

PR #16980, opened 2018-10-23
Update Synchronicer.dk to enforce HTTPS

kimsey0:
Synchronicer.dk no longer has any mixed content problems.


kimsey0:
@Hainish: I can't see exactly what's wrong from the Travis log. It looks like the Selenium driver just can't start Firefox. Should I do anything to handle this?

Hainish:
@kimsey0 this should be fixed in https://github.com/EFForg/https-everywhere/pull/16986 - re-running tests now.

J0WI:
Thanks!

PR #16983, opened 2018-10-24
Add ruleset for ftsafe.com

jcoscia:


J0WI:
@jcoscia would be great if you could add the remaining host.

ivysrono:
> 
> 
> See also `[abcserver.ftsafe.com](http://abcserver.ftsafe.com)`

Not exist.

PR #16991, opened 2018-10-28
Add ruleset: sa-mp.com

PixellUp:
Add ruleset: forum.sa-mp.com


PixellUp:
@J0WI `lists.sa-mp.com` & `server.sa-mp.com` are for internal use and can't be accessed via a browser. If you insist I shall include them too.

J0WI:
They run a http service and are accessible in a browser.

Anyway, the current syntax is not valid and the build is failing. Only one `<ruleset>` is allowed per file.

pipboy96:
Sorry, this can't be merged as is. I will attempt to create a corrected PR.

PR #16992, opened 2018-10-28
[familycouncil.gov.hk] New ruleset

Marco129:
New ruleset for [Hong Kong Family Council](http://www.familycouncil.gov.hk)


Marco129:
@J0WI Updated, thanks!

PR #17010, opened 2018-11-01
Early return for URLS that are already HTTPS

zoracon:
This follows up on the work in PR #16924.


cschanaj:
This will bring significant performance improvement to HTTPSE. Thanks for the PR. 

However, I am guessing that the users will no longer be able to disable certain rulesets from the UI. It might worth investigations. cc @Hainish @J0WI @Bisaloo

PR #17012, opened 2018-11-02
Forbid HTTPS rewrites in relaxng

Bisaloo:
Follow up of #15194, #16924

PR #17017, opened 2018-11-03
Update MyEdAccount.com.xml

cschanaj:
#17012


J0WI:
The domain seems to be for sale, so let's delete it.

PR #17020, opened 2018-11-03
Update Telus.xml

cschanaj:
#17012


zoracon:
I didn't realize there was duplicate work being done to cover #17012. I had merged in the clean up deleting the handful of turned off rulesets in #17043. We can restore these since you took the time to update them.

cschanaj:
done, see also #17021, #17023

PR #17022, opened 2018-11-03
Delete rules for Solus Project

Jacalz:
They had a rough time and had to temporarily switch to https://getsol.us and
their new site does not accept connections over http, it automatically defaults to
using https. It makes this rule unnecessary...


Bisaloo:
Do you have a link to an announcement with more info? 

You say "temporarily switch". If that's the case, we should probably just disable this rule for now.

Jacalz:
A link to it the announcement: https://getsol.us/2018/09/08/supercharging-solus-infrastructure/ 

The thing is that the rule was already useless before they switched over because they automatically changed over to https even if you tried using http. They have the https setup correctly so there is no reason to have a rule for it...

Jacalz:
Okay, now I see 👍 The thing is that they might have gotten the acces to the site back if xyou believe the following link:
https://www.phoronix.com/?page=news_item&px=Solus-Open-Letter

I think we should wait and see what happens, if they switch back we can simply close this 👍

PR #17025, opened 2018-11-04
Fix extension popup shows rulesets from previous pages

cschanaj:
STR:

  1. Visit http://time.com/ using a build from b056c3d7f01bd6e6921f0a4b4a2173b60f11862a from #17010 
  2. Visit https://bloomberg.com/
  3. Notice that rulesets applied to http://time.com/ shows up when the URL is changed to https://bloomberg.com/ 
  4. Expect rulesets from previous page are clear and only rulesets applied to current page are shown.

Issues 

02a9220 breaks browser test because #17010 contains a bugs. Step 3 can be repeated infinitely for different domains, as long as all the subsequent pages are HTTPS only. So this bug cannot be ignored.

  1. For non-HTTPS pages, if `main_frame` is redirected by HTTPSE, no ruleset will be shown on the popup. Otherwise, rulesets for page's resources will be shown (if any). 

  2. For HTTPS pages, no ruleset be shown unless the page serve mixedcontent 

  3. No rulesets will be shown for most of the time (at least for me), this might confuse some users that HTTPSE is not working. Perhaps there should be a brief explanation in the release notice on this matter. 

To enable `test/selenium/test_popup.py::TestPopup::test_rule_shown`, we need to perform the test against a `mixedcontent` pages ... which seem an anti-pattern for me. So I have disabled the test in 2bf5f79, I am open for discussions on this. 

cc @jsha @zoracon @Hainish This PR will fix a bug from #17010 and should be merged before the next release. thanks.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -320,6 +320,14 @@ function onBeforeRequest(details) {
     return;
   }
 
+  // Clear the content shown in the extension popup.
+  // This needed to be done before this listener returns,
+  // otherwise, the extension page might include rulesets
+  // from previous page.
+  if (details.type == "main_frame") {
+    appliedRulesets.removeTab(details.tabId);
+  }
+
   let uri = new URL(details.url);
 
   /**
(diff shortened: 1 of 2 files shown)


jsha:
Thanks for spotting this, @cschanaj! I'll take a look at this PR with @zoracon on Monday.

Hainish:
Thanks @jsha and @cschanaj.  Please flag me for review when you open the PR.

zoracon:
@Hainish FYI this is already a PR but we reviewed and tested and it looks good to go. Just going to merge this in with some urgency since PR #17010 is in fact breaking the build

PR #17043, opened 2018-11-05
Clean up HTTPS to HTTPS rulesets

zoracon:
In #16924 we fixed many rulesets that were HTTPS to HTTPS, however when we ran the RELAX NG validator, it caught a handful of rulesets that were failing regardless of them being turned off or not. This PR helps in that clean up and as well assists in helping with @Bisaloo's PR (#17012) that helped to clear out the checks in the whitelist test that checked for the HTTPS to HTTPS flag, which is not longer needed.

To be more specific, this PR mainly deleted those problematic rulesets that were default off and converted a few to trivial rulesets.


cschanaj:
why simply delete those rulesets rather than updating them? there is only a handful of rulesets and making minimal changes (if we ignore the fetch test) to pass the validations test is possible.

zoracon:
> why simply delete those rulesets rather than updating them? there is only a handful of rulesets and making minimal changes (if we ignore the fetch test) to pass the validations test is possible.

Many of the rulesets in this particular PR were already turned off, therefore not in use. I updated the ones that were to help pass the validations needed. Overall I wanted to get this up so this work could help #17012 out with the contribution they made as well. In #16924, many were updated to pass coverage since they were very much still on use. This PR mostly addresses the more problematic of that bunch.

Issue #13412, opened 2017-10-29
Only show a ruleset if at least one redirection took place.


ghost:
Type: code issue
ping @Hainish @cowlicks.

Possibly also if the redirect could have taken place if the request were made over HTTP.

jeremyn:
As I understand it, a ruleset appears in the UI if the page URL or the URLs of any secondary request match that ruleset. So this issue is already solved if we allow the condition "Possibly also if the redirect could have taken place if the request were made over HTTP." Is that right?

If we don't allow that condition, so that for example `Freerangekitten.com` would not appear in the UI if you directly visit https://freerangekitten.com , then you wouldn't be able to disable that ruleset, since it's not in the UI to toggle on or off.

koops76:
@jeremyn No, it simply checks for targets, not for actual possible redirection.

Bisaloo:
> @jeremyn No, it simply checks for targets, not for actual possible redirection.

That's what I observed as well. It was pretty obvious with the google rulesets. Not so much now that some of them have been removed.

Note that this differs from the behaviour of the legacy HTTPS-Everywhere.

cschanaj:
I think that I can confirm @koops76 and @Bisaloo 's observations. Given a `tabId`, for each request, we look for rulesets which have matching `target`. Among these `ruleset`, if it is `active` (not `default_off`), we show it as an entry of "Stable rules"; otherwise, it is regarded as "Experimental rules". So, whether or not these rulesets make a redirection is not taken into account. 

While this issue shouldn't be too difficult to fix, but I am not sure what is the expected behavior for those `default_off` rulesets? By definition, they do not make redirections until an user activate them. However, if they do not show up, it might not be possible for non-technical user to activate these rulesets. 

cc @Hainish @cowlicks 

https://github.com/EFForg/https-everywhere/blob/e958e4bf4e8f3be30a8d183a3ed305ccb72a85b1/chromium/background.js#L285-L290

jeremyn:
All: thanks for the clarification. So is the point of this issue to take into account complications like `exclusion`s and complex `rule`s when deciding when to show a ruleset in the UI?

Hainish:
Thanks for the issue, and @cschanaj's explanation is correct: we show all *potentially* applicable rulesets, whether or not they are actually triggered on a page.  This is behavior that _is_ consistent with the legacy extension, and I believe should not be altered.

[...]

Issue #12076, opened 2017-08-24
Ruleset is shown as activated even when the request is blocked


gloomy-ghost:
STR:
1. Install uMatrix and block all 3rd parties.
2. Open https://zh.moegirl.org/Mainpage
3. Check activated rulesets

Expected:
No ruleset should be shown

Actual:
Ruleset for 3rd parties resources are shown
![default](https://user-images.githubusercontent.com/13086421/29653320-6c17c7ae-88fd-11e7-8b32-fdaee0c314fa.png)

Can be reproduced on both Chrome and Firefox

jeremyn:
`https://github.com/Synzvato/decentraleyes/issues/148#issuecomment-285121088` might be interesting. It links to https://developer.chrome.com/extensions/webRequest , which under the section "Conflict Resolution" says:

> Only one extension is allowed to redirect a request or modify a header at a time. If more than one extension attempts to modify the request, the most recently installed extension wins and all others are ignored. An extension is not notified if its instruction to modify or redirect has been ignored.

Here I guess both uMatrix and HTTPS Everywhere are modifying the request -- uMatrix by blocking, HTTPS Everywhere by rewriting -- and uMatrix's action is  "winning" but HTTPS Everywhere isn't notified so it still displays the relevant rulesets.

So in Chrome at least, it may not be possible to do what you want.

numismatika:
I think this a bigger problem that should be adressed upstream at both browsers root.
If you have multiple addons interfering and listening to redirects you get a lot of wrong info.
If you have an Adblocker like UBO, HTTPS-Everywhere & SmartHTTPS, Info Addons like SixOrNot&Cert Patrol and a Privacy aware addon like Decentral Eyes, the state shown by these addons can become wrong rather fast. Not to mention irritating console errors and csp warnings.
You have blocked requests, rewritten requests and you want to also save the states of those requests before and after the interference so you can properly show info about them.

pipboy96:
This is not possible to fix since HTTPS Everywhere can't know whether the request will be blocked by the others addons you have.

Issue #9688, opened 2017-04-26
Indication when being redirected by HTTPS Everywhere


C0rn3j:
When I visit a site that has both HTTP and HTTPS enabled, I'd like to have an indication of that, as I like to contact projects that still do this and help/nag them to disable HTTP.

ATM the only way I know of to figure this out is by disabling the extension and then trying to load the site over HTTP again.

jeremyn:
HTTPS Everywhere doesn't actually know if a site supports HTTP or not. It just rewrites URLs using a regex. There's a relevant discussion in issue https://github.com/EFForg/https-everywhere/issues/8239 about some concerns with adding the ability to HTTPS Everywhere to check HTTP requests.

PR #17047, opened 2018-11-08
Show only rulesets which performed an upgrade

cschanaj:
As a side effect of #17010, the popup no longer shows rulesets which do not perform any upgrade on HTTPS pages, this PR makes it the default behavior on HTTP pages as well. It also introduce room for performance improvement like #16951.

There was concerns that users might need granular control of all the ruleset states to avoid site breakages, but #16546 is merged so I guess this is less likely an issue. 

A controversial side effect could be the fact that `default_off` rulesets will not be visible anymore. That's why I have closed #16951 before #17010 emerged. 

cc @Hainish @jsha @zoracon @Bisaloo @J0WI @jeremyn for opinions

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -406,10 +406,11 @@ function onBeforeRequest(details) {
   let newuristr = null;
 
   for (let ruleset of potentiallyApplicable) {
-    if (details.url.match(ruleset.scope)) {
-      appliedRulesets.addRulesetToTab(details.tabId, details.type, ruleset);
-      if (ruleset.active && !newuristr) {
-        newuristr = ruleset.apply(uri.href);
+    if (ruleset.active && details.url.match(ruleset.scope)) {
+      newuristr = ruleset.apply(uri.href);
+      if (newuristr !== null) {
+        appliedRulesets.addRulesetToTab(details.tabId, details.type, ruleset);
+        break;
       }
     }
   }


Bisaloo:
Related issues: https://github.com/EFForg/https-everywhere/issues/9688, https://github.com/EFForg/https-everywhere/issues/12076, https://github.com/EFForg/https-everywhere/issues/13412

Bisaloo:
> A controversial side effect could be the fact that default_off rulesets will not be visible anymore.

Currently, most `default_off` rulesets are terribly outdated. The lack of clear explanations of why a ruleset is `default_off` (#9906) also make it nearly impossible for users to know if a ruleset can safely be enabled or not (even for technical savvy users).

As far as I know, ruleset maintainers do not merge `default_off` rulesets anymore, causing [their number to steadily decline with time](https://github.com/EFForg/https-everywhere/wiki/Ruleset-Statistics) so I don't think that's a huge issue. `default_off` rulesets could have been useful and it was interesting to keep them around in the UI as a test but it's not practical enough to make sense today IMHO.

J0WI:
I'd prefer to keep the `default_off` rules in the UI, there is no other practical way to enable them.

Hainish:
If the number of `default_off` rules is declining, I don't see much of a problem in keeping them in the UI as @J0WI suggests.

Relatedly, I've had questions come up over what a `Stable Rule` is.  We might want to change this title to something else, or remove it entirely and only label the `Unstable Rules` when they are present.

Hainish:
@cschanaj:

> As a side effect of #17010, the popup no longer shows rulesets which do not perform any upgrade on HTTPS pages

It's actually worse than that, it doesn't allow you to disable rulesets even if they have performed an upgrade.  Try `http://freerangekitten.com` (our own HTTPS Everywhere test site) for instance.

Hainish:
I've just pushed a stopgap fix for #17087.  I'm going to close this PR as well, since the behavior is now consistent.

PR #17052, opened 2018-11-10
Add Evote-ch.ch.xml

J0WI:


src/chrome/content/rules/Evote-ch.ch.xml
```diff
@@ -0,0 +1,23 @@
...
+
+	f: secure connection failed
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Evote-ch.ch">
+	<target host="www.evote-ch.ch" />
```
Bisaloo:
```suggestion
	<target host="evote-ch.ch" />
	<target host="www.evote-ch.ch" />
```

Bisaloo:
Looks like this domain is now preloaded. Sorry for the extra work.

PR #17053, opened 2018-11-10
Upgrade Travis base distro to xenial

J0WI:
@Hainish could you double check for any regressions (especially the IPv6 issue #14425)?


J0WI:
Looks like there is a regression :(

PR #17057, opened 2018-11-11
Simplify and Update Save_the_Children.org.uk.xml

cschanaj:


src/chrome/content/rules/Save_the_Children.org.uk.xml
```diff
@@ -1,49 +1,24 @@
...
-		- Images, on:
-
-			- blogs from $self *
-			- shop from www *
-			- www from $self *
-
-	* Secured by us
+	Non-functional hosts
+		Couldn't connect to server:
+		- savethechildren.org.uk
```
Bisaloo: I get a time out. Is that what you mean here?

Bisaloo:
Can you also edit the filename to match the domain name please?

Otherwise, LGTM.

PR #17069, opened 2018-11-15
Add 179.ru

VoidPhantom:
There are a few domain names whose only purpose appears to be a redirect. For example, <http://wiki.179.ru/> redirects to <https://server.179.ru/wiki/>. For those domain names, HTTPS does not work; but non-root URLs like <http://wiki.179.ru/foo> simply give an error. I tried to express that using the exclusion.

### DIFF
--- a/src/chrome/content/rules/179.ru.xml
+++ b/src/chrome/content/rules/179.ru.xml
@@ -35,7 +35,7 @@
 	<exclusion
 		pattern="^http://(bak|monitor|sbory|support|t3|wiki)\.179\.ru/.+$" />
 	<rule from="^http://(bak|monitor|sbory|support|t3|wiki)\.179\.ru/$"
-		to="https://server.179.ru/wiki" />
+		to="https://server.179.ru/wiki/" />
 
 	<rule from="^http:" to="https:" />
 </ruleset>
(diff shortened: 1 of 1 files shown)


VoidPhantom:
I’m somewhat confused. Do I need to add more tests?

src/chrome/content/rules/179.ru.xml
```diff
@@ -0,0 +1,41 @@
...
+	<target host="open.179.ru" />
+	<target host="sbory.179.ru" />
+	<target host="server.179.ru" />
+	<target host="support.179.ru" />
+	<target host="t3.179.ru" />
+	<target host="wiki.179.ru" />
+
+	<securecookie host=".+" name=".+" />
+
+	<exclusion
```
ivysrono: Add 7+ test rules.

ivysrono:
@VoidPhantom To be simple, You can add another ruleset like：
```
<ruleset name="179.ru-Problematic">
	<target host="bak.179.ru" />
	<target host="monitor.179.ru" />
	<target host="sbory.179.ru" />
	<target host="support.179.ru" />
	<target host="t3.179.ru" />
	<target host="wiki.179.ru" />

	<rule from="^http://(bak|monitor|sbory|support|t3|wiki)\.179\.ru/$"
		to="https://server.179.ru/wiki/" />

</ruleset>
```

PR #17090, opened 2018-11-27
Ruleset for vescudero.net domain added

vescudero:


J0WI:
Since your domain is already preloaded in all browsers we support, there is no need to add a ruleset for it.
Thank you anyway!

PR #17100, opened 2018-11-30
Ruleset for Demo

zoracon:
This is for  a deno, please don't merge

### DIFF
--- a/src/chrome/content/rules/Example.com.xml
+++ b/src/chrome/content/rules/Example.com.xml
@@ -1,0 +1,15 @@
+<!--
+	Nonfunctional hosts in *.example.com:
+
+	h: http redirect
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Example.com.xml
```diff
@@ -0,0 +1,15 @@
...
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Example.com">
+	<target host="example.com" />
+	<target host="www.example.com" />
```
andresbase: Thank you for your contribution

zoracon: :shipit:

PR #17102, opened 2018-12-01
Run hsts-prune and fix impacted rulesets

Bisaloo:


src/chrome/content/rules/Bitly_branded_short_domains.xml
```diff
@@ -10967,7 +10967,6 @@ This file generated by Bitly on Fri, 10 Nov 2017 17:24:38 UTC
 	<target host="go.bcoach.info" />
 	<target host="go.beardben.com" />
 	<target host="go.beckatl.com" />
-	<target host="go.beeming.net" />
```
J0WI: This target is not yet preloaded.

src/chrome/content/rules/Despora.de.xml
```diff
@@ -1,10 +0,0 @@
-<ruleset name="Despora.de">
-
-	<target host="despora.de" />
-	<target host="www.despora.de" />
```
J0WI: This target is not yet preloaded (only `^` is).

src/chrome/content/rules/G5.gov.xml
```diff
@@ -1,8 +0,0 @@
-<ruleset name="G5 - US Department of Education">
```
J0WI: This is not yet preloaded.

src/chrome/content/rules/Bitly_branded_short_domains.xml
Bisaloo: Hum, that's weird, I use `hsts-prune` to remove those targets. I didn't remove them myself

Bisaloo: Where is it missing?

J0WI: https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsSTSPreloadList.inc

J0WI: Ah, got it!

Bisaloo: As far as I can tell, they only list domains. They don't enumerate all subdomains. Otherwise, there would be much entries starting with `www`.

src/chrome/content/rules/Despora.de.xml
Bisaloo: This is because only the main domain is listed, right? Like the others.

J0WI: The host is not preloaded with the `includeSubDomains` flag in Firefox. Also, the server uses an invalid HSTS configuration: https://hstspreload.org/?domain=despora.de
So this host should be removed at all (#13682).

Bisaloo: Okay, it's not surprising the script got confused here. I opened an issue to fix this issue directly with the website: https://github.com/despora/diaspora/issues/12.

J0WI: @Bisaloo could you revert the removal of `Despora.de.xml` for now that we can ship this in #17157?

Bisaloo: @J0WI, done

PR #17111, opened 2018-12-02
Create Multinationales.org.xml

Bisaloo:


Bisaloo:
Thank you!

btw @J0WI, I think you changed your settings to "Merge with merge commit" instead of "rebase and squash".

J0WI:
@Hainish this just happened again. Has the merge policy in the repo setting been changed?

Hainish:
@J0WI In my experience whatever merge method you chose last is set as the default the next time you go to merge.

J0WI:
That was the case for me too until a few days (?) ago. Now it's always set to merge after logout.
The merge button can be configured in the repo settings (does not affect cli AFAIK). Should we force squash & merge?

Hainish:
It can be configured to enable/disable certain merging methods, but not to prioritize merging methods.  I still use merge commits for releases, but could conceivably just do this on the command line.

Do you see any other problems with just disabling the other merge methods completely?

J0WI:
Something has definitively been changed how GitHub is memorizing merge strategies.
If I look at the recent commit it sometimes worked after selecting it manually, but most of the time it didn't memorize my preference.

PR #17112, opened 2018-12-02
Create SocalistWorker.org.xml

Bisaloo:


src/chrome/content/rules/SocialistWorker.org.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	www.socialist.org 404 on deep links
```
J0WI: I assume this should be `www.socialistworker.org`.
However, I think this is expected. Plain HTTP does also throw a 404.
And there is a wildcard DNS.

Bisaloo: Oh, I still need to add a comment about the wildcard DNS. I will do it later today!

PR #17121, opened 2018-12-04
Update ruleset: Consumer Reports

injust:


injust:
Blocked on #17122 (for dangling reference removal)

injust:
Blocked on #17122 (for removal of dangling reference to `RightNow-clients.xml`)

PR #17123, opened 2018-12-04
add bootstrap.saltstack.com and repo.saltstack.com

bryceml:


src/chrome/content/rules/SaltStack.com.xml
```diff
@@ -11,9 +11,13 @@
 -->
 <ruleset name="SaltStack.com">
 	<target host="saltstack.com" />
+	<target host="bootstrap.saltstack.com" />
 	<target host="docs.saltstack.com" />
+	<target host="repo.saltstack.com" />
 	<target host="www.saltstack.com" />
```
injust: Please sort the `www` subdomain below the root domain, as defined [here](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#target-ordering).

src/chrome/content/rules/SaltStack.com.xml
```diff
@@ -11,9 +11,13 @@
...
 
+	<rule from="^http://saltstack\.com/"
```
injust: This doesn't appear to be necessary, as HTTPS appears to function correctly on `saltstack.com`.
An explicitly defined redirect rule is usually used when the HTTP version redirects, but the HTTPS version has an invalid certificate or some other error.

bryceml: I work at saltstack and we are trying to move to an s3 bucket to do the redirection which doesn't have https on saltstack.com.  I was fixing this in advance.

injust: Oops, didn't expect that to be the case. Sorry!

bryceml:
technical issues, dropping the first commit.

src/chrome/content/rules/SaltStack.com.xml
```diff
@@ -11,8 +11,10 @@
 -->
```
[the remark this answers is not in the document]
bryceml: ssc.saltstack.com has some older ssl settings and didn't work for me so I left it out.  Let me know what you think of https://github.com/EFForg/https-everywhere/commit/265b93f078f9772acf5f8278147966061c25d6ff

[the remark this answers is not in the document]
bryceml: I got it using TLS 1.2 so we're good now.

PR #17125, opened 2018-12-05
Update ruleset: RaceForward.org

injust:


src/chrome/content/rules/RaceForward.org.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="RaceForward.org">
+	<target host="raceforward.org" />
+	<target host="www.raceforward.org" />
+	<target host="assets.raceforward.org" />
```
Bisaloo: Do you have a test url for this?

injust: No, and I have absolutely no recollection of how I found the subdomain, especially because I wasn't using Sublist3r.

injust: Think I might have found it because it was in the SNI for one of the certificates.

Bisaloo: Okay, it's not a big deal. Let's keep it this way.

PR #17129, opened 2018-12-05
Add ruleset for samba.org

Jochen-A-Fuerbacher:


J0WI:
This host is already preloaded. Thanks anyway!

PR #17134, opened 2018-12-07
Create ruleset: PlantOGram.com

injust:


Bisaloo:
To be honest, I'm always a bit uneasy merging this kind of ruleset because I'm afraid it might break the check out process. What do you think? Do you believe that's an unfounded fear?

J0WI:
There is a server-side HTTPS redirect. So I'd not expect any issues.

Issue #16394, opened 2018-08-26
Warning page do not work on Chrome Incognito mode


cschanaj:
Type: code issue

## Platform

Browser: Google Chrome Version 68.0.3440.106 (Official Build) (64-bit)
Platform: Debian 9

## Step to reproduce

1. Start Google Chrome in Incognito mode and check "Block all unencrypted requests"
2. Visit http://time.com/ and expect the site to be blocked due to redirect loop and the warning page implemented in #16053 shown up
3. Observe Google Chrome display its own warning page

![screenshot_2018-08-26_10-39-04](https://user-images.githubusercontent.com/23636124/44624278-54b31100-a91c-11e8-8744-7497c5dab8a0.png)

Giltyhub:
I can reproduce it as well.

bcyphers:
It looks like this happens with any chrome-extension:// URL you try to open in incognito mode. I just tried accessing some Privacy Badger resources and had the same issue. The logic is working correctly, but we need to find a way to work around this.

cschanaj:
IMHO, redirect the page to an external resource on eff.org seem promising. First, it avoid breaking the code. It will also allow EFF to provide more detailed information to the extension users. Since this might have privacy implication, there should be an option for user to opt-out the redirect.

P.S. According to the docs, `"incognito": "split"` is not supported by Firefox (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito)

cc @jeremyn @J0WI @gloomy-ghost @Bisaloo opinions?

Bisaloo:
I don't really like the idea of redirecting to an external resource (such as `eff.org`). As you said, it has many important privacy implications, such as referers which may information about visited pages.

A fourth option would be to keep things as they are right now. It's not optimal but the default page shown in the OP screenshot is still quite explicit and should work well enough.

J0WI:
I'm not so happy with redirecting to another site. Isn't there any other solution? How does e.g. uBlock Origin handle this case?

cschanaj:
It seem that uBlock Origin has no problem with redirecting the users to the warning pages, tested with http://com.com/. If we can apply the same solution uBO used, I agree that it should be preferred.

bcyphers:
@cschanaj uBlock uses `"incognito": "split"` in Chrome (https://github.com/gorhill/uBlock/blob/master/platform/chromium/manifest.json). This bug doesn't appear to affect Firefox.

[...]

PR #17135, opened 2018-12-07
extension.getURL is deprecated, use runtime.getURL instead

Hainish:


Hainish:
Thought this might solve https://github.com/EFForg/https-everywhere/issues/16394, but unfortunately it doesn't.  Still needed, thoguh.

Issue #16915, opened 2018-10-18
Add new ruleset for uab.edu


TotalCaesar659:
Type: new ruleset
Domain: http://uab.edu/

PR #17144, opened 2018-12-13
Add ruleset: UAB.edu

dnoetzel:
Fixes issue #16915.

### DIFF
--- a/src/chrome/content/rules/UAB.edu.xml
+++ b/src/chrome/content/rules/UAB.edu.xml
@@ -1,0 +1,13 @@
+<!--
+	Nonfunctional hosts in *.uab.edu:
+		- mail.uab.edu
+		- web.uab.edu
+-->
+<ruleset name="UAB.edu">
+	<target host="uab.edu" />
+	<target host="www.uab.edu" />
+	<target host="go.uab.edu" />
+	<target host="sites.uab.edu" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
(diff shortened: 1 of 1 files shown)


J0WI:
It would be great if you could add as many targets as possible: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains

src/chrome/content/rules/UAB.edu.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Nonfunctional hosts in *.uab.edu:
+		- mail.uab.edu
+		- web.uab.edu
+-->
+<ruleset name="UAB.edu">
```
J0WI: Please rename to `UAB.edu (partial)`

dnoetzel:
@J0WI 
> It would be great if you could add as many targets as possible: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains

Thanks for pointing me in the right direction. `Sublist3r` reports 1,467 domains associated with `uab.edu`, 110 of which were reachable only via HTTP and 269 via HTTPS, but  of those 269 subdomains, 98 of them now just redirect to some location on `www.uab.edu` (e.g. `http://sa.psychiatry.uab.edu` ultimately redirects to `https://www.uab.edu/medicine/substanceabuse/`). While I'm sure that wildcards could cut down on the number of entries somewhat, listing out hundreds of targets seems like it would bloat the ruleset. How would you recommend I go about listing these targets?

J0WI:
Thanks!

PR #17148, opened 2018-12-15
Create ruleset: dmm.com/dmm.co.jp

saschanaz:
Warning: *.dmm.co.jp includes NSFW contents so take care when reviewing.


saschanaz:
Not sure why dlsoft fails, maybe a regional problem?

src/chrome/content/rules/Dmm.co.jp.xml
```diff
@@ -0,0 +1,13 @@
+<ruleset name="Dmm.co.jp">
+	<target host="dmm.co.jp" />
+	<target host="www.dmm.co.jp" />
+	<target host="avatar-contents.dmm.co.jp" />
+	<target host="display.dmm.co.jp" />
+	<target host="games.dmm.co.jp" />
+	<target host="pics.dmm.co.jp" />
+	<target host="p.dmm.co.jp" />
+
+	<rule from="^http://dmm\.co\.jp/"
```
zoracon: Have you found cases where this rule was needed on top of the trivial rule below? They seem to have a directive already for this domain to redirect to the "www" iteration.

src/chrome/content/rules/Dmm.com.xml
```diff
@@ -0,0 +1,15 @@
...
+	<target host="www.dmm.com" />
+	<target host="avatar-contents.dmm.com" />
+	<target host="img-freegames.dmm.com" />
+	<target host="p.dmm.com" />
+	<target host="pics.dmm.com" />
+	<target host="rcv.ixd.dmm.com" />
+	<target host="stat.i3.dmm.com" />
+	<target host="trac.i3.dmm.com" />
+
+	<rule from="^http://dmm\.com/"
```
zoracon: Same comment as above

src/chrome/content/rules/Dmm.co.jp.xml
saschanaz: @zoracon Are you sure that https://dmm.co.jp/ and  https://dmm.com/ are accessible there? I can't access them. Please make sure that your browser is not automatically adding `www`.

PR #17152, opened 2018-12-18
Added limeknight.com

555:


Bisaloo:
Hi, thank you for your contribution!

The tests are failing because the certificate is only valid for `www.limeknight.com`, not `limeknight.com`. I see you're the website owner, could you fix this on your side please?

Also, for website owners, we recommend [HSTS preloading](https://scotthelme.co.uk/hsts-preloading/) your website. This offers the same level of protection as HTTPS Everywhere and it benefits to all visitors, not just HTTPS Everywhere users.

Please let me know if you have any question!

pipboy96:
As @Bisaloo said, please use HSTS instead.

Issue #16985, opened 2018-10-25
Rebrand "Block all unencrypted requests" feature


Hainish:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: other

"Block all unencrypted requests" is a phrase which is overly verbose and hard to reference.  We keep on using the BAUR initialism when we don't want to type it out.

I suspect this feature will be used more and more as the web gets better and better HTTPS coverage.  To make this feature easily referenceable and neater, I propose we rebrand it to something a bit nicer.

In the past we've used the phrase "HTTP Nowhere Mode", but this isn't properly descriptive since this mode blocks FTP connections now as well.

We will do some user testing on what makes the most intuitive sense for this feature, but I'd like to throw a few options out and see if they float or swim.

1. Secure-Only Mode
2. Encrypted-Only Mode
3. Turbo-Secure Mode
4. Security Ensured Mode
5. Encryption Ensured Mode

Before coming up with other options, I was leaning towards (1) as a personal preference.  But listing out these options, I actually like (5) the best.  Next to any of these options in the drop-down, I'd like some help bubble text saying "This blocks all unencrypted requests" or something that better explains what this does.

cc @zoracon @J0WI @Bisaloo @cschanaj

jsha:
Part of my mental model for this: I would like to make it easy for people to contact website owners and say "Hi, your website doesn't work in HTTPS Everywhere's <Secure Only Mode>, please fix that."

So it should be fairly short, sound very positive so the site owner is like "ooh yeah I should support that," and be reasonably self explanatory (though it will still need some explanation)

Bisaloo:
I haven't given too much thought about it but what about "HTTPS only"? Although it may not be 100% technically accurate, that's probably the most intelligible option for non-technical users.

I would be wary of phrases with "secure" because it may give the false (and already way too common) impression that the other party is trustworthy.

For some reason, I like (1) and (2). I don't know why but "XXX-only" options sound better to me. Maybe because it's easier to understand even for non-native speakers (?)

J0WI:
I just checked and Firefox uses "secure" for HTTPS (if click on the padlock), but for me security is not only HTTPS.

Hainish:
[...]

PR #17156, opened 2018-12-18
Change HTTP Block language

zoracon:
- For better user clarity for when they request un-encrypted requests
- In reference to the conversation had in #16985
- New UX is under development from #16669, but this can be a quick fix for the release tomorrow

### DIFF
--- a/src/chrome/locale/en/https-everywhere.dtd
+++ b/src/chrome/locale/en/https-everywhere.dtd
@@ -8,7 +8,7 @@
 <!ENTITY https-everywhere.menu.donate_eff_imperative "Donate to EFF">
 <!ENTITY https-everywhere.menu.observatory "SSL Observatory Preferences">
 <!ENTITY https-everywhere.menu.globalEnable "Enable HTTPS Everywhere">
-<!ENTITY https-everywhere.menu.blockUnencryptedRequests "Block all unencrypted requests">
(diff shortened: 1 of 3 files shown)


Hainish:
This will make it so that previous translations of "Block all unecrypted requests" will now be the translation for "Encrypt All Sites Eligible."  Is this the desired behavior?

We may want to change the localization string variable name too, since this will clear the existing translations for this variable.  Then again, since doing a release tomorrow, that will leave very little time and probably all non-English languages will have this string untranslated for the next release.  Maybe immediately after release, we can switch the underlying variable name.  That'll give us a lot of runway for that to be translated properly.

src/chrome/locale/en/https-everywhere.dtd
```diff
@@ -8,7 +8,7 @@
 <!ENTITY https-everywhere.menu.donate_eff_imperative "Donate to EFF">
 <!ENTITY https-everywhere.menu.observatory "SSL Observatory Preferences">
 <!ENTITY https-everywhere.menu.globalEnable "Enable HTTPS Everywhere">
-<!ENTITY https-everywhere.menu.blockUnencryptedRequests "Block all unencrypted requests">
+<!ENTITY https-everywhere.menu.blockUnencryptedRequests "Encrypt All Sites Eligible">
```
Hainish: I might also add " (EASE)" to this, to make it easy (no pun intended) to refer to.

PR #17160, opened 2018-12-20
Update Travis to latest LTS, ensure ipv6 works for fetch test (Fixes #17141)

### LINKED ISSUE #17141: Upgrade Travis base distro
We use quite an old base in Travis for our tests. Trusty will be EOL in 2019, but we have several issues (mainly IPv6) that block us from a simple upgrade. I failed to upgrade in #17053.

@Hainish, since you're already involved in the testing infrastructure, it would be great if you could schedule this for 2019 or [documenting](https://github.com/EFForg/https-everywhere/issues/7257) the necessary steps. You mentioned somewhere that you considered switching to another CI service. Or is it possible to use different bases for fetch tests than for everything else?

If we have to do major changes anyway we should also consider to cleanup our [backlog](https://github.com/EFForg/https-everywhere/labels/ruleset-testing), especially #14937

cc @zoracon @Bisaloo @cschanaj @gloomy-ghost

Type: other

Hainish:


Hainish:
In https://travis-ci.org/EFForg/https-everywhere/builds/470733182?utm_source=github_status&utm_medium=notification, I test this with a fetch test, which successfully tests `ipv6.haproxy.org`

PR #17166, opened 2018-12-21
Re-activate and Update Cisco.xml

cschanaj:


src/chrome/content/rules/Cisco.xml
```diff
@@ -128,76 +120,9 @@
...
 	<target host="tools3.cisco.com" />
 	<target host="www.cisco.com" />
-	<target host="www1.cisco.com" />
-	<target host="www2.cisco.com" />
-	<target host="www3.cisco.com" />
-
-	<!--	Complications:
-				-->
-	<target host="home.cisco.com" />
-	<target host="homesupport.cisco.com" />
```
J0WI: Please also remove the comments for them.

PR #17170, opened 2018-12-21
Update PlayStation.xml

cschanaj:
`us` works while `www.us` does not (note that `www.us` behave differently on different path, I simply removed the complicated rules) 

```bash
$ curl www.us.playstation.com/
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="http://www.playstation.com/en-us/home/">here</a>.</p>
</body></html>

$ curl www.us.playstation.com/test-path/
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="http://us.playstation.com/test-path/">here</a>.</p>
</body></html>

```

Also added more subdomains.

### DIFF
--- a/src/chrome/content/rules/PlayStation.xml
+++ b/src/chrome/content/rules/PlayStation.xml
@@ -27,9 +27,7 @@
 
 		- ^ (mismatched)
 		- metrics.aem ᵐ
-		- static.blog ᴬ
 		- uk ᵐ
-		- us ᴬ
 		- cdn.us ³
 		- service1.us ᵉ
 		- www.us ᵐ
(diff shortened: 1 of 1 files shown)


cschanaj:
@J0WI this is intended to be a follow up PR of #17168, would you be interested in reviewing this as well?

J0WI:
Thanks!

PR #17175, opened 2018-12-21
Add Nic.ir.xml

J0WI:


pipboy96:
Website has some mixed content issues, so default_off is fine for now I guess.

src/chrome/content/rules/Nic.ir.xml
```diff
@@ -1,18 +1,11 @@
...
-	m: certificate mismatch
-	r: connection refused
-	s: self-signed certificate
-	t: timeout on https
+	Expired:
+		- tools
+		- whois
 -->
-<ruleset name="Nic.ir">
+<ruleset name="Nic.ir" default_off="cert-chain">
```
J0WI: @pipboy96 this does only affect `rrmt.nic.ir`.

pipboy96: There's other breakage, not sure if splitting this into two rulesets and making one default_off and whitelisting the other is worth it.

pipboy96: Feel free to make a follow-up PR if you want to. I will merge it but I don't want to make one myself.

J0WI: #17634

PR #17176, opened 2018-12-22
Re-activate and Update 7-Eleven.com.xml

cschanaj:


src/chrome/content/rules/7-Eleven.com.xml
```diff
@@ -1,84 +1,22 @@
...
-		<test url="http://sites.7-eleven.com/franchise/home" />
-
-
-	<!--	Not secured by server:
-					-->
-	<securecookie host="^\.7-eleven\.com$" name="^__qca$" />
-
-
-	<rule from="^http://7-eleven\.com/"
-		to="https://www.7-eleven.com/" />
```
J0WI: Shouldn't we keep this?

cschanaj: I am more inclined to remove this rule since `^` gives error code 400 and does not work over http(s).

PR #17181, opened 2018-12-23
 Re-activate and Update WP.pl.xml 

cschanaj:


src/chrome/content/rules/WP.pl.xml
```diff
@@ -26,7 +20,6 @@ Fetch error: http://banki.wp.pl/ => https://banki.wp.pl/: (7, 'Failed to connect
 
 	Nonfunctional hosts in *wp.pl:
```
J0WI: Most of them seems to work now.

PR #17182, opened 2018-12-23
Re-activate and Update Metro.co.uk.xml

cschanaj:


cschanaj:
@J0WI just a reminder to squash and merge. Anyway, thank you!

PR #17195, opened 2018-12-27
remove dena.com

xruins:
For HTTP access, dena.com and its subdomains now return 301 Moved Permanently and redirect HTTPS sites.


Bisaloo:
Hi, thank you for your PR!

301 redirects to HTTPS are less secure than the protection HTTPS Everywhere is providing: https://www.eff.org/deeplinks/2018/12/how-https-everywhere-keeps-protecting-users-increasingly-encrypted-web. We should therefore keep that rule.

Are you by any chance affiliated with that website?

Issue #17192, opened 2018-12-26
HTTP nowhere mode hijacks legit Firefox warning pages.


M83tUt3:
Type: code issue

Using the "block all unencrypted requests" feature on Firefox, I often see that the addon redirects me to the "HTTPS Everywhere noticed you were navigating to a non-HTTPS page, and tried to send you to the HTTPS version instead..." message on pages that would've otherwise displayed certain Firefox warnings.
This happens for example:
1) On any non existing site. Put some bogus in the address bar with HTTP nowhere enabled and instead of FF showing the "Server Not Found" page, the addon redirects you.
2) On sites using self signed certificates, where Firefox would usually warn you and give you the option to (temporarily) accept the certificate.

I imagine it happens on any occasion where Firefox would display such a warning page.

The problem with this behaviour is that when HTTPS everywhere redirects me, I have no clue if the site indeed does not support HTTPS, or if it's just non-existant, uses a self-signed certificate, or has any other issue that would cause Firefox to warn me before actually loading the site.

Bisaloo:
I agree that (if possible), HTTPS Everywhere should not display an error on non-existing pages. This qualifies as a bug IMO.

For the rest, see #8239 for background info on this decision.

M83tUt3:
I see, you've had plenty discussion about it already. This is probably not the place to start a new one but I'll have to say, as a BAUR user I was well aware that the option could cause failures to load a page. I agree that the addon displaying an error when the site has no support for HTTPS is a good thing, but by also erroring out on pages using self signed certs for example, it does more harm than good imo.
If it could redirect only when HTTPS is simply not available rather then there being some other issue, that would be perfect of course.

zoracon:
Thanks for filing this. A good work around for this is taking a second look at the kind of messages we filter for when we receive requests with common SSL misconfigs. I will look into this one ASAP

zoracon:
@M83tUt3 As a temp solution as I get some fixes up in a PR, you can also disable HTTPSE for that URL, then navigating to the HTTPS version to add the cert.

M83tUt3:
Great that work's being done to fix this! I usually just temporarily disable EASE and reload the page to see what's up, but indeed, whitelisting would work too.

scarf:
I noticed this happens with Firefox 60.4.0esr with BAUR enabled.  Sometimes I can push the 'back' button and the normal Firefox warning page appears, allowing me to add an exception for the site and accept the invalid certificate.
[...]

Issue #17157, opened 2018-12-18
Release 2019.1.7 forthcoming


zoracon:
Type: other

cc @J0WI, @Bisaloo, @Hainish, @jsha, @andresbase

Rather short notice, however this release will be somewhat minor but please put comment any extra needed checks, call outs, or notes that may be needed.

Release Notes (incomplete):
* Change "Block all unencrypted requests" language to "Encrypt all sites eligible"
* Amend `check_rules.py` fetch test to disable rules only if _all_ rules are problematic,
and comment rules out if other rules are functional in the set
* HSTS Prune and updates

cschanaj:
It seems that the PR labeller is crashed, #17197 and #17199 should be marked as top-1k.

Bisaloo:
Sorry, I guess that's me again. It happens every time I submit a HSTS-prune PR

zoracon:
Due to some scheduling conflicts we are pushing the release to Monday since a Friday release is not ideal for obvious reasons. Bear with us!

J0WI:
#16595 fixes #15862, would be great to get some feedback there. But this could also be served in a ruleset update and does not block a new release.

Bisaloo:
If the release happens today, the date should be 201**9**.1.**7** :wink:

Bisaloo:
This update reverted my EASE setting. That could be very problematic in some threat models. I could swear there was an issue for this but I can't find it anymore. Anyone remembers this?

zoracon:
Looks like an oversight on my part led to the "EASE" language not making it into this release. Merging #17156 into master and scheduling another release for this month soon. 

> This update reverted my EASE setting. That could be very problematic in some threat models. I could swear there was an issue for this but I can't find it anymore. Anyone remembers this?
@Bisaloo  Do you think the issue was resolved and returned? Or was this an issue that was closed? I'll search around for it in the meantime.

Bisaloo:
I'm not sure. I don't know if it was open of closed and I'm even starting to wonder if I'm not mixing things up with similar but different issues :confused: . I'll keep looking.

Issue #16418, opened 2018-08-27
New "HTTPS unavailable" interstitial page misses some cases


jsha:
Type: code issue

I've got the latest HTTPS Everywhere, and tested http://http.badssl.com/ with "Block All HTTP Requests" enabled; the interstitial works great!

However, when I visit http://brokenballots.com, I get the browser's "Connection timed out" error.
Disabling "Block all HTTP Requests" allows me to visit the site (which is a redirect).

Also, visiting http://timbergroves.com/bb/ gives ERR_SSL_PROTOCOL_ERROR instead of the interstitial.

bcyphers:
Related to #16394?

cschanaj:
AFAIK, this is not related to #16394. These errors happen when the requests are made and should be handled in `onErrorOccurred(details)`.

https://github.com/EFForg/https-everywhere/blob/ca0db7bd0b975b0d7f1933508cbb85ea02fcf408/chromium/background-scripts/background.js#L586-L590

I tried to fix this with `chrome.tab.update` to force redirect a tab to the interstitial page, but it could not distinguish errors which are not caused by HTTPSE, e.g. `Could not resolve host`. So I guess a more sophisticated mechanism is needed to fix this.

bcyphers:
Yeah, this is tricky. It'll be hard to know whether any given error was the result of HTTPS Everywhere's rewrite instead of a regular host problem. Should we try listing out the errors that are probably a result of HTTPSE?

cschanaj:
> string. The error description. This string is an internal error string, may vary from one browser to another, and is not guaranteed to stay the same between releases.

This might require some maintenance effort, since the strings might change from time to time according to [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onErrorOccurred). However, I think this is a good workaround for SSL errors which hopefully the messages will not change frequently.

Hainish:
In Chrome I can enumerate a class of errors that are likely due to HTTPS misconfigurations:

```js
details.error.indexOf("net::ERR_SSL_") == 0 ||                               
details.error.indexOf("net::ERR_CERT_") == 0 ||                              
details.error.indexOf("net::ERR_CONNECTION_") == 0
```

For Firefox, this is very tricky.  I received these three errors in `details.error` for three different HTTPS bad configurations:

`NS_ERROR_CONNECTION_REFUSED`
[...]

PR #17201, opened 2018-12-29
Add option to opt-out EASE mode warning page

cschanaj:
@Hainish @zoracon This PR add an advanced option for users to opt-out EASE mode warning page in order to workaround with #17192. 

Any hope for merging before #17157? 

---

For future reference, this PR also disabled the selenium test for the `#showCounter` sanity check. This is because adding new options slow down the asynchronous `chrome.storage.get`. This resulted in failing selenium test. 

In 9612b19, I tried to avoid the bulk get operations and 3/4 selenium tests passed. However, standalone callback for each option introduce unnecessary code complexity. 

In 84b6910, I tried to increase the timeout to 10 seconds for each reload. This slows down the selenium test terribly and 4/4 tests failed. 

In 21fe45b and c225836, I disabled the selenium test for the `#showCounter` sanity check and added comments in the source code. Travis passed.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -37,9 +37,10 @@ async function initializeAllRules() {
  *    isExtensionEnabled: Boolean
  *  }
  */
-var httpNowhereOn = false;
-var showCounter = true;
-var isExtensionEnabled = true;
+let httpNowhereOn = false;
+let showCounter = true;
+let showEaseModeWarningPage = true;
+let isExtensionEnabled = true;
 let disabledList = new Set();
 
 function initializeStoredGlobals(){
@@ -49,10 +50,12 @@ function initializeStoredGlobals(){
       showCounter: true,
       globalEnabled: true,
       enableMixedRulesets: false,
+      showEaseModeWarningPage: true,
       disabledList: [],
     }, function(item) {
       httpNowhereOn = item.httpNowhere;
       showCounter = item.showCounter;
+      showEaseModeWarningPage = item.showEaseModeWarningPage;
       isExtensionEnabled = item.globalEnabled;
       for (let disabledSite of item.disabledList) {
         disabledList.add(disabledSite);
@@ -105,6 +108,10 @@ chrome.storage.onChanged.addListener(async function(changes, areaName) {
       rules.settings.enableMixedRulesets = changes.enableMixedRulesets.newValue;
       initializeAllRules();
     }
+    if ('showEaseModeWarningPage' in changes) {
+      showEaseModeWarningPage = changes.showEaseModeWarningPage.newValue;
+    }
+
     if ('debugging_rulesets' in changes) {
       initializeAllRules();
     }
@@ -302,7 +309,11 @@ let simpleHTTPNowhereRedirect = new Map();
 const cancelUrl = chrome.runtime.getURL("/pages/cancel/index.html");
 
 function redirectOnCancel(shouldCancel, originURL){
-  return shouldCancel ? {redirectUrl: newCancelUrl(originURL)} : {cancel: false};
+  if (shouldCancel) {
+    return showEaseModeWarningPage ? {redirectUrl: newCancelUrl(originURL)} : {cancel: true};
+  } else {
+    return {cancel: false};
+  }
 }
 
 function newCancelUrl(originURL){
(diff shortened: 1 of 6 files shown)


zoracon:
@cschanaj Sorry for missing this discussion and contribution. We plan on doing another release this month so I am hoping to get more tweaks in. 

Overall if the user decides to opt-out of the page it sort of defeats the purpose of BAUR (soon to be EASE) being enabled. They can choose to uncheck BAUR if this mode does not fit their needs. But issue #17192 brings up a great point about self-signed certs and sites that don't exist. I'll get to work on that ASAP.

However the Selenium change you made is a good improvement. We been having issues since that test is a bit flaky. So thank you for that!

Bisaloo:
> Overall if the user decides to opt-out of the page it sort of defeats the purpose of BAUR (soon to be EASE) being enabled.

I'm not sure I understand what you mean. For a very long time, there was no interstitial error page and users would get the default browser error. EASE would nonetheless still force all connections to HTTPS.

cschanaj:
This PR was intended to allow the users to opt-out the warning page without leaving the EASE mode, this prevents certain privacy/ security risks disabling EASE mode. Since #17192 looks like an advanced-users-only problem, I have put it in the advanced option tab. 

In #17192, it was mentioned that identifying the errors in Firefox is tricky and subjected to upstream changes. I would think an option to opt-out the warning page is a good thing to have.

zoracon:
> > Overall if the user decides to opt-out of the page it sort of defeats the purpose of BAUR (soon to be EASE) being enabled.
> 
> I'm not sure I understand what you mean. For a very long time, there was no interstitial error page and users would get the default browser error. EASE would nonetheless still force all connections to HTTPS.

Yes, they would get the default browser error in EASE mode, but would not have access to add exempt http links. The interstitial page is not only a an alternative to the browser default page, but a place to venture forth and add to the "exempt links" list we implemented when EASE came about.

zoracon:
> This PR was intended to allow the users to opt-out the warning page without leaving the EASE mode, this prevents certain privacy/ security risks disabling EASE mode. Since #17192 looks like an advanced-users-only problem, I have put it in the advanced option tab.
> 
> In #16418, it was mentioned that identifying the errors in Firefox is tricky and subjected to upstream changes. I would think an option to opt-out the warning page is a good thing to have.

If advanced users get the option to opt-out of the interstitial, do you also conclude that they would not want access to the "Sites Disabled" list? We not only provide an alternative interstitial, but this option as well with that page.

zoracon:
After some discussion with bringing up these points I much rather focus on the bugs in the case of advanced users and improve the interstitial experience. Providing an opt out would ultimately not help us see the potential bugs advanced users come across. The interstitial appears not only as a warning but a "one stop" for adding exemptions. 

As we incorporate new UX in the near future, maybe we can add the network code in a "read more" for advanced users.

cschanaj:
I'd consider adding a domain to the "Site Disable" the last resort when dealing with page breakage, it is arguably less secure than the default mode if the disabled domain serves mostly upgrade-able content. Otherwise, quitting EASE mode is a more promising resolution for average users.

However, if advanced users are not allowed to leave the EASE mode. The interstitial breaks the "Add a rule for this site" feature rendering it impossible for advanced users to add the "user rule" from the extension popup. 

I would expect when an advance found that `^` is unreachable over HTTPS, they will try to prepend `www` to the URL and see if it works. If `www` works, visit `^` again and add an trivial snapping rule, so that the breakage will not show up for a while. This can be done in completely and conveniently in EASE mode without type the whole URL again in the address bar. 

If this PR is not needed, I am ok closing it. thanks.

zoracon:
> I'd consider adding a domain to the "Site Disable" the last resort when dealing with page breakage, it is arguably less secure than the default mode if the disabled domain serves mostly upgrade-able content. Otherwise, quitting EASE mode is a more promising resolution for average users.
> 
> However, if advanced users are not allowed to leave the EASE mode. The interstitial breaks the "Add a rule for this site" feature rendering it impossible for advanced users to add the "user rule" from the extension popup.
> 
> I would expect when an advance found that `^` is unreachable over HTTPS, they will try to prepend `www` to the URL and see if it works. If `www` works, visit `^` again and add an trivial snapping rule, so that the breakage will not show up for a while. This can be done in completely and conveniently in EASE mode without type the whole URL again in the address bar.
> 
> If this PR is not needed, I am ok closing it. thanks.

This helps us to think about a good prompt for adding a rule for the "sites disabled" and get more users to easily request or add a rule. I will work to improve that from this disfromcussion. 

I can close out this but I am flagging this for the UX development that is in the works.

cschanaj:
Thanks.

PR #17213, opened 2019-01-02
add archive.ph

jc86035:
I was redirected here from another archive.is site a few days ago, and its metadata (on the website and at WHOIS) indicates that it's the same site as archive.is. (The SSL certificate became valid on `29 December 2018, 8:05:27 AM GMT`; presumably this is the first one.)


J0WI:
closed by #18137

PR #17215, opened 2019-01-03
Create tianya.cn-mixedcontent.xml

ivysrono:


pipboy96:
I'm not really sure if it's appropriate to create new mixedcontent rulesets, seeing how Tor Browser is no longer any different from Firefox and Chrome, mixed content wise, e.g. all active mixed content is blocked. If this behavior causes a rewritten website to break, rewrite should not be done.

pipboy96:
@ivysrono Are there any other issues than MCB with this ruleset?

ivysrono:
No more issues. BTW, most MCB are fixed by https://github.com/ivysrono/UpgradeMixedContentByBlacklist

PR #17222, opened 2019-01-06
Create Overpass-api.de.xml

nmxcgeo:
Every subdomain found on Duckduckgo is tested and included as testurl.
Thus I expected to have sufficient tests. Can someone please explain to me, why this fails?

Greetings Nmxcgeo


nmxcgeo:
Ok, I've found the problem. Waiting for feedback now, if there's anything left to be done.

J0WI:
We have a policy to avoid wildcards:
https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#left-wildcards
Could you just transform them into targets?

J0WI:
Thanks!

PR #17232, opened 2019-01-08
Create Cucumber.io.xml

Bisaloo:


src/chrome/content/rules/Cucumber.io.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="Cucumber.io">
+
+	<target host="cucumber.io" />
+	<target host="clinic.cucumber.io" />
+	<target host="cukenfest.cucumber.io" />
+	<target host="docs.cucumber.io" />
+	<target host="handbook.cucumber.io" />
+	<target host="thoughts.cucumber.io" />
```
J0WI:
```
aruba.cucumber.io
shouty.cucumber.io
```

Bisaloo: For future reference, `aruba.cucumber.io` has active mixed content from `fonts.googleapis.com`

J0WI:
Thanks!

PR #17234, opened 2019-01-11
Remove Network message for unknown hosts

zoracon:
This refers to #17192, where sites that don't exist get blocked by our EASE mode page.

The conclusive sentiment here on my part is that with or without EASE (BAUR) a site that doesn't exist (http or https), will pose little threat to our users. Since all major browsers provide a page explaining that the site does not exist.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -650,7 +650,6 @@ function onErrorOccurred(details) {
       details.error.indexOf("net::ERR_CONNECTION_") == 0 ||
       details.error.indexOf("net::ERR_ABORTED") == 0 ||
       details.error.indexOf("NS_ERROR_CONNECTION_REFUSED") == 0 ||
-      details.error.indexOf("NS_ERROR_UNKNOWN_HOST") == 0 ||
       details.error.indexOf("NS_ERROR_NET_TIMEOUT") == 0 ||
       details.error.indexOf("NS_ERROR_NET_ON_TLS_HANDSHAKE_ENDED") == 0 ||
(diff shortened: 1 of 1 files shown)


Hainish:
I wonder why this was added in the first place.  This error may triggered in an overly-broad manner.  Can we check different errors that are returned with https://www.badssl.com/ to ensure this doesn't remove the interstitial for any hosts that are valid?

For testing, I've typically just done `curl -O badssl.html https://www.badssl.com/` and replaced all the HTTPS links with HTTP within the HTML.  This is necessary since direct links to HTTPS pages do not trigger the interstitial, and it's easier than copy/pasting the URL and removing the 's' each time.

zoracon:
> > I wonder why this was added in the first place. This error may triggered in an overly-broad manner. Can we check different errors that are returned with https://www.badssl.com/ to ensure this doesn't remove the interstitial for any hosts that are valid?
> > For testing, I've typically just done `curl -O badssl.html https://www.badssl.com/` and replaced all the HTTPS links with HTTP within the HTML. This is necessary since direct links to HTTPS pages do not trigger the interstitial, and it's easier than copy/pasting the URL and removing the 's' each time.
> 
> I will check that out as well to make sure this doesn't hit other cases. I truly hope "unknown hosts" doesn't mean something else :(

@Hainish I went through badssl.com and have not seen `"NS_ERROR_UNKNOWN_HOST"` occur for other SSL errors. I did see `""NS_ERROR_CONNECTION_REFUSED""` occur sometimes for non existent sites. Maybe this is where the murky issues lied out with these errors? Maybe that jogs a memory?

PR #17256, opened 2019-01-19
Create Interbix.net.xml

Jayson-Fong:


src/chrome/content/rules/Interbix.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Interbix.net (partial)">
+	<target host="interbix.net" />
+	<target host="www.interbix.net" />
+	<target host="archive.interbix.net" />
```
J0WI: `doclib.interbix.net`

src/chrome/content/rules/Interbix.net.xml
```diff
@@ -0,0 +1,8 @@
...
+	<target host="press.interbix.net" />
+  <securecookie host=".+" name=".+"/>
+  <rule from="^http:" to="https:"/>
```
J0WI: Please use only tabs to indent lines

src/chrome/content/rules/Interbix.net.xml
J0WI: After adding this target, you can also remove the `partial` flag in the name

PR #17259, opened 2019-01-20
Do not run on https requests

pipboy96:
Now when https->https rewrites are completely removed, we can skip all https requests.


pipboy96:
ignore this, this breaks upgrade-insecure-requests

PR #17260, opened 2019-01-21
Create Gigabyte.us.xml

pipboy96:


pipboy96:
For people who wish to verify validity of this URL: https://dns.google.com/query?name=download.gigabyte.us&type=CNAME&dnssec=true

src/chrome/content/rules/Gigabyte.us.xml
```diff
@@ -0,0 +1,30 @@
...
+		- event
+		- reseller
+		- businesscenter
+		- rebate
+		- brix
+		- rma
+-->
+<ruleset name="Gigabyte.us (partial)">
+	<target host="store.gigabyte.us" />
+	<target host="download.gigabyte.us" />
```
J0WI: cert does not match for me

pipboy96: try now.

PR #17261, opened 2019-01-21
Create Gigabyte.eu.xml

pipboy96:


pipboy96:
For people who wish to verify validity of this url: https://dns.google.com/query?name=download.gigabyte.eu&type=CNAME&dnssec=true

src/chrome/content/rules/Gigabyte.eu.xml
```diff
@@ -0,0 +1,27 @@
...
+		- join
+		- brix
+		- g1
+		- upgrade
+
+	Expired:
+		- gaming
+-->
+<ruleset name="Gigabyte.eu (partial)">
+	<target host="download.gigabyte.eu" />
```
J0WI: cert does not match for me

pipboy96: try now.

PR #17262, opened 2019-01-21
Create download.gigabyte.asia.xml

pipboy96:


pipboy96:
For people who wish to verify validity of this URL: https://dns.google.com/query?name=download.gigabyte.asia&type=CNAME&dnssec=true

Issue #17271, opened 2019-01-24
Redirect loop for kooora.com


tsl143:
HTTPS Everywhere seems to cause an infinite redirect loop on kooora.com.

As soon as I disable HTTPS Everywhere for kooora.com, the problem is resolved.

Maybe useful-
`koora.com`, redirects to `http://www.kooora.com/default.aspx`

PR #17272, opened 2019-01-24
Disable rule

tsl143:
Fixes: #17271 

kooora.com falls in redirect loop
http://kooora.com <=> http://www.kooora.com/default.aspx


Bisaloo:
I should have asked before but can you use double quotes instead of single quotes as well please?

tsl143:
> I should have asked before but can you use double quotes instead of single quotes as well please?

No worries, I too should have been careful to see double quotes 🤦

Bisaloo:
Thank you!

PR #17276, opened 2019-01-25
Create CeCILL.info.xml

Bisaloo:


src/chrome/content/rules/CeCILL.info.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="CeCILL.info">
+
+	<target host="cecill.info" />
+	<target host="www.cecill.info" />
+
+	<rule from="^http://www\.cecill\.info/"
+		to="https://cecill.info/" />
```
J0WI: Please add a comment for the cert issue.

PR #17279, opened 2019-01-26
redirect uwaterloo csclub mirrors

jtracey:
The certificate is only valid for *.csclub.uwaterloo.ca, but these other subdomains rank higher in searches.

(For context, this is one of the most popular official Canadian mirrors for a bunch of linux distros and open source projects.)


J0WI:
Thanks!

PR #17281, opened 2019-01-27
Delete Cyphdbyhiddenbhs.onion.xml

pipboy96:
Doesn't seem to be reachable either over http or https.


J0WI:
WFM

J0WI:
@pipboy96 `torsocks` is much easier to use: `torsocks curl https://cyphdbyhiddenbhs.onion` :wink:

pipboy96:
@J0WI If I understand it correctly, `torsocks` hooks the program's calls to network functions in order to function. Doesn't exactly seem like a clean solution. I would prefer using curl's native capabilities if possible.

PR #17285, opened 2019-01-27
Add Rsyslog.com.xml

J0WI:
#15750


src/chrome/content/rules/Rsyslog.com.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+	Nonfunctional hosts in *.rsyslog.com:
+		- rsyslog.com (m)
+		- alpine.rsyslog.com (t)
```
Bisaloo: cookbook.rsyslog.com (m)

J0WI: Thanks, added.

PR #17287, opened 2019-01-27
Remove X.org.xml in favor of X.Org.xml

J0WI:
There was a race condition during the review process of #15753 and #17131.
@cschanaj


cschanaj:
Sorry I didn't notice that there is already a merged ruleset.

PR #17291, opened 2019-01-27
Add Nuttcp.net.xml

J0WI:
#15750


Bisaloo:
That's weird, the certificate is fine on Chrome but I can reproduce your error in Firefox.

PR #17300, opened 2019-01-28
 Add Rephial.org.xml to ruleset

RemakingEden:
I have added a new ruleset for Rephial.org.

This is my first ruleset and opensource contribution so please let me know if I did anything wrong and you need something changing. Thank you

### DIFF
--- a/src/chrome/content/rules/Rephial.org.xml
+++ b/src/chrome/content/rules/Rephial.org.xml
@@ -1,13 +1,13 @@
 <!--
     Redirect to HTTP:	
     buildbot.rephial.org
-
+	www.rephial.org
+	www.trac.rephial.org
+	
 -->
 <ruleset name="Rephial.org">
 	<target host="rephial.org" />
-	<target host="www.rephial.org" />
 	<target host="trac.rephial.org" />
-	<target host="www.trac.rephial.org" />
 
 	<rule from="^http:" to="https:" />
 </ruleset>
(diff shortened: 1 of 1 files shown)


RemakingEden:
Ive just noticed I have not added a test URL. Is this mandatory? If it is I can add it and submit a new pull request

src/chrome/content/rules/Rephial.org.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+    Redirect to HTTP:	
+    buildbot.rephial.org
+
+-->
+<ruleset name="Rephial.org">
+	<target host="rephial.org" />
+	<target host="www.rephial.org" />
```
zoracon: @RemakingEden Thank you for your contribution! I'm glad HTTPSE could be your first open source project. Glad to have you.

It looks like those TLS certs are only for `trac.rephial.org` and `rephial.org`, but not `www.trac.rephial.org` and `www.rephial.org`. Many web devevlopers miss adding both in for proper HTTPS configuration coverage for their domain

1. You can take several actions. To fix the test, you can add comments above in the that these `www` hosts are nonfunctional
2. Contact the website admin to amend their TLS Cert
They seem to use Let's Encrypt, so assuming they probably used Certbot to install...they can use this link for reference
```suggestion
```
https://certbot.eff.org/docs/using.html#changing-a-certificate-s-domains

src/chrome/content/rules/Rephial.org.xml
```diff
@@ -0,0 +1,14 @@
...
+	<target host="trac.rephial.org" />
+	<target host="www.trac.rephial.org" />
```
zoracon:
```suggestion
```

RemakingEden:
@zoracon Thanks for the reply, i'm happy to be here!

PR #17335, opened 2019-01-30
Delete GetItHere.xyz.xml

pipboy96:
nxdomain


pipboy96:
> Can you remove the reference here please? https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/GameCopyWorld.com.xml

Done.

PR #17338, opened 2019-01-30
Run HSTS-prune and fix impacted rulesets

Bisaloo:


J0WI:
Depends on https://github.com/EFForg/https-everywhere/pull/17507

PR #17345, opened 2019-01-31
Create ruleset: Abercrombie.com

injust:


injust:
Looks like the CI build is failing due to the server blocking requests. Even cURL fails unless you give it a browser UA.
Adding a whitelist entry.

ivysrono:
`Access Denied` all?

PR #17347, opened 2019-02-01
Change the open URL link to a button

patch-malone:
When trying to visit a page that isn't available as HTTPS a user may be shown a cancel page where they can either cancel the insecure visit or choose to override the restriction. The link of the URL shown at the bottom of that page misleadingly goes to an internal extension URL. Instead, change the actionable item into a button with a clear label describing the action.

### Before
before screenshot

### After
Notice the button at the bottom of the page
after screenshot

### DIFF
--- a/chromium/pages/cancel/index.html
+++ b/chromium/pages/cancel/index.html
@@ -8,11 +8,21 @@
     <link rel="stylesheet" href="style.css" type="text/css" />
   </head>
   <body>
-    <h1 id="https-everywhere"><img src="../../images/banner-red.png" alt="HTTPS Everywhere" /></h1>
-    <p data-i18n="cancel_he_blocking_explainer"></p>
-    <p id="url-paragraph"><span id="url-label">URL: </span><a href="#" id="originURL"></a></p>
-    <script src="../translation.js"></script>
-    <script src="../util.js"></script>
-    <script src="ux.js"></script>
+    <div class="content">
+      <h1 id="https-everywhere"><img src="../../images/banner-red.png" alt="HTTPS Everywhere" /></h1>
+      <p data-i18n="cancel_he_blocking_explainer"></p>
+
+      <p id="url-paragraph">
+        <span id="url-label">URL: </span><a href="#" id="url-value">PLACEHOLDER</a>
+      </p>
+
+      <form method="get" action="#" id="url-actions-form">
+        <button type="button" name="open" formaction="#" id="open-url-button" data-i18n="cancel_open_page">Open insecure page</button>
+      </form>
+
+      <script src="../translation.js"></script>
+      <script src="../util.js"></script>
+      <script src="ux.js"></script>
+    </div>
   </body>
 </html>
--- a/chromium/pages/cancel/style.css
+++ b/chromium/pages/cancel/style.css
@@ -1,20 +1,41 @@
 body{
   margin-top: 6em;
-  left: 50%;
   position: relative;
-  overflow: hidden;
+  text-align: center;
 }
 
-h1 img{
-  position: relative;
-  left: -289px;
-}
-
-p{
+.content{
   width: 600px;
-  position: relative;
-  left: -300px;
+  margin: auto;
+  text-align: left;
+}
+
+h1{
+  text-align: center;
+}
+
+form, button, p{
   font-size: 12pt;
   font-family: sans-serif;
   line-height: 150%;
+}
+
+form{
+  overflow: auto;
+  margin-bottom: 1em;
+}
+
+form button{
+  padding: .5em 1em;
+  background-color: #aaa;
+  color: #fff;
+  border: none;
+  border-radius: 4px;
+  cursor: pointer;
+}
+
+#open-url-button{
+  display: inline-block;
+  float: left;
+  background-color: #ec1e1e;
 }
--- a/chromium/pages/cancel/ux.js
+++ b/chromium/pages/cancel/ux.js
@@ -28,15 +28,20 @@ function replaceLink(explainer) {
 function displayURL() {
   const cancelURL = new URL(window.location.href);
   const originURL = decodeURI(cancelURL.searchParams.get('originURL'));
-  const originURLLink = document.getElementById('originURL');
-  originURLLink.innerText = originURL;
+  const originURLLink = document.getElementById('url-value');
+  const openURLButton = document.getElementById('open-url-button');
 
-  originURLLink.addEventListener("click", function() {
+  originURLLink.innerHTML = originURL;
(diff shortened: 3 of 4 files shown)


pipboy96:
Needs some work with CSS styling, but the idea is good.

pipboy96:
@patch-malone Wait a couple of minutes, I will write a style.

chromium/pages/cancel/index.html
```diff
@@ -8,11 +8,18 @@
...
-    <script src="../translation.js"></script>
-    <script src="../util.js"></script>
-    <script src="ux.js"></script>
+    <div class="content">
+      <h1 id="https-everywhere"><img src="../../images/banner-red.png" alt="HTTPS Everywhere" /></h1>
+      <p data-i18n="cancel_he_blocking_explainer"></p>
+      <p id="url-paragraph">
+        <span id="url-label">URL: </span><a href="#" id="url-value">PLACEHOLDER</a>
+      </p>
+      <form method="get" action="#" id="url-form">
```
pipboy96: no need for `<form>` here.

chromium/pages/cancel/index.html
```diff
@@ -8,11 +8,18 @@
...
+        <button name="open" formaction="#" id="open-url-button">Open page insecurely</button>
```
pipboy96: no need for `formaction` either.

chromium/pages/cancel/style.css
```diff
@@ -1,20 +1,34 @@
...
+form, form button{
+  font-size: 12pt;
+}
+
+form button{
+  padding: 8px 16px;
+}
+
+#open-url-button{
+  background: red;
```
pipboy96:
```css
button {
	cursor: pointer;
	display: inline-block;
	margin: 0;
	padding: 1em 1em;
	background: repeating-linear-gradient( 45deg, red, red 7.5px, salmon 7.5px, salmon 15px );
	color: #fff;
	font-weight: bold;
	text-align: center;
	vertical-align: middle;
	-moz-user-select: none;
  user-select: none;
	border: 0;
	border-radius: 5px;
}
```

chromium/pages/cancel/index.html
pipboy96: It would be better if it was more obvious that this option permanently whitelists the website.

chromium/pages/cancel/index.html
```diff
@@ -8,11 +8,18 @@
...
-    <p data-i18n="cancel_he_blocking_explainer"></p>
-    <p id="url-paragraph"><span id="url-label">URL: </span><a href="#" id="originURL"></a></p>
-    <script src="../translation.js"></script>
-    <script src="../util.js"></script>
-    <script src="ux.js"></script>
+    <div class="content">
+      <h1 id="https-everywhere"><img src="../../images/banner-red.png" alt="HTTPS Everywhere" /></h1>
+      <p data-i18n="cancel_he_blocking_explainer"></p>
+      <p id="url-paragraph">
+        <span id="url-label">URL: </span><a href="#" id="url-value">PLACEHOLDER</a>
```
pipboy96: make this a `<span>` that autoselects if you click it.

chromium/pages/cancel/index.html
patch-malone: Should the notice be part of the button or in subtext underneath?

pipboy96: Make the button say: "Open insecurely and do not ask again".

chromium/pages/cancel/style.css
patch-malone: Thanks for providing the CSS. The screenshot has been updated.

chromium/pages/cancel/index.html
patch-malone: Are you wanting an anchor element instead of the button (and form)?

pipboy96: @patch-malone `<button>` doesn't need a form.

patch-malone: What about the button's `formaction`? I think it's better to persist the URL in the standard HTML attribute (`formaction`) on `<button>` itself rather than dynamically reading it from another part of the page because that will make it clear to someone inspecting source where the button takes them, along with making the value available to screenreaders.

chromium/pages/cancel/index.html
patch-malone: The button text has been updated.

chromium/pages/cancel/index.html
patch-malone: I can understand hiding the internal extension URL though. What do you recommend?

pipboy96: Change the button to `<a>` then. That would make more sense in this case.

patch-malone: The `<button>` has been changed to an `<a>` tag.

chromium/pages/cancel/index.html
patch-malone: I think it's valuable to know whether you've visited the URL before. A user may remember that they visited a website a few years ago but not quite remember the exact URL. If they search for it on http://excite.com/ and click an imposter link the :visited status can expose that.

pipboy96: @patch-malone You need to make the link show a message box that asks if it's ok to visit unencrypted website then.

patch-malone: I see. Do you think the <span> is better?

pipboy96: @patch-malone Yes.

pipboy96:
I have another idea. To make the page seem more like an error page, the bg should be yellow, possibly striped too.

pipboy96:
You can apply these styles to `html`:
```css
background: repeating-linear-gradient(45deg, #ffdc00, #ffdc00 10px, #e5c600 10px, #e5c600 20px);
min-height: 100vh; /* this fixes weird looping */
```

pipboy96:
Maybe instead of closing the page you should make it go back to the page that had unencrypted link and if that's impossible (page was opened directly) you should simply hide the button?

patch-malone:
Tab history isn't accessible. `history.back()` doesn't work correctly in this instance and `document.referrer` is blank. I'll remove the button. It can be added in a separate PR.

patch-malone:
@Bisaloo Are any other changes required before merging?

Bisaloo:
Let's ask @zoracon. I know some people at the EFF are usually involved in UI PR review but I'll let @zoracon ping them if necessary.

zoracon:
@Bisaloo I can pull this down and review. The interstitial page does need admittedly need work. We have UX dev plans in the making #16669, but I see no harm with looking this change over to see how it plays out.

PR #17352, opened 2019-02-01
Make it possible to open debugging ruleset editor on Firefox

pipboy96:
On latest Firefox Nightly, pressing Ctrl-Z on options page does nothing if you are viewing it from Firefox UI, and if you are viewing it directly it just shows a "popup blocked" message.

### DIFF
--- a/chromium/pages/options/index.html
+++ b/chromium/pages/options/index.html
@@ -53,6 +53,10 @@
       </div>
     </div>
 
+    <div class="section-wrapper">
+      <a href="/pages/debugging-rulesets/index.html" target="_blank" data-i18n="options_debuggingRulesets"></a>
+    </div>
+
     <script src="ux.js"></script>
     <script src="../translation.js"></script>
     <script src="../util.js"></script>
--- a/chromium/pages/options/index.html
+++ b/chromium/pages/options/index.html
@@ -53,9 +53,7 @@
       </div>
     </div>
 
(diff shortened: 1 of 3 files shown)


pipboy96:
screenshot

Hainish:
I'm not a fan of exposing this to users.  This functionality was added specifically for maintainers to make it easier to test rulesets.  Creating an additional route to add custom rulesets adds UX complexity, and will cause people to rely on this even if we add a warning to it.

Popups are probably not blocked intentionally for extensions in Firefox, I'm wondering if this is a regression we can open a bug report for in bugzilla..

pipboy96:
@Hainish There is already a bug open, it seems to be stalled.

BTW can we simply remove this UI and ask people who want to use it to directly type the URL for the editor if you want it to be hidden? Just make sure to document it for devs to be able to use it.

pipboy96:
https://bugzilla.mozilla.org/show_bug.cgi?id=1429624

Last activity 9 months ago, no follow-up.

pipboy96:
wait, can we make <kbd>Ctrl</kbd> + <kbd>Z</kbd> make this link visible and set a "developer mode" flag in extension storage?

Hainish:
@pipboy96 yes, that makes sense.  I'd merge that.

pipboy96:
@Hainish Ok, I close this and start work on this suggestion.

PR #17362, opened 2019-02-02
Added new ruleset - Rrpproxy.net.xml

RemakingEden:
Another domain for https://github.com/EFForg/https-everywhere/issues/2324


Bisaloo:
It depends on where you're running `git checkout -b newbranchname`. 

If you're running it from `master`, you should be fine. But if you run it from a branch where you already made changes, those will be copied to the new branch (`newbranchname`).

Otherwise, you could run `git checkout -b newbranchname master` to always create your new branch from master.

RemakingEden:
@Bisaloo Thanks mate, that's really helpful. Still getting my head around everything. I will close these requests and open new correct ones.

PR #17365, opened 2019-02-03
[LINBIT.com] Merge rulesets and add targets

Bisaloo:


src/chrome/content/rules/LINBIT.com.xml
```diff
@@ -0,0 +1,11 @@
+<ruleset name="LINBIT.com (partial)">
+
+	<target host="linbit.com" />
+	<target host="www.linbit.com" />
+	<target host="blogs.linbit.com" />
+	<target host="docs.linbit.com" />
+	<target host="downloads.linbit.com" />
+
```
cschanaj: Would you add `my.linbit.com`? otherwise lgtm.

PR #17369, opened 2019-02-04
Added ruleset - Rb-Media-group.de.xml

RemakingEden:
Messed up with the other commit. This is a clean one.

### DIFF
--- a/src/chrome/content/rules/Rb-media-group.de.xml
+++ b/src/chrome/content/rules/Rb-media-group.de.xml
@@ -1,0 +1,59 @@
+<!--
+	Nonfunctional hosts in *.rb-media-group.de:
+	http redirect
+        apc4.rb-media-group.de
+        ats2.rb-media-group.de
+        backups1.rb-media-group.de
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Rb-media-group.de.xml
```diff
@@ -0,0 +1,59 @@
...
+        statistik.rb-media-group.de
+        support.rb-media-group.de
+        switch-s2.rb-media-group.de
+        transfer.rb-media-group.de
+        v13.rb-media-group.de
+        webmail.rb-media-group.de
+        vertrieb.rb-media-group.de
+	connection refused
+        autodiscover.rb-media-group.de
+        community.rb-media-group.de
```
J0WI: WFM

RemakingEden: What does that mean sorry?

J0WI: This host works for me over HTTPS, so please consider adding it as target.

PR #17371, opened 2019-02-04
Removed ruleset - Hobocomp.com.xml

RemakingEden:
nxdomain


pipboy96:
Domain resolves but IP is dead. Should be ok to remove though.

PR #17374, opened 2019-02-04
Removed ruleset - T3Blog.com.xml

RemakingEden:
nxdomain


RemakingEden:
Can anyone who originally saw this domain confirm if this is the same blog? https://t3-blog.com/ 

If it is i will change the ruleset rather than deleting.

pipboy96:
domain still resolves but broken.

pipboy96:
@RemakingEden 
They are not.
https://web.archive.org/web/20170720051946/https://www.t3blog.com/

RemakingEden:
> domain still resolves but broken.

What do we do in this case? Do we usually remove or is there other things we check for? 

Thanks for clarifying the difference between all of these, it's good to know.

pipboy96:
@RemakingEden We may delete.

J0WI:
Thanks!

PR #17392, opened 2019-02-06
 Update Microsoft.xml 

cschanaj:


zoracon:
Was about to run the merge but looks like there are fetch issues with `http://assets.microsoft.com/`

PR #17395, opened 2019-02-06
Run duplicate-whitelist-cleanup.sh

pipboy96:


Bisaloo:
Thank you for your contribution but this is run as part of the release process.

PR #17401, opened 2019-02-07
Update Aftenposten.xml

pipboy96:
There are hundreds of subdomains, I'm not going to go through each of them.


src/chrome/content/rules/Aftenposten.xml
```diff
@@ -1,123 +1,27 @@
...
+	<target host="onlinesos.aftenposten.no" />
+	<target host="oslopuls.aftenposten.no" />
+	<target host="reise.aftenposten.no" />
+	<target host="rubrikkcache.aftenposten.no" />
+	<target host="sport.aftenposten.no" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http://sport\.aftenposten\.no/.*" to="https://www.aftenposten.no/100Sport/" />
+	<test url="http://sport.aftenposten.no/index.html" />
```
J0WI: I think this rule does not work as you expect. Please also add a comment for the mismatch on `sport.aftenposten.no`.

pipboy96: @J0WI It's supposed to remove the path part of the URL.

pipboy96: @J0WI Done.

J0WI: `.*` is not necessary, the regex will match anyway.

pipboy96: @J0WI It would result in path not being removed, e.g. `https://www.aftenposten.no/100Sport/yhojszma7r6v2esfjk4ri5xo7q`

PR #17405, opened 2019-02-07
Remove wildcard targets from HypeMachine.xml

cschanaj:


cschanaj:
@J0WI Travis passed. thanks!!

J0WI:
I suspected that `x.hypem.com` is already covered by Bitly, but for some reason it didn't show up when I searched for it...
Could you revert this please, because the list was updated by Bitly itself in the past.

cschanaj:
@J0WI sorry for the delay. I have force-pushed a new commit accordingly. thanks.

J0WI:
LGTM, thanks!

PR #17408, opened 2019-02-07
Replace empty securecookie name tag with trivial wildcard

cschanaj:
Honestly, I didn't expect there exists securecookie tag with empty name attributes... they work as if `(?:)` in Chrome, so I guess they don't break anything. IMHO, we should change them to functionally equivalent trivial wildcard for better consistence

```bash
sed -i 's/name=""/name=".+"/g' *.xml
```


cschanaj:
cc @zoracon for reviews and comments

PR #17415, opened 2019-02-08
Add ruleset: Bashy.io

dnoetzel:


src/chrome/content/rules/Bashy.io.xml
```diff
@@ -0,0 +1,13 @@
+ <!--
+	Nonfunctional hosts in *.bashy.io:
+		- bashy.io
```
J0WI: Could be redirected to `www`.

dnoetzel:
`https://launchpage.bashy.io/` gives a mixed content warning, because it tries to load resources from `http://www.opensourcebash.org/`, but since that domain is down, no content is actually loaded from an insecure resource. With that in mind, do you have any recommendations on how the `launchpage` host should be handled?

J0WI:
Even if the resources would exists, it seems to be passive content only. You can ignore it.

src/chrome/content/rules/Bashy.io.xml
```diff
@@ -1,13 +1,14 @@
- <!--
+<!--
 	Nonfunctional hosts in *.bashy.io:
-		- bashy.io
```
J0WI: Please keep this comment to document the special rule below.

J0WI:
Thanks!

dnoetzel:
Thank you for all of the help. Sorry for all of the back and forth this PR has required.

PR #17420, opened 2019-02-09
 Enhance utils/trivialize-rules/trivialize-cookie-rules.js 

cschanaj:
This will remove the restrictions on the number of `securecookie` and `rule` tags and trivialize `securecookie` in a more generic way.


cschanaj:
cc @zoracon for reviews and comments

pipboy96:
@cschanaj I have some objections to the current code, specifically about the regexes and async parts. Will make a review.

pipboy96:
@cschanaj This is what I currently have:
```js
'use strict'

/**
 * For future contributors, this script was written to trivialize a special
 * form of securecookie rules which is equal to trivial securecookie rule
 * under the current HTTPSe implementation. This allows trivialize-rules.js to
 * simplify rulesets otherwise impossible due to the lack of support of
 * unbounded wildcard RegExps in explode-regexp.js (by design).
 *
 * As stated in trivialize-rules.js, a securecookie rule works if there is an
 * existing rule covering the exact same domain.
 *
 * In particular, <securecookie host=".*example.com" .../> covers all
 * subdomains in cookie.host given at least one working rule cover
 * the same subdomains.
 *
 * If there is at least one such rule, for e.g. www.example.com, the mentioned
 * securecookie rules can be reduced to
 *     <securecookie host="^\.?www\.example\.com$" .../>
 * Else, the securecookie rule do not take into effect and thus can be ignored
 * during the runtime.
 *
 * So, regardless of the existence of any rule, for all targets listed in a
 * ruleset, the securecookie rule can be interpreted as
 *     <securecookie host=".+" .../>
 *
 * For simplicity, this implementation require securecookie#host includes a
 * tailing $
 */

const util = require('util')
const path = require('path')
const xml2js = require('xml2js')
const fs = require('graceful-fs')

const readdir = util.promisify(fs.readdir)
const readFile = util.promisify(fs.readFile)
const writeFile = util.promisify(fs.writeFile)
const parseString = util.promisify(xml2js.parseString)

const rulesDir = 'src/chrome/content/rules'

const isTrivial = securecookie => securecookie.host === '.+' && securecookie.name === '.+'

;(async () => {
  const dir = (await readdir(rulesDir)).filter(filename => filename.endsWith('.xml'))
  const filePromises = dir.map(async filename => {
    const fileContents = await readFile(path.join(rulesDir, filename), 'utf8')
    const ruleset = (await parseString(fileContents)).ruleset
    const targets = ruleset.target.map(target => target.$.host)
    const securecookies = ruleset.securecookie
      ? ruleset.securecookie.map(sc => sc.$)
      : []

    if (!securecookies.length || (securecookies.length === 1 && isTrivial(securecookies[0]))) {
      return
    }

    for (const securecookie of securecookies) {
      if (!securecookie.name === '.+') {
        return
      }

      if (!securecookie.host.startsWith('^.+') &&
        !securecookie.host.startsWith('^.*') &&
        !securecookie.host.startsWith('.+') &&
        !securecookie.host.startsWith('.*') &&
        !securecookie.host.startsWith('(?:.*\\.)?') &&
        !securecookie.host.startsWith('(?:.+\\.)?') &&
        !securecookie.host.startsWith('^(?:.*\\.)?') &&
        !securecookie.host.startsWith('^(?:.+\\.)?')) {
        return
      }
    }

    // make sure each domains and its subdomains are covered by at least
    // one securecookie rule
    const securedDomains = new Map()

    for (let target of targets) {
      // we cannot handle the right-wildcards based on the argument above
      if (target.includes('.*')) {
        return
      }
      if (target.includes('*.')) {
        // www may collide with unrelated <securecookie> tags.
        target = target.replace('*.', 'wildcard-rule-test')
      }
      securedDomains.set(target, false)
      securedDomains.set('.' + target, false)
    }

    for (const securecookie of securecookies) {
      const pattern = new RegExp(securecookie.host)
      for (const [key] of securedDomains) {
        if (pattern.test(key)) {
          securedDomains.set(key, true)
        }
      }
    }

    // If each value of map is true, return.
    if (!([...securedDomains.values()].every(a => a))) {
      return
    }

    // Remove every <securecookie> tag except the last one and replace the last <securecookie> tag with a trivial one.
    const scRe = /([ \t]*)<securecookie\s+host\s*=\s*["'].*["']\s*name\s*=\s*["'].*["']\s*\/>[ \t]*/g

    let len = fileContents.match(scRe).length
    const source = fileContents.replace(scRe, (_, p1) => {
      // Determine if we are at the last match or not.
      if (!--len) {
        // We are at the last match. Replace the last tag with a trivial one.
        return `${p1}<securecookie host=".+" name=".+" />`
      } else {
        // Remove all <securecookie> tags except the last one.
        return ''
      }
    })

    await writeFile(path.join(rulesDir, filename), source, 'utf8')
  })
  await Promise.all(filePromises)
})()
```

cschanaj:
@zoracon @pipboy96 I'm closing this in favor of #17438 with some refactoring and modification to the syntax thanks to @pipboy96 . Please leave your comments there if needed. thanks.

PR #17431, opened 2019-02-09
[DataMineLab.com] Simplify ruleset

Bisaloo:


src/chrome/content/rules/DataMineLab.com.xml
```diff
@@ -1,15 +1,14 @@
-<ruleset name="DataMineLab.com (partial)">
+<ruleset name="DataMineLab.com">
 
 	<target host="dataminelab.com" />
-	<target host="*.dataminelab.com" />
+	<target host="www.dataminelab.com" />
```
cschanaj: would you add a top comment on the certificate mismatch? otherwise lgtm.

cschanaj:
thanks!

PR #17466, opened 2019-02-13
Update Al-Madina.com.xml

pipboy96:
No evidence of mixed content issues.


src/chrome/content/rules/Al-Madina.com.xml
```diff
@@ -1,9 +1,4 @@
-<!--
-	Mismatched:
-		madina.al-madina.com, CN=*.prod.ams3.secureserver.net.
-		waseet.al-madina.com, CN=*.prod.ams3.secureserver.net.
```
J0WI: Why did you removed them?

pipboy96:
@J0WI Done.

PR #17484, opened 2019-02-16
Add ruleset: LeanCrew.com

dnoetzel:


src/chrome/content/rules/LeanCrew.com.xml
```diff
@@ -0,0 +1,20 @@
...
+		- www
+		- admin
+		- eww
+		- mail
+		- vvv
+		- vwww
+		- wew
+		- wqw
+		- wvw
+		- ww2
```
J0WI: This domain uses a wildcard DNS record, so you can remove this comments and just describing this behavior.

PR #17489, opened 2019-02-21
Adjust CSS for long http links

zoracon:
- Email tracking links get long and break layout, so cap overflow with an ellipsis


wold5:
Long urls are still shown for me, stretching the logo to fullscreen (v2021.4.15, Firefox 88.0, Ubuntu derivative)

Adding CSS property "overflow-x: scroll;" to ".copy_block" solves it, though less elegantly.

PR #17274, opened 2019-01-25
Disable "Show Counter" test (temporarily) for Chrome

zoracon:
- Show counter has been failing lately for many PRs
- Suspected cause: deprecation warning
- Pushing fix for deprecation
Message
(Warnings.warn('use options instead of chrome_options', DeprecationWarning))

Edit: Checks out to not be the issue, will investigate further for a solution

Going to skip the test for now in Chrome. If not hashed out by Release time I will merge for the sake of people's PRs

PR #17490, opened 2019-02-21
Add patch for Chrome run time issue

zoracon:
Caught that at times "item" is not defined during "show counter" test. Causing Chrome and Chrome Beta test to fail. Adding this in as a stop gap so this test can continue for future PRs.

Related: #17274

### DIFF
--- a/chromium/pages/options/ux.js
+++ b/chromium/pages/options/ux.js
@@ -20,6 +20,10 @@ document.addEventListener("DOMContentLoaded", () => {
   };
 
   sendMessage("get_option", defaultOptions, item => {
+    // Sometimes item does not define on runtime in Chrome
+    if( !item ){
+      item = defaultOptions;
+    }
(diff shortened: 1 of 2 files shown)


zoracon:
Looks like I now botched the Firefox flow with this change, gonna look into the test itself more to see how to improve on it

chromium/pages/options/ux.js
```diff
@@ -20,6 +20,10 @@ document.addEventListener("DOMContentLoaded", () => {
   };
 
   sendMessage("get_option", defaultOptions, item => {
+    // Sometimes item does not define on runtime in Chrome
+    if(!item) {
+      item = defaultOptions;
+    }
```
pipboy96:
```js
item = item || defaultOptions
```
?

zoracon:
This has been resolved through amending the selenium test, closing this out

PR #17491, opened 2019-02-21
Add ruleset: FLIF.info

dnoetzel:


src/chrome/content/rules/FLIF.info.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+	Nonfunctional hosts in *.flif.info:
+		- www
```
Bisaloo: Please say that it's mismatched / has an invalid certificate (as you prefer)

src/chrome/content/rules/FLIF.info.xml
```diff
@@ -0,0 +1,9 @@
...
+-->
+<ruleset name="FLIF.info">
+	<target host="flif.info" />
```
Bisaloo: Please add `www.flif.info` with a rule to redirect it to https://flif.info/

Bisaloo:
Thank you!

PR #17492, opened 2019-02-22
Update utils/merge-rulesets.py

cschanaj:
cc @Hainish @zoracon this will slightly reduce the maximum memory usage by around 0.6% while keeping the ruleset format backward-compatible.


Hainish:
Thanks!

Issue #17206, opened 2018-12-31
Browser tests fail on Fedora 29


tsilvap:
Type: code issue
OS: Fedora 29 (GNU/Linux)

How to reproduce:
1. Clone `https-everywhere` repo locally and `cd` into the project root.
2. Run `bash install-dev-dependencies.sh` (no errors).
3. Run `bash test.sh`. It exits with this error message: `Chrome: HTTP to HTTPS redirection failed`

Here is the complete log if that's needed:
https://gist.github.com/tsilvap/e1feb74fa4730fa5c7f285e999d846ce

I have no idea what could be the issue here. I also tried this inside a pipenv running Python 3.6 (since Fedora uses Python 3.7 by default, I thought this could be the issue) but I get the same error.

oleblaesing:
I'm also using Fedora 29 (Python 3.7) and I can't even run `bash install-dev-dependencies.sh` successfully, because I run into this error everytime I try. Please let me know if you need some more information about my environment.

```
https-everywhere [master] ✘ bash install-dev-dependencies.sh

Warning: Installing the development dependencies for HTTPS Everywhere
may alter your system, installing requirements both within the package
management system and also external binaries.

Are you sure you want to continue? [y/N]: y

Last metadata expiration check: 1:33:10 ago on So 03 Feb 2019 15:59:32 CET.
Package firefox-65.0-1.fc29.x86_64 is already installed.
Package gcc-8.2.1-6.fc29.x86_64 is already installed.
Package git-2.20.1-1.fc29.x86_64 is already installed.
Package libcurl-devel-7.61.1-7.fc29.x86_64 is already installed.
Package libxml2-devel-2.9.8-5.fc29.x86_64 is already installed.
Package libxslt-devel-1.1.32-3.fc29.x86_64 is already installed.
Package python2-devel-2.7.15-11.fc29.x86_64 is already installed.
Package redhat-rpm-config-118-1.fc29.noarch is already installed.
Package xorg-x11-server-Xvfb-1.20.3-3.fc29.x86_64 is already installed.
Package which-2.21-11.fc29.x86_64 is already installed.
Package findutils-1:4.6.0-21.fc29.x86_64 is already installed.
Package procps-ng-3.3.15-4.fc29.x86_64 is already installed.
Package openssl-1:1.1.1a-1.fc29.x86_64 is already installed.
Package openssl-devel-1:1.1.1a-1.fc29.x86_64 is already installed.
[...]

PR #17494, opened 2019-02-22
Fix install-dev-dependencies.sh on Fedora

Bisaloo:
Tested in https://github.com/EFForg/https-everywhere/issues/17206#issuecomment-466161962

Source: https://stackoverflow.com/questions/21530577/fatal-error-python-h-no-such-file-or-directory

This was not updated since the switch to python3


Bisaloo:
@zoracon, I took the liberty of requesting a review from you. Please let me know if you'd prefer I don't do it in the future.

zoracon:
> @zoracon, I took the liberty of requesting a review from you. Please let me know if you'd prefer I don't do it in the future.

Requesting review is fine with me, especially for PRs that fix issues

PR #17495, opened 2019-02-22
Delete ruleset: QupZilla.com

injust:


Bisaloo:
Since the project got renamed to Falkon (https://en.wikipedia.org/wiki/Falkon#History), I don't expect them to renew the domain.

PR #17496, opened 2019-02-22
Default_off Libgen.xml

Bisaloo:


Giltyhub:
Since libgen.pw just redirects to ambry.pw (which doesn't have any SSL problems on my end) wouldn't it be a good idea to just make a redirect libgen.pw -> ambry.pw?

Bisaloo:
I don't think it would be a good idea to redirect to `ambry.pw`. As soon as the redirect changes, HTTPS Everywhere will cause an illegitimate redirect. Especially for pirate sites such as this one that regularly switch domains.

A ruleset for ambry.pw on the other hand would be within the scope of this project and probably a good idea.

PR #17501, opened 2019-02-23
Create ThisAirbnbDoesNotExist.com.xml

Giltyhub:


src/chrome/content/rules/ThisAirbnbDoesNotExist.com.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	Mismatch:
+		www.thisairbnbdoesnotexist.com
```
gloomy-ghost: works for me, but the whole domain seems just an alias for thisrentaldoesnotexist.com

Giltyhub: @gloomy-ghost thanks, seems like they just fixed it, added!

PR #17504, opened 2019-02-23
Create AlFaseeh.com.xml

Giltyhub:


Bisaloo:
For the record, MCB here but it's from facebook so I'm fine with it.

https://www.alfaseeh.com/vb/faq.php

@cschanaj, @J0WI, @zoracon, what's your opinion on this?

J0WI:
I'm also seeing mixed scripts from `http://www.alfaseeh.com/`.

Bisaloo:
You're right, I missed that. @Giltyhub, could you add `platform="mixedcontent"` please (with a link to https://www.alfaseeh.com/vb/faq.php as an example).

@J0WI, more generally, we are okay with MCB breaking ads or tracking scripts, what's your opinion about fb like buttons?

J0WI:
> What's your opinion about fb like buttons?

I would consider them as tracking. There was a similar situation in https://github.com/EFForg/https-everywhere/pull/4459.

zoracon:
@J0WI @Bisaloo unless there is a large push to accommodate tracking content then I am fine with it in this case

Giltyhub:
@Bisaloo i don't think the lack of scripts on that faq page break any crucial functionality

Bisaloo:
@Giltyhub, I cannot really verify this since this page is in a language I don't understand. Since you're a regular contributor on this project and this does not pose any security issue (worst case scenario is degraded UX), I am going to trust you on this and go ahead with the merge.

PR #17512, opened 2019-02-24
Update GameSpot.xml

pipboy96:


src/chrome/content/rules/GameSpot.xml
```diff
@@ -1,86 +1,12 @@
...
-		- auth
-
-
-	Observed cookie domains:
-
-		- ^ ¹
-		- . ²
-		- auth ³
-		- www ¹
-
```
zoracon: The comments  above are probably still useful to keep around fro future updates. Unless there was some reason that they'd only add to further confusion.

pipboy96: @zoracon Done.

Issue #9971, opened 2017-05-20
Option to turn mixedcontent rulesets on


Bisaloo:
How can we turn all mixedcontent rulesets on at once with the WebExtensions HTTPS-Everywhere?

The following command doesn't seem to work anymore:

```
extensions.https_everywhere.enable_mixed_rulesets;true
```

Related: #9958

jeremyn:
@Bisaloo WebExtensions add-ons shouldn't count on having access to `about:config`, see https://wiki.mozilla.org/WebExtensions/FAQ#Will_I_have_access_to_about:config_or_the_preferences.3F . So it makes sense that HTTPS Everywhere isn't doing anything with that setting.

ivysrono:
#8506 enable mixedcontent on all platform by default.

Bisaloo:
@ivysrono, that's not really the purpose of this issue. It is a great find and helps a lot with mixedcontent issue but it doesn't turn on HTTPS-Everywhere `platform="mixedcontent"` rulesets.

jeremyn:
A more general version of this issue is a way to turn on any combination of `platform`s, not just `mixedcontent`, using a checklist.

Hainish:
A lot of the rules that have been marked as `platform="mixedcontent"` were broken when the browsers first introduced MCB.  I suspect many sites have fixed their mixed content problems in response to MCB being the default behavior now, but these rules should be manually checked rather than blanket-enabled, or else we risk blocking access to site resources.  Currently there are ~1325 such rules:

user@https-everywhere ~/workspace/https-everywhere/rules (webextensions) $ git grep "mixedcontent" | wc -l
1325

I suspect that our tagging of these rules as `mixedcontent` is old enough that these rules should probably be reviewed anyway.  In https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#general-info, I've made a note for those interested in contributing to prioritize the `mixedcontent` rules.

jeremyn:
@Hainish I think this issue is to add the option to enable all the `mixedcontent` rules, not that we should enable them all by default.

Hainish:
We could add an option to localStorage, as I note in this comment: https://trac.torproject.org/projects/tor/ticket/21323#comment:15

[...]

PR #17518, opened 2019-02-24
Delete GO.com.xml

pipboy96:
Ruleset is extremely complex, effectively disabled (`platform="mixedcontent"`) and most of the website doesn't support HTTPS.


J0WI:
With this removal we lost a lot of information and also some popular targets like `abc.go.com` :confused:

pipboy96:
@J0WI These targets were effectively disabled for everyone anyway.

J0WI:
There is a preference to enable all mixedcontent platform rules.

pipboy96:
@J0WI Currently not even Tor Browser has MCB disabled, and disabling it greatly reduces security of HTTPS.

J0WI:
I was referring to #9971

pipboy96:
@J0WI I was going to discuss no longer treating mixedcontent breakage as "platform", but treat it simply as a broken ruleset, due to security implications of allowing active (or even passive, given the possibility of a 0day for "passive" formats like svg images) mixed content, and the fact Tor Browser now does not treat mixed content in a different way than normal browsers, and enabling active missing content is not an "official", easily enabled feature in either FF or Chromium (about:config in FF, console flag in Chromium).

pipboy96:
Also the fact that we will most likely eventually make an actual interface for viewing all rulesets and enabling/disabling them without visiting the websites they are triggered on. Users can simply enable all broken rulesets with default_off reason being `mixedcontent` or similar if they want to.

J0WI:
In fact HTTPS Everywhere can mitigate mixed content issues. The reason why they are disabled is because the browsers security feature block mixed content before we can rewrite the url to HTTPS:
https://www.eff.org/https-everywhere/rulesets#mixed-content-blocking-mcb
https://bugzilla.mozilla.org/show_bug.cgi?id=878890

There are concepts to fix this, but most unexpired users will just blame the extension or the browser to be broken:
https://bugzilla.mozilla.org/show_bug.cgi?id=1435733

The Tor Browser has a completely different security philosophy and blocks content such as JavaScript and SVGs entirely (depending on your [security level](https://tb-manual.torproject.org/security-slider/)).

pipboy96:
@J0WI Currently the only way to both prevent mixed content and also try to fix it if possible is to enable EASE and disable all MCB. There's also an ugly way of doing something similar by injecting CSP headers into each response with `upgrade-insecure-requests` enabled.

zoracon:
Generally, yes we accept MCB rules. The reason i deleted was due to the many non functional hosts and complex regex applied just to make it work. I'll gladly add this back in myself as a rule since it is a 1k coverage rule and I'll admit I missed that label in my review yesterday. I'll look into this domain and put up a PR later.

PR #17527, opened 2019-02-25
Add ruleset: Casual-Effects.com

dnoetzel:


src/chrome/content/rules/Casual-Effects.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="Casual-Effects.com">
```
pipboy96:
```xml
<!--
	Mismatched:
		- mail
-->
```

src/chrome/content/rules/Casual-Effects.com.xml
```diff
@@ -0,0 +1,6 @@
...
+	<target host="casual-effects.com" />
+	<target host="www.casual-effects.com" />
```
pipboy96:
```xml
	<target host="cgpp-net.casual-effects.com" />
	<target host="www.cgpp-net.casual-effects.com" />
	<target host="codeheartjs.casual-effects.com" />
	<target host="endlyss.casual-effects.com" />
	<target host="graphicscodex.casual-effects.com" />
	<target host="rocketgolfing.casual-effects.com" />
```

src/chrome/content/rules/Casual-Effects.com.xml
```diff
@@ -0,0 +1,6 @@
...
+
```
pipboy96:
```xml
```

src/chrome/content/rules/Casual-Effects.com.xml
dnoetzel: My mistake; I forgot to run `sublist3r` for this domain.

src/chrome/content/rules/Casual-Effects.com.xml
dnoetzel: This is the first time I have encountered this while contributing rulesets. So that I can write better rulesets in the future, how does one tell when the `securecookie` tag is required?

pipboy96: Unless the website breaks with this rule, you should usually add it.

PR #17531, opened 2019-03-01
New rules for additifs-alimentaires.net

Harest:


src/chrome/content/rules/additifs-alimentaires.net.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="additifs-alimentaires.net">
+	<target host="additifs-alimentaires.net" />
+	<target host="www.additifs-alimentaires.net" />
```
pipboy96: `<target host="joomla.additifs-alimentaires.net" />`

src/chrome/content/rules/additifs-alimentaires.net.xml
```diff
@@ -0,0 +1,7 @@
...
+
```
pipboy96: `<securecookie host=".+" name=".+" />`

src/chrome/content/rules/additifs-alimentaires.net.xml
```diff
@@ -0,0 +1,7 @@
...
+	<rule from="^http:"
+		to="https:" />
```
pipboy96: Keep this on one line, like this: `<rule from="^http:" to="https:" />`.

Harest: It's from the first copy/pasta of [the contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#example-ruleset-after-style-guidelines-are-applied-with-test-urls) (example ... after ... section). I'll change that.

src/chrome/content/rules/additifs-alimentaires.net.xml
Harest: Cf. contributor guide : `For securecookie tags, if you know that all cookies on the included targets can be secured (which in particular means that the cookies are not used by any of its non-securable subdomains)`. How do you know it's the case ? Or is it now enforced for every new ruleset ?

pipboy96: @Harest Generally, if `<securecookie>` doesn't cause any issues, it should be added.

Harest: Okay, i'll keep that in mind and add it then. Thanks for the info.

src/chrome/content/rules/additifs-alimentaires.net.xml
Bisaloo: I'd actually prefer we don't add internal subdomains. We've had complaints from domain owners about this in multiple instances in the past.

There is almost no chance a user will visit it anyways so there is not real risk to drop it IMO.

@Harest, could you revert that please?

Bisaloo:
For future reference, there is active mixed content triggering MCB at https://www.additifs-alimentaires.net/livres.php but it doesn't look like switching to HTTPS breaks anything important.

pipboy96:
@Bisaloo IMO if only thing broken by upgrade to HTTPS is trackers/ads, it's fine.

src/chrome/content/rules/additifs-alimentaires.net.xml
Harest: Damn i just saw that by luck @Bisaloo, i had no notification. I'll remove it right now.

Bisaloo:
Thank you!

PR #17538, opened 2019-03-02
Create sis001.xml

15556133244:


src/chrome/content/rules/sis001.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="sis001">
+	<target host="sis001.com" />
+	<target host="www.sis001.com" />
+
+	<!--	mismatched	-->
+	<target host="sis001.us" />
+	<target host="www.sis001.us" />
```
pipboy96: put these is a separate file.

15556133244: sis001.us === sis001.com

PR #17540, opened 2019-03-02
Create sex8.xml

15556133244:


src/chrome/content/rules/sex8.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="sex8">
+	<target host="sex8.cc" />
+	<target host="www.sex8.cc" />
+
+	<!--	mismatched	-->
+	<target host="s8bbs.com" />
+	<target host="www.s8bbs.com" />
```
pipboy96: put these in a separate file.

15556133244: s8bbs.com === sex8.cc

PR #17544, opened 2019-03-04
Update 126.net.xml

ivysrono:


ivysrono:
@pipboy96 It's better to understand that put special rules with the corresponding target hosts.

pipboy96:
@ivysrono Rules are ruleset-global, and their position in the ruleset (except the order) does not matter.

pipboy96:
I get dropped connection on lofter.ph.126.net and imgsize.ph.126.net from Sweden, but imglf1.ph.126.net is 404.

ivysrono:
Both OK from China, Now.
BTW, you can see ruleset for baidu.com https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Baidu.xml
Move all rules to the bottom of the ruleset may hard to userstand.

J0WI:
Thanks!

PR #17551, opened 2019-03-06
Ensure that when the extension is updated, older rulesets are cleared…

Hainish:
In https://trac.torproject.org/projects/tor/ticket/29454 there's a discussion about how Tor Browser users are often getting older versions of rulesets.

Currently, if any rulesets have been successfully downloaded and verified from the update channels (https://www.https-rulesets.org/ for instance) , this will overwrite the extension built-in rulesets.

Polling of the update channels occurs every 24 hours, or on browser startup if it's been longer than 24 hours since the last poll.

In Tor Browser this is a problem, since the tor circuit hasn't been established at browser startup, when the poll request is made.  It fails, and the rulesets aren't updated.

For some Tor Browser users, they've successfully downloaded rulesets once, but subsequently they've gotten fails for the above reason.  This means they're stuck on an old version of the rulesets, even after the extension itself is updated and they have newer rules available within the extension.

This PR does a few things:

- When an update to the extension is made, we clear the https://www.https-rulesets.org/ rulesets from the browser. We'll then be using the extension-bundled rulesets.
- We only update the extension to the latest rulesets from https://www.https-rulesets.org/ if the timestamp for those rulesets is newer than the extension itself. This will prevent the case where HTTPS Everywhere downloads the newest ruleset, then an update to the extension is made which clears the https-rulesets.org rulesets, and then you need to download the rulesets from https-rulesets.org again.

### DIFF
--- a/chromium/background-scripts/update.js
+++ b/chromium/background-scripts/update.js
@@ -2,7 +2,7 @@
 
 "use strict";
 
-let combined_update_channels;
+let combined_update_channels, extension_version;
 const { update_channels } = require('./update_channels');
 
 // Determine if we're in the tests.  If so, define some necessary components.
@@ -14,9 +14,14 @@ if (typeof window === "undefined") {
     pako = require('../external/pako-1.0.5/pako_inflate.min.js'),
     { TextDecoder } = require('text-encoding'),
     chrome = require("sinon-chrome"),
-    window = { atob, btoa, chrome, crypto, pako, TextDecoder };
+    window = { atob, btoa, chrome, crypto, pako, TextDecoder },
+    fs = require('fs');
+
+  extension_version = JSON.parse(fs.readFileSync('./manifest.json')).version;
 
   combined_update_channels = update_channels;
+} else {
+  extension_version = chrome.runtime.getManifest().version;
 }
 
 (function(exports) {
@@ -28,6 +33,9 @@ let store,
 
 // how often we should check for new rulesets
 const periodicity = 86400;
+
+const extension_date = new Date(extension_version.split('.').slice(0,3).join('-'));
+const extension_timestamp = extension_date.getTime() / 1000;
 
 let imported_keys;
 
@@ -184,7 +192,7 @@ async function applyStoredRulesets(rulesets_obj) {
           const rulesets_string = new TextDecoder("utf-8").decode(rulesets_byte_array);
           const rulesets_json = JSON.parse(rulesets_string);
 
-          resolve({json: rulesets_json, scope: update_channel.scope});
+          resolve({json: rulesets_json, scope: update_channel.scope, replaces: update_channel.replaces_default_rulesets});
         } else {
           resolve();
         }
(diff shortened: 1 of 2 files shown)


chromium/background-scripts/update.js
```diff
@@ -198,11 +201,15 @@ async function applyStoredRulesets(rulesets_obj) {
 
   const channel_results = (await Promise.all(rulesets_promises)).filter(isNotUndefined);
 
-  if(channel_results.length > 0) {
-    for(let channel_result of channel_results) {
-      rulesets_obj.addFromJson(channel_result.json.rulesets, channel_result.scope);
+  let replaces = false;
+  for(let channel_result of channel_results) {
+    if(channel_result.replaces == true) {
```
pipboy96: Use `===`, else it will match any trueish value, e.g. `[]` or `{}`.

chromium/background-scripts/update.js
```diff
@@ -198,11 +201,15 @@ async function applyStoredRulesets(rulesets_obj) {
 
   const channel_results = (await Promise.all(rulesets_promises)).filter(isNotUndefined);
 
-  if(channel_results.length > 0) {
-    for(let channel_result of channel_results) {
-      rulesets_obj.addFromJson(channel_result.json.rulesets, channel_result.scope);
+  let replaces = false;
+  for(let channel_result of channel_results) {
```
pipboy96: `const`?

chromium/background-scripts/update.js
```diff
@@ -271,12 +284,33 @@ function destroyTimer() {
...
+  });
+}
+
 async function initialize(store_param, cb) {
   store = store_param;
   background_callback = cb;
 
   await loadUpdateChannelsKeys();
 
+  if (await store.local.get_promise('extensionTimestamp', 0) != extension_timestamp) {
```
pipboy96: `!==` should be used to avoid edge cases with values that are loosely equal (e.g. `0` and empty string or `false`).

chromium/background-scripts/update.js
```diff
@@ -271,12 +284,33 @@ function destroyTimer() {
   }
 }
 
+function clear_replacement_update_channels() {
+  let keys = [];
+  for (let update_channel of combined_update_channels) {
```
pipboy96: `const`?

chromium/background-scripts/update.js
cschanaj: Is it safe to assume `channel_result.replaces == true` implies the content of `channel_result.json.rulesets` originates from EFF? My understanding is that if an evil `update_channel` crafted the response such that `channel_result.replaces` is `true`,  the bundled rules from EFF might never apply.

Hainish: That is the assumption - there is no way in the update channels UX to add an update channel that has the replacement flag set to true, and I don't plan on exposing that to the UX either.  This means that adding an update channel will not result in your default rulesets being overwritten.

cschanaj: I see your point and I believe this is a reasonable assumption. (though I guess this should be explicitly documented somewhere, possibly in this or later PR).

PR #17553, opened 2019-03-07
Display the version even for rulesets that are bundled with the exten…

Hainish:
…sion to avoid confusion

### DIFF
--- a/chromium/pages/popup/ux.js
+++ b/chromium/pages/popup/ux.js
@@ -198,18 +198,15 @@ document.addEventListener("DOMContentLoaded", function () {
   rulesets_versions.addSpan = function(update_channel_name, ruleset_version_string) {
     let timestamp_span = document.createElement("span");
     timestamp_span.className = "rulesets-version";
-    timestamp_span.innerText = chrome.i18n.getMessage("about_rulesets_version") + " " + update_channel_name + ": " + ruleset_version_string;
+    timestamp_span.innerText = `${chrome.i18n.getMessage("about_rulesets_version")} ${update_channel_name}: ${ruleset_version_string}`;
     this.appendChild(timestamp_span);
   }
 
   sendMessage("get_ruleset_timestamps", null, timestamps => {
-    let replace = timestamps.reduce((reduced, [update_channel, timestamp]) => {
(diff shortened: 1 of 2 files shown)


chromium/pages/popup/ux.js
```diff
@@ -194,16 +194,30 @@ document.addEventListener("DOMContentLoaded", function () {
   version_info.innerText = the_manifest.version;
 
   let rulesets_versions = e('rulesets-versions');
+
+  rulesets_versions.addSpan = function(update_channel_name, ruleset_version_string) {
+    let timestamp_span = document.createElement("span");
+    timestamp_span.className = "rulesets-version";
+    timestamp_span.innerText = chrome.i18n.getMessage("about_rulesets_version") + " " + update_channel_name + ": " + ruleset_version_string;
```
pipboy96:
```js
timestamp_span.innerText = `${chrome.i18n.getMessage("about_rulesets_version")} ${update_channel_name}: ${ruleset_version_string}`
```

chromium/pages/popup/ux.js
```diff
@@ -194,16 +194,30 @@ document.addEventListener("DOMContentLoaded", function () {
...
+    this.appendChild(timestamp_span);
+  }
+
   sendMessage("get_ruleset_timestamps", null, timestamps => {
+    let replace = timestamps.reduce((reduced, [update_channel, timestamp]) => {
```
cschanaj: I think the `replace` here is the same as the `replaces` in https://github.com/EFForg/https-everywhere/commit/de97a001c8e625ffa33ed05363a533b6ce4f755b#diff-893c26ee11675d91a6893982afc0f38aR217, it would be a bit more consistent to use `replaces` with `s`

```javascript
let replaces = timestamps.some([update_channel, timestamp] => {
   return update_channel.replaces_default_rulesets && timestamp > 0;
});
```

pipboy96: @cschanaj You need parenthesis around `[update_channel, timestamp]`.

pipboy96: @cschanaj Also no need of `return`:
```js
let replaces = timestamps.some(([update_channel, timestamp]) => (
  update_channel.replaces_default_rulesets && timestamp > 0
));
```

PR #17559, opened 2019-03-08
[TRIVIAL] Unmark Verisign-Inc.com.xml executable

maciejsszmigiero:
Verisign-Inc.com.xml is a XML ruleset file, there is no need for it to be
executable.


maciejsszmigiero:
> Nice find!

Thanks!

cschanaj:
thanks!

PR #17438, opened 2019-02-12
Refactor and enhance trivialize-cookie-rules.js

cschanaj:

PR #17565, opened 2019-03-10
Update some securecookie with no coverage

cschanaj:
this is a side product of #17438. since the quantity discovered is relative small, I am not modify the script there to include the check. cc  @J0WI @Bisaloo @Hainish  @zoracon


Hainish:
Thanks, looks good!

PR #17573, opened 2019-03-11
Re-enable Safe-mail.net.xml (fix #17267)

### LINKED ISSUE #17267: Invalid test URL in Safe-mail.net ruleset
Type: ruleset/website issue

Right now, there is a [ruleset for Safe-mail.net](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Safe-mail.net.xml) that is off by default, apparently because the test URL http://rimon.safe-mail.net/ could not be accessed. My guess is that it should be possible to make the ruleset functional again by using the test URL `http://orange.safe-mail.net/` instead of `http://rimon.safe-mail.net/`.

J0WI:


zoracon:
Closes #17267

PR #17574, opened 2019-03-11
 Run node utils/trivialize-rules/trivialize-cookie-rules.js

cschanaj:
### **[DO NOT MERGE]**

related #17438 

To reproduce https://github.com/EFForg/https-everywhere/pull/17574/commits/da207e4eab8e2a98d7c6ace1c1d5433511fae62a

```bash
# travis log from 9fce6df validations test
wget https://api.travis-ci.org/v3/job/504953396/log.txt

# list all xml with failed validations and hash them into a temporary file
cd rules
grep ^ERROR ~/log.txt | grep -Eo '[^/ ]+.xml' | sort -u | xargs sha256sum > /tmp/tmp.txt

# https-everywhere
cd ..
cat /tmp/tmp.txt | sed 's/  /,1,0,/g >> utils/ruleset-whitelist.csv 

```

PR #17578, opened 2019-03-13
Run node utils/trivialize-rules/trivialize-cookie-rules.js

cschanaj:
this is a re-open of #17574, related #17438


src/chrome/content/rules/AIA-Surety.xml
```diff
@@ -1,18 +1,15 @@
...
 	<target host="aiasurety.com" />
-	<target host="*.aiasurety.com" />
+	<target host="becomeanagent.aiasurety.com" />
+	<target host="www.aiasurety.com" />
 	<target host="expertbail.com" />
 	<target host="www.expertbail.com" />
 
 
-	<securecookie host="^(?:.*\.)?(?:aiasurety|expertbail)\.com$" name=".+" />
+	<securecookie host=".+" name=".+"/>
```
pipboy96:
```suggestion
	<securecookie host=".+" name=".+" />
```

`sed -e 's/<securecookie host=".+" name=".+"\/>/<securecookie host=".+" name=".+" \/>/g'`

utils/trivialize-rules/trivialize-cookie-rules.js
```diff
@@ -65,7 +65,7 @@ let isTrivial = securecookie => {
     }
 
     for (let securecookie of securecookies) {
-      if (!securecookie.name === ".+") {
+      if (".+" !== securecookie.name) {
```
pipboy96:
```suggestion
      if (securecookie.name !== ".+") {
```

cschanaj: I agree this looks better, but I am not sure why they don't seem to work the same on my machine though...

pipboy96: @cschanaj Add code like:
```js
if ((securecookie.name !== ".+") !== (".+" !== securecookie.name)) {
  console.log(securecookie.name)
}
```
and run your script. Most likely reason is name being empty or undefined.

cschanaj: thanks for your suggestion. It turns out that I might made a mistake when comparing the output of the script.

src/chrome/content/rules/AIA-Surety.xml
cschanaj: a better approach is to modify the script to include the whitespace. however, I want to avoid re-doing the same PR and add complications to the review process. Will submit another PR to fix this later. thanks.

cschanaj:
@Hainish would you be interested to take a look and merge this at your convenience?

PR #17577, opened 2019-03-12
bgp.he.net now redirects to HTTPS, so I don't think its nonfunctional

theel0ja:

PR #17584, opened 2019-03-13
Update HurricaneElectric.xml

pipboy96:
See also #17577.

### DIFF
--- a/src/chrome/content/rules/HurricaneElectric.xml
+++ b/src/chrome/content/rules/HurricaneElectric.xml
@@ -5,6 +5,8 @@
 	www.hurricaneelectric.net does not exist.
 
 	Mismatched:
+		- he.com
+		- www.he.com
 		- hurricaneelectric.net
 -->
 <ruleset name="Hurricane Electric (partial)">
--- a/src/chrome/content/rules/HurricaneElectric.xml
+++ b/src/chrome/content/rules/HurricaneElectric.xml
@@ -8,9 +8,6 @@
 		- he.com
 		- www.he.com
 		- hurricaneelectric.net
-
-	Timeout:
-		- code.he.net
 -->
 <ruleset name="Hurricane Electric (partial)">
 	<target host="he.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/HurricaneElectric.xml
```diff
@@ -1,43 +1,27 @@
...
+	<target host="dns.he.net" />
+	<target host="ipv6.he.net" />
 
-	<rule from="^http://((?:admin|code|cust|dns|ipv6|www)\.)?he\.net/"
-		to="https://$1he.net/" />
+	<target host="hurricaneelectric.net" />
 
+	<securecookie host=".+" name=".+" />
+	
+	<rule from="^http://(www\.)?he\.com/" to="https://$1he.net/" />
```
J0WI: This rule can be for `www` only, `^` is already covered by the generic rule below.

src/chrome/content/rules/HurricaneElectric.xml
```diff
@@ -1,43 +1,27 @@
...
-
-
-	Problematic domains:
-
-		- (www.)hu.com *
-		- hurricaneelectric.com *
-
-	* Alternative domains, same content, using he.net certificate
-
+	www.hurricaneelectric.net does not exist.
```
J0WI: Please keep a comment for the mismatch on `hurricaneelectric.net`.

J0WI:
@pipboy96 there are only some minor things left, would be great if you could finish this.

src/chrome/content/rules/HurricaneElectric.xml
pipboy96: The redirect is for `.com` → `.net`.

src/chrome/content/rules/HurricaneElectric.xml
pipboy96: :+1:

src/chrome/content/rules/HurricaneElectric.xml
J0WI: Ah, therefore the comment for `.com` should be kept

J0WI:
All tests passed in #18073.

PR #15210, opened 2018-04-21
Create michelbergerhotel.com.xml

DavidXochipilli:
New ruleset for the website of the [Michelberger](https://michelbergerhotel.com)

PR #17592, opened 2019-03-14
Create MichelbergerHotel.com.xml

pipboy96:


pipboy96:
Closes #15210.

src/chrome/content/rules/MichelbergerHotel.com.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+	Self-signed:
+		- mail
+		- vpn
+-->
```
J0WI: Wildcard DNS record

pipboy96: @J0WI I'm not sure what to do here. I guess I should simply add a note in the comment and only include the domains that actually have content (not 404).

pipboy96: `remote` sends the same response as the non-existent domains, so I removed it.

J0WI: What about the hosts in https://github.com/EFForg/https-everywhere/pull/17592#pullrequestreview-215209963?

pipboy96: `shop`: no response at `:443`.
`store`: 404 on both http and https, can't find working url.

J0WI: Try with `http://shop.michelbergermonkey.com/germany/` or `http://store.michelbergermonkey.com/germany/`

pipboy96: @J0WI Redirect to HTTP.

J0WI: Could you add them to the list?

pipboy96: @J0WI Done.

pipboy96: @J0WI Done.

J0WI: Thanks!

PR #17596, opened 2019-03-15
Fix DeprecationWarning in HTTPS Everywhere Checker

pipboy96:
Example of warning:
```
test/rules/src/https_everywhere_checker/check_rules.py:311: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  config = SafeConfigParser()
```


pipboy96:
@zoracon :ping_pong:

PR #17597, opened 2019-03-15
Remove auto-pass coverage test when not needed

pipboy96:
:ping_pong: @zoracon


pipboy96:
Closing for now.

pipboy96:
The previous commit removed the files that are already `default_off` since they are already not subject to fetch or coverage tests. There's no point excluding them again in the `ruleset-whitelist.csv`.

PR #17622, opened 2019-03-22
Remove byond rule

MrStonedOne:
per: http://www.byond.com/forum/?post=2438855 its breaking functionality on the website because of how its interacting with `form` targets.


Bisaloo:
@MrStonedOne, thank you for your PR!

Unfortunately, this file has been updated since you opened this PR and there is no some conflicts. Could you have a look please?

I'm also wondering whether we should simply default off this ruleset instead of removing it. It would allow us to add a link to the forum post in the top comment and thus keep memory of this. Otherwise, another contributor may propose a ruleset for this website again in a couple of months.

ZeWaka:
Hello! This is still breaking, so it would be great to have a resolution.

pipboy96:
@ZeWaka Merge conflict fixed.

PR #17623, opened 2019-03-22
Update default timestamp for deterministic builds

### LINKED ISSUE #17618: https-everywhere: xpi timestamp of files is 1/1/1980
This is Debian bug

https://bugs.debian.org/925177

The timestamp of files in the distributed xpi version of https-everywhere dates back to January 1980 on Linux systems. Obviously this is not accurate. 

Once https-everywhere shipped a script in utils/create_xpi.py that was responsible for creating the xpi file. It just lacked a date_time parameter.

https://salsa.debian.org/webext-team/https-everywhere/blob/99fcc50e0bb5617c64e55457a7b3e502013e0dc8/debian/patches/use-newer-timestamp-date.patch

Debian applied the aforementioned patch to change the timestamp to a recent value. The create_xpi.py script does not exist anymore but maybe you want to correct this anyway. You can easily reproduce this issue by unpacking the xpi file on a Debian GNU/Linux system.

Bisaloo:
Should fix #17618. ~~I'll test it later today.~~ Tested and it looks fine.

### DIFF
--- a/utils/zipfile_deterministic.py
+++ b/utils/zipfile_deterministic.py
@@ -77,7 +77,7 @@ LZMA_VERSION = 63
 # we recognize (but not necessarily support) all features up to that version
 MAX_EXTRACT_VERSION = 63
 
-DEFAULT_DATE = (1980,1,1,0,0,0)  # hard-coded timestamp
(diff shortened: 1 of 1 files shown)


pipboy96:
@Bisaloo Maybe set the date to the last commit date, instead of hardcoding it?

PR #17632, opened 2019-03-25
Update s8bbs.com.xml

pipboy96:


pipboy96:
I did not redirect s8bbs.com to sex8.cc because IPs are different and accessing s8bbs.com over HTTPS results in SSL alert no matter which SNI is sent.

PR #17637, opened 2019-03-27
Replace all instances of `"/>` with `" />`

pipboy96:
Reproduce with `for file in rules/*.xml; do sed -i 's_"/>_" />_g' $file; done`.


zoracon:
This PR is triggering many instances with rules that have fetch problems and issues. Since this is about spacing, this is not high priority for backlog cleaning and involves a total of 1067 files. 

I have to make the call to close this for now since it's not reviewable and time consuming to appease the need for a trailing space. Please continue to focus on the PRs that have new rulesets or need revisions.

pipboy96:
@zoracon I would like to eventually require consistent style for every new PR as a part of Travis, and eventually convert the old ones. I guess this will wait then.

pipboy96:
@zoracon We may simply regenerate ruleset-whitelist.csv, but I think you should do that directly, instead of a PR since apparently you are allowed to commit without review.

zoracon:
> @zoracon We may simply regenerate ruleset-whitelist.csv, but I think you should do that directly, instead of a PR since apparently you are allowed to commit without review.

I do not submit major changes without review

PR #17638, opened 2019-03-27
Replace all instances of `"\s+>` with `">`

pipboy96:
Reproduce with `for file in rules/*.xml; do sed -r -i 's_"\s+>_">_g' $file; done`.


pipboy96:
This may wait.

PR #17639, opened 2019-03-28
[PusherApp.com] Simplify ruleset

Bisaloo:


Bisaloo:
@pipboy96, please avoid editing directly on my branches, unless you have no other choices (eg. I stopped reponding, deleted my account, etc.) I suppose this goes for other contributors' branches as well.

Bisaloo:
Plus I disagree with your changes. The targets I listed as non functional are indeed non functional.

pipboy96:
@Bisaloo In which cases editing someone's branch is fine? BTW did you know you can simply `git reset --hard` to a point before the changes and force push to completely delete the commit?

pipboy96:
I was going to remove the failing targets after Travis was finished and then manually test the rest, and request review from @cschanaj.

pipboy96:
Excuse me for editing this anyway, I thought that it's fine to edit branches if you don't merge your own changes.

pipboy96:
The reason why I readded failing targets back was for Travis.

src/chrome/content/rules/PusherApp.com.xml
```diff
@@ -0,0 +1,21 @@
+<!--
+	For other Pusher coverage, see Pusher.com.xml.
+
+	Invalid certificate:
```
pipboy96:
```suggestion
	Mismatched:
```

src/chrome/content/rules/PusherApp.com.xml
```diff
@@ -0,0 +1,21 @@
...
+		blog.pusherapp.com
+		js.pusherapp.com
```
pipboy96: You can only leave the subdomain part:
```
	- blog
	- js
```

src/chrome/content/rules/PusherApp.com.xml
```diff
@@ -0,0 +1,21 @@
...
+
+-->
+<ruleset name="PusherApp.com">
+
+	<target host="pusherapp.com" />
+	<target host="www.pusherapp.com" />
+	<target host="api.pusherapp.com" />
+	<target host="app.pusherapp.com" />
+
```
pipboy96: I just tested these targets locally, and they are working:
```
	<target host="ws.pusherapp.com" />
	<target host="wss.pusherapp.com" />
```

src/chrome/content/rules/PusherApp.com.xml
```diff
@@ -0,0 +1,21 @@
...
+	<securecookie host="^app\.pusherapp\.com$" name=".+" />
```
pipboy96: Any specific reason for only adding this domain?

src/chrome/content/rules/PusherApp.com.xml
```diff
@@ -0,0 +1,21 @@
...
+
+	<rule from="^http:"
+		to="https:" />
```
pipboy96: No `\n` here, it's not needed, it fits on one line perfectly.

pipboy96:
@Bisaloo Just clarifying, I was aware that these targets may be indeed broken, I wanted Travis to test them, and then remove the non-working targets, like I usually do.

pipboy96:
@Bisaloo BTW, if you would have deleted your account or fork, it would be impossible to edit your PRs, so they would have to be either merged as is or closed.

Bisaloo:
I'd prefer to keep the style as it is (comments and newlines). It's the style used in many other rulesets, not only my own, and the style used in our contributing guide.

For `securecookie`, the consensus was to default to not adding them: https://github.com/EFForg/https-everywhere/issues/13762. I did not run any tests so I left it as it was.

`ws` and `wss` cause `HTTP 426` and we have no working URL to test them.

pipboy96:
@Bisaloo Please mention `ws` and `wss`. For me it's enough that the responses are exactly same. From the name it also seems that these subdomains are used for WebSocket communication, and there's no "working URL" to be talking about.

We also need some standardized style of top comment. I already asked if it's fine to mass remove Unicode characters that were used in some old styles and require all characters to be ASCII unless the domain in question itself has these characters. That could prevent redirecting to a malicious websites using Unicode tricks, like `disċordapp.com`.

cschanaj:
> Just clarifying, I was aware that these targets may be indeed broken, I wanted Travis to test them, and then remove the non-working targets, like I usually do.

@pipboy96 I suppose reviewers are expected to test the rulesets in an independent environment, since Travis can be biased. It would also be easier for others to reviews if there is no unnecessary changes. 

Besides, please avoid sending request reviews/ or pinging me frequently, my works here are totally voluntary and I will be looking into a PR myself if I am interested in reviewing it. It might also save you some time expecting a response from me in some cases.

pipboy96:
@cschanaj Thanks for explanation. However, how is Travis more biased than any random internet connection? The only way to test in a relatively unbiased way is to test from multiple locations, which is what we did for fetch-test.

Bisaloo:
By testing yourself locally and then on travis, you are basically checking from two different locations (3 if you're reviewing a ruleset where a contributor has done the same)

pipboy96:
@Bisaloo Makes sense, but I usually want to test on Travis since this is what matters when deciding whether the merge is fine, and whitelisting requires/should require extremely strong reasons.

Bisaloo:
Test failures are not related to my changes. I'm not sure when it happened. Should I fix it here?

cschanaj:
@Bisaloo I guess the test is failing test because `.active` is removed from the public suffix list in our upstream dependency. see #17643

pipboy96:
@ivysrono this is fine to merge.

PR #17642, opened 2019-03-29
[PyData.org] Update ruleset

fuglede:
Several previously problematic subdomains now work. Some font-related mixed content issues warnings will show up on `blaze` and `xarray` but nothing problematic.


fuglede:
CI error should be unrelated to this PR:

> ERROR src/chrome/content/rules/Bitly_branded_short_domains.xml: Target 'get.active' missing eTLD

J0WI:
Thanks!

PR #17643, opened 2019-03-29
Remove get.active from Bitly_branded_short_domains.xml

cschanaj:
It is breaking Travis due to an upstream change in our dependency https://github.com/publicsuffix/list/commit/0e2a405f597a3c1be456d704b42bdd5e0d4954bb. It should be safe to remove this `target` since it is already gone from the DNS. Also note that this is the only `.active` domain we have. 

Related: #17639, #17642


Bisaloo:
Leaving others some time to have a look and I'll merge this tomorrow if nobody has done it by then.

Bisaloo:
Actually, on second thought and careful check, this should be safe. Plus, it confuses other contributors when they try to submit PR.

cschanaj:
thanks!

PR #17644, opened 2019-03-29
[SphinxDoc.org] New ruleset

fuglede:


fuglede:
CI failure is unrelated:

> ERROR src/chrome/content/rules/Bitly_branded_short_domains.xml: Target 'get.active' missing eTLD

Bisaloo:
Thank you for your PR! Could you please:

- [ ] Add a top comment explaining why we need a specific rule for `sphinx-doc.org`
- [ ] Rename the file to `Sphinx-doc.org.xml` (or `Sphinx-Doc.org.xml`). I prefer when the filename is exactly the domain name. That way, you can know which domain is handled without even opening the file
- [ ] Update the `ruleset name` to `Sphinx-doc.org` to match the filename

Bisaloo:
Thank you!

PR #17656, opened 2019-03-31
Fix bug in Chromium that makes popup links always load in non-incognito mode

### LINKED ISSUE #4247: Donate to EFF in private mode (chromium)
If Chrome is in private mode, the links in the menu of HTTPS Everywhere ("about https everywhere", "donate to eff") are opened in a non-private window. If there is a quick fix, that would be great.

Might be related to #4212 ...

pipboy96:
Closes #4247.


Hainish:
Works for me - thanks!

PR #17665, opened 2019-04-01
Add Saveyourinternet.eu.xml

J0WI:


Bisaloo:
`Sublist3r` finds may more subdomains. Each one is made from the country code of EU countries + `.saveyourinternet.eu`. Eg. `at.saveyourinternet.eu`, `be.saveyourinternet.eu`, etc.

Are you interested in adding them?

J0WI:
Is it worth to use a wildcard?

PR #12319, opened 2017-09-01
Remove support for wildcard in the middle

Bisaloo:
#4369

PR #17669, opened 2019-04-01
Remove wild-card-in-the-middle check in rules.js

zoracon:
Related to removing wildcard in the middle support #4369
Related #12319

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -562,21 +562,9 @@ RuleSets.prototype = {
       return nullIterable;
     }
 
-    // Replace each portion of the domain with a * in turn
-    let segmented = host.split(".");
-    for (let i = 0; i < segmented.length; i++) {
-      let tmp = segmented[i];
-      segmented[i] = "*";
-
-      results = (this.targets.has(segmented.join(".")) ?
-        new Set([...results, ...this.targets.get(segmented.join("."))]) :
-        results);
-
-      segmented[i] = tmp;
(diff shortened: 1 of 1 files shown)


Bisaloo:
I'm not really familiar with node but wouldn't this break right wildcards as well? 

We still have a couple of right wildcards in rulesets such as [`Google.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Google.xml). We can remove them by fixing #13682 first.

cschanaj:
> 
> 
> I'm not really familiar with node but wouldn't this break right wildcards as well?
> 
> We still have a couple of right wildcards in rulesets such as [`Google.xml`](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Google.xml). We can remove them by fixing #13682 first.

I agree with @Bisaloo that it might break right-wildcards. We need a standalone check for the right-wildcards if we want to preserve them. IMHO fixing #13682 is a better option though.

AFAIK, this might also break left wildcards such that `www.asset.example.com` will not be matched by `*.asset.example.com` but `*.example.com`. I think this can be fixed by adjusting the starting index in the lower for-loop.

zoracon:
I'll test with wild cards on the right. it likely breaks those looking at the logic. There's logic still looking through the left below the deleted logic though. I'll close this for now if this does break those on the right.

zoracon:
Looks like this does in fact break the ones on the right, closing for now.

PR #17672, opened 2019-04-03
Updated URL to new Atlas URL

Gitoffthelawn:
Chose to use https://atlas.eff.org/index.html instead of https://atlas.eff.org/ because links to go back to "HTTPS Everywhere Atlas" main page from the Atlas itself go to the index.html file.  This is also a little safer in the event of a server misconfiguration or hack.

### DIFF
--- a/chromium/pages/popup/index.html
+++ b/chromium/pages/popup/index.html
@@ -71,7 +71,7 @@
     </div>
 
     <footer>
-      <a id="viewAllRules" href="https://www.eff.org/https-everywhere/atlas" target="_blank" data-i18n="menu_viewAllRules"></a><br>
+      <a id="viewAllRules" href="https://atlas.eff.org/index.html" target="_blank" data-i18n="menu_viewAllRules"></a><br>
       <br>
       <a id="aboutTitle" href="https://www.eff.org/https-everywhere" target="_blank" data-i18n="about_title"></a><br>
       <br>


Bisaloo:
Thank you for your contribution! Could you change your target branch to `master` please? 

I agree with this change in the principle but I think it would be good to make sure with the EFF that the domain `atlas.eff.org` is here to stay and not a temporary redirect. cc @zoracon

Gitoffthelawn:
@Bisaloo You're welcome!  I had a 50/50 of selecting the correct branch, and I guessed wrong.  So I can learn from my mistake, why should pull requests be for the `master` branch instead of the `release` branch?

In the meantime, I'll update the master branch.

Gitoffthelawn:
@Bisaloo I created a pull request for the `master` branch, and received the following error:

> Merging is blocked
> The base branch restricts merging to authorized users. Learn more about protected branches.
> You’re not authorized to merge this pull request.

Bisaloo:
> So I can learn from my mistake, why should pull requests be for the `master` branch instead of the `release` branch?

It's just the way we work in this project. All changes go to the `master` branch while the `release` branch contains code from the current release. But other projects may have different conventions. Usually, if repos are well configured, your PR should automatically target the correct branch when you create them.

> In the meantime, I'll update the master branch.

I don't know if you are already aware of it but [you can change the target branch of a PR](https://stackoverflow.com/a/24159161). But it can sometimes lead to complex situations so maybe that's why you created a new PR.

> @Bisaloo I created a pull request for the master branch, and received the following error:

Interesting. I did not notice that. The settings of this repo were recently updated to prevent direct pushing on the master branch. I did not expect it would generate warnings for users. You should not worry about it though, because you are not trying to push directly to master.

Gitoffthelawn:
@Bisaloo Thanks Hugh for all the helpful information!

PR #17680, opened 2019-04-04
Add Css3test.com.xml

J0WI:


src/chrome/content/rules/Css3test.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Css3test.com">
+	<target host="css3test.com" />
+	<target host="www.css3test.com" />
```
Bisaloo: This target has an invalid certificate (`www.github.com`) but can be redirected to https://css3test.com/

J0WI:
SSL Labs reports inconsistent results: https://www.ssllabs.com/ssltest/analyze.html?d=css3test.com

Bisaloo:
As far as I can tell, it has been updated and now consistently fails for `www`. Could you have another look please?

The `unittests` test also consistently fails. You might need to rebase your branch on top of current master to fix it.

J0WI:
Confirmed that the results are now consistent.

PR #17710, opened 2019-04-09
Remove .onion rulesets that pass tests from ruleset whitelist

pipboy96:


pipboy96:
@zoracon Please investigate test failures from Tor not being completely initialized.

src/chrome/content/rules/privacyintyqcroe.onion.xml
```diff
@@ -1,7 +1,7 @@
-<!--
+<!--i
```
J0WI: typo

Issue #17711, opened 2019-04-09
Release 2019.5.2 forthcoming


zoracon:
Type: other

cc @J0WI, @Bisaloo, @cschanaj, @Hainish, @jsha, @andresbase, @pipboy96 

## Planning a release for May ~1st~ 2nd
Delayed: Due to some extra work around broken tests with the new UX branch and a few other loose ends, the release will be done tomorrow, pacific U.S. time. Thanks for the reviews. This a major release with many changes so making sure to release the extension with the least friction possible.

Wanted to get in some patches and UI rehaul (stage 1) for the pop up menu and UX/UI tweaks to EASE interstitial in. Visually this will be a big change. This CSS work I am wrapping up is heavily based on the extensive feedback from #16669 and the consequent work done internally, iterating off of Glenn's work.

## What "stage 1" entails:
- More explanation of each mode
- Element changes to make more visually clear to user what is on/off
- Layout changes for visually better intake of information

## Release Notes (So far):
- UI changes in extension menu (#17854)
- EASE interstitial UI tweaks (#17347)
- Remove support for wildcard in the middle (#12319)
- Update default timestamp for deterministic builds (#17623)
- Refactor and enhance trivialize-cookie-rules.js (#17438)
- Run HSTS-prune and fix impacted rulesets (#17338)
- Update HSTS preload max age (#17564)
-  Fix DeprecationWarning in HTTPS Everywhere Checker  (#17596 )
-  Fix Chromium local store exception (#17557) 
- Remove middle wildcard support in rules.js (#17715)

cschanaj:
It would be great if the next release include #17557 since the there is UI breakage on C73+. It might also worth merging #17596 to get ride of the depreciation warning on Travis. Thanks!

pipboy96:
See also #17656, #17352, #17498, #17674

pipboy96:
@zoracon Did you add privacy policy URL to UI?

pipboy96:
@Hainish @zoracon We still have a GPL2 (not GPL2 or later, thus incompatible with GPL3) file in our repo: https://github.com/EFForg/https-everywhere/blob/master/test/rules/src/https_everywhere_checker/gvgen.py. Is GraphViz generation feature important/still used? Can we remove it safely?

J0WI:
What is the state of #12776 and #17490?

pipboy96:
@J0WI https://github.com/EFForg/https-everywhere/pull/17490 does not pass tests.

[...]

PR #17715, opened 2019-04-12
Remove middle wildcard support in rules.js

cschanaj:
This replaces #17669 to drop support for middle wildcards in the runtime. hopefully #17711

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -562,10 +562,10 @@ RuleSets.prototype = {
       return nullIterable;
     }
 
-    // Replace www.example.com with *.example.com, however, this
-    // SHOULD NOT replace example.com with *.com
+    // Replace www.example.com with www.example.*
+    // eat away from the right for once and only once
     let segmented = host.split(".");
-    if (segmented.length > 2) {
+    if (segmented.length > 1) {
       let tmp = segmented[segmented.length - 1];
 
       results = (this.targets.has(segmented.join(".")) ?
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -566,7 +566,7 @@ RuleSets.prototype = {
     // eat away from the right for once and only once
     let segmented = host.split(".");
     if (segmented.length > 1) {
-      let tmp = segmented[segmented.length - 1];
+      const tmp = segmented[segmented.length - 1];
       segmented[segmented.length - 1] = "*";
 
       results = (this.targets.has(segmented.join(".")) ?
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -567,6 +567,7 @@ RuleSets.prototype = {
     let segmented = host.split(".");
     if (segmented.length > 1) {
       let tmp = segmented[segmented.length - 1];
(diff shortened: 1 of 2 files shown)


pipboy96:
@cschanaj Can you add a test case that checks if middle wildcard causes error and gets ignored as it's supposed to?

pipboy96:
This may not necessarily work, since some rulesets depend on `*.example.com` also matching `a.b.example.com`.

chromium/background-scripts/rules.js
```diff
@@ -562,22 +562,23 @@ RuleSets.prototype = {
...
 
-    // Replace each portion of the domain with a * in turn
+    // Replace www.example.com with www.example.*
+    // eat away from the right for once and only once
     let segmented = host.split(".");
-    for (let i = 0; i < segmented.length; i++) {
-      let tmp = segmented[i];
-      segmented[i] = "*";
+    if (segmented.length > 1) {
+      let tmp = segmented[segmented.length - 1];
```
Hainish: `const` can be used here

chromium/test/rules_test.js
```diff
@@ -203,15 +203,23 @@ describe('rules.js', function() {
           assert.deepEqual(res3, new Set(value), 'wildcard matches sub domains');
         });
 
-        it('matches middle wildcards', function() {
-          let target = 'sub.*.' + host;
+        it('matches right wildcards', function() {
+          let target = host + '.*';
```
Hainish: `const` can be used here

chromium/test/rules_test.js
```diff
@@ -203,15 +203,23 @@ describe('rules.js', function() {
...
           this.rsets.targets.set(target, value);
 
-          let res1 = this.rsets.potentiallyApplicableRulesets('sub.star.' + host);
+          let res1 = this.rsets.potentiallyApplicableRulesets(host + '.tld');
```
Hainish: `const` can be used here

chromium/test/rules_test.js
```diff
@@ -203,15 +203,23 @@ describe('rules.js', function() {
...
           assert.deepEqual(res1, new Set(value), 'default case');
 
-          let res2 = this.rsets.potentiallyApplicableRulesets('sub.foo.bar.' + host);
-          assert.isEmpty(res2, new Set(value), 'only matches one label');
+          let res2 = this.rsets.potentiallyApplicableRulesets(host + '.tld.com');
```
Hainish: `const` can be used here

chromium/test/rules_test.js
```diff
@@ -203,15 +203,23 @@ describe('rules.js', function() {
...
+          assert.isEmpty(res2, 'wildcard matches second level domains');
+        });
+
+        it('ignore middle wildcards', function() {
+          let target = 'www.*.' + host;
```
Hainish: `const` can be used here

chromium/test/rules_test.js
```diff
@@ -203,15 +203,23 @@ describe('rules.js', function() {
...
+          this.rsets.targets.set(target, value);
+
+          let res1 = this.rsets.potentiallyApplicableRulesets('www.cdn.' + host);
```
Hainish: `const` can be used here

PR #17736, opened 2019-04-13
Avoid redundant statement branching in safeToSecureCookie()

cschanaj:
#17711

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -614,7 +614,7 @@ RuleSets.prototype = {
    * @returns {*} ruleset or null
    */
   shouldSecureCookie: function(cookie) {
-    let hostname = cookie.domain;
+    var hostname = cookie.domain;
     // cookie domain scopes can start with .
     while (hostname.charAt(0) == ".") {
       hostname = hostname.slice(1);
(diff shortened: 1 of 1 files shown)


pipboy96:
@cschanaj simply changing `var` to `let` should probably be done as a separate PR.

chromium/background-scripts/rules.js
```diff
@@ -614,7 +614,7 @@ RuleSets.prototype = {
    * @returns {*} ruleset or null
    */
   shouldSecureCookie: function(cookie) {
-    var hostname = cookie.domain;
+    let hostname = cookie.domain;
```
pipboy96: Keep as `var`, this should be done as part of ESLint.

chromium/background-scripts/rules.js
```diff
@@ -624,12 +624,11 @@ RuleSets.prototype = {
       return null;
     }
 
-    var potentiallyApplicable = this.potentiallyApplicableRulesets(hostname);
+    let potentiallyApplicable = this.potentiallyApplicableRulesets(hostname);
```
pipboy96: Keep as `var`, this should be done as part of ESLint.

chromium/background-scripts/rules.js
```diff
@@ -627,9 +627,8 @@ RuleSets.prototype = {
     var potentiallyApplicable = this.potentiallyApplicableRulesets(hostname);
     for (let ruleset of potentiallyApplicable) {
       if (ruleset.cookierules !== null && ruleset.active) {
-        for (let cookierules of ruleset.cookierules) {
-          var cr = cookierules;
-          if (cr.host_c.test(cookie.domain) && cr.name_c.test(cookie.name)) {
+        for (let cookierule of ruleset.cookierules) {
```
Hainish: `const` can be used here

PR #17737, opened 2019-04-14
Remove duplicated reverb.echo.nasa.gov, gone from dns

cschanaj:
Seemingly a bug that `duplicate-whitelist-cleanup.sh` failed to remove this `target`.


src/chrome/content/rules/NASA.gov.xml
```diff
@@ -558,7 +558,6 @@
 	<target host="neba.nasa.gov" />
 	<target host="towerfts.nsbf.nasa.gov" />
 	<target host="*.towerfts.nsbf.nasa.gov" />
-	<target host="nasasearch.nasa.gov" />
```
Bisaloo: `dig @8.8.8.8 nasasearch.nasa.gov` works for me

cschanaj: fixed. thanks!

Bisaloo:
Thank you!

PR #17742, opened 2019-04-16
Do not call safeToSecureCookie until necessary

cschanaj:
In the past, we call safeToSecureCookie for every new cookie regardless we will ever secure the cookie or not. It is more sane to delay the call until we know it is necessary. The proposed change also reduce calls to potentiallyApplicableRulesets to avoid search in a huge (cached) Map object .

IMHO, the code logic is more complicated but I am not sure if it worth the trade off. Given the complexity of this change, I don't think it is necessary to merge it before the next release. 

P.S. This is also likely the last major change I will make before the next release. thanks!!

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -612,38 +612,15 @@ RuleSets.prototype = {
   /**
    * Check to see if the Cookie object c meets any of our cookierule criteria for being marked as secure.
    * @param cookie The cookie to test
-   * @returns {*} ruleset or null
+   * @returns {*} true or false
    */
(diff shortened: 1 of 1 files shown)


pipboy96:
@cschanaj Can you add some unit tests e.g. to verify that safeToSecureCookie is not called when it's not necessary?

cschanaj:
@Hainish @zoracon this PR make improvement to the checking for securing cookies. would you be interested in taking a look?

P.S. I do not have the time to add unit test for it at the moment. Might do it at a later PR.

Hainish:
LGTM, thanks!

PR #17746, opened 2019-04-17
Delay all requests until rewriting is possible

### LINKED ISSUE #17716: HTTPS-E should defer all network activity at the start until it loads up
[//]: #

Type: code issue

when first starting firefox nightly+https-e on a vm with low amount of ram+cpu resources and then loading up a website that has an https everywhere ruleset and that i expected to immediately go to https but was instead loading up over http

noscript webextension does something like that if i remember

pipboy96:
**Extremely experimental. Please test before merging. I did brief testing on Chromium but not Firefox.**

This is supposed to close #17716 and another security vulnerability related on how HTTPS Everywhere is caching rewrites.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -21,7 +21,7 @@ async function initialize() {
   await all_rules.loadFromBrowserStorage(store, update.applyStoredRulesets);
   await incognito.onIncognitoDestruction(destroy_caches);
 }
-initialize();
+const extensionReady = initialize();
 
 async function initializeAllRules() {
   const r = new rules.RuleSets();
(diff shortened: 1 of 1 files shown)


pipboy96:
Reason for letting HTTPS requests through is that `onBeforeRequest` is also catching queries to `https-rulesets.org`.

Hainish:
I'd love to get metrics on how much of a gap in time that we're closing with this PR.  Another problem I can foresee is if someone adds a custom, HTTP endpoint for a ruleset update channel.  Someone may not deem HTTPS necessary since the rulesets are verified with a signing key anyway.  I would argue for the additional protections that HTTPS guarantees, but someone may not think that necessary for the scope of the rulesets they publish.

pipboy96:
@Hainish simply require HTTPS for custom ruleset sources, that's it.

Hainish:
@pipboy96 so http://example.onion/ would be invalid?

Hainish:
Onion services are self-validating - keeping this option open for custom ruleset channels makes sense.

pipboy96:
@Hainish I have added a code that does `performance.now()` at the start of the extension and on `extensionReady` promise, the result was 654 ms on Firefox.

Hainish:
Thanks for the performance impact assessment.  Can you add a clause to the condition which allows `.onion` TLDs?

Hainish:
In local testing I'm getting successes for Chromium, yet Travis is not.  I wonder if this is due to differences in the runtime environment, bandwidth, or something else.

pipboy96:
@Hainish My first guess would be dependency versions.

pipboy96:
rerunning tests

PR #17761, opened 2019-04-19
Create handelsbanken.se.xml

hstseverywhere:


src/chrome/content/rules/handelsbanken.se.xml
```diff
@@ -0,0 +1,57 @@
...
+        - affarsnyttigt
+        - event
+        - aw
+        - www.forskningsstiftelserna
+        - hcm
+        - link01
+        - mob
+        - newsletter
+        - tv
+        - w3c
```
J0WI: `mob.handelsbanken.se` (please use the whole domain for all comments)

J0WI:
Thanks!

PR #17791, opened 2019-04-20
Create icabanken.se.xml

hstseverywhere:


src/chrome/content/rules/icabanken.se.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="icabanken.se">
+	<target host="icabanken.se" />
+	<target host="www.icabanken.se" />
+	<target host="appserver.icabanken.se" />
```
J0WI: `dialog.icabanken.se`

PR #17800, opened 2019-04-21
Update Fastmail.xml

hstseverywhere:
https://hstspreload.org/?domain=fastmail.com


J0WI:
For some reason this has not triggered our `preload` test, but in general we do not duplicate browsers internal HSTS rules:

PR #17806, opened 2019-04-22
Update PayPal.xml

hstseverywhere:
https://hstspreload.org/?domain=paypal.com

### DIFF
--- a/src/chrome/content/rules/PayPal.xml
+++ b/src/chrome/content/rules/PayPal.xml
@@ -9,7 +9,14 @@
 
 		- paypal.112.2o7.net
 		- where-spotlight.s3.amazonaws.com
+		
+	HSTS:
+	
+		- paypal.com 
 -->
+
+<!--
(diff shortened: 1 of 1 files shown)


pipboy96:
I believe PayPal cannot be removed since it's not preloaded in the usual way, also commenting out the ruleset in this way doesn't work.

pipboy96:
cc @J0WI @Bisaloo they should know more.

Bisaloo:
Yes, paypal.com is missing the `includeSubdomains` and the `preload` directives ([source](https://securityheaders.com/?q=paypal.com&followRedirects=on)). It therefore doesn't meet our criteria for removal.

@pipboy96 is right. paypal.com was added manually under the `custom` policy ([source](https://chromium.googlesource.com/chromium/src/+/refs/heads/master/net/http/transport_security_state_static.json#764)). #13682 should fix this.

PR #17809, opened 2019-04-22
Delete Torproject.xml

hstseverywhere:
https://hstspreload.org/?domain=torproject.org


J0WI:
Please do not remove them manually, we have a tool for it that we use before releases:
https://github.com/EFForg/https-everywhere/tree/master/utils/hsts-prune

wesinator:
@J0WI has that tool not been run in a while ?
torproject.org has been preloaded for a few years I think.

PR #17811, opened 2019-04-22
Update Bing.xml

hstseverywhere:
https://hstspreload.org/?domain=bing.com


pipboy96:
Commenting out the entire ruleset is not going to work.

PR #17813, opened 2019-04-22
Update ManilaPrinciples.org.xml

hstseverywhere:
https://hstspreload.org/?domain=manilaprinciples.org


pipboy96:
I can't find any subdomain except `www`. Adding newlines in the comment is also unneccesary. Sorry, this can't be merged.

PR #17815, opened 2019-04-22
EFF domains without HSTS

hstseverywhere:


pipboy96:
Can you attempt to discover subdomains (using Sublist3r and Amass) and separate the ruleset into several files?

src/chrome/content/rules/EFF.xml
```diff
@@ -1,17 +1,26 @@
-<!--
-	Other EFF rulesets:
-		- Freerangekitten.com.xml
-		- IFightSurveillance.org.xml
-		- ManilaPrinciples.org.xml
-		- Panopticlick.com.xml
+<ruleset name="EFF">
 
--->
+	<target host="freerangekitten.com" />
```
cschanaj: IIRC, this domain is intentionally leave unprotected, and reserved like `example.com`. And this domain is also used in the unit tests.

cschanaj: IIRC, this domain is intentionally leave unprotected, and reserved like example.com. And this domain is also used in the unit tests.

J0WI:
Maybe the @EFForg wants to preload them anyway?
@zoracon @Hainish

pipboy96:
@J0WI That would be the actual solution. We still may need to add the ruleset until they will be preloaded in all targets incl. FF ESR.

src/chrome/content/rules/EFF.xml
```diff
@@ -1,17 +1,26 @@
-<!--
-	Other EFF rulesets:
-		- Freerangekitten.com.xml
-		- IFightSurveillance.org.xml
-		- ManilaPrinciples.org.xml
-		- Panopticlick.com.xml
```
J0WI: The targets added in this PR are already covered in these files. We shouldn't merge them back into one file.

src/chrome/content/rules/EFF.xml
J0WI: https://github.com/EFForg/https-everywhere/pull/9095

hstseverywhere:
Sorry I messed up the "thread" I think. Not used to GitHub

pipboy96:
Your PR has been successfully merged (https://github.com/EFForg/https-everywhere/commit/60d65b37427c1680b02d9e2e3241c4e8d2662900).

PR #17819, opened 2019-04-23
Create Fastmail-User.FM.xml

hstseverywhere:


src/chrome/content/rules/Fastmail-User.FM.xml
```diff
@@ -0,0 +1,24 @@
...
+                                  $1                      $2
+-->
+<ruleset name="Fastmail User Websites">
+	<target host="user.fm" />
+	<target host="*.user.fm" />
+		<test url="http://robm.fastmail.fm.user.fm/articles/arthurs_walk.html" />
+		<test url="http://www.robm.fastmail.fm.user.fm/articles/arthurs_walk.html" />
+		<test url="http://robm.fastmail.fm.user.fm/" />
+
+	<rule from="^http://([\.a-zA-Z0-9_-]*)\.user\.fm/([\s\S]*)$" to="https://user.fm/$1/$2" />
```
Bisaloo: This regex seems unnecessarily complex. What about

```xml
	<rule from="^http://([\w.-]+)\.user\.fm/" to="https://user.fm/$1/" />
```

src/chrome/content/rules/Fastmail-User.FM.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+          For other Fastmail coverage, see Fastmail.xml 
+
+
+		  Rewrite rule for user.fm illustrated:
```
Bisaloo: We don't need this explanation with the simplified regex

src/chrome/content/rules/Fastmail-User.FM.xml
```diff
@@ -0,0 +1,24 @@
...
+                 ^^^^^^^^^^^^^^^^         ^^^^^^^^^^^^^^^^^^^^^^^^^^
+                        $1                             $2
+
+      => https://user.fm/www.robm.fastmail.fm/articles/arthurs_walk.html
+                         ^^^^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^^^
+                                  $1                      $2
+-->
+<ruleset name="Fastmail User Websites">
+	<target host="user.fm" />
+	<target host="*.user.fm" />
```
Bisaloo: Please add a comment explaining why this wildcard is necessary.

src/chrome/content/rules/Fastmail-User.FM.xml
```diff
@@ -0,0 +1,24 @@
+<!--
+          For other Fastmail coverage, see Fastmail.xml
```
Bisaloo: Please also add a comment in `Fastmail.xml` pointing to this ruleset.

src/chrome/content/rules/Fastmail-User.FM.xml
```diff
@@ -0,0 +1,24 @@
...
+                        $1                             $2
+
+      => https://user.fm/www.robm.fastmail.fm/articles/arthurs_walk.html
+                         ^^^^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^^^
+                                  $1                      $2
+-->
+<ruleset name="Fastmail User Websites">
+	<target host="user.fm" />
+	<target host="*.user.fm" />
+		<test url="http://robm.fastmail.fm.user.fm/articles/arthurs_walk.html" />
```
Bisaloo: Are all targets subdomains of `fm.user.fm`? If so, it would probably be better to put the wildcard on this domain.

hstseverywhere: No. That is from the tld of the hosted website. If www.example.com was hosted there, it would also be available as http://www.example.com.user.fm and https://user.fm/www.example.com 

A few examples from Sublist3r:

www.rjj.id.au.user.fm
www.hairtodyefor.biz.user.fm
www.3karpfen.ch.user.fm
www.bilikfamily.com.user.fm
telescope.cslwin.fastmail.com.user.fm
antelopeweb.fmail.co.uk.user.fm

src/chrome/content/rules/Fastmail-User.FM.xml
hstseverywhere: Does this added comment clear it up?

"Fastmail user websites
		If www.example.com was hosted on Fastmail, it would also be available 
		as http://www.example.com.user.fm/ and https://user.fm/www.example.com/ "

src/chrome/content/rules/Fastmail-User.FM.xml
hstseverywhere: Fastmail user websites
		If www.example.com was hosted on Fastmail, it would also be available 
		as http://www.example.com.user.fm/ and https://user.fm/www.example.com/

Bisaloo:
I tested more targets and it looks like this rewrite breaks relative paths. See for example http://pjohansson.fastmail.com.user.fm.

I don't how this can be fixed on our side so this PR should probably be closed.

PR #17831, opened 2019-04-25
Add Calend.ru.xml

pipboy96:


src/chrome/content/rules/Calend.ru.xml
```diff
@@ -0,0 +1,25 @@
+<!--
+	Mismatched:
+		- amp
+		- bbs
+		- ilserver
+		- mail
+		- sub
+
```
J0WI: wildcard DNS

pipboy96: @J0WI Done.

PR #17840, opened 2019-04-27
Update Debian.net.xml

J0WI:


src/chrome/content/rules/Debian.net.xml
```diff
@@ -1,39 +1,24 @@
 <!--
 	For other Debian coverage, see Debian.xml.
 
+	Mismatched:
+		- debain.net
```
Bisaloo: debian.net (typo in domain name)

J0WI: Thanks, fixed.

PR #17842, opened 2019-04-28
WIP: Add script to trivialize some left-wilcard targets

cschanaj:
This script will trivialize at least 1.3k left-wildcard targets (nearly 1/3 of their population). In some sense, it seems too good to be true, hope I didn't mess up anything :smile: 

P.S. benchmark with memusage.js show the memory usage will be increased by 0.6 MB because of this change... if this is not justified, I will close this PR.

### DIFF
--- a/utils/trivialize-rules/trivialize-targets.js
+++ b/utils/trivialize-rules/trivialize-targets.js
@@ -1,0 +1,162 @@
+"use strict";
+
+let util = require("util");
+let path = require("path");
+let xml2js = require("xml2js");
+
+let fs = require("graceful-fs");
+let readdir = util.promisify(fs.readdir);
+let readFile = util.promisify(fs.readFile);
+let parseString = util.promisify(xml2js.parseString);
+
+let chalk = require("chalk");
+let validUrl = require("valid-url");
+let escapeStringRegexp = require("escape-string-regexp");
+let { explodeRegExp, UnsupportedRegExp } = require("./explode-regexp");
+
+let rulesDir = "src/chrome/content/rules";
+
+(async () => {
+  let filenames = (await readdir(rulesDir)).filter(fn => fn.endsWith(".xml"));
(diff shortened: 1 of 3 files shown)


pipboy96:
@cschanaj This is justified since it reduces ruleset complexity and improves ruleset testing coverage, which are currently very important tasks.

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,162 @@
...
+    // note that "trivial" does not mean a trivial rule here
+    function isTrivialRewrite(rule) {
+      let retval = true;
+
+      try {
+        explodeRegExp(rule.from, url => {
+          if (validUrl.isUri(url)) {
+            let domain = new URL(url).hostname;
+            rewriteDomains.add(domain);
+            retval &= true;
```
pipboy96: This is a no-op since any of `[true, false]` when ran through `a => a && true` will stay identical.

pipboy96:
Also, the actual `package.json` is missing, even though `package-lock.json` is there.

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,162 @@
...
+          } else {
+            retval = false;
+          }
+        });
+      } catch (e) {
+        // assume it is a non-trivial rewrite here
```
pipboy96: Can you add a logger here and see if this gets executed, and what triggers it?

pipboy96:
Also makes sense to replace rules that are effectively equal to `^http:` :arrow_right: `https:` after these changes. Will try to implement that.

cschanaj:
> Also makes sense to replace rules that are effectively equal to `^http:`  `https:` after these changes. Will try to implement that.

AFAIK, `trivialize-rules.js` already do part of this. It might be helpful if you would like to implement that yourself.

PR #17844, opened 2019-04-29
Run HSTS-prune and update references

Bisaloo:


Bisaloo:
- `despora.de` send two HSTS headers and one is missing directives for preloading
- `edas.info` does not consistently send the `includeSubdomains` and `preload` directives

PR #17850, opened 2019-04-29
Add Xwiki.org.xml

J0WI:


src/chrome/content/rules/Xwiki.org.xml
```diff
@@ -0,0 +1,29 @@
+<!--
+	Plaintext response:
```
cschanaj:
```bash
$ curl https://lists.xwiki.org/mailman/listinfo
curl: (35) error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protoco
```

I get SSL errors

J0WI: curl fails to perform a SSL/TLS handshake. You can verify the plaintext response by forcing http on port 443: http://lists.xwiki.org:443/

cschanaj: Yes. I can reproduce the plaintext response that SSL/TLS is not performed over port 443. thanks for explaining!

PR #17853, opened 2019-04-30
Add git reference to OOB ruleset releases, close #17641

### LINKED ISSUE #17641: How can I find out a Git commit corresponding to the out-of-band ruleset update?
E.g. how can I know which state of the tree the update is based on? I would like to create a script that automatically picks up new updates and tries to reproduce them from corresponding Git commits, and if they do not match, notify me so I can contact someone at EFF.

cschanaj:
I haven't test this, but it should not break things too crazily.

### DIFF
--- a/utils/sign-rulesets/async-request.sh
+++ b/utils/sign-rulesets/async-request.sh
@@ -15,7 +15,8 @@ trap 'rm $SIGNED_SHA256SUM_BASE64' EXIT
 
 mkdir -p $2
 TIMESTAMP=`date +%s`
-echo "{ \"timestamp\": $TIMESTAMP, \"rulesets\":" "`cat $RULESETS_FILE`" "}" | tr -d '\n' | gzip -nc > $2/default.rulesets.$TIMESTAMP.gz
+REFERENCE=`git rev-parse HEAD`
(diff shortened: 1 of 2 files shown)


utils/sign-rulesets/async-request.sh
```diff
@@ -15,7 +15,8 @@ trap 'rm $SIGNED_SHA256SUM_BASE64' EXIT
 
 mkdir -p $2
 TIMESTAMP=`date +%s`
-echo "{ \"timestamp\": $TIMESTAMP, \"rulesets\":" "`cat $RULESETS_FILE`" "}" | tr -d '\n' | gzip -nc > $2/default.rulesets.$TIMESTAMP.gz
+REFERENCE=`git rev-parse HEAD`
+echo "{ \"timestamp\": $TIMESTAMP, \"reference\": \"$REFERENCE\", \"rulesets\":" "`cat $RULESETS_FILE`" "}" | tr -d '\n' | gzip -nc > $2/default.rulesets.$TIMESTAMP.gz
```
zoracon: Are the escaped quotes around `$REFERENCE` needed?

cschanaj: The time stamp do not need an escape quote because it is an integer which is natively supported by the JSON format. The reference is an hex string, so I think it need an escaped quote.

Hainish: Looks good, please add this to `standalone.sh` as well and it'll be ready to merge.

cschanaj: done. thanks!

PR #17855, opened 2019-05-01
Add script to trivialize some left-wildcard targets

cschanaj:
this supersedes #17842 and is ready for review and merging. thanks.

P.S. there is no need to rush for the major release in #17711 .

cc @pipboy96 @zoracon @Hainish

### DIFF
--- a/utils/trivialize-rules/trivialize-targets.js
+++ b/utils/trivialize-rules/trivialize-targets.js
@@ -189,7 +189,7 @@ let rulesDir = "src/chrome/content/rules";
         process.exit(1);
       }
 
-      [, indent ] = matches;
+      [, indent] = matches;
       let sub = value.map(v => `\n${indent}<target host=\"${v}\" />`).join("") + "\n";
       content = content.replace(regex, sub);
     });
@@ -234,7 +234,7 @@ let rulesDir = "src/chrome/content/rules";
           process.exit(1);
         }
 
-        [, indent ] = matches;
+        [, indent] = matches;
         content = content.replace(regex, "\n");
       }
     }
--- a/utils/trivialize-rules/trivialize-targets.js
+++ b/utils/trivialize-rules/trivialize-targets.js
@@ -1,0 +1,253 @@
+"use strict";
+
+let util = require("util");
+let path = require("path");
+let xml2js = require("xml2js");
+
+let fs = require("graceful-fs");
+let readdir = util.promisify(fs.readdir);
+let readFile = util.promisify(fs.readFile);
+let parseString = util.promisify(xml2js.parseString);
+
+let chalk = require("chalk");
+let validUrl = require("valid-url");
+let escapeStringRegexp = require("escape-string-regexp");
+let { explodeRegExp, UnsupportedRegExp } = require("./explode-regexp");
+
+let rulesDir = "src/chrome/content/rules";
+
+(async () => {
+  let filenames = (await readdir(rulesDir)).filter(fn => fn.endsWith(".xml"));
+  let filePromises = filenames.map(async filename => {
+    function createTag(tagName, colour, print) {
+      return (strings, ...values) => {
+        let result = `[${tagName}] ${chalk.bold(filename)}: ${strings[0]}`;
+        for (let i = 1; i < strings.length; i++) {
+          let value = values[i - 1];
+          if (value instanceof Set) {
+            value = Array.from(value);
+          }
+          value = Array.isArray(value) ? value.join(", ") : value.toString();
+          result += chalk.blue(value) + strings[i];
+        }
+        print(colour(result));
+      };
+    }
+
+    let warn = createTag("WARN", chalk.yellow, console.warn);
+    let info = createTag("INFO", chalk.green, console.info);
+    let fail = createTag("FAIL", chalk.red, console.error);
+
+    let content = await readFile(path.join(rulesDir, filename), "utf8");
+    let { ruleset } = await parseString(content);
+
+    let rules = ruleset.rule.map(rule => rule.$);
+    let targets = ruleset.target.map(target => target.$.host);
+
+    // make sure ruleset contains at least one left wildcard targets
+    if (!targets.some(target => target.includes("*."))) {
+      return;
+    }
+
+    // but does not contain right widcard targets
+    if (targets.some(target => target.includes(".*"))) {
+      return;
+    }
+
+    // should not contain trivial rules
+    if (rules.some(rule => rule.from === "^http:" && rule.to === "https:")) {
+      return;
+    }
+
+    let ruleToIsSimpleMap = new Map();
+    let ruleToIsSnappingMap = new Map();
+
+    let targetToSupportedExplodedDomainsMap = new Map();
+
+    let explodedDomains = new Set();
+    let unsupportedExplodedDomains = new Set();
+    let unusedTargets = new Set();
+
+    // (1) We check if all rules can be exploded to valid urls
+    //     (a) if true, continue to (2)
+    //     (b) we cannot trivial targets for this ruleset, skipping
+    function isExplosiveRewrite(rule) {
+      let explodedUrls = new Array();
+
+      try {
+        explodeRegExp(rule.from, url => explodedUrls.push(url));
+      } catch (e) {
+        if (!(e instanceof UnsupportedRegExp)) {
+          fail(e.message);
+        }
+        return false;
+      }
+
+      // make sure each exploded URL is valid
+      if (!explodedUrls.every(url => validUrl.isUri(url))) {
+        return false;
+      }
+
+      let isSimpleToAllExplodedUrls = true;
+      let isSnappingToAllExplodedUrls = false;
+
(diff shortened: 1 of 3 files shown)


utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+let rulesDir = "src/chrome/content/rules";
+
+(async () => {
+  let filenames = (await readdir(rulesDir)).filter(fn => fn.endsWith(".xml"));
+  let filePromises = filenames.map(async filename => {
+    function createTag(tagName, colour, print) {
+      return (strings, ...values) => {
+        let result = `[${tagName}] ${chalk.bold(filename)}: ${strings[0]}`;
+        for (let i = 1; i < strings.length; i++) {
+          let value = values[i - 1];
```
pipboy96:
```
        for (let i = 0; i < strings.length - 1; i++) {
          let value = values[i];
```

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+          if (value instanceof Set) {
+            value = Array.from(value);
+          }
+          value = Array.isArray(value) ? value.join(", ") : value.toString();
+          result += chalk.blue(value) + strings[i];
```
pipboy96:
```
          result += chalk.blue(value) + strings[i - 1];
```

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+    // (5) Check if we can trivialize the rules. Need to satisfy all below conditions:
+    //     i)   there is no unused target, i.e. unusedTargets.size == 0
+    //     ii)  every rule is "simple" as in ruleToIsSimpleMap
+    //     iii) at least one rule is a non-snapping rule
+    //
+    //     (a) if all of the conditions are met, append a trivial rule after all
+    //         existing rule; and remove the non-snapping rules.
+    //     (b) else, do not trivialize the rules
+    let condition1 = (unusedTargets.size == 0);
+    let condition2 = [...ruleToIsSimpleMap.entries()].every(([,value]) => value === true);
```
pipboy96: `([,value]) => value`

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+    let condition3 = [...ruleToIsSnappingMap.entries()].some(([,value]) => value === false);
```
pipboy96: `([,value]) => !value`

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+    let fail = createTag("FAIL", chalk.red, console.error);
+
+    let content = await readFile(path.join(rulesDir, filename), "utf8");
+    let { ruleset } = await parseString(content);
+
+    let rules = ruleset.rule.map(rule => rule.$);
+    let targets = ruleset.target.map(target => target.$.host);
+
+    // make sure ruleset contains at least one left wildcard targets
+    if (!targets.some(target => target.includes("*."))) {
```
pipboy96: At this point this may be replaced with `target.startsWith("*.")`.

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+      return;
+    }
+
+    // but does not contain right widcard targets
+    if (targets.some(target => target.includes(".*"))) {
```
pipboy96: `target.endsWith(".*")`

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+
+      let isSimpleToAllExplodedUrls = true;
+      let isSnappingToAllExplodedUrls = false;
+
+      for (let url of explodedUrls) {
+        let { protocol, hostname, pathname } = new URL(url);
+
+        // if a rule do not rewrite all path for any URL, it is not a simple rule
+        // i.e. a rule is simple only if it rewrite all path for all URLs
+        if (!(protocol == "http:" && pathname == "/*")) {
```
pipboy96: `==` :-1:
'===' :sunglasses:

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+          isSimpleToAllExplodedUrls = false;
+        }
+
+        // if a rule is snapping to any URL, it is a snapping rule
+        if (url.replace(new RegExp(rule.from), rule.to) !== url.replace(/^http:/, "https:")) {
+          isSnappingToAllExplodedUrls = true;
+        }
```
pipboy96: Are you trying to test here if domain name is changed?

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+      }
+      unsupportedExplodedDomains.add(domain);
+      return false;
+    }
+
+    // assume each target support no exploded domain initially
+    targets.forEach(target => targetToSupportedExplodedDomainsMap.set(target, []));
+    if (![...explodedDomains].every(domain => isSupported(domain))) {
+      warn`ruleset rewrites domains ${[...unsupportedExplodedDomains]} unsupported by ${targets}`;
+      return ;
```
pipboy96: `return;`

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,252 @@
...
+    //     (a) if true, continue to (2)
+    //     (b) we cannot trivial targets for this ruleset, skipping
+    function isExplosiveRewrite(rule) {
+      let explodedUrls = new Array();
+
+      try {
+        explodeRegExp(rule.from, url => explodedUrls.push(url));
+      } catch (e) {
+        if (!(e instanceof UnsupportedRegExp)) {
+          fail(e.message);
```
pipboy96: Shouldn't you interrupt execution on error that is not `UnsupportedRegExp`?

utils/trivialize-rules/trivialize-targets.js
cschanaj: in fact, this part of code is copied from `trivialize-rules.js`. I believe the index is correct because `strings[0]` is already used above while `strings[i]` and `values[i - 1]` works like key-value pairs.

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,256 @@
...
+        let regex = new RegExp(regexSource);
+
+        let matches = content.match(regex);
+        if (!matches) {
+          // should be unreachable.
+          warn`unexpected regular expression error`;
+          process.exit(1);
+        }
+
+        [, indent ] = matches;
```
pipboy96: `[, indent]` (no need of ` `).

utils/trivialize-rules/trivialize-targets.js
```diff
@@ -0,0 +1,256 @@
...
+(async () => {
+  let filenames = (await readdir(rulesDir)).filter(fn => fn.endsWith(".xml"));
+  let filePromises = filenames.map(async filename => {
+    function createTag(tagName, colour, print) {
+      return (strings, ...values) => {
+        let result = `[${tagName}] ${chalk.bold(filename)}: ${strings[0]}`;
+        for (let i = 1; i < strings.length; i++) {
+          let value = values[i - 1];
+          if (value instanceof Set) {
+            value = Array.from(value);
```
pipboy96: `value = [...value]` has the same result.

cschanaj: I would like to keep this part of code somewhat sync-ed with `trivialize-rules.js`. If they are functionally equivalent, I am not going to change it. but thanks anyway :smile:

pipboy96: @cschanaj Can you move this bit into a library?

pipboy96:
@cschanaj Is there a reason to prefer `"` over `'`? Same issue exists in the extension itself. Both are used without any consistency.

utils/trivialize-rules/trivialize-targets.js
cschanaj: this function works only within the closure. if we move it into a library, I am afraid we need to change the function prototype. I would like to avoid making the unrelated changes given the logics of the script in this PR is already very complicated.

pipboy96: @cschanaj Thanks for explaining.

cschanaj:
this directory use `prettier` as the preferred linter (a choice from its original author). I just follow his practices.

cschanaj:
@Hainish would you be interested in looking into this, possibly before the next ruleset release?

Hainish:
I've reviewed and made a few changes here: https://github.com/EFForg/https-everywhere/pull/18062

@cschanaj please review my changes and make sure the clarifications are actually accurate.  Otherwise, good to merge.  Great work, and thanks for this!

PR #17877, opened 2019-05-03
Fix typo causing abccommercial.com unprotected

cschanaj:
there was a typo in the complicated rule for `abccommercial.com`

### DIFF
--- a/src/chrome/content/rules/ABC-Online.xml
+++ b/src/chrome/content/rules/ABC-Online.xml
@@ -5,12 +5,12 @@
 
 	Non-functional hosts in *.abc.net.au
 		Timeout:
-		- abc.net.au
 		- shop.abc.net.au
 -->
 <ruleset name="ABC Online (partial)">
 
 	<!-- *.abc.net.au -->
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/ABC-Online.xml
```diff
@@ -3,48 +3,21 @@
...
-		- (www.)radioaustralianews.net.au ³
-		- (www.)triplejunearthed.com ³
-
-	¹ 503, akamai
-	² 504, akamai
-	³ Dropped
-
+	Non-functional hosts in *.abc.net.au
+		Timeout:
+		- abc.net.au
```
Bisaloo: This target works fine for me. Could you double check please?

Bisaloo: If travis is the only issue, I think we should whitelist this ruleset. The target looks fine on ssllabs: https://www.ssllabs.com/ssltest/analyze.html?d=abc.net.au

cschanaj:
Travis pass. thanks.

PR #17878, opened 2019-05-03
Update CadSoft.xml

cschanaj:
there is no rule for `*.www.cadsoft.com`, cookie cannot be secured this way.

### DIFF
--- a/src/chrome/content/rules/CadSoft.xml
+++ b/src/chrome/content/rules/CadSoft.xml
@@ -1,16 +1,13 @@
+<!--
+	Non-functional hosts:
+		Mismatch:
+		- cadsoft.de
+-->
 <ruleset name="CadSoft">
 
 	<target host="cadsoft.de" />
 	<target host="www.cadsoft.de" />
-	<!--	* for cross-domain cookie.	-->
-	<target host="*.www.cadsoft.de" />
 
-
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/CadSoft.xml
```diff
@@ -1,16 +1,13 @@
...
-	<!--	* for cross-domain cookie.	-->
-	<target host="*.www.cadsoft.de" />
 
-
-	<securecookie host="^\.?www\.cadsoft\.de$" name=".+" />
-
-
-	<!--	Cert only matches www.
-					-->
 	<rule from="^http://(?:www\.)?cadsoft\.de/"
```
Bisaloo: Could you update this to use a catching group instead, as recommended in the contributing guide, please?

```suggestion
	<rule from="^http://(www\.)?cadsoft\.de/"
```

src/chrome/content/rules/CadSoft.xml
```diff
@@ -1,16 +1,13 @@
+<!--
+	Non-functional hosts:
+		Mismatch:
+		- cadsoft.de
+-->
 <ruleset name="CadSoft">
```
Bisaloo: Please update the ruleset name to `CadSoft.de`

```suggestion
<ruleset name="CadSoft.de">
```

cschanaj:
Travis pass. Thanks!!

PR #17887, opened 2019-05-04
Update Indiana_State_University.xml

cschanaj:
`^` was not supported by any rewrite.  #17965

### DIFF
--- a/src/chrome/content/rules/Indiana_State_University.xml
+++ b/src/chrome/content/rules/Indiana_State_University.xml
@@ -6,7 +6,7 @@
 		Mixed content blocking (MCB) triggered:
 		- library.indstate.edu
 -->
-<ruleset name="Indiana State University">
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Indiana_State_University.xml
```diff
@@ -1,30 +1,16 @@
...
-	<target host="*.indstate.edu" />
-		<exclusion pattern="^http://(?:www\.)?indstate\.edu/(?!css/|favicon\.ico|images/|js/)" />
-
-
-	<securecookie host="^isuportal\.indstate\.edu$" name=".+" />
-
-
-	<rule from="^http://((?:isuportal|www1?)\.)?indstate\.edu/"
-		to="https://$1indstate.edu/" />
+<ruleset name="Indiana State University">
```
J0WI: Still partial, isn't it?

PR #17889, opened 2019-05-04
Fix typo causing *.nexway.com unprotected

cschanaj:


src/chrome/content/rules/Nexway.com.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="Nexaway.com">
```
J0WI: Would you like to add more targets or keep it `(partial)`?

cschanaj: I would prefer to keep it `partial` since my intention is to fix the typo here.

Issue #17434, opened 2019-02-11
Add ability to disable HTTPS Everywhere for specific domains without first visiting the site


raman325:
Type: feature request

Hi, I've run into some issues using HTTPS Everywhere while on sites on my corporate networks, or on sites that are "public" but are proxied via Microsoft Application Proxy to our internal network. I'm not sure why this is creating problems (in this case it was our JIRA instance), but in trying to get around it, I've noticed that I can only disable HTTPS Everywhere for a site if I am on the site and disable it through the extensions right click menu.

It would be great if I could explicitly whitelist domains that HTTPS Everywhere would be disabled for without having to do it for every site as needed (in my case, I'd disable it for my corporate domains). I think it would make sense to add this functionality into the extension settings, maybe even under "Advanced".

Thanks for your consideration!

pipboy96:
I'll attempt to implement this.

bershanskiy:
@pipboy96 Any update on this? If you are not working on this, may I pick up this issue?

pipboy96:
@bershanskiy No, it's fine.

bershanskiy:
#17890

PR #17890, opened 2019-05-04
Add ability to disable HTTPS Everywhere for specific domains without first visiting the site (#17434)

bershanskiy:
# WORK IN PROGRESS! DO NOT MERGE!

This allows user to add a domain name to the whitelist from within Options menu (without visiting the site). No domain validation is performed.

This is an incredibly simplistic implementation (e.g., no domain validation or Punycode encoding or any CSS work) so please tell me which things I should change.

### Preview

eff

I decided to put the form under the disabled list title on "General" page because it seemed more natural than putting it on "Advanced" page.

#### Domain names validation
What is the preferred validation method? Internally, HTTPS Everywhere seems to prefer IDN in Punycode and does not allow wild cards. Should I enforce this and display error messages? What 

#### Preferred code style
1. There are a few variables declared as `let` even though they can be declared as `const`. Is this intentional?
2. Variable names are not always consistent (some use underscores, some are camel-cased). What's preferred?
3. There are many tiny refactoring things that can be done in the Options UI. I omitted them from this commit for now for simplicity sake.

CC @raman325 @pipboy96


pipboy96:
@bershanskiy Use camelCase and `const` where possible.

pipboy96:
It would make sense to allow left-side wildcards, such as `*.github.io`.

bershanskiy:
@pipboy96 Thanks for the prompt reply.

> Use camelCase and const where possible.

Will do.

> It would make sense to allow left-side wildcards, such as *.github.io.

Currently it allows wildcards, as well as anything else you put in the box. In fact, there is no validation performed at all (which is definitely bad). Could you write-up exact criteria for input validation?

I understand it as follows:
 - fully-qualified domain names like `github.com`; IDN should be converted to Punycode, but invalid Punycode (domain labels that start with `xn--` but are valid domain name and not valid Punycode) are allowed.
 - left-hand side wildcards like `*.github.io`
 - all mentioned above but with a port number, e.g. `github.io:1000` or `*.github.io:1000`

I did not make sure the background scripts support wildcards yet. But I'll definitely do that before merging.

pipboy96:
@bershanskiy I will write a JS function to validate domains.

pipboy96:
I think port numbers are not allowed.

bershanskiy:
@pipboy96 
> I will write a JS function to validate domains.

Great. Will you also implement the error message display for invalid entries or should I do that?

> I think port numbers are not allowed.

Could you double-check? The popup UI [`disableOnSite()` and `enableOnSite()`](https://github.com/EFForg/https-everywhere/blob/master/chromium/pages/popup/ux.js#L361) explicitly use [`url.host`](https://developer.mozilla.org/en-US/docs/Web/API/URL/host) and not [`url.hostname`](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname). Also, when I open a site with a non-default port and disable HTTPS-E on it, the Popup does display the port.

Is there anything I should do to advance this PR besides changing to camelCase  and `const`?

pipboy96:
@bershanskiy Generally, HTTPS Everywhere doesn't seem to handle custom ports well, e.g. https://example.com and https://example.com:443 are treated like different origins.

pipboy96:
You would also need to `.trim` and `.toLowerCase()` the domain.

pipboy96:
Probably some kind of punycode encoder/decoder would also be needed.

pipboy96:
This is code to encode/decode punycode: https://gist.github.com/pipboy96/cd3b89bf9598e90c0c2dda4e9fe9c286

pipboy96:
Regex: `/^(\*|[a-z0-9_-]+)(\.[a-z0-9_-]+)*$/`.

pipboy96:
I recommend you to show IDNs in this way: xn--80acde.xn--j1aef (абвг.ком) but obviously store in encoded form.

bershanskiy:
@pipboy96 
> I recommend you to show IDNs in this way: xn--80acde.xn--j1aef (абвг.ком) but obviously store in encoded form.

Will do.

Also, could you go over the `validateDomain`:
```
  function validateDomain(domain) {
    // Checks whether input is a valid domain and returns it in a canonocal form
    // TODO: should apply Punycode before or after toLowerCase()?
    // TODO: what if toASCII errors out?
    domain = domain.trim();
    try {
      domain = toASCII(domain);
    } catch(e) {
      // This domain is not representable as Punycode
      return null
    }
    domain = domain.toLowerCase();
    const pattern = /^(\*|[a-z0-9_-]+)(\.[a-z0-9_-]+)*$/
    const match = pattern.test(domain);
    if (match !== true){
      domain = null;
    }
    return domain;
  }
```

What is the proper order of `trim()`, `toLowerCase()`, and `toASCII(...)`? I guess `trim()` goes first to remove the obviously unneeded spaces, but what's next `toLowerCase()` or `toASCII(...)`. I believe the IDN codepoint tables are designed so that `toLowerCase()` is a noop, but I do not have a solid source for that.

The linter does not see the definition of Punycode functions:
```
  291:25  error  'toUnicode' is not defined          no-undef
  328:16  error  'toASCII' is not defined            no-undef
```
May be you have an idea where to add the proper includes/ exclusions?

pipboy96:
toLowerCase may do weird things for non-ASCII chars depending on OS locale. Do it after toASCII.

bershanskiy:
@pipboy96 
> toLowerCase may do weird things for non-ASCII chars depending on OS locale. Do it after toASCII.

That's what I was thinking too.
But is it possible that Punycode is different depending on capitalization of the letters in the domain? I think Punycode automatically lowers the case, e.g. Punycode of `абвг.ком` and `АБвг.ком` is `xn--80acde.xn--j1aef`, but Unicode of `xn--80acde.xn--j1aef` is `абвг.ком`. I don'thave time to work on that today, so I'll look it up tomorrow.

pipboy96:
Should remove linter error.

pipboy96:
I was going to lint anyway, but thanks!

chromium/pages/options/index.html
```diff
@@ -4,12 +4,21 @@
     <meta charset="utf-8">
     <title></title>
     <link href="style.css" rel="stylesheet">
+    <script src="../util.js" defer></script>
+    <script src="../punycode.js" defer></script>
+    <script src="ux.js" defer></script>
+    <script src="../translation.js" defer></script>
```
pipboy96: `async`?

pipboy96: Also move to end of `<body>`.

bershanskiy:
> This (6293476, 0cda424) should remove linter error.

Linter was also complaining about [indentation](https://travis-ci.org/EFForg/https-everywhere/jobs/528353620), so I changed that. However, Travis still complains about "preload", which should not be impacted by this PR, since it does not change any rules...

pipboy96:
@bershanskiy hsts-prune needs `npm audit` to be run.

chromium/pages/options/index.html
bershanskiy: I did not use `async` because scripts have dependencies on each other. `ux.js` and `translation.js` can be made into `async` because they wait for "DOMContentLoaded" anyway, but they do need `util.js` and `punycode.js` to be loaded in advance. I think I can make `translation.js` into `async`, but not the others.

bershanskiy: I [moved](https://github.com/EFForg/https-everywhere/pull/17890/commits/395bb05414aa5209604a7994ad0f860c27fa90de) those to the end of `<body>` and made `translation.js` `async`, but `ux.js` still needs `util.js` and `punycode.js` to run before it so I mad them `defer` to preserve order.

pipboy96:
@bershanskiy Can you fix merge conflicts?

bershanskiy:
@pipboy96 I resolved conflicts on this one via the GitHub online interface as well, but I'll test it for real too. I'll comment when I'm done.

pipboy96:
@bershanskiy Do you have any plans to continue work on this?

zoracon:
Closing out for now, can reopen at request of the author.

PR #17899, opened 2019-05-05
Add LuaJIT.org.xml

pipboy96:


src/chrome/content/rules/LuaJIT.org.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+	luajit.org has a wildcard DNS record, all other subdomains are mismatched.
+
+	Incomplete certificate chain:
+		- wiki
```
J0WI: I got a mismatch

J0WI:
Ref #15750

src/chrome/content/rules/LuaJIT.org.xml
pipboy96: https://whatsmychaincert.com/?wiki.luajit.org

J0WI: Both is true. It's an incomplete, mismatching cert.

pipboy96: ✅

PR #17907, opened 2019-05-05
Run npm audit fix on hsts-prune

cschanaj:
cc @Hainish @zoracon


pipboy96:
This sometimes causes `preloaded` test to fail.

PR #17909, opened 2019-05-06
Update AkademikerForsakring.se.xml (remove last non-punycode unicode domain)

pipboy96:

PR #17910, opened 2019-05-06
Update schema.xml

pipboy96:
Follow-up to #17909.

### DIFF
--- a/test/validations/relaxng/schema.xml
+++ b/test/validations/relaxng/schema.xml
@@ -29,7 +29,7 @@
       <element name="target">
         <attribute name="host">
           <data type="string">
-            <param name="pattern">(\*\.)?([a-z0-9äö_-]+\.)+([a-z0-9äö-]+|\*)</param>
+            <param name="pattern">(\*\.)?([a-z0-9_-]+\.)+([a-z0-9-]+|\*)</param>
           </data>
         </attribute>
       </element>


Bisaloo:
What's the reasoning behind disallowing non-ASCII characters? Are they causing any issue?

pipboy96:
The last target that wasn't converted to Punycode was fixed. Also, I believe any target with these characters would be no-op, since HTTPS Everywhere doesn't decode Punycode.

cschanaj:
> The last target that wasn't converted to Punycode was fixed. Also, I believe any target with these characters would be no-op, since HTTPS Everywhere doesn't decode Punycode.

I got the following extension log by pasting akademikerförsäkring.se into Chrome. It seems that the URL is Punycoded automatically before being passed to the extension. So I think they are not no-op. 

```
Ruleset cache hit for www.xn--akademikerfrskring-xtb17a.se items:0
util.js:26 onBeforeRequest returning shouldCancel: false
util.js:26 Ruleset cache hit for www.akademikerforsakring.se items:1
util.js:26 onBeforeRequest returning shouldCancel: false
util.js:26 Ruleset cache hit for cdn.optimizely.com items:1
util.js:26 onBeforeRequest returning shouldCancel: false
util.js:26 Ruleset cache hit for www.akademikerforsakring.se items:1
util.js:26 onBeforeRequest returning shouldCancel: false
util.js:26 Ruleset cache hit for www.akademikerforsakring.se items:1
util.js:26 onBeforeRequest returning shouldCancel: false
```

cschanaj:
> The last target that wasn't converted to Punycode was fixed. Also, I believe any target with these characters would be no-op, since HTTPS Everywhere doesn't decode Punycode.

It seems that they are in fact no-op :thinking:

Hainish:
Thanks!

PR #15532, opened 2018-05-26
Improve popup page performance and reduce memory usage

cschanaj:
ping @Hainish

PR #17917, opened 2019-05-07
Update chromium/pages/popup/ux.js, fix #17916

### LINKED ISSUE #17916: Toggle to enable/ disable rulesets do not work on Chrome/ Firefox
RT cc @Hainish @zoracon

cschanaj:
As a result of the UI redesign, #15532 is partially reverted. 

P.S. Ideally we should have a unittest on this when the UI stabilize.


zoracon:
Thank you for catching this. Will try to get this in as soon as possible.

zoracon:
Also going to write a test for this one

Issue #17912, opened 2019-05-06
Rulesets that rewrite to s3.amazonaws.com will break in September 2020


jsha:
## Remaining rulesets

- [x] Issuu_Aps.xml
- [x] FeedBlitz.xml
- [x] DaWanda.xml
- [x] Filemobile.xml
- [x] Compendium.xml
- [x] Smowtion.xml
- [x] SMF_for_Free.xml
- [x] Scribd.com.xml
- [x] SBNdev.net.xml
- [x] JanRain.xml
- [x] Roster-Teeth.xml
- [x] Wishabi.xml
- [x] IPR_Software.com.xml (default_off)
- [x] Intuit.xml (default_off)
- [x] PR-Newswire.xml (default_off)

### Awaiting merge

- [x] Quizsnack.com.xml (#19580)
- [x] RPXNow.com.xml (#19581)
- [x] KIXEYE.xml (#19583)
- [x] GEBlogs.com.xml (#19584)

## Original issue comment

Per https://forums.aws.amazon.com/ann.jspa?annID=6776,

> Amazon S3 currently supports two request URI styles in all regions: path-style (also known as V1) that includes bucket name in the path of the URI (example: //s3.amazonaws.com/<bucketname>/key), and virtual-hosted style (also known as V2) which uses the bucket name as part of the domain name (example: //<bucketname>.s3.amazonaws.com/key). In our effort to continuously improve customer experience, the path-style naming convention is being retired in favor of virtual-hosted style request format. Customers should update their applications to use the virtual-hosted style request format when making S3 API requests before September 30th, 2020 to avoid any service disruptions. Customers using the AWS SDK can upgrade to the most recent version of the SDK to ensure their applications are using the virtual-hosted style request format. 

In other words, all rulesets that have a `to=` using the s3.amazonaws.com hostname will stop working. We should update all of these before then. Here's the current list of 115 affected rulesets:

```
$ grep to=.*s3.amazonaws.com src/chrome/content/rules/*
3D_Vision_Live.com.xml:		to="https://s3.amazonaws.com/photos.3dvisionlive/$2/images/$1/cached.thumb.square.208.208.0.0" /-->
AboutUs.xml:		to="https://s3.amazonaws.com/au-site-static-assets/" />
ActBlue.xml:		to="https://s3.amazonaws.com/i.actblue.com/" />
AdExcite.xml:		to="https://adexcite.s3.amazonaws.com/" />
[...]

PR #17923, opened 2019-05-08
Update DocumentCloud.xml

pipboy96:
Related: #17912.


pipboy96:
@zoracon to match the previous version of ruleset?

zoracon:
> @zoracon to match the previous version of ruleset?

No just the standard trivial cookie tag, but you already added it in

PR #17928, opened 2019-05-09
Delete NHS.xml

suny000:
As soon


pipboy96:

pipboy96:
Website still supports HTTPS and is not preloaded, can you try fixing the ruleset instead?

Issue #17944, opened 2019-05-10
Patch Release 2019.5.13 forthcoming


zoracon:
Type: release

This will be another patch release addressing the following as priority:
- Translation errors (Fixed with #17942, #17919)
- Stable rules bugs (#17916, Fixed with #17917, #17954)
- Unit test for ruleset toggle (#17952)

Will try to also fix the following:
CSS animation flaw with EASE toggle

cschanaj:
#17907 is a trivial fix on the hsts-prune utility. would you take a look on that?

zoracon:
Release is out

PR #17954, opened 2019-05-13
Fix counter show number of experimental rulesets

cschanaj:
Step to reproduce:

1. Visit https://hk.yahoo.com/ or any site with some experimental rulesets applicable

2. Observe the ruleset counter show wrong number: it shows the number of experimental rulesets instead of the number of all applicable rulesets. 

If possible, I hope this can be merged before #17944 since it disrupts the UX.


zoracon:
I actually had a branch with a fix for this since I noticed as well. But you beat me to it.

PR #17959, opened 2019-05-15
Added ruleset for Equiac.com

FrancisPlaza:


src/chrome/content/rules/Equiac.com.xml
```diff
@@ -0,0 +1,10 @@
+<ruleset name="Equiac.com">
+	<target host="equiac.com" />
+	<target host="www.equiac.com" />
+	<target host="chat.equiac.com" />
```
Bisaloo: This target times out over both HTTP and HTTPS for me. Could you double check please? It looks fine on travis as well so it's probably just me.

FrancisPlaza: @Bisaloo This should be OK now.

Bisaloo: Yes, indeed! Thank you! It wasn't clear to me that you were affiliated with this website.

For you information, we usually recommend that website owners [HSTS preload](https://scotthelme.co.uk/hsts-preloading/) their websites. It offers the same level of protection as HTTPS Everywhere but will benefit all visitors and not just HTTPS Everywhere users.

PR #17964, opened 2019-05-18
Fix typo causing *.uni.lu unprotected

cschanaj:
#17965


J0WI:
Temporary downtime, let's retry later.

Issue #17965, opened 2019-05-18
Fix rulesets with problematic targets/ rewrites 


cschanaj:
The following problematic rulesets contain either unprotected targets/ invalid rewrites found using the script in #17855. These should be fixed if possible.

- [ ] Academia-Press.xml
- [ ] AEMI.xml
- [ ] Alluremedia.xml
- [ ] Argus_Leader.xml
- [ ] ARM.xml
- [ ] Axis_Bank.xml
- [ ] Branson_Zipline.xml
- [ ] BT_Wi-fi.xml
- [ ] Clicksor.com.xml
- [ ] Commission_Junction.xml
- [ ] Coupons-Inc.xml
- [ ] Demand-Media-mismatches.xml
- [ ] Dial_Direct.xml
- [ ] Digiweb.xml
- [ ] DuckCorp.xml
- [ ] E-junkie.xml
- [ ] EmuNewz-Network.xml
- [ ] Entertainment-Consumers-Association.xml
- [ ] Erowid.xml
- [ ] Experian.xml
- [ ] EZTV.xml
- [ ] Factory_Expo_Home_Centers.xml
- [ ] FierceMarkets.xml
- [ ] Flaska.net.xml
- [ ] Florida_Today.xml #17906 
- [ ] FlyersRights.org.xml #17904 
- [ ] FOX-Toolkit.xml
- [ ] Fujitsu.xml
- [ ] Globat.xml
- [ ] Goldkeys.net.xml
- [ ] GTCs-Online.xml
- [ ] gulli.com.xml
- [ ] Halifax.xml
- [ ] Herdict.xml
- [ ] Hi.nl.xml
- [ ] Holidaybreak.xml
- [ ] HurricaneElectric.xml
- [ ] Indiana_State_University.xml #17887 
- [ ] Infused-Systems-mismatches.xml
- [ ] ION-Audio.xml
- [ ] Lohud.com.xml
- [ ] MediaMind.xml
- [ ] MediaSpan.xml
- [ ] Media-Storehouse.xml
- [ ] Monkeys-Audio.xml
- [ ] MSN-mismatches.xml
- [ ] MyLikes.xml
- [ ] NewspaperDirect-mismatches.xml
- [ ] Nowy_BIP.xml
- [ ] NPR.org.xml
- [ ] Online.nl.xml
- [ ] OU.edu.xml
- [ ] Overclockers.com.xml
- [ ] Patriotsuperbowl.us.xml
[...]

PR #17969, opened 2019-05-18
Delete MediaSpan.xml

cschanaj:
`^` and `www` are gone from DNS for both domains. #17965


Hainish:
@J0WI are you looking for an additional reviewer?  If it looks good please feel free to merge.

J0WI:
I wanted to make sure that this is not a temporary downtime.

Hainish:
Ah I see, thanks for clarifying.

PR #17974, opened 2019-05-19
Create miaopai.com.xml

ivysrono:


src/chrome/content/rules/miaopai.com.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="miaopai.com">
+	<target host="miaopai.com" />
```
cschanaj: would you add comments on the incomplete certificate chain?

src/chrome/content/rules/miaopai.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Mismatched:
```
cschanaj: I got expired certificate: see https://www.ssllabs.com/ssltest/analyze.html?d=miaopai.com . It could be the server behavior differently on different region, though.

PR #18000, opened 2019-05-25
Update and rename Duolingo.com.xml to Duolingo.xml

ivysrono:


pipboy96:
I think there may be more subdomains, please use Amass to get them.

src/chrome/content/rules/Duolingo.xml
```diff
@@ -0,0 +1,28 @@
+<!--
+	Mismatched:
+		^duolingo.cn		equal to www.duolingo.cn
+	Timeout:
+		sponsorships.duolingo.com
+-->
+<ruleset name="Duolingo.com">
+	<target host="duolingo.com" />
+	<target host="*.duolingo.com" />
```
pipboy96: Please avoid <kbd>*</kbd> if it's even possible.

src/chrome/content/rules/Duolingo.xml
```diff
@@ -0,0 +1,28 @@
...
+	<target host="englishtest.duolingo.cn" />
+
+	<securecookie host=".+" name=".+" />
+
+	<exclusion pattern="^http://sponsorships\.duolingo\.com/" />
+		<test url="http://sponsorships.duolingo.com/" />
+		<test url="http://www.duolingo.com/" />
+		<test url="http://api.duolingo.com/" />
+		<test url="http://englishtest.duolingo.com/" />
+		<test url="http://de.duolingo.com/" />
```
pipboy96: Can you do the opposite, e.g. only include working targets.

ivysrono: are there tools to find subdomains all?

pipboy96: @ivysrono https://github.com/OWASP/Amass

pipboy96:
Did you test all subdomains of `.cn` too?

ivysrono:
> 
> 
> Did you test all subdomains of `.cn` too?

I did not found more subdomains of .cn

PR #18011, opened 2019-05-25
Update Alukah.net.xml

Giltyhub:


src/chrome/content/rules/Alukah.net.xml
```diff
@@ -1,17 +1,16 @@
 <!--
 	Timeout:
-		alukah.net
-		api.alukah.net
 		en.alukah.net
 		majles.alukah.net
 	Refused:
 		fatawa.alukah.net
```
Bisaloo: radio.alukah.net

Giltyhub: @Bisaloo merci, added

PR #18015, opened 2019-05-26
Update Media-Storehouse.xml

cschanaj:
#17965 `mediastorehouse.com` was not protected.

### DIFF
--- a/src/chrome/content/rules/Media-Storehouse.xml
+++ b/src/chrome/content/rules/Media-Storehouse.xml
@@ -11,7 +11,7 @@
 		- images.mediastorehouse.net
 		- www.mediastorehouse.net
 -->
-<ruleset name="Media Storehouse (partial)">
+<ruleset name="Media Storehouse">
 	<!-- *.mediastorehouse.com -->
 	<target host="mediastorehouse.com" />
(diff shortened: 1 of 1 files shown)


J0WI:
> #17965 mediastorehouse.com was not protected.

Isn't this covered by

https://github.com/EFForg/https-everywhere/blob/b371e9331a64baf6fa372e74faa1343434775f93/src/chrome/content/rules/Media-Storehouse.xml#L16

https://github.com/EFForg/https-everywhere/blob/b371e9331a64baf6fa372e74faa1343434775f93/src/chrome/content/rules/Media-Storehouse.xml#L31-L32

src/chrome/content/rules/Media-Storehouse.xml
```diff
@@ -1,34 +1,32 @@
...
 
-	Nonfunctional:
-
-		- images.mediastorehouse.net	(times out; not on secure nor images...com)
-
+		SSL peer certificate was not OK:
+		- images.mediastorehouse.net
+		- www.mediastorehouse.net
 -->
-<ruleset name="Media Storehouse (partial)">
```
J0WI: There are still a lot of uncovered subdomains. Maybe you just want to keep the `partial`?

PR #18045, opened 2019-05-31
Create Birdtree.org.xml

Bisaloo:


pipboy96:
There are more:
```
calendar.birdtree.org
email.birdtree.org
files.birdtree.org
mobilemail.birdtree.org
pop.birdtree.org
imap.birdtree.org
fax.birdtree.org
smtp.birdtree.org
```

Bisaloo:
@pipboy96, these are all internal subdomains as far as I can tell. I'm not sure we should add them. It unnecessarily increase the noise in rulesets.

pipboy96:
@Bisaloo I'm not sure about procedures in this case either.

Bisaloo:
I think it depends on the reviewer so you're correct we should probably adopt a clear and common guideline.

@cschanaj @gloomy-ghost @J0WI

J0WI:
Hosts like `pop.birdtree.org` without any HTTP service can be ignored.

pipboy96:
@J0WI I (softly) disagree since if user had accidentally accessed it over unencrypted HTTP, it's possible for an attacker to do SSLstrip-like attack as long as the user doesn't notice that the connection is unencrypted. Of course this only applies if corresponding HTTPS service exists.

gloomy-ghost:
Oh, I remember internal subdomains were brought up in #7662, but it wasn't really the focus.

I still hold my opinion about potential breakage that the risk is acceptable. However, after joining the project for a long time, I feel like we simply won't have the resources to maintain them with little people actually benefit from these ruleset. So similar with small blog sites - it's fine if someone finds it important enough to propose a ruleset, and it can be merged if a reviewer happens to check on it, but we shouldn't encourage it.

cschanaj:
In my opinions, we should ignore subdomains without any HTTP(s) services, like `pop` and `smtp` in this case. They increase maintenance burden since we lack an established way to perform any form of test.

Besides, I usually prefers not to add coverage to internal subdomains which require authentications. A notable exception is when the rulesets come from an affiliated/ authorized user to the organization. Otherwise, it seems too risky to rewrite internal connections blindly, especially for large corporate networks.

Bisaloo:
@pipboy96, I've seen you recommend Amass to enumerate subdomains. Why do you prefer it over Sublist3r?
If we update the `Enumerating subdomains` section of the contributing guide, we could possibly change that as well.

J0WI:
> Besides, I usually prefers not to add coverage to internal subdomains which require authentications.

I have no doubts if they already use/force HTTPS for the authentication.

pipboy96:
@Bisaloo Faster, especially if brute mode is on compared to Sublist3r with same brute mode setting, no Google bans, written in Go.

pipboy96:
@J0WI Playing devil's advocate, maybe it redirects back to HTTP after authentication (it's unlikely though).

PR #18049, opened 2019-06-01
Update KOOORA.com.xml

ComputerJy:
Added a trivial subdomain

### DIFF
--- a/src/chrome/content/rules/KOOORA.com.xml
+++ b/src/chrome/content/rules/KOOORA.com.xml
@@ -1,18 +1,13 @@
 <!--
-		Error code 29 (SSL not supported):
-
-	o forum.kooora.com
-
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/KOOORA.com.xml
```diff
@@ -12,7 +12,8 @@
 	<target host="www.kooora.com" />
 	<target host="m.kooora.com" />
 	<target host="o.kooora.com" />
-
+	<target host="img.kooora.com" />
```
J0WI: Targets should be [sorted alphabetically](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#target-ordering) (except `www` which comes right after `^`).

PR #17980, opened 2019-05-21
DRY out loading of rulesets, make everything parsed through addFromJson

Hainish:

PR #18052, opened 2019-06-01
Replace Array() with []

pipboy96:
This should also be slightly faster (in real world performance, to much less degree than demonstrated in this example):

```js
let c = null

const a = performance.now()

for (let i = 0; i < 1e8; i++) {
  c = Array()
}

const b = performance.now()

console.log(b - a)

// 2634.315000002971
```

```js
let c = null

const a = performance.now()

for (let i = 0; i < 1e8; i++) {
  c = []
}

const b = performance.now()

console.log(b - a)

// 1824.0100000039092
```

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -231,7 +231,7 @@ RuleSets.prototype = {
   addFromXml: function(ruleXml, scope) {
     const rulesets_xml = ruleXml.getElementsByTagName("ruleset");
 
-    let rulesets = Array();
+    let rulesets = [];
     for (let ruleset_xml of rulesets_xml) {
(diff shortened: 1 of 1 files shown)


cschanaj:
related #17980

pipboy96:
@cschanaj Not sure how this is directly related. Thanks for the link though.

cschanaj:
#17980 is merged recently, contributors who didn't rebase their local branch recently might have difficulties follow the code. there was only one instance of `Array()` prior to that.

pipboy96:
@pipboy96 Thanks for explaining.

Hainish:
https://stackoverflow.com/questions/7375120/why-is-arr-faster-than-arr-new-array interesting.  Thanks!

PR #18057, opened 2019-06-04
Add Ghidra-SRE.org.xml

pipboy96:


src/chrome/content/rules/Ghidra-SRE.org.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+	For other National Security Agency coverage, see National-Security-Agency.xml
```
Bisaloo: Can you also add a link to this ruleset in `National-Security-Agency.xml` please?

pipboy96: @Bisaloo Done.

PR #18058, opened 2019-06-05
Release 2019.6.4

Hainish:


Hainish:
cc @J0WI, @Bisaloo, @cschanaj, @Hainish, @jsha, @andresbase, @pipboy96

This release was made due to a medium-level vulnerability in the interstitial `cancel` page.  The vulnerability could trigger a clickjacking attack in Chrome browser.  This would not affect Firefox due to randomization of the extension ID.  Clickjacking is further mitigated in Tor Browser due to first-party isolation.  See https://twitter.com/vladimir_metnew/status/1135226568601690112.

pipboy96:
@Hainish Maybe we need to add CSP as part of defense in depth?

pipboy96:
@Hainish It is also possible to use `cancel.html` link to trick the user into disable HTTPS Everywhere for some website, even when EASE is off.

Example: `chrome-extension://gcbommkclmclpchllfjekcdonpmejbdp/pages/cancel/index.html?originURL=http://s3.amazonaws.com/ygl-photos/9867O5a6c86190f0ba`

If user presses the button, it will go to unencrypted version of page and also will disable HTTPS Everywhere on entire `s3.amazonaws.com` domain.

Should we disable `cancel.html` if EASE is off, or maybe use something like HMAC to test if we created this link?

cschanaj:
I think @pipboy96 is right that phishing is possible. EFF might have to look into another way to securely pass the original URL to the extension. 

Regarding the suggestions we have, disable the cancel page outside EASE alone does not work because an attacker might add the phishing content via an HTTPS page under his/ her control. 

A signed URL using runtime generated key could work, but the implementation can be tricky.

Hainish:
Tricking users into disabling HTTPS Everywhere for a specific site is a risk, but since (1) it requires active participation of the user, and (2) this action has to be taken once per domain, I consider this a low risk scenario.  I think this still warrants fixing, but not a new release to be made specifically for this scenario.

Bisaloo:
Is there a specific reason chrome is not randomizing extension IDs? It may be a good idea to file a ticket to fix this on the browser side as well.

pipboy96:
@Bisaloo They are unlikely to do that now since that would probably break something.

Hainish:
In fact, each approach has its advantages and disadvantages.

Chromium deployed WebExtensions first, and Firefox chose to randomize extension IDs as a security measure.  But random IDs can also be a browser fingerprinting vector.  If an extension inserts some resource into the page that includes the extension ID (e.g. if Privacy Badger swapped out social media widgets this way instead of using `data` URLs) this unique ID can be discovered by the remote site, and used as a supercookie.

pipboy96:
@Hainish Nice one! Didn't think about this in such way. Do you think always random nonce or nonce that is only used for one browsing session is fine?

Hainish:
If it was once per browsing session then that would limit the supercookie lifetime.  Perhaps also making the ID randomized when cookies are cleared, and randomizing the ID for every first party.  But that does seem like a lot of effort on the browsers part...

PR #18063, opened 2019-06-06
Minor style fixes to cschanaj script

pipboy96:
Feel free to partially merge/cherrypick if needed.


Hainish:
Looks good and confirmed it still performs as intended

pipboy96:
@Hainish can you also do `s/new Array()/[]/`? I accidentally missed it.

Hainish:
@pipboy96 I don't think it necessary to change all instances of this since this is only performed by admins and only once in a while, the performance gains will be minimal.  `new Array()` is still valid even if it is slightly optimized.

pipboy96:
@Hainish should https://github.com/EFForg/https-everywhere/pull/18065 be closed or can you still merge, even though it's not really 100% necessary?

Hainish:
Merged.

PR #18064, opened 2019-06-06
Fix ruleset editor

pipboy96:
This is how appearing effect looks like: https://raw.githubusercontent.com/pipboy96/https-everywhere/raw/tmp/fix-ruleset-editor.mp4


Hainish:
Thanks!

PR #18066, opened 2019-06-06
 Remove rulesets from coverage whitelist if they pass the coverage test

pipboy96:
Feel free to cherry-pick the first commit and run `ruleset-whitelist-cleanup.sh` separately.


Hainish:
This seems to have changed the ordering of some lines without changing anything else.  The previous ordering was sorted with https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-whitelist-cleanup.sh#L41

pipboy96:
@Hainish I have noticed that. Seems this script has somehow lost its case insensitivity.

PR #18068, opened 2019-06-07
Run node utils/trivialize-rules/trivialize-targets.js

cschanaj:
This is a follow-up PR for #17855 

There is around 500+ targets which are gone from DNS, however I am not going to handle the fetch errors in the PR. 

Besides, it surprises me that the node `URL` object accepts hostname with leading dots. This make the script from #18068 a little buggy. Since it is not possible to whitelist rulesets from schematic errors, Travis failed. The problematic rulesets are fixed in eef2855  

```bash
$ node
> new URL('https://.example.com/')
URL {
  href: 'https://.example.com/',
  origin: 'https://.example.com',
  protocol: 'https:',
  username: '',
  password: '',
  host: '.example.com',
  hostname: '.example.com',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }

```

```
$ node --expose-gc utils/memusage.js

[Before]
Initial usage: 42 MB
Maximum usage: 77.8 MB

[After]
Initial usage: 42.3 MB
Maximum usage: 77.6 MB
```


cschanaj:
cc @Hainish @J0WI @Bisaloo @pipboy96 this is ready for review. note that this might cause a merge conflict with  #18066 . thanks.

Hainish:
I know it's our normal policy to trivialize rules wherever possible, but have we done any memory analysis on this?  I'm a bit worried about balooning the all_rules object...

pipboy96:
@Hainish I believe complex regexes themselves use up lots of memory.

Hainish:
I missed the memory analysis above! Thanks for this, merging

PR #18093, opened 2019-06-14
Adding https-everywhere-lib submodule, which moves much of the

Hainish:
functionality of `rules.js` into WASM for memory efficiency and
performance.

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -223,7 +223,9 @@ RuleSets.prototype = {
     this.ruleActiveStates = await this.store.get_promise('ruleActiveStates', {});
     try {
       this.wasm_rs = wasm.RuleSets.new();
-    } catch(e) {}
+    } catch(e) {
+      util.log(util.WARN, 'Falling back to pure JS implementation.');
+    }
     await applyStoredFunc(this);
     await this.loadStoredUserRules();
     await this.addStoredCustomRulesets();
@@ -575,7 +577,7 @@ RuleSets.prototype = {
     let results;
     if (this.wasm_rs) {
       let pa = this.wasm_rs.potentially_applicable(host);
-      results = [...pa].map(ruleset => {
+      results = new Set([...pa].map(ruleset => {
         let rs = new RuleSet(ruleset.name, ruleset.default_state, getScope(ruleset.scope), ruleset.note);
 
         if (ruleset.cookierules) {
@@ -598,7 +600,7 @@ RuleSets.prototype = {
           rs.exclusions = null;
         }
         return rs;
-      });
+      }));
     } else {
       // Let's begin search
       // Copy the host targets so we don't modify them.
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -2,7 +2,8 @@
 
 (function(exports) {
 
-const util = require('./util');
+const util = require('./util'),
+  wasm = require('./wasm');
 
 let settings = {
   enableMixedRulesets: false,
@@ -220,6 +221,9 @@ RuleSets.prototype = {
   loadFromBrowserStorage: async function(store, applyStoredFunc) {
     this.store = store;
     this.ruleActiveStates = await this.store.get_promise('ruleActiveStates', {});
+    try {
+      this.wasm_rs = wasm.RuleSets.new();
+    } catch(e) {}
     await applyStoredFunc(this);
     await this.loadStoredUserRules();
     await this.addStoredCustomRulesets();
@@ -241,11 +245,20 @@ RuleSets.prototype = {
 
   addFromJson: function(ruleJson, scope) {
     const scope_obj = getScope(scope);
-    for (let ruleset of ruleJson) {
-      try {
-        this.parseOneJsonRuleset(ruleset, scope_obj);
-      } catch(e) {
-        util.log(util.WARN, 'Error processing ruleset:' + e);
+
+    if (this.wasm_rs) {
+      this.wasm_rs.add_all_from_js_array(
+        ruleJson,
+        settings.enableMixedRulesets,
+        this.ruleActiveStates,
+        scope);
+    } else {
+      for (let ruleset of ruleJson) {
+        try {
+          this.parseOneJsonRuleset(ruleset, scope_obj);
+        } catch(e) {
+          util.log(util.WARN, 'Error processing ruleset:' + e);
+        }
       }
     }
   },
@@ -322,7 +335,15 @@ RuleSets.prototype = {
    */
   addUserRule : function(params, scope) {
     util.log(util.INFO, 'adding new user rule for ' + JSON.stringify(params));
-    this.parseOneJsonRuleset(params, scope);
+    if (this.wasm_rs) {
+      this.wasm_rs.add_all_from_js_array(
+        [params],
+        settings.enableMixedRulesets,
+        this.ruleActiveStates,
+        scope);
+    } else {
+      this.parseOneJsonRuleset(params, scope);
+    }
 
     // clear cache so new rule take effect immediately
     for (const target of params.target) {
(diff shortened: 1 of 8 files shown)


Hainish:
This gracefully falls back to pure JS for contexts where WASM isn't supported, like the Tor Browser.

In some manual benchmarking with the Chromium Task Manager I've seen the memory consumption go down from ~105M to ~85M.

The [`https-everywhere-lib`](https://github.com/EFForg/https-everywhere-lib) this refers to is a Rust library which compiles to WASM and has to be rebuilt if changes are made.  There are instructions there for building and testing this library via Docker.

Hainish:
WASM is included via the wasm_bindgen function rather than modularly because modular includes are not supported for WASM just yet.  This corresponds to the `-t no-modules` flag in the library compilation.

Hainish:
This adds roughly 100k to the extension download size, bringing it from 1.7M to 1.8M.

pipboy96:
Not really sure about Rust, maybe we should have written it in pure WASM (WAT)?

pipboy96:
Great otherwise.

cschanaj:
I believe rust is a reasonable language of choice, as it provide some built-in libraries which ease our effort. Will try to look into the rust code and test ASAP. thanks @Hainish 👍

chromium/background-scripts/rules.js
```diff
@@ -546,51 +572,80 @@ RuleSets.prototype = {
...
 
-    // Let's begin search
-    // Copy the host targets so we don't modify them.
-    let results = (this.targets.has(host) ?
-      new Set([...this.targets.get(host)]) :
-      new Set());
+    let results;
+    if (this.wasm_rs) {
+      let pa = this.wasm_rs.potentially_applicable(host);
+      results = [...pa].map(ruleset => {
```
cschanaj: `results` should be a `Set` instead of an `Array`. Otherwise, `cached_item.size` is `undefined` when there is a cache hit.

chromium/background-scripts/rules.js
```diff
@@ -220,6 +221,9 @@ RuleSets.prototype = {
   loadFromBrowserStorage: async function(store, applyStoredFunc) {
     this.store = store;
     this.ruleActiveStates = await this.store.get_promise('ruleActiveStates', {});
+    try {
+      this.wasm_rs = wasm.RuleSets.new();
+    } catch(e) {}
```
cschanaj: maybe we should log the errors here?

pipboy96: We should, definitely.

Hainish:
> I believe rust is a reasonable language of choice, as it provide some built-in libraries which ease our effort. Will try to look into the rust code and test ASAP. thanks @Hainish +1
> 
> P.S. I agree with @pipboy96 that the .wasm file should be deterministic and compiled when the extension is built.

Consecutive compilations of the WASM binary have resulted in the same bit-for-bit result on my local development docker environment.  Can you test if this is true for you as well?

I'm reluctant to add additional compile-time requirements due to the fact that we'd need to add the entire Rust build chain.  I understand the argument about opacity - perhaps we can flag any PR which changes the submodule's commit with a tag which indicates that the resulting `wasm` file should be re-checked?

pipboy96:
@Hainish Well, commit IDs are SHA-1 based so theoretically they can be forged. Can you sign the file in the same way the rulesets are signed and verify the signature in JS code before any other interaction with the file takes place?

pipboy96:
In this case, signature failure won't break the extension, it will just run slower, and TB will not even need to attempt signature check since TB doesn't support WASM.

Hainish:
> @Hainish Well, commit IDs are SHA-1 based so theoretically they can be forged. Can you sign the file in the same way the rulesets are signed and verify the signature in JS code before any other interaction with the file takes place?

Can you explain how you see an attack scenario taking place?  The SHA-1 sum would be changed, but the submodule URI would not.  We already implicitly trust the information security infrastructure of Github to accurately deliver code based on the sha1sum they have stored for a commit.  Additional sha256sum checks can be run directly on the `wasm` code.  The only attack scenario I can foresee is Github delivering a forged commit based on a collision in the SHA-1 hash, but this is not new - this attack scenario already exists for the main HTTPS Everywhere repo.  We haven't increased the attack surface.

I think that this is part of a larger problem that the `git` infrastructure needs to be migrated to a more collision-resistant hashing algorithm, but I don't think that's our responsibility to address.

chromium/background-scripts/rules.js
Hainish: I'll convert this to a `Set`.  Currently, `this.wasm_rs.potentially_applicable` returns a `Set`, but we have to call `map` on the result because WebAssembly does not have access to the `CookieRule` or `Rule` objects.  This may be solvable, in which case we won't have to run this `map` at all.  I'll also look into this as a longer-term solution.

pipboy96:
@Hainish We use signed commits/tags in some places, which may reduce the risks (but probably not by a large amount). I also had in mind the possibility WASM module could only be downloaded as needed (if browser supports WASM).

> Can you explain how you see an attack scenario taking place?

No, I don't have any specific scenario in mind. I just think an opaque binary in Git repo may seem suspicious to the users, and may make people think something's wrong. It's not much a security issue, but the PR one.

Hainish:
@cschanaj resolved those issues, let me know if you plan on doing any additional reviews.

chromium/background-scripts/rules.js
```diff
@@ -223,7 +223,9 @@ RuleSets.prototype = {
     this.ruleActiveStates = await this.store.get_promise('ruleActiveStates', {});
     try {
       this.wasm_rs = wasm.RuleSets.new();
-    } catch(e) {}
+    } catch(e) {
+      util.log(util.WARN, 'Falling back to pure JS implementation.');
```
pipboy96: can you actually output `e` too?

chromium/background-scripts/wasm.js
```diff
@@ -2,10 +2,15 @@
...
-const { RuleSets } = wasm_bindgen;
+const util = require('./util'),
+  { RuleSets } = wasm_bindgen;
 
 async function initialize() {
-  await wasm_bindgen(chrome.runtime.getURL('wasm/https_everywhere_lib_bg.wasm'));
+  try {
+    await wasm_bindgen(chrome.runtime.getURL('wasm/https_everywhere_lib_bg.wasm'));
+  } catch(e) {
+    util.log(util.WARN, 'The wasm library has not loaded correctly for some reason.');
```
pipboy96: ditto

pipboy96:
@Hainish this doesn't include the line error happened on, but with current design you can't really fix that since `util.log` only accepts one string argument. I will add this on my roadmap.

Hainish:
@cschanaj have you had a chance to reproduce the `wasm` file?

pipboy96:
@Hainish I may try reproducing right now.

pipboy96:
@Hainish I was unable to reproduce the file, the file size I got was `87686` and SHA256 was `aa197d91e85134819e8e37f22423292ea525fd619951303922b78339d2cc811c`.

Hainish:
@pipboy96 you followed the instructions in the `README.md` file?

Hainish:
@pipboy96 can you do a `docker images` and let me know the hash for your `https-everywhere-lib` and `rust:latest` images?

pipboy96:
@Hainish 

>

pipboy96:
@Hainish 

    https-everywhere-lib   latest              506bcb77b0d5        25 minutes ago      2.09GB
    rust                   latest              cc8fdbea4a2a        6 days ago          1.7GB

Hainish:
We have the same hash for `rust:latest`: I'll make sure that we lock to that commit.  I should commit `Cargo.lock` to ensure all crate versions are the same - one moment let me do that.

Hainish:
@pipboy96 I've made some changes to increase reproducibility.  Can you pull, build, and compile once again?

pipboy96:
@Hainish 

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

I, pipboy96 (pipboy96@protonmail.com), hereby testify that a file called https_everywhere_lib_bg.wasm with SHA-256 hash of a30d6b8c309882edcbff4ee1c07154f53a6f9a95fae4d3d8ffcb66fbf8ce07ee and size of 87289 bytes fully corresponds to a Git commit with hash 58d40d6fb964532c6d9a90e9dcb6b51f19004331 and was successfully reproduced by me from its source code.
-----BEGIN PGP SIGNATURE-----

iI4EARYIADYWIQROV8sHuKtZUjpKQXlrXPDIQKF3dAUCXQf7rRgccGlwYm95OTZA
cHJvdG9ubWFpbC5jb20ACgkQa1zwyEChd3SQqQD/XQPebTj/PiRkzFOCfXnH8aDT
vofCc8XrNH0rgbivQoIA/AmiqzEGhqC8EpDlSEySMXlTWdzibSvQc7vHoLq34ckB
=bRPl
-----END PGP SIGNATURE-----
```

pipboy96:
@cschanaj Please also try to reproduce.

cschanaj:
@pipboy96 @Hainish I am able to reproduce the wasm files locally. thanks.

pipboy96:
@Hainish Are we ready to merge?

Hainish:
@pipboy96 yes

PR #18101, opened 2019-06-18
Add more hosts to Omnirom.org.xml

J0WI:


src/chrome/content/rules/Omnirom.org.xml
```diff
@@ -1,6 +1,17 @@
...
+		- jira.omnirom.org
+-->
 <ruleset name="Omnirom.org">
-<target host="omnirom.org"/>
-<target host="www.omnirom.org"/>
-<rule from="^http:" to="https:" />
-</ruleset>
+	<target host="omnirom.org"/>
+	<target host="www.omnirom.org"/>
+	<target host="blog.omnirom.org"/>
```
Bisaloo: For the record, there is mixed content from `fonts.googleapis.com`

src/chrome/content/rules/Omnirom.org.xml
```diff
@@ -1,6 +1,17 @@
...
+	<target host="dl.omnirom.org"/>
+	<target host="docs.omnirom.org"/>
+	<target host="gerrit.omnirom.org"/>
+	<target host="jenkins.omnirom.org"/>
+	<target host="lists.omnirom.org"/>
```
Bisaloo: I get a 502 with the default landing page, could you please add a test url such as http://lists.omnirom.org/pipermail/maintainers/2015-August/date.html?

J0WI: Done

PR #18109, opened 2019-06-20
Making the options and cancel pages more mobile friendly

Hainish:
The options page in Firefox for Android looks weird, and opening it in a new tab looks much better.  We could do this with `"open_in_tab": true` in `manifest.json`, but opening the options page in a separate tab in desktop also looks a bit weird.  Thus the slightly hacky "Settings" link.

### DIFF
--- a/chromium/pages/options/ux.js
+++ b/chromium/pages/options/ux.js
@@ -5,6 +5,20 @@
 
 "use strict";
 
+if (~navigator.userAgent.indexOf("Android")) {
+  const url = new URL(window.location.href);
+  if (!url.searchParams.get('redirected')) {
+    url.searchParams.set('redirected', true);
+    document.body.innerText = "";
+    let link = document.createElement("a");
+    link.href = url.href;
+    link.target = "_blank";
+    link.className = "settings";
+    link.innerText = chrome.i18n.getMessage("options_settings");
+    document.body.appendChild(link);
+  }
+}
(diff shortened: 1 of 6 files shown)


chromium/pages/options/ux.js
```diff
@@ -5,6 +5,20 @@
 
 "use strict";
 
+if (~navigator.userAgent.indexOf("Android")) {
```
pipboy96: ?

Hainish: https://wsvincent.com/javascript-tilde/

Hainish: Looks like we can start using `.includes` as well, since we've specified https://github.com/EFForg/https-everywhere/blob/master/chromium/.eslintrc.json#L8

pipboy96: Can you do `!== -1` instead, for ease of reading, or `String.includes` since it's supported already.

chromium/pages/options/ux.js
```diff
@@ -5,7 +5,7 @@
 
 "use strict";
 
-if (~navigator.userAgent.indexOf("Android")) {
+if (navigator.userAgent.includes("Android")) {
```
pipboy96: 🎉

pipboy96:
@Hainish can you log into mattermost?

PR #18114, opened 2019-06-21
[USRowing.org] Default_off ruleset

Bisaloo:


pipboy96:
If ruleset is completely broken, better just delete it.

PR #18115, opened 2019-06-21
[LightTheNight.org] Simplify and enable ruleset 

Bisaloo:


src/chrome/content/rules/LightTheNight.org.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	MCB:
+		pages.lightthenight.org (e.g. https://pages.lightthenight.org/va/Frdrcksb13/TeamTomas)
+
+	No working URL known:
+		donate.lightthenight.org
```
J0WI: Wildcard DNS

PR #18139, opened 2019-06-27
Update Github-Pages.xml

Giltyhub2:
googletrends.github.io/GOPDebate1 returns 404, and i don't get any mixed content warning with the other googletrends page

### DIFF
--- a/src/chrome/content/rules/Github-Pages.xml
+++ b/src/chrome/content/rules/Github-Pages.xml
@@ -16,6 +16,7 @@
 		<test url="http://duckietv.github.io/DuckieTV/" />
 
 	<exclusion pattern="^http://googletrends\.github\.io/iframe-scaffolder/?" />
+		<test url="http://googletrends.github.io/iframe-scaffolder" />
 		<test url="http://googletrends.github.io/iframe-scaffolder/" />
 
 	<exclusion pattern="^http://schizoduckie\.github\.io/DuckieTorrent/?" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Github-Pages.xml
```diff
@@ -15,11 +15,6 @@
 		<test url="http://duckietv.github.io/DuckieTV" />
 		<test url="http://duckietv.github.io/DuckieTV/" />
 
-	<!-- See https://github.com/EFForg/https-everywhere/issues/2507 -->
-	<exclusion pattern="^http://googletrends\.github\.io/(GOPDebate1|iframe-scaffolder)/" />
```
Bisaloo: - https://googletrends.github.io/: MCB from `font.googleapis.com` (begnin)
- https://googletrends.github.io/Sandy/index.html: MCB from `libs.cartocdn.com` and `maps.google.com`.

I'm not sure we should change this for the time being.

Giltyhub2: @Bisaloo i've added it back and only removed gopdebate1

src/chrome/content/rules/Github-Pages.xml
```diff
@@ -15,9 +15,8 @@
...
-		<test url="http://googletrends.github.io/GOPDebate1/" />
+	<exclusion pattern="^http://googletrends\.github\.io/iframe-scaffolder/?" />
```
Bisaloo: Why not use the simpler

```suggestion
	<exclusion pattern="^http://googletrends\.github\.io/iframe-scaffolder" />
```

?

Giltyhub2: @Bisaloo my mistake, changed now thanks!

PR #18153, opened 2019-07-01
Refactor code for the browser session storage

cschanaj:


pipboy96:
@cschanaj Is the goal here to generalize `AppliedRulesets` to universal session storage?

cschanaj:
@pipboy96 Yes, it also avoid using multiple `Map` in the global scope for similar things.

chromium/background-scripts/background.js
```diff
@@ -187,37 +187,60 @@ chrome.browserAction.onClicked.addListener(e => {
...
+ * A centralized storage for browsing data within the browser session.
  */
-function AppliedRulesets() {
-  this.active_tab_rules = new Map();
-  this.active_tab_main_frames = new Map();
+function BrowserSession() {
+  this.tabs = new Map();
+  this.requests = new Map();
 
   let that = this;
```
pipboy96: Is there any reason for this? Wouldn't just turning the next function into an arrow function make `this` a part of its closure?

pipboy96:
Can you also turn this into a ES6 class?

Hainish:
Thanks!

Hainish:
This is a great refactor.

PR #18154, opened 2019-07-01
Update OpenStreetMap.de.xml

nmxcgeo:
Hi,

I included all working hosts from Sublist3r and noted the rest under "Doesn't work".

Did I handle the special chars correctly?

How is the procedure with mixed content domains?
I didn't include it because it completely breaks the site's function.

Greetings
Nmxcgeo

### DIFF
--- a/src/chrome/content/rules/OpenStreetMap.de.xml
+++ b/src/chrome/content/rules/OpenStreetMap.de.xml
@@ -1,17 +1,40 @@
 <!--
 	For OpenStreetMap.de
-
-		Works:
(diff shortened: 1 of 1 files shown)


J0WI:
There is a [`platform="mixedcontent"`](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#general-info) attribute, but this would require a dedicated ruleset file.
Excluding the host is fine for me and punycode is the correct way to handle special chars.

Thanks!

PR #18162, opened 2019-07-05
Create mhfa.com.au.xml

sunnz:


pipboy96:
Just a suggestion, generally, having same ruleset name and filename makes things easier.

sunnz:
> Just a suggestion, generally, having same ruleset name and filename makes finding the ruleset in the repo much easier.

Sure, ruleset name updated to be the same as the filename now. :)

src/chrome/content/rules/mhfa.com.au.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="mhfa.com.au">
+  <target host="mhfa.com.au" />
```
pipboy96: Do not use spaces for indentation, use `\t` instead.

sunnz:
Changed indentation from spaces to tabs

J0WI:
Thanks!

PR #18167, opened 2019-07-06
Create TafasirStore.com.xml

Giltyhub2:


J0WI:
This seems to be related to `Salla.sa.xml`. Could you document this in comments?

Giltyhub2:
@J0WI thanks, done

Giltyhub2:
> Please do the same on `Salla.sa.xml`

I don't think it makes sense for that one since this site just redirects to an online store on Salla.sa and it isn't officially related to that website.

J0WI:
We always add cross ruleset references on both sides. If they are not related you can revert 627a868469a52057f655ce07ae3b20227d829b16.

Giltyhub2:
@J0WI thanks, i will take that into consideration next time, meanwhile i reverted my change

PR #18178, opened 2019-07-09
Update jenkins.io ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/jenkins.io.xml
```diff
@@ -1,12 +1,13 @@
 <!--
 	Invalid certificate:
-	  nginx.azure.jenkins.io
-	  lettuce.jenkins.io
```
J0WI: Why did you remove them?

Jochen-A-Fuerbacher: Sublist3r didn't list them.

J0WI: They're still up

J0WI:
Thanks!

PR #18210, opened 2019-07-16
Create Wellcome-HK-Lucky-Draws.xml

ipaucek4680:
This website is used for [Wellcome](https://en.wikipedia.org/wiki/Wellcome)'s (*Wellcome* is a large supermarket chain in Hong Kong) lucky draws.

### DIFF
--- a/src/chrome/content/rules/easy2winluckydraw.com.xml
+++ b/src/chrome/content/rules/easy2winluckydraw.com.xml
@@ -1,5 +1,0 @@
-<ruleset name="easy2winluckydraw">
-	<target host="www.easy2winluckydraw.com" />
-
-	<rule from="^http:" to="https:" />
-</ruleset>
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/easy2winluckydraw.com.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="easy2winluckydraw">
```
J0WI: Please also add `easy2winluckydraw.com` (without www) as target.

ipaucek4680:
@J0WI 
I have added `easy2winluckydraw.com` (without www) as target. In addition, I have found some domains which _were_ used for lucky draws (by the same company) and have added them to the ruleset.

J0WI:
Thanks!

Issue #16733, opened 2018-10-14
 Add new ruleset for radioreference.com


ghost:
Type: new ruleset
Domain: http://www.radioreference.com/

phuzion:
Should be fixed with the merge of #18222.

PR #18222, opened 2019-07-22
Create Radioreference.com.xml

phuzion:
Should solve #16733.

### DIFF
--- a/src/chrome/content/rules/Radioreference.com.xml
+++ b/src/chrome/content/rules/Radioreference.com.xml
@@ -1,0 +1,23 @@
+<!--
+	Nonfunctional hosts in *.radioreference.com:
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Radioreference.com">
+	<target host="radioreference.com" />
+	<target host="www.radioreference.com" />
+	<target host="api.radioreferenc.com" />
+	<target host="forums.radioreference.com" />
+	<target host="m.radioreference.com" />
+	<target host="register.radioreference.com" />
+	<target host="u.radioreference.com" />
+	<target host="wiki.radioreference.com" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Radioreference.com.xml
```diff
@@ -0,0 +1,23 @@
...
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
+<ruleset name="Radioreference.com">
+	<target host="radioreference.com" />
+	<target host="www.radioreference.com" />
+	<target host="api.radioreference.com" />
+	<target host="forums.radioreference.com" />
+	<target host="m.radioreference.com" />
```
J0WI: `proxy.radioreference.com`

src/chrome/content/rules/Radioreference.com.xml
```diff
@@ -0,0 +1,23 @@
+<!--
+	Nonfunctional hosts in *.radioreference.com:
+
+	h: http redirect
+	m: certificate mismatch
+	r: connection refused
+	s: self-signed certificate
+	t: timeout on https
+-->
```
J0WI: You can remove this comments.

phuzion: I've removed these in the latest commit.

src/chrome/content/rules/Radioreference.com.xml
phuzion: proxy.radioreference.com is NXDOMAIN according to what I've found. Would you still like me to add it to the ruleset?

J0WI: No, you can ignore this.

src/chrome/content/rules/Radioreference.com.xml
```diff
@@ -0,0 +1,14 @@
+<ruleset name="Radioreference.com">
+	<target host="radioreference.com" />
```
J0WI: Connection times out

phuzion: I've removed the root domain, as it times out on HTTP and HTTPS.

J0WI: Thanks!

PR #18232, opened 2019-07-29
Add Ace-book.net.xml

pipboy96:
See also #12960.


pipboy96:
The website has active mixed content warning but it redirects to HTTPS on its own anyway (however, it doesn't use HSTS and is not in preload list) and it doesn't affect the website's functionality.

Issue #18212, opened 2019-07-17
Direct HTTP link to a Firefox extension doesn't redirect correctly if EASE is enabled


kidhanis:
Type: code issue

This might be related to #17774.
I've tested with HTTPS Everywhere version 2019.6.27 and the latest Beta and Nightly versions of Firefox.

### Steps to reproduce:
1. Enable EASE.
2. Find an HTTPS URL to an `.xpi` file, change it to HTTP, and visit the modified URL.  For example, try clicking on http://www.eff.org/files/https-everywhere-2019.6.27-eff.xpi.
### Expected results:
The browser prompts the user to download and to install the extension.
### Actual results:
The page gets redirected to the `moz-extension://` blocking page and the `.xpi` file cannot be downloaded.

pipboy96:
I think the reason is Firefox doesn't let you tamper with .xpi requests.

cschanaj:
> 
> 
> I think the reason is Firefox doesn't let you tamper with .xpi requests.

But this is strange, `www.eff.org` is preloaded https://hstspreload.org/?domain=www.eff.org. This request should, thoretically, always go through HTTPS...

pipboy96:
Should be fixed by #18233.

pipboy96:
This issue still exists, cc @zoracon.

zoracon:
Revisiting, when I click on  http://www.eff.org/files/https-everywhere-2019.6.27-eff.xpi, i get the install prompt. Is that still not the case for others?

PR #18233, opened 2019-07-29
Remove NS_BINDING_ABORTED from list of errors triggering EASE error page (fixes XPI installation on Firefox)

pipboy96:


pipboy96:
Should fix #18212.

PR #18234, opened 2019-07-29
Create Dotwhat.net.xml

ryanbr:


src/chrome/content/rules/Dotwhat.net.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset name="Dotwhat.net/">
+    <target host="dotwhat.net" />
+    <target host="www.dotwhat.net" />
+    <target host="webmail.dotwhat.net" />
+    <target host="ftp.dotwhat.net" />
+    <target host="cpanel.dotwhat.net" />
+    <target host="ftp.dotwhat.net" />
```
J0WI: duplicated target and mismatching cert.

ryanbr: Noted and fixed

J0WI: Thanks!

PR #18244, opened 2019-07-30
Add ruleset for dw.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/DW.com.xml
```diff
@@ -0,0 +1,45 @@
+<!--
+	Invalid certificate:
+		- multimedia.dw.com
+		- nl.dw.com
+		- webmo.dw.com
```
cschanaj: `webmo.dw.com` no longer serve an invalid certificate, would you add it to the list of `target` and update the comments?

ivysrono: No   NOT TRUSTED
https://www.ssllabs.com/ssltest/analyze.html?d=webmo.dw.com&latest

cschanaj:
thanks!

PR #18250, opened 2019-07-30
Re-enabling Cryptome.xml

Giltyhub2:


src/chrome/content/rules/Cryptome.xml
```diff
@@ -1,20 +1,18 @@
...
-  <target host="www.cryptome.info" />
+	<target host="cryptome.info" />
+	<target host="www.cryptome.info" />
 
-  <rule from="^http://(www\.)?cryptome\.info/"
-          to="https://cryptome.org/"/>
+	<securecookie host=".+" name=".+" />
 
-  <rule from="^http:" to="https:" />
+	<rule from="^http://(www\.)?cryptome\.info/" to="https://cryptome.org/"/>
```
J0WI: They do not have the same content

Giltyhub2: @J0WI yes, but it says that urls with .info should be changed to .org instead of doing the automatic redirect, I see no problem wih this approach, no?

Giltyhub2: @J0WI *ping*

J0WI: Please have a look at our policy: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects

Giltyhub2: @J0WI thanks, pushed a fix

J0WI: Thanks!

PR #18254, opened 2019-07-31
Run node utils/trivialize-cdn-rules/index.js

cschanaj:


cschanaj:
6299281 is intended to fixed the merge conflict with #18256

PR #18263, opened 2019-07-31
Revert icons back to previous state

Hainish:


Hainish:
We should rework how to present the HTTPSE icon, and introduce it along with changes to the interstitial page as well as https://www.https-rulesets.org/ and the infographic on the addons page, as well as the EFF.org HTTPSE banner graphic.  For lack of a better solution right now I'm reverting the icon to its previous state

pipboy96:
@Hainish We probably should use a SVG version of the icon when possible.

PR #18274, opened 2019-08-03
Create GitFoo.com.xml

Giltyhub2:


src/chrome/content/rules/GitFoo.com.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="GitFoo.com">
```
pipboy96: Mismatched: `mail`

Giltyhub2: @pipboy96 :heart:

PR #18282, opened 2019-08-04
Update and rename My_Healthcare_Contacts.xml to minavardkontakter.se.xml

hstseverywhere:


src/chrome/content/rules/minavardkontakter.se.xml
```diff
@@ -0,0 +1,21 @@
+<!-- See also: 1177.se.xml 
+
+		Problematic subdomains:
+			
+			- cert error: 
+				- designverktyget.sob.minavardkontakter.se
+				- personal.designverktyget.sob.minavardkontakter.se
+
+			- nxdomain:
+				- kontakt.minavardkontakter.se
```
J0WI: You can safely remove this

hstseverywhere: You mean the the nxdomain, right?

PR #18275, opened 2019-08-04
Create Ichijinsha.xml

ipaucek4680:

PR #18286, opened 2019-08-05
Create Ichijinsha.xml

ipaucek4680:
(This is a re-submission of #18275)
@pipboy96 

Note: I skipped `next.ichijinsha.co.jp` because HTTP authorization is needed and I am unable to test this domain.


src/chrome/content/rules/Ichijinsha.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Mismatched:
+		- www.shop
```
cschanaj: Would you add a comment for `next.ichijinsha.co.jp` ? Say:

```
    Authentication required:
        - next
```

ipaucek4680:
@cschanaj
I have updated the ruleset.

cschanaj:
thanks!

PR #18289, opened 2019-08-05
Create xael.org.xml

TotalCaesar659:

### DIFF
--- a/src/chrome/content/rules/xael.org.xml
+++ b/src/chrome/content/rules/xael.org.xml
@@ -1,8 +1,24 @@
+<!--
+	Invalid certificate:
+		vps3temp.xael.org
+
+	Refused:
+		vps4.xael.org
+
+	Time out:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/xael.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="xael.org">
+	<target host="xael.org" />
+	<target host="www.xael.org" />
```
J0WI: `photos.xael.org`
`stats.xael.org`

ivysrono: https://travis-ci.org/EFForg/https-everywhere/jobs/586367665#L320

Refused all?

TotalCaesar659: Yes, I think it may be closed until the site returns online.

PR #18296, opened 2019-08-07
Create tset.de.xml

TotalCaesar659:

### DIFF
--- a/src/chrome/content/rules/tset.de.xml
+++ b/src/chrome/content/rules/tset.de.xml
@@ -1,0 +1,6 @@
+<ruleset name="tset.de">
+	<target host="tset.de" />
+	<target host="www.tset.de" />
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/tset.de.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="tset.de">
```
J0WI: Needs `default_off="cert-chain"` attribute

src/chrome/content/rules/tset.de.xml
```diff
@@ -0,0 +1,6 @@
...
+	<target host="tset.de" />
+	<target host="www.tset.de" />
```
J0WI: `hg.tset.de`

PR #18303, opened 2019-08-09
Create Iraqueer.org.xml

Bisaloo:


pipboy96:
There are also `ar-iraqueer.org` and `ku-iraqueer.org`. Please also try `mail` sybdomain.

pipboy96:
@Bisaloo Can you make rulesets for Arabic and Kurdish versions of the website?

Bisaloo:
I've added everything in a single ruleset as the number of subdomains is really limited but I can split them if necessary.

src/chrome/content/rules/Iraqueer.org.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Time out:
+		mail.iraqueer.org
+		mail.ar-iraqueer.org
+
```
pipboy96: `mail.ku-iraqueer.org`?

Bisaloo: Seems broken over both http and https

PR #18304, opened 2019-08-09
Run hsts-prune and update impacted rulesets

Bisaloo:


src/chrome/content/rules/Marx.org.xml
```diff
@@ -0,0 +1,15 @@
...
+-->
+<ruleset name="Marx.org">
+
+	<target host="marx.org" />
+	<target host="www.marx.org" />
+
+	<rule from="^http://(www\.)?marx\.org/"
+		to="https://www.marxists.org/" />
+
+</ruleset>
```
pipboy96: Is it possible to remove this redirect?

src/chrome/content/rules/Nstatic.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="Nstatic.net">
+
+	<target host="nstatic.net" />
+
+	<rule from="^http:" to="https:" />
+
+</ruleset>
```
pipboy96: Are there any subdomains?

Bisaloo: I'd rather keep the changes in this PR as minimal as possible. Only run hsts-prune and fix the result. Enhancements can take place in future PRs

pipboy96: I will make such PR after this is merged.

src/chrome/content/rules/Marx.org.xml
Bisaloo: I've removed this ruleset entirely as the redirect wasn't correct. See 17f3362

src/chrome/content/rules/Nstatic.com.xml
ivysrono: merged.

PR #18309, opened 2019-08-10
Create surina.net.xml

TotalCaesar659:


src/chrome/content/rules/surina.net.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="surina.net">
+	<target host="surina.net" />
+	<target host="www.surina.net" />
```
J0WI:
```
audioshift.surina.net
soundtouch.surina.net
```

PR #18310, opened 2019-08-12
Update acast.com (fixes bbc.co.uk podcast playback)

pipboy96:
High priority since current ruleset breaks bbc.co.uk (Alexa rank 109).

### DIFF
--- a/src/chrome/content/rules/acast.com.xml
+++ b/src/chrome/content/rules/acast.com.xml
@@ -16,11 +16,9 @@
 	<target host="flex.acast.com" />
 	<target host="media.acast.com" />
 	<target host="rss.acast.com" />
-	<target host="status.acast.com" />
 
 	<securecookie host=".+" name=".+" />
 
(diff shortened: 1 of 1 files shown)


pipboy96:
To speed up review I decided to not add any new subdomains.

src/chrome/content/rules/acast.com.xml
```diff
@@ -1,64 +1,26 @@
...
-	<rule from="^http://acast\.com/"
-		to="https://www.acast.com/" />
-
-	<rule from="^http://status\.acast\.com/"
-		to="https://acast.statuspage.io/" />
-
-	<rule from="^http:"
-		to="https:" />
-
+	<rule from="^http://status\.acast\.com/" to="https://acast.statuspage.io/" />
```
cschanaj: I think we should avoid cross-domain rewrites in new rulesets, would you remove this? otherwise lgtm.

pipboy96: @cschanaj Done.

PR #18351, opened 2019-08-28
Add download.schneider-electric.com host target

wesinator:


src/chrome/content/rules/Schneider-Electric.com.xml
```diff
@@ -71,5 +72,8 @@
 
 	<rule from="^http://(?:www\.)?schneider-electric\.com/(download/|navigation/|(?:product|solution)s(?:$|[?/])|resource/|site/(?:home|marketing_center)/(?:image|j)s/|site/tasks/sites/|stat/)"
 		to="https://www.schneider-electric.com/$1" />
```
J0WI: Could you also remove the obsolete exclusions and the comments above?

wesinator: shouldn't these updates be done under a separate PR ?

J0WI: Would be great if you can fix it here. This kind of regex has some performance impact and we can also get rid of the whitelist entry.

J0WI:
Thanks!

PR #18372, opened 2019-09-03
Add rulesets telenet.xml and cinsscore.xml

karelvanhecke:


karelvanhecke:
telenet.xml:

`<rule from="^http://www\.telenet\.be/"
            to="https://www2.telenet.be/" />`

This rule is required because https://www.telenet.be redirects to http://www.telenet.be, which would cause a loop of redirections when upgraded to HTTPS by HTTPS everywhere.

cinsscore.xml:

Certificate is only valid for `cinsscore.com`, not for the `www` subdomain, nor `ciarmy.com`.
All should be redirected to `cinsscore.com`.

Including HTTPS requests to `(www\.)?ciarmy\.com` and `www\.cinsscore\.com`. But rule checks seem to fail when trying to redirect HTTPS requests. (any insight on this?)

ivysrono:
Please split them to two PRs.

PR #18373, opened 2019-09-04
Update hsbc.com.hk.xml

cschanaj:
add www.file.online-banking.hsbc.com.hk

### DIFF
--- a/src/chrome/content/rules/hsbc.com.hk.xml
+++ b/src/chrome/content/rules/hsbc.com.hk.xml
@@ -8,9 +8,6 @@
 		- assetmanagement.hsbc.com.hk
 		- business.hsbc.com.hk
 
-		SSL error:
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/hsbc.com.hk.xml
```diff
@@ -8,9 +8,13 @@
 		- assetmanagement.hsbc.com.hk
 		- business.hsbc.com.hk
 
+		SSL error:
+		- www.security.online-banking.hsbc.com.hk
```
J0WI: This might be specific to Travis, I'm not able to reproduce it. Would you like to try again and whitelist it?

cschanaj: Look like the problem is fixed on the server side after the build. thanks.

PR #18376, opened 2019-09-04
Usf patch

cwbell-usf:
Updated the University of South Florida rulesets with all known protected public-facing URLs.


cwbell-usf:
Legacy (broken) rules are breaking that should work - investigating and will open new request.

PR #18378, opened 2019-09-04
Usf patch

cwbell-usf:
Updated URLs for usf.edu with all known protected public-facing sites. Three sites are excluded (see comments) due to configuration issues. Contacting USF IT to check on possible resolution.


cwbell-usf:
missed one...

PR #18379, opened 2019-09-04
Add timus.online.xml

pipboy96:


src/chrome/content/rules/timus.online.xml
```diff
@@ -0,0 +1,14 @@
+<!--
+    Mismatched:
+        - www
+        - acm
+        - ftp
+        - mirror
+-->
+<ruleset name="timus.online">
+    <target host="timus.online" />
+    <target host="acm.timus.ru" />
```
cschanaj: Would you split `timus.ru` into a separate ruleset and add comments on the ssl error? otherwise lgtm

pipboy96: I remember you said before that cross-origin redirects are now discouraged. Should I remove this target completely? There is nothing on `timus.ru` that can be rewritten otherwise.

cschanaj: Sure. It would be ideal if you do not mind to remove `timus.ru` completely.

pipboy96: @cschanaj done.

cschanaj:
thanks!

PR #18380, opened 2019-09-04
Usf patch

cwbell-usf:
Updates to make the USF rule functional and add all known public-facing domains. Three sites remain to be resolved, working with USF IT to determine cause.


pipboy96:
@cwbell-usf Thanks for contribution! May I ask if you are a USF employee? If yes, I recommend using HSTS preloading, which would make this ruleset unnecessary or at least make it much simpler.

cwbell-usf:
I am, and though I can't quite go into why (opsec) we currently can't roll out hsts for most of the domains. Things move slow in an org this big, so this is a stopgap until we can enforce natively.

src/chrome/content/rules/University-of-South-Florida.xml
```diff
@@ -1,68 +1,51 @@
...
-	<target host="vpn.usfsp.edu" />
+	<target host="www.cutr.usf.edu" />
+	
+	<target host="health.usf.edu" />
+	
 	<target host="www.usfsp.edu" />
 	<target host="sharemypc.usfsp.edu" />
+	
+	<target host="wusf.usf.edu" />
+	<target host="www.wusf.org" />
```
J0WI: `wusfnews.wusf.usf.edu`

src/chrome/content/rules/University-of-South-Florida.xml
```diff
@@ -1,68 +1,51 @@
...
 <!--
-Disabled by https-everywhere-checker because:
-Fetch error: http://usfsp.edu/ => https://usfsp.edu/: (51, "SSL: no alternative certificate subject name matches target host name 'usfsp.edu'")
-
-Disabled by https-everywhere-checker because:
-Fetch error: http://usfsp.edu/ => https://usfsp.edu/: (51, "SSL: no alternative certificate subject name matches target host name 'usfsp.edu'")
-	For problematic rules, see University-of-South-Florida-mismatches.xml.
-
-
-	Nonfunctional usf.edu subdomains:
```
J0WI: Please take care that you're not removing any information that is still valid.

src/chrome/content/rules/University-of-South-Florida.xml
cwbell-usf: That domain has been deprecated and will no longer be maintained

src/chrome/content/rules/University-of-South-Florida.xml
cwbell-usf: I wasn't aware that I did? I removed rules that do/will not have valid SSL protection on them or that are no longer in production/do not exist/are not maintained. 

The usfsp.edu no SAN one in particular will not be resolved.

src/chrome/content/rules/University-of-South-Florida.xml
J0WI: Would be good to protect it as long as it's still up.

src/chrome/content/rules/University-of-South-Florida.xml
J0WI: E.g. `system.usf.edu` is still a a non-functional target. We use to document them to make sure that we haven't missed them: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts

cwbell-usf: Good point... Ok, I guess I'll work on documenting those and resub a new PR later.

cwbell-usf:
I'll open a new one when I have more documentation.

PR #18386, opened 2019-09-06
Updating lib-wasm library

Hainish:


pipboy96:
@Hainish Should I try reproducing?

Hainish:
@pipboy96 only if you want, the tests are in place in the lib-wasm library and travis has reproduced.

pipboy96:
@Hainish So reproducing is now automatic? Cool. I will reproduce anyway and post the result though.

pipboy96:
@Hainish Reproduced!

PR #18388, opened 2019-09-06
Add rule for e-hentai.org

PRB0t:
Thanks to all the reviewers!
--
Automated submit by PRB0t

PR #18393, opened 2019-09-09
Add E-Hentai.org.xml

pipboy96:
See also #18388.

### DIFF
--- a/src/chrome/content/rules/E-Hentai.org.xml
+++ b/src/chrome/content/rules/E-Hentai.org.xml
@@ -1,0 +1,34 @@
+<!--
+	Cloudflare error:
+		- g
+
+	Timeout:
+		- monitor
+		- shingata
+		- test
+
+	Mismatched:
+		- wiki
+-->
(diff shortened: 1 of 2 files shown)


src/chrome/content/rules/E-Hentai.org.xml
```diff
@@ -0,0 +1,34 @@
...
+	<target host="r.e-hentai.org" />
+	<target host="repo.e-hentai.org" />
+	<target host="ul.e-hentai.org" />
+	<target host="ulv.e-hentai.org" />
+	<target host="upload.e-hentai.org" />
+	<target host="wiki.e-hentai.org" />
+	<target host="xml.e-hentai.org" />
+
+	<rule from="^http://g\.e-hentai\.org/" to="https://e-hentai.org/" />
+	<rule from="^http://wiki\.e-hentai\.org/" to="https://ehwiki.org/" />
```
pipboy96: These two rules may be removed, but both of them mirror website's own 301 redirects.

PR #18396, opened 2019-09-09
Create meldmerge.org.xml

TotalCaesar659:

### DIFF
--- a/src/chrome/content/rules/meldmerge.org.xml
+++ b/src/chrome/content/rules/meldmerge.org.xml
@@ -1,0 +1,9 @@
+<!--
+        Invalid certificate:
+                www.meldmerge.org
+-->
+<ruleset name="meldmerge.org">
+	<target host="meldmerge.org" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/meldmerge.org.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+        Invalid certificate:
+                www.meldmerge.org
```
pipboy96: Do not mix spaces and tabs please.

src/chrome/content/rules/meldmerge.org.xml
```diff
@@ -0,0 +1,9 @@
...
+-->
+<ruleset name="meldmerge.org">
+	<target host="meldmerge.org" />
+
+	<rule from="^http:" to="https:" />
```
pipboy96: Please do this in order to also cover `www`.

```
	<target host="meldmerge.org" />
	<target host="www.meldmerge.org" />

	<rule from="^http://www\.meldmerge\.org/" to="https://meldmerge.org/" />
```

src/chrome/content/rules/meldmerge.org.xml
TotalCaesar659: Sorry, wrong copy.

PR #18410, opened 2019-09-18
Add rule for RustFest.eu

dns2utf8:

- some subdomains do not always redirect to https


pipboy96:
Please enumerate subdomains using [Amass](https://github.com/OWASP/Amass).

dns2utf8:
Thank you for running the amass tool, I have never heard of it.
May I ask why you would like to have it's output?

pipboy96:
You should not use a wildcard but list individual domains and document the ones that do not work.

dns2utf8:
Thank you :)

Issue #17867, opened 2019-05-03
Usability issues with new chrome extension layout


catthou:
Type: feature request

I am using https everywhere chrome extension 2019.5.2.1 and I have two big usability concerns with it.

1) "Disable https everywhere on this site" should be a top-level menu item. I try to keep EASE on all the time and I need to disable https for specific sites more often than I like to admit. Right now it's a three-button pain to open (extension button -> See more -> Disable https everywhere on this site), and another three to enable it again when I'm done on the site. 6 total presses vs the 4 from the earlier version of the extension.

2) The "See more" and down arrow accompanying it need to have more clickable area. They're rather difficult to click on, and on top of that there's a gap of unclickable area between the text and the button image.

3) When encryption is disabled via "Disable https everywhere on this site" the chrome icon no longer turns red - it's just a bland, unnoticeable gray. This is disturbing as I like to be notified when a website is using unencrypted traffic. The red was easier to notice and didn't clash with my darker chrome theme, where was the bland gray is easy to miss, and I worry I might forget to turn it back on for that site when I'm done.

Edit: added third problem.

zoracon:
Thank you for the early feedback.

For #3, the "Disable https everywhere on this site" was always gray, the red was actually for EASE mode. Which is apart of the reason we leaned away from that. Due to the fact "red" indicated a negative connotation.

KyleSanderson:
Yeah this is awful... Can we please get the old UI back @zoracon ?

zoracon:
> Yeah this is awful... Can we please get the old UI back @zoracon ?

If you could list specific features you do not like, that would be more helpful. If you generally don't like the UI at all, then not much can be helped there besides through constructive suggestion. This is from months of UX and UI work with the community that was publicly discussed here: https://github.com/EFForg/https-everywhere/issues/16669

KyleSanderson:
@zoracon There's crazy whitespace... This makes the panel nearly completely 100% unusable, my ancient bank did similar years ago and they received so much pushback they had to restart (and compress the whitespace). Checkout PrivacyBadger (which still has a lot of whitespace... but is 300% better).

![Worst thing ever to be witnessed](https://i.imgur.com/03O6DAM.png)
![Slightly Better](https://i.imgur.com/2lVJrED.png)

Please revert this immediately.

jadchaar:
[...]

PR #18417, opened 2019-09-19
Feature: Navigate to HTTP in EASE Once

### LINKED ISSUE #16998: Firefox http allow once
Type: feature request

When using "Block all unencrypted requests" and viewing a page that can only be viewed in http, it would be nice to be able to allow the http once/1hr/24h/this session only/etc vs just adding it permanently to a whitelist.

zoracon:
Closes #16998 

This is a first try at implementing this feature. Still in review on what this looks like in the front-end.

Functionally, this is based on the window closing, not the tab. I had contemplated tab based closings, but decided that it's probably better for windows based closure instead.

Note: Windows based is likely more convenient, but less safe than per tab. However since we already have a permanently managed list, that reasoning seems probably unnecessary now.


chromium/background-scripts/background.js
```diff
@@ -114,6 +115,9 @@ if (chrome.windows) {
   chrome.windows.onFocusChanged.addListener(function() {
     updateState();
   });
+
+  // Grant access to HTTP site only during session, clear once window is closed
+  chrome.windows.onRemoved.addListener( () => httpOnceList.clear() );
```
pipboy96: How would this behave with multiple windows? Shouldn't it apply for entire browsing session?

chromium/background-scripts/background.js
```diff
@@ -147,7 +151,7 @@ function updateState () {
     }
     const tabUrl = new URL(tabs[0].url);
 
-    if (disabledList.has(tabUrl.host) || iconState == "disabled") {
+    if (disabledList.has(tabUrl.host) || httpOnceList.has(tabUrl.host) || iconState == "disabled") {
```
pipboy96: Is host guaranteed to be lowercase, punycoded etc?

pipboy96:
Good overall.

chromium/background-scripts/background.js
zoracon: The very nature of the URL JavaScript object doesn't necessarily seem to enforce this ask, but since it grabs the host, which are modeled to be case insensitive, this has been left as is.

chromium/background-scripts/background.js
zoracon: I can try to manage by window ID and flag if the window is incognito I believe. However to fully manage sessions the extension would need the sessions permission, but I believe I can programmatically manage with the windows API alone.

chromium/background-scripts/background.js
pipboy96: Do other parts of the code do the same? If yes, this should be fine for now.

chromium/background-scripts/background.js
pipboy96: I think we should investigate split incognito mode. Regular and incognito parts of the extension can't communicate directly but they share storage.

chromium/background-scripts/background.js
zoracon: Yes, the URL object has been utilized this way. Will keep as is for now.

chromium/background-scripts/background.js
zoracon: Agreed, will finish this piece up tomorrow.

chromium/background-scripts/background.js
```diff
@@ -654,11 +664,21 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse) {
...
-  function storeDisabledList() {
+  function storeDisabledList(message) {
+
     const disabledListArray = Array.from(disabledList);
-    store.set({disabledList: disabledListArray}, () => {
-      sendResponse(true);
-    });
+    const httpOnceListArray = Array.from(httpOnceList);
+
+    if ('once' === message) {
```
cschanaj: Is there any reason to store the session-based exception list to browser storage? A similar purposed per-session only `urlBlacklist` is also implemented without using persistent storage.

chromium/pages/cancel/style.css
```diff
@@ -47,8 +47,12 @@ form button{
   text-overflow: ellipsis;
 }
 
-#open-url-button{
-  display: inline-block;
-  float: left;
+#open-url-button, #http-once-button {
```
cschanaj: Is there any follow up on the UX redesign project mentioned in https://github.com/EFForg/https-everywhere/issues/17867#issuecomment-489494058 ? We have introduced a lots of new UI elements since the discussion in #16669, it would be great to have further updates on it. 

Sorry for derailing from the PR here.

pipboy96: @cschanaj I want to try improving the design but not sure if I will have enough time for that.

cschanaj: @pipboy96 I have some opinions on the current design too but I would like to improve it without affecting EFF's plan. Some issues like #18482 could be easily resolve with a grid system as mentioned in #17867 . That's why I am asking for a update on the issue. :smiley:

zoracon: @pipboy96 @cschanaj As far as design, I am leaving that to our internal design team when we wrap up things with Ura Design Labs, who is currently doing great work on UX research and helping to incorporate a design system that makes the most sense.

You can track design updates here for now: https://github.com/uracreative/works/issues/29

pipboy96: @zoracon Is there some chat work on this design happens on (besides this issue)?

zoracon: @pipboy96 we discuss with Ura labs through their chat but mainly they are taking formal feedback on the issue. Since they wanted a central place for updates and changes.

pipboy96: @zoracon Is there any decision on how we are going to implement the design technically?

zoracon: Still awaiting a response, while I wait I am going to just push up the code amendments I have.

chromium/background-scripts/background.js
zoracon: I had it set up this way due to the fact this interfaces with the user and I needed the messaging system.

pipboy96: @zoracon not sure if keeping Yoda comparisons is a good idea, confusing comparison with assignment is almost impossible in JS especially with use of `===`.

chromium/background-scripts/background.js
```diff
@@ -564,7 +571,10 @@ function onHeadersReceived(details) {
 
     // Do not upgrade resources if the first-party domain disbled EASE mode
     // This is needed for HTTPS sites serve mixed content and is broken
-    if (disabledList.has(browserSession.getTab(details.tabId, 'first_party_host', null))) {
+    if (
+    disabledList.has(browserSession.getTab(details.tabId, 'first_party_host', null)) ||
+    httpOnceList.has(browserSession.getTab(details.tabId, 'first_party_host', null)))
```
pipboy96: These two need two more spaces.

.github/pull_request_template.md
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
pipboy96: This is definitely not supposed to be here.

chromium/background-scripts/background.js
zoracon: For now I have if all windows are closed regardless of type, since Incognito splitting proposed some issues I want to iron out.

PR #18422, opened 2019-09-21
Add welmers.net.xml

bast1aan:
Added rules for domains running webservers that I own.

### DIFF
--- a/src/chrome/content/rules/welmers.net.xml
+++ b/src/chrome/content/rules/welmers.net.xml
@@ -1,0 +1,18 @@
+<ruleset name="welmers.net">
+
+	<!--	Direct rewrites:
+				-->
+	<target host="welmers.net" />
+	<target host="www.welmers.net" />
+	<target host="www4.welmers.net" />
+	<target host="www6.welmers.net" />
+	<target host="users.welmers.net" />
+	<target host="gallery.welmers.net" />
+
+
+	<securecookie host="^(www.?|users|gallery)\.welmers\.net$" name=".+" />
+
+	<rule from="^http:"
+		to="https:" />
+
+</ruleset>


pipboy96:
@bast1aan Thanks! Since you are a website owner, you may use [HSTS preloading](https://hstspreload.org/) instead. It's more effective since it doesn't require the user to have HTTPS Everywhere installed. The user just needs to use a modern version of Chrome or Firefox.

bast1aan:
Hi @pipboy96 unfortunately the domain is also running services that may be included in http-only sites or used for http-only traffic. Because many applications run on the hosts, some not maintained by myself, I cannot set HSTS on the domains at this moment, or force to redirect to https. So I'd like to leave it to the user if they want to be forced to use https or not, by submitting this information to https-everywhere.

pipboy96:
@bast1aan Can you tell which services should allow unencrypted connections so we could work something out? On the opposite, are there parts that should always be redirected to HTTPS?

bast1aan:
Hi, services on the domains that should use https are already forced and redirected to https, mainly internal services requiring logins. 

The things that I would like to leave to the user are for instance public_html directories served by users.welmers.net. I also have a few subdomains (not mentioned in the xml file) on servers that run on port 80 only because of restrictions in the underlying infrastructure. They do not run services requiring login so I currently have no urgency fixing this.

I also run a cached CentOS and Debian mirror for internal usage that should stay on http and a couple of APIs with customers that request them from http because they still haven't fixed it. (the parts that do not require user credentials of course) I currently do not want to force them to use https.
Another problem is sites that still contain http:// links and would break if it is forwarded to https (forms for instance). 

I know, old legacy crap mainly, but I won't get it fixed very soon, or rather remove it as soon as I could instead of fixing it.

The only thing I could come up with that could be made fully https by myself is gallery.welmers.net maybe.

bast1aan:
BTW is there a way I can fix the Travis CI? I can't see what is going wrong exactly.

pipboy96:
@bast1aan It's a random bug. I don't know how to fix it.

src/chrome/content/rules/welmers.net.xml
```diff
@@ -0,0 +1,18 @@
...
+				-->
+	<target host="welmers.net" />
+	<target host="www.welmers.net" />
+	<target host="www4.welmers.net" />
+	<target host="www6.welmers.net" />
+	<target host="users.welmers.net" />
+	<target host="gallery.welmers.net" />
+
+
+	<securecookie host="^(www.?|users|gallery)\.welmers\.net$" name=".+" />
```
pipboy96: Can you send the cookies with `Secure` flag on the server side? If HTTPS Everywhere rewrites a cookie to include `Secure` flag and then user uninstalls HTTPS Everywhere the flag will not be removed and it won't be available on insecure version of website.

PR #18431, opened 2019-09-23
Create babr24.com.xml

TotalCaesar659:


src/chrome/content/rules/babr24.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="babr24.com">
+	<target host="babr24.com" />
+	<target host="www.uvnc.com" />
```
pipboy96: `www.babr24.com`?

TotalCaesar659: Yes, forgot about my template.

PR #18435, opened 2019-09-24
Create Embedresponsively.com.xml

ryanbr:


ryanbr:
Was requested by a user in our forum;

https://community.brave.com/t/upgrade-connections-to-https-not-working/84298/3

src/chrome/content/rules/Embedresponsively.com.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+    Non-functional hosts:
+       Timeout:
+       - ssh.embedresponsively.com
```
Bisaloo: I get invalid certificate here as well, not a time out. Could you double check please?

ryanbr: Hmm I got a time out last I check, but got a different response from the peer now. Adjusted comment.

PR #18461, opened 2019-09-29
Remove duplicates code.highcharts.com

cschanaj:


src/chrome/content/rules/Highcharts.com.xml
```diff
@@ -1,39 +1,20 @@
...
+		- highcharts.com
 -->
-
 <ruleset name="Highcharts.com (partial)">
 	<target host="highcharts.com" />
-	<rule from="^http://highcharts\.com/" to="https://www.highcharts.com/" />
-
 	<target host="www.highcharts.com" />
+	<target host="api.highcharts.com" />
 	<target host="cloud.highcharts.com" />
```
J0WI: `code.highcharts.com`
Otherwise this target would be removed completely.

PR #18466, opened 2019-09-30
Update RickyGervais.com.xml

TotalCaesar659:


ivysrono:
I still get MCB because of http://www.rickygervais.com/beta_rg.css

pipboy96:
I guess for now this website is not fixable.

PR #18467, opened 2019-09-30
Create logicalincrements.com.xml

TotalCaesar659:


src/chrome/content/rules/logicalincrements.com.xml
```diff
@@ -0,0 +1,21 @@
...
+	Redirect to HTTP:
+		logicalincrements.com
+
+	Refused:
+		blog.logicalincrements.com
+
+	Time out:
+		a.logicalincrements.com
+-->
+<ruleset name="logicalincrements.com">
```
pipboy96:
```
	<target host="logicalincrements.com" />
```

src/chrome/content/rules/logicalincrements.com.xml
```diff
@@ -0,0 +1,21 @@
...
+	<target host="www.logicalincrements.com" />
+	<target host="backend.logicalincrements.com" />
+	<target host="backend2.logicalincrements.com" />
+	<target host="images.logicalincrements.com" />
+
```
pipboy96:
```
	<rule from="^http://logicalincrements\.com/"
		to="https://www.logicalincrements.com/" />
``

File: src/chrome/content/rules/logicalincrements.com.xml (hunk 1 above)
Comment (pipboy96) [reply src/chrome/content/rules/logicalincrements.com.xml]: This one is still needed for this rule to work.

File: src/chrome/content/rules/logicalincrements.com.xml [hunk 3]
```diff
@@ -0,0 +1,20 @@
+<!--
```
pipboy96: Keep the comment, just make it like this:

```
	Redirects to HTTP:
		- ^ (fixed by this ruleset)
```

PR #18471, opened 2019-10-01
Add ruleset for my-aime.net

blin00:
Note: `my-aime.net` has HSTS for subdomains, but it isn't preloaded and it doesn't set secure cookies.

### DIFF
--- a/src/chrome/content/rules/My-Aime.net.xml
+++ b/src/chrome/content/rules/My-Aime.net.xml
@@ -1,0 +1,11 @@
+<!--
+	Timeout:
+		- mx
+-->
+<ruleset name="My-Aime.net">
+	<target host="my-aime.net" />
+
+	<securecookie host=".+" name=".+" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


src/chrome/content/rules/My-aime.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="my-aime.net">
+    <target host="my-aime.net" />
+    <target host="*.my-aime.net" />
```
pipboy96: There are no subdomains, even `www`.

src/chrome/content/rules/My-aime.net.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="my-aime.net">
```
pipboy96:
```
<!--
	Timeout:
		- mx
-->
```

src/chrome/content/rules/My-aime.net.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Timeout:
+		- mx
+-->
+<ruleset name="my-aime.net">
```
Bisaloo:
```suggestion
<ruleset name="My-aime.net">
```

To match the file name and `ruleset` `name`.

blin00: done (capitalized the A as well)

PR #18477, opened 2019-10-02
Add jabref.org

koppor:
This adds a simple ruleset for https://www.jabref.org/


pipboy96:
Please use [Amass](https://github.com/OWASP/Amass) to enumerate subdomains instead of using wildcard records in the future.

koppor:
Thank you for updating the PR 👍.

PR #18480, opened 2019-10-03
Fix redirect on feedback.ebay.co.uk

ryanbr:
User was having an issue feedback not using/redirecting using https: 

https://community.brave.com/t/non-https-links-not-loading-pages/85838

1. `https://www.ebay.co.uk/itm/401786990949`
2. On the right-side, under Seller Information.
3. Click on the Seller number '54388'
4. `http://feedback.ebay.co.uk` is attempted.

Does this patch look okay?


ryanbr:
User indicated this was resolved on their end.

Issue #18482, opened 2019-10-03
broken popup layout when language is set to Chinese


gloomy-ghost:
Type: code issue

![图片](https://user-images.githubusercontent.com/13086421/66163149-290c6980-e68c-11e9-84cc-d81d9f9f5685.png)

cc: @ivysrono

pipboy96:
cc @zoracon.

ivysrono:
Mobile +1
![图片](https://user-images.githubusercontent.com/2649544/66176816-bbfbd280-e691-11e9-918e-ff71da706fe1.png)

PR #18495, opened 2019-10-06
 Fix styles when display language is not English, issue #18482 

cschanaj:
This PR fixes the UI distortion when the display language is not English

- #18482

### DIFF
--- a/chromium/pages/popup/style.css
+++ b/chromium/pages/popup/style.css
@@ -76,7 +76,6 @@ footer a.button {
 
 /* Elements */
 .settings_block {
-  display: inline-block;
   padding: 5px 1em 20px 1em;
   border-bottom: var(--light-grey) solid 1px;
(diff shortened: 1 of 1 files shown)


chromium/pages/popup/style.css
```diff
@@ -76,7 +76,6 @@ footer a.button {
 
 /* Elements */
 .settings_block {
```
zoracon: Seeing a display issue with this branch where the grey horizontal line is now crossing the "see more" text with English. Going to see if this can be amended for both by aligning `display: block` with padding.
image

cschanaj: I added additional `padding` to to make the "See more" element align better in ea8ee20

cschanaj:
@zoracon would you take a look at the fix and merge this possibly before the next release?

PR #18524, opened 2019-10-11
Add rule for randaris.app

pointhi:
randaris.app should be https only now


pipboy96:
`.app` domains are already HTTPS-only.

pointhi:
ok, didn't knew that

PR #7392, opened 2016-10-20
Add basic ruleset for jku.at subdomains

pointhi:
Adding some important subdomains of jku.at which should be used with https only.

PR #18525, opened 2019-10-11
Add partial ruleset for JKU.at

pointhi:
This list contains at least a set of domains which should always be encrypted.

### List related PRs if any

- successor of #7392


J0WI:
Thanks!

PR #18528, opened 2019-10-12
Create smspower.org.xml

TotalCaesar659:


src/chrome/content/rules/smspower.org.xml
```diff
@@ -0,0 +1,23 @@
...
+		pop.smspower.org
+		imap.smspower.org
+		blog.smspower.org
+		smtp.smspower.org
+-->
+<ruleset name="smspower.org">
+	<target host="smspower.org" />
+	<target host="www.smspower.org" />
+
+	<rule from="^http://smspower\.org/" to="https://www.smspower.org/" />
```
J0WI: Please add a comment why this rule is required

J0WI: @TotalCaesar659 could you fix this too?

PR #18563, opened 2019-10-15
Add unit tests for util.js

cschanaj:


chromium/test/util_test.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const assert = require('chai').assert,
+  utils = require('../background-scripts/util');
+
+
+describe('utils.js', function() {
+  describe('getNormalisedHostname', function() {
+    it("removes tailing dots", function() {
+      assert.strictEqual(utils.getNormalisedHostname("example.com."), "example.com");
```
pipboy96: Please don't mix `'` and `"`.

cschanaj: fixed. thanks :smile:

PR #18565, opened 2019-10-15
Ignore loopback and private network addresses (such as 192.168.0.1)

### LINKED ISSUE #18474: Exclude all local subnets from HTTPSE
[//]: #

Type: code issue

Coming in from email:

> Hello,
> I would like to know if HTTPS everywhere can be disabled globally for private ip addresses (10.0.0.0/8 172.16.0.0/12 192.168.0.0/16) since they don't need https most of the time being 
in the same subnet and it would be very inconvenient to disable HTTPS everywhere for every single ip address.

What do people think about this request?

pipboy96:
Closes #18474.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -8,7 +8,8 @@ const rules = require('./rules'),
   util = require('./util'),
   update = require('./update'),
   { update_channels } = require('./update_channels'),
-  wasm = require('./wasm');
+  wasm = require('./wasm'),
+  ipUtils = require('./ip_utils');
 
 
 let all_rules = new rules.RuleSets();
@@ -304,6 +305,14 @@ function onBeforeRequest(details) {
   // Normalise hosts with tailing dots, e.g. "www.example.com."
   uri.hostname = util.getNormalisedHostname(uri.hostname);
 
+  let ip = ipUtils.parseIp(details.hostname);
+
+  let isLocalIp = false;
+
+  if (ip !== -1) {
+    isLocalIp = ipUtils.isLocalIp(ip);
+  }
+
   if (details.type == "main_frame") {
     // Clear the content from previous browser session.
     // This needed to be done before this listener returns,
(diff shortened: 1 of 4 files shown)


chromium/background-scripts/background.js
```diff
@@ -327,9 +336,9 @@ function onBeforeRequest(details) {
     (uri.protocol === 'http:' || uri.protocol === 'ftp:') &&
     uri.hostname.slice(-6) !== '.onion' &&
     uri.hostname !== 'localhost' &&
-    !/^127(\.[0-9]{1,3}){3}$/.test(uri.hostname) &&
     uri.hostname !== '0.0.0.0' &&
```
pipboy96: @zoracon Should this be moved into `ip_utils`?

chromium/test/ip_utils_test.js
```diff
@@ -0,0 +1,82 @@
...
+      assert(parseIp('1.1.1.1...') === -1)
+    })
+
+    it('rejects an IP address-like string with omitted parts', () => {
+      assert(parseIp('.1.1.1') === -1)
+      assert(parseIp('1..1.1') === -1)
+      assert(parseIp('1.1..1') === -1)
+      assert(parseIp('1.1.1.') === -1)
+    })
+
```
cschanaj:
```suggestion
    it('rejects an IP address-like string with invalid parts', () => {
      assert(parseIp('192.168.1.256') === -1)
      assert(parseIp('192.168.256.1') === -1)
      assert(parseIp('192.256.1.1') === -1)
      assert(parseIp('256.168.1.1') === -1)
      assert(parseIp('256.168.1.-1') === -1)
    })

```

cschanaj: not sure if this help to improve the coverage to 100%

pipboy96:
@cschanaj I think there is no rule about use of `;`, but @zoracon would be a better person to ask.

pipboy96:
@cschanaj Thanks! It now has 100% coverage.

chromium/background-scripts/background.js
zoracon: Go ahead and move that into `ip_utils`

zoracon:
> @cschanaj I think there is no rule about use of `;`, but @zoracon would be a better person to ask.

@pipboy96 let's stick to using semicolons since I want as much consistency and readability as we can in the repo. There was no hard rule in place but it is general practice. The file is generally clean but want to avoid future potential debugging issues like concatenated statements.

pipboy96:
@zoracon any idea what causes this error?

zoracon:
@pipboy96 Merged in that web-crypto temporary fix. Will investigate more

taokann:
Hello, I searched the web for this issue and found here that's already resolved.
Unfortunately I have the right version (2019.11.7) but I still have to manually whitelist all local IPs.
I read the PR's code, I don't know the project well but it seems that the feature is only added in the Chromium version. Is it right ? Will it be available for Firefox too ? I'm using Firefox 68.3.0 ESR with Debian Buster.

pipboy96:
@taokann Please wait until the next HTTPS Everywhere extension release, which would include #18755 patch that fixes a typo in this PR.

PR #18567, opened 2019-10-16
Create grafica.app.xml

jumde:
Adding ruleset for grafica.app

### DIFF
--- a/src/chrome/content/rules/grafica.app.xml
+++ b/src/chrome/content/rules/grafica.app.xml
@@ -1,0 +1,8 @@
+<!--
+	NXDOMAIN: www
+-->
+<ruleset name="grafica.app">
+	<target host="grafica.app" />
+
+	<rule from="^http:" to="https:" />
+</ruleset>


src/chrome/content/rules/grafica.app.xml
```diff
@@ -0,0 +1,8 @@
+<!--
+grafica.app/ 
+-->
```
zoracon: Could you please clarify this comment?

jumde: @zoracon I added this to highlight that the rule set will apply to all `grafica.app` urls, I can remove it if it feels unnecessary.

pipboy96: @jumde Yes, it's not really needed here.

jumde: @pipboy96 @zoracon - Updated.

src/chrome/content/rules/grafica.app.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="grafica.app">
```
pipboy96: Add before this line:

```
<!--
	NXDOMAIN: www
-->
```

jumde: @pipboy96 - Added

jumde:
Thanks for the approval @pipboy96

Issue #18055, opened 2019-06-04
3-state logic per-site: encrypt everything, apply rulesets, disable


mikhaelkh:
Type: code issue

There's 3 possible states of HTTPS Everywhere: EASE, rulesets, disabled.
But only 2 of them are available at any given moment: either rulesets & disabled or EASE & disabled.
I want to use EASE by default, and when it breaks the site, I don't want to completely disable HTTPS Everywhere on the site straightaway, I want to apply rulesets on the site. And then if rulesets are breaking the site, only then I want to disable extension on the site until the ruleset issue is being resolved.

Bisaloo:
Thanks for your report!

This sounds like a duplicate of https://github.com/EFForg/https-everywhere/issues/17554. Could you please post there? It may help prioritize this issue.

Issue #18069, opened 2019-06-07
Possible extension refresh?


Saiv46:
Type: other
There's still a bunch of open and old issues which is not still resolved/finished.

The release process: #2697
Improving UI/UX: #1291, #17868
Performance:
#12232: Extension excessive memory usage (applies to Chrome too)
#16551: Use more performant IndexedDB instead of serialized WebStorage

Solutions:
* Refactor *everything* in a new branch and migrate rulesets into other repo.
* Rewrite code and make modules for most browsers and environments (see #17268).

pipboy96:
Can you explain in more details what do you have in mind?

Saiv46:
~~Sorry for the long response, my laptop charger just died.~~

In details:

* #981: There was a discussion about improving the release process, does it was resolved in release #14730?
* #2697: Why there's still no separate repository for only the rulesets? As I see, the migration process will not be so smooth for the ruleset maintainers.
* #1291, #1292: @J0WI has already mentioned that some features are missing from Chrome/WebExtension version.
* #17868: I don't see any progress in the discussion of UI/UX feedback.
* #12232: While it says that this extension using 40-80 MB of memory, mine is using over 120 MB.

I suppose the issues above can be resolved by making "beta_2.0" (or something like that) branch and apply backward-incompatible changes there. 

Also, maybe it's better to make some things from scratch than rewriting that.

pipboy96:
Currently I see 90MB memory usage.

pipboy96:
Generally, I like this idea. We should discuss it with actual EFF employees though.

Giltyhub:
One can mention #17143 as well

cschanaj:
> #12232: While it says that this extension using 40-80 MB of memory, mine is using over 120 MB.

I might have missed something but I don't see any claim on the memory usage in #12232 matched your description. I am guessing that you are referring to #16092. This measurement calculate the theoretical memory consumption by the rulesets, not the extension itself. 

I am open to the idea of refreshing the extension since the current code have some assumption which limit the performance. I believe this will also imply a breaking change to the ruleset format. If https://github.com/EFForg/https-everywhere/issues/17143#issuecomment-447156379 can be resolved, we should rewrite the whole extension in WASM, such that the improvement can be very obvious. However, I am concerned that #17268 will limit our effort to fully rewrite the extension given the uncertainty on the Google's side.

pipboy96:
[...]

PR #18590, opened 2019-10-19
Make HTTPS Everywhere always ON by default

cschanaj:
This PR will remove the On/Off switch from the extension UI and effectively make it an always-on extension. From many issues, e.g. #18055 it seems that the On/Off switch has been a cause of confusion for normal users to understand how HTTPS Everywhere operate. 

This change is also aligned with UX suggestion from https://github.com/uracreative/works/issues/29#issuecomment-530731106


pipboy96:
We would need to fix the exclusions before doing that.

cschanaj:
@pipboy96 would you mind to explain a little bit on the exclusion issue? We already have a feature to disable the extension per site since #16546. I might have missed something tho.

pipboy96:
@cschanaj I mean allowing to whitelist websites without visiting them beforehand.

zoracon:
Sorry for the delay on responding to this. Right now, taking away the ability to turn the extension "on/off" would likely not be ideal for many users out there who want to temporarily turn the extension off without going to the preferences panel.

We have not yet moved to implement Ura Design Labs suggestions, and the expectations that it will be implemented are still being negotiated with the feedback they will implement.

I do agree and it's backed by research and comments that people are very much confused by the 3 states this extension can have, But taking the on/off switch is a big design decision and can trigger more confusion & frustration with users if we do this without warning. Which is why we can't accept this change at the moment.

On addressing the confusion users have with the states, I plan on utilizing this guide for onboarding new and existing users: https://extensionworkshop.com/documentation/develop/onboard-upboard-offboard-users/

Right now Privacy Badger has a great onboarding page and they have had success with it. So I plan to create a page similar for HTTPS Everywhere after I address Manifest V3 testing.

pipboy96:
@zoracon Thanks for clarifying. Browsers can already temporarily disable extensions, so we don't really "lose" any functionality. Primary issue is to avoid confusing our users.

cschanaj:
> ...taking away the ability to turn the extension "on/off" would likely not be ideal for many users out there who want to temporarily turn the extension off without going to the preferences panel.

In addition to the extension toggle embeded in the browser, users can disable HTTPSE on a session basis since the latest release (#18417). I am more inclined to think that the users rarely needed to disable the entire extension to avoid site breakages, such scenarios are more frequent but limited to advanced users/ developers only. 

An on-boarding will help to address the confusions on how HTTPSE work, it is definitely a good idea :+1:

zoracon:
Just to update, it is in motion with internal design to not only create an installation/onboarding page, but a style guide as well

cschanaj:
Update: #18993 is merged and released in version `2020.3.16`

cschanaj:
@zoracon https://github.com/EFForg/https-everywhere/issues/18069#issuecomment-705849864 mentions that the UX redesign works will be postponed to next year. Would you be interested to review/ close this PR to avoid the merge conflicts?

zoracon:
I will close for now, but still have the idea to bring up the on-boarding page to our design team to address this.

PR #18596, opened 2019-10-20
Fix WASM implementation missed wildcard targets, closes #18589

### LINKED ISSUE #18589: WASM implementation sometimes fails to find rules
Type: code issue

With EASE enabled, http://meta.unix.stackexchange.com/ leads to the interstitial page while we would expect it to lead to https://unix.meta.stackexchange.com/ because of:

https://github.com/EFForg/https-everywhere/blob/b11aa252a38932ce09bb50b87adcc1f14382c59c/src/chrome/content/rules/Stack-Exchange.xml#L137-L139

I think this used to work. Am I missing something obvious here?

cschanaj:
It is found that the WASM implementation of HTTPS Everywhere does not support wildcard targets. This can be reproduce by setting `javascript.option.wasm` to `false` on the `about:config` page on Firefox with #18589 .

This PR was initially intended for adding wildcard support for disable host so it includes some refactorization. Besides, I am unable to add back the unit test for `nullIterable` because #18563 is blocking.

Note that this affect browsing outside of EASE mode since `2019.6.27` (current stable).

P.S. it seems that this straight forward fix will increase the memory usage by quite a bit.... 

### List related PRs if any

- #18093, #18563 

- #18589

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -562,42 +562,40 @@ RuleSets.prototype = {
     }
 
     let results;
-    let expressions = util.getWildcardExpressions(host);
 
     if (this.wasm_rs) {
-      for (const expression of expressions) {
-        let pa = this.wasm_rs.potentially_applicable(expression);
-        results = new Set([...pa].map(ruleset => {
-          let rs = new RuleSet(ruleset.name, ruleset.default_state, getScope(ruleset.scope), ruleset.note);
+      let pa = this.wasm_rs.potentially_applicable(host);
+      results = new Set([...pa].map(ruleset => {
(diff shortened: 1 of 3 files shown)


pipboy96:
@cschanaj Current code doesn't redirect anything for some reason from what I see in Travis.

cschanaj:
@pipboy96 On further investigations, it seems that the following part of code is the root cause of the issue. This issue should be fixed there ideally, so I am partially reverting the changes here 

https://github.com/EFForg/https-everywhere-lib-core/blob/master/src/rulesets.rs#L386-L393

cschanaj:
I am closing this. Since this PR is a refactor work rather than fixing the captioned issue. Superseded by https://github.com/EFForg/https-everywhere-lib-core/pull/1

PR #18603, opened 2019-10-23
Create topekazoo.org.xml

TotalCaesar659:


src/chrome/content/rules/topekazoo.org.xml
```diff
@@ -0,0 +1,19 @@
+<!--
+	Invalid certificate:
+		email.topekazoo.org
+		imap.topekazoo.org
+		mail.topekazoo.org
+		pop.topekazoo.org
+		smtp.topekazoo.org
+		store.topekazoo.org
```
J0WI: WFM

src/chrome/content/rules/topekazoo.org.xml
```diff
@@ -0,0 +1,19 @@
...
+		webmail.topekazoo.org
+
+	Refused:
+		mx.topekazoo.org
```
J0WI: Same for plain HTTP, so this can be removed from the comment section.

TotalCaesar659: Do you mean that I don't need to add all these invaild domains to comment? Cause many other reviewers asked me to add some subdomains or check all ones. I'm not good in subdomain classification, and even contributing.md doesn't help me muchю

J0WI: If they don't run any HTTP service like `mx.topekazoo.org` you can skip them. All others with certificate issues etc. should be documented.

src/chrome/content/rules/topekazoo.org.xml
TotalCaesar659: Invalid certificate for me, but OK, I'll move it.

PR #18615, opened 2019-10-24
Create rainews.it.xml

TotalCaesar659:


src/chrome/content/rules/rainews.it.xml
```diff
@@ -0,0 +1,55 @@
...
+		televisionary.blog.rainews.it
+		thalassa.blog.rainews.it
+		viadellaconciliazione.blog.rainews.it
+		files24.rainews.it
+		www.files24.rainews.it
+		ls.rainews.it
+		medicalc.glazdik.ru
+-->
+<ruleset name="rainews.it">
+	<target host="rainews.it" />
```
J0WI: Mismatching cert, but can be redirected to `www`.

Issue #18618, opened 2019-10-24
Release 2019.11.x forthcoming [On Hold]


zoracon:
Type: other

Release Notes (So far):
- #18417 EASE HTTP Once Exception
- #18565 Add Private network IPs to exclusion for HTTPSE
- #18153 Refactor code for the browser session storage
- #18263 Revert icons back to previous state
- #18452 Normlise disabled site hostname and migrate stored list
- #18602 hsts-prune

Alerting: @J0WI , @TotalCaesar659 , @Bisaloo , @pipboy96 , @ivysrono , @cschanaj

cschanaj:
@zorazon the fix for #18589 should be merged before the next release since most of the wildcard rules is not working right now. It affects all browsers which support WASM by default.

ivysrono:
#18495

cschanaj:
#18602

zoracon:
I really want to get https://github.com/EFForg/https-everywhere-lib-core/pull/1 In this release. So holding off until Monday. If it can't happen then, I'll plan for a patch release when it occurs after this one.

zoracon:
Update: Was in the process of release but came across some strange issues with the translations files and having a few process problems. Will be retrying first thing in the morning. Thank you for your patience.

pipboy96:
@zoracon Should this be closed?

zoracon:
@pipboy96 , yes for some reason the merge didn't close this out

PR #18619, opened 2019-10-24
Allow users to whitelist hosts from the option page

### LINKED ISSUE #17434: Add ability to disable HTTPS Everywhere for specific domains without first visiting the site

### LINKED ISSUE #18503: Allow end user to add *.neverssl.com into the disabled sites list
Type: query

I find that ``neverssl.com`` is unusable on EASE mode, even when its domain is added to my "HTTPS Everywhere Sites Disabled" list, as it currently redirects you to a url like ``lxzmwvtdsnrhkfbc.neverssl.com`` (where part before the dot is random data) to fix problems with stale in-browser cache

Add-ons_Manager

So I like a way to add ``*.neverssl.com`` into the disabled sites list so that ``neverssl.com`` would work for me on EASE mode

I have written a custom ruleset for ``neverssl.com`` and added it into firefox via debugging rulesets but it still doesn't work with EASS

Debugging_Rulesets

It would be the good if ``*.neverssl.com`` can be excluded in https-everywhere. If not, please allow the user to add it themselves via a ruleset, or adding disabled sites with wildcard in the extension's preference UI, or both, so that ``neverssl.com`` continues to work for users that choose to use EASS mode

cschanaj:
This PR will:

- Add UI to add disabled site without require a first visit (close #17434)
- Add wildcard support for disabled site (close #18503)
- Allow us to get rid of the double-negation toggle (follow-up ~#18565~ #18590)

Note that:

- ~this PR have temporarily removed the unit test for the `nullIterable` object because some refactoring works are blocked by #18563. I will submit a PR to restore the unit test once that is merged.~ See 8a6bace

- wildcard matching in this PR is not really working with WASM enabled because of #18589. So the review of this PR will be much easier with the fix in https://github.com/EFForg/https-everywhere-lib-core/pull/1 merged. 

P.S. Personally, I hope this can be merged before a release is made (#18618) if it is not too urgent.

**EDIT:** Updated the incorrect PR reference.


chromium/background-scripts/util.js
```diff
@@ -110,8 +110,8 @@ function getWildcardExpressions(host) {
 
   // now eat away from the left, with *, so that for x.y.z.google.com we
   // check *.y.z.google.com, *.z.google.com and *.google.com
-  for (let i = 1; i <= segmented.length - 2; i++) {
-    const tmp = "*." + segmented.slice(i, segmented.length).join(".");
+  for (let i = 1; i < segmented.length - 1; i++) {
+    const tmp = ["*", ...segmented.slice(i, segmented.length)].join(".");
```
pipboy96: Good idea, makes more sense IMO.

cschanaj:
@zoracon would you take a look at this PR as well?

zoracon:
Reviewing today

pipboy96:
I think it would also make sense to allow the user to remove temporary exclusions created by "Open insecure page for this session only" option on EASE error page.

chromium/pages/options/ux.js
```diff
@@ -331,6 +327,18 @@ document.addEventListener("DOMContentLoaded", () => {
     }
   });
 
+  const add_disabled_site = document.getElementById("add-disabled-site");
+  const disabled_site_input = document.getElementById("disabled-site");
+  disabled_site_input.setAttribute("placeholder", chrome.i18n.getMessage("options_enterDisabledSite"));
+
+  add_disabled_site.addEventListener("click", function() {
+    if (disabled_site_input.value !== "") {
```
zoracon: Could you also add some validation here as well?

cschanaj:
> wildcard matching in this PR is not really working with WASM enabled because of #18589. So the review of this PR will be much easier with the fix in EFForg/https-everywhere-lib-core#1 merged.

I am so sorry that I was wrong on this statement. It seems there is uncommited changes on my local branch and the wildcard matching is not working because the refactoring work was partially done.

cschanaj:
@zoracon @pipboy96 this is ready for a second round of reviews. thanks.

cschanaj:
ping @zoracon for a second review. thanks.

zoracon:
Pulling this down and testing today. Apologies for the delay

pipboy96:
@zoracon I think we should also allow adding whitelisted domains/disabled rulesets and maybe force/disable EASE from managed storage (enterprise policy).

zoracon:
Pulling this down and testing today. Apologies for the delay.

PR #18637, opened 2019-10-25
[Labeller Dockerfile] pin node version

J0WI:
Labeller is broken with current versions.


pipboy96:
I was planning to rewrite the labeller at some point.

J0WI:
That would be great. Node 8 and some of the packages are already EOL.

pipboy96:
@J0WI I think GitHub Actions may be useful for this scenario.

PR #18655, opened 2019-11-01
Create ruleset: Flyerify.com

injust:


J0WI:
This may break 3rd lvl subdomains

injust:
@J0WI My bad, forgot about how to handle wildcard certs.
Fixed now, thanks!

PR #18658, opened 2019-11-01
Create ruleset: AntiqueAppliances.com

injust:


src/chrome/content/rules/AntiqueAppliances.com.xml
```diff
@@ -0,0 +1,7 @@
+<ruleset name="AntiqueAppliances.com">
```
zoracon: Could you add a comment above this ruleset addressing any non functional hosts, for example, `autodiscover.antiqueappliances.com`?

PR #18667, opened 2019-11-07
Ensure that the translations script does not attempt to convert nonexistant translation files

Hainish:


pipboy96:
@Hainish Can we somehow get rid of this conversion step?

zoracon:
> @Hainish Can we somehow get rid of this conversion step? Am I correct that it's impossible without changes on Tor Project's side?

@pipboy96 Since are translations are directly ran by Tor Project's translations efforts, correct. We can not modify the process in a way that separates for now.

pipboy96:
@zoracon Can't we just move them into GitHub repo and handle them the same way as code? (I just ask, I do not propose this change immediately)

zoracon:
> @zoracon Can't we just move them into GitHub repo and handle them the same way as code? (I just ask, I do not propose to change anything now)

Statically, yes. But dynamically as strings change those translations efforts are not handled by solely by us. It takes bigger effort and capacity we do not have internally.

PR #18685, opened 2019-11-15
Add ruleset: Tim.it

scop:
### List related PRs if any

- #15312

### DIFF
--- a/src/chrome/content/rules/Tim.it.xml
+++ b/src/chrome/content/rules/Tim.it.xml
@@ -1,4 +1,4 @@
-<ruleset name="Tim.it">
+<ruleset name="Tim.it (partial)">
 	<target host="tim.it" />
 	<target host="www.tim.it" />
 
--- a/src/chrome/content/rules/Tim.it.xml
+++ b/src/chrome/content/rules/Tim.it.xml
@@ -1,16 +1,8 @@
-<!--
-	Nonfunctional hosts in *.tim.it:
-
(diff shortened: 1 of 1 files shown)


zoracon:
The following errors happened with these hosts:
```
ERROR src/chrome/content/rules/Tim.it.xml: Fetch error: http://tim.it/ => https://tim.it/: (35, 'OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to tim.it:443 ')

ERROR src/chrome/content/rules/Tim.it.xml: Fetch error: http://www.tim.it/ => https://www.tim.it/: (35, 'OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to www.tim.it:443 ')
```

Please amend ruleset

src/chrome/content/rules/Tim.it.xml
```diff
@@ -0,0 +1,21 @@
+<ruleset name="Tim.it (partial)">
+	<target host="tim.it" />
```
zoracon: It seems like this base host has an error connecting to HTTPS. Could you double check to see if HTTPS is available?

scop: Seems flaky at the moment:
```
$ curl -I https://tim.it
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to tim.it:443 
$ curl -I https://tim.it
[...]
```

PR #18690, opened 2019-11-16
Add pypi.org

scop:


pipboy96:
This website is already redirected to HTTPS in all supported browsers due to HSTS preloading.

PR #18722, opened 2019-12-01
Update ComiXology.com.xml

TotalCaesar659:


ivysrono:
https://travis-ci.org/github/EFForg/https-everywhere/jobs/690285578#L328
ERROR src/chrome/content/rules/Amazon.xml: Dangling reference to ComiXology.com.xml

PR #18732, opened 2019-12-04
Create Lionbrand.com

zoracon:

### DIFF
--- a/src/chrome/content/rules/Lionbrand.com.xml
+++ b/src/chrome/content/rules/Lionbrand.com.xml
@@ -1,7 +1,7 @@
 <!--
-	Refused:
-      Seems like most functioning subdomains do not have https
-      *.lionbrand.com
+Refused:
+  Seems like most functioning subdomains do not have https
+  *.lionbrand.com
 -->
 <ruleset name="Lionbrand.com">
 	<target host="lionbrand.com" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Lionbrand.com.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+	Refused:
+      Seems like most functioning subdomains do not have https
```
pipboy96: `s/      /\t\t/`

src/chrome/content/rules/Lionbrand.com.xml
```diff
@@ -0,0 +1,11 @@
+<!--
+Refused:
```
pipboy96: Maybe this line should also have some indentation?

zoracon: Sure

pipboy96:
You currently have both spaces and tabs in your code. Maybe only one of these should be used?

zoracon:
> You currently have both spaces and tabs in your code. Maybe only one of these should be used?
My code editor is set up with tabs at 2 spaces, Ive only used tabs. @pipboy96,

pipboy96:
@zoracon

pipboy96:
@zoracon LGTM!

zoracon:
> @zoracon LGTM!

Turns out my tabs in my code editor was inserting spaces :/. IDEs.

pipboy96:
@zoracon Try setting up your IDE to use EditorConfig: https://github.com/EFForg/https-everywhere/blob/master/.editorconfig

PR #18733, opened 2019-12-07
Create 1freehosting.com.xml

ivysrono:


src/chrome/content/rules/1freehosting.com.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+  Mismatched:
+    cpanel.1freehosting.com
```
pipboy96: Please use tabs to indent.

PR #18736, opened 2019-12-12
Remove unused imports

camjc:


camjc:
Is there a linter to automatically stop this type of thing merging we should apply?

zoracon:
> Is there a linter to automatically stop this type of thing merging we should apply?
There is no requirements for a linter here in this case

PR #18744, opened 2019-12-16
Re-activate and simplify Stanford.edu.xml

cschanaj:
Note that this is an automatically generated ruleset. So I would like to keep it simple and avoid complex rewrites.


pipboy96:
@cschanaj What did you use to generate it?

cschanaj:
I use my own bot https://github.com/cschanaj/xhttpse2 for most PR I submit. IMHO, it works quite well for most cases.

PR #18749, opened 2019-12-18
Remove 3D_Vision_Live.com.xml

zoracon:
Reason: Host no longer functional
Related #17912


src/chrome/content/rules/NVidia.xml
```diff
@@ -23,11 +22,16 @@
 		- news.developer ²
 		- ngvpn04.nvidia.com ²
 		- ngvpn12.nvidia.com ²
+		- ngvpn15.nvidia.com ²
+    - ngvpn31.nvidia.com ²
+    - ngvpn32.nvidia.com ²
+    - ngvpn33.nvidia.com ²
```
cschanaj: the indents seem to be off here

PR #18758, opened 2019-12-21
add entry for blog.fefe.de

sjjh:
must have forgotten this commit.


sjjh:
sorry for the noise. I overlooked my commit  four years ago. (By now) the url blog.fefe.de is already present in the file [src/chrome/content/rules/Fefe.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Fefe.xml). Thus this commit and pull request is unnecessary. I tried to delete it and apparently failed. I'll try again and if do not succeed, please feel free to delete it.

PR #18766, opened 2019-12-25
Add boards.4channel.org to ruleset

BaderSZ:
Added a missing ruleset for the website. Quite recently, 4chan adopted the usage of boards.4channel.* in their service.


cschanaj:
thanks!

PR #18767, opened 2019-12-27
5ch.net

Ohbi3urao8:
5ch.net list

### DIFF
--- a/src/chrome/content/rules/5ch.net.xml
+++ b/src/chrome/content/rules/5ch.net.xml
@@ -1,0 +1,33 @@
+<!--
+	Nonfunctional hosts in *5ch.net:
+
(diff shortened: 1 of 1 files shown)


Ohbi3urao8:
why canceled ?

pipboy96:
Do not use `*`, enumerate subdomains using [Amass](https://github.com/OWASP/Amass).

zoracon:
You can also use [Sublist3r](https://github.com/aboul3la/Sublist3r) as well for enumerating a full set of subdomains.

pipboy96:
@zoracon Amass is a bit faster, especially in brute mode.

pipboy96:
Closed in favor of #19126.

PR #18770, opened 2019-12-28
Re-activate and Update TwitCasting.tv.xml

cschanaj:


src/chrome/content/rules/TwitCasting.tv.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Note: *.twitcasting.tv has a wildcard DNS record and certificate.
```
J0WI: Shall we list `movie.twitcasting.tv` as `refused`?

PR #18773, opened 2019-12-29
Update playground.ru.xml

TotalCaesar659:


TotalCaesar659:
There is redirect to www, this fix is unnecessary now.

Issue #18434, opened 2019-09-24
Firefox will disable support for TLS 1.0 and TLS 1.1


cschanaj:
Type: code issue

> In March of 2020, Firefox will disable support for TLS 1.0 and TLS 1.1.

We might have to update Travis to accommodate the change, see https://blog.mozilla.org/security/2018/10/15/removing-old-versions-of-tls

https://github.com/EFForg/https-everywhere/blob/master/test/rules/src/https_everywhere_checker/http_client.py#L62

Google Chrome will deprecate TLS 1.0 and TLS 1.1 in C72, see https://security.googleblog.com/2018/10/modernizing-transport-security.html

**EDIT**: Fixed URL to https-everywhere-checker

pipboy96:
@cschanaj Chrome 77 still has TLS 1.0 and TLS 1.1 enabled. IMO March 2020 is not a date that should cause immediate concern. I feel this should be closed for time being.

cschanaj:
I think this is similar to #17912 in the sense that once the browsers removed the support for TLS 1.0 and TLS 1.1, we will be redirecting our users to malfunction pages. IMHO We should act early by requiring TLS 1.2 or later on Travis to minimize the impact it might have.

pipboy96:
@cschanaj Makes sense, we need @zoracon's opinion though.

pipboy96:
To be honest, any website that doesn't support at least TLS 1.2 may reasonably be expected to have other issues, such as mixed content or lack of support of secure ciphers. I am for this change.

zoracon:
@cschanaj Considering that these browsers will likely put up security warnings for sites on TLS 1.0 and 1.1. I want to check out and investigate the SSL warning they will likely be thrown in the error object (unless someone already knows this). Because we will have to update Travis and some of the JS in `background.js` that accounts for SSL warnings.

>Be aware that these changes will appear in pre-release versions of Firefox (Beta, Developer Edition, and Nightly) earlier than March 2020.  We will announce specific dates when we have more detailed plans.

Gonna dig through these pre-release versions and correspond with a contact to account for that future SSL warning that will likely be thrown

TLDR;
I support this, just don't want to make this change until I confirm that it has been disabled in Beta (unless someone knows that they have or not already in this thread)

pipboy96:
@zoracon I may test right now.

pipboy96:
@zoracon Current macOS Nightly still has TLS v1.0 enabled.

[...]

PR #18781, opened 2020-01-03
Update UST.hk.xml

cschanaj:
related to #18434 #18700


cschanaj:
Note that `validations` passed in 4a3bb1c, closing .

Issue #16468, opened 2018-09-04
Validations test not reliably catching duplicate targets


injust:
Type: code issue

The validations test for [this build](https://travis-ci.org/EFForg/https-everywhere/builds/423522907) passed, even though a target (`<target host="s.wsj.net" />`) was included twice in the same ruleset.
However, the same duplicate-target mistake caused [the build before that](https://travis-ci.org/EFForg/https-everywhere/builds/423522278) to fail the validations test.

cschanaj:
I am able to reproduce this issue when I created #18782 with an outdated branch on my fork. I tried to add coverage to `lbpics.ust.hk` without realizing that it was already done in ee97815b508 . When Travis run the validations, my PR is merged to the `master` gracefully (yet unfortunately) without merge conflicts. This resulted in 2 occurences of `lbpics.ust.hk` in the merged branch, and thus the failing `validations` test. 

While this is not a bug, a clearer debug message on Travis is most welcomed.  

cc @zoracon  @Bisaloo @J0WI @pipboy96

PR #18782, opened 2020-01-03
Update UST.hk.xml

cschanaj:
Supersedes #18781 Related to #18434 #18700 #16468


src/chrome/content/rules/UST.hk.xml
```diff
@@ -12,11 +12,12 @@
 		- ustlib.ust.hk
 
 		SSL connect error:
```
J0WI: Note: they are still working in the current versions, including Tor.

PR #18807, opened 2020-01-15
Update Redmine.org.xml

J0WI:


J0WI:
> Could you also trivialize the cookie rule while you're at it please?

Done

src/chrome/content/rules/Redmine.org.xml
```diff
@@ -17,23 +10,20 @@
 		- www.redmine.org
 
 -->
```
pipboy96: IMO this comment and the next one is unnecessary, since there is no reason to list nonexistent domains and domains which set insecure cookies.

PR #18813, opened 2020-01-20
Fix Non-www requests on costco.com

ryanbr:
We had a report of https://costco.com not responding and can confirm the non-www isn't responding.

Was reported here, https://community.brave.com/t/https-anywhere-stops-costco-com-loading/103072/2

Any questions let me know.

### DIFF
--- a/src/chrome/content/rules/Costco.xml
+++ b/src/chrome/content/rules/Costco.xml
@@ -9,6 +9,7 @@
 
 		- video *
 		- www2 ²
+        - Non-www (redirection needed)
 
 	* Shows another domain
 	² Reset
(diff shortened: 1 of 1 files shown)


pipboy96:
This is how the ruleset should look like:

```
<!--
	Other Costco rulesets:
		- Costco-static.com.xml

	Timeout:
		- ^
		- www2
-->
<ruleset name="Costco.com (partial)">
	<target host="costco.com" />
	<target host="www.costco.com" />
	<target host="customerservice.costco.com" />

	<rule from="^http://costco\.com/" to="https://www.costco.com/" />

	<rule from="^http:"	to="https:" />
</ruleset>
```

J0WI:
They may block our CI and we have to add them to the [whitelist](https://github.com/EFForg/https-everywhere/blob/master/utils/ruleset-whitelist.csv).

J0WI:
But the initial issue seems to be fixed anyway so this PR is superfluous.

PR #18818, opened 2020-01-26
Add gitlab subdomain to FreeDesktop.org ruleset

1ace:
What is says on the tin :)

We use GitLab for pretty much all FreeDesktop.org projects now, so let's add a rule to always load it via https :+1:


J0WI:
_Triggering Travis_

J0WI:
Thanks!

PR #18819, opened 2020-01-26
[Mesa3d.org] add ruleset to always load via https

1ace:


src/chrome/content/rules/Mesa3d.org.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="Mesa3d.org">
+
+	<target host="mesa3d.org"/>
```
pipboy96:
```suggestion
	<target host="mesa3d.org" />
```

src/chrome/content/rules/Mesa3d.org.xml
```diff
@@ -0,0 +1,8 @@
...
+	<target host="www.mesa3d.org"/>
```
pipboy96:
```suggestion
	<target host="www.mesa3d.org" />
```

src/chrome/content/rules/Mesa3d.org.xml
```diff
@@ -0,0 +1,8 @@
...
+
+	<rule from="^http:" to="https:"/>
```
pipboy96:
```suggestion
	<rule from="^http:" to="https:" />
```

PR #18831, opened 2020-02-01
Add ruleset for themeparkreview.com

Jochen-A-Fuerbacher:


src/chrome/content/rules/ThemeParkReview.com.xml
```diff
@@ -0,0 +1,30 @@
+<!--
+	Invalid certificate:
+		- ftp.themeparkreview.com
+		- ns1.themeparkreview.com
+		- ns2.themeparkreview.com
+		- server2.themeparkreview.com
+		- www.web1.themeparkreview.com
+
+	PR_END_OF_FILE_ERROR:
```
pipboy96: What does that mean?

Jochen-A-Fuerbacher: This is the error, Firefox says when it tries to connect to that host. It also says `Secure Connection Failed`.

I also tried it with curl:
```
$ curl https://mail.themeparkreview.com/
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to mail.themeparkreview.com:443
```

Looks like, it is some kind of SSL error.

pipboy96: The correct error string is `Connection closed:`.

PR #18842, opened 2020-02-03
Linux-KVM.org

MilanKral:


src/chrome/content/rules/Linux-KVM.org.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	m: certificate mismatch
+	    - linux-kvm.org
```
pipboy96: Please replace 4 spaces with a tab.

PR #18845, opened 2020-02-04
Fix and update welt.de ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/Welt.de.xml
```diff
@@ -1,45 +1,30 @@
...
-	Incomplete certificate chain:
-		- stepstone.welt.de
+	Mismatch:
+		- apps.welt.de
 
 	Refused:
-		- biowetter.welt.de
+		- epaper.apps.welt.de
 		- donnerunddoria.welt.de
-		- drakensang.welt.de
```
J0WI: `epaper.welt.de`

src/chrome/content/rules/Welt.de.xml
```diff
@@ -1,45 +1,30 @@
...
+		- reisewetter.welt.de
+		- schmid.welt.de
 		- service.welt.de
-		- tv.welt.de
+		- to.welt.de
 
-	Incomplete certificate chain:
-		- stepstone.welt.de
+	Mismatch:
+		- apps.welt.de
```
J0WI: WFM

Jochen-A-Fuerbacher: What is the meaning of `WFM`?

PR #18849, opened 2020-02-04
Fix and update zdf ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/zdf.xml
```diff
@@ -1,79 +1,74 @@
...
 	<target host="quiz.tivi.de" />
 	<target host="typtest.tivi.de" />
-	<target host="weihnachtsbaumspiel.tivi.de" />
+	<target host="vr.tivi.de" />
+		<test url="http://vr.tivi.de/checkpoint/" />
 
 	<target host="heute-show.de" />
 	<target host="www.heute-show.de" />
+	
+	<rule from="^http://(www\.)?heute-show\.de/" to="https://www.zdf.de/comedy/heute-show/" />
```
J0WI: Please add a comment for it

PR #18859, opened 2020-02-06
Update FAZ ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/FAZ.xml
```diff
@@ -1,83 +1,68 @@
...
-		- boersenlexikon.faz.net
-		- boersenspiel.faz.net
-		- faz-community.faz.net
-		- fazschule.net
-		- jobsuche-hochschulanzeiger.fazjob.net
+	Timeout:
+		- literaturkalender.faz.net
 		- kulturkalender.faz.net
-		- lebenswege.faz.net
-		- media[01]
```
J0WI: `media[0-9]` does still exist

PR #18869, opened 2020-02-08
Add ruleset for opentopomap.org

Jochen-A-Fuerbacher:


src/chrome/content/rules/OpenTopoMap.org.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+	This domain contains a wildcard DNS record.
```
J0WI: Expired: `garmin.opentopomap.org`

Issue #18887, opened 2020-02-10
Madison-Logic.xml invalid match group


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

added in 5b17541c682

```xml
        <rule from="^http://www\.madisonlogic\.com/images/(arrow|bg|headBg|linkbutton_background|mainNav)\.gif$"
                to="https://www.madisonlogic.com/images/$2.gif"/>
```

Same as https://github.com/EFForg/https-everywhere/issues/18886

PR #18888, opened 2020-02-10
Madison-Logic.xml: Fix regex replacement group

jayvdb:
Fixes https://github.com/EFForg/https-everywhere/issues/18887

### DIFF
--- a/src/chrome/content/rules/Madison-Logic.xml
+++ b/src/chrome/content/rules/Madison-Logic.xml
@@ -6,64 +6,27 @@
 
 	Problematic domains:
 
-		- (www.)madisonlogic.com
-
-			- Data differ
-			- $ redirects to Login.aspx
-			- css/ & graphics/ 404
-
-
-	Partially covered domains:
-
-		- (www.)madisonlogic.com
-
+		- (www|blog.)madisonlogic.com	SSL certificate problem: unable to get local issuer certificate
 
 	Fully covered domains:
 
 		- madisonlogic.com subdomains:
 
-			- blog
 			- clk
 			- jsc
 			- st
 
-
-	Mixed content:
-
-		- Script on blog from blog *
-
-		- css:
-
-			- On blog from blog *
-			- On blog from fonts.googleapis.com **
-
-		- Images on blog from blog ***
-
-	* Secured by us, just a facebook widget
-	** Secured by us
-	*** Secured by us, doesn't trip MCB anyway
-
 -->
 <ruleset name="Madison Logic (partial)">
 
-	<target host="www.madisonlogic.com" />
-	<target host="blog.madisonlogic.com" />
 	<target host="clk.madisonlogic.com" />
 	<target host="jsc.madisonlogic.com" />
 	<target host="st.madisonlogic.com" />
 
 
-	<!--	Tracking cookies:
-					-->
-	<securecookie host="^\.madisonlogic\.com$" name="^__(?:utm\w|zlcid)$" />
-
-
-	<rule from="^http://www\.madisonlogic\.com/images/(arrow|bg|headBg|linkbutton_background|mainNav)\.gif$"
-		to="https://www.madisonlogic.com/images/$1.gif"/>
-
 	<!--	Included on 3rd-party websites.
 						-->
-	<rule from="^http://(blog|clk|jsc|st)\.madisonlogic\.com/"
-		to="https://$1.madisonlogic.com/" />
+	<rule from="^http:"
+		to="https:" />
 
 </ruleset>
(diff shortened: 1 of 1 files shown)


jayvdb:
> ERROR src/chrome/content/rules/Madison-Logic.xml: Fetch error: http://dinclinx.com/ => https://clk.madisonlogic.com/: (6, 'Could not resolve host: dinclinx.com')
> ERROR rules/Madison-Logic.xml: Fetch error: http://www.madisonlogic.com/ => https://www.madisonlogic.com/: (60, 'SSL certificate problem: unable to get local issuer certificate')
```
I guess those two mean that the only thing to do is delete or mark the ruleset as failing.

File: src/chrome/content/rules/Madison-Logic.xml [hunk 1]
```diff
@@ -49,14 +44,11 @@
 	*** Secured by us, doesn't trip MCB anyway
 
 -->
-<ruleset name="Madison Logic (partial)">
+<ruleset name="Madison Logic (partial)" default_off="failing tests">
```
cschanaj: Some targets including `clk`, `jsc` and `st` work over HTTPS. It would be great if you can keep their rules. Please also try to document the non-functional host in the top comment, as suggested in [CONTRIBUTING.md#non-working-hosts](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts)

src/chrome/content/rules/Madison-Logic.xml
```diff
@@ -66,15 +58,12 @@
...
-		to="https://clk.madisonlogic.com/" />
-
 	<rule from="^http://www\.madisonlogic\.com/images/(arrow|bg|headBg|linkbutton_background|mainNav)\.gif$"
-		to="https://www.madisonlogic.com/images/$2.gif"/>
+		to="https://www.madisonlogic.com/images/$1.gif"/>
 
 	<!--	Included on 3rd-party websites.
 						-->
-	<rule from="^http://(blog|clk|js-agent|jsc|st)\.madisonlogic\.com/"
+	<rule from="^http://(blog|clk|jsc|st)\.madisonlogic\.com/"
```
cschanaj: This can be replaced by a trivial rules `<rule from="^http:" to="https:" />`

src/chrome/content/rules/Madison-Logic.xml
```diff
@@ -66,15 +58,12 @@
 	<securecookie host="^\.madisonlogic\.com$" name="^__(?:utm\w|zlcid)$" />
```
cschanaj: Since there is no working URL known for the working targets, would you please remove the securecookie tags?

src/chrome/content/rules/Madison-Logic.xml
jayvdb: Re-enabled and removed only non-functioning sites.

Is that the preferred style at https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#non-working-hosts - I've looked at an awful lot of the rules and not seen that style.

I can see it appears on ~1000 rules.  If that is the new style, I'll be happy to update my PRs accordingly.

src/chrome/content/rules/Madison-Logic.xml
jayvdb: With the leading `.`, that looks like a domain cookie, it which case there are still working targets if as you asked for clk, jsc and st to be enabled.

cschanaj: > With the leading ., that looks like a domain cookie

This is a common confusion on the implementation of the `securecookie` tag. In fact, HTTPSE is only capable of protecting cookies set for a domain in the list of `target` if a matching `securecookie` tag is found. If you would like to protect the cookies anyway, you can use a trivial securecookie rule `<securecookie host=".+" name=".+" />`

src/chrome/content/rules/Madison-Logic.xml
```diff
@@ -6,57 +6,21 @@
...
-			- $ redirects to Login.aspx
-			- css/ & graphics/ 404
-
-
-	Partially covered domains:
-
-		- (www.)madisonlogic.com
-
+		- (www|blog.)madisonlogic.com	SSL certificate problem: unable to get local issuer certificate
+		  (Hosting by wpengine.com)
```
cschanaj: Since the hosting information might change over time, we don't include this information in the ruleset comment. Would you remove this line ?

jayvdb: This indicates who to bug about getting the ssl certs fixed, and it is more likely that the cert problem will be fixed before they change hosting provider, and the entire entry would then be removed from the Problematic hosts entry.

Anyways, happy to follow maintainer directions.

src/chrome/content/rules/Madison-Logic.xml
jayvdb: Thanks for clearing that up.

cschanaj:
Thanks!

PR #18895, opened 2020-02-10
Add ruleset for psychology-tools.com

psychology-tools:


pipboy96:
@psychology-tools Thanks for contribution! Since you are a website owner, you may use HSTS preloading instead. It's more effective since it doesn't require the user to have HTTPS Everywhere installed. The user just needs to use a modern version of Chrome or Firefox.

psychology-tools:
Unfortunately for legacy reasons, we have to permit some URLs to be accessible over http for a small number of api users.

I’m trying to get that number down to zero, and redirect to https everywhere it won’t break their service. So once I get them all to upgrade, I will most certainly do this.

But until that happens, I just wanted to make sure at least no one was using http over tor. If/when I am able to do HSTS, I plan open a PR to amend it (or removing the ruleset since it will be moot).

pipboy96:
@psychology-tools Ruleset will be automatically removed when HSTS rule will be added and applied by Chrome, Firefox and Tor Browser.

psychology-tools:
> @psychology-tools Ruleset will be automatically removed when HSTS rule will be added and applied by Chrome, Firefox and Tor Browser.

Even better! 

If it were just up to me, it would already be implemented... Hopefully it won’t have to be in there long.

pipboy96:
@psychology-tools I see some more subdomains that were not included for some reason:

```
amp.psychology-tools.com
beta.psychology-tools.com
lamp.psychology-tools.com
lamp-sf02.psychology-tools.com
tumblr.psychology-tools.com
```

Should they also be included?

pipboy96:
@psychology-tools Also, for some reason http://psychk3hrpclistd.onion/ does not work. Is this expected?

psychology-tools:
> [psychk3hrpclistd.onion/](http://psychk3hrpclistd.onion/)

This is not expected. Probably a change in the private network between instances. Checking on that now.

pipboy96:
@psychology-tools This is OK to merge, unfortunately I can't merge this by myself. We need to wait for another reviewer.

psychology-tools:
> > [psychk3hrpclistd.onion/](http://psychk3hrpclistd.onion/)
> 
> This is not expected. Probably a change in the private network between instances. Checking on that now.

IDK... I just upgraded it to a [v3 onion url ](http://q4yv4muuviiwsag4m6bypay42c7ay3sgir2j4xbeu4a76qnxbq3sgwqd.onion) and now it seems to be working, which I suppose is neither here nor there for https purposes.

Thanks for the guidance (and incidental bug fix) @pipboy96, appreciate it.

pipboy96:
@psychology-tools No problem, I'm always glad to help!

PR #18901, opened 2020-02-11
Update tz.de ruleset

Jochen-A-Fuerbacher:


pipboy96:
See travis:

    failure: Hostname on.tz.de shows up in 2 different rulesets.

PR #18909, opened 2020-02-11
Add ruleset for codeblocks.org

Jochen-A-Fuerbacher:


src/chrome/content/rules/Codeblocks.org.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+	This domain contains a wildcard DNS record.
+
+	Invalid certificate:
+		www.codeblocks.org
+
+	Mismatch:
+		forums.codeblocks.org
+		wiki.codeblocks.org
```
cschanaj: these 2 work for me, would you please check again ?

Jochen-A-Fuerbacher: Both subdomains show on https an other content than on http. On https both subdomains show the content of the frontpage as shown on https://codeblocks.org.

cschanaj: I see, can you mark them as `Different content`?

cschanaj:
Thanks!

Issue #18918, opened 2020-02-13
AmazonAWS.com.xml failing test overlaps with Law_Business_Research_CDN.xml


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

'AmazonWebServices' (AmazonAWS.com.xml) test http://s3.lbrcdn.net.s3-external-3.amazonaws.com/ fails, I believe because it matches a rule now in ruleset `Law_Business_Research_CDN.xml`.

PR #18919, opened 2020-02-13
AmazonAWS.com.xml: Remove failing test

jayvdb:
Hostname is now in Law_Business_Research_CDN.xml

Fixes https://github.com/EFForg/https-everywhere/issues/18918


jayvdb:
I assume is it safe to also remove the exclusion from this rule.

The other approach would be to remove the target from Law_Business_Research_CDN.xml .  For parsing that is a better approach, to avoid domains being split across multiple rulesets, but there are lots of other domains split across multiple rulesets so I guess that goal isnt achievable.

cschanaj:
> The other approach would be to remove the target from Law_Business_Research_CDN.xml 

As you might note that `Law_Business_Research_CDN.xml` is not updated since 2014. There is some recent efforts to avoid merging multiple domain targets in the same ruleset. So I would recommend you to do it the other way.  

Besides, the rewrites in `Law_Business_Research_CDN.xml` will fail in September 2020 (#17912). It would be great if you could remove it in the PR as well.

src/chrome/content/rules/AmazonAWS.com.xml
```diff
@@ -1,6 +1,8 @@
 <!--
 	For other Amazon coverage, see Amazon.xml.
 
+        For s3.lbrcdn.net.s3-external-3.amazonaws.com, see Law_Business_Research_CDN.xml
```
cschanaj: would you please remove this comment as well?

cschanaj:
Thanks!

PR #18930, opened 2020-02-15
Rename and update xfce.org ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/XFCE.org.xml
```diff
@@ -1,52 +1,53 @@
...
-			 squeeze.xfce.org
-			 svn.xfce.org
-			 thunar.xfce.org
-			 translations.xfce.org
-			 www.tx-us.xfce.org
-			 www.us.xfce.org
-			 users.xfce.org
-			 www-test.xfce.org
-			 xfc.xfce.org
+	This domain contains a wildcard DNS record.
```
J0WI: Cannot reproduce

src/chrome/content/rules/XFCE.org.xml
```diff
@@ -1,52 +1,53 @@
...
+		ns[1-4].xfce.org
+		pyxfce.xfce.org
+		squeeze.xfce.org
+		svn.xfce.org
+		thunar.xfce.org
+		translations.xfce.org
+		users.xfce.org
+		www-test.xfce.org
+		xarchiver.xfce.org
+		xfc.xfce.org
```
J0WI:
```
www.tx-us.xfce.org
www.us.xfce.org
```

src/chrome/content/rules/XFCE.org.xml
Jochen-A-Fuerbacher: Try something like https://this-is-a-wildcard-dns-record.xfce.org/ - it works (with a invalid certificate, anyway).

Issue #18944, opened 2020-02-18
Twenga.com non-operational


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Twenga.com.xml
https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Twenga.com-falsemixed.xml

https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Twenga-Solutions.com.xml

https://twenga.com uses cert for https://twenga-solutions.com , and possibly that could become a redirect rule.

https://blog.twenga.com times out, and https://blog.twenga-solutions.com/ is 503

https://dynstatic.twenga.com and https://r.twenga.com and http://i00.twenga.com/cms/img_75577.jpg (etc) all use the wrong cert and http://i00.twenga-solutions.com/cms/img_75577.jpg is 503

Issue #18943, opened 2020-02-18
raptrcdn.com


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/raptrcdn.com.xml

raptrcdn.com is in DNS

https://d1ut3nrduxpo26.cloudfront.net/ isnt in DNS.

A few of the old subdomains like, b[0-3], b7. all point to d1ut3nrduxpo26.cloudfront.net

community.raptrcdn.com is also in DNS, but points to non-existent d3klvv18lzknz8.cloudfront.net

admin.raptrcdn.com isnt in DNS.

Looks like it is dead for a while, and the ruleset should be deleted.

Issue #18942, opened 2020-02-18
polyvore.com and polyvoreimg.com.xml: All targets non-existent


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/polyvoreimg.com.xml

Non-existent hosts:
* https://secure.polyvoreimg.com
* https://ak1.polyvoreimg.com
* https://ak2.polyvoreimg.com
* https://cfc.polyvoreimg.com

Rule pattern also captures non-existent https://akwww.polyvoreimg.com, but that isnt a target.  This is why it turned up on my radar.

https://polyvoreimg.com and https://www.polyvoreimg.com have a cert which only covers `*.polyvore.com, polyvore.com` , but they are not currently targets.

My guess, without understanding the websites in question, is all existing targets in this ruleset should be removed, and polyvoreimg.com and www.polyvoreimg.com added as targets redirecting to `https://polyvore.com`

jayvdb:
Interesting .. http://polyvoreimg.com/cgi/img-thing/size/m/tid/170548310.jpg redirects to https://www.ssense.com/en-id?utm_source=polyvore.com&utm_medium=redirect

All tests (except favicon), e.g. http://polyvore.com/cgi/img-thing/size/m/tid/170548310.jpg (http) and https://polyvore.com/cgi/img-thing/size/m/tid/170548310.jpg (https) redirect to https://www.ssense.com/en-us/women?utm_source=polyvore.com&utm_medium=redirect

Issue #18940, opened 2020-02-18
Myftp.utechsoft.com.xml cert problem


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/Myftp.utechsoft.com.xml

https://myftp.utechsoft.com/ has an expired cert since 2017

The page contains a meta redirect and note to go to https://filecatalyst.com/ instead.

Do we update the rule to point to https://filecatalyst.com/ or delete the ruleset?

pipboy96:
I would delete the ruleset. If it was a HTTP redirect I would suggest making a redirect.

jayvdb:
> If it was a HTTP redirect I would suggest making a redirect.

Do you mean that ... if the website was a http redirect instead of a meta redirect, then a replacement rule would be preferred.

I doubt it matters much in this case as few users would be manually visiting an old website, but I guess ppl could have bookmarks they havent updated.  Deleting the ruleset would mean any http bookmark would start being a http hit again, and the user may see the note before the meta redirect, but if they havent updated their bookmark already, the slight delay of the meta redirect also probably wont trigger them enough to update the bookmark.

Anyways, delete it is!

PR #18949, opened 2020-02-18
Broken ruleset deletions

jayvdb:

- #18939
- #18940
- #18942
- #18943
- #18944


src/chrome/content/rules/Peer1.ca.xml
```diff
@@ -1,19 +0,0 @@
...
-<ruleset name="Peer1.ca (partial)">
-
-	<target host="*.peer1.com" />
-
-
-	<securecookie host="^(?:www\.)?partners\.peer1\.ca$" name=".+" />
-
-
-	<rule from="^http://(www\.)?partners\.peer1\.ca/"
-		to="https://$1partners.peer1.ca/" />
```
jayvdb: note https://partners.peer1.com (.com vs .ca) does exist, however it redirects to https://www.cogecopeer1.com/en/partners/ which redirects to https://aptum.com/en/partners/ which is 404

But this rule never worked, so there is little reason to try rescuing it by creating a Peer1.com  and/or cogecopeer1.com ruleset for previous operating names

jayvdb:
(Rebased to ensure it passes the new whitelist check; no changes occurred)

J0WI:
We tend to `default_off` rules with major issues first. If the target host does not exist anymore it's safe to remove them completely. Could you please split this PR? This allows us to review each case individually.

PR #18950, opened 2020-02-18
Exempt localhost from header rewrite

hansottowirtz:
While developing a website that requests `http://localhost:8080`, I noticed that HTTPS Everywhere rewrites the `Access-Control-Allow-Origin` header to `https://localhost:8080`. This is undesirable with `localhost`.


pipboy96:
@hansottowirtz Do you have EASE mode on (is HTTPS Everywhere icon red)?

hansottowirtz:
Unfortunately I mistook the repository, this problem occurs with Privacy Badger. I will open a PR over there.

PR #18952, opened 2020-02-18
Orion_Magazine.org.xml: Add .com target

jayvdb:
Rule was for .com, which wasnt listed as a target.

.com is not functional; redirect to .org which is functional.

Related to https://github.com/EFForg/https-everywhere/issues/12297

### DIFF
--- a/src/chrome/content/rules/Orion_Magazine.org.xml
+++ b/src/chrome/content/rules/Orion_Magazine.org.xml
@@ -4,7 +4,7 @@
 	<target host="www.orionmagazine.org" />
 
 
-	<rule from="^http://(www\.)?orionmagazine\.com/"
-		to="https://$1orionmagazine.com/" />
+	<rule from="^http:"
+		to="https:" />
 
 </ruleset>


J0WI:
`orionmagazine.com` is for sale, let's just remove it.

src/chrome/content/rules/Orion_Magazine.org.xml
```diff
@@ -1,10 +1,19 @@
...
 
 	<target host="orionmagazine.org" />
 	<target host="www.orionmagazine.org" />
+	<target host="orionmagazine.com" />
+	<target host="www.orionmagazine.com" />
 
 
-	<rule from="^http://(www\.)?orionmagazine\.com/"
-		to="https://$1orionmagazine.com/" />
+	<rule from="^http://(?:www\.)?orionmagazine\.(?:org|com)/"
```
J0WI: Please avoid non-capturing groups

jayvdb: I dont understand why you are wanting me to avoid non-capturing groups, but in any case removing orionmagazine.com as requested heavily modified this line.

Is it still problematic.  If so, why?

J0WI: For simple rules [we prefer](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#regex-conventions) something like
```xml
<!-- Certificate not valid for www -->
<rule from="^http://www\.example\.com/" to="https://example.com/" />
```

but in this case `<rule from="^http:" to="https:" />` seems to [work just fine](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#snapping-redirects).

J0WI:
Thanks!

Issue #18866, opened 2020-02-07
Full Fetch Test: Pending for March 2020


zoracon:
Announcing to our ruleset maintainers that I plan to run this soon. So far, I have amended old config files and currently updating `check_rules.py` from the suggestions given from the [last  fetch test.](https://github.com/EFForg/https-everywhere/issues/13829)

## The tasks listed from the last time that are complete:
- [x] Instead of disabling the entire ruleset, remove individual hosts (#1998)
- [x] The report that is generated when a ruleset is disabled (beside the `Disabled by https-everywhere-checker because` text) should include a timestamp

## Before Each Run
- [x] Update the cipher list in https://github.com/EFForg/https-everywhere/blob/c7da7b7eb704a697cc3d44e406c1912e45b954a9/test/rules/src/https_everywhere_checker
/http_client.py#L48-L49 with ciphers from an updated version of Firefox ESR (https://github.com/EFForg/https-everywhere/pull/18964)
- [x] Create an issue (or send an email) announcing that there will be a forthcoming run of the full fetch test
- [x] Update certificate bundle https://github.com/EFForg/https-everywhere/blob/master/utils/create-platform-certs/README.md
- [x] Update User agent in `http.checker.config` #19004

Also, I wanted to be able to run this test and have it be reviewable for the ruleset maintainers when it is pushed. So I open to what can make PR's more reviewable in the future.

I am also going to track this in a project in the repo so that way progress can be tracked with fetch tests in the future and hopefully run this quarterly and not every 2-3 years anymore.

J0WI:
We should review some of the failed top-* rulesets before submitting a new ruleset update. Probalby also for .onion rules, because the results might not be very reliable there.

zoracon:
> We should review some of the failed top-* rulesets before submitting a new ruleset update. Probalby also for .onion rules, because the results might not be very reliable there.

Should I hold off on the ruleset update for this week?

J0WI:
> Should I hold off on the ruleset update for this week? Or are you referencing the Fetch Test update?

The scheduled release is not affected. The fetch test will cause a bunch of disabled rules and we probably do not want to ship a ruleset update to the API right afterwards that disables important rules for all users.

J0WI:
The whitelist needs to be updated after each run. See also #18967.

pipboy96:
@zoracon We also need to update the certificate bundle.

zoracon:
[...]

PR #18964, opened 2020-02-20
Update Cipher List for Pending Fetch Test

zoracon:
Updating cipher suite to TLS 1.2 for March Fetch Test

TLS 1.0., 1.1 complete deprecation in both Chrome and Firefox roll out in March

Related PR: 
#18700

Related Issue:
#18866


J0WI:
The current versions of ESR/Tor are still supporting TLS 1.0, 3DES an CBC ciphers.

pipboy96:
@J0WI We need to test for intersection of supported browsers' cipher lists, so even if only Chrome removed TLS 1.0, 3DES or CBC ciphers, we would have to disable it anyway.

J0WI:
https://bugzilla.mozilla.org/show_bug.cgi?id=1227524
https://bugs.chromium.org/p/chromium/issues/detail?id=691888

pipboy96:
@J0WI Should we replicate this behavior by making a connection without 3DES support and if if fails, making another with 3DES support enabled?

J0WI:
I think we should just keep it in the cipher list until it's disabled by default in browsers.

test/rules/src/https_everywhere_checker/http_client.py
```diff
@@ -46,8 +46,9 @@ def getCAPath(self, platform):
 class FetchOptions(object):
     """HTTP fetcher options like timeouts."""
 
-    # The default list of cipher suites that ships with Firefox 51.0.1
-    _DEFAULT_CIPHERLIST = "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA:AES128-SHA:AES256-SHA:DES-CBC3-SHA"
+    # The default list of cipher suites that ships with Firefox 68.5.0 ESR
+    # TLS 1.2 Recommemdation: https://wiki.mozilla.org/Security/Server_Side_TLS
+    _DEFAULT_CIPHERLIST = "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384"
```
pipboy96: `DHE-RSA-AES128-GCM-SHA256` and `DHE-RSA-AES256-GCM-SHA384` are only supported by Firefox and not Chrome. I think it should not be included here.

Also to avoid confusion, the list at https://wiki.mozilla.org/Security/Server_Side_TLS is intended to be used as a HTTPS server configuration, not for use in a browser.

I suggest keeping only:

```
ECDHE-ECDSA-AES128-GCM-SHA256
ECDHE-RSA-AES128-GCM-SHA256
ECDHE-ECDSA-CHACHA20-POLY1305
ECDHE-RSA-CHACHA20-POLY1305
ECDHE-ECDSA-AES256-GCM-SHA384
ECDHE-RSA-AES256-GCM-SHA384
```

(these are supported by both Firefox and Chrome, and are all considered fully secure at the moment)

zoracon: Cross checking this list with https://clienttest.ssllabs.com and https://wiki.mozilla.org/Security/Cipher_Suites, I also agree to keep these as well, the other listed weak ciphers I will double check. But for now this list seems like the cross supported suites.

zoracon: For the future, I will add a comment linking into the wiki on how to update this cipher list

pipboy96: @zoracon I think the list should only include the ciphers supported by all three supported browsers (Chrome, Firefox and Tor), since only websites that successfully load over HTTPS on all three should be rewritten to avoid issues.

zoracon: I also cross checked between Tor for this list as well. It looks like it checks out for all 3 browsers. Pushing this change soon.

PR #18976, opened 2020-02-23
Create bundesbank.de ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/Bundesbank.de.xml
```diff
@@ -0,0 +1,54 @@
...
+	Timeout:
+		ebics-t1
+		extranet-t
+		newsletter
+		s504pmgwz1
+		s999mgwxx001a
+		s999mgwxx001p
+		s999mgwxx002a
+		s999mgwxx002p
+		smv
```
J0WI: I get a mismatch, but it works on port 8443

Jochen-A-Fuerbacher: I still get timeouts on both ports, 443 and 8443.

Issue #18910, opened 2020-02-11
Common rules for +/- www.


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

There are a lot of rules which I believe can be simplified because their targets cover a simple set of hostnames, but they are using literal host names in the rules unnecessarily IIUC.  e.g. the following 13 all have a single glob target, and a single rule which could be the basic `{'from': '^http:', 'to': 'https:'}` as long as `*.foo.com` target doesnt apply to `foo.com`.

- *.actionkit.com
- *.csmres.co.uk
- *.stcllctrs.com
- *.mdgms.com
- *.service.mirror-image.net
- *.vo.msecnd.net
- *.secure.nonprofitsoapbox.com
- *.ntwk45.com
- *.ovpn.to
- *.omeda.com
- *.staticwars.com
- *.thankyou4caring.org
- *.zerista.com

There are many more cohorts like this, but to find them easily another type of conversion is needed (first).

There are a lot of rules which are only also adding or removing `www.` when upgrading to https.  e.g. `zuerich.ch.xml` and `zarunet.org.xml` .  These +/- `www.` rules can be detected and replaced with a common pattern and replacement, i.e. 
```py
REMOVE_WWW_RULE = ('^http://(?:www\\.)?([^/]+)/', 'https://$1/')
ADD_WWW_RULE = ('^http://([^/]+)/', 'https://www.$1/')
```

If those commonly usable rules are used instead, it is fairly easy to find pairs (or more) of rules which can be reduced to simpler rules, like `Computeruniverse.xml`.

[...]

PR #18980, opened 2020-02-24
Stale too many redirects (batch 1)

jayvdb:
Re-enables 19 rulesets which were disabled due to too-many-redirects which is no longer accurate.

Related to https://github.com/EFForg/https-everywhere/issues/18981

The failing rules in the "fetch" job all pass locally when running the checker manually.


jayvdb:
I removed the rulesets which were failing in https://travis-ci.org/EFForg/https-everywhere/jobs/660136038 ; again they all worked for me locally, but I'll defer them to the next batch.

I can split the commit "Lots of too-many-redirects fixes" into nine separate commits if requested, one for each ruleset.

If it helps I can also move to a new PR the few changes in this PR which are not re-enabling a ruleset, as that is a reasonable division in this problem.

src/chrome/content/rules/Sensiolabs.com.xml
```diff
@@ -1,25 +1,14 @@
...
     <target host="www.sensiolabs.com" />
-    <target host="support.sensiolabs.com" />
-    <target host="www.support.sensiolabs.com" />
+    <target host="blog.sensiolabs.com" />
     <target host="insight.sensiolabs.com" />
     <target host="www.insight.sensiolabs.com" />
     <target host="connect.sensiolabs.com" />
-    <target host="www.connect.sensiolabs.com" />
 
     <securecookie host="^(www\.)?(support\.|insight\.|connect\.)?sensiolabs\.com" name=".+" />
```
J0WI:
```suggestion
    <securecookie host=".+" name=".+" />
```

src/chrome/content/rules/DMLP.org.xml
```diff
@@ -1,36 +1,23 @@
...
+	<test url="http://dmlp.org/favicon.ico" />
+	<test url="http://www.dmlp.org/favicon.ico" />
 
-	<rule from="^http://(?:www\.)?citmedialaw\.org/"
-		to="https://www.citmedialaw.org/" />
-
-	<rule from="^http://(?:www\.)?dmlp\.org/(?!favicon\.ico|files/|misc/|modules/|sites/)"
-		to="https://www.dmlp.org/" />
+	<rule from="^http://(?:www\.)?([^/]+)/"
+		to="https://www.$1/" />
```
J0WI: You can just use multiple simple redirects here.

jayvdb: I used this per https://github.com/EFForg/https-everywhere/issues/18910

If you still want me to change, should I be using four literal or two basic rules here?

jayvdb: ping @J0WI

J0WI: I prefer two basic rules like in the example [here](https://github.com/EFForg/https-everywhere/issues/18910#issuecomment-600088525)

J0WI: ping @jayvdb

Issue #18983, opened 2020-02-26
Return a simple mapping of trivial rulesets for given update channel


Hainish:
Type: feature request

Reference: https://trac.torproject.org/projects/tor/ticket/28005

Exciting news - Tor Project is using our [update channels](https://github.com/EFForg/https-everywhere/blob/master/docs/en_US/ruleset-update-channels.md) in their alpha release to assist with the discovery of next-generation onion services!

To move this forward, Tor Browser developers have asked us to provide a hook in HTTPS Everywhere to return an object which contains a mapping of { host, to } for simple rulesets in a given update channel.  This should work for rulesets in the format specified in the [example ruleset channel](https://people.torproject.org/~acat/misc/rulesets/) Tor has provided us.

Hainish:
I'll begin work on it this week.

pipboy96:
@Hainish Wouldn't it make more sense to separate this from HTTP to HTTPS redirects, and how would it work if we deprecated all `to="http://*"` rules?

Hainish:
@pipboy96 I think in order to support this, making deprecation of all `to="http://*"` URLs an option in the codepath that we can default to `true` rather than ripping out all the code for this is preferable.

pipboy96:
@Hainish You mean Tor Browser's HTTPS Everywhere version is now going to be different from the regular one?

Hainish:
@pipboy96 I think we could add a hook for HTTPSE to flip that option programmatically. Having a different extension in Tor Browser seems to invite a long-term maintenance burden

pipboy96:
@Hainish I get it, but how is that flag going to be activated? WebExts can no longer read user.js? Will we just turn it on based on user agent?

cschanaj:
The JSON in `default.rulesets.1579180748.gz` looks like this when prettified. 
```
{
    "timestamp": 1579180748,
    "reference": "4437a4168878143ce3fdc63bbd355d537b1cea57",
    "rulesets":
    [
        {
            "name": "Nytimes.securedrop.tor.onion",
            "target":
            [
                "nytimes.securedrop.tor.onion"
            ],
            "rule":
            [
                {
                    "from": "^http[s]?://nytimes.securedrop.tor.onion/",
                    "to": "https://nyttips4bmquxfzw.onion/"
                }
            ]
        }
    ]
}
```

So HTTPS Everywhere is going to list of `*.tor.onion` address specifically for the Tor Browser?

Hainish:
[...]

PR #18994, opened 2020-02-28
Adding 'get_simple_rules_ending_with' message and functionality for Tor

Hainish:
Closes https://github.com/EFForg/https-everywhere/issues/18983

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -738,7 +738,7 @@ RuleSets.prototype = {
       for(let [host, rulesets] of this.targets) {
         if (host.endsWith(ending) &&
             rulesets.length == 1 &&
-            rulesets[0].active == true &&
+            rulesets[0].active === true &&
             rulesets[0].exclusions == null
         ) {
           for (let rule of rulesets[0].rules) {
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -723,6 +723,36 @@ RuleSets.prototype = {
   },
 
   /**
+   * Get a list of simple rules (active, with no exclusions) for all hosts that
+   * are in a single ruleset, and end in the specified ending.
+   * @param ending Target ending to search for
+   * @returns A list of { host, from_regex, to, scope_regex }
+   */
+  getSimpleRulesEndingWith: function(ending) {
+    let results;
+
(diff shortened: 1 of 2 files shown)


chromium/background-scripts/rules.js
```diff
@@ -722,6 +722,36 @@ RuleSets.prototype = {
...
+    let results;
+
+    if (this.wasm_rs) {
+      results = this.wasm_rs.get_simple_rules_ending_with(ending);
+    } else {
+      results = [];
+      for(let [host, rulesets] of this.targets) {
+        if (host.endsWith(ending) &&
+            rulesets.length == 1 &&
+            rulesets[0].active == true &&
```
cschanaj:
```suggestion
            rulesets[0].active &&
```

chromium/background-scripts/rules.js
```diff
@@ -722,6 +722,36 @@ RuleSets.prototype = {
...
+            rulesets[0].exclusions == null
+        ) {
+          for (let rule of rulesets[0].rules) {
+            if (rule.from_c.test("http://" + host + "/")) {
+              results.push({ host, from_regex: rule.from_c.toString(), to: rule.to, scope_regex: rulesets[0].scope.toString() });
```
cschanaj: Both `from_regex` and `scope_regex` are String, IMHO it is quite confusing to have the `_regex` part in their names. They are the same as the `from` attribute from a ruleset and `scope` attribute from a ruleset channel, is there any reason not reusing the names?

Hainish: Yeah, so `sendResponse` [must](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage) return a JSON-ifiable object, which is why we convert the RegExp to a string.

Given that this message is being used by an external project, I thought it safer to specify in the key that the `from` and `scope` should be used as a regular expression, thus the `_regex` suffix.

chromium/background-scripts/rules.js
Hainish: `active` should either be true or false in rulesets, and I'd like to maintain parity with the [`lib-core` implementation](https://github.com/EFForg/https-everywhere-lib-core/blob/master/src/rulesets.rs#L387) here, but this should probably be `===` instead.

PR #18995, opened 2020-02-29
update Ecosia.xml subdomains, short URL

csavard-sudo:
Added wildcard to include subdomains and added "ecosia.co" shortlink as it is used periodically on their site.


csavard-sudo:
> 
> 
> Please add more test urls to ensure that the wildcard does not break anything. https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#left-wildcards

@J0WI done - I included ample test urls in the latest commit, particularly from important subdomains such as the cdn and blog.

J0WI:
Thanks!

PR #19025, opened 2020-03-16
Use native bytes value as the default sort order

cschanaj:
As per the documentation `sort(1)`

```
***  WARNING  ***  The locale specified by the environment affects sort
       order.  Set LC_ALL=C to get the traditional sort order that uses native
       byte values.
```

### DIFF
--- a/utils/ruleset-whitelist-cleanup.sh
+++ b/utils/ruleset-whitelist-cleanup.sh
@@ -38,5 +38,5 @@ done) < "$WLIST"
 
 # Sorting by the 4th column (ruleset name)
 TMPFILE=`mktemp`
-(head -n1 "$WLIST" && tail -n +2 "$WLIST" | sort -t"," -b -u -k4) > "$TMPFILE"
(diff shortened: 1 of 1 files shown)


zoracon:
Could you specify impact it has on the sort order. I get the localization aspect but was this causing some sort of error if the native bytes weren't in use?

Not questioning to be contrary, just genuinely trying to understand what the underlying issue was.

pipboy96:
@zoracon The main issue is that sorting may be inconsistent between machines that have different locales.

cschanaj:
@zoracon It is like what @pipboy96 described, when different contributors use a different locale, the diff of the `ruleset-whitelist.csv` might become difficult to review. This PR will make the automated change much more reproducible.

zoracon:
Thank you both for explaining a bit further, I'm not as sharp as possible of seeing multiple cases for contributors around the world. I also like a running record of commentary for changes. So I appreciate it and will merge this in. Thank you!

PR #19049, opened 2020-03-22
Update EMS ruleset

Jochen-A-Fuerbacher:


Jochen-A-Fuerbacher:
The host `news` cannot be resolved. I added the stats host to the comment.

PR #19061, opened 2020-03-23
Create cleverpush.com ruleset

Jochen-A-Fuerbacher:


src/chrome/content/rules/CleverPush.com.xml
```diff
@@ -0,0 +1,19 @@
...
+<ruleset name="CleverPush.com">
+
+	<target host="cleverpush.com" />
+	<target host="*.cleverpush.com" />
+		<test url="http://www.cleverpush.com" />
+		<test url="http://abendblatt.cleverpush.com" />
+		<test url="http://api.cleverpush.com" />
+		<test url="http://demo.cleverpush.com" />
+		<test url="http://developers.cleverpush.com" />
+		<test url="http://static.cleverpush.com" />
```
J0WI: Missing trailing `/`. Please also add a hint why the wildcard is required here.

J0WI: Thanks!

PR #19099, opened 2020-03-28
Create yourcmc.ru.xml

Deimos715:
http://yourcmc.ru/wiki/Grive2


src/chrome/content/rules/yourcmc.ru.xml
```diff
@@ -0,0 +1,9 @@
+<!--
+	Invalid certificate:
+		www.yourcmc.ru
```
J0WI: Please also check:
```
gtwp.yourcmc.ru
svn.yourcmc.ru
www.svn.yourcmc.ru
vmx.yourcmc.ru
www.vmx.yourcmc.ru
```

PR #19103, opened 2020-03-31
Update hko.gov.hk.xml

cschanaj:


src/chrome/content/rules/hko.gov.hk.xml
```diff
@@ -1,25 +1,32 @@
...
 		- kids.hko.gov.hk
 		- pda.hko.gov.hk
 		- wap.hko.gov.hk
-		- m.weather.gov.hk (sometimes)
 
 		Incomplete certificate chain error:
 		- m.hko.gov.hk
+
+	Non-functional hosts in *.weather.gov.hk:
+		Incomplete certificate chain error:
```
J0WI: I'm getting inconsistent results

PR #19108, opened 2020-03-31
Delete rules for my (old) domain.

Flameeyes:
None of these are useful as none of the hostnames are currently in use. Also, they should all be on HSTS.


pipboy96:
@Flameeyes Please add your website to https://hstspreload.org/, this will make `hsts-prune` remove your website's ruleset automatically.

Flameeyes:
The flameeyes.eu domains are actually phased out, so adding them to the preload list is kind of pointless.

pipboy96:
@Flameeyes Thanks for clarification.

pipboy96:
@zoracon Can you check why Travis doesn't restart?

Issue #8242, opened 2017-01-24
Don't mess with localhost URLs


strugee:
With "block all unencrypted requests" enabled, HTTPS Everywhere tries to rewrite localhost URLs to use HTTPS, which is unideal. We probably shouldn't do that.

Also relevant: https://www.w3.org/TR/secure-contexts/#localhost

exegete46:
+1 👍 - For all the web developers out there.

Kibouo:
+1
Really annoying that it messes with localhost

Ryuno-Ki:
Affects both, `http://127.0.0.1:<PORT>` as well as `http://localhost:<PORT>`.
Pretty annoying.

Ryuno-Ki:
https://github.com/EFForg/https-everywhere/blob/9a243c6cd71f612a5b539c83f59a469eb5950fae/rewriter/rewriter.js#L69 looks like the line in question.

At this point I would long have changed the regex to use `RegExp` … so it is possible to break it into smaller part, concatenate it and pass it to `new RegExp()`.

strugee:
Please stop commenting with +1s; it generates a lot of spam for subscribers/watchers.

@Ryuno-Ki you should PR!

Ryuno-Ki:
@strugee I have no experience with browser add-ons and would need some guidance for setting up dev environment and testing my patch.

strugee:
@Ryuno-Ki https://github.com/EFForg/https-everywhere#getting-started has some instructions for getting started with HTTPS Everywhere development. I've only dabbled in HTTPS Everywhere development but I'd be happy to try and help, or I'm sure you could ping one of the core developers.

Hainish:
@Ryuno-Ki Let me know if you plan to PR, otherwise I can add a localhost exception.

Hainish:
https://github.com/EFForg/https-everywhere/commit/1dafcc9578ab3c9287dc4f811878269fdd57f9ae resolves this.  It will be deployed in 5.2.12

Mikaela:
I am not sure this is fixed or has been reintroduced as HTTPS Everywhere is now conflicting with IPFS Companion using `*.localhost` addresses and those get https and errors when EASE mode is enabled.

* https://github.com/ipfs-shipyard/ipfs-companion/issues/855

pipboy96:
@Mikaela Please don't bump old issues if possible. Create a new issue instead. Thanks for reporting the issue though, this will be looked into.

PR #19125, opened 2020-04-07
Exclude localhost subdomains from EASE redirection

pipboy96:
Related: https://github.com/EFForg/https-everywhere/issues/8242#issuecomment-610247779.


cschanaj:
@pipboy96 would you add comments on `*.localhost` will always resolve the the respective IP loopback address (as per [RFC6761](https://tools.ietf.org/html/rfc6761#section-6.3)) in the comment?

P.S. I am not really sure about the specific standard which mandate this behavior, please feel free to correct me if I made a mistake.

PR #19126, opened 2020-04-07
Create 5ch.xml

maboroshin:
### List related ruleset

Currently `2ch.net` is being redirected to `5ch.net`.

- [2ch.xml](https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/2ch.net.xml)

P.S. ( September 8, 2020)
- 1st test group: Non-forum server: relatively easy for servers to be rejected.
- 2nd test group: Forum server

Checked on September 8, 2020 at [here](https://ss1.xrea.com/z4s85ttt.s1007.xrea.com/5chkanshi/disp_status.php)

### DIFF
--- a/src/chrome/content/rules/5ch.xml
+++ b/src/chrome/content/rules/5ch.xml
@@ -1,44 +1,44 @@
 <ruleset name="5ch.net">
-        <target host="5ch.net" />
-        <target host="*.5ch.net" />
+	<target host="5ch.net" />
+	<target host="*.5ch.net" />
 
-        <test url="http://5ch.net/" />
-        <test url="http://www2.5ch.net/" />
-        <test url="http://itest.5ch.net/" />
-        <test url="http://premium.5ch.net/" />
-        <test url="http://be.5ch.net/" />
-        <test url="http://i.5ch.net/" />
-        <test url="http://headline.5ch.net/" />
-        <test url="http://newsnavi.5ch.net/" />
-        <test url="http://info.5ch.net/" />
-        <test url="http://o.5ch.net/" />
-        <test url="http://stat.5ch.net/" />
-        <test url="http://dig.5ch.net/" />
-        <test url="http://qb5.5ch.net/" />
-        
-        <test url="http://agree.5ch.net/" />
-        <test url="http://asahi.5ch.net/" />
-        <test url="http://egg.5ch.net/" />
-        <test url="http://fate.5ch.net/" />
-        <test url="http://hayabusa9.5ch.net/" />
-        <test url="http://hebi.5ch.net/" />
-        <test url="http://himawari.5ch.net/" />
-        <test url="http://krsw.5ch.net/" />
-        <test url="http://lavender.5ch.net/" />
-        <test url="http://leia.5ch.net/" />
-        <test url="http://mao.5ch.net/" />
-        <test url="http://matsuri.5ch.net/" />
-        <test url="http://medaka.5ch.net/" />
-        <test url="http://mevius.5ch.net/" />
-        <test url="http://nhk2.5ch.net/" />
-        <test url="http://rio2016.5ch.net/" />
-        <test url="http://rosie.5ch.net/" />
-        <test url="http://swallow.5ch.net/" />
-        <test url="http://tanuki.5ch.net/" />
+	<test url="http://5ch.net/" />
+	<test url="http://www2.5ch.net/" />
+	<test url="http://itest.5ch.net/" />
+	<test url="http://premium.5ch.net/" />
+	<test url="http://be.5ch.net/" />
+	<test url="http://i.5ch.net/" />
+	<test url="http://headline.5ch.net/" />
+	<test url="http://newsnavi.5ch.net/" />
+	<test url="http://info.5ch.net/" />
+	<test url="http://o.5ch.net/" />
+	<test url="http://stat.5ch.net/" />
+	<test url="http://dig.5ch.net/" />
+	<test url="http://qb5.5ch.net/" />
+	
+	<test url="http://agree.5ch.net/" />
+	<test url="http://asahi.5ch.net/" />
+	<test url="http://egg.5ch.net/" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
+<ruleset name="5ch.net">
+        <target host="5ch.net" />
+        <target host="*.5ch.net" />
```
pipboy96, line 3: Please do not use wildcard targets if possible.

maboroshin: Is it better this way?
`<target host="\w+.5ch.net" />`

**5ch** (formerly 2ch) is the largest forum in Japan. 5ch has fewer servers. But it's fluid rather than fixed.

The former 2ch (not **5ch**) has grown to over 100 subdomains: [sample](http://qb5.5ch.net/)

pipboy96: The `host` attribute is not a regex, so this won't work. If it's actually impossible to list all subdomains, a wildcard is acceptable.

maboroshin: I think a wildcard is good.

erhaps they are worried about connection errors with HTTPS. But, SSL has been used since the previous 2ch (current 5ch). The web operator is SSL-enabled, can also connect to both via HTTP. It won't cause any problems.

On the other hand, the official smartphone loading program (still output HTTP) has been stalled in development. Programmers are not active. And external third party web tools also.

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
...
+
+        <test url="http://5ch.net/" />
+        <test url="http://www2.5ch.net/" />
+        <test url="http://itest.5ch.net/" />
+        <test url="http://premium.5ch.net/" />
+        <test url="http://be.5ch.net/" />
+        <test url="http://i.5ch.net/" />
+        <test url="http://headline.5ch.net/" />
+        <test url="http://newsnavi.5ch.net" />
```
cschanaj, line 12:
```suggestion
        <test url="http://newsnavi.5ch.net/" />
```

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
...
+        <test url="http://info.5ch.net" />
```
cschanaj, line 13:
```suggestion
        <test url="http://info.5ch.net/" />
```

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
...
+        <test url="http://o.5ch.net/" />
+        <test url="http://stat.5ch.net" />
```
cschanaj, line 15:
```suggestion
        <test url="http://stat.5ch.net/" />
```

cschanaj (changes requested):
Thank you for your PR. Travis did not pass because you missed some trailing slashes for the test URLs. You may apply the suggestions to fix this. Besides, would you please use tabs for indentations as suggested by our [contributing guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#indentation--misc-stylistic-conventions) ?

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
...
+        <test url="http://medaka.5ch.net/" />
+        <test url="http://mevius.5ch.net/" />
+        <test url="http://nhk2.5ch.net/" />
+        <test url="http://rio2016.5ch.net/" />
+        <test url="http://rosie.5ch.net/" />
+        <test url="http://swallow.5ch.net/" />
+        <test url="http://tanuki.5ch.net/" />
+
+        <securecookie host="^\." name="^(?:__cfduid|cf_clearance)$" />
+        <securecookie host="^\w" name=".+" />
```
cschanaj, line 40: You may use the below `securecookie` to protect all of the cookies set by `5ch.net`. Since a wildcard is used, this should be fine and offer better protection. 

```suggestion
        <securecookie host=".+" name=".+" />
```

maboroshin (commented):
@cschanaji Thanks for the assist.
Seems like a good change. Now we just have to wait for the merge?

cschanaj:
Please modify the ruleset by replacing the spaces with tabs for indentations and adding the trailing slashes. If Travis pass, I shall merge this PR.

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,44 @@
+<ruleset name="5ch.net">
+	<target host="5ch.net" />
+	<target host="*.5ch.net" />
+
+	<test url="http://5ch.net/" />
```
cschanaj, line 5: An implicit test URL http://5ch.net/` is added automatically for non-wildcard `target`. Please also remove this line so Travis will not complain about duplicated test URLs.

maboroshin: ＠cschanaj Thank you. I was reading the guide, now.😅

src/chrome/content/rules/5ch.xml
```diff
@@ -0,0 +1,45 @@
+<ruleset name="5ch.net">
+	<target host="*.5ch.net" />
+
+	<test url="http://5ch.net/" />
```
cschanaj, line 4: You need `5ch.net` in the list of `target` in order to enforce HTTPS on `5ch.net`.  However, you don't need a test URL for `5ch.net`. This is because `<target host="5ch.net" />` implicitly created a test URL. Travis will signal an error if you add a duplicated tests. Sorry if I have been causing confusions.

```suggestion
	<target host="5ch.net" />
	<target host="*.5ch.net" />
```

cschanaj (changes requested):
Otherwise, LGTM

src/chrome/content/rules/5ch.xml
maboroshin: @cschanaj I see!

cschanaj (approved):

PR #18938, opened 2020-02-17
General tidy-up to allow matching rules with targets

jayvdb:
**NOTE:** This PR includes the commits in https://github.com/EFForg/https-everywhere/pull/18938 - please lets get that PR merged first.

### List related issues if any

- #12297

I have a fairly functional tool for matching rules with targets in order to detect mismatches.

This PR includes some of the fixes for issues detected, and also regex syntax changes which make the matching easier.

Many of the fixes are in new smaller more focused PRs:

- https://github.com/EFForg/https-everywhere/pull/18949 - Broken ruleset deletions
- https://github.com/EFForg/https-everywhere/pull/18957 - Resolve error fixes

I appreciate that some of these syntax changes in this PR may create other problems that I am as yet unaware of -- I'm happy to drop parts of these changes and adjust my algorithm accordingly.

~~Scania.xml is one example where constraining the `\d+` is a "nice to have", but the change means that new hostnames brought online by Scania wouldnt automatically be included in the rule.  However it does mean that it is feasible for tools to grok the hostname sequence and generate tests dynamically, or detect if another hostname has been brought online, and check whether the regex needs updating.~~ (Scania.xml had bigger problems that this)

PR #19147, opened 2020-04-11
E-rewards.com.xml: Add base domain used in rule

jayvdb:
Related issues:  #17965 #12297 

### List related PRs if any

- From https://github.com/EFForg/https-everywhere/pull/18938

### DIFF
--- a/src/chrome/content/rules/E-rewards.com.xml
+++ b/src/chrome/content/rules/E-rewards.com.xml
@@ -1,4 +1,7 @@
 <ruleset name="E-rewards.com">
-  <target host="www.e-rewards.com"/>
-  <rule from="^http://(?:www\.)?e-rewards\.com/" to="https://www.e-rewards.com/"/>
+	<target host="e-rewards.com"/>
+	<target host="www.e-rewards.com"/>
+
+	<rule from="^http:"
+		to="https:"/>
 </ruleset>


src/chrome/content/rules/E-rewards.com.xml
```diff
@@ -1,4 +1,7 @@
 <ruleset name="E-rewards.com">
-  <target host="www.e-rewards.com"/>
-  <rule from="^http://(?:www\.)?e-rewards\.com/" to="https://www.e-rewards.com/"/>
+	<target host="e-rewards.com"/>
+	<target host="www.e-rewards.com"/>
+
```
cschanaj, line 4: flare.e-rewards.com

jayvdb: Also https://beta.e-rewards.com

And https://pages.e-rewards.com fails with `not enough values to unpack (expected 2, got 1)`

And then there are lots of other hosts with errors of all types, including http://oa.e-rewards.com & http://owa.e-rewards.com which are cert errors as the cert is valid for `*.dynata.com, dynata.com`, but both of those are "Our website is currently unavailable.".

Would this be a reasonable time to use `*.e-rewards.com`

cschanaj: Personally, I would rather whitelisting the ruleset for either coverage / fetch test instead of using a wildcard target to avoid tests. This is because there is a lot of different checking performed by Travis important for the ruleset quality. We also discourage wildcard targets unless the domain have a wildcard DNS record, to reduce the long term maintenance effort. 

If you would like to keep this ruleset as-is (without adding more hosts), I am ok to merge this given there was incomplete coverage. :smile:

jayvdb: Given I have 500 odd files to fix, maintainers merging valid fixes rather than requiring I enhance them would be highly appreciated.

Note that using a wildcard was not to avoid tests.  It should be very obvious from my contributions that I am trying to get more testing to occur.  I would have accompanied a wildcard rule with tests for known working hosts.  However the state of this domain is clearly in flux, and/or deteriorating, so IMO it isnt great use of time to document all of the broken hosts which far outnumber the ones which are working.

jayvdb:
https://travis-ci.org/github/EFForg/https-everywhere/jobs/674664348 needs to be restarted; it should succeed I believe, and I dont seem to have permission to restart Travis jobs.

cschanaj:
Travis will restart automatically if you close and re-open the PR.

PR #19148, opened 2020-04-11
Exhale.xml: Fixes and enhancements

jayvdb:
Related issues:  #17965 #12297 

### List related PRs if any

- From https://github.com/EFForg/https-everywhere/pull/18938


cschanaj:
By convention we avoid non-capturing RegEx rules now. The non-necessary snapping rewrite can also be removed. Would you replace the existing rules with the following? 
```suggestion
	<rule from="^http://(www\.)?4exhale\.org/"
		to="https://exhaleprovoice.org/" />
		
	<rule from="^http:" to="https:" />
```

jayvdb:
Ruleset enhanced as requested @cschanaj .

PR #19165, opened 2020-04-14
Trivialize rules in 4sysops.com.xml

pipboy96:


src/chrome/content/rules/4sysops.com.xml
```diff
@@ -1,22 +1,14 @@
 <!--
-	www.4sysops.com: Mismatched
+	Connection closed:
+		- feeds.4sysops.com
```
cschanaj, line 3: `curl https://feeds.4sysops.com/` gives me ssl connect error.

PR #19182, opened 2020-04-20
Fix minor UX issues on the options page

pipboy96:


pipboy96:
@zoracon Would it be a good idea to also make `<textarea>`s resize automatically?

zoracon:
> @zoracon Would it be a good idea to also make `<textarea>`s resize automatically?

I have to further discuss this point with a designer, do not think it's necessary at the moment.

PR #19193, opened 2020-04-24
Update file and contents for DogsbodyTechnology

dogsbody:
Thank you for a great tool

### DIFF
--- a/src/chrome/content/rules/DogsbodyTechnologyLtd.xml
+++ b/src/chrome/content/rules/DogsbodyTechnologyLtd.xml
@@ -1,5 +1,6 @@
 <ruleset name="Dogsbody Technology Ltd">
 	<target host="dogsbody.com" />
+	<target host="www.dogsbody.com" />
 	<target host="analytics.dogsbody.com" />
 	<target host="email.dogsbody.com" />
 	<target host="mirrors.dogsbody.com" />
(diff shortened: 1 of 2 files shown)


pipboy96:
If possible, use [HSTS preloading](https://hstspreload.org) instead. This will upgrade the connection even if the user does not have HTTPS Everywhere installed.

dogsbody:
Thank you and apologies for getting the ordering slightly wrong :+1: 

> Also, maybe you should also add `webhook01.dogsbodytechnology.com`?

This hostname will literally be deleted within the next month with no redirects. It seemed wrong to list this now or as something that was broken as it's going away but think you for raising it.

> If possible, use [HSTS preloading](https://hstspreload.org) instead. This will upgrade the connection even if the user does not have HTTPS Everywhere installed.

We are definitely working towards this goal :100:  :+1:

pipboy96:
@dogsbody Great! Thanks for applying the changes.

dogsbody:
I can't quite work out why the CI tests are failing.

I can see it's only on Chrome and `test_httpnowhere_blocks` but can't find any details of what that is testing for. Any hints appreciated :+1:

pipboy96:
@dogsbody This error can be safely ignored, it's not related to your ruleset.

PR #19212, opened 2020-04-30
Create Cht.sh.xml

IceCodeNew:
Propose a rule for cht.sh.

This site is using the wildcard certs so `*.cht.sh` should work fine.

### DIFF
--- a/src/chrome/content/rules/Cht.sh.xml
+++ b/src/chrome/content/rules/Cht.sh.xml
@@ -1,0 +1,10 @@
+<ruleset name="Cht.sh">
+	<target host="cht.sh" />
+	<target host="*.cht.sh" />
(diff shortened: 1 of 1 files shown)


pipboy96:
cheat.sh is already redirected: https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/cheat.sh.xml

IceCodeNew:
> cheat.sh is already redirected: https://github.com/EFForg/https-everywhere/blob/master/src/chrome/content/rules/cheat.sh.xml [](/EFForg/https-everywhere/blob/HEAD@{2020-04-30T19:58:53Z}/src/chrome/content/rules/cheat.sh.xml)

I see. I will delete the related part from this PR.

PR #19213, opened 2020-04-30
Create Ping.pe.xml

IceCodeNew:
Propose a rule for ping.pe.

Though this site is using the wildcard cert `*.ping.pe`, but as I run through some test it turns out that `*.ping.pe` doesn't have a CNAME record. So any other subdomains that are not list here should count the NX error.


pipboy96:
Please also add `mrtg.ping.pe`.

IceCodeNew:
> Please also add `mrtg.ping.pe`.

You got it.

PR #19214, opened 2020-04-30
Update cheat.sh.xml

IceCodeNew:
Add some basic test urls so Cheat.sh.xml can meet the CI test request.

Also add a target host `*.cheat.sh` as this site is using the wildcard cert so `*.cheat.sh` should work fine.

### DIFF
--- a/src/chrome/content/rules/cheat.sh.xml
+++ b/src/chrome/content/rules/cheat.sh.xml
@@ -1,6 +1,10 @@
 <ruleset name="cheat.sh">
 	<target host="cheat.sh" />
-	<target host="www.cheat.sh" />
+	<target host="*.cheat.sh" />
+		<test url="http://valid.cheat.sh/curl" />
+		<test url="http://example.cheat.sh/wget" />
+		<test url="http://test.cheat.sh/rpm" />
+		<test url="http://cheat.sh/apt" />
 
 	<rule from="^http:" to="https:" />
 </ruleset>


src/chrome/content/rules/cheat.sh.xml
```diff
@@ -1,6 +1,10 @@
 <ruleset name="cheat.sh">
 	<target host="cheat.sh" />
-	<target host="www.cheat.sh" />
+	<target host="*.cheat.sh" />
```
ivysrono, line 3: Could you list all to target hosts?

IceCodeNew: > Could you list all to target hosts?

I'd already run Amass to enum all the hosts that's valiable. And it turns out that any subdomain of cheat.sh is accessible.
Refer: https://github.com/EFForg/https-everywhere/pull/19212/files

ivysrono: How much subdomain?

IceCodeNew: Any subdomain is accessible, as this site's cert has the wildcard SAN and proper DNS record configure.

image
image

ivysrono: IMO, you can list all of them to test and add some comment at top.

pipboy96: In this case, you can just use `*.cheat.sh` and add a comment saying "cheat.sh has both a wildcard DNS record and a wildcard certificate".

IceCodeNew: > IMO, you can list all of them to test and add some comment at top.

I have added some subdomains for the CI test in the initial time (Please refer to the following line [L4-L7](https://github.com/EFForg/https-everywhere/pull/19214/files/31167e991d23b4c2d2d2d5ffd3576ace6e5d71dd#diff-488c30226c28290ff94eaa054ab5f348R7) ). And that's exactly what I called `update cheat.sh.xml` for this very rule.
Maybe we are having some sort of language barrier here, I really can't get your point of asking me to `list all of them`. I have been answering this several times and IMO it's just impossible to do. Please be note that `this site has a wildcard SAN and wildcard DNS record` :-)

IceCodeNew: > In this case, you can just use `*.cheat.sh` and add a comment saying "cheat.sh has both a wildcard DNS record and a wildcard certificate".

Hey there, the reason I keep the line `<target host="www.cheat.sh" />` untouched is that there are loads of rules written in this shape.
I have no idea why they are written like this, but I think it's better just to follow them.
image

IceCodeNew: @pipboy96 @ivysrono FYI, `cheat.sh` and `cht.sh` are the same site, so after I make PR #19212 and this PR #19214, I basically just sync all the changes I append to PR #19212 for passing CI test back to this PR #19214 .
IMO, it would OK to overwhelm the failed CI test and approve this PR for merge. Or we may have to lay a second glance to PR #19212 .

PR #19245, opened 2020-05-07
Create boincstats.com.xml

csavard-sudo:


src/chrome/content/rules/boincstats.com.xml
```diff
@@ -0,0 +1,38 @@
+<ruleset name="boincstats.com (partial)">
+<!--
```
ivysrono, line 2: Move comments to the top.
Use tab to replace space.

PR #19251, opened 2020-05-08
Add tekram.com

braiam:
Rule only targets the two domains that don't use redirect.
In the backend, tekram seems to use bigcommerce for everything,
which supports https and in the site it's being included as https.

It's strange that they set the hsts header with max-age=0, yet
backend consistently use https for everything.

Added all cookies as securecookies, while there are cookies
with the flag, some SESSION ones weren't.

### DIFF
--- a/src/chrome/content/rules/tekram.com.xml
+++ b/src/chrome/content/rules/tekram.com.xml
@@ -1,9 +1,8 @@
 <ruleset name="tekram.com">
-        <target host="tekram.com" />
-        <target host="www.tekram.com" />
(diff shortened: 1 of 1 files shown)


ivysrono:
Use tab to replace space.

src/chrome/content/rules/tekram.com.xml
```diff
@@ -0,0 +1,9 @@
+<ruleset name="tekram.com">
+        <target host="tekram.com" />
```
ivysrono, line 2: Use tab to replace space.

PR #19254, opened 2020-05-09
Fix JS errors found by LGTM.com

cschanaj:


pipboy96:
May I recommend another rule to add: https://eslint.org/docs/rules/eqeqeq

cschanaj:
> May I recommend another rule to add: https://eslint.org/docs/rules/eqeqeq

I am afraid this is too far from the scope of this PR. I am interested in adding `"extends": "eslint:recommended"` and fix the errors in a later PR. We shall add more _Best Practices_ rules including `eqeqeq` there as well.

PR #19286, opened 2020-05-19
Release 2020.5.19

zoracon:
Closes #19281

Issue #19110, opened 2020-04-01
Issue with redirect


Makishima:
Type: ruleset/website issue
# Domain: http://we.the-village.ru
I'm trying to follow a link in this post https://www.facebook.com/Villagemsk/posts/4292138030811979 but I'm stuck on https://amp.gs/0Xc7 without getting redirected

pipboy96:
Do you have EASE mode on? This domain doesn't support HTTPS, and is not redirected by current rulesets.

Makishima:
Yes, EASE mode is on. So how can I resolve this problem?

pipboy96:
@Makishima Please downgrade to previous release. Sadly we made a broken release and COVID-19 lockdown prevents us from withdrawing it or making a new one. More information #19029. If you need help doing that, answer here please.

zoracon:
Resolved with https://github.com/EFForg/https-everywhere/pull/19288

PR #19288, opened 2020-05-20
Release 2020.5.20

zoracon:
Related #19286 

Chrome was a corrupted release. Rereleasing to fix issue.


Makishima:
I'm sorry but release 2020.5.20 don't fix my problem from #19110 Still stuck on redirect with "404 page not found"

Issue #19029, opened 2020-03-17
Known issue: Release 2020.3.16 breaks whitelisting of some URLS and "open insecure page" button for them.


ivan-kolmychek:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

Update to `Release 2020.3.16` with new feature that allows manually whitelisting hosts seems to break whitelisting for few URLs that I'm using.

For example, if I have EASE enabled when I visit URL like `http://abcd.local:4000` (which points to something along the lines of `127.x.y.z` in my hosts file), I see the `HTTPS Everywhere noticed you were navigating to a non-HTTPS page`, even though I had whitelisted it in previous versions by pressing "open insecure page". Looking at the list of hosts I see an entry for `abcd.local:4000` in there but it doesn't seem to be doing anything. Deleting it doesn't change anything, re-adding it also seems to have no effect.

Neither pressing "open insecure page" nor pressing "open insecure page for this session only" nor manually whitelisting it seems to work, so it seems that the only currently available workaround is to disable EASE.

ivan-kolmychek:
Forgot to mention that it all is happening in Chromium 80.0.3987.132.

cschanaj:
#18760 introduced a regression for the whitelisting features, I have just filed a fix #19032.

jmgurney:
And there appears to be no plans to do a release for a couple weeks:
https://github.com/EFForg/https-everywhere/pull/19032#issuecomment-600146695

pipboy96:
@jmgurney You may thank COVID-19 for that.

jmgurney:
Thanks to @zoracon for providing a work around for this issue:
https://github.com/EFForg/https-everywhere/pull/19032#issuecomment-600841064

Installing the previous 2019.11.7 fixes this issue for me.  I do not know if auto updates will work in the future, or if this work around will have to regularly reapplied due to auto updates.

jmgurney:
Another option is to turn off EASE (Encrypt All Sites Eligible), but if you have that turned on, likely you do for a very valid reason, and disabling that is likely worse than downgrading.

[...]

PR #19292, opened 2020-05-22
Revert #19029 and #18760

### LINKED ISSUE #19291: 2020.5.20 does not fix port whitelisting issue
Type: code issue

When upgrading to 2020.5.20, I experience the issue that if Encrypt All Sites Eligible is on, that sites w/ a port number and are included in the whitelist are broken and unreachable.  If I downgrade to 2019.11.7, things work fine.  See previously related issue:
https://github.com/EFForg/https-everywhere/issues/19029

When I got to a site w/ a port number that is different that 443, and the that site uses http, but the 443 port is available, I get an error that the proxy server is refusing connections.  I normally use a proxy server to browser the web, and w/ 2019.11.7, things work.  The site w/ port number is correct listed in the HTTPS Everywher Sites Disabled list, and so should work as normal.  The only change is switching between 2020.5.20 and 2019.11.7, no other changes are made.

This is with Firefox 76.0.1.

cschanaj:
***Delete section if irrelevant***

### List related PRs if any

- #19029, #18760

- fix #19291 

cc @zoracon let's revert the change for the moment.

### DIFF
--- a/chromium/background-scripts/background.js
+++ b/chromium/background-scripts/background.js
@@ -360,7 +360,7 @@ function onBeforeRequest(details) {
 
     // Check if an user has disabled HTTPS Everywhere on this site.  We should
     // ensure that all subresources are not run through HTTPS Everywhere as well.
-    browserSession.putTab(details.tabId, 'first_party_host', uri.hostname, true);
+    browserSession.putTab(details.tabId, 'first_party_host', uri.host, true);
   }
 
   if (isExtensionDisabledOnSite(browserSession.getTab(details.tabId, 'first_party_host', null))) {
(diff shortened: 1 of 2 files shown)


zoracon:
@cschanaj 
Do you know of a good test case url I can test this on? Otherwise I will just create one where it can be tested against.

cschanaj:
@zoracon you may use the test case in #19293 . The URL in #19293 should work for this PR as well. You may add `www.pettigrew.lib.nc.us:8080` to the whitelist and observe in the Network console that request to `:8080` is over HTTP in EASE

Another test is to add `example.com:8080` to the whitelist and observe `http://example.com/` is a request over HTTPS.

PR #19296, opened 2020-05-23
Update University-of-Michigan.xml

Berkeley-Reject:
Add more subdomains of `umich.edu`


J0WI:
Can you explain why you removed some targets?

Issue #19330, opened 2020-06-03
Travis fetch test should not pass for revoked certificates


cschanaj:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

RT, see #19329 . Note that Firefox shows a `Secure Connection Failed` message when visiting https://revoked.badssl.com .

It seems fairly feasible to check the CSL and detect revoked certificate used by websites on Travis.

https://www.namecheap.com/support/knowledgebase/article.aspx/9968/38/how-to-check-the-certificate-revocation-status

The libcurl built-in checking is not very helpful, as it beaks all other checking. 

> Note that if this option is enabled but the server does not support the TLS extension, the verification will fail. 

https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYSTATUS.html

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

Issue #19034, opened 2020-03-18
Cipher mismatch in ruleset tests


J0WI:
The following error appears in multiple PRs:
`OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to host.example`

This is a regression of #18964 which has disabled AES-CBC ciphers.

Type: other

cschanaj:
I agree with @pipboy96 that we should have a tool to check for the the supported cipher list for each browser as discussed in https://github.com/EFForg/https-everywhere/pull/18964#issuecomment-589989760 . Manual cross checking require a lot time-consuming parsing work which could possibly go wrong. The ideal way is to setup a cipher checking server (with docker?) to handle the ciphers checking systematically and automatically.

PR #19339, opened 2020-06-10
Fetch Test: Completely Disabled Rulesets

zoracon:
**Date ran: 2020-04-17**
Rulesets with all hosts that are nonfunctional in ways defined by `check_rules.py` were completely disabled.

Logic: (If all hosts nonfunctional, disable)

Fetch Test Rubric: https://github.com/EFForg/https-everywhere/issues/18866

**Thinking** 
I wanted to clear out completely disabled rulesets first before moving on to round 2. That way we can make a more sane path of patching rules.

If for some reason that seems too intense, please consider the amount of hosts we load that are nonfunctional and weigh what outcomes are best for our users and load.

I am aware of issues ~#19034~ and #19330 and want to amend these before moving to round two..

**Attachments**
- Diff patch of Full Fetch test and Diff Patch of disabled rules (applied to this PR): https://www.dropbox.com/s/rfd2i4o5n04cle6/diff%20patches.zip?dl=0

Let me know if you have issues with this link - Just click "continue to view" if it prompts you to login

**Comments for review**
- The adjusted rulesets are from the effects of having dangling references and then since they were touched, being run through proper coverage themselves. I wanted this PR to be just about those fully disabled but that's not possible with strict testing structure we have to prevent coverage problems in testing.


zoracon:
Merging in this batch to wrap up this portion the work and running a ruleset release to end this segment and have the rulesets reflect the changes.

With this portion of changes, it's become apparent the fetch errors will be persistent with each push no matter how many times I run it consecutively, which means I will definitely not do these fetch tests the same as prior years and the way it was run in April. I plan to make some amendments to `check_rules.py` before running it again for smaller batches of rulesets. But completely disabled rulesets here should be cleaned up at the very least.

J0WI:
Thank you for working on this! :tada:

zoracon:
> Thank you for working on this!

And thank you for the feedback!

PR #19350, opened 2020-06-21
Remove Minalyzer.com.xml and Chambal.xml

pipboy96:
Related: #17912.

No working subdomains found.


cschanaj:
The certificate for `accounts.chambal.com` expired for a while (since 2018), would you update that as well?

pipboy96:
@cschanaj `chambal.com` has no working subdomains too.

Issue #19023, opened 2020-03-16
Update install-dev-dependencies.sh to work with Debian Buster


Jochen-A-Fuerbacher:
The script `install-dev-dependencies.sh` doesn't work with Debian Buster at the moment. The error is:

```
E: Unable to locate package python3.6-dev
E: Couldn't find any package by glob 'python3.6-dev'
E: Couldn't find any package by regex 'python3.6-dev'
E: Package 'chromedriver' has no installation candidate
```

In Debian Buster there are no `python3.6-dev` and `chromedriver` packages available. However, there are [python3.7-dev](https://packages.debian.org/buster/python3.7-dev) and [chromium-driver](https://packages.debian.org/buster/chromium-driver) packages available.

The installation script should be updated so that installs the requiered dependencies also on Debian Buster.

cschanaj:
related #18951

zoracon:
After I run a release going to work on just setting these dependencies to Python 3 support. This is snowballing into issues every where. Thank you for your patience and submitting this issue.

ShenZhouHong:
This issue can be reproduced on the additional following operating systems: Ubuntu 16.04, and Ubuntu 14.04:

## Ubuntu 16.04.6-server-amd64
```
shen@ubuntu16-test:~/https-everywhere$ ./install-dev-dependencies.sh

Warning: Installing the development dependencies for HTTPS Everywhere
may alter your system, installing requirements both within the package
management system and also external binaries.

Are you sure you want to continue? [y/N]: y

[sudo] password for shen:
Hit:1 http://us.archive.ubuntu.com/ubuntu xenial InRelease

...

Reading state information... Done
E: Unable to locate package python3.6-dev
E: Couldn't find any package by glob 'python3.6-dev'
E: Couldn't find any package by regex 'python3.6-dev'
```

## Ubuntu 14.04.6-server-amd64
```
shen@ubuntu14-test:~/https-everywhere$ ./install-dev-dependencies.sh

Warning: Installing the development dependencies for HTTPS Everywhere
may alter your system, installing requirements both within the package
management system and also external binaries.

Are you sure you want to continue? [y/N]: y

Ign http://us.archive.ubuntu.com trusty InRelease

...

Reading state information... Done
E: Unable to locate package python3.6-dev
E: Couldn't find any package by regex 'python3.6-dev'
```

PR #19368, opened 2020-06-25
switched to python3.8

vporton:
I was assigned https://github.com/acatarineu/tor-browser/issues/10

While working on it, I was unable to install `lxml` for Python 3.6.

As a solution, I did systematic replace `python3.6` -> `python3.8` in the source of HTTPSE.

I will base my solution to https://github.com/acatarineu/tor-browser/issues/10 on Python 3.8. If this is not acceptable, we can downgrade back to Python 3.6 later. I will work with 3.8.

Maybe wee should write just `python3` not `python3.6` or `python3.8`?


zoracon:
I will take the time to sit down and look historically at why we just don't use `python3.6`. I have wanted to use `python3` for some time but there were various other reasons as to why it stayed. There are various issues like this one:
- https://github.com/EFForg/https-everywhere/issues/19023#issuecomment-599675378

And reasoning like here: 
https://github.com/EFForg/https-everywhere/issues/6749#issuecomment-363281281

Which is arguably dated, I will push this to the forefront and out of the backlog.

zoracon:
Also, the main reason for failure is due to Travis also is hardcoded to `python3.6`
https://github.com/EFForg/https-everywhere/blob/master/.travis.yml#L5

zoracon:
Closing this for the time being because we have to do extensive testing to Python3 migration. A sweeping change with just bumping up the version isn't addressing that portion. But we are in the midst of changing and I am noting the issue with packages like `lxml`. Hopefully I can have it up soon.

PR #19374, opened 2020-06-26
Create tenhou.net

mk0x9:
New ruleset for [tenhou.net](https://tenhou.net/).

### DIFF
--- a/src/chrome/content/rules/Tenhou.net.xml
+++ b/src/chrome/content/rules/Tenhou.net.xml
@@ -1,0 +1,12 @@
+<!--
+	mismatch: blog.tenhou.net
+-->
+
+<ruleset name="Tenhou.net">
+	<target host="tenhou.net" />
+	<target host="cdn.tenhou.net" />
+
+	<securecookie host=".+" name=".+" />
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Tenhou.net.xml
```diff
@@ -0,0 +1,12 @@
+<!--
+	mismatch: blog.tenhou.net
+-->
+
+<ruleset name="Tenhou.net">
+	<target host="tenhou.net" />
+	<target host="cdn.tenhou.net" />
+
```
pipboy96, line 8:
```suggestion
	<target host="opid.tenhou.net" />

```

mk0x9: I've seen this domain in Sublist3r output, but I'm not really sure what is it. It's serving empty 200 HTTP OK, this might be non-functioning private OpenID endpoint. Tenhou.net itself a online [riichi mahjong](https://en.wikipedia.org/wiki/Japanese_Mahjong) server, and I've never stumbled on this subdomain whenever I played there.

I would suggest not to add it, as it might break something and not to bloat the extension.

PR #19388, opened 2020-07-10
Create filipvel.ddns.net.xml

Silverdimond:


pipboy96:
@Silverdimond This website doesn't seem to load. Also, if it's your website, the recommended way to redirect it is to use [HSTS preloading](https://hstspreload.org/), since it doesn't require HTTPS Everywhere to be installed.

PR #19399, opened 2020-07-17
Add a ruleset for yes24.com

andromedarabbit:
https://yes24.com


andromedarabbit:
`Non-2xx HTTP code: http://yes24.com/ (200) => https://yes24.com/ (404)` 

It seems like this should not be merged. I would just close the pull request.

PR #19400, opened 2020-07-19
Create yuji.ne.jp ruleset

yujixr:
This new ruleset adds rules for [yuji.ne.jp](https://yuji.ne.jp)


J0WI:
Reopen to trigger Travis

PR #19410, opened 2020-07-26
Delete wuala.com

jayvdb:
https://wuala.com/ says

> Wuala has shut down on November 15, 2015

### List related PRs if any

- Extracted from https://github.com/EFForg/https-everywhere/pull/18949


J0WI:
But for `wuala.com` and `www.wuala.com` the rule is still valid.

jayvdb:
But the premise of the rule, which is that it is a commonly used part of the internet, is no longer valid.  It is just a single page in two languages, and everyone who needs to read it, read it ages ago.

It already has http->https, just doesnt have hsts.  If we filled https-everywhere with all sites in that category, irrespective of how many users it has, the dataset would be so enormous that it is unusable.

pipboy96:
@jayvdb 

> But the premise of the rule, which is that it is a commonly used part of the internet, is no longer valid.

There is no such premise required for adding a rule to HTTPS Everywhere, hence "Everywhere" in its name. We accept all correctly functioning rulesets, including ones targeting very obscure websites with next to no visitors, and only reason for not including a website that supports HTTPS is the website having a HSTS header and being preloaded in both Chrome, Firefox and Tor Browser.

J0WI:
See also https://www.eff.org/https-everywhere/faq#why-does-https-everywhere-include-rules-for-sites-like-paypal-that-already-require-https-on-all-their-pages

jayvdb:
This isnt about "accepting" rules.  This is about deleting a *broken* rule for a website which no longer exists.  There is a new host owned by someone else on the same name, and somehow that is preventing it being deleted.

And of course I can find many cases where maintainers have deleted rules for living hosts.  So I'll leave this broken rule to the maintainers.

PR #19419, opened 2020-07-31
Update Nautil.us.xml

rany0:


src/chrome/content/rules/Nautil.us.xml
```diff
@@ -18,16 +14,20 @@
...
+	<target host="nautil.us" />
+	<target host="www.nautil.us" />
 		<!--exclusion pattern="http://www\.nautil\.us/" /-->
 
 
 
 	<rule from="^http://static\.nautil\.us/"
 		to="https://d3chnh8fr629l6.cloudfront.net/" />
+	<rule from="^http[s]?://www\.nautil\.us/"
+		to="https://nautil.us/" />
```
zoracon, line 30: Please remove an HTTPS to HTTPS rule. Rather than taking in a HTTPS to HTTPS request,either set an exclusion pattern, or set it `^http..` and not `^http[s]...`. HTTPS Everywhere should only attempt to redirect HTTP requests.

rany0: Is this acceptable now?

zoracon (approved):

PR #19428, opened 2020-08-02
Create radio.garden.xml

TotalCaesar659:


war59312:
Radio Garden is current broken. Stuck in loop. https forward back to http

PR #19438, opened 2020-08-03
Create glitch.me.xml

TotalCaesar659:
http://concrete-hardboard.glitch.me/

### DIFF
--- a/src/chrome/content/rules/glitch.me.xml
+++ b/src/chrome/content/rules/glitch.me.xml
@@ -1,4 +1,5 @@
 <ruleset name="glitch.me">
+	<target host="glitch.me" />
 	<target host="*.glitch.me" />
 
 	<rule from="^http:" to="https:" />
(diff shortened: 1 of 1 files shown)


zoracon:
I don't normally promote wildcards but in this case where so many use Glitch.me, it fits in here. If you don't mind converting this to support the many `glitch.me` sites I think that would be more beneficial.

zoracon (changes requested):
I don't normally promote wildcards but in this case where so many use Glitch.me, it fits in here. If you don't mind converting this to support the many glitch.me sites I think that would be more beneficial.

pipboy96:
@TotalCaesar659 Use a wildcard! There are way too many domains to redirect individually (more than 1000).

zoracon (changes requested):
Just need to add some tests for the wildcard.

Have a list here from enumerating the domain:
http://spiky-well-vein.glitch.me/
http://pwainstall.glitch.me
http://vyper.glitch.me

PR #19446, opened 2020-08-03
Create scsportsmans.org.xml

TotalCaesar659:


zoracon (changes requested):
Please address:
```
ERROR src/chrome/content/rules/scsportsmans.org.xml: Time: 2020-08-03 01:58:23

 Fetch error: http://scsportsmans.org/ => https://www.scsportsmans.org/: (6, 'Could not resolve host: www.scsportsmans.org')
```

PR #19448, opened 2020-08-03
Create radeon.com.xml

TotalCaesar659:


zoracon (approved):

zoracon (commented):
Looks like base domain has timeout issues:
```
ERROR src/chrome/content/rules/radeon.com.xml: Fetch error: http://www.radeon.com/ => https://www.radeon.com/: (28, 'Operation timed out after 30001 milliseconds with 0 bytes received')
```

Will run check again to see if this is amended and was temporary.

PR #19455, opened 2020-08-09
Create aicevote.com ruleset

yujixr:
This new ruleset adds rules for [aicevote.com](https://aicevote.com)


pipboy96:
@yujixr Thanks for contribution! Since you are the owner of the website, and your website is already sending the correct HSTS header, there's no need to add it to HTTPS Everywhere. Instead, please add it to https://hstspreload.org/ (I already did that, so you don't have to do that on your own).

PR #19458, opened 2020-08-10
Delete Encrypt_Everything.ca.xml

kla951:


pipboy96:
@kla951 The website seems to still work, there's no need to delete it, instead the ruleset should be updated.

pipboy96:
Replaced by #19467.

PR #19474, opened 2020-08-12
Update External Dependency: Pako

zoracon:
This is an attempt to monitor this dependency in a way that can catch any vulnerabilities with this version of Pako in case it needs to be updated.

If, for some reason this is not a good approach for a historical reason I am missing, please let me know. This not something I am particularly pushing if there is a good reason why I shouldn't. My thinking here is that our automated security flows will catch an issue before I personally can and provide an automated PR I can review.


cschanaj:
I remember that the version of the dependencies to the extension is hard-coded to maintain a deterministic build. See discussions in #13560

zoracon:
@cschanaj This makes sense. I think I will just simply add Pako and CodeMirror to package.json to be able to be monitor security issues within our flow with these versions. And keep the what is there in the external directory for now until it needs to be updated as needed by an alert.

PR #19510, opened 2020-08-22
Add rules for Archive.today and related sites

Nintendofan885:
Add rules for Archive.today and related sites

each domain has the following subdomains:
archive.xx
www.archive.xx
blog.archive.xx


Nintendofan885:
Whoops. Turns out it already exists as
https://github.com/EFForg/https-everywhere/blob/a56624ae44f2c6d343089ee2eed23da7eb96163d/src/chrome/content/rules/Archive.is.xml

PR #19512, opened 2020-08-22
Add intr0.com.xml

Voltairine-de-Cleyre:
`add intr0.com & *.intr0.com to HTTPS Everywhere`


pipboy96:
This domain uses HSTS preloading, and is already redirected by all browsers as the result (see https://hstspreload.org/?domain=intr0.com).

PR #19522, opened 2020-08-24
Hotfix: Github Labeller Action

zoracon:
Trying to solve the issue of forked repo PR's access to Action jobs


zoracon:
After doing more detailed research: Realizing no one with a forked repo will be able to access the workflow in the Alexa Labeller :(. Will remove the Action for now since it seems this is a wall we can't pass. Github Token has READ access only as well, so blocking ability to label.

Reference:
"Secrets are not passed to workflows that are triggered by a pull request from a fork."
https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

PR #19479, opened 2020-08-14
Adding error message when the host being disabled is invalid

Hainish:

PR #19531, opened 2020-08-29
Fix Travis lint errors and regressions on wildcards, follow up #19479

cschanaj:
This PR fixes a regression from #19479 on wildcard support for disabled site (added since #18619) and some Travis lint errors. 

@zoracon is the UX redesign work being blocked by the pandemic? It would be much nicer to have a consistent CSS styles/ framework to work with when adding new UI elements. IMHO, it would be better if we could provide visual clues to our users that wildcards are supported.

### DIFF
--- a/chromium/pages/options/ux.js
+++ b/chromium/pages/options/ux.js
@@ -331,27 +331,19 @@ document.addEventListener("DOMContentLoaded", () => {
   const disabled_site_input = document.getElementById("disabled-site");
   const add_disabled_site_invalid_host = document.getElementById('add-disabled-site-invalid-host');
   disabled_site_input.setAttribute("placeholder", chrome.i18n.getMessage("options_enterDisabledSite"));
-  function isValidHost(host) {
-    try {
-      new URL(`http://${host}/`);
-      return true;
-    } catch {
-      return false;
-    }
-  }
-  add_disabled_site.addEventListener("click", function() {
-    const host = disabled_site_input.value;
 
-    if (isValidHost(host)) {
-      hide(add_disabled_site_invalid_host);
-      sendMessage("disable_on_site", disabled_site_input.value, okay => {
-        if (okay) {
-          chrome.tabs.reload();
-        }
-      });
-    } else {
-      show(add_disabled_site_invalid_host);
-    }
+  add_disabled_site.addEventListener("click", function() {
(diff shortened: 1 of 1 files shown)


chromium/pages/options/ux.js
```diff
@@ -331,27 +331,19 @@ document.addEventListener("DOMContentLoaded", () => {
   const disabled_site_input = document.getElementById("disabled-site");
   const add_disabled_site_invalid_host = document.getElementById('add-disabled-site-invalid-host');
   disabled_site_input.setAttribute("placeholder", chrome.i18n.getMessage("options_enterDisabledSite"));
-  function isValidHost(host) {
-    try {
-      new URL(`http://${host}/`);
-      return true;
-    } catch {
```
pipboy96, line 338: It's possible to fix the lint error without removing this code completely, just replace this line with:

```
    } catch (_) {
```

cschanaj: It seems that the URL constructor throw TypeError only. And it will not throw error for some common invalid URLs.  

```node
> try { new URL('https://*...,example.com/'); console.log('ok'); } catch(error) { console.log(error); }
ok

> try { new URL(123); console.log('ok'); } catch(error) { console.log(error); }
TypeError [ERR_INVALID_URL]: Invalid URL: 123
    at onParseError (internal/url.js:256:9)
    at new URL (internal/url.js:332:5)
    at repl:1:7
    at Script.runInThisContext (vm.js:120:18)
    at REPLServer.defaultEval (repl.js:433:29)
    at bound (domain.js:427:14)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at REPLServer.onLine (repl.js:760:10)
    at REPLServer.emit (events.js:327:22)
    at REPLServer.EventEmitter.emit (domain.js:483:12) {
  input: '123',
  code: 'ERR_INVALID_URL'
}
undefined
```

The `disable_on_site` will validate the hostname and returns an error by default. IMHO we should rely on a single-source-of-truth implementation for validations.

https://github.com/EFForg/https-everywhere/blob/2bfa7efe3575f8b38827f5eb278d47ff14608095/chromium/background-scripts/background.js#L899-L907

pipboy96: Seems like a good idea indeed.

PR #19533, opened 2020-08-29
Rename and update Justia.xml

pipboy96:
Related: #17912.


src/chrome/content/rules/Justia.com.xml
```diff
@@ -0,0 +1,158 @@
+<ruleset name="Justia.com (partial)">
```
cschanaj, line 1: Would you add a comment that this domain has a wildcard DNS records and uses a wildcard certificate? https://dns.google.com/query?name=*.justia.com&type=A&dnssec=true

cschanaj (changes requested):

src/chrome/content/rules/Justia.com.xml
pipboy96: @cschanaj Done.

PR #19534, opened 2020-08-29
Bump ECMAScript version to ES2019 to make ESLint pass

pipboy96:
See: https://travis-ci.org/github/EFForg/https-everywhere/jobs/722242042; https://travis-ci.org/github/EFForg/https-everywhere/jobs/722242042.

### DIFF
--- a/chromium/.eslintrc.json
+++ b/chromium/.eslintrc.json
@@ -5,7 +5,7 @@
     "webextensions": true
   },
   "parserOptions": {
-    "ecmaVersion": 2017,
+    "ecmaVersion": 2019,
     "sourceType": "script"
   },
   "rules": {


pipboy96:
@Hainish @zoracon It seems we may need to replace the ESLint version we use with a newer one.

cschanaj:
It seems that the browser supports for ES2019 features are fine for our migrations. 

https://kangax.github.io/compat-table/es2016plus/#test-optional_catch_binding

P.S. #19531 will remove the ES2019 code and pass the current eslint requirements.

cschanaj:
@zoracon @pipboy96 eslint is failing with another message now.

```
error  Parsing error: Invalid ecmaVersion
```

PR #19536, opened 2020-08-30
Updated/extended Oregon State University rulesets

MarcelWaldvogel:
* Re-enabled rulesets, as those certificate problems have been fixed
* Mixed-content has been fixed
* osufoundation.org now self-redirects to www.osufoundation.org
* osulibrary.oregonstate.edu works and redirects to
  library.oregonstate.edu
* web.engr.oregonstate.edu and ecampus.oregonstate.edu now work
* More consistently cross-referenced the OSU files

PR #19540, opened 2020-08-31
Re-enabled OSU Foundation

MarcelWaldvogel:
* osufoundation.org now works, both with/without www prefix

This is part of a series of splits resulting from #19536 as asked

### DIFF
--- a/src/chrome/content/rules/osufoundation.org.xml
+++ b/src/chrome/content/rules/osufoundation.org.xml
@@ -2,10 +2,13 @@
 	For other Oregon State University coverage, see
 
 
-	^osufoundation.org: dropped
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/osufoundation.org.xml
```diff
@@ -2,23 +2,15 @@
...
-
-	<!--	Complications:
-				-->
 	<target host="osufoundation.org" />
+	<target host="www.osufoundation.org" />
 
 
-	<rule from="^http://osufoundation\.org/"
-		to="https://www.osufoundation.org/"/>
-
```
pipboy96, line 21: This rule is still required; https://osufoundation.org still times out.

pipboy96 (changes requested):

PR #19568, opened 2020-09-09
Feature: EASE Copy Url

### LINKED ISSUE #19385: Copy button on EASE warning page
[//]: #

Type: Feature request

### Description
It would be nice, if HTTPS-everywhere has a copy button present just beside of URL (for which it is showing warning message) to copy URL directly to clipboard.

### Use-case
Sometimes I need know to URL for which HTTPS-everywhere showing warning message in EASE mode. HTTPS-everywhere do show URL on warning page, but longer URL replaced with `...` sign.

ease-copy-button

Currently, I've to extract this URL from browser's address bar and then decode it, which may be harder for average user, but adding a copy button make things easier for average users.

PS: What do you think, is it good idea to add a copy button? If so, can I make PR for this?

zoracon:
Closes #19385

Adds feature to copy unencrypted link from EASE warning page

### DIFF
--- a/chromium/pages/cancel/ux.js
+++ b/chromium/pages/cancel/ux.js
@@ -63,7 +63,7 @@ function displayURL() {
 
   // Copy URL Feature on EASE
   copyButton.addEventListener("click", async event => {
-    copyButton.innerText = 'Copy URL';
+    copyButton.innerText = chrome.i18n.getMessage("cancel_copy_url");;
     if (!navigator.clipboard) {
       // Clipboard API not available
       return
--- a/chromium/pages/cancel/ux.js
+++ b/chromium/pages/cancel/ux.js
@@ -45,6 +45,7 @@ function displayURL() {
   const originURLLink = document.getElementById('url-value');
   const openURLButton = document.getElementById('open-url-button');
   const openHttpOnce = document.getElementById('http-once-button');
+  const copyButton = document.getElementById('copy-url');
   const url = new URL(originURL);
 
   originURLLink.innerText = originURL;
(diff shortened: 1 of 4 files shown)


chromium/pages/cancel/ux.js
```diff
@@ -60,6 +61,22 @@ function displayURL() {
...
+      return
+    }
+    try {
+      await navigator.clipboard.writeText(originURL);
+      copyButton.innerText = chrome.i18n.getMessage("cancel_copied_url");
+    } catch (err) {
+      console.log(err);
+      console.log('failed');
+    }
+  });
```
pipboy96, line 78:
```suggestion
  // Copy URL Feature on EASE

  function copyLinkAlternate() {
    let isSuccessful = false;

    const sel = window.getSelection();

    try {
      sel.removeAllRanges();

      const range = document.createRange();
      range.selectNode(originURLLink);

      sel.addRange(range);

      isSuccessful = document.execCommand("copy");

      sel.removeAllRanges();

      return isSuccessful;
    } catch (err) {
      console.error(err);

      sel.removeAllRanges();

      return false;
    }
  }

  async function copyLink() {
    try {
      await navigator.clipboard.writeText(originURL);
      return true;
    } catch (err) {
      return copyLinkAlternate();
    }
  }

  let restoreTimeout = null;

  copyButton.addEventListener("click", async () => {
    if (await copyLink()) {
      copyButton.innerText = chrome.i18n.getMessage("cancel_copied_url");

      if (restoreTimeout !== null) {
        clearTimeout(restoreTimeout);
      }

      restoreTimeout = setTimeout(() => {
        copyButton.innerText = chrome.i18n.getMessage("cancel_copy_url");
      }, 1500);
    }
  });
```

pipboy96 (changes requested):
`navigator.clipboard.writeText` seems to be supported by all browsers we support, but just in case let's also add a backup method. Also I added a timeout which resets the button text back after 1.5 seconds.

chromium/pages/cancel/ux.js
zoracon: I assume `execCommand` was used for older browsers. I will accept the commit, pull down and test.

PR #19593, opened 2020-09-13
Rename and update Ecwid.com.xml

cschanaj:
#19578


src/chrome/content/rules/Ecwid.com.xml
```diff
@@ -0,0 +1,20 @@
+<!--
+	Non-functional hosts:
+		Incomplete certificate chain error:
+		- ideas.ecwid.com
```
J0WI, line 4: Looks like customers get their own subdomain and all of them support HTTPS

cschanaj: Do you think adding a top comment on this would be sufficient to inform the future contributors? there is over 1k subdomains discovered by Sublist3r, and it will take significant time to test them all.

J0WI: Either this or a wildcard target.

cschanaj: I would prefer to keep this ruleset as-is. Thank you.

J0WI (approved):

PR #19594, opened 2020-09-13
Rename and update CityEmail.com.xml

cschanaj:
#19578

### DIFF
--- a/src/chrome/content/rules/CityMail.com.xml
+++ b/src/chrome/content/rules/CityMail.com.xml
@@ -2,16 +2,15 @@
 	For other Wizard Internet Services coverage, see Wizard-Internet-Services.xml.
 
 -->
-<ruleset name="City Mail">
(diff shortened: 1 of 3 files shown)


src/chrome/content/rules/CityMail.com.xml
```diff
@@ -2,16 +2,15 @@
 	For other Wizard Internet Services coverage, see Wizard-Internet-Services.xml.
 
 -->
-<ruleset name="City Mail">
+<ruleset name="CityMail.com">
```
pipboy96, line 5:
```suggestion
<ruleset name="CityEmail.com">
```

pipboy96 (changes requested):

src/chrome/content/rules/CityMail.com.xml
cschanaj: good catch. thanks :+1:

Issue #19416, opened 2020-07-29
PR_END_OF_FILE_ERROR error doesn't trigger EASE error page


kpcyrd:
### Steps to reproduce
- Make sure "HTTPS Everywhere" and "Encrypt all sites eligible" is on
- Try to visit http://blog.cmpxchg8b.com/ (pasting the url into your address bar if necessary)

### Expected results
I'm able to visit the blog over insecure http

### Actual results
https-everywhere detects _something_ on port 443 and redirects, but firefox then fails with `An error occurred during a connection to blog.cmpxchg8b.com. PR_END_OF_FILE_ERROR`.

### Setup
Firefox Developer Edition on Linux

pipboy96:
EASE mode is expected to break websites often. The only solution is manually adding an exception for this website (go to HTTP version of the website and click "Disable HTTPS Everywhere for this website" in the HTTPS Everywhere popup menu).

kpcyrd:
I'd argue the 2nd E in EASE doesn't work correctly, if the server does not respond to a tls client hello https-everywhere should display the "are you sure you want to visit this insecure site" instead of forcibly redirecting. Teaching users to disable https-everywhere if something doesn't work is encouraging bad security practice, especially since it's not obvious that https-everywhere is at fault here.

pipboy96:
@kpcyrd I think it's possible to add `PR_END_OF_FILE_ERROR` to the list of errors intercepted by EASE, but I'm not sure if it will have undesired side effects. cc @zoracon.

pipboy96:
For reference: Chrome error code on this page is `ERR_CONNECTION_CLOSED`.

zoracon:
I'll investigate this error. We have a static list of SSL Errors that occur that we add to. I will audit if this should make the list.

zoracon:
@pipboy96, if you have time this week. Could you replicate this error? I am getting a route to HTTPS for this host. Even when I'm not using EASE.

pipboy96:
@zoracon This website supports HTTPS now.

kpcyrd:
You can replicate the original error like this:
```
socat tcp-listen:443,reuseaddr,fork stdio < /dev/null > /dev/null
```
This makes it impossible to visit the website hosted on port 80 if EASE is enabled.

PR #19605, opened 2020-09-21
Add cmpxchg8b.com.xml

pipboy96:
Related: #19416.

### DIFF
--- a/src/chrome/content/rules/cmpxchg8b.com.xml
+++ b/src/chrome/content/rules/cmpxchg8b.com.xml
@@ -1,0 +1,11 @@
+<!--
+	Mismatched:
+		- cmpxchg8b.com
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/cmpxchg8b.com.xml
```diff
@@ -0,0 +1,13 @@
+<!--
+	Mismatched:
+		- cmpxchg8b.com
+		- www.cmpxchg8b.com
+
+	Incomplete certificate chain:
```
J0WI, line 6: I cannot reproduce anymore

pipboy96: @J0WI Fixed.

PR #19621, opened 2020-09-28
Remove Smowtion

zoracon:
Related #17912 

HTTPS Not Supported

<details>
<summary>Tl;DR Summary</summary>
<blockquote>
...In other words, all rulesets that have a to= using the s3.amazonaws.com hostname will stop working. We should update all of these before then. Here's the current list of 115 affected rulesets:
</blockquote>
</details>


src/chrome/content/rules/Smowtion.xml
```diff
@@ -35,17 +40,9 @@
...
-
 
 	<rule from="^http://ad\.smowtion\.com/"
 		to="https://ad.yieldmanager.com/" />
 
-
-	<rule from="^http://px\.smowtion\.com/"
-		to="https://s3.amazonaws.com/px.smowtion.com/" />
-
 	<rule from="^http:" to="https:" />
```
cschanaj, line 47: this trivial rule does not apply to any target, it can be safely removed.

src/chrome/content/rules/Smowtion.xml
```diff
@@ -14,12 +14,17 @@
 	Nonfunctional subdomains:
 
 		- (www.) *
+    - applicants
```
cschanaj, line 17: the indents for the new comment are off here.

cschanaj (changes requested):
There is no target works over HTTPS and the snapping rewrite serves advertisement content only. I am inclined to think that you may remove this ruleset completely. I shall merge this either you fix the Travis error/ remove this ruleset.

PR #19626, opened 2020-09-30
Remove S3 References from Scribd

zoracon:
Related #17912 

<details>
<summary>Tl;DR Summary</summary>
<blockquote>
...In other words, all rulesets that have a to= using the s3.amazonaws.com hostname will stop working. We should update all of these before then. Here's the current list of 115 affected rulesets:
</blockquote>
</details>

### DIFF
--- a/src/chrome/content/rules/Scribd.com.xml
+++ b/src/chrome/content/rules/Scribd.com.xml
@@ -1,15 +1,5 @@
 <!--
 	CDN buckets:
-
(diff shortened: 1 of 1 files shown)


cschanaj:
For the records, `*.scribd.com` does not have a wildcard DNS records so it would be better to enumerate the subdomains insteads of using a wildcard target. However, I will merge this given the urgency of #17912 . Improvement to this rulesets shall be made in another PR shall any contributor be interested.

PR #19634, opened 2020-10-01
Update Intuit.xml related ruleset

cschanaj:
#17912


zoracon (approved):
I just put my branch up, but you beat me to it

cschanaj:
We should definitely avoid merging new complex rulesets like this in the future as it could be a nightmare to most contributors/ maintainers to update/ modify. It took me literally hours to update this ruleset even when using multiple scripts and robots. Anyway, it is great to see #17912 is completed. thanks for merging.

Issue #7691, opened 2016-11-24
Automatic generation of rulesets


galeksandrp:
https-everywhere-bot got mad?????

@J0WI @jeremyn

[Submit app comparison (+sources)](https://github.com/EFForg/https-everywhere/issues/7691#issue-191596811) | [Generators comparison (+sources)](https://github.com/EFForg/https-everywhere/issues/7691#issuecomment-267221375) | [Domain enumerators](https://github.com/EFForg/https-everywhere/pull/4279#issuecomment-250968108)
--- | --- | ---

Submit app comparison

_Note that :white_check_mark: is clickable_

Backend | Frontend | Prod | License | Dependencies | GitHub caps
--- | --- | --- | --- | --- | ---
https://github.com/galeksandrp/https-everywhere/commit/webhook | [galeksandrp/httpse][4] | [:white_check_mark:][3] | MIT | Travis GitHub Docker (https://github.com/galeksandrp/https-everywhere/commit/docker-httpse)
https://github.com/galeksandrp/https-everywhere/commit/webhook-2 | https://github.com/galeksandrp/https-everywhere/commit/gh-pages | [:white_check_mark:][1] | MIT | Travis GitHub Docker (https://github.com/galeksandrp/https-everywhere/commit/docker-httpse) | public_repo
https://github.com/galeksandrp/https-everywhere/commit/webhook-client | https://github.com/galeksandrp/galeksandrp.github.io/commit/master | [:white_check_mark:][2] | MIT | Travis GitHub Docker (https://github.com/galeksandrp/https-everywhere/commit/docker-httpse) | public_repo email

Generators support

Author/Link | `@Foorack` https://github.com/EFForg/https-everywhere/issues/6912#issuecomment-248211667 | https://github.com/gloomy-ghost/ew/commit/master | https://github.com/galeksandrp/https-everywhere/commit/check-sublist3r | `@pgerber` #7706 | https://github.com/Foorack/httpse-ruleset-generator/commit/master | https://github.com/guiweber/ssl-yeah/commit/master
--- | --- | --- | --- | --- | --- | ---
[galeksandrp/httpse][4] | [:heavy_check_mark:][template-foorack] || [:white_check_mark:][template] | [:heavy_check_mark:][template-pgerber] | [:heavy_check_mark:][template-foorack2]
[...]

Issue #6912, opened 2016-09-19
Would you provide one tool to generate rules?


ivysrono:
To search subdomains , get cert status, auto check in local, and so on.

jeremyn:
That would be nice. I mentioned wanting those sorts of tools [here](https://github.com/EFForg/https-everywhere/issues/6322#issuecomment-246866651).

fuglede:
I'm not sure what you mean by cert status and auto-checking in local exactly, but it sounds like what our tests achieve? Those you can (and probably should) run locally.

It would be nice to have a subdomain finder utility as well though (cf. #4279).

jeremyn:
It should be straightforward for one script, given some wildcard domains, to generate a ruleset that pre-resolves the to-do list in the various pull requests I link in issue #6863 , including the one for [`udn.com`](https://github.com/EFForg/https-everywhere/pull/6824#issuecomment-246125911) that @renyouguo was involved with.

Creating a script to use Google to filter domains, run curl to note basic failures, and create an XML file to note the results should be within the skill-set of a beginner programmer. It would be slightly more complicated to identify mixed content and to crawl sites looking for `static.`-type domains, but not too much harder. [Here](https://www.google.com/#q=selenium+mixed+content)'s a simple Google search discussing solutions to the mixed content problem. Humans should only be needed to write `exclusion`s and custom `rule`s.

The real challenge for anyone wanting to make this happen is being willing to focus on a medium-sized task like this, make it work within the existing ecosystem, and endure the inevitable bikeshedding from myself and others.

Foorack:
@jeremyn This idea has been catching my interest for a while. I've done an attempt on it already but it's not finished yet. It can be optimized a lot but this is **only a proof-of-concept**. To test it I have been running it against a portion of the europa.eu domain as it has 788 functional subdomains. :+1: 
My personal test approach is as following:

1. First use Sublist3r (without the bruteforce option!) which uses Google, Bing, dnsdumpster as well as a few other sources to find all subdomains. This does generate some false-positives which I sort out in next step.
2. Curl with head probe to see if the subdomain responds to basic http. If not then should be safe to assume it was a false-positive.
3. Compiles ruleset with python script.

I still planning on implementing automatic mixed-content checking with bramu's mixed-content scan https://github.com/bramus/mixed-content-scan but most features such as different content, https->http redirect, timeouts, certificate-chain errors and so on.

[...]

PR #19638, opened 2020-10-01
Update CONTRIBUTING.md

zoracon:
Update Nonfunctional Hosts Standard

We have this in generated rules with make-trivial-rule script. It should be reflected in the Docs

Related: #14936

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -281,18 +281,20 @@ It is useful to list hosts that do not work in the comments of a `ruleset`.  Thi
 For easy reading, please avoid using UTF characters unless in the rare instances that they are part of the hostname itself.
 
 Example:
-
 ```xml
 <!--
-	Nonfunctional hosts in *.$lower:
-		# Example: foobar.$lower (m)
-
-	h: http redirect
-	m: certificate mismatch
-	r: connection refused
-	s: self-signed certificate
-	t: timeout on https
-
+	Invalid certificate:
+		8marta.glavbukh.ru
+		forum2.glavbukh.ru (incomplete certificate chain)
+	Redirect to HTTP:
+		8marta2013.glavbukh.ru
+		den.glavbukh.ru
+	Refused:
+		e.glavbukh.ru
+		www.e.glavbukh.ru
+	Time out:
+		psd.glavbukh.ru
+		str.glavbukh.ru
 -->
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -284,21 +284,14 @@ Example:
 
 ```xml
 <!--
-				Invalid certificate:
-								8marta.glavbukh.ru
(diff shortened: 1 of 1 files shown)


pipboy96:
@zoracon Let's change `make-trivial-rule` to match the docs instead, since we usually use that format.

pipboy96:
Also, is there evidence that this script is still used by people? Maybe we can completely remove it?

zoracon:
> Also, is there evidence that this script is still used by people? Maybe we can completely remove it?

@pipboy96 I am not sure who uses this and who doesn't. I am going to research the background on why this comment for hosts was structured this way. There's a lot of discussion from a few years ago concerning this topic. I want to decide on a simplistic format going forward.

So my question is the format in `make-trivial-rule` more useful or is the more verbose format in the document better to keep?

zoracon:
To add, we also reference the `make-trivial-rule` tool in our Update Channels documentation
https://github.com/Hainish/https-everywhere/blob/master/docs/en_US/ruleset-update-channels.md

Which being actively experimented with at the moment.

cschanaj:
The style in make-trivial-rule was being discussed in #6868 and the _agreement_ (tagged master list) was made by a number of contributors. However, I agree with @pipboy96  that the style in the documentation (grouped list) and its variants are more commonly seen. Personally, I prefer the style in the current documentation. 

I believe the grouped list styles gained its popularity after the agreement is made. There was a time many contributors created their own bots for automatic generation of rulesets (#7691), and the grouped list styles turned out to be easier to parsers and formatters (e.g. https://github.com/EFForg/https-everywhere/issues/6912#issuecomment-248211667) 

cc @jeremyn @J0WI @Hainish

zoracon:
Thank you for the background @cschanaj. I am inclined to go with the previous consensus instead of going down that path again. The conversation seemed to have come to a good middle ground between the styles. I know the more verbose categorization seems to be dominant, but going forward would like to create a script that can easily parse this out and enforce this style.

CONTRIBUTING.md
```diff
@@ -281,25 +281,20 @@ It is useful to list hosts that do not work in the comments of a `ruleset`.  Thi
...
+		forum2.glavbukh.ru (incomplete certificate chain)
+	Redirect to HTTP:
+		8marta2013.glavbukh.ru
+		den.glavbukh.ru
+	Refused:
+		e.glavbukh.ru
+		www.e.glavbukh.ru
+	Time out:
+		psd.glavbukh.ru
+		str.glavbukh.ru
```
J0WI, line 297: Should we use an [example domain](https://www.iana.org/domains/reserved) or a domain we control for this?

zoracon: We can, I can adjust to reflect that or you can.

J0WI: #19899 feel free to adjust/copy my changes.

zoracon:
Merged in related PR, closing this one out

PR #19660, opened 2020-10-09
Clean up UTF characters from ruleset files.

pipboy96:
This PR doesn't replace the Unicode arrows (← and →) and various superscript characters due to a very high number of affected ruleset files (on the order of thousands). They will be addressed in a separate PR.

For motivation, see [ruleset style guide](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#ruleset-style-guide):
>For easy reading, please avoid using UTF characters unless in the rare instances that they are part of the hostname itself.


pipboy96:
@zoracon There are also some Unicode characters (’ instead of ') in `make.sh`:

https://github.com/EFForg/https-everywhere/blob/9bc37342929eff8bc9b3c6ac85401eff9a2ccc97/make.sh#L24
https://github.com/EFForg/https-everywhere/blob/9bc37342929eff8bc9b3c6ac85401eff9a2ccc97/make.sh#L37

Should they also be replaced?

J0WI (approved):

cschanaj (approved):

zoracon:
Apologies for letting this get stale, if you could solve that conflict I can push.

zoracon:
Seeing fetch errors with several rulesets. Going to clean these up and do a follow up release that will occur after the one that's planned now.

PR #19693, opened 2020-10-17
SmartCard Signing for Rulesets

zoracon:
Concept for distributed signing for rulesets by EFF maintainers. Subject to review first before merging. 

Other maintainers are also welcome to give feedback!

Internal documentation will be updated reflecting changes.


utils/sign-rulesets/smartcard-sign.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "{ \"timestamp\": $TIMESTAMP, \"reference\": \"$REFERENCE\", \"rulesets\":" "`cat $RULESETS_FILE`" "}" | tr -d '\n' | gzip -nc > $2/default.rulesets.$TIMESTAMP.gz
+
+echo 'Hash for signing: '
+sha256sum $2/default.rulesets.$TIMESTAMP.gz | cut -f1 -d' '
+
+openssl dgst -sha256 -binary $2/default.rulesets.$TIMESTAMP.gz > default.rulesets.$TIMESTAMP.sha256
+
+pkcs15-crypt -s -k 02 --sha-256 -i default.rulesets.$TIMESTAMP.sha256 -o $2/rulesets-signature.$TIMESTAMP.sha256 -f openssl
+
+openssl dgst -sha256 -verify $1 -signature $2/rulesets-signature.$TIMESTAMP.sha256 $2/default.rulesets.$TIMESTAMP.gz
```
Hainish, line 35: Ideally, we'd like the ruleset signatures to be in the same format as that generated by `async-request.sh` and `standalone.sh`:

https://github.com/EFForg/https-everywhere/blob/3504bf141c178b51ef08c9412886a318df234809/utils/sign-rulesets/standalone.sh#L18

This is so different ruleset signatories can use the same key in different ways.  Your signatures are in RSA PKCS1 v1.5 format, whereas we've been using RSA-PSS format.  Both are supported by the [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign), but unfortunately it looks like Yubikeys only support PKCS1 v1.5 currently (though the higher-end YubiHSM 2 supports PSS as well).

If we are introducing a change that will break interoperability, we should make this explicit.  This would involve adding a field to [`update_channels.js`](https://github.com/EFForg/https-everywhere/blob/master/chromium/background-scripts/update_channels.js) specifying the format rulesets are expected to be signed in, and add logic to verify these signatures according to their expected format.  This will allow SecureDrop and other projects which rely on HTTPS Everywhere for Tor Browser instance discovery to retain their signature algorithm.

Hainish: @redshiftzero would SmartCard signing be useful to SecureDrop?

Hainish: Currently, we're only verifying with RSA-PSS:
https://github.com/EFForg/https-everywhere/blob/3504bf141c178b51ef08c9412886a318df234809/chromium/background-scripts/update.js#L59

One additional option would be instead of specifying the signature algorithm in [`update_channels`](https://github.com/EFForg/https-everywhere/blob/master/chromium/background-scripts/update_channels.js), we instead try to verify first using `RSA-PSS`, and if that fails then trying `RSASSA-PKCS1-v1_5`.  This would definitely be easier to implement, and wouldn't break backwards compatibility, but defeats some of the security properties of `RSA-PSS` and just kind of strikes me as a bad practice.  It may be acceptable given the trade-offs, though.

pipboy96: @Hainish What do you think about using P-256 or P-384 for YubiKey-signed channels? I'm aware that these curves were previously [criticized by DJB](https://safecurves.cr.yp.to/), but I would personally prefer using ECC over `RSASSA-PKCS1-v1_5`, a quite dated signing scheme.

Hainish: @pipboy96 do YubiKeys support signing with these curves?

pipboy96: @Hainish They do.

zoracon: I think it's possible we can do this check on just EFF's rulesets, and non-EFF update channels can stay with RSA-PSS. Preserving the security benefits of RSA-PSS for them. That was my general idea, but I could be missing something on how that could impact Tor or how the update channels are bundled. When I wrote this up I knew it would present a challenge on the update channels, which is why I wanted @Hainish to provide input, `RSASSA-PKCS1-v1_5` isn't broken _yet_, but it is older.

If this presents as a detriment I can just archive this work for other aspects of our code base later. I wanted to at least attempt distributed signing and this was my first demo of that.

pipboy96: @zoracon Are these changes being made for EFF or for SecureDrop?

zoracon: @pipboy96 Meant for EFF, but the change I presented impacts SecureDrop if without the needed modifications to update channels

zoracon: For context @pipboy96 https://securedrop.org/news/introducing-onion-names-securedrop/

zoracon:
Closing out for now, But will archive this PoC in case of future developments

Issue #12118, opened 2017-08-26
List of uncovered domains for potential new rules


ghost:
Can you please add rules for the following websites:
https://www.indeed.co.uk/
https://www.ulster.ac.uk/
https://www.shu.ac.uk/
https://www.hud.ac.uk/
https://www2.mmu.ac.uk/
~~https://solarmoviez.to/~~ gone
~~https://fmovies.is/~~ ~~now https://123movies.com/~~
https://aib.ie/
https://www.bankofireland.com/
https://www.rbs.com/
https://www.ebs.ie/
https://www.kbc.ie/
https://www.home.barclays/
https://www.sc.com/en/
https://www.brownshipley.com/
https://www.northerntrust.com/
https://ria.ru/

PR #19718, opened 2020-10-23
Add Indeed.co.uk.xml

pipboy96:
Related: #12118.


src/chrome/content/rules/Indeed.co.uk.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset name="indeed.co.uk">
```
J0WI, line 1: `blog.indeed.co.uk` (redirects to http)

pipboy96: @J0WI Thanks!

PR #19733, opened 2020-10-28
Create 25thandclement.com.xml

TotalCaesar659:


src/chrome/content/rules/25thandclement.com.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="25thandclement.com">
+	<target host="25thandclement.com" />
+	<target host="www.25thandclement.com" />
```
J0WI, line 3: See also `wilbur.25thandclement.com`

Issue #19826, opened 2020-12-01
http://test-http.surge.sh/ redirect loop


diracdeltas:
http://test-http.surge.sh needs an exclusion because it redirects to HTTP

zoracon:
Please disable HTTPS Everywhere for this site. It will stop the issue.

PR #19827, opened 2020-12-01
Add exclusion pattern to surge.sh

diracdeltas:
Fixes https://github.com/EFForg/https-everywhere/issues/19826

## Type of change

- [ ] New Ruleset
- [ ] Existing Ruleset

## For Rulesets: Testing

- [ ] Travis CI completed without errors (triggered upon pull request)

## Existing Contributors
If you have open PRs that have not been closed or merged, please revisit them by searching: `https://github.com/EFForg/https-everywhere/issues?q=is%3Aopen+assignee%3A[yourusernamehere]`.

### DIFF
--- a/src/chrome/content/rules/Surge.sh.xml
+++ b/src/chrome/content/rules/Surge.sh.xml
@@ -8,7 +8,10 @@
 		<test url="http://web-cdn.surge.sh/" />
 		<test url="http://ioniclib.surge.sh/docs/" />
 
-
+  <!--
+    This is an HTTP test page which intentionally redirects to HTTP.
+  -->
+  <exclusion pattern="^http://test-http\.surge\.sh/$" />
 	<rule from="^http:"
 		to="https:" />
 


CrazyPython:
`test-http.surge.sh` is only intended to demo how a `surge.sh` site can be broken.

The entire surge.sh domain should be removed, except for an allowlist of potential domains that do want HTTPS.

ryanbr:
@CrazyPython  Maybe a good example to work around an infinite loop, so expiring the redirections after X attempts.

cschanaj:
> Maybe a good example to work around an infinite loop, so expiring the redirections after X attempts.

The website `test-http.surge.sh` is a typical bad HTTPS implementation that HTTPS Everywhere failed to address

1. use JavaScript to rewrite the HTTPS protocol to HTTP 
2. send `200` on both HTTP and HTTPS connections 
3. same content on both HTTP and HTTPS connections

(1) and (2) basically means we cannot detect the redirect by the HTTP status code or by logging the same requestId in webextensions. 

(3) however, I don't think there is a way to keep the functional HTTPS pages using current webext API.

```javascript
<script>
if (window.location.protocol === "https:") {
    window.location.href = 'http:' + window.location.href.slice().slice(6);
}
</script>
``` 

A possible solution is to cancel a redirect (allowlist a domain) to the same URLs visited on the main frame after several attempts within a certain period of time. Given that modern browsers are rolling out their own HTTPS rewriter, should we add this new feature? cc @zoracon @J0WI @pipboy96

src/chrome/content/rules/Surge.sh.xml
```diff
@@ -11,5 +11,6 @@
 
 	<rule from="^http:"
 		to="https:" />
+  <exclusion pattern="http://test-http\.surge\.sh/" />
```
J0WI, line 14:
```suggestion
  <exclusion pattern="^http://test-http\.surge\.sh/$" />
```

Please move the exclusion above the rule and add a comment documenting the reason.

J0WI:
> A possible solution is to cancel a redirect (allowlist a domain) to the same URLs visited on the main frame after several attempts within a certain period of time. Given that modern browsers are rolling out their own HTTPS rewriter, should we add this new feature?

From the very beginning until the transition to WebExtensions API HTTPS Everywhere 2e754f2a53bae907d1621f323dd769c5e6be6ee used to have such a heuristic. IMHO the use case is very limited nowadays.

src/chrome/content/rules/Surge.sh.xml
diracdeltas: Done!

pipboy96:
@J0WI An unrelated question, why do you use `window.location.href.slice().slice(6)` instead of simply using `window.location.href.slice(6)` here?

src/chrome/content/rules/Surge.sh.xml
J0WI: Ah sorry, every exclusion needs an explicit test url.

PR #19781, opened 2020-11-14
Open option page in a new tab

cschanaj:
Related #19780 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring existing code
- [ ] New Ruleset
- [ ] Existing Ruleset

## For Rulesets: Testing

- [x] Travis CI completed without errors (triggered upon pull request)

## Existing Contributors
If you have open PRs that have not been closed or merged, please revisit them by searching: `https://github.com/EFForg/https-everywhere/issues?q=is%3Aopen+assignee%3A[yourusernamehere]`.

PR #19844, opened 2020-12-10
Introduce Modular CSS

zoracon:
Related #19781

## Type of change
- [x] Refactoring existing code

For more unity in the UI, using a base templates method that can be imported and used in different pages of the extension


zoracon:
Merging to attend to related PR

PR #19860, opened 2020-12-17
Create sunxdcc.com.xml

TotalCaesar659:
## Type of change

- [x] New Ruleset


zoracon:
Incorrect file path for rule. Please amend to be under `src/chrome/content/rules`

zoracon (changes requested):
Incorrect file path for rule. Please amend to be under `src/chrome/content/rules`

J0WI (approved):
LGTM

PR #19869, opened 2020-12-19
Update rewriter script

ypid:
## Type of change

- [ ] New Ruleset
- [ ] Existing Ruleset
- [x] Code maintenance

### DIFF
--- a/utils/rewriter/rewriter.js
+++ b/utils/rewriter/rewriter.js
@@ -25,10 +25,17 @@ var ruleSets = null;
 function processDir(dir) {
   var stream = readdirp({
     root: dir,
-    fileFilter: ['*.html', '*.js', '*.rb', '*.erb', '*.mustache',
-      '*.scala', '*.c', '*.cc', '*.cpp', '*.cxx',
-      '*.java', '*.go', '*.php', '*.css', '*.pl', '*.py',
-      '*.rhtml', '*.sh', '*.yaml']
+    directoryFilter: [
+      '!.*', // Protect `.git` and other VCS directories.
+      '!*modules',
+      '!LICENSES', // https://reuse.software/ convention.
+    ],
+    fileFilter: [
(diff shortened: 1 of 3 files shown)


utils/rewriter/rewriter.js
```diff
@@ -37,10 +25,17 @@ var ruleSets = null;
...
+      '!.*', // Protect `.git` and other VCS directories.
+      '!*modules',
+      '!LICENSES', // https://reuse.software/ convention.
+    ],
+    fileFilter: [
+      '!*.png', '!*.svg', '!*.ico', '!*.xcf',
+      '!*.pdf',
+
+      '!LICENSE*', '!COPYING*', // Prefer https://reuse.software/.
+    ],
```
J0WI, line 38: I'd still prefer an allowlist to avoid any binary file. E.g. JPEG is also missing from the list.

ypid: I am not really happy with that either. My idea was that I did not want to miss files. As I wrote, after the rewrite the diff will have to be reviewed anyway.

What do you think of using this https://github.com/gjtorikian/isBinaryFile instead?

0xACE: Drop it. https://github.com/gjtorikian/isBinaryFile is fruitless. Stick to a white list and use this tool carefully if you intend to use it as a sanitizer... I didn't review your code, but I don't understand why'd you remove the only documentation on top...

ypid: > I didn't review your code, but I don't understand why'd you remove the only documentation on top...

I moved it to README.md so that it will be shown. Preview: https://github.com/ypid/https-everywhere/tree/fix/rewriter/utils/rewriter

zoracon (approved):

ypid:
@zoracon You approved this but closed it without merging?

PR #19910, opened 2021-01-30
Adding bloom-filter based redirects

Hainish:


pipboy96:
@Hainish There's a slight problem with the code. You do not check the rulesets, which you need to, since bloom filters have false positives. (for example https://github.com/EFForg/https-everywhere-lib-core/master/tests/hosts.bf matches `91.com`. Nevertheless, this is a good step forward. Thanks for your work.

cschanaj:
There was some discussions on using bloom filters to improve lookup and memory usage in https://github.com/EFForg/https-everywhere/pull/12160#issuecomment-326167967 back when 2017. Some of the issues with bloom filters can be briefly concluded as follows:

1. false positives
2. lookup performance
3. implementation

For (1) and (2), because bloom filters are probabilistic, false positives need to be verified by performing a secondary/ teritary lookup. Personally, I prefer to use (a) multiple bloom filters (with different hash functions) to reduce the collision rate. We should also make sure there is no collision generated in the top-1m list before each release. This adds a benefit of reducing the bundled rulesets size significantly by taking away the targets with trivial rules. It also reduce the bandwidths used to deliver ruleset updates. (b) using ES6 `Set()` for simplicity and quick implementation, which the benefits are obvious. 

For (3), any chance there will be a JavaScript implementation of the bloom filters lookup?  

Apart from the issues discussed earlier, it will not be easy to disable a particular rule when the users encounter broken websites in the bloom filters list. Does it also means that we will expect the users to _disable the extension permanently on broken webpages (via the extension popup)_?

Hainish:
@pipboy96 

> @Hainish There's a slight problem with the code. You do not check the rulesets, which you need to, since bloom filters have false positives. (for example https://github.com/EFForg/https-everywhere-lib-core/master/tests/hosts.bf matches `91.com`). Nevertheless, this is a good step forward. Thanks for your work.

The false positivity rate is parameterized at the point of generation of the filter.  See https://github.com/Hainish/create-bloom-filter
In this test case, I chose a false positivity rate of 1 in 100.  This is to keep the test file size at a minimum.  With the production bloom filters, we should

1. use a low false positivity rate (1 in 10 million, for instance),
2. calculate the false positives as much as it's feasible and create a negative bloom filter that we also check against, and
3. inform users of the ability to 'disable on this site' when a bloom upgrade occurs but fails.

This will make encountering a false positive in real world scenarios very unlikely, and gives recourse to users.

@cschanaj I hadn't planned on implementing this in JS.  The major constituency that still has WebAssembly disabled is Tor Browser users.  From discussions with Tor Browser core developers, we're under the impression that they will turn on HTTPS Only mode in the near future.  The way this is currently implemented, if a `"format": "bloom"` update channel is included in a browser when WASM is disabled, the bloom filters will be downloaded and saved, but the bloom objects will not be created by the binding generator, which is compiled into WASM.  This will result in a safe fallback, but if a user turns WASM back on, the bloom objects will be successfully created, no harm done.

Hainish:
This was tested with the following `update_channels.js` file:

```
/* exported update_channels */

'use strict';

(function (exports) {

exports.update_channels = [{
  name: 'Bloom Test',
  format: 'bloom',
  jwk: {"kty":"RSA","n":"DZwlYjrh6dzQADYJD8wkREEyIbyhi4I4pT45MK2MZsXqxk-1ezcYBd3yfPojGgLu6CBxj0cIrQVbiZMHwN4A1rG5VPx2jUuNt-LYMbuBfwTS6ziTdjfHHpq4ZqttrMWUR1BH44C4OHaN9XW5NjWiIMxzWhoMrPz53yLAAR-9ltPjZmtmekXN-EUnim0SAoGGqNX41rIbjTGrN_jo-wqLri1-yINNFMQ0WH901kIGEqKNQk_LjLmwbO8c5L45U4NJ-D1lJgixVa5LeqdynWOA9FKUof3ITzUi1FzORoLo0kr2sR0YyhA1UiA9aag0uyHH5B1XmQZxD2SMlBrrbNxh2CNZqq1V9_vLWdHiTVPidST6QZDpNLj2mG9zJkJnslt0BzoxSY8-J0SbWfk-rwBOptiPeWWFEhQsa5zSQFQGlOiqF66FG-Tof70y-WkEFNh-datXa6rty4Zdf-WvWzqBVKodzdP7qJE-YD6pXQJ4syrwlyB_g9eMGy0YyGQ3TxtWpYd_ZTIT1Hc1slhrlmJYmUZ3AeIt1GvN322Cq9XN6YhBnohKa7phj-8LFYtZ3zXDUQzXSjRQrOBz49NrQYX2HKgnY2zjCL_lvG3ZzNUfqJofMHR1gruWSk8-ZodMX0B8MM7EWD4kK3SOcUnppR0-Lo4QjRgDM7fLJtH7R7WEZUM","e":"AQAB"},
  update_path_prefix: 'https://www.https-rulesets.org/test/bloom',
}];

})(typeof exports === 'undefined' ? require.scopes.update_channels = {} : exports);
```

Hainish:
You can now add & delete bloom format update channels in the options UX

zoracon:
After further testing and review, this change is ready and will be utilized very soon for upcoming ruleset changes. More information will expand on this in the near future. Thank you for doing this @Hainish.

zoracon (approved):

PR #19943, opened 2021-02-23
Create uplus.co.kr.xml

alexkoala:
## Type of change

- [x] New Ruleset
- [ ] Existing Ruleset

### DIFF
--- a/src/chrome/content/rules/LG-Uplus.xml
+++ b/src/chrome/content/rules/LG-Uplus.xml
@@ -1,0 +1,10 @@
+<ruleset name="LG Uplus">
+	<target host="uplus.co.kr"/>
+	<target host="www.uplus.co.kr"/>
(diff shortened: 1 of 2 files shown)


zoracon (commented):
There are numerous subdomains associated with this domain. If you could address this with a [wildcard](https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#wildcards-in-targets) rule that would be optimal. Or, if you discover that a lot these subdomains do not support HTTPS, you can note that as well in a comment. I often ignore the dev.* and qa.* domains and focus on the public facing resources for typical site visitors.

I ran an enumerator for subdomains. Here is an example tool here: https://github.com/EFForg/https-everywhere/blob/master/CONTRIBUTING.md#enumerating-subdomains

PR #19957, opened 2021-03-03
Update WordPress.xml: don't secure wordpress_logged_in cookie

### LINKED ISSUE #19954: onCookieChange overwrites cookies if both secure and non-secure one are sent
Type: code issue/website issue

When logging into the WordPress.com service, the login endpoints sends two variants of the `wordpress_logged_in` session cookie:
```
Set-Cookie: wordpress_logged_in=secret
Set-Cookie: wordpress_logged_in=secret; secure=true; SameSite=None
```

The goal is to have valid session cookie for all the following cases:
- the site detects a logged-in user when accessed over an unencrypted `http://` URL. There are still rare cases that need this.
- the site can load authenticated iframes or resources (images, videos) from another domain (CDN-like). It needs the `SameSite=None` attribute to support that usage.

Cookies with `SameSite=None` also need a `secure=true` attribute. Otherwise they would be rejected by the browser. That's why the non-secure cookie doesn't and can't have that attribute.

Now the HTTPS Everywhere extension automatically converts all incoming cookies to `secure=true`. The result is that a secure `wordpress_logged_in` cookie _without_ the `SameSite` attribute is stored and used. The second one, with `secure=true` and `SameSite=None`, is overwritten/ignored.

That breaks the cross-domain authenticated requests where the `SameSite=None` attribute is required.

Is there a way around that? Can we specify a wordpress.com ruleset that fixes the issue just for the one site? Or can `onCookieChange` be patched to cover this case?

jsnajdr:
The wordpress.com site is already sending both secure and non-secure versions.
The non-secure version doesn't have a SameSite=None attribute, and securing it
would overwrite the secure cookie that has it.

Fixes #19954

## Type of change

- [ ] New Ruleset
- [x] Existing Ruleset

### DIFF
--- a/src/chrome/content/rules/WordPress.xml
+++ b/src/chrome/content/rules/WordPress.xml
@@ -28,7 +28,7 @@
 		<test url="http://www.wp.com/music" />
 		<test url="http://get.wp.com/" />
 
-	<securecookie host=".+" name=".+" />
+	<securecookie host=".+" name="^(?!wordpress_logged_in$).+" />
 
 	<rule from="^http:" to="https:"/>
 </ruleset>


zoracon (approved):

zoracon:
I can run a ruleset release later today to reflect these changes for WordPress

Issue #18951, opened 2020-02-18
Shell scripts, hashbangs and docs require python3.6


jayvdb:
[//]: # (Thank you for reporting an issue to HTTPS Everywhere.)
[//]: # (We welcome input from users on improving this project.)
[//]: #
[//]: # (Please help us by following this issue template.)
[//]: # (You can delete all blank lines and all lines starting)
[//]: # (with the comment marker, such as this one.)

Type: code issue

[//]: # (Include any other relevant information below. Thank you again for)
[//]: # (helping to improve HTTPS Everywhere.)

Shell scripts like the test coverage scripts explicitly refer to `python3.6`, which is increasingly not present on Linux distros which often provide python3.7 or python3.8 now.  Also iirc, Python installers for Windows do not provide Python3.6.exe/cmd , and may not even provide a Python3.exe - should re-check that, as Windows users should be able to help with issues like https://github.com/EFForg/https-everywhere/issues/18745 , and docs saying `python3.6` is going to confuse them.

`git grep python3.6` has 48 results; very varied.

Suggest using just `python3` now, and check the minimum supported version inside the scripts, especially any which dont work on python3.4+ which is a really safe minimum, and provide a useful error message if an insufficient version is installed.

cschanaj:
As per https://trac.torproject.org/projects/tor/ticket/29097

> ... which ensures that Firefox builds are deterministic, is derived from the Python 3.6 standard library. We've targeted 3.6 for this reason, as well as to enforce similar tooling across different contributor development environments.

In https://github.com/EFForg/https-everywhere/issues/6749#issuecomment-363281281, @Hainish explained the rationale behind hard-coding Python 3.6 dependency approximately 2 years ago, which seems unfit for this moment when Python 2.7 is obsolete and Ubuntu 18.04 LTS is released. 

We use the `bionic` release for Travis CI and docker image. I think it is fine to use ```#!/bin/usr/env python3``` as the Ubuntu LTS ships with Python 3.6 by default if the builds remains deterministic. If possible, we proceed with #9993 in the long term.

zoracon:
I have been slowly moving into using Python3 instead. This has been on the backlog as a ticket so I don't lose sight of this.

kristapsk:
I also don't have Python 3.6 on my laptop anymore, so `test/firefox.sh` don't work. Also, Python 3.6 has EOL in 9 months (23 Dec 2021).

zoracon:
[...]

PR #19987, opened 2021-03-24
[REFACTOR] Move to Python 3

### LINKED ISSUE #19023: Update install-dev-dependencies.sh to work with Debian Buster

zoracon:
Python3.6 causing multiple issues with newer Ubuntu versions >= 18.04

Fixes #19023, #18951

## Type of change

- [x] Refactoring existing code


cschanaj:
The docker image is still hard-coding python3.6 (https://github.com/EFForg/https-everywhere-docker-base/blob/76634d212acc5f40ee9a5c4e01758173d2ef9408/Dockerfile#L9), IMHO it should be moved to python3 too.

zoracon:
@cschanaj Thanks for catching that. Will update those as well

PR #19994, opened 2021-03-28
Added rules for allegrolokalnie.pl

Teeed:
## Type of change
- [x] New Ruleset


src/chrome/content/rules/Allegrolokalnie.pl.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="AllegroLokalnie.pl">
+	<target host="allegrolokalnie.pl" />
+	<target host="www.allegrolokalnie.pl" />
```
J0WI, line 3: `zobacz.allegrolokalnie.pl`

Teeed: Thanks, added that :)

PR #19999, opened 2021-04-01
Create emule-rus.net.xml

TotalCaesar659:
## Type of change

- [x] New Ruleset


src/chrome/content/rules/emule-rus.net.xml
```diff
@@ -0,0 +1,16 @@
+<!--
+	Invalid certificate:
+		ftp.emule-rus.net
+		webmail.emule-rus.net
+
+	Redirect to HTTP:
```
J0WI, line 6: Cannot reproduce

PR #20010, opened 2021-04-08
Added rsbsrt.ru

KOLANICH:
## Type of change

- [x] New Ruleset

### DIFF
--- a/src/chrome/content/rules/rsbsrt.ru.xml
+++ b/src/chrome/content/rules/rsbsrt.ru.xml
@@ -1,0 +1,7 @@
+<ruleset name="rsbsrt.ru">
+	<target host="www.rsbsrt.ru" /> <!-- mismatch, wrong content -->
+	<rule from="^http://www\.rsbsrt\.ru/" to="https://rsbsrt.ru/" />
+
+	<target host="rsbsrt.ru" />
+	<rule from="^http:" to="https:" />
+</ruleset>


src/chrome/content/rules/rsbsrt.ru.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="rsbsrt.ru">
+	<target host="rsbsrt.ru" />
```
J0WI, line 2: `www.rsbsrt.ru` (mismatch, wrong content)

KOLANICH: Done

src/chrome/content/rules/rsbsrt.ru.xml
```diff
@@ -0,0 +1,6 @@
...
+	<target host="www.rsbsrt.ru" /> <!-- mismatch, wrong content -->
+
+	<rule from="^http://(www\.)?" to="https://" />
```
J0WI, line 5: Invalid regex: https://travis-ci.com/github/EFForg/https-everywhere/jobs/499842904#L332-L335

KOLANICH: >missing caret (^) in 'from' attribute: it should be "^http:" and not "http:"

Obviously wrong, the caret is present, it is obvious.

>missing trailing slashes in 'from' or 'to' when specifying full hostnames: 
it should be "https://eff.org/" and not "https://eff.org"

Not needed here.

The regex is valid and OK, it is the static analyser used flawed. Please don't call it invalid anymore, it is the static analyser that is invalid.

PR #20012, opened 2021-04-08
Added www.opensource.org

KOLANICH:
Fixes #(issue)

## Type of change

- [x] Existing Ruleset


src/chrome/content/rules/Open_Source.org.xml
```diff
@@ -5,6 +5,7 @@
 <ruleset name="Open Source.org (partial)">
 
 	<target host="opensource.org" />
+	<target host="www.opensource.org" />
```
cschanaj, line 9: Would you please 

- [ ] Add `lists.opensource.org`
- [ ] Remove the obsolete top comment
- [ ] Delete the ruleset file `Open_Source.org-problematic.xml`

cschanaj (changes requested):
Changes required.

src/chrome/content/rules/Open_Source.org.xml
KOLANICH: Done.

cschanaj (approved):

PR #20013, opened 2021-04-08
Added valgrind.org

KOLANICH:
## Type of change

- [x] New Ruleset


src/chrome/content/rules/valgrind.org.xml
```diff
@@ -0,0 +1,6 @@
+<ruleset name="valgrind.org">
```
J0WI, line 1: non-functional `njn.valgrind.org` and `svn.valgrind.org`

KOLANICH: Done.

PR #20014, opened 2021-04-08
Added candelatech.com

KOLANICH:
## Type of change
- [x] New Ruleset

### DIFF
--- a/src/chrome/content/rules/candelatech.com.xml
+++ b/src/chrome/content/rules/candelatech.com.xml
@@ -1,0 +1,7 @@
+<ruleset name="candelatech.com">
+	<target host="candelatech.com" />
+	<rule from="^http://candelatech\.com/" to="https://www.candelatech.com/" /> <!-- mismatch -->
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/candelatech.com.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="candelatech.com">
+	<target host="candelatech.com" />
```
J0WI, line 2: mismatch => `www`

J0WI (changes requested):

src/chrome/content/rules/candelatech.com.xml
```diff
@@ -0,0 +1,7 @@
...
+	<rule from="^http://candelatech\.com" to="https://www.candelatech.com" /> <!-- mismatch -->
```
J0WI, line 3:
```suggestion
	<rule from="^http://candelatech\.com/" to="https://www.candelatech.com/" /> <!-- mismatch -->
```

PR #20019, opened 2021-04-08
Added libusb.info

KOLANICH:
## Type of change

- [x] New Ruleset

### DIFF
--- a/src/chrome/content/rules/libusb.info.xml
+++ b/src/chrome/content/rules/libusb.info.xml
@@ -1,0 +1,13 @@
+<ruleset name="www.libusb.info">
+
+	<!-- Non-functional
+	api.libusb.info
+	log.libusb.info
+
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/libusb.info.xml
```diff
@@ -0,0 +1,5 @@
+<ruleset name="www.libusb.info">
+	<target host="www.libusb.info" />
+	<target host="libusb.info" />
+	<rule from="^http://(www\.)?" to="https://" />
```
J0WI, line 4: This syntax is invalid.

Please also add a comment for the non functional hosts `api.libusb.info` and `log.libusb.info` on the top.

KOLANICH: https://github.com/libusb/libusb/issues/910

PR #20030, opened 2021-04-08
No longer different content on git.infradead.org

KOLANICH:
## Type of change
- [x] Existing Ruleset


src/chrome/content/rules/Infradead.org.xml
```diff
@@ -51,17 +52,16 @@
 
 		Different content:
 			 - ftp.infradead.org
-			 - git.infradead.org
 -->
 <ruleset name="Infradead.org (partial)">
 	<target host="www.infradead.org" />
 	<target host="bombadil.infradead.org" />
```
J0WI, line 58:
```suggestion

```

PR #20036, opened 2021-04-08
Added more domains to funet.fi

KOLANICH:
Fixes #(issue)

## Type of change

- [x] Existing Ruleset

### DIFF
--- a/src/chrome/content/rules/Funet.fi.xml
+++ b/src/chrome/content/rules/Funet.fi.xml
@@ -2,8 +2,13 @@
 
 	<!--	Direct rewrites:
 				-->
+	<target host="www.funet.fi" />
 	<target host="haka.funet.fi" />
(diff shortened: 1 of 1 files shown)


src/chrome/content/rules/Funet.fi.xml
```diff
@@ -2,8 +2,11 @@
 
 	<!--	Direct rewrites:
 				-->
+	<target host="www.funet.fi" />
 	<target host="haka.funet.fi" />
 	<target host="haka-ds.funet.fi" />
+	<target host="nic.funet.fi" />
+	<target host="ftp.funet.fi" />
```
J0WI, line 9: `www.ftp.funet.fi`
`www.nic.funet.fi`

KOLANICH: Done.

PR #20046, opened 2021-04-13
Fix invalid ruleset scope typings

cschanaj:
This PR makes compiling `scope` into `RegExp` automatic. Pass the string literal version of the `scope` parameter everywhere, except for when constructing rulesets/ ~WASM calls~. 

- `getScope` expected a required string parameter, but get none. It is also called multiple time when passed to `addFromJson`

## Type of change

- [ ] New Ruleset
- [ ] Existing Ruleset

### DIFF
--- a/chromium/background-scripts/rules.js
+++ b/chromium/background-scripts/rules.js
@@ -17,8 +17,8 @@ const trivial_cookie_rule_c = /.+/;
 const scopes = new Map();
 
 /**
- * Returns the scope object from the map for the given scope string
- * @param {string} scope
+ * Compile literal ruleset scope into RegExp
+ * @param {string} scope literal ruleset scope
  * @returns {RegExp}
  */
 function getScope(scope) {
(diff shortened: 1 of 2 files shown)


chromium/background-scripts/rules.js
```diff
@@ -232,26 +241,33 @@ RuleSets.prototype = {
...
   addFromJson: function(ruleJson, scope) {
-    const scope_obj = getScope(scope);
-
     if (this.wasm_rs) {
       this.wasm_rs.add_all_from_js_array(
         ruleJson,
         settings.enableMixedRulesets,
         this.ruleActiveStates,
-        scope);
+        getScope(scope));
```
Hainish, line 254: Why change this string to a scope in JavaScript?  The scope will be cached when passed to Rust at any rate.  It looks like this passes a RegExp to the Rust bindings, which it isn't expecting.

Hainish (changes requested):
See comment

chromium/background-scripts/rules.js
cschanaj: Look like a mistake made when I was trying to figure out the types of `scope` in different calls. The original source code uses string `scope` too. It should be a string. Changes made in f87434f

As the `getScope` function is being called in `Ruleset` constructor exclusively, I added 1c20e82 to reflect this. Only the internal representation of the `scope` is a `RegExp` now. Thanks.

Hainish:
LGTM, thanks!

PR #20071, opened 2021-05-10
Remove Cloudflare __cfduid cookie configs

migueldemoura:
This cookie is no longer set for any requests proxied via Cloudflare and
thus the configs that were added for it can be removed.

Further context from Cloudflare's blog:
https://blog.cloudflare.com/deprecating-cfduid-cookie/

---

Not sure if the `name` attribute of `<securecookie>` always has to be a regex, but if not then you could simplify some of the expressions down to a straight strncmp.

## Type of change

- [ ] New Ruleset
- [x] Existing Ruleset


zoracon:
@migueldemoura Yes, go ahead and remove those domains from the rule. The PR should build fine afterwards.

migueldemoura:
Thanks @zoracon, I've removed the `placehold.it` config file.

zoracon (approved):

PR #20096, opened 2021-07-04
Add bitcoinkpis.com

kristapsk:
## Type of change

- [X] New Ruleset


zoracon:
We have stopped accepting new rulesets as of May 31st, 2021. Please see: https://github.com/EFForg/https-everywhere/discussions/20099. 

We will still accept requests on rulesets already in our list that are causing significant breakage for users through the summer. However in autumn, we will begin the plan to ultimately move our crowdsourced rulesets out of the extension.

